PR #6051, opened 2021-03-09
Fixes #6049 - Default provider [files] section always executed

### LINKED ISSUE #6049: Default provider [files] section always executed
**Jetty version**
10.0.x

**Description**
`java -jar start.jar --add-modules=http,logging-logback` copies `resources/jetty-logging.properties`, but it should not.

Interestingly, 
`java -jar start.jar --add-modules=logging-logback,http` does *not* copy `resources/jetty-logging.properties`.

Since an explicit provider has been configured, the default provider should be out of the picture.

sbordet:
Keeping only enabled modules when processing the modules,
so that default provider modules don't get processed.

### DIFF
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/BaseBuilder.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/BaseBuilder.java
@@ -187,6 +187,7 @@ public class BaseBuilder
                 DirectoryStream.Filter<Path> filter = new DirectoryStream.Filter<Path>()
                 {
                     PathMatcher iniMatcher = PathMatchers.getMatcher("glob:**/start.d/*.ini");
+
                     @Override
                     public boolean accept(Path entry) throws IOException
                     {
@@ -195,7 +196,9 @@ public class BaseBuilder
                 };
                 List<Path> paths = new ArrayList<>();
                 for (Path path : Files.newDirectoryStream(startd, filter))
+                {
                     paths.add(path);
+                }
                 paths.sort(new NaturalSort.Paths());
 
                 // Read config from start.d
@@ -212,12 +215,16 @@ public class BaseBuilder
                 try (FileWriter out = new FileWriter(startini.toFile(), true))
                 {
                     for (String line : startLines)
+                    {
                         out.append(line).append(System.lineSeparator());
+                    }
                 }
 
                 // delete start.d files
                 for (Path path : paths)
+                {
                     Files.delete(path);
+                }
                 Files.delete(startd);
             }
         }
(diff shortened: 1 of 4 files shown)


jetty-start/src/main/java/org/eclipse/jetty/start/BaseBuilder.java
```diff
@@ -264,56 +271,65 @@ public boolean accept(Path entry) throws IOException
             StartLog.warn("Use of both %s and %s is deprecated", getBaseHome().toShortForm(startd), getBaseHome().toShortForm(startini));
 
         builder.set(useStartD ? new StartDirBuilder(this) : new StartIniBuilder(this));
-        newlyAdded.stream().map(modules::get).forEach(module ->
-        {
-            String ini = null;
-            try
+        newlyAdded.stream()
```
gregw, line 277: This method is a monster with lots of complex conditionals and loops.
I think perhaps now we probably need to split and/or better document each section of this method.

jetty-start/src/main/java/org/eclipse/jetty/start/BaseBuilder.java
```diff
@@ -264,56 +271,65 @@ public boolean accept(Path entry) throws IOException
...
+            .map(modules::get)
+            .filter(Module::isEnabled)
+            .forEach(module ->
             {
-                if (module.isSkipFilesValidation())
-                {
-                    StartLog.debug("Skipping [files] validation on %s", module.getName());
-                }
-                else
+                StartLog.debug("%s.isEnabled() = %b", module.getName(), module.isEnabled());
```
gregw: don't need to log isEnabled, as it is always true

gregw (dismissed):
I think the change is good, but I also think we need to better describe the algorithm in the build method so we can actually be sure the change is good.

jetty-start/src/main/java/org/eclipse/jetty/start/BaseBuilder.java
joakime: But not for this PR, a new PR for that is a better choice.

gregw: Then this PR at least needs more comments to explain how/why it fixes the problem

gregw:
@sbordet what is the status of this?

gregw:
@sbordet I'm going to remove my review, so you can fix my comments and get somebody else to re-review in your timezone rather than wait 24 hours for me.

sbordet:
@gregw @joakime nudge

joakime (approved):

gregw (approved):

PR #6053, opened 2021-03-10
Issue #6052 - make ModuleLocation optional on Android

### LINKED ISSUE #6052: Cleanup TypeUtil and ModuleLocation to allow jetty-client/hybrid to work on Android
**Jetty version**
9.4.x

**Java version**
Android

**OS type/version**
Android

**Description**
When attempting to use `jetty-client-<ver>-hybrid.jar` on Android.
The `TypeUtil` has an optional dependency on `ModuleLocation` which uses `MethodHandle.invoke` and `MethodHandle.invokeExact`, both of which are not supported by Android (yet)

Attempt to make the use of `ModuleLocation` in a more optional way.

joakime:
Change how `ModuleLocation` is loaded by `TypeUtil` to allow it be entirely optional.

Also, change `jetty-client-<ver>-hybrid.jar` to generate a minimized jar, which will only include the dependent classes that are actually used, not all classes from http/io/util.

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/TypeUtil.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/TypeUtil.java
@@ -19,9 +19,6 @@
 package org.eclipse.jetty.util;
 
 import java.io.IOException;
-import java.lang.invoke.MethodHandle;
-import java.lang.invoke.MethodHandles;
-import java.lang.invoke.MethodType;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
@@ -42,8 +39,6 @@ import java.util.function.Function;
 
 import org.eclipse.jetty.util.log.Log;
 import org.eclipse.jetty.util.log.Logger;
-
-import static java.lang.invoke.MethodType.methodType;
 
 /**
  * TYPE Utilities.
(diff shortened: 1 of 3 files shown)


jetty-util/src/main/java/org/eclipse/jetty/util/TypeUtil.java
```diff
@@ -177,38 +173,32 @@
...
+    private static final Function<Class<?>, URI> MODULE_LOCATION;
 
     static
     {
-        List<MethodHandle> locationMethods = new ArrayList<>();
-
-        MethodHandles.Lookup lookup = MethodHandles.lookup();
-        MethodType type = methodType(URI.class, Class.class);
-
+        // Lookup order in LOCATION_METHOD is important.
```
sbordet: Typo: METHOD*S*.

jetty-util/src/main/java/org/eclipse/jetty/util/TypeUtil.java
```diff
@@ -177,38 +173,32 @@
...
+        LOCATION_METHODS.add(TypeUtil::getCodeSourceLocation);
+        Function<Class<?>, URI> moduleFunc = null;
         try
         {
-            locationMethods.add(lookup.findStatic(TypeUtil.class, "getCodeSourceLocation", type));
-            ModuleLocation moduleLocation = null;
-            try
+            Class<?> clazzModuleLocation = Class.forName(TypeUtil.class.getPackage().getName() + ".ModuleLocation");
```
sbordet: I would be stricter here about the class loader.
Use this instead:

```suggestion
            Class<?> clazzModuleLocation = TypeUtil.class.getClassLoader().loadClass(TypeUtil.class.getPackage().getName() + ".ModuleLocation");
```

sbordet (changes requested):

sbordet (approved):

PR #6055, opened 2021-03-10
add an easy way to configure useInputDirectByteBuffers/useOutputDirectByteBuffers for HttpConfiguration

olamy:


jetty-server/src/main/config/modules/server.mod
```diff
@@ -91,3 +91,7 @@ etc/jetty.xml
 # jetty.scheduler.name=
 # jetty.scheduler.deamon=false
 # jetty.scheduler.threads=-1
+
+# Whether to use direct ByteBuffers for reading or writing
+jetty.httpConfig.useInputDirectByteBuffers=true
+jetty.httpConfig.useOutputDirectByteBuffers=true
```
lachlan-roberts: These should be commented out like the rest of the lines, and description should probably start with two `##` like the rest of them.

Maybe this should also be put directly under the other `jetty.httpConfig` setters.

olamy: correct. fixed (arghh we don't have checkstyle configured for mod file 😄)

lachlan-roberts (approved):
LGTM

PR #6068, opened 2021-03-17
jetty 9.4.x hazelcast remote distribution test

olamy:


janbartel:
@olamy  can you add an explanation of why these changes are necessary? Particularly as we already have tests for embedded and remote hazelcast.

tests/test-distribution/src/test/resources/session-store-hazelcast-remote.ini
```diff
@@ -0,0 +1,16 @@
...
+# Enables session data store in a remote Hazelcast Map
+# ---------------------------------------
+jetty.webapp.addSystemClasses+=,org.eclipse.jetty.hazelcast.
+jetty.webapp.addServerClasses+=,-org.eclipse.jetty.hazelcast.
+
+--module=session-store-hazelcast-remote
+
+#jetty.session.hazelcast.mapName=jetty-distributed-session-map
+jetty.session.hazelcast.hazelcastInstanceName=JETTY_DISTRIBUTED_SESSION_INSTANCE
+jetty.session.hazelcast.onlyClient=false
```
joakime: @janbartel this is the reason this whole effort got started.

If `jetty.session.hazelcast.onlyClient=false` then we have classloader issues from within hazelcast.

janbartel: Can you clarify: if using hazelcast-embedded module (where onlyClient is false by default) does this problem occur or not?

olamy:
@janbartel AFAICT we do not have test for remote hazelcast :) and not even for distribution. (as I remember it's because of this https://github.com/hazelcast/hazelcast/issues/17114) I started long time ago a branch called `jetty-9.4.x_hazelcast_remote_testing` (just created a PR here https://github.com/eclipse/jetty.project/pull/6075 )

tests/test-distribution/src/test/resources/session-store-hazelcast-remote.ini
janbartel: OK, I see what is happening. 

The test is setting the SessionSerializer in the _hazelcast_ config file so it is _hazelcast_ that is trying to load the SessionSerializer from the webapp classloader.  The way we recommend and have tested with is to let jetty set up the SessionSerializer in the HazelcastSessionDataStoreFactory, in which case there is no classloading issue.

I guess there is no harm in adding these 2 lines to the .mod file to support the usecase illustrated by the test.

joakime: Yes, I agree, these 2 lines to adjust the webapp classloader should be in the mod file.

olamy:
actually depends on question posted here https://github.com/hazelcast/hazelcast/issues/18508

tests/test-distribution/src/test/resources/session-store-hazelcast-remote.ini
olamy: fixed

tests/test-distribution/src/test/resources/jetty-logging.properties
```diff
@@ -1,2 +1,4 @@
 org.eclipse.jetty.util.log.class=org.eclipse.jetty.util.log.StdErrLog
 #org.eclipse.jetty.LEVEL=DEBUG
+
+org.eclipse.jetty.tests.distribution.LEVEL=DEBUG
```
janbartel: Probably don't want to commit that at DEBUG

tests/test-webapps/test-simple-session-webapp/src/main/java/org/eclipse/jetty/test/session/Chocolate.java
```diff
@@ -0,0 +1,35 @@
...
+//
+//  You may elect to redistribute this code under either of these licenses.
+//  ========================================================================
+//
+
+package org.eclipse.jetty.test.session;
+
+import java.io.Serializable;
+
+public class Chocolate implements Serializable
```
janbartel, line 23: I approve of the name of this class :)

tests/test-webapps/test-simple-session-webapp/src/main/webapp/index.jsp
```diff
@@ -0,0 +1,5 @@
+<html>
```
janbartel: Probably don't need this file: you only ever hit the `/session` url, don't you? If it is really necessary to have it, lets make it a static index.html, rather than bringing `jsp` into the mix.

tests/test-distribution/src/test/resources/hazelcast-server.xml
```diff
@@ -0,0 +1,52 @@
...
+      <multicast enabled="false"/>
+      <aws enabled="false"/>
+    </join>
+  </network>
+
+  <properties>
+    <property name="hazelcast.rest.enabled">true</property>
+  </properties>
+
+<!--  <serialization>-->
```
janbartel: Should this be commented out?

tests/test-distribution/src/test/resources/hazelcast-server.xml
```diff
@@ -0,0 +1,52 @@
...
+<hazelcast xsi:schemaLocation="http://www.hazelcast.com/schema/config hazelcast-config-3.8.xsd"
+           xmlns="http://www.hazelcast.com/schema/config"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+
+  <network>
+    <reuse-address>true</reuse-address>
+    <port port-count="20">5701</port>
+    <join>
+      <tcp-ip enabled="true">
+<!--        <member-list>-->
```
janbartel, line 32: Should this be commented out?

tests/test-distribution/src/test/java/org/eclipse/jetty/tests/distribution/HazelcastSessionDistributionTests.java
```diff
@@ -0,0 +1,258 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class HazelcastSessionDistributionTests extends AbstractDistributionTest
+{
+    private static final Logger HAZELCAST_LOG = LoggerFactory.getLogger("org.eclipse.jetty.tests.distribution.HazelcastLogs");
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(HazelcastSessionDistributionTests.class);
+
+    @Test
+    public void testHazelcastRemoteOnlyClient() throws Exception
```
janbartel, line 66: Add a comment explaining the test (for me there's something backwards about the `onlyClient` hazelcast option and I always get confused about what is being the client and what is being the server).

janbartel (changes requested):
Just a couple of changes and then it looks fine.

tests/test-distribution/src/test/resources/hazelcast-server.xml
olamy: yes and no :) as the test is still commented but yes when hazelcast problem will be solved

tests/test-distribution/src/test/java/org/eclipse/jetty/tests/distribution/HazelcastSessionDistributionTests.java
```diff
@@ -0,0 +1,263 @@
...
+            String jettyVersion = System.getProperty("jettyVersion");
+            DistributionTester distribution = DistributionTester.Builder.newInstance()
+                .jettyVersion(jettyVersion)
+                .mavenLocalRepository(System.getProperty("mavenRepoPath"))
+                .build();
+
+            String[] args1 = {
+                "--create-startd",
+                "--approve-all-licenses",
+                "--add-to-start=resources,server,http,webapp,deploy`,jmx,servlet,servlets,session-store-hazelcast-remote"
```
janbartel: There's a back quote char after `deploy`.

janbartel (changes requested):

tests/test-distribution/src/test/java/org/eclipse/jetty/tests/distribution/HazelcastSessionDistributionTests.java
olamy: good catch. arghh and CI cannot see it as it's not executed :)

janbartel (approved):

PR #6073, opened 2021-03-18
Fixes #6072 - jetty server high CPU when client sends TLS record length > 17408 (jetty-9.4.x)

### LINKED ISSUE #6072: High CPU usage when TLS client sends TLS Record data exceeding length 17408
**Jetty version**
9.4.36.v20210114

**Java version**
openjdk 8u272

**OS type/version**
linux 

**Description**

When using a socket to send data with a size greater than 17408, an endless loop will occur in the jetty server.
The following demo can reproduce this problem：

``` java
// server 
public class MyServer {
    public static void startServer() throws Exception {
         SslContextFactory sslContextFactory = new SslContextFactory.Server();
         HttpConnectionFactory httpFactory = new HttpConnectionFactory();
         Server server = new Server();
         ServerConnector connector = new 
         ServerConnector(server,null,null,null,1,-1,AbstractConnectionFactory.getFactories(sslContextFactory ,httpFactory));
         connector.setPort(9988);
         connector.setHost("localhost");
         server.setConnectors(new Connector[]{connector});
         ServletContextHandler context = new ServletContextHandler();
         context.setContextPath("/");
         HandlerCollection handlerCollection= new HandlerCollection();
         handlerCollection.setHandlers(new Handler[]{context,new DefaultHandler()});
         server.setHandler(handlerCollection);
         server.start();
         server.join();
    }
        startServer();
     }
}



public class MyClient {
    // client 
    private static byte[] buildMessage() {
       byte[] bytes = new byte[20005];
       bytes[0] = 22;  // record type
       bytes[1] = 3;   // major version
       bytes[2] = 3;   // minor version
       bytes[3] = 78; // record length 2 bytes
       bytes[4] = 32;  // record length
    
       bytes[5] = 1; // message type
       bytes[6] = 0; // message length 3 bytes
       bytes[7] = 78;
       bytes[8] = 23;
      
      for( int i = 9; i < bytes.length; i++) {
          bytes[i] = 1;
      }
      return bytes;
    }
    
    public static void sendMessage() throws Exception {
        byte[] bytes = buildMessage();
        SocketFactory socketFactory = SocketFactory.getDefault();
        Socket socket = socketFactory.createSocket("localhost",9988);
        socket.getOutputStream().write(bytes);
        socket.close();
    }

        sendMessage();
    }
}
```

sbordet:
…408.

Avoid spinning if the input buffer is full.

Co-authored-by: Joakim Erdfelt <joakim.erdfelt@gmail.com>

### DIFF
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
@@ -729,13 +729,15 @@ public class SslConnection extends AbstractConnection implements Connection.Upgr
                                     return filled = -1;
 
                                 case BUFFER_UNDERFLOW:
-                                    if (netFilled > 0)
+                                    if (BufferUtil.space(_encryptedInput) == 0)
                                     {
-                                        if (BufferUtil.space(_encryptedInput) > 0)
-                                            continue; // try filling some more
                                         BufferUtil.clear(_encryptedInput);
                                         throw new SSLHandshakeException("Encrypted buffer max length exceeded");
                                     }
+
+                                    if (netFilled > 0)
+                                        continue; // try filling some more
+
                                     _underflown = true;
                                     if (netFilled < 0 && _sslEngine.getUseClientMode())
                                     {
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
@@ -730,7 +730,12 @@ public class SslConnection extends AbstractConnection implements Connection.Upgr
 
                                 case BUFFER_UNDERFLOW:
                                     if (netFilled > 0)
-                                        continue; // try filling some more
+                                    {
+                                        if (BufferUtil.space(_encryptedInput) > 0)
+                                            continue; // try filling some more
+                                        BufferUtil.clear(_encryptedInput);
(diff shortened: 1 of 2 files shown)


gregw (approved):

jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
```diff
@@ -730,7 +730,12 @@ public int fill(ByteBuffer buffer) throws IOException
 
                                 case BUFFER_UNDERFLOW:
                                     if (netFilled > 0)
-                                        continue; // try filling some more
+                                    {
+                                        if (BufferUtil.space(_encryptedInput) > 0)
+                                            continue; // try filling some more
+                                        BufferUtil.clear(_encryptedInput);
+                                        throw new SSLHandshakeException("Encrypted buffer max length exceeded");
+                                    }
```
gregw: I think the following is better/safer:

```suggestion
                                        continue; // try filling some more
                                        
                                    if (netFilled == 0 && BufferUtil.space(_encryptedInput) == 0)
                                    {
                                        BufferUtil.clear(_encryptedInput);
                                        throw new SSLHandshakeException("Encrypted buffer max length exceeded");
                                    }
```

gregw (changes requested):
Changed my mind.... I think we should check the size on any zero fill, not just the last non zero

gregw (approved):

PR #6078, opened 2021-03-22
Fix #6076 Protect from null local certificates

### LINKED ISSUE #6067: Easier access to invalid client certificates
**Jetty version**
10.0.x

**Description**
During the TLS handshake, in case of `needClientAuth`, the client may send an invalid (e.g. expired) certificate.
The validation checks are performed by the `TrustManager` and if they fail there is no way to access the expired client certificate, for example in `SslHandshakeListener.handshakeFailed()`, as it is not exposed via `SSLSession.getPeerCertificate()`, etc.

The only option would be to wrap the `TrustManager`, but that requires subclassing `SslContextFactory.Server` and overriding `getTrustManager()`, whose signature is likely to change in light of #6054.

Would be great to have a more stable way to provide hooks into the `TrustManager` in a simpler way.

### LINKED ISSUE #6076: Embedded Jetty throws null pointer exception
**Jetty version**
11.0.1
**Java version**
11
**OS type/version**
MacOX
**Description**
I'm using [this code][1] verbatim to run an embedded Jetty server. 

HTTP calls work totally fine. However, HTTPS calls start failing after sometime with a null pointer exception

    java.lang.NullPointerException
    	at org.eclipse.jetty.server.SecureRequestCustomizer.customize(SecureRequestCustomizer.java:247)
    	at org.eclipse.jetty.server.SecureRequestCustomizer.customize(SecureRequestCustomizer.java:205)
    	at org.eclipse.jetty.server.HttpChannel.lambda$handle$0(HttpChannel.java:395)
    	at org.eclipse.jetty.server.HttpChannel.dispatch(HttpChannel.java:656)
    	at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:391)
    	at org.eclipse.jetty.server.HttpChannel.run(HttpChannel.java:351)
    	at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.runTask(EatWhatYouKill.java:333)

Debugging code I found that the host is null in this code block in Jetty's org.eclipse.jetty.server.SecureRequestCustomizer:

    String sniHost = (String)sslSession.getValue(SslContextFactory.Server.SNI_HOST);

  [1]: https://github.com/eclipse/jetty.project/blob/jetty-9.4.5.v20170502/examples/embedded/src/main/java/org/eclipse/jetty/embedded/Http2Server.java

gregw:
Fix #6076 Protect from null local certificates

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/SecureRequestCustomizer.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/SecureRequestCustomizer.java
@@ -53,6 +53,7 @@ public class SecureRequestCustomizer implements HttpConfiguration.Customizer
     public static final String JAVAX_SERVLET_REQUEST_CIPHER_SUITE = "javax.servlet.request.cipher_suite";
     public static final String JAVAX_SERVLET_REQUEST_KEY_SIZE = "javax.servlet.request.key_size";
     public static final String JAVAX_SERVLET_REQUEST_SSL_SESSION_ID = "javax.servlet.request.ssl_session_id";
+    public static final String X509_CERT = "org.eclipse.jetty.server.x509_cert";
 
     private String sslSessionAttribute = "org.eclipse.jetty.servlet.request.ssl_session";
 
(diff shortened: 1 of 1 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/SecureRequestCustomizer.java
```diff
@@ -244,7 +245,9 @@ protected void customize(SSLEngine sslEngine, Request request)
         if (isSniRequired() || isSniHostCheck())
         {
             String sniHost = (String)sslSession.getValue(SslContextFactory.Server.SNI_HOST);
-            X509 cert = new X509(null, (X509Certificate)sslSession.getLocalCertificates()[0]);
+            Certificate[] certificates = sslSession.getLocalCertificates();
```
sbordet: Isn't much simpler to throw early here, rather than doing `:?` and null checks later?

sbordet (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/SecureRequestCustomizer.java
gregw: yep.  Also I think the X509 instance should be cached, as it does a moderate amount of work in its constructor!

sbordet (approved):

PR #6083, opened 2021-03-22
Fix #6082 Reduce SSL buffer compaction

### LINKED ISSUE #6082: SslConnection compacting
From fixing #6072, it became apparent that the SslConnection compacts buffers frequently.  Performance can be improved if compaction is only done when needed.

gregw:
Fix #6082. Only compact when buffer is underflown.  

Note that BufferUtil will also do a cheap "compact" when flipping empty buffers.

### DIFF
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
@@ -711,12 +711,9 @@ public class SslConnection extends AbstractConnection implements Connection.Upgr
                                     return filled = -1;
 
                                 case BUFFER_UNDERFLOW:
-                                    // Can we compact?
-                                    if (_encryptedInput.position() != 0)
-                                    {
-                                        BufferUtil.compact(_encryptedInput);
+                                    // Continue if we can compact?
+                                    if (BufferUtil.compact(_encryptedInput))
                                         continue;
-                                    }
 
                                     // Are we out of space?
                                     if (BufferUtil.space(_encryptedInput) == 0)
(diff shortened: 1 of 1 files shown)


gregw:
Running our various SSL tests prior to this PR indicated that compaction was frequently done, with 4, 10 or more compactions done on many tests.
With this PR, compaction is seldom done. In fact it has only been observed in the SslServerBytesTest which needs java11.  So the issue is that this change is now probably a little under tested.  However it should be a significant performance gain.

jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
```diff
@@ -712,12 +711,21 @@ public int fill(ByteBuffer buffer) throws IOException
                                     return filled = -1;
 
                                 case BUFFER_UNDERFLOW:
+                                    // Can we compact?
+                                    if (_encryptedInput.position() != 0)
+                                    {
+                                        BufferUtil.compact(_encryptedInput);
```
lorban: Beware: `compact()` is an optional operation that may not do anything. If that happens, aren't you going to enter an infinite loop?

So shouldn't you check the returned boolean of `BufferUtil.compact` and only continue if it was true?

lorban (changes requested):
One detail I'm unsure about.

jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
gregw: ah that's much better.  Avoids double access of position too!

lorban (approved):

sbordet (approved):

PR #6088, opened 2021-03-23
Issue #6085 Fix reference counts for multiple valid cookies for sessions

### LINKED ISSUE #6085: Jetty keeps Sessions in use after "Duplicate valid session cookies" Message
**Jetty version**
9.4.26

**Java version**
openjdk version "10.0.2" 2018-07-17

**OS type/version**
CentOS Linux 7 (core)

**Description**
After getting org.eclipse.jetty.http.BadMessageException: 400: Duplicate valid session cookies jetty keeps the duplicate sessions in use state (request counter in session >0). Thats why these session will not be invalidated by the HouseKeeper-Thread.
Is this a bug in jetty?

janbartel:
Closes #6085 

If a request has multiple session cookies, and the sessions for more than one of them are valid, then the request counts for the valid sessions were not being properly decremented.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/session/SessionHandler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/session/SessionHandler.java
@@ -1670,7 +1670,7 @@ public class SessionHandler extends ScopedHandler
                         //retrieve the session, which increments the reference count
                         HttpSession s = getHttpSession(id);
                         //associate it with the request so its reference count
-                        //will be decremented as the request exits
+                        //will be decremented as the request completes
                         if (s != null && isValid(s))
                             baseRequest.enterSession(s);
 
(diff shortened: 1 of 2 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/session/SessionHandler.java
```diff
@@ -1667,7 +1667,12 @@ else if (!DispatcherType.REQUEST.equals(baseRequest.getDispatcherType()))
                         if (LOG.isDebugEnabled())
                             LOG.debug("Got Session ID {} from cookie {}", id, sessionCookie);
 
+                        //retrieve the session
```
gregw:
```suggestion
                        //retrieve the session, which increments the reference count
```

jetty-server/src/main/java/org/eclipse/jetty/server/session/SessionHandler.java
```diff
@@ -1667,7 +1667,12 @@ else if (!DispatcherType.REQUEST.equals(baseRequest.getDispatcherType()))
...
                         HttpSession s = getHttpSession(id);
+                        //associate it with the request so its reference count
+                        //will be decremented as the request exits
```
gregw:
```suggestion
                        //will be decremented as the request completes
```

gregw (approved):
OK other than some niggles.
There is some repeated logic, but I can't see how to clean it up without risking changing behaviour.

PR #6090, opened 2021-03-24
Improve WebAppProvider scanning

gregw:
Cherry picked from 9
Ensure that WebAppProvider Filter always canonicalises the file passed in from the Scanner. Thus, both the monitored directory is canonical as well as the file it is being compared against.


janbartel (approved):
LGTM modulo other PR for Files.isSame

PR #6091, opened 2021-03-24
jetty 9.4.x make tests working on environment without docker and port 8080 already used

olamy:


jetty-cdi/src/test/java/org/eclipse/jetty/embedded/EmbeddedWeldTest.java
```diff
@@ -167,7 +167,7 @@ public void testServletContext(String mode) throws Exception
     @Test
     public void testWebappContext() throws Exception
     {
-        Server server = new Server(8080);
+        Server server = new Server(0);
```
janbartel, line 170: The PR is about infinispan, but this is a change to a Weld cdi test?

jetty-cdi/src/test/java/org/eclipse/jetty/embedded/EmbeddedWeldTest.java
```diff
@@ -199,7 +199,7 @@ public void testWebappContext() throws Exception
     @Test
     public void testWebappContextDiscovered() throws Exception
     {
-        Server server = new Server(8080);
+        Server server = new Server(0);
```
janbartel, line 202: See previous comment.

janbartel (changes requested):
If you remove the changes to the Weld CDI test and then this should be ok.

jetty-cdi/src/test/java/org/eclipse/jetty/embedded/EmbeddedWeldTest.java
olamy: I have updated the PR title to explain the goal of this PR. 
Please note each change (testing if docker available and one for the hardcoded port) are already different commits.
So having 2 PRs will have the same result at the end with 2 different commits in the git history and each will have a correct description of the change.

janbartel (approved):

PR #6093, opened 2021-03-24
Use Files.isSameFile to check Resource equality

gregw:
Avoid using canonical and instead use Files.isSameFile

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
@@ -312,6 +312,18 @@ public abstract class Resource implements ResourceFactory, Closeable
     public abstract boolean isContainedIn(Resource r) throws MalformedURLException;
 
     /**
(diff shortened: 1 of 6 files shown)


jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -311,6 +311,18 @@ protected void finalize()
...
+    /**
+     * Return true if the passed Resource represents the same resource as the Resource.
+     * For many resource types, this is equivalent to {@link #equals(Object)}, however
+     * for resources types that support aliasing, this maybe some other check (e.g. {@link java.nio.file.Files#isSameFile(Path, Path)}).
+     * @param resource The resource to check
+     * @return true if the passed resource represents the same resource.
+     */
+    public boolean isSame(Resource resource)
+    {
+        return equals(resource);
```
janbartel, line 323: Probably should implement this for FileResource, even if it is deprecated.

janbartel (changes requested):

janbartel (approved):

PR #6094, opened 2021-03-24
Do not store WebSocketComponents in Server scope.

lachlan-roberts:

### DIFF
--- a/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/WebSocketServerComponents.java
+++ b/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/WebSocketServerComponents.java
@@ -83,11 +83,11 @@ public class WebSocketServerComponents extends WebSocketComponents
 
         // These components may be managed by the server but not yet started.
         // In this case we don't want them to be managed by the components as well.
-        if (server.isManaged(inflaterPool))
+        if (server.contains(inflaterPool))
             serverComponents.unmanage(inflaterPool);
-        if (server.isManaged(deflaterPool))
+        if (server.contains(deflaterPool))
             serverComponents.unmanage(deflaterPool);
-        if (server.isManaged(bufferPool))
+        if (server.contains(bufferPool))
             serverComponents.unmanage(bufferPool);
 
         servletContext.setAttribute(WEBSOCKET_COMPONENTS_ATTRIBUTE, serverComponents);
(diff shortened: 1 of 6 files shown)


sbordet (changes requested):
@lachlan-roberts please let's review the lifecycle of `WebSocketServerComponent` created in `ensureWebSocketComponents` like we discussed.

jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/WebSocketServerComponents.java
```diff
@@ -35,31 +38,70 @@
...
+        WebSocketComponents serverComponents = new WebSocketServerComponents(inflaterPool, deflaterPool, bufferPool);
+
+        // These components may be managed by the server but not yet started.
+        // In this case we don't want them to be managed by the components as well.
+        if (server.isManaged(inflaterPool))
+            serverComponents.unmanage(inflaterPool);
+        if (server.isManaged(deflaterPool))
+            serverComponents.unmanage(deflaterPool);
+        if (server.isManaged(bufferPool))
+            serverComponents.unmanage(bufferPool);
```
gregw: We've very rarely, if ever needed to do this kind of explicit management of the management of beans.

Typically if a server bean is shared with handlers, it is added as auto in all places and the fact that the server starts it first is enough to make it managed there and not managed in the handlers.

Is this an issue here because the component is lazily added to the server and thus the order of initialisation is unreliable? 
 
Also, is it not the case that if the component is on the server, then it should not be managed here?
Ie, even if the components are created by a context, if they are put on the server then they can be managed by the server.

So at the very least, we need some good comments/doco describing the various options for the lifecycles and sharing of these components.

gregw (changes requested):
needs some more comments at least

jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/WebSocketServerComponents.java
lachlan-roberts: > Is this an issue here because the component is lazily added to the server and thus the order of initialisation is unreliable?

This is an issue because the server may start the `ServletContextHandler` before it starts its `InflaterPool` / `DeflaterPool`. As they are all beans on the server it depends on what order they're added to what order they are started, and we create the `WebSocketServerComponents` in the `ServletContainerInitializer` when the `ServletContextHandler` is being started.

So without this change, this would work:
```java
server.addManaged(inflaterPool);
server.addManaged(deflaterPool);
server.setHandler(contextHandler);
```
But this would not work:
```java
server.setHandler(contextHandler);
server.addManaged(inflaterPool);
server.addManaged(deflaterPool);
```

> Also, is it not the case that if the component is on the server, then it should not be managed here?

The `WebSocketComponents` is now never taken from the server. One is always created for each `ServletContext`.

gregw (approved):

sbordet (approved):

PR #6095, opened 2021-03-24
Use Files.isSameFile to check Resource equality (#6093)

gregw:
cherry picked from 9
Use Files.isSameFile to check Resource equality
Avoid using canonical and instead use Files.isSameFile

### DIFF
--- a/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/WebAppProvider.java
+++ b/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/WebAppProvider.java
@@ -15,7 +15,6 @@ package org.eclipse.jetty.deploy.providers;
 
 import java.io.File;
 import java.io.FilenameFilter;
-import java.io.IOException;
 import java.util.Locale;
 
 import org.eclipse.jetty.deploy.App;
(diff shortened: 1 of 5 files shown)


jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/WebAppProvider.java
```diff
@@ -75,62 +74,45 @@
...
-                    return false;
-                
-                String lowerName = name.toLowerCase(Locale.ENGLISH);
+            String lowerName = name.toLowerCase(Locale.ENGLISH);
 
-                File canonical = new File(dir, name).getCanonicalFile();
-                Resource r = Resource.newResource(canonical);
-                if (getCanonicalMonitoredResources().contains(r) && r.isDirectory())
+            Resource resource = Resource.newResource(new File(dir, name));
+            for (Resource m : getMonitoredResources())
```
lorban: Minor improvement:

```java
if (getMonitoredResources().stream().anyMatch(resource::isSame))
    return false;
```

jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/WebAppProvider.java
```diff
@@ -75,62 +74,45 @@
         @Override
         public boolean accept(File dir, String name)
         {
-            if (dir == null)
+            if (dir == null || !dir.exists())
```
lorban: Should we use `dir.canRead()` instead of `dir.exists()`? This would also reject unreadable directories. Or do we want to keep unreadable dirs to make the scanner report the error and/or retry them later?

jetty-util/src/test/java/org/eclipse/jetty/util/resource/PathResourceTest.java
```diff
@@ -113,4 +114,16 @@ public void testDefaultFileSystemGetFile() throws Exception
         File file = resource.getFile();
         assertThat("File for default FileSystem", file, is(exampleJar.toFile()));
     }
+
+    @Test
+    public void testSame() throws Exception
```
lorban: I would add an assertion that checks that symlink'ed paths behave as they should as this is the original root cause.

Something like this (except that I do not know if that would work on windows):

```java

Path epath2 = Files.createSymbolicLink(MavenTestingUtils.getTargetPath().resolve("testSame-symlink"), epath.getParent()).resolve("example.jar");
PathResource ePathResource2 = new PathResource(epath2);

assertThat(ePathResource.isSame(ePathResource2), Matchers.is(true));
assertThat(ePathResource.equals(ePathResource2), Matchers.is(false));
```

lorban (changes requested):
Looking good, but lacks a test with a symlink IMHO.

lorban (approved):

sbordet (approved):

jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/WebAppProvider.java
jmcc0nn3ll: I think it is legit to configure directories that may provide webapps that don't exist at the start time of the server, to be created and fleshed out whilst the server is running.  Scanner should just check for existence during its normal course of business and incorporate things as they show up.

gregw: perhaps, but let's make changes like that in another PR after we have fixed the CVEs

PR #6096, opened 2021-03-24
Issue #6085 Fix reference counts for multiple valid cookies for sessions

### LINKED ISSUE #6085: Jetty keeps Sessions in use after "Duplicate valid session cookies" Message

janbartel:
cherry  pick from 9.4.x

issue #6085

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/session/SessionHandler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/session/SessionHandler.java
@@ -1614,7 +1614,12 @@ public class SessionHandler extends ScopedHandler
                         if (LOG.isDebugEnabled())
                             LOG.debug("Got Session ID {} from cookie {}", id, sessionCookie);
 
+                        //retrieve the session, which increments the reference count
                         HttpSession s = getHttpSession(id);
+                        //associate it with the request so its reference count
+                        //will be decremented as the request completes
+                        if (s != null && isValid(s))
+                            baseRequest.enterSession(s);
 
                         if (requestedSessionId == null)
                         {
@@ -1640,6 +1645,10 @@ public class SessionHandler extends ScopedHandler
                         }
                     }
                 }
+
+                //if we wound up with a single valid session
+                if (session != null && isValid(session))
+                    baseRequest.setSession(session);  //associate the session with the request
             }
         }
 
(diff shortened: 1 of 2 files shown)


lorban (approved):
This is literally identical to https://github.com/eclipse/jetty.project/pull/6088 so LGTM.

jetty-server/src/main/java/org/eclipse/jetty/server/session/SessionHandler.java
```diff
@@ -1614,7 +1614,12 @@ else if (!DispatcherType.REQUEST.equals(baseRequest.getDispatcherType()))
                         if (LOG.isDebugEnabled())
                             LOG.debug("Got Session ID {} from cookie {}", id, sessionCookie);
 
+                        //retrieve the session, which increments the reference count
                         HttpSession s = getHttpSession(id);
+                        //associate it with the request so its reference count
+                        //will be decremented as the request completes
+                        if (s != null && isValid(s))
+                            baseRequest.enterSession(s);
```
sbordet, line 1622: In all other cases, the calls to `enterSession()` and `setSession()` are in 2 consecutive lines.
Is there a reason to split them in this case?

Can this snippet be moved down to line 1650?

Also, the `else if (requestedSessionId.equals(id))` branch does not assign the `session` variable... Given the comment, I think it should?

sbordet (dismissed):
Kinda looks ok, but I think it could be improved by making the `enterSession()` and `setSession()` calls in consecutive lines, like it is done in other places.

sbordet (changes requested):
See comment.

jetty-server/src/main/java/org/eclipse/jetty/server/session/SessionHandler.java
gregw: I REALLY think we have to stop do "logic cleanups" in the middle of important time critical fixes.  
Such cleanups without clear reasons have broken things before.

janbartel: No, line 1622 cannot be moved down. The whole point of the PR is that when we receive a request that has multiple valid session id cookies, we need to associate both of them with the request so that the request will decrement the reference count on both of them on the way out (the ref count was incremented at line 1618 when each session is loaded).

If the requestedSessionId.equals(id) we don't assign the session object because both ids are the same, and thus the session object will be the same, thus no point in re-assigning it.

sbordet (approved):

PR #6097, opened 2021-03-24
Exposed the configuration option to scan manifests on the JAR scanner.

tet-lenovo:
We precompile JSPs and it takes a long time to run because we are scanning lots of JAR files that will never contain TLDs. This patch simply exposes the configuration option on the JAR scanner.

Signed-off-by: Tony Thompson <tthompson1@lenovo.com>


janbartel (changes requested):
Thanks for the PR, looks good! Only thing is, can you please also update the documentation for the jspc maven plugin to document this new option?

janbartel (approved):
LGTM. Thanks!

PR #6107, opened 2021-03-26
Make the JavaxWebSocketClientContainer.getContainer(HttpClient) method static.

lachlan-roberts:
- Make the JavaxWebSocketClientContainer.getContainer(HttpClient) method static.
- Introduce utility methods to add and remove beans from containers.


gregw (commented):
Shouldn't your websocket stuff use these new methods?

lachlan-roberts:
@gregw the only place the method was used is from the no args `getContainer()` method of `JavaxWebSocketClientContainerProvider`. There are no other usages to update.

gregw (approved):

PR #6108, opened 2021-03-26
Fixes #6105 - HttpConnection.getBytesIn() incorrect for requests with…

### LINKED ISSUE #6105: HttpConnection.getBytesIn() incorrect for requests with chunked content
**Jetty version**
9.4.38.v20210224

**Java version**
11.0.7+10

**OS type/version**
MacOS 10.15.7

**Description**
`HttpConnection.bytesIn` is only aggregated when `fillRequestBuffer()` is called from `onFillable()`, not from `fillAndParseForContent()`. The result is that `HttpConnection.getBytesIn()` reports a value lower than than the actual number of bytes filled. I have been able to reproduce this for POST requests using chunked encoding.

sbordet:
… chunked content

Moved recording of bytes to fillRequestBuffer(),
so they are accounted also for async reads.
Added test case.

### DIFF
--- a/jetty-server/src/test/java/org/eclipse/jetty/server/HttpConnectionTest.java
+++ b/jetty-server/src/test/java/org/eclipse/jetty/server/HttpConnectionTest.java
@@ -37,6 +37,8 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import java.util.stream.Stream;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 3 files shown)


jetty-server/src/test/java/org/eclipse/jetty/server/HttpConnectionTest.java
```diff
@@ -1361,6 +1367,68 @@ public void testCONNECT() throws Exception
...
+        {
+            @Override
+            public void handle(String target, Request jettyRequest, HttpServletRequest request, HttpServletResponse response) throws IOException
+            {
+                jettyRequest.setHandled(true);
+                IO.copy(request.getInputStream(), IO.getNullStream());
+
+                HttpConnection connection = HttpConnection.getCurrentConnection();
+                long bytesIn = connection.getBytesIn();
+                assertThat(bytesIn, greaterThan(dataLength));
```
gregw, line 1387: Can't we do better than this?   We should know exactly how many bytes have been sent on a connection.
At the very least we need to check it is greater than `dataLength + headerBytes`

gregw (changes requested):

jetty-server/src/test/java/org/eclipse/jetty/server/HttpConnectionTest.java
sbordet: @gregw I don't want to count exactly bytes for headers, as they may depend on dates, version strings that have additional digits (e.g. 10.0.9 -> 10.0.10), etc.
I just want to make sure that "resumed" reads count all the content -- I verified that the test failed before the fix.

gregw (approved):

jetty-server/src/test/java/org/eclipse/jetty/server/HttpConnectionTest.java
gregw: hmmmm

Issue #5798, opened 2020-12-11
jetty-runner startup error with jetty-10


cor3000:
**Jetty version** 10.0.0

**Java version** openjdk version "11.0.9.1" 2020-11-04

**OS type/version** Linux Mint 20 

**Description**

Starting any war file or webapp folder (I tried so far) using the jetty-runner results in the exceptions below
accessing localhost:8080 results in an error code 503 

The same war files work with jetty-runner 9.4.35.

```
~/temp (mytheme)$ java -jar jetty-runner-10.0.0.jar anyapp.war
WARNING: jetty-runner is deprecated.
         See Jetty Documentation for startup options
         https://www.eclipse.org/jetty/documentation/
2020-12-11 16:21:22.136:INFO :oejr.Runner:main: Runner
2020-12-11 16:21:22.268:INFO :oejs.Server:main: jetty-10.0.0; built: 2020-12-02T21:06:43.949Z; git: b9645a17373e4e9b7f30b6c0a07defcea2cb660b; jvm 11.0.9.1+1-Ubuntu-0ubuntu1.20.04
2020-12-11 16:21:22.495:INFO :oeja.AnnotationConfiguration:main: Error:  for o.e.j.w.WebAppContext@6a41eaa2{/,file:///home/robert/workspace/temp/,STARTING}{file:///home/robert/workspace/temp/}
2020-12-11 16:21:22.803:INFO :oejw.StandardDescriptorProcessor:main: NO JSP Support for /, did not find org.eclipse.jetty.jsp.JettyJspServlet
2020-12-11 16:21:22.808:WARN :oejs.BaseHolder:main: Unable to load class org.eclipse.jetty.servlet.listener.ELContextCleaner
java.lang.ClassNotFoundException: org.eclipse.jetty.servlet.listener.ELContextCleaner
	at org.eclipse.jetty.webapp.WebAppClassLoader.loadClass(WebAppClassLoader.java:544)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	at org.eclipse.jetty.util.Loader.loadClass(Loader.java:64)
	at org.eclipse.jetty.servlet.BaseHolder.doStart(BaseHolder.java:102)
	at org.eclipse.jetty.servlet.ListenerHolder.doStart(ListenerHolder.java:75)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:98)
[...]

PR #6115, opened 2021-03-30
Issue #5798 fix jetty runner startup error and add jsp support as well

olamy:


jetty-runner/src/main/java/org/eclipse/jetty/runner/Runner.java
```diff
@@ -78,7 +78,9 @@
             org.eclipse.jetty.plus.webapp.EnvConfiguration.class.getCanonicalName(),
             org.eclipse.jetty.plus.webapp.PlusConfiguration.class.getCanonicalName(),
             org.eclipse.jetty.annotations.AnnotationConfiguration.class.getCanonicalName(),
-            org.eclipse.jetty.webapp.JettyWebXmlConfiguration.class.getCanonicalName()
+            org.eclipse.jetty.webapp.JettyWebXmlConfiguration.class.getCanonicalName(),
+            org.eclipse.jetty.webapp.WebAppConfiguration.class.getCanonicalName(),
+            org.eclipse.jetty.webapp.JspConfiguration.class.getCanonicalName()
         };
```
janbartel, line 87: Seems like a nice simple change. Is there any way to extend the current integration test to actually test the runner?

olamy: good point. I need to figure out how to do it

olamy: @janbartel test added

jetty-runner/src/it/demo-simple-webapp-runner/pom.xml
```diff
@@ -0,0 +1,137 @@
...
+          <systemPropertyVariables>
+          </systemPropertyVariables>
+          <dependenciesToScan>
+            <dependency>org.eclipse.jetty:jetty-runner</dependency>
+          </dependenciesToScan>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+</project>
```
janbartel, line 139: @olamy would any of this test setup be simplified at all if it moved over into tests/test-runner instead?  The integration test setup is impressive, but looks really complex.

olamy: could be but it will be simply the same poms etc... move to tests/test-runner. 
What if we want to add more tests? I don't want to turn having tests/test-runner-2 etc...

jetty-runner/src/main/java/org/eclipse/jetty/runner/Runner.java
```diff
@@ -164,6 +170,7 @@ public void usage(String error)
         System.err.println(" --out file                          - info/warn/debug log filename (with optional 'yyyy_mm_dd' wildcard");
         System.err.println(" --host name|ip                      - interface to listen on (default is all interfaces)");
         System.err.println(" --port n                            - port to listen on (default 8080)");
+        System.err.println(" --port-file path                    - file to write the connector port ");
```
joakime: This description is a bit vague.

jetty-runner/src/it/demo-simple-webapp-runner/pom.xml
joakime: This uses the `jetty-runner.jar` as built, and without the maven classloader causing false success.
Does tests/test-runner satisfy these 2 (kinda important) aspects of testing jetty runner?

olamy: will only move this IT test as a simple project under tests/test-runner. But I don't like it. I prefer this structure using the invoker plugin within jetty-runner project.

jetty-runner/src/main/java/org/eclipse/jetty/runner/Runner.java
olamy: not sure. what about `file to write a single line with the port used to serve http requests` ?

olamy: @joakime description updated

olamy:
@joakime nudge

jetty-runner/src/main/java/org/eclipse/jetty/runner/Runner.java
```diff
@@ -164,6 +170,7 @@ public void usage(String error)
         System.err.println(" --out file                          - info/warn/debug log filename (with optional 'yyyy_mm_dd' wildcard");
         System.err.println(" --host name|ip                      - interface to listen on (default is all interfaces)");
         System.err.println(" --port n                            - port to listen on (default 8080)");
+        System.err.println(" --port-file path                    - file to write a single line with the port used to serve http requests ");
```
joakime: This should probably be more than just the port.
It should likely contain the value of `Server.getURI()` (after `Server.start()`)so that the scheme + network address + port + context path are all communicated.

joakime: Would `--server-uri-file path` be a better option?

joakime (commented):
I think this can be improved.

jetty-runner/src/test/java/org/eclipse/jetty/maven/jettyrunner/it/IntegrationTestJettyRunner.java
```diff
@@ -31,13 +31,12 @@
...
     {
-        int port = findPortUsed();
+        String serverUri = findServerUri();
         HttpClient httpClient = new HttpClient();
         try
         {
             httpClient.start();
-            ContentResponse response =
-                httpClient.newRequest("localhost", port).path("/").send();
+            ContentResponse response = httpClient.newRequest(serverUri).send();
```
joakime, line 39: Keep in mind that the serverUri will contain the contextPath of the first ContextHandler, if it isn't `/` then you'll need to `URI.create(serverUri).resolve("/")` to use the path at root.

olamy: but here we have only one war deployed.

olamy: interesting the documentation says `[[--path /path] context]*n` but looking at the code we do not support this `n` ;)

joakime (approved):

Issue #6106, opened 2021-03-25
WebSocket/CDI integration is broken in Jetty 10


rdicroce:
**Jetty version**
10.0.1

**Java version**
11.0.8

**OS type/version**
Windows 10 Pro 20H2

**Description**
Classes annotated with `@ServerEndpoint` do not receive CDI integration. You can reproduce the problem easily by upgrading https://github.com/jetty-project/embedded-jetty-weld to Jetty 10. The code needs a few tweaks since some things were moved around, but after that's fixed, you'll notice that TimeSocket blows up due to a NPE because the logger hasn't been injected. After digging around in the code a bit, the cause appears to be that nothing ever calls the ObjectFactory to decorate the actual endpoint class.

Workaround: specify a Configurator on the `@ServerEndpoint`. However, this does not work:
```java
	@Override
	public <T> T getEndpointInstance(Class<T> endpointClass) throws InstantiationException {
		return CDI.current().select(endpointClass).get();
	}
```
At least, it doesn't work if you're using Weld and the endpoint class is ApplicationScoped. In that case, Weld will return a proxy subclass, which creates a problem in JavaxWebSocketServerFrameHandlerFactory#getMetadata() because it looks for `@ServerEndpoint` but that annotation is not `@Inherited`. Therefore the annotation is not found and the configuration is ignored.

Instead, this works, if all you need is injection:
```java
	@Override
	public <T> T getEndpointInstance(Class<T> endpointClass) throws InstantiationException {
		return new Unmanaged<>(endpointClass).newInstance().produce().inject().get();
	}
```

lachlan-roberts:
@rdicroce this issue is fixed by PR #6106, I have verified that https://github.com/jetty-project/embedded-jetty-weld now works if you run it against the current 10.0.x branch. This fix will be released in 10.0.3.

rdicroce:
@lachlan-roberts Thanks for the fix!

PR #6119, opened 2021-03-31
Issue #6106 - Fix WebSocket/CDI integration in Jetty 10

lachlan-roberts:
Issue #6106


joakime (changes requested):
The `javax.websocket` components need a `DecoratedObjectFactory.decorate()` call as well, if they were created by Jetty. (as opposed to being handed to Jetty already instantiated by the application).

* javax.websocket.Encoder
* javax.websocket.Decoder
* javax.websocket.ClientEndpointConfig.Configurator
* javax.websocket.server.ServerEndpointConfig.Configurator

joakime:
We should also support CDI on `Extension` instance we create as well.

joakime:
There's also (in jetty-10+) the `org.eclipse.jetty.websocket.core.server.Handshaker.newInstance()` that should go through the DecoratedObjectFactory.

gregw (changes requested):
tell me when ready to review again...

lachlan-roberts:
@joakime we already use the `DecoratedObjectFactory` to create extensions. And I don't think we need it for `Handshaker.newInstance()` because that is just creating an internal class, you can't provide a custom handshaker for that method to create.

joakime (changes requested):
I'd like to see a testcase of a `javax.websocket` Server Endpoint that has an `@Inject` of the servlet `HttpSession` as that is a very common use case.

gregw (dismissed):
wow! Big!!!

lachlan-roberts:
> I'd like to see a testcase of a javax.websocket Server Endpoint that has an @Inject of the servlet HttpSession as that is a very common use case.

@joakime I think this isn't going to be very simple, so should be taken out to a separate issue.

We would need to somehow propagate the `@RequestScope` context for the lifecycle of the websocket connection. Otherwise the websocket will try access the injected `HttpSession` from a different thread from the `WebSocketUpgradeFilter` so it won't be in CDI scope anymore and we get an error trying to use it. We decorate the endpoint instance during the call to `WebSocketUpgradeFilter`, but it only injects a weld CDI wrapper of the object which doesn't work when used in a different thread.

joakime:
> > I'd like to see a testcase of a javax.websocket Server Endpoint that has an @Inject of the servlet HttpSession as that is a very common use case.
> 
> @joakime I think this isn't going to be very simple, so should be taken out to a separate issue.
> 
> We would need to somehow propagate the `@RequestScope` context for the lifecycle of the websocket connection. Otherwise the websocket will try access the injected `HttpSession` from a different thread from the `WebSocketUpgradeFilter` so it won't be in CDI scope anymore and we get an error trying to use it. We decorate the endpoint instance during the call to `WebSocketUpgradeFilter`, but it only injects a weld CDI wrapper of the object which doesn't work when used in a different thread.

That's not how it works.
The `@Inject HttpSession httpSession` will cause the `HttpSession` from the servlet layer to be added/set to the Endpoint during our `DecoratedObjectFactory.decorate(endpoint)` call.

The (server) endpoint we create should have the HttpSession populated from the servlet layer during the time of the upgrade itself.
There is no update to that endpoint by CDI after the upgrade is complete.

lachlan-roberts:
@joakime it is how it works, I have a test for this which fails (which I have not committed to this branch).

It successfully injects the `HttpSession` into the WebSocket Endpoint, but its class is `org.jboss.weldx.servlet.http.HttpSession$$Proxy$_$$_WeldClientProxy`. It works fine if accessed in the same thread that's doing the call to `doFilter()`, but if accessed from another thread, it will throw `ContextNotActiveException`.

```
Caused by: org.jboss.weld.contexts.ContextNotActiveException: WELD-001303: No active contexts for scope type javax.enterprise.context.SessionScoped
	at org.jboss.weld.manager.BeanManagerImpl.getContext(BeanManagerImpl.java:647)
	at org.jboss.weld.bean.ContextualInstanceStrategy$DefaultContextualInstanceStrategy.getIfExists(ContextualInstanceStrategy.java:89)
	at org.jboss.weld.bean.ContextualInstance.getIfExists(ContextualInstance.java:63)
	at org.jboss.weld.bean.proxy.ContextBeanInstance.getInstance(ContextBeanInstance.java:87)
	at org.jboss.weld.bean.proxy.ProxyMethodHandler.invoke(ProxyMethodHandler.java:105)
	at org.jboss.weldx.servlet.http.HttpSession$$Proxy$_$$_WeldClientProxy.getId(Unknown Source)
	at org.eclipse.jetty.cdi.tests.websocket.JavaxWebSocketCdiTest$CdiHttpSessionSocket.onMessage(JavaxWebSocketCdiTest.java:401)
```

joakime:
> @joakime it is how it works, I have a test for this which fails (which I have not committed to this branch).
> 
> It successfully injects the `HttpSession` into the WebSocket Endpoint, but its class is `org.jboss.weldx.servlet.http.HttpSession$$Proxy$_$$_WeldClientProxy`. It works fine if accessed in the same thread that's doing the call to `doFilter()`, but if accessed from another thread, it will throw `ContextNotActiveException`.
> 
> ```
> Caused by: org.jboss.weld.contexts.ContextNotActiveException: WELD-001303: No active contexts for scope type javax.enterprise.context.SessionScoped
> 	at org.jboss.weld.manager.BeanManagerImpl.getContext(BeanManagerImpl.java:647)
> 	at org.jboss.weld.bean.ContextualInstanceStrategy$DefaultContextualInstanceStrategy.getIfExists(ContextualInstanceStrategy.java:89)
> 	at org.jboss.weld.bean.ContextualInstance.getIfExists(ContextualInstance.java:63)
> 	at org.jboss.weld.bean.proxy.ContextBeanInstance.getInstance(ContextBeanInstance.java:87)
> 	at org.jboss.weld.bean.proxy.ProxyMethodHandler.invoke(ProxyMethodHandler.java:105)
> 	at org.jboss.weldx.servlet.http.HttpSession$$Proxy$_$$_WeldClientProxy.getId(Unknown Source)
> 	at org.eclipse.jetty.cdi.tests.websocket.JavaxWebSocketCdiTest$CdiHttpSessionSocket.onMessage(JavaxWebSocketCdiTest.java:401)
> ```

OK, You've sold me on this.

We should ...
1. Open a new issue for the inject of HttpSession in endpoints.
2. This new HttpSession testcase should be disabled and labelled to look at the new issue number.

lachlan-roberts:
> 1. Open a new issue for the inject of HttpSession in endpoints.
> 2. This new HttpSession testcase should be disabled and labelled to look at the new issue number.

@joakime I have made these changes, see #6174 for the new issue.

jetty-websocket/websocket-javax-server/src/main/java/org/eclipse/jetty/websocket/javax/server/internal/JavaxWebSocketCreator.java
```diff
@@ -159,8 +159,6 @@ public Object createWebSocket(ServerUpgradeRequest req, ServerUpgradeResponse re
             // [JSR] Step 6: create endpoint class
             Class<?> endpointClass = config.getEndpointClass();
             Object endpoint = config.getConfigurator().getEndpointInstance(endpointClass);
-            // Do not decorate here (let the Connection and Session start first)
-            // This will allow CDI to see Session for injection into Endpoint classes.
```
joakime, line 163: Why was this comment removed?
Seems like an important reminder for future devs.

joakime (commented):
That comment removal seems odd.

jetty-websocket/websocket-javax-server/src/main/java/org/eclipse/jetty/websocket/javax/server/internal/JavaxWebSocketCreator.java
lachlan-roberts: The comment is incorrect. We don't support the injection of the WebSocket Session into Endpoint classes.

We don't decorate there because we decorate in the `FrameHandlerFactory` instead, and we do this because it can be done in the same way between the Jetty and Javax sides because you can provide your own creator on the Jetty side so we cannot decorate there.

joakime (dismissed):

joakime (approved):

PR #6120, opened 2021-04-01
Fixes #2627 - Improve QoSFilterTest

### LINKED ISSUE #2627: Test failure: QoSFilterTest.testBlockingQosFilter
https://jenkins.webtide.net/job/jetty.project/job/jetty-9.4.x/391/testReport/junit/org.eclipse.jetty.servlets/QoSFilterTest/linux_jdk9___Test___jdk9___testBlockingQosFilter/

gregw:
Reenable disabled test for #2627 with longer sleep time


jetty-servlets/src/test/java/org/eclipse/jetty/servlets/QoSFilterTest.java
```diff
@@ -255,7 +253,7 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) t
                         __maxSleepers = __sleepers;
                 }
 
-                Thread.sleep(50);
+                Thread.sleep(500);
```
lachlan-roberts: I can't see how this will actually fix the problem, the failure was coming from `rethrowExceptions(futures)` so I think one of the workers is throwing. But maybe it is worth just enabling the test to see if it is still flaky.

Also I think we should remove the warning in each of the tests and always test:
```java
assertEquals(TestServlet.__maxSleepers, maxQos);
```

No one is going to be checking for warnings in tests so it is not useful.

lachlan-roberts (approved):

PR #6122, opened 2021-04-01
Issue #6109 - Introduce carriers for HttpClient transports

### LINKED ISSUE #6109: Introduce carriers for HttpClient transports
**Jetty version**
10.0.x

**Description**
Java 16 introduced UnixDomain support (#6043).
HTTP3 over QUIC is another option.

We need a way to "carry" protocol-specific bytes (e.g. HTTP/1.1 bytes or HTTP/2 bytes or HTTP/3 bytes) using different "carriers", i.e. via TCP, UnixDomain, or QUIC (over UDP).

A proposal is the following:

```java
ClientConnector clientConnector = new CompositeClientConnector();
clientConnector.add(new UnixDomainClientConnector("/path8080", "localhost:*"));
clientConnector.add(new QUICClientConnector("quic.cloudflare.com:443"));
clientConnector.add(new TCPClientConnector("*"));

HttpClient httpClient = new HttpClient(new HttpClientTransportDynamic(clientConnector));
httpClient.start();
```

The idea is that when we need to open a new carrier endpoint, we call `ClientConnector.connect(SocketAddress, Map)`.
The `CompositeClientConnector` will do a match on the carriers it manages to see which one matches best (so there must be some "weight" or partial ordering to pick one among many that match).

In this way, all the other concepts will work (e.g. SOCKS4 proxying, as well as HTTP proxying) because they will be orthogonal to the carriers.
Only at the moment that you have to physically establish a carrier endpoing, we go to the `ClientConnector`, that will decide what type of carrier creates the endpoint.

sbordet:
Introduced org.eclipse.jetty.io.Connectable.
Retrofitted HttpClientTransport implementation to take a Connectable
instead of ClientConnector.

In future, a different Connectable implementation can be passed to
HttpClientTransport implementation so that it may delegate to concrete
ClientConnector implementation that use TCP, or Unix Domain sockets,
or QUIC over UDP.

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/AbstractConnectorHttpClientTransport.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/AbstractConnectorHttpClientTransport.java
@@ -14,21 +14,22 @@
 package org.eclipse.jetty.client;
 
 import java.net.InetSocketAddress;
-import java.time.Duration;
 import java.util.Map;
 
 import org.eclipse.jetty.client.api.Connection;
 import org.eclipse.jetty.io.ClientConnector;
+import org.eclipse.jetty.io.Connectable;
 import org.eclipse.jetty.util.Promise;
 import org.eclipse.jetty.util.annotation.ManagedAttribute;
 import org.eclipse.jetty.util.annotation.ManagedObject;
+import org.eclipse.jetty.util.component.Container;
 
 @ManagedObject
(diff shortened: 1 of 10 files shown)


sbordet:
Converted back to draft, I think it needs more discussion.

sbordet:
Closed this draft PR as for Unix-Domain sockets it's not necessary.

If we need a different abstraction layer to support QUIC, then we will work it out when it's needed.

Issue #5684, opened 2020-11-17
Review disabled tests in Jetty 9.4.x


janbartel:
We should only have very few disabled tests, and they should be disabled preferably with a condition.

See Issue #6327 for tests disabled in Jetty 10.0.x

This is the list of `@Disabled` for jetty-9.4.x:

 - [ ] [jetty-client/src/test/java/org/eclipse/jetty/client/ExternalSiteTest.java:40](https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-client/src/test/java/org/eclipse/jetty/client/ExternalSiteTest.java#L40) - `@Disabled`
 - [x] [jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientRedirectTest.java:295](https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientRedirectTest.java#L295) - `    @Disabled`
 - [x] [jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTLSTest.java:266](https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTLSTest.java#L266) - `    @Disabled("No longer viable, TLS protocol behavior changed in 8u272")`
 - [ ] [jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTest.java:709](https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTest.java#L709) - `    @DisabledIfSystemProperty(named = "env", matches = "ci") // TODO: SLOW, needs review`
 - [ ] [jetty-client/src/test/java/org/eclipse/jetty/client/HttpConnectionLifecycleTest.java:220](https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-client/src/test/java/org/eclipse/jetty/client/HttpConnectionLifecycleTest.java#L220) - `    @DisabledIfSystemProperty(named = "env", matches = "ci") // TODO: SLOW, needs review`
[...]

PR #6143, opened 2021-04-07
Issue #5684 Reduce intensity and rename SameNodeLoadTest

janbartel:
* Rename the session SameNodeLoadTest to ConcurrencyTest.
* Peg back the amount of requests from 100 each for 10 clients to 10 each for 10 clients to make the test run faster
* Re-enable the test on ci.


janbartel:
nudge @olamy

olamy (approved):
great! nice simple change to not overload CI ;)

PR #6145, opened 2021-04-08
Fixes #6105 - HttpConnection.getBytesIn() incorrect for requests with chunked content

### LINKED ISSUE #6105: HttpConnection.getBytesIn() incorrect for requests with chunked content

sbordet:
… chunked content

Moved recording of bytes to fillRequestBuffer(),
so they are accounted also for async reads.
Added test case.
Fixed test that was too strictly comparing HttpConnection.bytesIn,
that now report a correct, but larger value.

(cherry picked from commit aed20abcbe157d06a7e70e0550586222b2aa433c)

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
@@ -263,9 +263,7 @@ public class HttpConnection extends AbstractConnection implements Runnable, Http
             {
                 // Fill the request buffer (if needed).
                 int filled = fillRequestBuffer();
-                if (filled > 0)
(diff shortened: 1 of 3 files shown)


sbordet:
@gregw forward port of #6108.

gregw (approved):

PR #6155, opened 2021-04-12
Issue #6153 Swap order of jetty maven plugin jvmArgs for EXTERNAL

### LINKED ISSUE #6153: jetty-maven-plugin does not correctly pass JVM arguments for external deployMode
**Jetty version**

jetty-maven-plugin: 10.0.2

**Java version**

```
$ java -version
openjdk version "11.0.11-ea" 2021-04-20
OpenJDK Runtime Environment (build 11.0.11-ea+4-post-Debian-1)
OpenJDK 64-Bit Server VM (build 11.0.11-ea+4-post-Debian-1, mixed mode, sharing)
```

**OS type/version**

debian/rolling

**Description**

I have been attempting use `jetty-maven-plugin` during development phase of a webapp but have come across an issue with the `jvmArgs` configuration parameter when running with `deployMode=EXTERNAL`.

From inspecting the code that generates the command which launches the java it treats the `jvmArgs` as arguments to the jetty jar, which means its not possible to enable the java debug server. That is, I want to set `jvmArgs` to `-agentlib:jdwp=transport=dt_socket,address=8000,server=y` so I can actually debug the code but this is instead passed to jetty and not jvm and then ignored. See excerpt of the code used to build the command line below:

https://github.com/eclipse/jetty.project/blob/e163b001c34704154edb45c7aa37908e0d9aafd7/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyHomeForker.java#L168-L187

I think it would be useful to have both a `jettyArgs` and `jvmArgs` configuration parameter as there are cases where it is useful to  pass options directly to jetty as well.

In case its of use, here is the plugin configuration (it tries to make use of both jetty and jvm arguments and fails):
```xml
          <configuration>
            <webApp>
              <contextPath>/${project.build.finalName}</contextPath>
            </webApp>
            <contextXml>${project.basedir}/jetty-context.xml</contextXml>
            <scanIntervalSeconds>10</scanIntervalSeconds>
            <scanTargets>
              <scanTarget>src/main/webapp/WEB-INF/web.xml</scanTarget>
            </scanTargets>
            <deployMode>EXTERNAL</deployMode>
            <jvmArgs>--lib=${project.basedir}/../jetty-auth/target/jetty-auth-1.0-SNAPSHOT.jar -agentlib:jdwp=transport=dt_socket,address=8000,server=y,suspend=n</jvmArgs>
            <modules>jaas,jsp</modules>
            <jettyProperties>
              <jetty.jaas.login.conf>${project.basedir}/jaas.conf</jetty.jaas.login.conf>
            </jettyProperties>
          </configuration>
```

janbartel:
Closes #6153

Move the addition of the `jvmArgs` to the forked command up to be just after `java` to ensure that they are interpreted by the jvm and not jetty.

### DIFF
--- a/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyHomeForker.java
+++ b/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyHomeForker.java
@@ -25,7 +25,6 @@ import java.nio.file.FileVisitOption;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.ArrayList;
@@ -185,9 +184,9 @@ public class JettyHomeForker extends AbstractForker
         cmd.add("java");
 
         //add any args to the jvm
-        if (!StringUtil.isBlank(jvmArgs))
+        if (StringUtil.isNotBlank(jvmArgs))
         {
-            Arrays.stream(jvmArgs.split(" ")).forEach((a) -> cmd.add(a.trim()));
+            Arrays.stream(jvmArgs.split(" ")).filter(a -> StringUtil.isNotBlank(a)).forEach((a) -> cmd.add(a.trim()));
         }
 
         cmd.add("-jar");
@@ -224,9 +223,9 @@ public class JettyHomeForker extends AbstractForker
         cmd.add(tmp.toString());
  
         //put any other jetty options onto the command line
-        if (!StringUtil.isBlank(jettyOptions))
+        if (StringUtil.isNotBlank(jettyOptions))
         {
-            Arrays.stream(jettyOptions.split(" ")).forEach((a) -> cmd.add(a.trim()));
+            Arrays.stream(jettyOptions.split(" ")).filter(a -> StringUtil.isNotBlank(a)).forEach((a) -> cmd.add(a.trim()));
         }
 
         //put any jetty properties onto the command line
--- a/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyHomeForker.java
+++ b/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyHomeForker.java
@@ -29,6 +29,7 @@ import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
@@ -74,6 +75,11 @@ public class JettyHomeForker extends AbstractForker
      */
     protected String[] modules;
 
+    /*
+     * Optional jetty commands
+     */
+    protected String jettyOptions;
+
     protected List<File> libExtJarFiles;
     protected Path modulesPath;
     protected Path etcPath;
@@ -81,6 +87,16 @@ public class JettyHomeForker extends AbstractForker
     protected Path webappPath;
     protected Path mavenLibPath;
     protected String version;
+
+    public void setJettyOptions(String jettyOptions)
+    {
+        this.jettyOptions = jettyOptions;
+    }
+
+    public String getJettyOptions()
+    {
+        return jettyOptions;
+    }
 
     public List<File> getLibExtJarFiles()
     {
@@ -169,15 +185,13 @@ public class JettyHomeForker extends AbstractForker
         cmd.add("java");
 
         //add any args to the jvm
-        if (jvmArgs != null)
+        if (!StringUtil.isBlank(jvmArgs))
         {
-            String[] args = jvmArgs.split(" ");
-            for (String a : args)
-            {
-                if (!StringUtil.isBlank(a))
-                    cmd.add(a.trim());
-            }
+            Arrays.stream(jvmArgs.split(" ")).forEach((a) -> cmd.add(a.trim()));
         }
+
+        cmd.add("-jar");
+        cmd.add(new File(jettyHome, "start.jar").getAbsolutePath());
 
         if (systemProperties != null)
         {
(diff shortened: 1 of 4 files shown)


kazkansouh:
Hello @janbartel, I have just given it a try. Yep, now the `jvmArgs` are passed to the jvm :smile: 

However, now I am unable to use the `--lib` argument which is for jetty. That said, I was misusing the parameter as it could have been managed through maven.

I think your first suggestion (in https://github.com/eclipse/jetty.project/issues/6153#issuecomment-817403075) of using the `--exec` is probably better as it maximises flexibility, albeit with nomenclature abuse (i.e. `jvmArgs` as jetty args). For me, the ideal solution would be to have both a `jvmArgs` and `jettyArgs` configuration parameter, but I am not sure whether its worth the effort to implement as it can be worked around?

joakime:
> However, now I am unable to use the `--lib` argument which is for jetty. That said, I was misusing the parameter as it could have been managed through maven.

Can you use a normal plugin `<plugin>/<dependencies>/<dependency>` instead of `--lib` ?

joakime (dismissed):
Good change.

We should probably fix the jvmArgs property to not need to split on space (in a future PR)

kazkansouh:
> Can you use a normal plugin `<plugin>/<dependencies>/<dependency>` instead of `--lib` ?

Yes I can, thanks (and had to refactor along these lines to test this PR). I just thought it might be good to allow flags to be passed down to jetty ...

joakime:
> > Can you use a normal plugin `<plugin>/<dependencies>/<dependency>` instead of `--lib` ?
> 
> Yes I can, thanks (and had to refactor along these lines to test this PR). I just thought it might be good to allow flags to be passed down to jetty ...

It's just always been easier to work with the classloader in maven, using the maven way, then it is to manipulate it after the fact.

janbartel:
@joakime  I added a new `<jettyOptions>` config param so you can pass extra commands to jetty for `EXTERNAL` mode. Please rereview.

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyHomeForker.java
```diff
@@ -169,16 +185,14 @@ protected ProcessBuilder createCommand()
...
-        if (jvmArgs != null)
+        if (!StringUtil.isBlank(jvmArgs))
         {
-            String[] args = jvmArgs.split(" ");
-            for (String a : args)
-            {
-                if (!StringUtil.isBlank(a))
-                    cmd.add(a.trim());
-            }
+            Arrays.stream(jvmArgs.split(" ")).forEach((a) -> cmd.add(a.trim()));
```
joakime: This doesn't eliminate blank entries from the `jvmArgs.split(" ")` call.

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyHomeForker.java
```diff
@@ -74,6 +75,11 @@
      */
     protected String[] modules;
 
+    /*
+     * Optional jetty commands
+     */
+    protected String jettyOptions;
```
joakime, line 80: Make this `protected String[] jettyOptions` and don't split on space.

joakime (changes requested):
Lets use the plugin parameters correctly so that we don't need to split on space, bonus is that we can now support options that have a space in the name.

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyHomeForker.java
olamy: but jvmArgs will be different? 
plugin such surefire (`argLine`) use String to avoid too big xml :) 
I vote for `String`  and using command line to pass parameters will be more complicated.

joakime: Then the split needs to be more advanced.

* not just on space.
* It should take into account escaped spaces. (be careful here `"program\\ files"` and `"program\ files"` are different behaviors)
* It should also honor quoted sequences. (that can have space in them. eg `--lib=foo.jar "context files/corp audit.xml"`).
* It should eliminate parts that are functionally blank, which is trivially easy to trigger on both the command line and in the XML using a single String. (all whitespace.  eg `--lib=foo.jar  \t  another.xml` or `--lib=foo.jar\n\tzed.jar`)

I'm not convinced that supporting `argLine` is an sufficient argument for this.

janbartel: The `jvmArgs` has been a single single string argument for years so far with no reported problems. The surefire plugin also uses just a single string for its jvm arg line. I don't see what's different about the jetty command args that would warrant making it so much more verbose to configure in the poms. Plus I think the `jvmArgs` and the new `jettyOptions` should both be the same: changing to `String[]` would be a breaking change for the `jvmArgs`.

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyHomeForker.java
```diff
@@ -167,24 +183,16 @@ protected ProcessBuilder createCommand()
...
-        if (jvmArgs != null)
+        if (!StringUtil.isBlank(jvmArgs))
         {
-            String[] args = jvmArgs.split(" ");
-            for (String a : args)
-            {
-                if (!StringUtil.isBlank(a))
-                    cmd.add(a.trim());
-            }
+            Arrays.stream(jvmArgs.split(" ")).forEach((a) -> cmd.add(a.trim()));
```
joakime: Old code also eliminated blank parts.
New code should too.

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyHomeForker.java
```diff
@@ -210,6 +222,12 @@ protected ProcessBuilder createCommand()
             tmp.append(",ext");
         tmp.append(",maven");
         cmd.add(tmp.toString());
+ 
+        //put any other jetty options onto the command line
+        if (!StringUtil.isBlank(jettyOptions))
+        {
+            Arrays.stream(jettyOptions.split(" ")).forEach((a) -> cmd.add(a.trim()));
```
joakime: Eliminate blank parts (like the old code for jvmArgs used to do)

joakime (changes requested):
Fix stream/add to also eliminate blank parts from the `split(" ")` call, like the old code used to do.

joakime (approved):

PR #6163, opened 2021-04-13
Jetty 9.4.x unconsumed content2

### LINKED ISSUE #6168: Improve handling of unconsumed content
**Jetty version**
9.4.x

**Description**
The case of unconsumed content, and `HttpInput.Interceptor` throwing an exception, should be handled better.
In particular, make sure that exceptions thrown by the interceptor are correctly handled.

sbordet:
Improved behavior in case of HttpInput.Interceptor throwing.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
@@ -366,7 +366,6 @@ public class HttpConnection extends AbstractConnection implements Runnable, Http
             }
         }
 
-        // TODO: must return _requestBuffer.remaining()!
         return 0;
     }
 
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
@@ -295,6 +295,14 @@ public class HttpConnection extends AbstractConnection implements Runnable, Http
                 }
             }
         }
+        catch (Throwable x)
+        {
+            if (LOG.isDebugEnabled())
+                LOG.debug("{} caught exception {}", this, _channel.getState(), x);
+            BufferUtil.clear(_requestBuffer);
+            releaseRequestBuffer();
+            close();
+        }
         finally
         {
             setCurrentConnection(last);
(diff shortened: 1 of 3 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -354,11 +359,14 @@ private int fillRequestBuffer()
...
             {
-                LOG.debug(e);
+                if (LOG.isDebugEnabled())
+                    LOG.debug(e);
                 _parser.atEOF();
                 return -1;
             }
         }
+
+        // TODO: must return _requestBuffer.remaining()!
```
gregw: Can you explain this TODO more?

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -295,6 +295,14 @@ else if (filled < 0)
...
             }
         }
+        catch (Throwable x)
+        {
+            if (LOG.isDebugEnabled())
+                LOG.debug("{} caught exception {}", this, _channel.getState(), x);
+            BufferUtil.clear(_requestBuffer);
+            releaseRequestBuffer();
+            close();
+        }
```
gregw, line 305: This is the only one of the new catch clauses  added that does not actually pass on the caught exception.
So perhaps add a comment saying why the close is enough... Actually, why are we not calling `getEndPoint().close(x);` ??

gregw (dismissed):
The core of this fix is good.
I remain very cautious about the extra catch blocks and have included some niggles.
I think we have more work to do even after  merging this one to proactively explore the error space more.  So please open another issue, which together with @lachlan-roberts work on interceptors,  will explore the error issues more.

lorban (dismissed):
LGTM

sbordet:
@gregw opened #6167.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
sbordet: It was a leftover.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
sbordet: Because there is no such method `EndPoint.close(Throwable)` in 9.4.x (it's only in 10).

gregw: @lachlan-roberts @lorban can you fix in 10 to pass the throwable to close in your work on #6167

lorban: @gregw will do.

Issue #6166, opened 2021-04-13
WebSocket MessageInputStream.read() spends a lot of time in ByteBuffer.compact()


lachlan-roberts:
**Jetty version**
10.0.2

**Description**
Originally reported by @dcitron in issue #6127.

> I did notice that `org.eclipse.jetty.websocket.core.internal.messages.MessageInputStream.read()` is taking a larger percentage of time in Jetty 10 than Jetty 9. In Jetty 9 it kicked around in `java.nio.Buffer.hasRemaining()` a bit, but in Jetty 10 it's taking quite a bit of time in the following call stack:
> ```
> MessageInputStream.read() 
>     -> org.eclipse.jetty.util.BufferUtil.append() 
>          -> org.eclipse.jetty.util.BufferUtil.flipToFill() 
>               -> java.nio.HeapByteBuffer.compact()
> ```
> In a simple test that I ran, processing 1,000 messages, it actually seemed to spend 24% of its time in `HeapByteBuffer.compact()` so maybe that's something you might want to look into further, or maybe it's just a memory-vs-CPU tradeoff.

lachlan-roberts:
This is fixed by #6170.

PR #6170, opened 2021-04-14
Issue #6166 - fix bug in the MessageInputStream byte array read

lachlan-roberts:
**Issue #6166**

The code for `MessageInputStream.read(final byte[] b, final int off, final int len)` seems to be wrong. Instead of doing `ByteBuffer.wrap(b, off, len).flip()` we should do something like:

```java
ByteBuffer buffer = ByteBuffer.wrap(b, off, len).slice();
BufferUtil.clear(buffer);
return read(buffer);
```

This will ensure that the buffer we get to represent the byte array always has a position and limit of 0, an offset of `off`, and a capacity of `len`.

The only time when `MessageInputStream.read(ByteBuffer)` will compact is when the buffer already has content and `limit == capacity`. So if we make this change, a call to `MessageInputStream.read(final byte[] b, final int off, final int len)` will never compact as the buffer is always empty with `position == limit`.

### DIFF
--- a/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/MessageInputStream.java
+++ b/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/MessageInputStream.java
@@ -95,7 +95,9 @@ public class MessageInputStream extends InputStream implements MessageSink
     @Override
     public int read(final byte[] b, final int off, final int len) throws IOException
     {
-        return read(ByteBuffer.wrap(b, off, len).flip());
+        ByteBuffer buffer = ByteBuffer.wrap(b, off, len).slice();
+        BufferUtil.clear(buffer);
+        return read(buffer);
     }
 
     public int read(ByteBuffer buffer) throws IOException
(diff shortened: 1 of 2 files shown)


jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/MessageInputStream.java
```diff
@@ -95,7 +95,9 @@ public int read() throws IOException
     @Override
     public int read(final byte[] b, final int off, final int len) throws IOException
     {
-        return read(ByteBuffer.wrap(b, off, len).flip());
+        ByteBuffer buffer = ByteBuffer.wrap(b, off, len).slice();
```
gregw, line 98: Why slice???

lachlan-roberts: Slice creates a new buffer with the capacity set to the limit of the previous buffer.
So the capacity of the buffer will be `len`, and not `b.length`. This is important so we never write more bytes than `len`.

sbordet (approved):

gregw (approved):

PR #6173, opened 2021-04-14
Fix #6157 consumeAll optimisation

### LINKED ISSUE #6157: 10.0.x: Perf regression in HttpInput.consumeAll()
**Jetty version**
10.0.x (all versions impacted)

**Description**
During the `HttpInput` refactoring that happened between 9.4.x and 10.0.x, a few harmless minor things were changed. Among those, the `HttpInput.consumeAll()` method used to succeed unconsumed content in 9.4.x while in 10.0.x unconsumed content is now failed.

This is pretty much equivalent in practice, even if the 10.0.x behavior is 'more correct' stricto sensu. except that failing content requires passing an exception instance.

This is exactly where the problem lies: in 10.0.x, `HttpInput.consumeAll()` creates a new `IOException` instance unconditionally. Since this method is on the fast path and creating an exception isn't cheap because of the stack trace collection, there's a visible perf difference between the two versions.

gregw:
Fix #6157 consumeAll no longer always creates a new Throwable

### DIFF
--- a/jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/HttpChannelOverFCGI.java
+++ b/jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/HttpChannelOverFCGI.java
@@ -20,6 +20,7 @@ import java.util.Locale;
 import java.util.Queue;
 import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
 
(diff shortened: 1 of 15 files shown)


gregw:
This changes some APIs and reaches deep inside HTTP2.... but can't think of a better way unless we revert to not failing content?

lorban (changes requested):
This change looks overly complicated. Why not simply using a statically allocated exception in consumeAll? The stack trace is irrelevant as it's never going to flow out to the user.

sbordet:
Closed in favor of #6197.

PR #6176, opened 2021-04-15
Issue #5817 - allow CustomRequestLog to be filtered with BiPredicate

### LINKED ISSUE #5817: Provide more filtering for CustomRequestLog
**Jetty version**
9.4.35+

**Description**
The current `CustomRequestLog` has only 1 filter option, which is the request uri path.

Provide more filtering options.

* Response status code (very useful)
* Content-Type (can narrow down specific behaviors)
* User-Agent (can identify specific user agents you are interested in)

I think just supporting a `BiPredicate<Request,Response>` on `CustomRequestLog` which is used in `.log(Request,Response)` should be the maximally flexible option.
Complex rules can be done with BiPredicate composition (`.and()`, `.or()`, and `.not()`) or with custom `BiPredicate` behaviors.

lachlan-roberts:
**closes #5817**

Allow filtering in `CustomRequestLog`. You can now set a `BiPredicate<Request,Response>` which will be evaluated for each request to be logged, it will only log to the `RequestLog` if the `BiPredicate` returns true.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/CustomRequestLog.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/CustomRequestLog.java
@@ -22,6 +22,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
+import java.util.function.BiPredicate;
 import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/CustomRequestLog.java
```diff
@@ -311,6 +313,15 @@ public CustomRequestLog(RequestLog.Writer writer, String formatString)
         }
     }
 
+    /**
+     * This allows you to set a custom filter to decide whether to log a request or omit it from the request log.
+     * @param filter - a BiPredicate which returns true if this request should be logged.
+     */
```
joakime, line 320: Might need to describe that this filter is evaluated after the other filtering, such as ignore paths?

joakime (changes requested):
Please fix the javadoc.

joakime (approved):

Issue #6020, opened 2021-03-02
Review Jetty Maven Plugin scanning defaults 


sbordet:
**Jetty version**
10.0.x

**Description**
By default, the Jetty Maven Plugin has `scan=0` which means when running `jetty:run` on a terminal, the following is printed:

```
Hit <enter> to redeploy:
```

However, this default is confusing because hitting `<enter>` on the terminal is useful to separate start-up log lines from application log lines, or application log lines between different requests.

Setting `scan=5` disables the above "hit enter" but prints:

```
[INFO] Scan interval ms = 5
```

The time unit is wrong, should be `seconds`.

Then, modifying `web.xml` and waiting for the re-scan, the following is printed:

```
[INFO] Restarting o.e.j.m.p.MavenWebAppContext@2c54c497{CometD Test WebApp,/,[file:///home/simon/opensource/cometd/cometd6/cometd-demo/src/main/webapp/, file:///home/simon/opensource/cometd/cometd6/cometd-demo/target/jetty_overlays/cometd-javascript-dojo-6_0_2-SNAPSHOT_war/, file:///home/simon/opensource/cometd/cometd6/cometd-demo/target/jetty_overlays/cometd-javascript-jquery-6_0_2-SNAPSHOT_war/, file:///home/simon/opensource/cometd/cometd6/cometd-demo/target/jetty_overlays/cometd-javascript-angular-6_0_2-SNAPSHOT_war/, file:///home/simon/opensource/cometd/cometd6/cometd-demo/target/jetty_overlays/cometd-javascript-examples-dojo-6_0_2-SNAPSHOT_war/, file:///home/simon/opensource/cometd/cometd6/cometd-demo/target/jetty_overlays/cometd-javascript-examples-jquery-6_0_2-SNAPSHOT_war/, file:///home/simon/opensource/cometd/cometd6/cometd-demo/target/jetty_overlays/cometd-javascript-examples-angular-6_0_2-SNAPSHOT_war/, file:///home/simon/opensource/cometd/cometd6/cometd-demo/target/jetty_overlays/cometd-javascript-examples-vanilla-6_0_2-SNAPSHOT_war/],AVAILABLE}{file:///home/simon/opensource/cometd/cometd6/cometd-demo/src/main/webapp/}
[INFO] Classes = /home/simon/opensource/cometd/cometd6/cometd-demo/target/classes
[ERROR] Error reconfiguring/restarting webapp after change in watched files
[...]

PR #6182, opened 2021-04-15
Updated Deployment Not enabled console message 

dana-whitelow:
Updated Deployment console message for a clearer status and instructions to turn on and change redeployment status for better user experience in
jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyRunMojo.java

### DIFF
--- a/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyRunMojo.java
+++ b/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyRunMojo.java
@@ -144,7 +144,7 @@ public class JettyRunMojo extends AbstractUnassembledWebAppMojo
     {
         if (scan < 0)
         {
-            getLog().info("Redeployment Disabled; Default Scan Interval = -1 ; Change Value to <scan>0<scan> for Manual or <scan>${Seconds}<scan> for automatic  ");
+            getLog().info("default deploy mode scan=-1:redeploy disabled; scan=0:redeploy manual on ENTER; scan > 0:redeploy automatic");
             return; //no automatic or manual redeployment
         }
         
--- a/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyRunMojo.java
+++ b/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyRunMojo.java
@@ -144,7 +144,7 @@ public class JettyRunMojo extends AbstractUnassembledWebAppMojo
     {
         if (scan < 0)
         {
-            getLog().info("Redeployment not enabled");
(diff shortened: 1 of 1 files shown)


jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyRunMojo.java
```diff
@@ -144,7 +144,7 @@ private void startScanner()
     {
         if (scan < 0)
         {
-            getLog().info("Redeployment not enabled");
+            getLog().info("Redeployment Disabled; Default Scan Interval = -1 ; Change Value to <scan>0<scan> for Manual or <scan>${Seconds}<scan> for automatic  ");
```
janbartel: Thanks for the pull request, I understand you're looking for more configuration info (which is in the doco btw). I think this log line might be a little wordy, how about something a bit more in line with the other configuration summaries the plugin produces, that is a little more succinct like:  

"scan=-1 redeploy disabled (scan=0:redeploy on ENTER; scan > 0:automatic redeploy check sec)"

janbartel (changes requested):

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyRunMojo.java
dana-whitelow: I would like to see  the word default because there is now behaviour change in this later version 11.0.2 compare to earlier versions with manual deploy was the default.  

Good user experience is where there is immediate cause and solution rather than having to look at documents, In my opinion.     

default scan=-1:redeploy disabled; scan=0:redeploy manual on ENTER; scan > 0:redeploy automatic

janbartel:
Some info for you: we just changed the default a few weeks ago. Previously all versions of jetty were no redeploy by default. With 10.0.0/11.0.0 we changed it to manual redeploy by default, and then we changed it back again to no redeploy for 11.0.2 with this issue #6020.

+1 if you want to use the word "default" in the message.

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyRunMojo.java
dana-whitelow: default deploy mode scan=-1:redeploy disabled;  scan=0:redeploy manual on ENTER;  scan > 0:redeploy automatic

I am happy with this. The default also may alert the user that there is other functionality available. 
Hot Deploy is a very useful feature, even a hyped!  feature in some circles, needs to be brought to the forefront attention of the user.
The user being a developer he will know implications into with word  default.
I did not suggest a changing the default behaviour because I am sure you had a good reason  for it. 
Personally my preference is default mode to be manual deploy mode.
may be add the word " default deploy mode" also with default ?

janbartel:
Closed in favour of other PR #6188

PR #6188, opened 2021-04-16
default deploy mode 

dana-whitelow:
inform user of default deployment mode feature is turned off,
how to turn on. Restructure wording

### DIFF
--- a/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyRunMojo.java
+++ b/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyRunMojo.java
@@ -144,7 +144,7 @@ public class JettyRunMojo extends AbstractUnassembledWebAppMojo
     {
         if (scan < 0)
         {
-            getLog().info("Redeployment not enabled");
+            getLog().info("default scan=-1:redeploy disabled; scan=0:redeploy manual on ENTER; scan > 0:redeploy automatic");
             return; //no automatic or manual redeployment
         }
         


joakime:
This seems to be a duplicate of #6182

dana-whitelow:
I  did a new one after the discussion Jan.

 

 
 

Sent: Friday, April 16, 2021 at 7:51 PM

sbordet:
The log line is printed only when `scan < 0`. 

If it is meant to detail how the plugin works, then it should always be printed.

But printing it all the times when it's actually well documented seems unnecessary?

I would rather add a clarification line about the redeployment options to `jetty:help` rather than printing it all the times.
That would make it more alike to other tools: you print the help and you see detailed information about what each option means.

janbartel:
@sbordet the plugin already documents all of the goals and all of the configurable options.

For example if you do:  `mvn help:describe -DgroupId=org.eclipse.jetty -DartifactId=jetty-maven-plugin -Dversion=10.0.3-SNAPSHOT -Dgoal=run -Ddetail`

You get the following (scroll down to see the `scan` parameter output:

```
jetty:run
  Description: This goal is used in-situ on a Maven project without first
    requiring that the project is assembled into a war, saving time during the
    development cycle.
    The plugin runs a parallel lifecycle to ensure that the 'test-compile'
    phase has been completed before invoking Jetty. This means that you do not
    need to explicity execute a 'mvn compile' first. It also means that a 'mvn
    clean jetty:run' will ensure that a full fresh compile is done before
    invoking Jetty.
    
    Once invoked, the plugin can be configured to run continuously, scanning
    for changes in the project and automatically performing a hot redeploy when
    necessary. This allows the developer to concentrate on coding changes to
    the project using their IDE of choice and have those changes immediately
    and transparently reflected in the running web container, eliminating
    development time that is wasted on rebuilding, reassembling and
    redeploying. Alternatively, you can configure the plugin to wait for an
    <enter> at the command line to manually control redeployment.
    
    You can configure this goal to run your unassembled webapp either
    in-process with maven, or forked into a new process, or deployed into a
    jetty distribution.
  Implementation: org.eclipse.jetty.maven.plugin.JettyRunMojo
  Language: java
  Before this mojo executes, it will call:
    Phase: 'test-compile'

  Available parameters:

    classesDirectory (Default: ${project.build.outputDirectory})
      Required: true
      The directory containing generated classes.

    contextHandlers
      List of other contexts to set up. Consider using instead the <jettyXml>
      element to specify external jetty xml config file. Optional.

    contextXml
      Location of a context xml configuration file whose contents will be
      applied to the webapp AFTER anything in <webApp>.Optional.

    deployMode (Default: EMBED)
      User property: jetty.deployMode
      Controls how to run jetty. Valid values are EMBED,FORK,EXTERNAL.

    env
      Extra environment variables to be passed to the forked process

    excludedGoals
      List of goals that are NOT to be used

    forkWebXml (Default: ${project.build.directory}/fork-web.xml)
      The file into which to generate the quickstart web xml for the forked
      process to use

    httpConnector
      A ServerConnector to use.

    jettyBase
      Location of jetty base directory

    jettyHome
      Location of jetty home directory

    jettyProperties
      Optional jetty properties to put on the command line

    jettyXmls
      List of jetty xml configuration files whose contents will be applied (in
      order declared) before any plugin configuration. Optional.

    jvmArgs
      User property: jetty.jvmArgs
      Arbitrary jvm args to pass to the forked process

    loginServices
      List of security realms to set up. Consider using instead the <jettyXml>
      element to specify external jetty xml config file. Optional.

    maxChildStartCheckMs (Default: 200)
      How long to wait in msec between checks to see if jetty has started
      correctly when running in FORK or EXTERNAL mode.

    maxChildStartChecks (Default: 10)
      Max number of times to check to see if jetty has started correctly when
      running in FORK or EXTERNAL mode.

    modules
      Optional list of other modules to activate.

    requestLog
      A RequestLog implementation to use for the webapp at runtime. Consider
      using instead the <jettyXml> element to specify external jetty xml config
      file. Optional.

    scan (Default: -1)
      Required: true
      User property: jetty.scan
      Controls redeployment of the webapp.
      1.  -1 : means no redeployment will be done
      2.  0 : means redeployment only occurs if you hit the ENTER key
      3.  otherwise, the interval in seconds to pause before checking and
        redeploying if necessary

    scanClassesPattern
      An optional pattern for includes/excludes of classes in the
      classesDirectory

    scanTargetPatterns
      List of directories with ant-style <include> and <exclude> patterns for
      extra targets to periodically scan for changes.Optional.

    scanTestClassesPattern
      An optional pattern for includes/excludes of classes in the
      testClassesDirectory

    server
      A wrapper for the Server object

    skip (Default: false)
      User property: jetty.skip
      Skip this mojo execution.

    stopKey
      Key to provide when stopping jetty on executing java -DSTOP.KEY=<stopKey>
      -DSTOP.PORT=<stopPort> -jar start.jar --stop

    stopPort
      Port to listen to stop jetty on executing -DSTOP.PORT=<stopPort>
      -DSTOP.KEY=<stopKey> -jar start.jar --stop

    supportedPackagings
      Default supported project type is war packaging.

    systemProperties
      System properties to set before execution. Note that these properties
      will NOT override System properties that have been set on the command
      line or by the JVM. They WILL override System properties that have been
      set via systemPropertiesFile. Optional.

    systemPropertiesFile
      User property: jetty.systemPropertiesFile
      File containing system properties to be set before execution Note that
      these properties will NOT override System properties that have been set
      on the command line, by the JVM, or directly in the POM via
      systemProperties. Optional.

    testClassesDirectory (Default: ${project.build.testOutputDirectory})
      Required: true
      The directory containing generated test classes.

    useProvidedScope (Default: false)
      Whether or not to include dependencies on the plugin's classpath with
      <scope>provided</scope> Use WITH CAUTION as you may wind up with
      duplicate jars/classes.

    useTestScope (Default: false)
      If true, the <testOutputDirectory> and the dependencies of
      <scope>test<scope> will be put first on the runtime classpath.

    webApp
      An instance of org.eclipse.jetty.webapp.WebAppContext that represents the
      webapp. Use any of its setters to configure the webapp. This is the
      preferred and most flexible method of configuration, rather than using
      the (deprecated) individual parameters like 'tmpDirectory', 'contextPath'
      etc.

    webXml (Default: ${project.baseDir}/src/main/webapp/WEB-INF/web.xml)
      The default location of the web.xml file. Will be used if
      <webApp><descriptor> is not set.
```

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyRunMojo.java
```diff
@@ -144,7 +144,7 @@ private void startScanner()
     {
         if (scan < 0)
         {
-            getLog().info("Redeployment not enabled");
+            getLog().info("default scan=-1:redeploy disabled; scan=0:redeploy manual on ENTER; scan > 0:redeploy automatic");
```
sbordet, line 147: @dana-whitelow can you please change this, rather than hardcoding the actual values in the log text, to be something like "Redeployment disabled. See jetty:help for other configuration options."?
Sorry and thanks.

sbordet (changes requested):

joakime:
> @sbordet the plugin already documents all of the goals and all of the configurable options.
> 
> For example if you do: `mvn help:describe -DgroupId=org.eclipse.jetty -DartifactId=jetty-maven-plugin -Dversion=10.0.3-SNAPSHOT -Dgoal=run -Ddetail`

For others that look at this PR in the future, there's an alternate syntax you can use here as well ...

``` shell
$ mvn org.eclipse.jetty:jetty-maven-plugin:10.0.2:help
```

or with detailed information about the goal parameters ...

``` shell
$ mvn org.eclipse.jetty:jetty-maven-plugin:10.0.2:help -Ddetail
```

PR #6197, opened 2021-04-19
consumeAll: fail content with static exception

lorban:
Simplest possible fix for #6157.

This compares to @gregw 's https://github.com/eclipse/jetty.project/pull/6173.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
@@ -30,6 +30,14 @@ import org.slf4j.LoggerFactory;
 class AsyncContentProducer implements ContentProducer
 {
     private static final Logger LOG = LoggerFactory.getLogger(AsyncContentProducer.class);
+    private static final Throwable UNCONSUMED_CONTENT_EXCEPTION = new IOException("Unconsumed content")
+    {
+        @Override
+        public synchronized Throwable fillInStackTrace()
+        {
+            return this;
+        }
+    };
 
     private final AutoLock _lock = new AutoLock();
     private final HttpChannel _httpChannel;
@@ -151,11 +159,15 @@ class AsyncContentProducer implements ContentProducer
     }
 
     @Override
-    public boolean consumeAll(Throwable x)
+    public boolean consumeAll()
     {
         assertLocked();
+        Throwable x = UNCONSUMED_CONTENT_EXCEPTION;
         if (LOG.isDebugEnabled())
-            LOG.debug("consumeAll [e={}] {}", x, this);
+        {
+            x = new IOException("Unconsumed content");
+            LOG.debug("consumeAll {}", this, x);
+        }
         failCurrentContent(x);
         // A specific HttpChannel mechanism must be used as the following code
         // does not guarantee that the channel will synchronously deliver all
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
@@ -150,10 +150,9 @@ public class HttpInput extends ServletInputStream implements Runnable
     {
         try (AutoLock lock = _contentProducer.lock())
         {
-            IOException failure = new IOException("Unconsumed content");
             if (LOG.isDebugEnabled())
-                LOG.debug("consumeAll {}", this, failure);
-            boolean atEof = _contentProducer.consumeAll(failure);
+                LOG.debug("consumeAll {}", this);
+            boolean atEof = _contentProducer.consumeAll();
             if (atEof)
                 _consumedEof = true;
 
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/BlockingContentProducer.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/BlockingContentProducer.java
@@ -80,9 +80,9 @@ class BlockingContentProducer implements ContentProducer
     }
 
     @Override
-    public boolean consumeAll(Throwable x)
+    public boolean consumeAll()
     {
-        boolean eof = _asyncContentProducer.consumeAll(x);
+        boolean eof = _asyncContentProducer.consumeAll();
         _semaphore.release();
         return eof;
     }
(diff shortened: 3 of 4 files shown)


lorban:
Here's a flame graph that clearly shows that exception's instantiation time is gone: [flame graph.zip](https://github.com/eclipse/jetty.project/files/6336988/flame.graph.zip). To be compared to the one in #6157.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
```diff
@@ -150,10 +151,9 @@ public boolean consumeAll()
     {
         try (AutoLock lock = _contentProducer.lock())
         {
-            IOException failure = new IOException("Unconsumed content");
             if (LOG.isDebugEnabled())
-                LOG.debug("consumeAll {}", this, failure);
-            boolean atEof = _contentProducer.consumeAll(failure);
+                LOG.debug("consumeAll {}", this);
+            boolean atEof = _contentProducer.consumeAll(CONSUME_ALL_IOEXCEPTION);
```
gregw: Can we use a dynamic exception of debug is turned on?

gregw (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
lorban: Sure.

lorban: Looking more carefully at this code, since `consumeAll()` is ran under a lock we could simply have a field that we initialize in `consumeAll()` with a simple `if (null) new Exception` pattern. WDYT?

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
```diff
@@ -150,10 +151,11 @@ public boolean consumeAll()
...
         try (AutoLock lock = _contentProducer.lock())
         {
-            IOException failure = new IOException("Unconsumed content");
+            if (_unconsumedContentException == null)
+                _unconsumedContentException = new IOException("Unconsumed content");
             if (LOG.isDebugEnabled())
-                LOG.debug("consumeAll {}", this, failure);
-            boolean atEof = _contentProducer.consumeAll(failure);
+                LOG.debug("consumeAll {}", this);
+            boolean atEof = _contentProducer.consumeAll(_unconsumedContentException);
```
sbordet: Can't we move the exception inside `ContentProducer`?

sbordet (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
```diff
@@ -151,11 +159,15 @@ public long getRawContentArrived()
...
-    public boolean consumeAll(Throwable x)
+    public boolean consumeAll()
     {
         assertLocked();
+        Throwable x = UNCONSUMED_CONTENT_EXCEPTION;
         if (LOG.isDebugEnabled())
-            LOG.debug("consumeAll [e={}] {}", x, this);
+        {
+            x = new IOException("Unconsumed content");
+            LOG.debug("consumeAll {}", this);
```
sbordet: I would log `x` too.

sbordet (dismissed):
Besides logging the exception as detailed in the comment, LGTM.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
lorban: As discussed, I've done it.

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
lorban: Added.

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
```diff
@@ -166,7 +166,7 @@ public boolean consumeAll()
         if (LOG.isDebugEnabled())
         {
             x = new IOException("Unconsumed content");
-            LOG.debug("consumeAll {}", this);
+            LOG.debug("consumeAll [e={}] {}", x, this);
```
joakime: This would be better as ...

``` java
LOG.debug("consumeAll {}", this, x);
```

lorban: Do you think logging this exception's stack trace here would be useful?

sbordet: Yes, that is the whole point of creating a fresh new `IOException`, to capture the stack trace so that we know from where the call comes from.
If we don't print the stack trace, then there is no point creating a fresh new `IOException`.

gregw (approved):

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
```diff
@@ -30,6 +30,14 @@
 class AsyncContentProducer implements ContentProducer
 {
     private static final Logger LOG = LoggerFactory.getLogger(AsyncContentProducer.class);
+    private static final Throwable UNCONSUMED_CONTENT_EXCEPTION = new IOException("Unconsumed content")
+    {
+        @Override
+        public synchronized Throwable fillInStackTrace()
```
ijuma, line 36: What is the reason for the `synchronized` here?

lorban: It is part of the signature of the `super` implementation, so it was kept by accident. Since it's useless, it got removed in a subsequent revision.

ijuma: Sounds good.

joakime: This is the current `jetty-10.0.x` of the same UNCONSUMED_CONTENT_EXCEPTION 

https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java#L33-L40

https://github.com/eclipse/jetty.project/blob/b139cbfc7bafd23689b24a406290bec2fa14070d/jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java#L33-L40

Jetty releases since 10.0.4 has the code block without the synchronized modifier.

PR #6198, opened 2021-04-20
Ensure that javax.websocket Encoder/Decoder destroy method is called.

lachlan-roberts:

### DIFF
--- a/jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/decoders/AvailableDecoders.java
+++ b/jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/decoders/AvailableDecoders.java
@@ -207,4 +207,9 @@ public class AvailableDecoders implements Iterable<RegisteredDecoder>
     {
         return registeredDecoders.stream();
     }
+
+    public void close()
+    {
+        registeredDecoders.forEach(RegisteredDecoder::destroyInstance);
+    }
 }
--- a/jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/decoders/AvailableDecoders.java
+++ b/jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/decoders/AvailableDecoders.java
@@ -13,6 +13,7 @@
 
 package org.eclipse.jetty.websocket.javax.common.decoders;
 
+import java.io.Closeable;
 import java.io.InputStream;
 import java.io.Reader;
 import java.nio.ByteBuffer;
@@ -29,7 +30,7 @@ import org.eclipse.jetty.websocket.core.exception.InvalidSignatureException;
 import org.eclipse.jetty.websocket.core.exception.InvalidWebSocketException;
 import org.eclipse.jetty.websocket.core.internal.util.ReflectUtils;
 
-public class AvailableDecoders implements Iterable<RegisteredDecoder>
+public class AvailableDecoders implements Iterable<RegisteredDecoder>, Closeable
 {
     private final List<RegisteredDecoder> registeredDecoders = new ArrayList<>();
     private final EndpointConfig config;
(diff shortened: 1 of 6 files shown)


jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/decoders/RegisteredDecoder.java
```diff
@@ -74,6 +78,23 @@ public boolean isType(Class<?> type)
         return (T)instance;
     }
 
+    public void destroyInstance()
```
sbordet, line 81: We have an interface `Destroyable`, any reason to not use it?

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/decoders/AvailableDecoders.java
```diff
@@ -207,4 +207,9 @@ public RegisteredDecoder getFirstRegisteredDecoder(Class<?> type)
     {
         return registeredDecoders.stream();
     }
+
+    public void close()
```
sbordet, line 213: Implement `Closeable` then?

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/encoders/AvailableEncoders.java
```diff
@@ -62,6 +67,23 @@ public boolean isType(Class<?> type)
             return objectType.isAssignableFrom(type);
         }
 
+        public void destroyInstance()
```
sbordet, line 71: Ditto

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/encoders/AvailableEncoders.java
```diff
@@ -286,4 +308,9 @@ public boolean test(Class<?> type)
     {
         return registeredEncoders.stream().anyMatch(registered -> registered.isType(type));
     }
+
+    public void close()
```
sbordet, line 314: Ditto

sbordet (changes requested):

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/decoders/RegisteredDecoder.java
joakime: Is the `Destroyable` interface exposed to the WebApp?

lachlan-roberts: > Is the Destroyable interface exposed to the WebApp?

I don't think it needs to be. `RegisteredDecoder` is an internal implementation anyway not something used by the WebApp.

lachlan-roberts: I don't think there is any benefit from implementing `Destroyable` though, we don't really need the JMX operations for destroy on this.

This usage doesn't really line up with the description of `Destroyable` from the javadoc:
> Typically a Destroyable is a {@link LifeCycle} component that can hold onto
>  resources over multiple start/stop cycles. A call to destroy will release all
>  resources and will prevent any further start/stop cycles from being successful.

sbordet: Fair enough.

sbordet (approved):

joakime (approved):

PR #6202, opened 2021-04-20
Documented WebSocketClient.

sbordet:


documentation/jetty-documentation/src/main/asciidoc/programming-guide/client/websocket/client-websocket.adoc
```diff
@@ -14,4 +14,188 @@
...
+
+However, it is recommended that you explicitly pass an `HttpClient` instance to `WebSocketClient` so that you can have control over the HTTP configuration as well:
+
+[source,java,indent=0]
+----
+include::../../{doc_code}/org/eclipse/jetty/docs/programming/client/websocket/WebSocketClientDocs.java[tags=startWithHttpClient]
+----
+
+You may create multiple instances of `WebSocketClient`, but typically one instance is enough for most applications.
+Creating multiple instances may be necessary for example when you need to specify different configuration parameters for different instances.
```
lachlan-roberts, line 54: I think this needs to be more specific, most configuration on the websocket client can also be configured through the Session, so there would be no need for multiple WebSocketClients.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/client/websocket/WebSocketClientDocs.java
```diff
@@ -0,0 +1,193 @@
...
+
+        // Create and start WebSocketClient.
+        WebSocketClient webSocketClient = new WebSocketClient(httpClient);
+        webSocketClient.start();
+
+        // The client-side WebSocket EndPoint that
+        // receives WebSocket messages from the server.
+        ClientEndPoint clientEndPoint = new ClientEndPoint();
+        // The server URI to connect to.
+        URI serverURI = URI.create("ws://domain.com/path");
```
lachlan-roberts: I think this should be `wss://` unless you are using HTTP/2 Cleartext.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/client/websocket/WebSocketClientDocs.java
```diff
@@ -0,0 +1,193 @@
...
+        // Use the dynamic HTTP/2 transport for HttpClient.
+        HTTP2Client http2Client = new HTTP2Client();
+        HttpClient httpClient = new HttpClient(new HttpClientTransportDynamic(new ClientConnectionFactoryOverHTTP2.HTTP2(http2Client)));
+
+        // Create and start WebSocketClient.
+        WebSocketClient webSocketClient = new WebSocketClient(httpClient);
+        webSocketClient.start();
+
+        ClientEndPoint clientEndPoint = new ClientEndPoint();
+        URI serverURI = URI.create("ws://domain.com/path");
```
lachlan-roberts: Same here.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/WebSocketDocs.java
```diff
@@ -0,0 +1,491 @@
...
+            session.setMaxTextMessageSize(16 * 1024);
+
+            // You may immediately send a message to the remote peer.
+            session.getRemote().sendString("connected", WriteCallback.NOOP);
+        }
+
+        @Override
+        public void onWebSocketClose(int statusCode, String reason)
+        {
+            // The WebSocket connection is closed.
```
lachlan-roberts, line 64: This is not always true, this description is only true for the `websocket-core` `onClosed()` notification. 

This notification can mean that we are fully closed, **or** that we have just received a close frame but not yet sent one. If we have not sent a close frame then we can still send frames from this method, and if a close frame is not sent by the end of this method one will be automatically sent by the implementation.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/websocket.adoc
```diff
@@ -0,0 +1,251 @@
...
+
+A WebSocket endpoint may annotate methods with `org.eclipse.jetty.websocket.api.annotations.*` annotations to receive WebSocket events:
+
+[source,java,indent=0]
+----
+include::{doc_code}/org/eclipse/jetty/docs/programming/WebSocketDocs.java[tags=annotatedEndpoint]
+----
+<1> You must annotate the class with the `@WebSocket` annotation to make it an endpoint.
+<2> For each event you are interested in, use the correspondent annotation.
+<3> Note how methods annotated with `@OnWebSocketMessage` may optionally declare `Session` as first parameter.
```
lachlan-roberts: I know in the section below you mention you can use `InputStream` and `Reader` as parameters for `@OnWebSocketMessage`. But you can also use `ByteBuffer payload` as an alternative to `byte[] payload, int offset, int len`.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/websocket.adoc
```diff
@@ -0,0 +1,251 @@
...
+A WebSocket endpoint is the entity that receives WebSocket events.
+
+The WebSocket events are the following:
+
+* The _connect_ event.
+This event is emitted when the WebSocket communication has been successfully established.
+Applications interested in the connect event receive the WebSocket _session_ so that they can use it to send data to the remote peer.
+* The _close_ event.
+This event is emitted when the WebSocket communication has been closed.
+// TODO: is this event emitted when the application calls Session.close()?
```
lachlan-roberts: No it is not. Only when we receive a close frame OR when the connection is closed (and we never received a close frame).

documentation/jetty-documentation/src/main/asciidoc/programming-guide/websocket.adoc
```diff
@@ -0,0 +1,251 @@
...
+[IMPORTANT]
+====
+You cannot initiate another send of any kind until the previous send is completed.
+
+For example, if you have initiated a text send, you cannot initiate a binary send, until the previous send has completed.
+
+Furthermore, if you have initiated a non-blocking send, you cannot initiate a blocking send, until the previous send has completed.
+====
+
+This requirement is necessary to avoid unbounded buffering that could lead to ``OutOfMemoryError``s.
```
lachlan-roberts, line 207: It might be worth mentioning the `RemoteEndpoint.setMaxOutgoingFrames()` setting.

lachlan-roberts (changes requested):
The sections on websocket endpoints and sessions is going to be required for the server guide as well, how are you going to share this between the two sections?

sbordet:
> The sections on websocket endpoints and sessions is going to be required for the server guide as well, how are you going to share this between the two sections?

The common docs are in the root directory and can be `include`d by both client and server. I have some something similar for HTTP/2.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/client/websocket/client-websocket.adoc
sbordet: Agreed and done.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/client/websocket/WebSocketClientDocs.java
sbordet: Done.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/client/websocket/WebSocketClientDocs.java
sbordet: Done.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/WebSocketDocs.java
sbordet: That is so messy. How can an application know if "we have not sent a close frame". 
What does "fully closed" mean? Also below you mention that if the connection has been closed _without_ receiving a CLOSE frame, then this is invoked?!?
How can an application distinguish all these cases?
I'd rather document "you can only do this one thing" rather than "this method is invoked in these many cases, but you cannot tell the difference among them, and depending on each case you may or may not do things".

documentation/jetty-documentation/src/main/asciidoc/programming-guide/websocket.adoc
sbordet: Done.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/WebSocketDocs.java
lachlan-roberts: > What does "fully closed" mean?

As in the connection has been closed, the session is not just `ISHUT`, or `OSHUT`.

> Also below you mention that if the connection has been closed without receiving a CLOSE frame, then this is invoked?!?

Yes, in the event of some connection failure then this will be invoked with some error status code such as 1006.
```
Abnormal Close is a synonym for {@link #NO_CLOSE}, used to indicate a close
condition where no close frame was processed from the remote side.
```

> How can an application distinguish all these cases?

The application can call `session.isOpen()` to know if they can still send messages to the remote endpoint.

I think the main point is that it can attempt to send an of an alternate close status from the one it received. The close does not throw so they don't even really need to check if the session is still open.

Up to you if you want to include any of this info in the documentation, might be too much of implementation details that people don't really need to know.

lachlan-roberts (approved):

PR #6214, opened 2021-04-26
Copy ServletHolder class/instance properly during startWebapp

### LINKED ISSUE #6280: Copy ServletHolder class/instance properly during startWebapp
When a forced-file (aka jsp-file) servlet holder is found, the held class is copied over from the servlet named "JSP". However, only the class name is copied, so if there is a class instance from a different classloader, the resulting servlet holder will fail.   Any class instance also needs to be copied.

joakime:
#6280


jetty-webapp/src/test/java/org/eclipse/jetty/webapp/ForcedServletTest.java
```diff
@@ -0,0 +1,239 @@
...
+            for (ServletHolder h : getServletHandler().getServlets())
+            {
+                if (h.getForcedPath() != null)
+                {
+                    System.err.println("FORCED " + h.getForcedPath());
+                    h.setHeldClass(RejectUncompiledJspServlet.class);
+                    h.setForcedPath(null);
+                }
+            }
+        }
```
gregw: Remove this method as that is the work around to go into the clients usage rather than the release.
```suggestion
```

jetty-webapp/src/test/java/org/eclipse/jetty/webapp/ForcedServletTest.java
```diff
@@ -0,0 +1,239 @@
...
+        {
+            System.err.printf("### Thread.cl = %s%n", Thread.currentThread().getContextClassLoader());
+
+            // This will result in a 404 for all requests that don't belong to a more precise servlet
+            forceServlet("default", ServletHandler.Default404Servlet.class);
+            addServletMapping("default", "/");
+
+            // This will result in any attempt to use an JSP that isn't precompiled and in the descriptor with status code 555
+            forceServlet("jsp", RejectUncompiledJspServlet.class);
+            rejectForcedJspPaths();
```
gregw: Remove this method as that is the work around to go into the clients usage rather than the release.
```suggestion
```

jetty-servlet/src/main/java/org/eclipse/jetty/servlet/ServletHolder.java
```diff
@@ -294,6 +294,22 @@ public void setForcedPath(String forcedPath)
...
+        // TODO adding this after setting the classname (or instead of) fixes the exceptions.
+        /*
+        Class<? extends Servlet> servletClass = holder.getHeldClass();
+        if (servletClass != null)
+            this.setHeldClass(servletClass);
+        */
+        if (holder.getServlet() != null)
+            this.setServlet(holder.getServlet());
+        if (holder.getServletInstance() != null)
+            this.setInstance(holder.getServletInstance());
```
gregw: @janbartel going to need some help on this one!

The fields that I think we need to consider copying are:
```
ServletHolder:
    Servlet _servlet
BaseHolder:
    Source _source
    Class<? extends T> _class
    String _className;
    T _instance;
```

The `_instance` and `_servlet` are the the same value, although `_serlvet` can be null until `initServlet()`, so it kind of records the init state as something separate to the started state.  If `getServlet()` is called once the handler is running, then `initServlet()` will be called.

The current code copies `_className` and `_servlet`, but not `_source` or `_class`... or any of the config fields.
So this method is a little conflicted?   Is it named to say it copies just the class, but it will copy an actual instance if it is there.  But not the config within the holder... so there is a risk that a single instance will be initialised twice with different configurations.

If different holders have the same `_servlet`, then it will be destroyed multiple times.

So we have to think what we are really trying to achieve here?  Do we want to copy just the class or the actual instance?    I think if we are doing this for real JSPs then we want to copy to instance as we don't want multiple instances of the `JspServlet`.    In which case we have to be careful about: exactly when the copy is done; what we do with conflicting config; how we stop multiple inits and destroys;

gregw (dismissed):
@janbartel can you review this and then let's chat.  I think we need to think very carefully about what it means to have 2 holders sharing the same instance.

jetty-servlet/src/main/java/org/eclipse/jetty/servlet/ServletHolder.java
```diff
@@ -294,6 +294,22 @@ public void setForcedPath(String forcedPath)
         _forcedPath = forcedPath;
     }
 
+    private void copyClassServlet(ServletHolder holder) throws ServletException
+    {
+        if (holder.getClassName() != null)
+            this.setClassName(holder.getClassName());
+        // TODO adding this after setting the classname (or instead of) fixes the exceptions.
```
janbartel: I'm fine with setting the held class. All we need to do is to ensure that the loaded class is copied over, and thus the name of the class set etc.

jetty-servlet/src/main/java/org/eclipse/jetty/servlet/ServletHolder.java
```diff
@@ -294,6 +294,22 @@ public void setForcedPath(String forcedPath)
...
+        /*
+        Class<? extends Servlet> servletClass = holder.getHeldClass();
+        if (servletClass != null)
+            this.setHeldClass(servletClass);
+        */
+        if (holder.getServlet() != null)
```
janbartel: I'm not sure this is correct. I don't think we have any current behaviour where we reuse an _instance_ of a servlet across ServletHolders.

jetty-servlet/src/main/java/org/eclipse/jetty/servlet/ServletHolder.java
```diff
@@ -294,6 +294,22 @@ public void setForcedPath(String forcedPath)
         _forcedPath = forcedPath;
     }
 
+    private void copyClassServlet(ServletHolder holder) throws ServletException
```
janbartel:
```suggestion
    private void copyServletClass(ServletHolder holder) throws ServletException
```

jetty-webapp/src/test/resources/alt-jsp-webdefault.xml
```diff
@@ -0,0 +1,195 @@
+<?xml version="1.0" encoding="utf-8"?>
```
janbartel: Can this file be simpler? Do we need all this?

janbartel (changes requested):

jetty-servlet/src/main/java/org/eclipse/jetty/servlet/ServletHolder.java
gregw: Ah, I got confused as I didn't realise the code to copy the instance was new.  
I'm still somewhat concerned by the duplicate configuration.

jetty-servlet/src/main/java/org/eclipse/jetty/servlet/ServletHolder.java
janbartel: In the case where the jsp_file has _not_ been precompiled into a servlet, the existing code copies all of the init params of the JspServlet over. 

The only code change we need to made is to copy the _class_ instead of it's _name_.

gregw: resolved after conversation with @janbartel

jetty-servlet/src/main/java/org/eclipse/jetty/servlet/ServletHolder.java
gregw: @janbartel found that config is already handled

jetty-webapp/src/test/resources/alt-jsp-webdefault.xml
joakime: We need this to trigger the conditions.
Well, we need a default descriptor that doesn't have "jsp" or "default" defined.
It can probably be pruned a great deal, but it should still exist.

jetty-webapp/src/test/java/org/eclipse/jetty/webapp/jsp/FakeJspServlet.java
```diff
@@ -0,0 +1,37 @@
+//
```
janbartel: If this and the other test servlets aren't referenced anywhere else, can we make them inner classes of the ForcedServletTest?

janbartel (changes requested):

jetty-webapp/src/test/java/org/eclipse/jetty/webapp/jsp/FakeJspServlet.java
gregw: done

jetty-webapp/src/test/java/org/eclipse/jetty/webapp/jsp/FakePrecompiledJSP.java
```diff
@@ -0,0 +1,37 @@
+//
```
janbartel: Make this an inner class.

jetty-webapp/src/test/java/org/eclipse/jetty/webapp/jsp/FakeJspServlet.java
janbartel: Make this one an inner class of ForcedServletTest too.

janbartel (changes requested):
Both of the remaining test servlet classes could be inner classes too.

janbartel (approved):
LGTM

PR #6220, opened 2021-04-28
Fixes #6208 - HTTP/2 max local stream count exceeded

### LINKED ISSUE #6208: HTTP/2 max local stream count exceeded
**Jetty version**
9.4.39.v20210325

**Java version**
1.8

**OS type/version**
MAC and linux

**Description**
When the header size reaches some value (In my demo, it's 2237B) and multiple requests send concurrently, some requests will not be sent to remote server and cause an TimeoutException and then an "java.lang.IllegalStateException: Max local stream count 100 exceeded". When the header size is smaller, it's ok. Please refer to this demo project to reproduce it. 
https://github.com/qingdaoheze/jetty-gzip-decode

@sbordet

sbordet:
Backported from Jetty 10 the "new stream" event so that the Stream can be set early on the client's `HttpChannelOverHTTP2`.
In this way, when a HEADERS frame stalled due to TCP congestion is failed, the corresponding Stream is closed and the connection released to the pool, fixing the "max stream exceeded" issue.

### DIFF
--- a/jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Stream.java
+++ b/jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Stream.java
@@ -67,6 +67,7 @@ public class HTTP2Stream extends IdleTimeout implements IStream, Callback, Dumpa
     private boolean remoteReset;
     private Listener listener;
     private long dataLength;
+    private boolean committed;
 
     public HTTP2Stream(Scheduler scheduler, ISession session, int streamId, boolean local)
     {
@@ -227,6 +228,18 @@ public class HTTP2Stream extends IdleTimeout implements IStream, Callback, Dumpa
     }
 
     @Override
+    public void commit()
+    {
+        committed = true;
+    }
+
+    @Override
+    public boolean isCommitted()
+    {
+        return committed;
+    }
+
+    @Override
     public boolean isOpen()
     {
         return !isClosed();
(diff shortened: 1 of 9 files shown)


sbordet:
@gregw @lorban see https://github.com/eclipse/jetty.project/issues/6208#issuecomment-828394112 for an explanation of the issue.

The fix is as follows:

- Backported from 10 the "new stream" event that happens as soon as a local stream is created. In this way we don't have to wait for the HEADERS frame to be written out to the network before having the `Stream` object available.
- If the HEADERS frame is not written due to TCP congestion, and then the corresponding request times out, now the `Stream` is available to send a reset to the remote peer.
- Sending a reset to the remote peer may be against the specification; I have introduced a `commit` flag to the stream that is set as soon as the first control frame for that stream is written to the network.
- By checking the commit flag we can drop RST_STREAM frames to avoid violating the specification.

lorban (approved):
LGTM after peer review.

jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Stream.java
```diff
@@ -226,6 +227,18 @@ public boolean isLocallyClosed()
         return closeState.get() == CloseState.LOCALLY_CLOSED;
     }
 
+    @Override
+    public void commit()
+    {
+        committed = true;
```
gregw, line 233: what is the memory barrier on this?

gregw (approved):
Other than slight concern on memory barrier... looks OK.
Logic felt a bit special for RST, but then I think it is.

jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Stream.java
sbordet: The flag is set and read from the flusher thread, so the memory barrier is guaranteed by `IteratingCallback`.

PR #6222, opened 2021-04-28
Fixes #6219 by adding option to the deploy module.

### LINKED ISSUE #6219: Support stopping deployment on context failure during startup
**Jetty version**

Jetty 10/11

**Description**

Currently when you try and load a webapp for something on the wrong side of the big bang, the way you realize is a handful of exceptions related to class not found for things like servlet Filter, which can be quite far back in the startup log, with the server still up and running regardless.

I wonder if we can improve this experience for users at all, either by outright failing the startup with a clear message stating the issue, or perhaps a startup option for the distribution that crawls through your webapp and calls attention to issues? Both? I have seen @joakime on the lists correcting people trying to run the wrong version of Jetty for their applications and perhaps it could be useful to just have a command we can respond with that will clearly articulate to the user what the source of the problem is, which jar files they have in their war that are problematic, etc.

I question the value to allowing startup when such a gross error is active, that got me thinking of having --strict and --relaxed modes on startup, with --strict being default and terminating upon exceptions detected on startup, that could help with automated deployment tooling that see successful startup regardless of actually being deployed.

Thoughts?

jmcc0nn3ll:
Adds a property to the jetty deploy module, defaulting to existing behavior to allows server to easily be configured to halt on deploy when a webapp fails to deploy.

### DIFF
--- a/jetty-deploy/src/main/config/etc/jetty-deploy.xml
+++ b/jetty-deploy/src/main/config/etc/jetty-deploy.xml
@@ -55,7 +55,7 @@
               </Set>
               <Set name="scanInterval"><Property name="jetty.deploy.scanInterval" default="1"/></Set>
               <Set name="extractWars"><Property name="jetty.deploy.extractWars" default="true"/></Set>
-              <Set name="stopOnWebappFail"><Property name="jetty.deploy.stopOnWebappFail" default="false"/></Set>
+              <Set name="stopOnFailedDeploy"><Property name="jetty.deploy.stopOnFailedDeploy" default="false"/></Set>
               <Set name="configurationManager">
                 <New class="org.eclipse.jetty.deploy.PropertiesConfigurationManager">
                   <!-- file of context configuration properties
--- a/jetty-deploy/src/main/config/etc/jetty-deploy.xml
+++ b/jetty-deploy/src/main/config/etc/jetty-deploy.xml
@@ -55,6 +55,7 @@
               </Set>
               <Set name="scanInterval"><Property name="jetty.deploy.scanInterval" default="1"/></Set>
               <Set name="extractWars"><Property name="jetty.deploy.extractWars" default="true"/></Set>
+              <Set name="stopOnWebappFail"><Property name="jetty.deploy.stopOnWebappFail" default="false"/></Set>
               <Set name="configurationManager">
                 <New class="org.eclipse.jetty.deploy.PropertiesConfigurationManager">
                   <!-- file of context configuration properties
--- a/jetty-deploy/src/main/config/modules/deploy.mod
+++ b/jetty-deploy/src/main/config/modules/deploy.mod
@@ -28,3 +28,7 @@ etc/jetty-deploy.xml
 
 # Whether to extract *.war files
 # jetty.deploy.extractWars=true
+
+# Whether to halt server on startup if webapp fails to deploy
+# false lets server continue running, true halts server
+# jetty.deploy.stopOnWebappFail=false
--- a/jetty-deploy/src/main/config/modules/deploy.mod
+++ b/jetty-deploy/src/main/config/modules/deploy.mod
@@ -29,6 +29,6 @@ etc/jetty-deploy.xml
 # Whether to extract *.war files
 # jetty.deploy.extractWars=true
 
-# Whether to halt server on startup if webapp fails to deploy
+# Whether to halt server on startup if context fails to deploy
 # false lets server continue running, true halts server
-# jetty.deploy.stopOnWebappFail=false
+# jetty.deploy.stopOnFailedDeploy=false
--- a/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/WebAppProvider.java
+++ b/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/WebAppProvider.java
@@ -55,13 +55,14 @@ import org.slf4j.LoggerFactory;
  * <p>For XML configured contexts, the ID map will contain a reference to the {@link Server} instance called "Server" and
  * properties for the webapp file as "jetty.webapp" and directory as "jetty.webapps".
  */
-@ManagedObject("Provider for start-up deployement of webapps based on presence in directory")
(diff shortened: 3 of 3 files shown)


jetty-deploy/src/main/config/etc/jetty-deploy.xml
```diff
@@ -55,6 +55,7 @@
               </Set>
               <Set name="scanInterval"><Property name="jetty.deploy.scanInterval" default="1"/></Set>
               <Set name="extractWars"><Property name="jetty.deploy.extractWars" default="true"/></Set>
+              <Set name="stopOnWebappFail"><Property name="jetty.deploy.stopOnWebappFail" default="false"/></Set>
```
gregw: Name would be better as `stopOnFailedDeploy` or `stopOnFailedContext` as it is not always a Webapp.

jetty-deploy/src/main/config/modules/deploy.mod
```diff
@@ -28,3 +28,7 @@ etc/jetty-deploy.xml
 
 # Whether to extract *.war files
 # jetty.deploy.extractWars=true
+
+# Whether to halt server on startup if webapp fails to deploy
+# false lets server continue running, true halts server
+# jetty.deploy.stopOnWebappFail=false
```
gregw: name change and CRLF

gregw (changes requested):

jetty-deploy/src/main/config/etc/jetty-deploy.xml
jmcc0nn3ll: good point, done

jetty-deploy/src/main/config/modules/deploy.mod
jmcc0nn3ll: and done

sbordet (changes requested):
I don't think it's correct to have a property on the deployer that stops the server.

Why it is not named after the property of the WebAppContext, i.e. `throwUnavailableOnStartupException`?

jmcc0nn3ll:
@sbordet IMO `throwUnavailableOnStartupException` is a configuration option a WebAppContext has can could be configured on a per webapp option, the `stopOnFailedDeploy` is the action that the `WebappProvider` in jetty-deploy can take to apply to everything it handles. It makes little sense to have to create custom xml files setting that throw* property for each WebAppContext.

Ultimately, I as a user should have the ability to set something in one place that calls a stop to startup should the webapp(s) I am deploying do not start up correctly, effectively to fail fast for me.  IMO this should be the default behavior for people leveraging automated tooling, it is generally worthless to have your deployment in some half enabled state but running. People will still need to have some sort of deployment check to make sure their application is running as expected, but if it fails within the Jetty startup phase then why not have it _fail_ fast.

I mean I suppose it could be moved up to a server.ini setting I suppose, i chose the low hanging fruit approach here.

sbordet:
> Ultimately, I as a user should have the ability to set something in one place that calls a stop to startup should the webapp(s) I am deploying do not start up correctly

I disagree or I don't understand the use case.

You have 25 perfectly deployed webapps, and you deploy the 26th. Why would you want the server to stop?
Also, you don't want to stop a server you are connected with JMX to monitor it: if you stop it, you cannot know what went wrong because the JMX communication will be closed.

Seems to me a development requirement that leaks into a production "feature" that is not really needed and actually dangerous to have.

joakime:
> You have 25 perfectly deployed webapps, and you deploy the 26th. Why would you want the server to stop?

A scenario is where the server as a whole is invalid if the 26th webapp isn't present.

> Also, you don't want to stop a server you are connected with JMX to monitor it: if you stop it, you cannot know what went wrong because the JMX communication will be closed.

This is mostly used on initial startup, the concerns about what happens on redeployment or hot deployment I don't see as a feature that this PR should attempt to cover.

jmcc0nn3ll:
@joakime @sbordet and I had a conversation on this PR and with @gregw comment on removing the 'Webapp' from the property name indicates there is likely something a little deeper going on.  While my PR focused on a general property name on jetty-deploy that in this specific case drilled down into a WebAppContext boolean, that same property perhaps ought to apply more generally to all contexts. 

Stated more succinctly, a property in the deploy module should allow for the deployment manager to fail startup if any managed context being deployed fails.  

The end result should be a 'fail fast' option for both development and production (a la automated deployment of an application stack).  

This end result is what begs the question if this should be a property of the jetty-deploy or if it should be a broader still jetty-server option.  Traditionally we have adopted a 'keep running at all costs' approach to server startup, and ultimately what I am advocating for is the ability to simply fail fast in situations that I as a developer or system admin know that if something fails during start the entire exercise is doomed and we should shut it all down.

That some unintended behavior is in play here is that with the PR above, the last output to the terminal is:

```
Usage: java -jar $JETTY_HOME/start.jar [options] [properties] [configs]
       java -jar $JETTY_HOME/start.jar --help  # for more information
```

This means things have busted their way out to start.jar and it is clueless to the state and just reports a usage violation.

gregw:
On reflection,  "stop" is the wrong verb for this property.    It really is just the same `throwUnavailableOnStartupException` property as individual contexts - ie the deployer will throw if one context fails to start.  What the caller of the deployer does with that is beyond the scope of the property.     So let's just keep it the same name.

janbartel:
Can I also draw your attention to the existing `global-webapp-common.mod` module https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-deploy/src/main/config/modules/global-webapp-common.mod. This creates a `${jetty.base}/etc/webapp-common.xml`  file into which you can place any kind of configuration of `WebAppContext` that the deployer applies to all webapps that it deploys - enabling the `throwUnavailableOnStartupException` could be put in there.

jmcc0nn3ll:
Well shoot Jan, that is easily the best solution here!

Where is that documented? And if not then we absolutely need to get that done @WalkerWatch

jmcc0nn3ll:
Jan's approach will solve this, just need @WalkerWatch to make sure it is fully documented!

PR #6225, opened 2021-04-29
Fixes #6208 - HTTP/2 max local stream count exceeded (#6220)

### LINKED ISSUE #6208: HTTP/2 max local stream count exceeded

sbordet:
Forward port of #6220 from jetty-9.4.x to jetty-10.0.x.

(cherry-picked from commit 2f19c67b410082eb74db4da464ad5fc39122d69b)


lorban (approved):
LGTM

Issue #6167, opened 2021-04-13
Review Http[Input|Output].Interceptor


sbordet:
**Jetty version**
10.0.x

**Description**
`Http[Input|Output].Interceptor` are underspecified and need to be reviewed and clarified.

This issue stems from work on #6163 and also #6156 and #6010.

Interceptors are not allowed to throw checked exceptions, so they have to wrap failures in runtime exceptions, which would be unnecessary.

Furthermore, if an interceptor throws, the exception handling needs to be careful in both blocking and non-blocking I/O.

@lachlan-roberts @lorban @gregw please add your comments about this matter.

sbordet:
@lorban please use this issue as an umbrella to port #6163 to Jetty 10.

However, this issue is also for `HttpOutput.Interceptor` so you may want to issue a PR for `HttpInput` but not close this issue until also `HttpOutput` is done.

lachlan-roberts:
The `HttpOutput.Interceptor` can fail the callback instead of throwing, however failing the callback causes the connection to be aborted with no 500 response even if the `HttpChannel` is not committed.

gregw:
An interesting question is that if there is a failure during an async write (either in interceptor or below) then should it be thrown from the call to async write?  If we don't throw, then the caller might loop on isReady and write, never seeing the exception. We can't call onError because we are already in onWritePossible. @sbordet how sure are we that we have not introduced such a loop with the new catches?

So i think we should throw.... but then do we call onError as well once the call to onWritePossible returns?

lorban:
`HttpInput` side is done.

sbordet:
Closing, as Jetty 12 does not have interceptors anymore.

PR #6226, opened 2021-04-30
Issue #6167 - Review failure cases of HttpOutput.Interceptor

lachlan-roberts:
**Issue #6167**

This PR adds extra testing around what happens if an `HttpOutput.Interceptor` throws or fails the callback.

Some changes were made to `HttpOutput` so that in most cases we will respond with a 500 response instead of just aborting the connection.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
@@ -191,6 +191,7 @@ public class HttpOutput extends ServletOutputStream implements Runnable
     private WriteListener _writeListener;
     private volatile Throwable _onError;
     private Callback _closedCallback;
+    private volatile boolean _abortOnCloseError;
 
     public HttpOutput(HttpChannel channel)
     {
@@ -266,7 +267,14 @@ public class HttpOutput extends ServletOutputStream implements Runnable
                 _firstByteTimeStamp = Long.MAX_VALUE;
         }
 
-        _interceptor.write(content, last, callback);
+        try
+        {
+            _interceptor.write(content, last, callback);
+        }
+        catch (Throwable t)
+        {
+            callback.failed(t);
+        }
     }
 
     private void onWriteComplete(boolean last, Throwable failure)
@@ -310,7 +318,10 @@ public class HttpOutput extends ServletOutputStream implements Runnable
         try
         {
             if (failure != null)
-                _channel.abort(failure);
+            {
+                if (_channel.isCommitted() || _abortOnCloseError)
+                    _channel.abort(failure);
+            }
 
             if (closedCallback != null)
             {
@@ -519,15 +530,7 @@ public class HttpOutput extends ServletOutputStream implements Runnable
         try (AutoLock l = _channelState.lock())
         {
             if (_onError != null)
-            {
-                if (_onError instanceof IOException)
-                    throw (IOException)_onError;
-                if (_onError instanceof RuntimeException)
-                    throw (RuntimeException)_onError;
-                if (_onError instanceof Error)
-                    throw (Error)_onError;
-                throw new IOException(_onError);
-            }
+                IO.throwIOException(_onError);
 
             switch (_state)
             {
@@ -613,6 +616,12 @@ public class HttpOutput extends ServletOutputStream implements Runnable
             {
                 // Do an async close
                 channelWrite(content, true, new WriteCompleteCB());
+
+                // If an asynchronous error has happened already we can throw it here.
+                // Otherwise the async call to onWriteComplete must abort if there is a failure as onError may not be called.
+                _abortOnCloseError = true;
+                if (_onError != null)
+                    IO.throwIOException(_onError);
             }
             else
             {
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
@@ -530,7 +530,7 @@ public class HttpOutput extends ServletOutputStream implements Runnable
         try (AutoLock l = _channelState.lock())
         {
             if (_onError != null)
-                IO.throwIOException(_onError);
+                IO.throwCheckedIOException(_onError);
 
             switch (_state)
             {
@@ -621,7 +621,7 @@ public class HttpOutput extends ServletOutputStream implements Runnable
                 // Otherwise the async call to onWriteComplete must abort if there is a failure as onError may not be called.
                 _abortOnCloseError = true;
                 if (_onError != null)
-                    IO.throwIOException(_onError);
+                    IO.throwCheckedIOException(_onError);
             }
             else
             {
@@ -700,6 +700,9 @@ public class HttpOutput extends ServletOutputStream implements Runnable
         ByteBuffer content = null;
         try (AutoLock l = _channelState.lock())
         {
+            if (_onError != null)
+                IO.throwCheckedIOException(_onError);
+
             switch (_state)
             {
                 case CLOSED:
@@ -753,7 +756,7 @@ public class HttpOutput extends ServletOutputStream implements Runnable
     private void checkWritable() throws IOException
     {
         if (_onError != null)
-            IO.throwIOException(_onError);
+            IO.throwCheckedIOException(_onError);
 
         if (_softClose)
                 throw new EofException("Closed");
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
@@ -750,8 +750,11 @@ public class HttpOutput extends ServletOutputStream implements Runnable
         }
     }
 
-    private void checkWritable() throws EofException
+    private void checkWritable() throws IOException
     {
+        if (_onError != null)
+            IO.throwIOException(_onError);
+
         if (_softClose)
                 throw new EofException("Closed");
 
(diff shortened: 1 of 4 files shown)


jetty-util/src/main/java/org/eclipse/jetty/util/IO.java
```diff
@@ -507,6 +507,24 @@ public static long write(GatheringByteChannel out, ByteBuffer[] buffers, int off
...
     }
 
+    /**
+     * Throw an {@link IOException} given a {@link Throwable}, if the supplied throwable is itself an {@link IOException}
+     * it will be thrown, otherwise it will be wrapped with {@link IOException}. If the supplied throwable is a
+     * {@link RuntimeException} or an {@link Error} it will be thrown without being wrapped in {@link IOException}.
+     * @param t the cause of the exception.
+     * @throws IOException if supplied throwable is not RuntimeException or Error.
+     */
+    public static void throwIOException(Throwable t) throws IOException
```
gregw: You need to work on the name and javadoc a little bit.     It only wraps checked exceptions as `IOException`
So perhaps `throwCheckedIOException` ???

jetty-server/src/test/java/org/eclipse/jetty/server/HttpOutputInterceptorTest.java
```diff
@@ -0,0 +1,441 @@
...
+                ServletOutputStream outputStream = response.getOutputStream();
+                baseRequest.getResponse().getHttpOutput().setBufferSize(0);
+                outputStream.setWriteListener(new WriteListener()
+                {
+                    @Override
+                    public void onWritePossible() throws IOException
+                    {
+                        while (outputStream.isReady())
+                        {
+                            outputStream.write(content);
```
gregw, line 227: Can you add another variation of this that does the write and then calls `flush` so that we can verify we get the same behaviour if there is an exception in `flush`

gregw (dismissed):
Well I like the approach.... other than my niggles about test, name and javadoc.
But I'd still like to hear from @sbordet regarding throwing from async close/write/flush.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
```diff
@@ -613,6 +616,12 @@ public void close() throws IOException
             {
                 // Do an async close
                 channelWrite(content, true, new WriteCompleteCB());
+
+                // If an asynchronous error has happened already we can throw it here.
+                // Otherwise the async call to onWriteComplete must abort if there is a failure as onError may not be called.
+                _abortOnCloseError = true;
+                if (_onError != null)
+                    IO.throwCheckedIOException(_onError);
```
sbordet, line 624: This piece of code after `channelWrite()` is in a race with the `WriteCompleteCB`, so I think you should only do the work in `WriteCompleteCB`.

jetty-server/src/test/java/org/eclipse/jetty/server/HttpOutputInterceptorTest.java
```diff
@@ -0,0 +1,525 @@
...
+    public void after() throws Exception
+    {
+        _stacklessLogging.close();
+        _server.stop();
+    }
+
+    @Test
+    public void testInterceptorCallbackFailed() throws Exception
+    {
+        addInterceptor(nextInterceptor -> new HttpOutput.Interceptor()
```
sbordet, line 87: `addInterceptor()` seems unnecessary. Why don't you just add the interceptor from the `AbstractHandler` that you have in every test? In this way you also won't need `_handlerCollection`.

Something like:

```java
@Test
public void testInterceptorCallbackFailed() 
{
    start(new AbstractHandler()
    {
        @Override
        public void handle(...) 
        {
            HttpOutput httpOutput = baseRequest.getResponse().getHttpOutput();
            httpOutput.setInterceptor(...);
            ...
        }
    });
}

File: jetty-server/src/test/java/org/eclipse/jetty/server/HttpOutputInterceptorTest.java [hunk 5]
```diff
@@ -0,0 +1,525 @@
...
+        });
+
+        CompletableFuture<Throwable> errorFuture = new CompletableFuture<>();
+        _handlerCollection.addHandler(new AbstractHandler()
+        {
+            @Override
+            public void handle(String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response) throws IOException
+            {
+                baseRequest.setHandled(true);
+                byte[] content = "content to be written".getBytes();
```
sbordet: Always specify the charset. Unnecessary here, but for good measure and consistency with rest of code.

sbordet (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
lachlan-roberts: This code is about getting the `close()` method to throw if the `WriteCompleteCB` has been failed before we get to this (like if the interceptor failed it). We do this here to report the exception back to the application without aborting the channel so we can send a 500 response.

jetty-server/src/test/java/org/eclipse/jetty/server/HttpOutputInterceptorTest.java
lachlan-roberts: Not all the interceptors added by the tests do the same thing, some fail the callback, some throw, some start a new thread and wait on a CountDownLatch before failing the callback.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
gregw, line 624: I think we should be more explicit that this code is in a known race, even though it will work the same:
```suggestion
                Throwable onError = _onError; // Was an exception reported already
                if (onError != null)
                    IO.throwCheckedIOException(onError);
```

@sbordet I told @lachlan-roberts that you wouldn't like this code... yes it is a race for a general type of async exception, but there is a class of exception that will be thrown during the processing of the close, and it is those that we should throw here so the application can know all was not OK.

jetty-server/src/test/java/org/eclipse/jetty/server/HttpOutputInterceptorTest.java
sbordet: @lachlan-roberts you can do what you need with the interceptor on the second line of the handler, no?

jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
sbordet: @gregw @lachlan-roberts IIUC this `if` block is the case where we do an async close, which is actually our educated interpretation of what we should do in case of async I/O, because it's underspecified.
I'm not sure it's worth reporting it with an exception, because we already report it with `onError()` (and if we don't we probably should).

I am worried about the case where we throw from close, but also fail the callback, and we will have one thread receiving the exception trying to write the error page, and to do that modify the channel and output states, and another thread failing the callback also modifying the channel and output states, and calling `onError()`.
I feel that we are just asking for troubles, garbled responses, etc.

If there is an exception in an async close, we call `onError()` and that's enough notifications to the application.

sbordet (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
lachlan-roberts: The problem is that we don't call `onError()` if there is a failure with the async close.
If there is the following code:
```java
httpOutput.close();
asyncContext.complete();
return;
```

then you can't call `onError()` because `asyncContext.complete()` was called and there was no call to `isReady()` which returned false.

gregw: Moreover,   we can't rely on calling onError for issues at the end of the stream, because when we have an exception or are at end of stream, then isReady() returns true, so the app will loop and not return from onWritePossible, so we will never be able to call onError even if we wanted to!

sbordet: If we don't call `onError()`, which is the usual path for errors in async I/O, why do we even bother?

The exception may or may not be thrown depending on thread scheduling etc, so it's not reliable behavior, we have a race in our code and we don't provide any reliable benefit to user code, just that sometimes, under certain conditions, if the thread scheduling aligns, maybe we throw.

This is the kind of code that in 6 months we are going to have an issue reported and ask ourselves "who put that racy code in there?"

I don't think it's worth it.

gregw: Because if we don't throw, then we end up aborting the connection rather than sending a 500

sbordet: @gregw _some_ times we send a 500; some other times we don't.
If it was possible to always send a 500, would be good -- even if it was racy.
But if we can't always send a 500, then I don't think it's worth it, also considering that calling `close()` explicitly is not that common.

gregw: But we also throw from write or flush, so it's just making it the same as those operations

sbordet: But in `write()` and `flush()` there is no race.

gregw:
Considering @lachlan-roberts research regarding this, I am switching this PR back to draft and Lachlan should work on a version that returns false from isReady if there is an exception that will be passed to onError.  

@lachlan-roberts can you write up your findings either here on in the actual issue.

gregw (changes requested):
Try changing so isReady() returns false if an exception has occurred

PR #6228, opened 2021-04-30
Fix #6227 Async timeout dispatch race

### LINKED ISSUE #6227: Better resolve race between `AsyncListener.onTimeout` and `AsyncContext.dispatch`
There is an intrinsic race between an application calling `AsyncContext.dispatch` and the container calling `AsyncListener.onTimeout`.
Currently when an async timeout expires, the `HttpChannelState` is atomically changed from `ASYNC` to `EXPIRE`, so that any subsequent call by the application to `AsyncContext.dispatch` will receive an ISE.

However, once the container starts calling any `AsyncListener.onTimeout` listeners, it switches the state to `EXPIRING` as it is allowed for those listeners to call `AsyncContext.dispatch` as part of their valid handling.   However, at this point the current implementation is unable to distinguish between a valid call to dispatch from a listener vs a late call to dispatch from the normal application logic.  So the dispatch is allowed, which can result in either the container throwing an ISE when is doesn't see `EXPIRING` state just before calling `onTimeout`, or the handling within an `onTimeout` listener may get an ISE as the request has already been dispatched.

This results at least in a noisy logged exception.  At worst, the application dispatch may be used even if onTimeout is called.  Better handling would be that once EXPIRING, only the same thread is allowed to call `dispatch` (or `complete for that matter`) so that only a listener will be able to call them and any late application calls are rejected with an ISE.

Whilst this race has been seen in the wild, it will be very difficult to reproduce in a unit test as there is no suitable extension point.

gregw:
Fix #6227 Only allow the thread calling onTimeout to call dispatch and complete once timeout has expired.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannelState.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannelState.java
@@ -155,6 +155,7 @@ public class HttpChannelState
     private boolean _asyncWritePossible;
     private long _timeoutMs = DEFAULT_TIMEOUT;
     private AsyncContextEvent _event;
+    private Thread _onTimeoutThread;
 
     protected HttpChannelState(HttpChannel channel)
     {
@@ -582,7 +583,10 @@ public class HttpChannelState
             switch (_requestState)
             {
                 case ASYNC:
+                    break;
                 case EXPIRING:
+                    if (Thread.currentThread() != _onTimeoutThread)
+                        throw new IllegalStateException(this.getStatusStringLocked());
                     break;
                 default:
                     throw new IllegalStateException(this.getStatusStringLocked());
@@ -646,36 +650,47 @@ public class HttpChannelState
                 throw new IllegalStateException(toStringLocked());
             event = _event;
             listeners = _asyncListeners;
+            _onTimeoutThread = Thread.currentThread();
         }
 
-        if (listeners != null)
+        try
         {
-            Runnable task = new Runnable()
+            if (listeners != null)
             {
-                @Override
-                public void run()
+                Runnable task = new Runnable()
                 {
-                    for (AsyncListener listener : listeners)
+                    @Override
+                    public void run()
                     {
-                        try
+                        for (AsyncListener listener : listeners)
                         {
-                            listener.onTimeout(event);
-                        }
-                        catch (Throwable x)
-                        {
-                            LOG.warn("{} while invoking onTimeout listener {}", x, listener, x);
+                            try
+                            {
+                                listener.onTimeout(event);
+                            }
+                            catch (Throwable x)
+                            {
+                                LOG.warn("{} while invoking onTimeout listener {}", x, listener, x);
+                            }
                         }
                     }
-                }
 
-                @Override
-                public String toString()
-                {
-                    return "onTimeout";
-                }
-            };
+                    @Override
+                    public String toString()
+                    {
+                        return "onTimeout";
+                    }
+                };
 
-            runInContext(event, task);
+                runInContext(event, task);
+            }
+        }
+        finally
+        {
+            synchronized (this)
+            {
+                _onTimeoutThread = null;
+            }
         }
     }
 
@@ -692,6 +707,11 @@ public class HttpChannelState
             switch (_requestState)
             {
                 case EXPIRING:
+                    if (Thread.currentThread() != _onTimeoutThread)
+                        throw new IllegalStateException(this.getStatusStringLocked());
+                    _requestState = _sendError ? RequestState.BLOCKING : RequestState.COMPLETE;
+                    break;
+
                 case ASYNC:
                     _requestState = _sendError ? RequestState.BLOCKING : RequestState.COMPLETE;
                     break;
--- a/jetty-servlets/src/main/java/org/eclipse/jetty/servlets/QoSFilter.java
+++ b/jetty-servlets/src/main/java/org/eclipse/jetty/servlets/QoSFilter.java
@@ -235,7 +235,8 @@ public class QoSFilter implements Filter
                             }
                             catch (IllegalStateException x)
                             {
-                                LOG.warn(x);
+                                if (LOG.isDebugEnabled())
+                                    LOG.debug(x);
                                 continue;
                             }
                         }


gregw:
@sbordet @joakime any idea how to write a unit test for this?   Can we try this branch in the wild with the known reproduction?

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannelState.java
```diff
@@ -646,36 +650,47 @@ protected void onTimeout()
...
-            };
+                    @Override
+                    public String toString()
+                    {
+                        return "onTimeout";
+                    }
+                };
 
-            runInContext(event, task);
+                runInContext(event, task);
```
sbordet, line 685: Method `onTimeout()` is called from the `HttpChannel.handle()` loop.

I wonder why are we explicitly calling `runInContext()` here, rather than just changing the state, returning, and letting the `handle()` loop do its job by looking at the state?

Not that it would change much, but at least "normalize" the place where we throw `IllegalStateException`.

A change for Jetty 10 in any case.

sbordet (changes requested):
The change looks right. 

However, I think that in `sendError()` there should be checks similar to what has been added to `dispatch()` and `complete()`.
I think these are the only 3 methods that can be called from an async listener, right?

joakime:
Do we want to fix the QoSFilter to set `AsyncContext.setTimeout(0)` if the suspendMs is set to 0? (or negative)
Not allowing the default AsyncContext timeout in those scenarios?

joakime:
See issue #6227 for link to reproduction projects.

I was able to test this branch, and the onTimeout WARNings are no longer happening.
However, the finally block in QoSFilter.doFilter is still triggering a WARN ...

```
16:56:33.207 [server-95] WARN  org.eclipse.jetty.servlets.QoSFilter - 
java.lang.IllegalStateException: s=HANDLING rs=EXPIRING os=OPEN is=IDLE awp=false se=false i=false al=1
	at org.eclipse.jetty.server.HttpChannelState.dispatch(HttpChannelState.java:589)
	at org.eclipse.jetty.server.AsyncContextState.dispatch(AsyncContextState.java:94)
	at org.eclipse.jetty.servlets.QoSFilter.doFilter(QoSFilter.java:233)
	at org.eclipse.jetty.servlet.FilterHolder.doFilter(FilterHolder.java:193)
	at org.eclipse.jetty.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1601)
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:548)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:233)
	at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1435)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:188)
	at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:501)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:186)
	at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1350)
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:141)
	at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:127)
	at org.eclipse.jetty.server.Server.handle(Server.java:516)
	at org.eclipse.jetty.server.HttpChannel.lambda$handle$1(HttpChannel.java:388)
	at org.eclipse.jetty.server.HttpChannel.dispatch(HttpChannel.java:633)
	at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:380)
	at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:277)
	at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:311)
	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:105)
	at org.eclipse.jetty.io.ChannelEndPoint$1.run(ChannelEndPoint.java:104)
	at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.runTask(EatWhatYouKill.java:336)
	at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.doProduce(EatWhatYouKill.java:313)
	at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.tryProduce(EatWhatYouKill.java:171)
	at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.run(EatWhatYouKill.java:129)
	at org.eclipse.jetty.util.thread.ReservedThreadExecutor$ReservedThread.run(ReservedThreadExecutor.java:383)
	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:882)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1036)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

Is this one we can squelch too?

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannelState.java
gregw: agree the runInContext looks strange here... but I'm not going to play with it... as you say, something for 10

gregw:
I'm not so sure about `sendError` needing the same logic.  It is somewhat more like an output method like `getOutputStream().write(....)` which we will not check against a timeout state.
I think we should keep the logic in the async lifecycle state machine: startAsync->timeout|dispatch|complete.  A write or a sendError can happen anywhere in that cycle.

More over, this change gives precedence to a `onTimeout` listener that may call `sendError` over an above a race with the application calling `dispatch` which may result in other writes etc.  So I think the case is covered.

gregw:
@joakime squelched

sbordet (approved):

gregw:
I think we should go ahead with this, even though there is a spec conversation underway. 
@sbordet @joakime if you are OK with that, I'll merge later today and prepare a PR for 10

joakime (approved):

Issue #6224, opened 2021-04-29
make jetty-jspc-maven-plugin @threadSafe


gbo80:
**Jetty version**
9.x, 10.x, 11.x

**Description**
Running the jetty-jspc-maven-plugin in a maven build that uses multiple threads (e.g. "mvn -T1C ...") works (so far I found no problems), but I get a warning that the plugin is not @threadSafe.
```
[WARNING] *****************************************************************
[WARNING] * Your build is requesting parallel execution, but project      *
[WARNING] * contains the following plugin(s) that have goals not marked   *
[WARNING] * as @threadSafe to support parallel building.                  *
[WARNING] * While this /may/ work fine, please look for plugin updates    *
[WARNING] * and/or request plugins be made thread-safe.                   *
[WARNING] * If reporting an issue, report it against the plugin in        *
[WARNING] * question, not against maven-core                              *
[WARNING] *****************************************************************
[WARNING] The following plugins are not marked @threadSafe in my-web:
[WARNING] org.eclipse.jetty:jetty-jspc-maven-plugin:9.4.40.v20210413
[WARNING] Enable debug to see more precisely which goals are not marked @threadSafe.
[WARNING] *****************************************************************
```
**Question**
Do we have to expect problems when running a multi-module project that contains multiple web projects with multiple threads?
We use the JSP precompiler in the maven build only for detecting compilation problems while building.
We do not use the precompiled JSPs in production.

**Change request**
Please make the plugin @threadSafe.

Thanks
Günther

janbartel:
@gbo80 I don't anticipate that you will have any threading issues with the jetty-jspc-maven-plugin, but we'll look at adding the appropriate annotation to the plugin.

gbo80:
Thank you very much!

olamy:
pr merged https://github.com/eclipse/jetty.project/commit/04e5c4e07716f0fe76a8707fa5b49e9ff9fb4d9c

trajano:
Can this be re-openned as Jetty 9.4.x still marks it as not-thread-safe

olamy:
@trajano please read here https://github.com/jetty/jetty.project/issues/7958 regarding 9.4.x support

PR #6230, opened 2021-05-02
Issue #6224 make jspc mojo maven threadsafe

olamy:


jetty-jspc-maven-plugin/src/main/java/org/eclipse/jetty/jspc/plugin/JspcMojo.java
```diff
@@ -241,12 +242,6 @@ protected TldScanner newTldScanner(JspCServletContext context, boolean namespace
     @Parameter
     private String targetVersion;
 
-    /**
-     * The JspC instance being used to compile the jsps.
```
janbartel: I think we should leave this as a parameter  - that way, users can configure it from the pom if they want.

janbartel (changes requested):
I think we should leave the jspc parameter.

jetty-jspc-maven-plugin/src/main/java/org/eclipse/jetty/jspc/plugin/JspcMojo.java
olamy: agree and not agree :) 
Because even if a user provide his own from a pom we will override some values.
Look at the code https://github.com/eclipse/jetty.project/blob/7555d0375aac6e0d9d51b0785d5259d80af3905d/jetty-jspc-maven-plugin/src/main/java/org/eclipse/jetty/jspc/plugin/JspcMojo.java#L336.
it's not very consistent and can be confusing for user who provide their own jspc as values will be override by some default values.

janbartel: Sure, however, the values that we are setting explicitly are able to be configured by params we've added to the plugin to integrate it with the jetty/unassembled webapp environment. There are dozens of other setters that exist on the JspC class that the user should be able to have access to that we don't want to have to add params to our plugin for - moreover, it's a bit of future-proofing: the JspC class may well change to add more setters and users would be able to access them without waiting for us to add params to the plugin.

olamy: fair enough but as a user I find it very confusing.
Using this:
```
<configuration>
  <jspc>
    <webXmlInclude>foo/bar/blabla.xml</webXmlInclude>
    few other attributes
  </jspc>
</configuration>
```
At the end the value used for `jspc.webXmlInclude` will be `${basedir}/target/webfrag.xml` because we always override with mojo field.
We should fixed that (well noboby complained so it's probably not an issue ;) ).
But what about `merging` the configuration? Pseudo code:
```
if jspc!=null && jsp.webXmlInclude1=null {
  // do not use value from mojo field
}
```
WDYT?

olamy: not related to this PR I opened another issue https://github.com/eclipse/jetty.project/issues/6235

olamy: @janbartel change reverted

janbartel (approved):
LGTM

PR #6236, opened 2021-05-05
Fixes #5306 - Default jetty.*.acceptors should be 1.

### LINKED ISSUE #5306: Default jetty.*.acceptors should be 1
**Jetty version**
10.0.x

**Description**
Given that connections are more and more persistent (due to protocols keeping them open for performance reasons), and many many cores, I think the current heuristic is too generous (up to 23 cores yields 2, up to 31 cores yields 3, more yields 4), and a fixed heuristic of `1` is plenty enough.

sbordet:
Changed the acceptor heuristic to default to 1, rather than calculating a value based on the number of cores.
Updated documentation.
Fixed a glitch in ManagedSelector.getMaxSelectedKeys() to return long, not double.


gregw:
Don't forget to link PR to issue: #5306

gregw (changes requested):
It is too confusing to have the default set to -1, which is a heuristic that sets it to 1.

Just keep the heuristic logic and change the default to 1

sbordet:
@gregw nudge

gregw (approved):

PR #6243, opened 2021-05-06
Fixes #6099 - Cipher preference may break SNI if certificates have di…

### LINKED ISSUE #6099: Cipher preference may break SNI if certificates have different key types
**Jetty version**
10.0.1

**Java version**
15.0.2

**OS type/version**
Windows 10.0.19042.867

**Description**
When the keystore contains certificates with keys of different types, cipher preference seems to take precedence over SNI.

Example setup:

- a keystore with:
  - an EC certificate with CN=ec.example.com, SAN = DNS:ec.example.com
  - an RSA certificate with CN=rsa.example.com, SAN = DNS:rsa.example.com
- `jetty.sslContext.sniRequired=false` (the default)

### Test 1 (fail)

```
curl -v -k --tlsv1.2 --resolve "rsa.example.com:443:127.0.0.1" https://rsa.example.com/
```

**Expected result**
The certificate for rsa.example.com is presented in the TLS handshake.

**Actual result**
ECDHE-ECDSA-AES256-GCM-SHA384 is the selected cipher.
The certificate for ec.example.com is presented in the TLS handshake, and a 400 Invalid SNI is returned. The following can be seen in the logs:

```
DEBUG:oejus.SslContextFactory:qtp1287934450-76: SNI matching for type=host_name (0), value=rsa.example.com
DEBUG:oejus.SslContextFactory:qtp1287934450-76: SNI host name rsa.example.com
DEBUG:oejus.SniX509ExtendedKeyManager:qtp1287934450-76: Chose delegate alias ec/EC on sun.security.ssl.SSLEngineImpl@44f79dfb
```

### Test 2 (success)

```
curl -v -k --tlsv1.2 --cipher ECDHE-RSA-AES256-GCM-SHA384 --resolve "rsa.example.com:443:127.0.0.1" https://rsa.example.com/
```

**Actual result**
The certificate for rsa.example.com is presented in the TLS handshake.

### Workaround
Setting `jetty.sslContext.sniRequired=true` fixes this behavior.

sbordet:
…fferent key types.

Updated the logic in SslContextFactory.Server.sniSelect(...) to check if there is
any certificate that matches, and if so return a null alias in the hope to be called
again and pick the right alias for the SNI.


jetty-util/src/main/java/org/eclipse/jetty/util/ssl/SslContextFactory.java
```diff
@@ -2306,9 +2316,15 @@ public String sniSelect(String keyType, Principal[] issuers, SSLSession session,
...
-                // No match, let the JDK decide unless unmatched SNIs are rejected.
                 if (matching.isEmpty())
-                    return isSniRequired() ? null : SniX509ExtendedKeyManager.SniSelector.DELEGATE;
+                {
+                    // There is no match for this SNI among the certificates valid for
+                    // this keyType; check if there is any certificate that matches this
+                    // SNI, as we will likely be called again with a different keyType.
+                    boolean anyMatching = aliasCerts().values().stream()
+                        .anyMatch(x509 -> x509.matches(sniHost));
+                    return isSniRequired() || anyMatching ? null : SniX509ExtendedKeyManager.SniSelector.DELEGATE;
```
gregw, line 2326: The problem with this algorithm is that if we have 5 matching certificates, this logic will be true just as much for the 5th as for calls 1 to 4.
I guess there is no where to remember a count easily?

sbordet: Good point. 
We can probably store an attribute in the handshake session -- not sure what though.
Initialize with all x509s, and remove those that don't match the `keyType`?

gregw: sounds plausible. Something like:
 + Look for Integer attribute in session:
   * if not found, then count matching SNIs and set the Integer.
 + Decrement the integer
    * If == 0 return DELEGATE
    * else return null

sbordet: @gregw on further analysis, I think the first version is enough.

Assuming the server has [ec.domain.com keyType=EC, rsa.domain.com keyType=RSA], then:

- case 1: 
client wants to connect, but only with keyType=EC, to rsa.domain.com;
we will be called with [ec.domain.com keyType=EC], which won't match the SNI; we know there is a matching certificate, we hope we will be called again (but we won't), so we return `null` (wrong-ish), which fails the connection, but there is nothing we can do.

- case 2:
client wants to connect to unknown.domain.com;
we will be called with [ec.domain.com keyType=EC], which won't match the SNI; we figure out there is no match at all; we return either `null` or `delegate` (correct).

- default case:
there is a match for both keyType and SNI, so eventually we find the right alias to return.

The proposed fix improves case 1: before we were immediately returning `delegate`, now we return `null` in the hope to be called again. 
It's not perfect, but if the client wants to connect with keyType=EC to a domain that does not support that keyType, I think failing the TLS handshake may be better than returning an EC certificate for a different domain.
Unfortunately this behavior is not configurable, as we have to return `null` in the hope to be called again.

gregw (approved):

PR #6246, opened 2021-05-10
Fixes #6099 - Cipher preference may break SNI if certificates have di…

### LINKED ISSUE #6099: Cipher preference may break SNI if certificates have different key types

sbordet:
…fferent key types.

Updated the logic in SslContextFactory.Server.sniSelect(...) to check if there is
any certificate that matches, and if so return a null alias in the hope to be called
again and pick the right alias for the SNI.

(cherry picked from commit 68296911b7ee58a9853b1baadbba549fb89f3904)


sbordet:
Forward port of #6243

gregw (approved):

PR #6252, opened 2021-05-10
Fixes #6250 - Lazily allocate HTTP2Stream data queue.

### LINKED ISSUE #6250: Lazily allocate HTTP2Stream data queue
**Jetty version**
10.0.x

**Description**
Most `HTTP2Stream` have no content, and allocating an `ArrayDeque` to never use it has proven to be an allocation hotspot.

Since it's always accessed inside a lock anyway, lazily allocate it.

sbordet:
Now the data queue is lazily allocated.

### DIFF
--- a/jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Stream.java
+++ b/jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Stream.java
@@ -55,7 +55,7 @@ public class HTTP2Stream extends IdleTimeout implements IStream, Callback, Dumpa
     private static final Logger LOG = LoggerFactory.getLogger(HTTP2Stream.class);
 
     private final AutoLock lock = new AutoLock();
-    private final Queue<DataEntry> dataQueue = new ArrayDeque<>();
+    private Queue<DataEntry> dataQueue;
     private final AtomicReference<Object> attachment = new AtomicReference<>();
     private final AtomicReference<ConcurrentMap<String, Object>> attributes = new AtomicReference<>();
     private final AtomicReference<CloseState> closeState = new AtomicReference<>(CloseState.NOT_CLOSED);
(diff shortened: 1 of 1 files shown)


jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Stream.java
```diff
@@ -235,12 +235,15 @@ public boolean isRemotelyClosed()
...
+        List<DataEntry> copy = List.of();
         try (AutoLock l = lock.lock())
         {
             dataDemand = 0;
-            copy = new ArrayList<>(dataQueue);
-            dataQueue.clear();
+            if (dataQueue != null)
+            {
+                copy = new ArrayList<>(dataQueue);
+                dataQueue.clear();
```
gregw: why not take the queue? Is the stream likely to be reused after a fail? will it need a queue?
```suggestion
                copy = dataQueue;
                dataQueue = null;
```

gregw (changes requested):

jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Stream.java
```diff
@@ -413,6 +416,8 @@ private void onData(DataFrame frame, Callback callback)
         DataEntry entry = new DataEntry(frame, callback);
         try (AutoLock l = lock.lock())
         {
+            if (dataQueue == null)
+                dataQueue = new ArrayDeque<>(8);
```
lorban: Why specifying the element count here? It wasn't when the `dataQueue` was initialized at the field level.

lorban (changes requested):

jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Stream.java
sbordet: If you don't specify it is by default `16`.
The value of `8` is arbitrary, any idea for a better value?
The default max frame size is 16 KiB, so `8` it will hold 128 KiB of data which is double the default flow control window of 64 KiB.
It halves the initial allocation without incurring in 3 expansions if I give an initial size of `1`.

gregw: Can you set it as a function of the actual flow control window?

gregw (approved):

lorban (approved):

PR #6257, opened 2021-05-11
Jetty 9.4.x 6254 queued request total timeout

### LINKED ISSUE #6254: Total timeout not enforced for queued requests
**Jetty version**
9.4.x

**Description**
The total timeout for requests that are queued in a `HttpDestination` is not always enforced.
Sending a request just before a previous queued one is about to expire resets the timeout of the already queued one.

sbordet:
@gregw first commit is cleanup and javadocs improvements, second commit is the actual fix.

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
@@ -567,14 +567,15 @@ public abstract class HttpDestination extends ContainerLifeCycle implements Dest
                 LOG.debug("{} timeouts check", this);
 
             long now = System.nanoTime();
-
+            long earliest = Long.MAX_VALUE;
             // Reset the earliest timeout so we can expire again.
             // A concurrent call to schedule(long) may lose an earliest
             // value, but the corresponding exchange is already enqueued
             // and will be seen by scanning the exchange queue below.
-            earliestTimeout.set(Long.MAX_VALUE);
+            earliestTimeout.set(earliest);
 
-            long earliest = Long.MAX_VALUE;
+            // Scan the message queue to abort expired exchanges
+            // and to find the exchange that expire the earliest.
             for (HttpExchange exchange : exchanges)
             {
                 HttpRequest request = exchange.getRequest();
@@ -596,8 +597,8 @@ public abstract class HttpDestination extends ContainerLifeCycle implements Dest
             // Schedule a timeout for the earliest exchange that may expire.
             // When the timeout expires, scan the exchange queue for the next
             // earliest exchange that may expire, and reschedule a new timeout.
-            long earliest = earliestTimeout.getAndUpdate(t -> Math.min(t, expiresAt));
-            if (expiresAt < earliest)
+            long prevEarliest = earliestTimeout.getAndUpdate(t -> Math.min(t, expiresAt));
+            if (expiresAt < prevEarliest)
             {
                 // A new request expires earlier than previous requests, schedule it.
                 long delay = Math.max(0, expiresAt - System.nanoTime());
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
@@ -597,7 +597,7 @@ public abstract class HttpDestination extends ContainerLifeCycle implements Dest
             // When the timeout expires, scan the exchange queue for the next
             // earliest exchange that may expire, and reschedule a new timeout.
             long earliest = earliestTimeout.getAndUpdate(t -> Math.min(t, expiresAt));
-            if (expiresAt != earliest)
+            if (expiresAt < earliest)
             {
                 // A new request expires earlier than previous requests, schedule it.
                 long delay = Math.max(0, expiresAt - System.nanoTime());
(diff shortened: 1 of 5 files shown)


jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
```diff
@@ -564,14 +564,17 @@ private TimeoutTask(Scheduler scheduler)
...
 
-            // Check all queued exchanges for those that have expired
-            // and to determine when the next check must be.
+            // Reset the earliest timeout so we can expire again.
+            // A concurrent call to schedule(long) may lose an earliest
+            // value, but the corresponding exchange is already enqueued
+            // and will be seen by scanning the exchange queue below.
+            earliestTimeout.set(Long.MAX_VALUE);
+
+            long earliest = Long.MAX_VALUE;
```
gregw:
```suggestion
            long earliest = Long.MAX_VALUE;
            earliestTimeout.set(earliest);
```

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
```diff
@@ -580,34 +583,27 @@ public void onTimeoutExpired()
...
-            // If subsequently that exchange is removed from the queue, the
-            // timeout is not cancelled, instead the entire queue is swept
-            // for expired exchanges and a new timeout is set.
-            long timeoutAt = nextTimeout.getAndUpdate(e -> Math.min(e, expiresAt));
-            if (timeoutAt != expiresAt)
+            // Schedule a timeout for the earliest exchange that may expire.
+            // When the timeout expires, scan the exchange queue for the next
+            // earliest exchange that may expire, and reschedule a new timeout.
+            long earliest = earliestTimeout.getAndUpdate(t -> Math.min(t, expiresAt));
+            if (expiresAt < earliest)
```
gregw: Is this right?   If expiresAt is after earliestTimeout, then earliest == earliestTimeout, which is < expiresAt, so do nothing... correct.
But if expiresAt is before earliestTimeout, then earliest == expiresAt, so the < also fails and we also do nothing - wrong?

Surely we need to do a schedule IFF earliest==expiresAt ?

```suggestion
            if (expiresAt == earliest)
```

gregw (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeout.java
```diff
@@ -102,8 +119,9 @@ public boolean schedule(long delay, TimeUnit units)
             if (_timeout.compareAndSet(timeout, new Timeout(newTimeoutAt, wakeup)))
             {
                 if (LOG.isDebugEnabled())
-                    LOG.debug("Installed timeout in {} ms, waking up in {} ms",
+                    LOG.debug("Installed timeout in {} ms, {} wake up in {} ms",
```
gregw, line 123: Use {} on multi-line log

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
sbordet: @gregw the update is a get-then-update, so `earliest` is the _old_ minimum. We don't know the new minimum, could be either `earliest` or `expireAt`.

The less-than check determines what the new minimum is; if it's `expireAt` then we need to schedule.

jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeout.java
sbordet: I am using `{}`, not sure what you mean here?

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
gregw: Ah, then `earliest` is the wrong name.  Should be `lastEarliest`

jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeout.java
gregw:
```java
if (LOG.isDebugEnabled())
{
   LOG.debug(something,
   somethingelse?or:something
   blah,
   blah);
}
```

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
gregw: or `expiringAt`

gregw (dismissed):
Other than the niggles in my comments, this is good.... but please change the variable name before merging

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
sbordet: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
sbordet: `prevEarliest`

jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeout.java
sbordet: Done.

PR #6267, opened 2021-05-12
Fixes #6251 - Use CyclicTimeout for HTTP2Streams.

### LINKED ISSUE #6251: Use CyclicTimeout for HTTP2Streams
**Jetty version**
10.0.x

**Description**
`HTTP2Stream` inherits from `IdleTimeout`, but most of the times its timeout is never enforced, as the stream completes before the timeout expires.

Use `CyclicTimeout` instead of `IdleTimeout`.

sbordet:
Introduced CyclicTimeouts to manage many entities that may timeout.
Rewritten HttpDestination request timeouts using CyclicTimeouts.
HTTP2Stream does not inherit from IdleTimeout anymore; now a
CyclicTimeouts in HTTP2Session manages the stream timeouts.

### DIFF
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeouts.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeouts.java
@@ -17,12 +17,14 @@ import java.util.Iterator;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
+import org.eclipse.jetty.util.component.Destroyable;
 import org.eclipse.jetty.util.thread.Scheduler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * <p>A {@link CyclicTimeout} that manages the timeouts for many entities.</p>
+ * <p>An implementation of a timeout that manages many {@link Expirable expirable} entities whose
+ * timeouts are mostly cancelled or re-scheduled.</p>
  * <p>A typical scenario is for a parent entity to manage the timeouts of many children entities.</p>
  * <p>When a new entity is created, call {@link #schedule(T)} with the new entity so that
  * this instance can be aware and manage the timeout of the new entity.</p>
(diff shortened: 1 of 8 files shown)


jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeouts.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.Iterator;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+
+import org.eclipse.jetty.util.thread.Scheduler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * <p>A {@link CyclicTimeout} that manages the timeouts for many entities.</p>
```
gregw: should say for many entities that each have a timeout that is expected to most be cancelled or rescheduled

jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeouts.java
```diff
@@ -0,0 +1,147 @@
...
+        /**
+         * <p>Returns the expiration time in nanoseconds.</p>
+         * <p>The value to return must be calculated taking into account {@link System#nanoTime()},
+         * for example:</p>
+         * {@code expireNanoTime = System.nanoTime() + timeoutNanos}
+         * <p>Returning {@link Long#MAX_VALUE} indicates that this entity does not expire.</p>
+         *
+         * @return the expiration time in nanoseconds, or {@link Long#MAX_VALUE} if this entity does not expire
+         */
+        public long getExpireNanoTime();
```
gregw, line 160:
```suggestion
        public long getExpireAtNanoTime();
```

jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeouts.java
```diff
@@ -0,0 +1,147 @@
...
+ *   <li>may expire at a future time; the iteration records the earliest expiration time among
+ *   all non-expired entities</li>
+ * </ul>
+ * <p>When the iteration is complete, this instance is re-scheduled with the earliest expiration time
+ * calculated during the iteration.</p>
+ *
+ * @param <T> the entity type
+ */
+public abstract class CyclicTimeouts<T extends CyclicTimeouts.Expirable> extends CyclicTimeout implements Iterable<T>
+{
```
gregw: I'm not sure this class is-a CyclicTimeout.  For a start this class works on expiry times rather than delays.  Also we don't want users calling the underlying schedule method directly.
```suggestion
public abstract class CyclicTimeouts<T extends CyclicTimeouts.Expirable> implements Iterable<T>
{
    private final CyclicTimeout timout;
```

gregw (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeouts.java
sbordet: Uhm. 

Either my `expireNanoTime` where `expire` is noun-ish, or `expire*s*At` where `expire` is verb-ish (so needs the `s`).

jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeouts.java
sbordet: Done.

jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeouts.java
sbordet: Done.
Also remove is-a `Iterable` for the same reasons, only subclasses should override `iterator()` and only the implementation should call it.

gregw (dismissed):

gregw (approved):

PR #6273, opened 2021-05-14
Fixes #6272 - Reduce allocation in HttpClient when notifying content …

### LINKED ISSUE #6272: Reduce allocation in HttpClient when notifying content listeners
**Jetty version**
10.0.x

**Description**
Response content is notified using a `ContentListeners` that is allocated for every response, and uses Java streams that are allocation expensive.

sbordet:
…listeners.

Avoid to allocate ContentListeners for every response.
Avoid using Java streams to reduce allocation.

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/HttpReceiver.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/HttpReceiver.java
@@ -72,8 +72,8 @@ public abstract class HttpReceiver
 
     private final AutoLock lock = new AutoLock();
     private final AtomicReference<ResponseState> responseState = new AtomicReference<>(ResponseState.IDLE);
+    private final ContentListeners contentListeners = new ContentListeners();
     private final HttpChannel channel;
-    private ContentListeners contentListeners;
     private Decoder decoder;
     private Throwable failure;
     private long demand;
@@ -291,7 +291,7 @@ public abstract class HttpReceiver
         ResponseNotifier notifier = getHttpDestination().getResponseNotifier();
         List<Response.ResponseListener> responseListeners = exchange.getConversation().getResponseListeners();
         notifier.notifyHeaders(responseListeners, response);
-        contentListeners = new ContentListeners(responseListeners);
+        contentListeners.reset(responseListeners);
         contentListeners.notifyBeforeContent(response);
 
         if (!contentListeners.isEmpty())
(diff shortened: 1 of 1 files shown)


lorban (approved):

jetty-client/src/main/java/org/eclipse/jetty/client/HttpReceiver.java
```diff
@@ -653,15 +653,22 @@ public String toString()
...
-            listeners = new ArrayList<>(responseListeners.size());
-            responseListeners.stream()
-                .filter(Response.DemandedContentListener.class::isInstance)
-                .map(Response.DemandedContentListener.class::cast)
-                .forEach(listeners::add);
+            demands.clear();
+            listeners.clear();
+        }
+
+        private void reset(List<Response.ResponseListener> responseListeners)
```
gregw, line 664: This is more an `initialise` than a `reset`

jetty-client/src/main/java/org/eclipse/jetty/client/HttpReceiver.java
```diff
@@ -653,15 +653,22 @@ public String toString()
...
+        {
+            clear();
+            for (Response.ResponseListener listener : responseListeners)
+            {
+                if (listener instanceof Response.DemandedContentListener)
+                    listeners.add((Response.DemandedContentListener)listener);
+            }
```
gregw, line 671: re my comment above, this code looks strange to be in a `reset` but reasonable in a `init` method

gregw (approved):
Other than a naming niggle - LGTM

jetty-client/src/main/java/org/eclipse/jetty/client/HttpReceiver.java
sbordet: @gregw I see `initialize()` as a once-off method. Here we are re-setting every time this class is reused, so I like `reset()` better.

gregw: how about `reinitialize` ?
Having a `clear` and a `reset` is just weird... so one of them must be wrong!

sbordet: I don't see why `clear` and `reset` are weird? 
One is to `clear` to avoid holding onto things, `reset` is "be ready again".
`reset` has been used this way in e.g. `IteratingCallback.reset()` and `HttpConnection.SendCallback.reset(...)`, here is the same.

PR #6281, opened 2021-05-16
Jetty 10.0.x servletholder copy #6280

gregw:
cherry pick of #6214 from 9.4 to 10 for #6280

### DIFF
--- a/jetty-webapp/src/test/java/org/eclipse/jetty/webapp/ForcedServletTest.java
+++ b/jetty-webapp/src/test/java/org/eclipse/jetty/webapp/ForcedServletTest.java
@@ -1,0 +1,259 @@
+//
+//  ========================================================================
+//  Copyright (c) 1995-2021 Mort Bay Consulting Pty Ltd and others.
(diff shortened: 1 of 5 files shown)


jetty-webapp/src/test/java/org/eclipse/jetty/webapp/ForcedServletTest.java
```diff
@@ -0,0 +1,269 @@
...
+            holder.setInitOrder(1);
+            holder.setName(name);
+            holder.setAsyncSupported(true);
+            handler.addServlet(holder);
+        }
+    }
+
+    public static class ForcedWebAppContext extends WebAppContext
+    {
+        {
```
janbartel: This doesnt need to be a class - you can add a bean to a standard WebAppContext.

janbartel (changes requested):
If you delete the ForcedWebAppContext class because its unnecessary, I'm good.

janbartel (approved):
Lgtm

PR #6293, opened 2021-05-18
Issue #6287 - fix classloading for WebSocketClient in webapp

### LINKED ISSUE #6287: Class loading broken for WebSocketClient used inside webapp
**Jetty version**
9.4.x

**Description**
The fix for #5859 resolved the leaking of `ClassLoader`s, but now causes class loading issues in web applications, see for example: https://github.com/cometd/cometd/issues/1037.

A web application using the WebSocket client may trigger the creation of a thread, which before #5859 was inheriting the webapp `ClassLoader` as context `ClassLoader`, but now the thread gets the server `ClassLoader` as the context `ClassLoader`.

If the task at hand needs to load a class from the webapp (for example due to JSON parsing using `oej.util.Loader.loadClass()`, it was succeeding before #5859, but fails now.

Obviously we want to keep the fix for #5859, but we need to explicitly set the context `ClassLoader` before calling webapp code (e.g. listeners, callbacks, etc.) in those components that are provided by the server, in particular `WebSocketClient`.

lachlan-roberts:
Issue #6287

### DIFF
--- a/jetty-websocket/websocket-core-client/src/main/java/org/eclipse/jetty/websocket/core/client/WebSocketCoreClient.java
+++ b/jetty-websocket/websocket-core-client/src/main/java/org/eclipse/jetty/websocket/core/client/WebSocketCoreClient.java
@@ -38,6 +38,7 @@ public class WebSocketCoreClient extends ContainerLifeCycle
     private static final Logger LOG = LoggerFactory.getLogger(WebSocketCoreClient.class);
     private final HttpClient httpClient;
     private final WebSocketComponents components;
+    private ClassLoader classLoader;
 
     // TODO: Things to consider for inclusion in this class (or removal if they can be set elsewhere, like HttpClient)
     // - AsyncWrite Idle Timeout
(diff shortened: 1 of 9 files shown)


jetty-websocket/websocket-core-client/src/main/java/org/eclipse/jetty/websocket/core/client/WebSocketCoreClient.java
```diff
@@ -61,12 +62,23 @@ public WebSocketCoreClient(HttpClient httpClient, WebSocketComponents webSocketC
...
     }
 
+    public ClassLoader getClassLoader()
+    {
+        return classLoader;
+    }
+
+    public void setClassLoader(ClassLoader classLoader)
+    {
+        this.classLoader = classLoader;
```
sbordet: I think it should be wrapped with `Objects.requireNonNull()`, as I don't see how we ever want this class loader to be `null`.

jetty-websocket/websocket-javax-tests/src/test/java/org/eclipse/jetty/websocket/javax/tests/ClientClassLoaderTest.java
```diff
@@ -0,0 +1,208 @@
...
+        public void onMessage(String message)
+        {
+            textMessages.add(message);
+        }
+    }
+
+    @WebServlet("/servlet")
+    public static class WebSocketClientServlet extends HttpServlet
+    {
+        private WebSocketContainer clientContainer;
```
sbordet: Allocate the WS container here and make the field `final`, as it's a more typical setup than creating a new container for every HTTP request.

sbordet (changes requested):
I would like to see a similar test for Jetty's API as well.
I know the fix is in core, but we want to make sure.

jetty-websocket/websocket-core-client/src/main/java/org/eclipse/jetty/websocket/core/client/WebSocketCoreClient.java
lachlan-roberts: I'm not even sure we really need the setter method on `WebSocketCoreClient`, it is not currently used. We always just use the ContextClassLoader at the time the `WebSocketCoreClient` was created.

But I have now changed it to use `Objects.requireNonNull()`.

sbordet:
@lachlan-roberts apparently the tests you added for this issue fail, can you check?

sbordet (approved):

PR #6294, opened 2021-05-18
Fix HttpChannelOverHTTP2 recycling order

### LINKED ISSUE #6284: Timeout on reading from Request with HTTPS and HTTP/2
**Jetty version**
10.0.2

**Java version/vendor**
openjdk version "11.0.11" 2021-04-20
OpenJDK Runtime Environment AdoptOpenJDK-11.0.11+9 (build 11.0.11+9)
OpenJDK 64-Bit Server VM AdoptOpenJDK-11.0.11+9 (build 11.0.11+9, mixed mode)

**OS type/version**
Windows 10, 64 bit

**Description**
When upgrading our Vaadin application from Jetty 9.4 to 10, it stopped working properly if accessed via HTTPS and HTTP/2. It the initial screen is loaded fine, but when the client tries to send another request, the server never answers.

I've reproduces the issue in [mperktold/vaadin14-embedded-jetty](https://github.com/mperktold/vaadin14-embedded-jetty), which is a fork of [mvysny/vaadin14-embedded-jetty](https://github.com/mvysny/vaadin14-embedded-jetty). It is basically a "Hello World" Vaadin application running in an embedded Jetty server, which is started programmatically.
In my fork, I've upgraded it to Jetty 10.0.2, Java 11, and Vaadin 19, and enabled HTTPS and HTTP/2 support. It loads fine, but when clicking "Say hello", the browser tries to send the click event to the server for being handled, but never gets a response. I've tested this with Chrome and Firefox, both with the same result.

When disabling HTTP/2, everything works fine.
In the branch [jetty-9](https://github.com/mperktold/vaadin14-embedded-jetty/tree/jetty-9), I downgraded to Jetty 9 again but kept the HTTPS and HTTP/2 support, and there everything works fine as well.

So far, my understanding is that after initializing the page, the browser sends an RST_STREAM (for whatever reason), so the Jetty server resets the connection. When receiving the next request of the same kind, Jetty fails to read the content. It tries to asynchronously read the whole request before passing it to the application, but fails. Then it passes the request to the application, which reads from the input stream, and gets a TimeoutException. Other kinds of requests still get handled correctly though, maybe because they are send over another connection?

I am not sure who is misbehaving, the browser, Jetty, Vaadin, or my configuration, but I thought that at least you have the best chance of understanding what is going on here.

lorban:
`HttpChannelOverHTTP2.recycle()` is implemented with an unfortunate order: since its `ContentDemander` is the first thing to be recycled, any other recycling call that may demand some content will leave the content demander in a corrupt state.

There is currently only one path to trigger this bug: if a servlet/handler makes use of `Request.getReader()` instead of `request.getInputStream()` as recycling the `Request` will read everything from the reader which does demand content, leaving the pending demand request in the demander.

Shuffling the recycle calls in a way that `ContentDemander` is the last object to be recycled guarantees that such corrupt state can never happen.

### DIFF
--- a/jetty-http2/http2-server/src/main/java/org/eclipse/jetty/http2/server/HttpChannelOverHTTP2.java
+++ b/jetty-http2/http2-server/src/main/java/org/eclipse/jetty/http2/server/HttpChannelOverHTTP2.java
@@ -209,11 +209,13 @@ public class HttpChannelOverHTTP2 extends HttpChannel implements Closeable, Writ
     @Override
     public void recycle()
     {
-        _expect100Continue = false;
-        _delayedUntilContent = false;
-        _contentDemander.recycle();
         super.recycle();
         getHttpTransport().recycle();
+        _expect100Continue = false;
+        _delayedUntilContent = false;
+        // The content demander must be the very last thing to be recycled
+        // to make sure any pending demanding content gets cleared off.
+        _contentDemander.recycle();
     }
 
     @Override
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
@@ -55,7 +55,8 @@ public class HttpInput extends ServletInputStream implements Runnable
 
     public void recycle()
     {
-        reopen();
+        if (LOG.isDebugEnabled())
+            LOG.debug("recycle {}", this);
     }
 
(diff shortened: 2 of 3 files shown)


sbordet (changes requested):
Test case?

lorban:
Test case is coming.

sbordet (approved):

PR #6300, opened 2021-05-20
add jetty-9.4.41.v20210516 infos in VERSION.TXT

olamy:


gregw (changes requested):
I didn't know we did this with a PR? Don't we just merge back from the release branch?

But if we are doing a manual PR can you add the missing #6280

olamy:
@gregw I'm not admin here so I cannot push directly to protected branches so this PR
I will add #6280

gregw (approved):

PR #6301, opened 2021-05-20
Issue #6268 Log max form size exceeded msg.

### LINKED ISSUE #6268: Warnings about "unable to parse form content" are not helpful for troubleshooting
****

**java version "1.8.0_291"**
**Linux/CentOS 7**

**Description**

****
I noticed that **_org.eclipse.jetty.server.Request#getParameters_**

```java
private MultiMap<String> getParameters()
{
if (!_contentParamsExtracted)
{
// content parameters need boolean protection as they can only be read
// once, but may be reset to null by a reset
_contentParamsExtracted = true;

            // Extract content parameters; these cannot be replaced by a forward()
            // once extracted and may have already been extracted by getParts() or
            // by a processing happening after a form-based authentication.
            if (_contentParameters == null)
            {
                try
                {
                    extractContentParameters();
                }
                catch (IllegalStateException | IllegalArgumentException e)
                {
                    throw new BadMessageException("Unable to parse form content", e);
                }
            }
        }
    .....
}
```
IllegalStateException  will be decorated into BadMessageException
And the Exception Message will be BadMessageException's message.

This is unreasonable.I care about IllegalStateException's message instead of BadMessageException's message!

e.g.
    if the body size exceed the limit,extractContentParameters() method will throw IllegalStateException with message _Form is larger than max length xxx_ 
    and in _org.eclipse.jetty.server.Request#getParameters#getParameters()_ will be be decorated into BadMessageException with message _Unable to parse form content_
    log will print _Unable to parse form content_ not _Form is larger than max length xxx_
    Think _Form is larger than max length xxx_ is more useful than  _Unable to parse form content_

janbartel:
Closes #6268

Log the `"Form is larger than max length ???"` and `"Cannot extract parameters with async IO"` messages _without stacktraces_ from `Request.extractFormParameters() `method to give better info on the server side rather than the generic `BadMessageException` `"Unable to parse form content"` message.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
@@ -436,7 +436,7 @@ public class Request implements HttpServletRequest
                 }
                 catch (IllegalStateException | IllegalArgumentException e)
                 {
-                    LOG.warn(e.getLocalizedMessage());
+                    LOG.warn(e.toString());
                     throw new BadMessageException("Unable to parse form content", e);
                 }
             }
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
@@ -436,6 +436,7 @@ public class Request implements HttpServletRequest
                 }
                 catch (IllegalStateException | IllegalArgumentException e)
                 {
+                    LOG.warn(e.getLocalizedMessage());
                     throw new BadMessageException("Unable to parse form content", e);
                 }
             }


jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -436,6 +436,7 @@ private void commitSession(Session session)
                 }
                 catch (IllegalStateException | IllegalArgumentException e)
                 {
+                    LOG.warn(e.getLocalizedMessage());
```
gregw: Normally we've just logged e.toString(). Is this better?

janbartel: toString() calls getLocalizedMessage() after prepending the classname. I'll change it to toString().

gregw (approved):

PR #6304, opened 2021-05-20
Issue #6302 - Treat empty path segments as ambiguous.

### LINKED ISSUE #6302: Treat empty path segments are ambiguous.
**Description**

Empty path segments `//` should be treated as ambiguous by `HttpUri`.

lachlan-roberts:

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
@@ -50,10 +50,29 @@ public interface HttpURI
 {
     enum Ambiguous
     {
+        /**
+         * URI contains ambiguous path segments e.g. <code>/foo/%2e%2e/bar</code>
+         */
         SEGMENT,
+
+        /**
+         * URI contains ambiguous empty segments e.g. <code>//</code>
+         */
         EMPTY,
+
+        /**
+         * URI contains ambiguous path separator within a URI segment e.g. <code>/foo/b%2fr</code>
+         */
         SEPARATOR,
+
+        /**
+         * URI contains ambiguous path encoding within a URI segment e.g. <code>/%2557EB-INF</code>
+         */
         ENCODING,
+
+        /**
+         * URI contains ambiguous path parameters within a URI segment e.g. <code>/foo/..;/bar</code>
+         */
         PARAM
     }
 
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
@@ -51,6 +51,7 @@ public interface HttpURI
     enum Ambiguous
     {
         SEGMENT,
+        EMPTY,
         SEPARATOR,
         ENCODING,
         PARAM
@@ -149,6 +150,11 @@ public interface HttpURI
      * @return True if the URI has a possibly ambiguous segment like '..;' or '%2e%2e'
      */
     boolean hasAmbiguousSegment();
+
+    /**
+     * @return True if the URI empty segment that is ambiguous like '//' or '/;param/'.
+     */
+    boolean hasAmbiguousEmptySegment();
 
     /**
      * @return True if the URI has a possibly ambiguous separator of %2f
@@ -381,6 +387,12 @@ public interface HttpURI
         }
 
         @Override
+        public boolean hasAmbiguousEmptySegment()
+        {
+            return _ambiguous.contains(Ambiguous.EMPTY);
+        }
+
+        @Override
         public boolean hasAmbiguousSeparator()
         {
             return _ambiguous.contains(Ambiguous.SEPARATOR);
@@ -453,7 +465,6 @@ public interface HttpURI
             .with("%2e%2e", Boolean.TRUE) // Is real dot dot segment not removed by normalisation.
             .with(".%2e", Boolean.TRUE)   // Is real dot dot segment not removed by normalisation.
             .with("%2e.", Boolean.TRUE)   // Is real dot dot segment not removed by normalisation.
-            .with("", Boolean.TRUE)       // An empty segment may be interpreted differently by file systems.
             .with("..", Boolean.FALSE)    // If followed by a parameter is not removed by dot dot normalisation.
             .with(".", Boolean.FALSE)     // If followed by a parameter is not removed by dot normalisation.
             .build();
@@ -469,6 +480,7 @@ public interface HttpURI
         private String _uri;
         private String _decodedPath;
         private final EnumSet<Ambiguous> _ambiguous = EnumSet.noneOf(Ambiguous.class);
+        private boolean _emptySegment;
 
         private Mutable()
         {
@@ -729,6 +741,12 @@ public interface HttpURI
         }
 
         @Override
+        public boolean hasAmbiguousEmptySegment()
+        {
+            return _ambiguous.contains(Ambiguous.EMPTY);
+        }
+
+        @Override
         public boolean hasAmbiguousSeparator()
         {
             return _ambiguous.contains(Mutable.Ambiguous.SEPARATOR);
@@ -905,6 +923,7 @@ public interface HttpURI
             boolean dot = false; // set to true if the path containers . or .. segments
             int escapedTwo = 0; // state of parsing a %2<x>
             int end = uri.length();
+            _emptySegment = false;
             for (int i = 0; i < end; i++)
             {
                 char c = uri.charAt(i);
@@ -1290,24 +1309,42 @@ public interface HttpURI
          */
         private void checkSegment(String uri, int segment, int end, boolean param)
         {
-            if (!_ambiguous.contains(Ambiguous.SEGMENT))
+            // We had a non last empty segment meaning it was ambiguous.
+            if (_emptySegment)
+                _ambiguous.add(Ambiguous.EMPTY);
+
+            if (end == segment)
             {
                 // Empty segments are only ambiguous if followed by a '#', '?' or end of string.
-                if (end == segment && (end == uri.length() || ("#?".indexOf(uri.charAt(end)) >= 0)))
+                if (end >= uri.length() || ("#?".indexOf(uri.charAt(end)) >= 0))
                     return;
 
-                // Look for segment in the ambiguous segment index.
-                Boolean ambiguous = __ambiguousSegments.get(uri, segment, end - segment);
-                if (ambiguous == Boolean.TRUE)
+                // If the first segment is empty then it is ambiguous.
+                if (segment == 0)
                 {
-                    // The segment is always ambiguous.
-                    _ambiguous.add(Ambiguous.SEGMENT);
+                    _ambiguous.add(Ambiguous.EMPTY);
+                    return;
                 }
-                else if (param && ambiguous == Boolean.FALSE)
+
+                // Otherwise only a non last empty segment is ambiguous.
+                if (!_emptySegment)
                 {
-                    // The segment is ambiguous only when followed by a parameter.
-                    _ambiguous.add(Ambiguous.PARAM);
+                    _emptySegment = true;
+                    return;
                 }
+            }
+
+            // Look for segment in the ambiguous segment index.
+            Boolean ambiguous = __ambiguousSegments.get(uri, segment, end - segment);
+            if (ambiguous == Boolean.TRUE)
+            {
+                // The segment is always ambiguous.
+                _ambiguous.add(Ambiguous.SEGMENT);
+            }
+            else if (param && ambiguous == Boolean.FALSE)
+            {
+                // The segment is ambiguous only when followed by a parameter.
+                _ambiguous.add(Ambiguous.PARAM);
             }
         }
     }
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
@@ -449,13 +449,13 @@ public interface HttpURI
          */
         private static final Index<Boolean> __ambiguousSegments = new Index.Builder<Boolean>()
             .caseSensitive(false)
-            .with("%2e", Boolean.TRUE)
-            .with("%2e%2e", Boolean.TRUE) // Is real dot dot segment not removed by normalisation
-            .with(".%2e", Boolean.TRUE)   // Is real dot dot segment not removed by normalisation
-            .with("%2e.", Boolean.TRUE)   // Is real dot dot segment not removed by normalisation
-            .with("", Boolean.TRUE)       // An empty segment may be interpreted differently by file systems
-            .with("..", Boolean.FALSE)    // If followed by a parameter is not removed by dot dot normalisation
-            .with(".", Boolean.FALSE)     // If followed by a parameter is not removed by dot normalisation
+            .with("%2e", Boolean.TRUE)    // Is real dot segment not removed by normalisation.
+            .with("%2e%2e", Boolean.TRUE) // Is real dot dot segment not removed by normalisation.
+            .with(".%2e", Boolean.TRUE)   // Is real dot dot segment not removed by normalisation.
+            .with("%2e.", Boolean.TRUE)   // Is real dot dot segment not removed by normalisation.
+            .with("", Boolean.TRUE)       // An empty segment may be interpreted differently by file systems.
+            .with("..", Boolean.FALSE)    // If followed by a parameter is not removed by dot dot normalisation.
+            .with(".", Boolean.FALSE)     // If followed by a parameter is not removed by dot normalisation.
             .build();
 
         private String _scheme;
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
@@ -51,27 +51,27 @@ public interface HttpURI
     enum Ambiguous
     {
         /**
-         * URI contains ambiguous path segments e.g. <code>/foo/%2e%2e/bar</code>
+         * URI contains ambiguous path segments e.g. {@code /foo/%2e%2e/bar}
          */
         SEGMENT,
 
         /**
-         * URI contains ambiguous empty segments e.g. <code>//</code>
+         * URI contains ambiguous empty segments e.g. {@code //}
          */
         EMPTY,
 
         /**
-         * URI contains ambiguous path separator within a URI segment e.g. <code>/foo/b%2fr</code>
+         * URI contains ambiguous path separator within a URI segment e.g. {@code /foo/b%2fr}
          */
         SEPARATOR,
 
         /**
-         * URI contains ambiguous path encoding within a URI segment e.g. <code>/%2557EB-INF</code>
+         * URI contains ambiguous path encoding within a URI segment e.g. {@code /%2557EB-INF}
          */
         ENCODING,
 
         /**
-         * URI contains ambiguous path parameters within a URI segment e.g. <code>/foo/..;/bar</code>
+         * URI contains ambiguous path parameters within a URI segment e.g. {@code /foo/..;/bar}
          */
         PARAM
     }
@@ -480,12 +480,12 @@ public interface HttpURI
          */
         private static final Index<Boolean> __ambiguousSegments = new Index.Builder<Boolean>()
             .caseSensitive(false)
-            .with("%2e", Boolean.TRUE)    // Is real dot segment not removed by normalisation.
-            .with("%2e%2e", Boolean.TRUE) // Is real dot dot segment not removed by normalisation.
-            .with(".%2e", Boolean.TRUE)   // Is real dot dot segment not removed by normalisation.
-            .with("%2e.", Boolean.TRUE)   // Is real dot dot segment not removed by normalisation.
-            .with("..", Boolean.FALSE)    // If followed by a parameter is not removed by dot dot normalisation.
-            .with(".", Boolean.FALSE)     // If followed by a parameter is not removed by dot normalisation.
+            .with("%2e", Boolean.TRUE)
+            .with("%2e%2e", Boolean.TRUE)
+            .with(".%2e", Boolean.TRUE)
+            .with("%2e.", Boolean.TRUE)
+            .with("..", Boolean.FALSE)
+            .with(".", Boolean.FALSE)
             .build();
 
         private String _scheme;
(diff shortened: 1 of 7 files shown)


jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -453,6 +453,7 @@ public URI toURI()
             .with("%2e%2e", Boolean.TRUE)
             .with(".%2e", Boolean.TRUE)
             .with("%2e.", Boolean.TRUE)
+            .with("", Boolean.TRUE)
             .with("..", Boolean.FALSE)
             .with(".", Boolean.FALSE)
```
gregw:
```suggestion
            .with("%2e%2e", Boolean.TRUE) // Is real dot dot segment not removed by normalisation 
            .with(".%2e", Boolean.TRUE)   // Is real dot dot segment not removed by normalisation 
            .with("%2e.", Boolean.TRUE)   // Is real dot dot segment not removed by normalisation 
            .with("", Boolean.TRUE)       // An empty segment may be interpreted differently by file systems
            .with("..", Boolean.FALSE)    // If followed by a parameter is not removed by dot dot normalisation
            .with(".", Boolean.FALSE)     // If followed by a parameter is not removed by dot normalisation
```

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -1252,7 +1252,7 @@ public String getRemoteUser()
     public RequestDispatcher getRequestDispatcher(String path)
     {
         // path is encoded, potentially with query
-
+        // TODO: why do we do this?
         path = URIUtil.compactPath(path);
```
gregw: Let's try commenting it out for now and seeing what fails
```suggestion
        // TODO: why do did we do this?
        // path = URIUtil.compactPath(path);
```

jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
```diff
@@ -1790,6 +1790,17 @@ public void testAmbiguousEncoding() throws Exception
         _connector.getBean(HttpConnectionFactory.class).getHttpConfiguration().setUriCompliance(UriCompliance.UNSAFE);
         assertThat(_connector.getResponse(request), startsWith("HTTP/1.1 200"));
     }
+
+    @Test
+    public void testDoubleSlash() throws Exception
+    {
+        _handler._checker = (request, response) -> true;
+        String request = "GET //../foo HTTP/1.0\r\n" +
```
gregw: This is a strange test. It has an authority of `..` and a path of `/foo`.  
Why is it a bad request?   Can you add a comment saying why it is bad and check that is really why a 400 is received

gregw:
I think we also need to add something in the documentation regarding ambiguous URIs, how to configure and alternatives if they are not exactly correct. @lachlan-roberts do you want to have a go at the doco, or should I push an attempt to this branch?

jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
lachlan-roberts: Ahh yes I need to fix this, I was just using it for debugging and forgot about it.
I will need to test out the compliance modes here.

lachlan-roberts:
@gregw you can push an attempt to this branch if you want.

gregw:
I have moved the doco work to #6312 , so this PR can proceed without waiting for it.

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -450,11 +450,12 @@ public URI toURI()
         private static final Index<Boolean> __ambiguousSegments = new Index.Builder<Boolean>()
             .caseSensitive(false)
             .with("%2e", Boolean.TRUE)
```
gregw: oops we missed one!

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -1281,6 +1282,21 @@ private void checkSegment(String uri, int segment, int end, boolean param)
...
+                        return;
+
+                    // If the segment starts after the URI string it means the string ended with a '/'.
+                    if (segment > uri.length() - 1)
+                        return;
+
+                    // It is only a true empty segment if it actually ends with a '/' character.
+                    if (uri.charAt(end) != '/')
+                        return;
+                }
```
gregw: Is is possible to avoid calling checkSegment if there really isn't a segment?

gregw: Hmmm if I comment this section out, all the tests still pass in HttpURITest, so at the very least it needs coverage

gregw (changes requested):

jetty-http/src/test/java/org/eclipse/jetty/http/HttpURITest.java
```diff
@@ -361,6 +361,13 @@ public void testCanonicalDecoded()
                 {".%2e", "..", EnumSet.of(Ambiguous.SEGMENT)},
                 {"%2e%2e", "..", EnumSet.of(Ambiguous.SEGMENT)},
 
+                // empty segment treated as ambiguous
+                {"/foo//bar", "/foo//bar", EnumSet.of(Ambiguous.SEGMENT)},
+                {"/foo//../bar", "/foo/bar", EnumSet.of(Ambiguous.SEGMENT)},
+                {"/foo///../../../bar", "/bar", EnumSet.of(Ambiguous.SEGMENT)},
+                {"/foo/./../bar", "/bar", EnumSet.noneOf(Ambiguous.class)},
+                {"/foo//./bar", "/foo//bar", EnumSet.of(Ambiguous.SEGMENT)},
+
```
gregw, line 379: Perhaps also test URIs like: `foo/bar`, `foo;/bar`, `;/bar`, `;?n=v`, `?n=v`
The last three of those have empty segments

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -1113,24 +1114,28 @@ else if (c == '/')
                         switch (c)
                         {
                             case ';':
-                                checkSegment(uri, segment, i, true);
+                                if (segment != i)
+                                    checkSegment(uri, segment, i, true);
```
gregw: I don't think that is right.  If segment == i, that is an empty segment.  So I think now you are not checking empty segments.

Eitherway, I think that another state is better than having lots of replicated if statements

So maybe have 2 PATH states.  You need one state that any valid terminator is also the end of a segment (eg `/`, `?`, `#`, `;` and EOS) and then you have another state that can see EOS, `?` or `#` without it being the end of a segment. 

So if you had just finished parsing `/foo/bar/` you'd be in second of these states, so if the next character is `?` that is not an empty segment, but if the next character is `;` that is an empty segment.  IE `/foo/bar/?n=v` has 2 segments and `/foo/bar/;?n=v` has 3 segments with the last being empty.

gregw (dismissed):

lachlan-roberts:
@gregw `DefaultServletTest.testListingWithSession()` is failing because it is trying a request to `/context/;JSESSIONID=1234567890` which is now classified as having an ambiguous segment.

What do you think about this, should I fix this test to remove the `;` from the URI?

gregw:
@lachlan-roberts Ouch! that's a tough one.   There definitely is an empty segment there, but we don't care because it is the last segment and we are happy to ignore empty segments when they are the last ones.... but at the time we check it, we don't know if it is the last one or not.

So how do we make `/foo/;param/bar` ambiguous, but `/foo/;param` not ?  ideas??? 
Perhaps an empty segment is only ambiguous if we have see another segment after it?
So the first time checkSegment is called with an empty segment, we just remember we have seen an empty segment, then if checkSegment is ever called again we say we are ambiguous.  That might be better than checking for the termination character???

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -51,6 +51,7 @@
     enum Ambiguous
     {
         SEGMENT,
+        EMPTY,
         SEPARATOR,
         ENCODING,
         PARAM
```
gregw: Let's add some javadoc for each of these giving examples

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -1251,10 +1251,6 @@ public String getRemoteUser()
     @Override
     public RequestDispatcher getRequestDispatcher(String path)
     {
-        // path is encoded, potentially with query
-
-        path = URIUtil.compactPath(path);
-
```
gregw, line 1257: @joakime Can you comment on why this was added?

I think removing it is safe, as it may mean that an application might not find a dispatcher that it previously did, but it should never result in a dispatcher being found that should not be???

jetty-http/src/test/java/org/eclipse/jetty/http/HttpURITest.java
```diff
@@ -401,10 +421,47 @@ public void testDecodedPath(String input, String decodedPath, EnumSet<Ambiguous>
...
+                {"/foo/?bar", EnumSet.noneOf(Ambiguous.class)},
+                {"/foo/#bar", EnumSet.noneOf(Ambiguous.class)},
+                {"/foo/;param", EnumSet.noneOf(Ambiguous.class)},
+                {"/foo/;param/bar", EnumSet.of(Ambiguous.EMPTY)},
+            }).map(Arguments::of);
+    }
+
+    @ParameterizedTest
+    @MethodSource("emptySegmentTests")
+    public void testEmptySegment(String input, EnumSet<Ambiguous> expected)
```
gregw: Can't these tests all be done in the test above?

gregw (commented):
I've had enough input into this PR that it needs another reviewer.
However, other than some niggles, it LGTM

jetty-http/src/test/java/org/eclipse/jetty/http/HttpURITest.java
lachlan-roberts: No because for this test we are using `HttpURI.from("GET", input)` instead of just `HttpURI.from(input)`, this starts us from state `PATH` so it can reproduce some failures the other one cannot.

There are tests for empty segments in the above one as well.

gregw: Then rename the test to `testFromPathQuery` and use `HttpURI.build().pathQuery(uri)` instead of the `from` method.    You can then add a bunch of other tests (eg %2f and %2e ones) that verify expected behaviour when parsing starting from a path or query

gregw: Also, like the other test, you should test what the resulting path actually is, not just the ambiguous set

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
joakime: This was added as a result of issue https://bugs.eclipse.org/482056 into Jetty 8.1.18 back in 2015 (as part of a "jetty 8 backport" effort that I don't recall the details of)

IIRC, it was because people were using it lazily, like ...

``` java
final String DATAPATH = "/data/controller.do";
request.getRequestDispatcher("/app/" + DATAPATH) 
```

which results in `getRequestDispatcher("/app//data/controller.do")` style of usage.

or 

``` java
request.getRequestDispatcher(request.getContextPath() + "/data/foo");
```

which can easily result in a double slash as well (when we used to return `/` on root context).

I'm good with changing this for Jetty 10+, being a bit more strict in the process.
We should probably make this a configurable switch though (or verify that these awkward request paths can be handled in getRequestDispatcher still via other means)

jetty-http/src/test/java/org/eclipse/jetty/http/HttpURITest.java
```diff
@@ -449,14 +480,58 @@ public void testDecodedPath(String input, String decodedPath, EnumSet<Ambiguous>
...
+                {"/path/%2f/..;/info", EnumSet.of(Ambiguous.SEPARATOR, Ambiguous.PARAM)},
+                {"/path/%2f/..;/%2e/info", EnumSet.of(Ambiguous.SEPARATOR, Ambiguous.PARAM, Ambiguous.SEGMENT)},
             }).map(Arguments::of);
     }
 
     @ParameterizedTest
-    @MethodSource("emptySegmentTests")
-    public void testEmptySegment(String input, EnumSet<Ambiguous> expected)
+    @MethodSource("pathFromQueryTests")
+    public void testPathFromQuery(String input, EnumSet<Ambiguous> expected)
```
gregw: `testFromPathQuery` or `testPathQuery` ???

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
gregw: If need be, you can always have a filter that is on FORWARD only that will look for // and do a new dispatch to the compacted path.

lachlan-roberts:
@sbordet / @joakime can you please review this PR

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -50,9 +50,29 @@
 {
     enum Ambiguous
     {
+        /**
+         * URI contains ambiguous path segments e.g. <code>/foo/%2e%2e/bar</code>
```
sbordet: Can you use the more modern `{@code /foo/%2e%2e/bar} instead?

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -449,12 +480,12 @@ public URI toURI()
          */
         private static final Index<Boolean> __ambiguousSegments = new Index.Builder<Boolean>()
             .caseSensitive(false)
-            .with("%2e", Boolean.TRUE)
-            .with("%2e%2e", Boolean.TRUE)
-            .with(".%2e", Boolean.TRUE)
-            .with("%2e.", Boolean.TRUE)
-            .with("..", Boolean.FALSE)
-            .with(".", Boolean.FALSE)
+            .with("%2e", Boolean.TRUE)    // Is real dot segment not removed by normalisation.
```
sbordet: I'd stick with American spelling and use `normali*z*ation` instead.

Also, the comments are not very clear. 

This is an `Index.Builder`, so it should just reference that -- no external references to normalization or such.

To me the comment should be: "`// %2e is considered ambiguous.` -- while I cannot understand what does "Is real dot segment not removed by normalisation." mean.

If I lookup `%2e` in this `Index` I get back `true` -- that's it.

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -1279,13 +1328,42 @@ else if (_path != null)
...
+                if (segment == 0)
+                {
+                    _ambiguous.add(Ambiguous.EMPTY);
+                    return;
+                }
+
+                // Otherwise only a non last empty segment is ambiguous.
+                if (!_emptySegment)
+                {
+                    _emptySegment = true;
```
sbordet, line 1354: I fail to understand why you need to remember `_emptySegment` across calls.
Here, can't you just call `_ambiguous.add(Ambiguous.EMPTY)` and get rid of the `_emptySegment` field?

Seems to me that `/a//p;/a?#` would add an empty segment multiple times -- sure it's a `Set` but I don't understand why we would do that?

If it is necessary, needs a comment.

jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
```diff
@@ -1790,6 +1790,23 @@ public void testAmbiguousEncoding() throws Exception
...
+            "Host: whatever\r\n" +
+            "\r\n";
+        _connector.getBean(HttpConnectionFactory.class).getHttpConfiguration().setUriCompliance(UriCompliance.DEFAULT);
+        assertThat(_connector.getResponse(request), startsWith("HTTP/1.1 400"));
+        _connector.getBean(HttpConnectionFactory.class).getHttpConfiguration().setUriCompliance(UriCompliance.LEGACY);
+        assertThat(_connector.getResponse(request), startsWith("HTTP/1.1 200"));
+        _connector.getBean(HttpConnectionFactory.class).getHttpConfiguration().setUriCompliance(UriCompliance.RFC3986);
+        assertThat(_connector.getResponse(request), startsWith("HTTP/1.1 200"));
+        _connector.getBean(HttpConnectionFactory.class).getHttpConfiguration().setUriCompliance(UriCompliance.UNSAFE);
+        assertThat(_connector.getResponse(request), startsWith("HTTP/1.1 200"));
```
sbordet, line 1808: Can you please add other combinations?

`//ambiguous/initial`
`/ambiguous//middle`
`/double//ambiguous//`
`//triple//ambiguous//`

sbordet (changes requested):

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
gregw: @sbordet There are comments explaining:
```java
            // We had a non last empty segment meaning it was ambiguous.
...
                // Empty segments are only ambiguous if followed by a '#', '?' or end of string.
...
                // If the first segment is empty then it is ambiguous.
...
                // Otherwise only a non last empty segment is ambiguous.
```

The `_emptySegment` field is true if there has been an empty segment, but that is not necessarily ambiguous.  It is only ambiguous if there is another segment found after an empty segment.  Hence the `_emptySegment` check at the start of the method.

Ah but there is a mistake in one of the comments.  I will comment separately

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -1279,13 +1328,42 @@ else if (_path != null)
          */
         private void checkSegment(String uri, int segment, int end, boolean param)
         {
-            if (!_ambiguous.contains(Ambiguous.SEGMENT))
+            // We had a non last empty segment meaning it was ambiguous.
```
gregw:
```suggestion
            // If we have previously seen an empty segment, then it was not the last segment and was thus ambiguous.
```

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -1279,13 +1328,42 @@ else if (_path != null)
...
+            if (_emptySegment)
+                _ambiguous.add(Ambiguous.EMPTY);
+
+            if (end == segment)
+            {
+                // Empty segments are only ambiguous if followed by a '#', '?' or end of string.
```
gregw:
```suggestion
                // Empty segments are not ambiguous if followed by a '#', '?' or end of string.
```

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -1279,13 +1328,42 @@ else if (_path != null)
...
+                if (end >= uri.length() || ("#?".indexOf(uri.charAt(end)) >= 0))
+                    return;
+
+                // If the first segment is empty then it is ambiguous.
```
gregw:
```suggestion
                // If this empty segment is the first segment then it is ambiguous.
```

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -1279,13 +1328,42 @@ else if (_path != null)
...
+                if (segment == 0)
+                {
+                    _ambiguous.add(Ambiguous.EMPTY);
+                    return;
+                }
+
+                // Otherwise only a non last empty segment is ambiguous.
```
gregw:
```suggestion
                // Otherwise remember we have seen an empty segment, which is check if we see a subsequent segment.
```

gregw (changes requested):
There is at least one negation error in the comments.
Also here are some suggested improvements to the wording of them

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
lachlan-roberts: I was asked by Greg to put these comments in. Should I just remove these comments? I don't think "`// %2e is considered ambiguous.`" adds much over just reading the code.

jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
lachlan-roberts: I don't think this would be the right place for these extra tests. This is just testing how the compliance modes handle and empty segment.

What combinations cause an empty segment is tested by `HttpUriTest`. We already cover these cases, but I have also added these ones exactly to be sure.

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
lachlan-roberts: I have removed the comments here.

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
lachlan-roberts: Done.

gregw (dismissed):

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -1279,13 +1328,42 @@ else if (_path != null)
          */
         private void checkSegment(String uri, int segment, int end, boolean param)
         {
-            if (!_ambiguous.contains(Ambiguous.SEGMENT))
+            // If we have previously seen an empty segment, then it was not the last segment and was thus ambiguous.
+            if (_emptySegment)
+                _ambiguous.add(Ambiguous.EMPTY);
```
gregw:
```suggestion
            // This method is called once for every segment parsed.
            // A URI like "/foo/" has two segments: "foo" and an empty segment.
            // Empty segments are only ambiguous if they are not the last segment
            // So if this method is called for any segment and we have previously seen an empty segment, then it was ambiguous 
            if (_emptySegment)
                _ambiguous.add(Ambiguous.EMPTY);
```

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -50,9 +50,29 @@
 {
     enum Ambiguous
     {
+        /**
+         * URI contains ambiguous path segments e.g. {@code /foo/%2e%2e/bar}
+         */
         SEGMENT,
+
+        /**
+         * URI contains ambiguous empty segments e.g. {@code //}
```
gregw:
```suggestion
         * URI contains ambiguous empty segments e.g. {@code /foo//bar} or {@code /foo/;param/}, but not {@code /foo/}
```

sbordet (approved):

PR #6306, opened 2021-05-20
Fix #6305 Optimise isProtectedTarget

### LINKED ISSUE #6305: Optimise `ContextHandler.isProtectedTarget`
The `isProtectedTarget` method is called on every new request.  It's implementation could be optimised to save trillions of CPU cycles!, but we must get it right as it is a key protection.

gregw:
Fix #6305 Optimise isProtectedTarget by using case insensitive Index.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
@@ -73,6 +73,7 @@ import org.eclipse.jetty.server.Request;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.util.Attributes;
 import org.eclipse.jetty.util.AttributesMap;
+import org.eclipse.jetty.util.Index;
 import org.eclipse.jetty.util.Loader;
 import org.eclipse.jetty.util.MultiException;
 import org.eclipse.jetty.util.StringUtil;
@@ -214,7 +215,7 @@ public class ContextHandler extends ScopedHandler implements Attributes, Gracefu
     private final List<ServletRequestAttributeListener> _servletRequestAttributeListeners = new CopyOnWriteArrayList<>();
     private final List<ContextScopeListener> _contextListeners = new CopyOnWriteArrayList<>();
     private final Set<EventListener> _durableListeners = new HashSet<>();
-    private String[] _protectedTargets;
+    private Index<Boolean> _protectedTargets = Index.empty(false);
     private final CopyOnWriteArrayList<AliasCheck> _aliasChecks = new CopyOnWriteArrayList<>();
 
     public enum Availability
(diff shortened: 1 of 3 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -1474,52 +1475,56 @@ public void handle(Runnable runnable)
...
-        }
 
-        for (int i = 0; i < _protectedTargets.length; i++)
-        {
-            String t = _protectedTargets[i];
-            if (StringUtil.startsWithIgnoreCase(target, t))
-            {
-                if (target.length() == t.length())
-                    return true;
+        Boolean exact = _protectedTargets.getBest(target);
```
janbartel: `Boolean` of `false` seems to me that it didn't match - I think that's confusing.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -1474,52 +1475,56 @@ public void handle(Runnable runnable)
...
-            return;
+            for (String t : targets)
+            {
+                if (!t.startsWith("/") || t.indexOf('/', 2) > 0)
+                    throw new IllegalArgumentException("Bad protected target: " + t);
+                // Index to TRUE if exact match required
+                builder.with(t, Boolean.TRUE);
+
+                // Index to FALSE for prefix matches
+                for (String s : new String[] {"/", "?", "#", ";"})
```
janbartel: The array should be a constant.

janbartel (changes requested):
If the Index class is more efficient, then great. A couple of niggles though.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -1474,52 +1475,56 @@ public void handle(Runnable runnable)
      */
     public boolean isProtectedTarget(String target)
     {
-        if (target == null || _protectedTargets == null)
+        if (target == null || _protectedTargets.isEmpty())
             return false;
 
-        while (target.startsWith("//"))
-        {
+        if (target.startsWith("//"))
```
lachlan-roberts, line 1491: Why do we only want to compact if it starts with `//`. We could have something like target=`/foo//////../foo-inf` which wouldn't be picked up by this check.

jetty-server/src/test/java/org/eclipse/jetty/server/handler/ContextHandlerTest.java
```diff
@@ -646,6 +646,13 @@ public void testProtected() throws Exception
         String[] protectedTargets = {"/foo-inf", "/bar-inf"};
         handler.setProtectedTargets(protectedTargets);
 
+        assertTrue(handler.isProtectedTarget("/foo-inf"));
+        assertTrue(handler.isProtectedTarget("/Foo-Inf"));
+        assertTrue(handler.isProtectedTarget("/FOO-INF"));
+        assertTrue(handler.isProtectedTarget("/foo-inf/"));
+        assertTrue(handler.isProtectedTarget("/FOO-inf?"));
+        assertTrue(handler.isProtectedTarget("/FOO-INF;"));
+        assertTrue(handler.isProtectedTarget("/foo-INF#"));
```
lachlan-roberts, line 655: Can we add some tests for paths that need to be compacted, `////foo-inf` etc.

gregw:
@lorban Is your performance testing able to spot any difference in this branch.  It is probably only very small improvement, but it is on every single request.

lachlan-roberts (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
gregw: Because we are matching only prefixes from the index.  So firstly we'd never see `/foo//////../foo-inf` here as it would be normalised to `/foo//////foo-inf`, then that will not match as it should not match. Even if the file system ignores the empty segments, it will the same as `/foo/foo-inf` which still shouldn't match.

Ultimately we only need to remove leading multiple //// but removing them all is simpler code and the extra cycles are not an issue as it should be very rare to see a target starting with `//`

lachlan-roberts (approved):

gregw:
@janbartel nudge

janbartel (approved):
+1

lorban:
@gregw my perf test measured a difference of about a 1.3% improvement with this change. Given the granularity of the test itself, I consider that the difference is in the noise.

PR #6308, opened 2021-05-21
Ensure buffers are returned to pool by MessageInputStream

lachlan-roberts:

### DIFF
--- a/jetty-websocket/websocket-common/src/main/java/org/eclipse/jetty/websocket/common/message/MessageInputStream.java
+++ b/jetty-websocket/websocket-common/src/main/java/org/eclipse/jetty/websocket/common/message/MessageInputStream.java
@@ -26,6 +26,7 @@ import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.TimeUnit;
 
 import org.eclipse.jetty.io.ByteBufferPool;
+import org.eclipse.jetty.io.NullByteBufferPool;
 import org.eclipse.jetty.util.BufferUtil;
 import org.eclipse.jetty.util.log.Log;
 import org.eclipse.jetty.util.log.Logger;
@@ -83,7 +84,7 @@ public class MessageInputStream extends InputStream implements MessageAppender
     {
         this.timeoutMs = timeoutMs;
         this.session = session;
-        this.bufferPool = (session instanceof WebSocketSession) ? ((WebSocketSession)session).getBufferPool() : null;
+        this.bufferPool = (session instanceof WebSocketSession) ? ((WebSocketSession)session).getBufferPool() : new NullByteBufferPool();
     }
 
     @Override
@@ -155,8 +156,21 @@ public class MessageInputStream extends InputStream implements MessageAppender
             if (remainingContent)
                 LOG.warn("MessageInputStream closed without fully consuming content {}", session);
 
-            state = State.CLOSED;
+
+            // Release any buffers taken from the pool.
+            if (activeBuffer != null && activeBuffer != EOF)
+            {
+                bufferPool.release(activeBuffer);
+                activeBuffer = null;
+            }
+
+            for (ByteBuffer buffer : buffers)
+            {
+                bufferPool.release(buffer);
+            }
+
             buffers.clear();
+            state = State.CLOSED;
             buffers.add(EOF);
         }
     }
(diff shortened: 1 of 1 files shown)


lachlan-roberts:
@olamy there is a failure for this branch but when I load it on jenkins to rerun I get 404 response.
How can I fix this?

olamy:
interesting as there is not failure here https://jenkins.webtide.net/job/jetty.project/view/change-requests/job/PR-6308/
As I was restarted Jenkins and doing some maintenance this morning maybe something went wrong during the build.
I will restart a build of the PR

jetty-websocket/websocket-common/src/main/java/org/eclipse/jetty/websocket/common/message/MessageInputStream.java
```diff
@@ -155,8 +156,21 @@ public void close()
             if (remainingContent)
                 LOG.warn("MessageInputStream closed without fully consuming content {}", session);
 
-            state = State.CLOSED;
+
+            // Release any buffers taken from the pool.
+            if (activeBuffer != null && activeBuffer != EOF)
```
gregw, line 161: can you add javadoc to the activeBuffer field saying that it is the head of the buffers queue (I think that is right?)

jetty-websocket/websocket-common/src/main/java/org/eclipse/jetty/websocket/common/message/MessageInputStream.java
```diff
@@ -309,11 +330,6 @@ public boolean markSupported()
 
     private ByteBuffer acquire(int capacity, boolean direct)
     {
-        ByteBuffer buffer;
-        if (bufferPool != null)
-            buffer = bufferPool.acquire(capacity, direct);
-        else
-            buffer = direct ? BufferUtil.allocateDirect(capacity) : BufferUtil.allocate(capacity);
-        return buffer;
+        return bufferPool.acquire(capacity, direct);
```
gregw, line 328: do we still need this method now? perhaps inline it.

jetty-websocket/websocket-common/src/main/java/org/eclipse/jetty/websocket/common/message/MessageInputStream.java
```diff
@@ -209,6 +223,13 @@ public int read(byte[] b, int off, int len) throws IOException
             // grab a fresh buffer
             while (activeBuffer == null || !activeBuffer.hasRemaining())
             {
+                if (activeBuffer != null)
+                {
+                    // All content consumed, release back to pool.
+                    bufferPool.release(activeBuffer);
```
gregw: why are we only releasing the active buffer before it is drained? Isn't there a case where we return below after consuming the active buffer and if read is not called again then it will not be released until close?

Isn't it better to release after the  ` if (!activeBuffer.hasRemaining())` branch below?

gregw (changes requested):

jetty-websocket/websocket-common/src/main/java/org/eclipse/jetty/websocket/common/message/MessageInputStream.java
lachlan-roberts: The loop condition here `while (activeBuffer == null || !activeBuffer.hasRemaining())` implies that we could have a non null `activeBuffer` in which case we would need to release it before getting a new one from the queue.

This whole loop seems strange to me, if the suspend/resume mechanism is working properly we should never have more than 1 buffer in the Queue at a time, and so we should be able to remove the queue and only use `activeBuffer`.

gregw: That sounds even better.  Can you verify that is the case? ie that you never have an activeBuffer and a non-zero queue?

gregw: Note that if you released active buffers when they are consumed, then there would be no need to release when you loop, as it will already have been released.

lachlan-roberts: I see why we need the Queue now, we put EOF buffer in with the last frame so we're not waiting forever.

But I have made the change to release buffers after they are consumed.

gregw: Wow, that is a bit of a cost to just deal with EOF state!  Couldn't that be another field that you only look at once activeBuffer is consumed.

Also, this still looks like the wrong spot for the release. Why is it not done at line 267 below?

gregw (approved):
LGTM other than some niggles

Issue #5379, opened 2020-10-03
Better handling for wrong SNI


sbordet:
**Jetty version**
9.4.x

**Description**
When the client sends a wrong SNI, and `jetty.ssl.sniHostCheck=true` Jetty should return a 400, but it returns 200.

sbordet:
@gregw strengthening the verifications when `jetty.ssl.sniHostCheck=true` fails a few tests that use a bogus certificate that has no CN and no SAN.

I don't think we should support bogus certificates like that, so I'm for do the strengthening and fix the tests.
WDYT?

joakime:
The 400 is returned from the SecureRequestCustomizer.  Is that present?

https://github.com/eclipse/jetty.project/blob/a76702baea01186d6349e62e4045968507e66649/jetty-server/src/main/java/org/eclipse/jetty/server/SecureRequestCustomizer.java#L246-L268

There's also a SNI check in SslContextFactory.Server

https://github.com/eclipse/jetty.project/blob/338ada6076ce454afeb10a88bc63e51dfb8c6dad/jetty-util/src/main/java/org/eclipse/jetty/util/ssl/SslContextFactory.java#L2146-L2158

PR #6311, opened 2021-05-21
Fixes #6276 - Support non-standard domains in SNI and X509. (#6296)

### LINKED ISSUE #6276: Support non-standard domains in SNI and X509
**Jetty version**
9.4.x

**Description**
Aggregates issues #6274 and #6275, as one is needed by the other for testing.

sbordet:
Improved support for IP addresses in X509 (after #5379).
Introduced SslContextFactory.Client.SniProvider to allow applications to specify the SNI names to send to the server.
Improved logging of SNI processing.
Skip X509 matching over IP addresses when the host does
not look like an IP address, to avoid reverse DNS lookup.

(cherry picked from commit 04df6d4ec2f232d94e8365b75dd7b38e68676abd)


sbordet:
Forward port of #6296.

jetty-util/src/main/java/org/eclipse/jetty/util/ssl/X509.java
```diff
@@ -182,21 +182,45 @@ public boolean matches(String host)
...
     }
 
+    private static boolean seemsIPAddress(String host)
+    {
+        // IPv4 is just numbers and dots.
+        String ipv4RegExp = "[0-9\\.]+";
+        // IPv6 is hex and colons and possibly brackets.
+        String ipv6RegExp = "[0-9a-fA-F:\\[\\]]+";
+        return host.matches(ipv4RegExp) ||
+            (host.matches(ipv6RegExp) && containsAtLeastTwoColons(host));
```
gregw: I missed this before the port... but shouldn't we precompile these regex's? 
Also they can be more precise: IPv4 must start and finish with a digit; IPv6 may have [] only on the outside and must have a : somewhere (which can be done in regex rather than another method).   i think this is specially important for ipv6 matching as `a-fA-F` can spell out words that might be local host names eg "fab" 

We may as well go for a complex address regex like https://stackoverflow.com/questions/53497/regular-expression-that-matches-valid-ipv6-addresses, which can be precompiled and will be quick enough to execute... ideally with a single expression rather than two.

gregw (changes requested):
sorry I missed this in original PR, but we can do better with the regex's

gregw (approved):

PR #6313, opened 2021-05-22
Update VERSION.TXT branch jetty-11.0.x

olamy:


gregw (changes requested):
This needs 10.03 and 9.4.41 changes in it

olamy:
I'm not sure if we do that when merging 10.0.x to 11.0.x or with this PR?
because merging #6314 will add 10.03 and 9.4.41 in VERSION.TXT

gregw (approved):

PR #6321, opened 2021-05-25
Issue #6118 Warn if hazelcast cfg file is missing SessionDataSerializer

### LINKED ISSUE #6118: Display a warning when Hazelcast configuration does not contain Jetty session serializer
It's possible to configure Hazelcast Session Manager with a single `hazelcast.xml` file but the Jetty session manager needs to have the serializer configured as is:
```
  <serialization>
    <serializers>
      <serializer type-class="org.eclipse.jetty.server.session.SessionData"
                  class-name="org.eclipse.jetty.hazelcast.session.SessionDataSerializer" />
    </serializers>
  </serialization> 
```
If it's not configured we should display a warning in the logs telling the user the Hazelcast Jetty session integration will not work

janbartel:
Fixes #6118 

Output a log warning if the user has supplied an optional hazelcast xml config file to configure hazelcast, and it is missing the serializer for the SessionData class.

### DIFF
--- a/jetty-hazelcast/src/main/java/org/eclipse/jetty/hazelcast/session/HazelcastSessionDataStoreFactory.java
+++ b/jetty-hazelcast/src/main/java/org/eclipse/jetty/hazelcast/session/HazelcastSessionDataStoreFactory.java
@@ -101,7 +101,7 @@ public class HazelcastSessionDataStoreFactory
                     {
                         config = new XmlClientConfigBuilder(configurationLocation).build();
                         if (config.getSerializationConfig().getSerializerConfigs().stream().noneMatch(s ->
-                        SessionData.class.getName().equals(s.getTypeClassName()) && s.getImplementation() instanceof SessionDataSerializer))
+                            SessionData.class.getName().equals(s.getTypeClassName()) && s.getImplementation() instanceof SessionDataSerializer))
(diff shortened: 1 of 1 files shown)


jetty-hazelcast/src/main/java/org/eclipse/jetty/hazelcast/session/HazelcastSessionDataStoreFactory.java
```diff
@@ -91,20 +96,22 @@ public SessionDataStore getSessionDataStore(SessionHandler handler)
...
                         config.getSerializationConfig().addSerializerConfig(sc);
-                        hazelcastInstance = HazelcastClient.newHazelcastClient(config);
                     }
                     else
                     {
-                        hazelcastInstance = HazelcastClient.newHazelcastClient(
-                            new XmlClientConfigBuilder(configurationLocation).build());
+                        config = new XmlClientConfigBuilder(configurationLocation).build();
+                        if (config.getSerializationConfig().getSerializerConfigs().stream().noneMatch(s ->
+                        SessionData.class.getName().equals(s.getTypeClassName()) && s.getImplementation() instanceof SessionDataSerializer))
```
olamy: if you fix this checkstyle issue it should be fine :P

olamy (changes requested):

olamy (approved):

PR #6332, opened 2021-05-28
Use RetainableByteBuffer and write a new pool for it

### LINKED ISSUE #6322: Use `RetainableByteBuffer` in `HttpConnection`
The `org.eclipse.jetty.server.HttpConnection`  class should use `org.eclipse.jetty.io.RetainableByteBuffer` rather than it's own reference counting of buffers.  This will eventually allow a more efficient buffer pool abstraction to be used that provides `RetainableByteBuffers` directly.

lorban:
Fix #6322 
`HttpConnection` can be modified to make use of `RetainableByteBuffer` with little effort, which indeed allows the use of a much more efficient pool type.

With the above change in place, we can make use of `o.e.j.u.Pool` to create a `RetainableByteBuffer` pool for all input buffer needs as that design does bring a significant perf improvement.

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
@@ -54,10 +54,12 @@ import org.eclipse.jetty.http.HttpHeader;
 import org.eclipse.jetty.http.HttpMethod;
 import org.eclipse.jetty.http.HttpParser;
 import org.eclipse.jetty.http.HttpScheme;
+import org.eclipse.jetty.io.ArrayRetainableByteBufferPool;
 import org.eclipse.jetty.io.ByteBufferPool;
 import org.eclipse.jetty.io.ClientConnectionFactory;
 import org.eclipse.jetty.io.ClientConnector;
 import org.eclipse.jetty.io.MappedByteBufferPool;
+import org.eclipse.jetty.io.RetainableByteBufferPool;
 import org.eclipse.jetty.io.ssl.SslClientConnectionFactory;
 import org.eclipse.jetty.util.Fields;
 import org.eclipse.jetty.util.Jetty;
@@ -193,12 +195,14 @@ public class HttpClient extends ContainerLifeCycle
             threadPool.setName(name);
             setExecutor(threadPool);
         }
+        int maxBucketSize = executor instanceof ThreadPool.SizedThreadPool
+            ? ((ThreadPool.SizedThreadPool)executor).getMaxThreads() / 2
+            : ProcessorUtils.availableProcessors() * 2;
         ByteBufferPool byteBufferPool = getByteBufferPool();
         if (byteBufferPool == null)
-            setByteBufferPool(new MappedByteBufferPool(2048,
-                executor instanceof ThreadPool.SizedThreadPool
-                    ? ((ThreadPool.SizedThreadPool)executor).getMaxThreads() / 2
-                    : ProcessorUtils.availableProcessors() * 2));
+            setByteBufferPool(new MappedByteBufferPool(2048, maxBucketSize));
+        if (getBean(RetainableByteBufferPool.class) == null)
+            addBean(new ArrayRetainableByteBufferPool(0, 2048, 65536, maxBucketSize));
         Scheduler scheduler = getScheduler();
         if (scheduler == null)
             setScheduler(new ScheduledExecutorScheduler(name + "-scheduler", false));
--- a/jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Connection.java
+++ b/jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Connection.java
@@ -23,10 +23,10 @@ import java.util.concurrent.atomic.AtomicLong;
 import org.eclipse.jetty.http2.frames.DataFrame;
 import org.eclipse.jetty.http2.parser.Parser;
 import org.eclipse.jetty.io.AbstractConnection;
-import org.eclipse.jetty.io.ByteBufferPool;
 import org.eclipse.jetty.io.Connection;
 import org.eclipse.jetty.io.EndPoint;
 import org.eclipse.jetty.io.RetainableByteBuffer;
+import org.eclipse.jetty.io.RetainableByteBufferPool;
 import org.eclipse.jetty.io.WriteFlusher;
 import org.eclipse.jetty.util.BufferUtil;
 import org.eclipse.jetty.util.Callback;
@@ -45,7 +45,7 @@ public class HTTP2Connection extends AbstractConnection implements WriteFlusher.
     private final Queue<Runnable> tasks = new ArrayDeque<>();
     private final HTTP2Producer producer = new HTTP2Producer();
     private final AtomicLong bytesIn = new AtomicLong();
-    private final ByteBufferPool byteBufferPool;
+    private final RetainableByteBufferPool retainableByteBufferPool;
     private final Parser parser;
     private final ISession session;
     private final int bufferSize;
@@ -53,10 +53,10 @@ public class HTTP2Connection extends AbstractConnection implements WriteFlusher.
     private boolean useInputDirectByteBuffers;
     private boolean useOutputDirectByteBuffers;
 
-    public HTTP2Connection(ByteBufferPool byteBufferPool, Executor executor, EndPoint endPoint, Parser parser, ISession session, int bufferSize)
+    protected HTTP2Connection(RetainableByteBufferPool retainableByteBufferPool, Executor executor, EndPoint endPoint, Parser parser, ISession session, int bufferSize)
     {
         super(endPoint, executor);
-        this.byteBufferPool = byteBufferPool;
+        this.retainableByteBufferPool = retainableByteBufferPool;
         this.parser = parser;
         this.session = session;
         this.bufferSize = bufferSize;
@@ -287,7 +287,7 @@ public class HTTP2Connection extends AbstractConnection implements WriteFlusher.
                             return task;
 
                         // If more references than 1 (ie not just us), don't refill into buffer and risk compaction.
-                        if (networkBuffer.getReferences() > 1)
+                        if (networkBuffer.isRetained())
                             reacquireNetworkBuffer();
                     }
 
@@ -415,16 +415,43 @@ public class HTTP2Connection extends AbstractConnection implements WriteFlusher.
         }
     }
 
-    private class NetworkBuffer extends RetainableByteBuffer implements Callback
+    private class NetworkBuffer implements Callback
     {
+        private final RetainableByteBuffer delegate;
+
         private NetworkBuffer()
         {
-            super(byteBufferPool, bufferSize, isUseInputDirectByteBuffers());
+            delegate = retainableByteBufferPool.acquire(bufferSize, isUseInputDirectByteBuffers());
+        }
+
+        public ByteBuffer getBuffer()
+        {
+            return delegate.getBuffer();
+        }
+
+        public boolean isRetained()
+        {
+            return delegate.isRetained();
+        }
+
+        public boolean hasRemaining()
+        {
+            return delegate.hasRemaining();
+        }
+
+        public boolean release()
+        {
+            return delegate.release();
+        }
+
+        public void retain()
+        {
+            delegate.retain();
         }
 
         private void put(ByteBuffer source)
         {
-            BufferUtil.append(getBuffer(), source);
+            BufferUtil.append(delegate.getBuffer(), source);
         }
 
         @Override
@@ -441,7 +468,7 @@ public class HTTP2Connection extends AbstractConnection implements WriteFlusher.
 
         private void completed(Throwable failure)
         {
-            if (release() == 0)
+            if (delegate.release())
             {
                 if (LOG.isDebugEnabled())
                     LOG.debug("Released retained {}", this, failure);
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
@@ -15,32 +15,40 @@ package org.eclipse.jetty.io;
 
 import java.nio.ByteBuffer;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Consumer;
 
 import org.eclipse.jetty.util.BufferUtil;
 import org.eclipse.jetty.util.Retainable;
 
 /**
- * A Retainable ByteBuffer.
- * <p>Acquires a ByteBuffer from a {@link ByteBufferPool} and maintains a reference count that is
- * initially 1, incremented with {@link #retain()} and decremented with {@link #release()}. The buffer
- * is released to the pool when the reference count is decremented to 0.</p>
+ * <p>A pooled ByteBuffer which maintains a reference count that is
+ * incremented with {@link #retain()} and decremented with {@link #release()}. The buffer
+ * is released to the pool when {@link #release()} is called one more time than {@link #retain()}.</p>
+ * <p>A {@code RetainableByteBuffer} can either be:
+ * <ul>
+ *     <li>in pool; in this case {@link #isRetained()} returns {@code false} and calling {@link #release()} throws {@link IllegalStateException}</li>
+ *     <li>out of pool but not retained; in this case {@link #isRetained()} returns {@code false} and calling {@link #release()} returns {@code true}</li>
+ *     <li>out of pool and retained; in this case {@link #isRetained()} returns {@code true} and calling {@link #release()} returns {@code false}</li>
+ * </ul>
+ * <p>Calling {@link #release()} on a out of pool and retained instance does not re-pool it while that re-pools it on a out of pool but not retained instance.</p>
  */
 public class RetainableByteBuffer implements Retainable
 {
-    private final ByteBufferPool pool;
     private final ByteBuffer buffer;
-    private final AtomicInteger references;
+    private final AtomicInteger references = new AtomicInteger();
+    private final Consumer<ByteBuffer> releaser;
+    private final AtomicLong lastUpdate = new AtomicLong(System.nanoTime());
 
-    public RetainableByteBuffer(ByteBufferPool pool, int size)
+    RetainableByteBuffer(ByteBuffer buffer, Consumer<ByteBuffer> releaser)
     {
-        this(pool, size, false);
+        this.releaser = releaser;
+        this.buffer = buffer;
     }
 
-    public RetainableByteBuffer(ByteBufferPool pool, int size, boolean direct)
+    public int capacity()
     {
-        this.pool = pool;
-        this.buffer = pool.acquire(size, direct);
-        this.references = new AtomicInteger(1);
+        return buffer.capacity();
     }
 
     public ByteBuffer getBuffer()
@@ -48,32 +56,66 @@ public class RetainableByteBuffer implements Retainable
         return buffer;
     }
 
-    public int getReferences()
+    public long getLastUpdate()
     {
-        return references.get();
+        return lastUpdate.getOpaque();
     }
 
+    /**
+     * Checks if {@link #retain()} has been called at least one more time than {@link #release()}.
+     * @return true if this buffer is retained, false otherwise.
+     */
+    public boolean isRetained()
+    {
+        return references.get() > 1;
+    }
+
+    public boolean isDirect()
+    {
+        return buffer.isDirect();
+    }
+
+    /**
+     * Increments the retained counter of this buffer. It must be done internally by
+     * the pool right after creation and after each un-pooling.
+     * The reason why this method exists on top of {@link #retain()} is to be able to
+     * have some safety checks that must know why the ref counter is being incremented.
+     */
+    void acquire()
+    {
+        if (references.getAndUpdate(c -> c == 0 ? 1 : c) != 0)
+            throw new IllegalStateException("re-pooled while still used " + this);
+    }
+
+    /**
+     * Increments the retained counter of this buffer.
+     */
     @Override
     public void retain()
     {
-        while (true)
-        {
-            int r = references.get();
-            if (r == 0)
-                throw new IllegalStateException("released " + this);
-            if (references.compareAndSet(r, r + 1))
-                break;
-        }
+        if (references.getAndUpdate(c -> c == 0 ? 0 : c + 1) == 0)
+            throw new IllegalStateException("released " + this);
     }
 
-    public int release()
+    /**
+     * Decrements the retained counter of this buffer.
+     * @return true if the buffer was re-pooled, false otherwise.
+     */
+    public boolean release()
     {
-        int ref = references.decrementAndGet();
+        int ref = references.updateAndGet(c ->
+        {
+            if (c == 0)
+                throw new IllegalStateException("already released " + this);
+            return c - 1;
+        });
         if (ref == 0)
-            pool.release(buffer);
-        else if (ref < 0)
-            throw new IllegalStateException("already released " + this);
-        return ref;
+        {
+            lastUpdate.setOpaque(System.nanoTime());
+            releaser.accept(buffer);
+            return true;
+        }
+        return false;
     }
 
     public int remaining()
@@ -99,6 +141,6 @@ public class RetainableByteBuffer implements Retainable
     @Override
     public String toString()
     {
-        return String.format("%s@%x{%s,r=%d}", getClass().getSimpleName(), hashCode(), BufferUtil.toDetailString(buffer), getReferences());
+        return String.format("%s@%x{%s,r=%d}", getClass().getSimpleName(), hashCode(), BufferUtil.toDetailString(buffer), references.get());
     }
 }
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
@@ -1,0 +1,58 @@
+//
+// ========================================================================
+// Copyright (c) 1995-2021 Mort Bay Consulting Pty Ltd and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.io;
+
+import java.nio.ByteBuffer;
+
+import org.eclipse.jetty.util.component.Container;
+
+/**
+ * <p>A {@link RetainableByteBuffer} pool.</p>
+ * <p>Acquired buffers <b>must</b> be released by calling {@link RetainableByteBuffer#release()} otherwise the memory they hold will
+ * be leaked.</p>
+ */
+public interface RetainableByteBufferPool
+{
+    /**
+     * Acquires a memory buffer from the pool.
+     * @param size The size of the buffer. The returned buffer will have at least this capacity.
+     * @param direct true if a direct memory buffer is needed, false otherwise.
+     * @return a memory buffer.
+     */
+    RetainableByteBuffer acquire(int size, boolean direct);
+
+    /**
+     * Finds a {@link RetainableByteBufferPool} implementation in the given container, or wrap the given
+     * {@link ByteBufferPool} with an adapter.
+     * @param container the container to search for an existing memory pool.
+     * @param byteBufferPool the {@link ByteBufferPool} to wrap if no memory pool was found in the container.
+     * @return the {@link RetainableByteBufferPool} found or the wrapped one.
+     */
+    static RetainableByteBufferPool findOrAdapt(Container container, ByteBufferPool byteBufferPool)
+    {
+        RetainableByteBufferPool retainableByteBufferPool = container == null ? null : container.getBean(RetainableByteBufferPool.class);
+        if (retainableByteBufferPool == null)
+        {
+            // Wrap the ByteBufferPool instance.
+            retainableByteBufferPool = (size, direct) ->
+            {
+                ByteBuffer byteBuffer = byteBufferPool.acquire(size, direct);
+                RetainableByteBuffer retainableByteBuffer = new RetainableByteBuffer(byteBuffer, byteBufferPool::release);
+                retainableByteBuffer.acquire();
+                return retainableByteBuffer;
+            };
+        }
+        return retainableByteBufferPool;
+    }
+}
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
@@ -35,6 +35,8 @@ import org.eclipse.jetty.io.AbstractEndPoint;
 import org.eclipse.jetty.io.ByteBufferPool;
 import org.eclipse.jetty.io.Connection;
 import org.eclipse.jetty.io.EndPoint;
+import org.eclipse.jetty.io.RetainableByteBuffer;
+import org.eclipse.jetty.io.RetainableByteBufferPool;
 import org.eclipse.jetty.io.WriteFlusher;
 import org.eclipse.jetty.util.BufferUtil;
 import org.eclipse.jetty.util.Callback;
@@ -107,10 +109,11 @@ public class SslConnection extends AbstractConnection implements Connection.Upgr
     private final AtomicLong _bytesIn = new AtomicLong();
     private final AtomicLong _bytesOut = new AtomicLong();
     private final ByteBufferPool _bufferPool;
+    private final RetainableByteBufferPool _retainableByteBufferPool;
     private final SSLEngine _sslEngine;
     private final DecryptedEndPoint _decryptedEndPoint;
     private ByteBuffer _decryptedInput;
-    private ByteBuffer _encryptedInput;
+    private RetainableByteBuffer _encryptedInput;
     private ByteBuffer _encryptedOutput;
     private final boolean _encryptedDirectBuffers;
     private final boolean _decryptedDirectBuffers;
@@ -188,10 +191,17 @@ public class SslConnection extends AbstractConnection implements Connection.Upgr
     public SslConnection(ByteBufferPool byteBufferPool, Executor executor, EndPoint endPoint, SSLEngine sslEngine,
                          boolean useDirectBuffersForEncryption, boolean useDirectBuffersForDecryption)
     {
+        this(RetainableByteBufferPool.findOrAdapt(null, byteBufferPool), byteBufferPool, executor, endPoint, sslEngine, useDirectBuffersForEncryption, useDirectBuffersForDecryption);
+    }
+
+    public SslConnection(RetainableByteBufferPool retainableByteBufferPool, ByteBufferPool byteBufferPool, Executor executor, EndPoint endPoint, SSLEngine sslEngine,
+                         boolean useDirectBuffersForEncryption, boolean useDirectBuffersForDecryption)
+    {
         // This connection does not execute calls to onFillable(), so they will be called by the selector thread.
         // onFillable() does not block and will only wakeup another thread to do the actual reading and handling.
         super(endPoint, executor);
         this._bufferPool = byteBufferPool;
+        this._retainableByteBufferPool = retainableByteBufferPool;
         this._sslEngine = sslEngine;
         this._decryptedEndPoint = newDecryptedEndPoint();
         this._encryptedDirectBuffers = useDirectBuffersForEncryption;
@@ -326,7 +336,7 @@ public class SslConnection extends AbstractConnection implements Connection.Upgr
     private void acquireEncryptedInput()
     {
         if (_encryptedInput == null)
-            _encryptedInput = _bufferPool.acquire(getPacketBufferSize(), _encryptedDirectBuffers);
+            _encryptedInput = _retainableByteBufferPool.acquire(getPacketBufferSize(), _encryptedDirectBuffers);
     }
 
     private void acquireEncryptedOutput()
@@ -339,7 +349,7 @@ public class SslConnection extends AbstractConnection implements Connection.Upgr
     public void onUpgradeTo(ByteBuffer buffer)
     {
         acquireEncryptedInput();
-        BufferUtil.append(_encryptedInput, buffer);
+        BufferUtil.append(_encryptedInput.getBuffer(), buffer);
     }
 
     @Override
@@ -409,7 +419,7 @@ public class SslConnection extends AbstractConnection implements Connection.Upgr
     @Override
     public String toConnectionString()
     {
-        ByteBuffer b = _encryptedInput;
+        ByteBuffer b = _encryptedInput == null ? null : _encryptedInput.getBuffer();
         int ei = b == null ? -1 : b.remaining();
         b = _encryptedOutput;
         int eo = b == null ? -1 : b.remaining();
@@ -431,7 +441,7 @@ public class SslConnection extends AbstractConnection implements Connection.Upgr
     {
         if (_encryptedInput != null && !_encryptedInput.hasRemaining())
         {
-            _bufferPool.release(_encryptedInput);
+            _encryptedInput.release();
             _encryptedInput = null;
         }
     }
@@ -672,14 +682,14 @@ public class SslConnection extends AbstractConnection implements Connection.Upgr
                             }
 
                             // Let's try reading some encrypted data... even if we have some already.
-                            int netFilled = networkFill(_encryptedInput);
+                            int netFilled = networkFill(_encryptedInput.getBuffer());
                             if (netFilled > 0)
                                 _bytesIn.addAndGet(netFilled);
                             if (LOG.isDebugEnabled())
                                 LOG.debug("net filled={}", netFilled);
 
                             // Workaround for Java 11 behavior.
-                            if (netFilled < 0 && isHandshakeInitial() && BufferUtil.isEmpty(_encryptedInput))
+                            if (netFilled < 0 && isHandshakeInitial() && (_encryptedInput == null || _encryptedInput.isEmpty()))
                                 closeInbound();
 
                             if (netFilled > 0 && !isHandshakeComplete() && isOutboundDone())
@@ -698,7 +708,7 @@ public class SslConnection extends AbstractConnection implements Connection.Upgr
                             try
                             {
                                 _underflown = false;
-                                unwrapResult = SslConnection.this.unwrap(_sslEngine, _encryptedInput, appIn);
+                                unwrapResult = SslConnection.this.unwrap(_sslEngine, _encryptedInput.getBuffer(), appIn);
                             }
                             finally
                             {
@@ -708,7 +718,7 @@ public class SslConnection extends AbstractConnection implements Connection.Upgr
                                 LOG.debug("unwrap net_filled={} {} encryptedBuffer={} unwrapBuffer={} appBuffer={}",
                                     netFilled,
                                     StringUtil.replace(unwrapResult.toString(), '\n', ' '),
-                                    BufferUtil.toSummaryString(_encryptedInput),
+                                    _encryptedInput,
                                     BufferUtil.toDetailString(appIn),
                                     BufferUtil.toDetailString(buffer));
 
@@ -729,13 +739,13 @@ public class SslConnection extends AbstractConnection implements Connection.Upgr
 
                                 case BUFFER_UNDERFLOW:
                                     // Continue if we can compact?
-                                    if (BufferUtil.compact(_encryptedInput))
+                                    if (BufferUtil.compact(_encryptedInput.getBuffer()))
                                         continue;
 
                                     // Are we out of space?
-                                    if (BufferUtil.space(_encryptedInput) == 0)
+                                    if (BufferUtil.space(_encryptedInput.getBuffer()) == 0)
                                     {
-                                        BufferUtil.clear(_encryptedInput);
+                                        BufferUtil.clear(_encryptedInput.getBuffer());
                                         throw new SSLHandshakeException("Encrypted buffer max length exceeded");
                                     }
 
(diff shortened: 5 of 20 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -60,8 +60,7 @@
     private final HttpGenerator _generator;
     private final HttpChannelOverHttp _channel;
     private final HttpParser _parser;
-    private final AtomicInteger _contentBufferReferences = new AtomicInteger();
-    private volatile ByteBuffer _requestBuffer = null;
+    private volatile RetainableByteBuffer _retainableByteBuffer;
```
gregw, line 65: I know this is a draft... but thought I'd lob in some comments anyway....

having a volatile reference to a class that itself holds a reference to a buffer is a bit too much. I'd like to see:

```suggestion
    private final RetainableByteBuffer _retainableByteBuffer;
```

which means the `RetainableByteBuffer` needs some modifications... but it will anyway if it is to end up being the pooled instance

lorban:
I've also modified `SslConnection` because its encrypted input buffer also appears as a hotspot when SSL is enabled, as show in the following flame graph: [profile.4.zip](https://github.com/eclipse/jetty.project/files/6575085/profile.4.zip)

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
lorban: Sounds like a case of premature optimization :smiley:. Let's first see how the `RetainableByteBuffer` pool API pans out.

lorban: So far, my perf tests tend to show that any loss due to the double indirection is largely compensated by the much more efficient pooling algorithm that this change allows.

jetty-io/src/main/java/org/eclipse/jetty/io/AdapterMemoryPool.java
```diff
@@ -0,0 +1,42 @@
...
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.io;
+
+import java.nio.ByteBuffer;
+import java.util.function.Consumer;
+
+public class AdapterMemoryPool implements MemoryPool<RetainableByteBuffer>
```
gregw: Javadoc needed.   Not sure why the name is "Adapter"?

Why not `RetainableMemoryPool`?    Hmmm but that is confusing with RBBP??? OK not seeing what this class is for?

jetty-io/src/main/java/org/eclipse/jetty/io/MemoryPool.java
```diff
@@ -0,0 +1,21 @@
...
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.io;
+
+public interface MemoryPool<T>
```
gregw: javadoc needed.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
```diff
@@ -27,20 +28,20 @@
  */
 public class RetainableByteBuffer implements Retainable
 {
```
gregw, line 37: javadoc above is now wrong.  No `ByteBufferPool` is passed.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
```diff
@@ -0,0 +1,154 @@
...
+        _minCapacity = minCapacity;
+        if (maxCapacity <= 0)
+            maxCapacity = 64 * 1024;
+        if ((maxCapacity % _factor) != 0 || _factor >= maxCapacity)
+            throw new IllegalArgumentException("The capacity factor must be a divisor of maxCapacity");
+
+        int length = maxCapacity / _factor;
+
+        _direct = new AtomicReferenceArray<>(length);
+        _indirect = new AtomicReferenceArray<>(length);
```
gregw: What is the saving making these arrays atomic?  Would it not be better to allocate all at once?      If the thinking is that most buckets will not be used, so we will not create most of those sub arrays, then we have to be careful we don't expose ourselves to DOS attacks where an attacker somehow forces all the buckets to be allocated. If we can't guarantee that we can prevent all the buckets being allocated then we should allocate them all up front so we know we have space.

Perhaps we need a better than a linear slice to allocate buckets.  If we use 8k and 16k buckets, will we really ever have need of 9-15k buckets?   What about a simple log bucket calculation so buckets go 1K, 2K, 4K, 8K, 16K, 32K & 64K - so 7 buckets for a fully populated pool. 

Hmm an empty bucket is just an empty COW array list right??? So not expensive to fully allocate all buckets even if linear... but I still think there is a DOS  issue with linear allocated buckets, so log buckets are better.

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
```diff
@@ -188,6 +189,9 @@ public AbstractConnector(
             pool = _server.getBean(ByteBufferPool.class);
         _byteBufferPool = pool != null ? pool : new ArrayByteBufferPool();
         addBean(_byteBufferPool);
+        RetainableByteBufferPool retainableByteBufferPool = _server.getBean(RetainableByteBufferPool.class);
+        retainableByteBufferPool = retainableByteBufferPool != null ? retainableByteBufferPool : new RetainableByteBufferPool();
+        addBean(retainableByteBufferPool);
```
gregw: might be worthwhile to be explict here with the management attribute.  ie only if it is a new pool should we manage it.   I know this probably happens automagically, but no harm being explicit.

jetty-io/src/main/java/org/eclipse/jetty/io/AdapterMemoryPool.java
```diff
@@ -0,0 +1,42 @@
...
+{
+    private final ByteBufferPool byteBufferPool;
+    private final Consumer<ByteBuffer> releaser;
```
gregw: why is this field needed?  `byteBufferPool::release` can be passed directly the the RBB

gregw (changes requested):
Just some initial feedback.  More review needed.

jetty-io/src/main/java/org/eclipse/jetty/io/AdapterMemoryPool.java
lorban: True, I added this field because I thought directly passing `byteBufferPool::release` to `RetainableByteBuffer` would need to allocate. Now I'm not certain about this.

jetty-io/src/main/java/org/eclipse/jetty/io/AdapterMemoryPool.java
lorban: `AdapterMemoryPool` adapts the old `ByteBufferPool` to the new `RetainableByteBufferPool` abstracted out as `MemoryPool<RetainableByteBuffer>`. This allows the input side to use `MemoryPool<RetainableByteBuffer>` only while still being able to work with `ByteBufferPool` implementations.

The reason for this mechanism is backward compatibility: there are components that used to expect a `ByteBufferPool` that now need a `MemoryPool<RetainableByteBuffer>`. Special care is going to be needed to figure out exactly what breaking changes we're okay with and what needs to be kept compatible.

jetty-io/src/main/java/org/eclipse/jetty/io/MemoryPool.java
lorban: Added some.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
lorban: Fixed.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
lorban: This mechanism is almost identical to the `ArrayByteBuffer` one. The only difference is that I used atomic ref arrays instead of plain arrays to avoid creating non-empty buckets that could be dropped as garbage. 

The current algorithm served us well until now, so if we want to modify it, I think that should be done as a separate PR. For instance, using log bucket sizes already opens questions about the SSL case which uses 17K buffers. And I do not think we're open for DoS attacks as AFAIK the clients have no control over the bucket sizes.

Though, I don't think pre-allocating the buckets would be that costly as those are indeed empty COW array lists, so I guess I could do that instead of using atomic ref arrays.

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
lorban: I once again just copied the logic used for the `_byteBufferPool` field. Do we want a different one for `retainableByteBufferPool`?

jetty-io/src/main/java/org/eclipse/jetty/io/AdapterMemoryPool.java
gregw: hmmm maybe it does allocated.    so leave as is unless you test that it doesn't

jetty-io/src/main/java/org/eclipse/jetty/io/AdapterMemoryPool.java
gregw: So for backwards compatibility we have the two cases to consider:
1. Somebody has written a component expecting to be able to find a ByteBufferPool as a bean on the server, but we have only added a RetainableByteBuffer.
2. We have a component that now expects to find a RetainableByteBuffer on the server, but the deployer has only added a ByteBufferPool

So this adaptor is useful for the later case, but do we have something that will work for the former case?


Perhaps there should be some static methods somewhere that capture both search algorithms:

1. look for a ByteBufferPool, if found use it, else look for a RetainableButeBufferPool, if found adapt it??? else create a new ByteBufferPool
2. look for a RetainableByteBufferPool, if found use it, else look for a ByteBufferPool, if found then adapt it (and emit a warning or at least an info), else create a new RetainableByteBufferPool.

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
gregw: I think it is worthwhile being explicit, since the order of creating/starting pools might not be straight forward with backwards compatibility logic.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
gregw: I think there will be performance savings with final arrays and preallocated buckets, with little cost.
I agree that we should punt log arrays to a different issue.

jetty-io/src/main/java/org/eclipse/jetty/io/MemoryPool.java
gregw: A bit minimal.  I think you need to say that this is a Pool of objects representing memory that can be acquired based on size.  The held instances may be the memory component itself (eg ByteBuffer) or an abstraction providing access to the memory component

jetty-io/src/main/java/org/eclipse/jetty/io/AdapterMemoryPool.java
lorban: Case 1 (_Somebody has written a component expecting to be able to find a ByteBufferPool as a bean on the server, but we have only added a RetainableByteBuffer._) cannot happen as we still rely on `ByteBufferPool` for output buffers, so there should always be one available.

Maybe we should log when we cannot find a `RetainableByteBufferPool` and have to use an adapter, but @sbordet usually does not like such warnings.

jetty-io/src/main/java/org/eclipse/jetty/io/MemoryPool.java
lorban: Thanks for the description, I improved the javadoc based on it.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
lorban: Agreed, I've changed the implementation to use pre-allocated `Pool` arrays instead.

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
lorban: Argh, can't do as `AbstractHTTP2ServerConnectionFactory` calls `getBean()` on the connector to get the `RetainableByteBufferPool` instance, so this bean must always be registered otherwise the legacy `ByteBufferPool` is going to be used instead.

gregw: I'm not saying don't register it.  I'm saying register it with `addBean(Object, boolean)` so you can pass in the managed status of it.
Currently if you just add a bean that is itself a lifecycle, it gets added in automode.   Then when the container is started, if it also has to start the auto bean, it changes it to managed.  If the autobean is already started it is changed to not-managed.

I'm suggesting that if a pool is found on the server, you should `addBean(pool, false)` as we should never manage a server bean.  If we create a new pool, then it should be `addBean(pool, true)` so that we explicitly manage it.

gregw: Something like:
```java
    RetainableByteBufferPool pool = _server.getBean(RetainableByteBufferPool.class);
    addBean(pool == null ? new RetainableByteBufferPool() : pool, pool == null);
```

```

File: jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java [hunk 8]
```diff
@@ -187,11 +191,18 @@ public SslConnection(ByteBufferPool byteBufferPool, Executor executor, EndPoint
 
     public SslConnection(ByteBufferPool byteBufferPool, Executor executor, EndPoint endPoint, SSLEngine sslEngine,
                          boolean useDirectBuffersForEncryption, boolean useDirectBuffersForDecryption)
+    {
+        this(new AdapterMemoryPool(byteBufferPool), byteBufferPool, executor, endPoint, sslEngine, useDirectBuffersForEncryption, useDirectBuffersForDecryption);
```
gregw: This should be findOrAdapt

jetty-io/src/main/java/org/eclipse/jetty/io/MemoryPool.java
```diff
@@ -0,0 +1,53 @@
...
+    void release(T buffer);
+
+    /**
+     * Find a {@link MemoryPool} of {@link RetainableByteBuffer} implementation in the given container, or wrap the given
+     * {@link ByteBufferPool} with an adapter.
+     * @param container the container to search for an existing memory pool.
+     * @param byteBufferPool the {@link ByteBufferPool} to wrap if no memory pool was found in the container.
+     * @return the memory pool found or the wrapped one.
+     */
+    static MemoryPool<RetainableByteBuffer> findOrAdapt(Container container, ByteBufferPool byteBufferPool)
```
gregw: I like the method, but it should be on `RetainableByteBufferPool`. I think that would allow AdaptorMemoryPool to be entirely private.

Also, perhaps there should be an option or variant to create a real RetainableByteBufferPool and add it to the container.   In websocket we name such methods `ensureRetainableByteBufferPool`.

jetty-io/src/main/java/org/eclipse/jetty/io/AdapterMemoryPool.java
gregw: Wouldn't it be better to have the RBBP wrapped to look like a plain BBP?    Although that will just increase contention... and if input/output buffers sizes are different, then there is little overlap.     OK.

gregw (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
lorban: Oh okay, I didn't get that but it makes sense now. I'll make this change.

jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
lorban: With an extra null check in the implementation, yes, that can be done. I'm okay with this idea as it would then be possible to make `AdapterMemoryPool` private as you pointed out below.

jetty-io/src/main/java/org/eclipse/jetty/io/MemoryPool.java
lorban: Agreed that this method should move to `RetainableByteBufferPool` to allow `AdapterMemoryPool` to be private.

I'm also going to have a look at what `ensureRetainableByteBufferPool()` would look like.

lorban: `ensureRetainableByteBufferPool` doesn't make much sense IMHO as there only are two places where `RetainableByteBufferPool` is instantiated and they both use different constructors and do not lookup and register in containers in the same way.

jetty-io/src/main/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPool.java
```diff
@@ -238,20 +236,20 @@ public static RetainableByteBufferPool findOrAdapt(Container container, ByteBuff
...
+            retainableByteBufferPool = new ByteBufferToRetainableByteBufferAdapterPool(byteBufferPool);
         return retainableByteBufferPool;
     }
 
     /**
      * An adapter class which exposes a {@link ByteBufferPool} as a
      * {@link RetainableByteBufferPool}.
      */
-    private static class AdapterMemoryPool implements RetainableByteBufferPool
+    private static class ByteBufferToRetainableByteBufferAdapterPool implements RetainableByteBufferPool
```
joakime: Wow, that's a mouthful, thankfully it's an internal / private class.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
```diff
@@ -193,12 +195,14 @@ protected void doStart() throws Exception
...
+            : ProcessorUtils.availableProcessors() * 2;
         ByteBufferPool byteBufferPool = getByteBufferPool();
         if (byteBufferPool == null)
-            setByteBufferPool(new MappedByteBufferPool(2048,
-                executor instanceof ThreadPool.SizedThreadPool
-                    ? ((ThreadPool.SizedThreadPool)executor).getMaxThreads() / 2
-                    : ProcessorUtils.availableProcessors() * 2));
+            setByteBufferPool(new MappedByteBufferPool(2048, maxBucketSize));
+        if (getBean(RetainableByteBufferPool.class) == null)
+            addBean(new DefaultRetainableByteBufferPool(0, 2048, 65536, maxBucketSize));
```
gregw, line 205:
```suggestion
            addBean(new DefaultRetainableByteBufferPool(0, 2048, 65536, maxBucketSize), true);
```

jetty-io/src/main/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPool.java
lorban: Any suggestion for a better name?

gregw: It's private so doesn't really matter as there is good context. `Adaptor` works for me.

gregw (approved):
Let's get this in early during the next release cycle!

lorban:
Looking back, I just noticed that `DefaultRetainableByteBufferPool` does not have an eviction mechanism to avoid the pool to accumulate an infinite amount of buffers. Duh!

I'm going to work on this right away to make sure this PR is not removing any existing functionality.

gregw (approved):

gregw (changes requested):
fix what ludo said

jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Connection.java
```diff
@@ -421,16 +421,44 @@ public void onConnectionFailure(int error, String reason)
         }
     }
 
-    private class NetworkBuffer extends RetainableByteBuffer implements Callback
+    private class NetworkBuffer implements Callback
```
sbordet, line 418: Why it is not extending `RetainableByteBuffer` anymore?
Seems a lot of code more for no gain.

jetty-io/src/main/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPool.java
```diff
@@ -0,0 +1,390 @@
...
+    }
+
+    /**
+     * Find a {@link RetainableByteBufferPool} implementation in the given container, or wrap the given
+     * {@link ByteBufferPool} with an adapter.
+     * @param container the container to search for an existing memory pool.
+     * @param byteBufferPool the {@link ByteBufferPool} to wrap if no memory pool was found in the container.
+     * @return the {@link RetainableByteBufferPool} found or the wrapped one.
+     */
+    public static RetainableByteBufferPool findOrAdapt(Container container, ByteBufferPool byteBufferPool)
```
sbordet: I would prefer this method to be in the interface, and the adapter class be a method-local class.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * <p>A {@link RetainableByteBuffer} pool.</p>
+ * <p>Acquired buffers <b>must</b> be released by calling {@link RetainableByteBuffer#release()} otherwise the memory they hold will
+ * be leaked.</p>
+ */
+public interface RetainableByteBufferPool
+{
+    /**
+     * Acquire a memory buffer from the pool.
```
sbordet: Nit: `Acquire` -> `Acquires`.

jetty-io/src/test/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPoolTest.java
```diff
@@ -0,0 +1,218 @@
...
+                RetainableByteBuffer buffer = pool.acquire(32768, false);
+                assertThat(buffer, is(notNullValue()));
+                buffer.release();
+            }
+        }
+
+        assertThat(pool.getDirectByteBufferCount(), is(4L));
+        assertThat(pool.getHeapByteBufferCount(), is(1L));
+        assertThat(pool.getDirectMemory(), is(52224L));
+        assertThat(pool.getHeapMemory(), is(32768L));
```
sbordet: Too many magic values. You can allocate variables and perform a sum here, e.g. `capacity1=1024` but then `acquire(capacity1/10, ...)`, etc.

jetty-io/src/test/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPoolTest.java
```diff
@@ -0,0 +1,218 @@
...
+
+        pool.clear();
+
+        assertThat(pool.getDirectByteBufferCount(), is(0L));
+        assertThat(pool.getHeapByteBufferCount(), is(0L));
+        assertThat(pool.getDirectMemory(), is(0L));
+        assertThat(pool.getHeapMemory(), is(0L));
+    }
+}
```
sbordet: I would have expected a test for `retain()`. What if a retained buffer is released? I would keep some safety net code.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -225,36 +228,32 @@ public void onFlushed(long bytes) throws IOException
...
                 LOG.debug("releaseRequestBuffer {}", this);
-            ByteBuffer buffer = _requestBuffer;
-            _requestBuffer = null;
-            _bufferPool.release(buffer);
+            if (_retainableByteBuffer.release() == 0)
+                _retainableByteBuffer = null;
         }
     }
 
     public ByteBuffer getRequestBuffer()
```
sbordet: You broke already some APIs, I would make this `private` or even remove the method and inline its 2 usages.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -302,15 +301,14 @@ else if (filled < 0)
         {
             if (LOG.isDebugEnabled())
                 LOG.debug("{} caught exception {}", this, _channel.getState(), x);
-            BufferUtil.clear(_requestBuffer);
```
sbordet, line 305: I don't think this `clear()` can be removed.
If you don't clear the buffer, then it's leaked because `releaseRequestBuffer()` does not release it if it still has content.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -406,15 +404,14 @@ private boolean upgrade()
         _channel.recycle();
         _parser.reset();
         _generator.reset();
-        if (_contentBufferReferences.get() == 0)
+        if (_retainableByteBuffer != null && _retainableByteBuffer.getReferences() == 1)
```
sbordet: I'd prefer a single null check expression, and inside an `if/else` on the number of references, more readable.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -629,21 +626,13 @@ public String toConnectionString()
...
-            int counter = _contentBufferReferences.decrementAndGet();
-            if (counter == 0)
-                releaseRequestBuffer();
-            // TODO: this should do something (warn? fail?) if _contentBufferReferences goes below 0
-            if (counter < 0)
-            {
-                LOG.warn("Content reference counting went below zero: {}", counter);
-                _contentBufferReferences.incrementAndGet();
-            }
+            _retainableByteBuffer.release();
```
sbordet, line 649: The safety code was removed, but I would keep it.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
```diff
@@ -15,62 +15,71 @@
...
     {
-        while (true)
-        {
-            int r = references.get();
-            if (r == 0)
-                throw new IllegalStateException("released " + this);
-            if (references.compareAndSet(r, r + 1))
-                break;
-        }
+        references.incrementAndGet();
```
sbordet: I would retain the safety code.

sbordet (dismissed):

jetty-io/src/main/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPool.java
```diff
@@ -0,0 +1,390 @@
...
+                    oldestEntry = entry;
+                    oldestEntryBucketIndex = i;
+                }
+            }
+
+            if (oldestEntry != null)
+            {
+                if (LOG.isDebugEnabled())
+                    LOG.debug("removing entry of capacity {} and age {}", oldestEntry.getPooled().capacity(), now - oldestEntry.getPooled().getLastUpdate());
+                oldestEntry.remove();
```
sbordet: What if `remove()` returns `false`? I don't think you can decrement the memory in that case.

jetty-io/src/main/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPool.java
```diff
@@ -0,0 +1,390 @@
...
+        if (LOG.isDebugEnabled())
+            LOG.debug("evicting {} bytes from {} pools", excess, (direct ? "direct" : "heap"));
+        long totalMemoryCleared = 0L;
+        long now = System.nanoTime();
+
+        Pool<RetainableByteBuffer>[] buckets = direct ? _direct : _indirect;
+        @SuppressWarnings("unchecked")
+        Pool<RetainableByteBuffer>.Entry[] oldestEntries = new Pool.Entry[buckets.length];
+
+        for (int i = 0; i < buckets.length; i++)
```
sbordet: Should you not iterate from the last, so the larger buffers get evicted first?

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -379,15 +377,15 @@ else if (filled < 0)
...
 
-        boolean handle = _parser.parseNext(_requestBuffer == null ? BufferUtil.EMPTY_BUFFER : _requestBuffer);
+        boolean handle = _parser.parseNext(_retainableByteBuffer == null ? BufferUtil.EMPTY_BUFFER : _retainableByteBuffer.getBuffer());
 
         if (LOG.isDebugEnabled())
             LOG.debug("{} parsed {} {}", this, handle, _parser);
 
         // recycle buffer ?
-        if (_contentBufferReferences.get() == 0)
+        if (_retainableByteBuffer != null && _retainableByteBuffer.getReferences() == 1)
```
sbordet: Please add a comment about why we test for `== 1`.

jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Connection.java
lorban: It would easily take as much code to allow `RetainableByteBuffer` to be sub-classed in such a way that an instance can delegate its lifecycle management to another instance.

I preferred to keep `RetainableByteBuffer` as focused as possible at the expense of 5 one-liner methods in this class.

jetty-io/src/main/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPool.java
lorban: Agreed. Let me do this.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
lorban: Done.

jetty-io/src/test/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPoolTest.java
lorban: Agreed. I cleaned up this test to get rid of magic constants. I also added a few more assertions that make it more obvious what buffer capacity is expected from a certain allocation.

jetty-io/src/test/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPoolTest.java
lorban: Agreed, I added a test for `retain`.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
lorban: Done.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
lorban: Done.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
lorban: This safety code is now in `RetainableByteBuffer.release()`

jetty-io/src/main/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPool.java
lorban: You're right, fixed that.

jetty-io/src/main/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPool.java
lorban: The whole eviction mechanism was greatly simplified after our discussion. Now a buffer is evicted from every single bucket, so that the iteration order doesn't matter anymore.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
lorban: I replaced all the `== 1` and `> 1` comparisons with a call to a more readable `boolean isRetained()` method and added some javadoc.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
lorban: You're right, but I replaced the call to `releaseRequestBuffer` with an explicit null check and release to make it more clear that this error handling is a bit of a special case.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
lorban: Okay, I could add it back after a little logic change that requires discriminating references increments due to user-code or due to pool code.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
```diff
@@ -15,65 +15,112 @@
...
+    }
+
+    public boolean isDirect()
+    {
+        return buffer.isDirect();
+    }
+
+    /**
+     * Increments the retained counter of this buffer.
+     * It must be done right after creation and after each un-pooling.
```
sbordet: Fix javadoc.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
```diff
@@ -15,65 +15,112 @@
...
 import java.nio.ByteBuffer;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Consumer;
 
 import org.eclipse.jetty.util.BufferUtil;
 import org.eclipse.jetty.util.Retainable;
 
 /**
  * A Retainable ByteBuffer.
```
sbordet: Remove this line, unnecessary.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -225,36 +227,32 @@ public void onFlushed(long bytes) throws IOException
...
-        if (_requestBuffer != null && !_requestBuffer.hasRemaining())
+        if (_retainableByteBuffer != null && !_retainableByteBuffer.hasRemaining())
         {
             if (LOG.isDebugEnabled())
                 LOG.debug("releaseRequestBuffer {}", this);
-            ByteBuffer buffer = _requestBuffer;
-            _requestBuffer = null;
-            _bufferPool.release(buffer);
+            if (_retainableByteBuffer.release())
+                _retainableByteBuffer = null;
```
sbordet, line 235: Can you please add an `else` branch with `throw ISE`?

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -302,15 +300,19 @@ else if (filled < 0)
...
             if (LOG.isDebugEnabled())
                 LOG.debug("{} caught exception {}", this, _channel.getState(), x);
-            BufferUtil.clear(_requestBuffer);
-            releaseRequestBuffer();
+            // Force the release of the _retainableByteBuffer.
+            if (_retainableByteBuffer != null)
+            {
+                while (!_retainableByteBuffer.release());
+                _retainableByteBuffer = null;
+            }
```
sbordet, line 316: Please let's go to `clear()` + `releaseRequestBuffer()` like it was before and it is in other places such as `WebSocketConnection`.
Please make sure `EndPoint.close(x)` is called even if `releaseRequestBuffer()` throws.

sbordet (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
lorban: Ok.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
lorban: Done.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
lorban: Sure.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
lorban: Ok.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -302,15 +302,25 @@ else if (filled < 0)
...
+                try
+                {
+                    releaseRequestBuffer();
+                }
+                catch (Throwable t)
+                {
+                    x.addSuppressed(t);
+                }
+            }
             getEndPoint().close(x);
```
sbordet: I'm nitpicking here, but if we surround this call, then we should surround also all the others -- not sure we want to do that.

How about this:

```suggestion
            try
            {
                if (LOG.isDebugEnabled())
                    LOG.debug("{} caught exception {}", this, _channel.getState(), x);
                if (_retainableByteBuffer != null)
                {
                    _retainableByteBuffer.clear();
                    releaseRequestBuffer();
                }
            }
            finally
            {
                getEndPoint().close(x);
            }
```

sbordet (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
lorban: Okay.

sbordet (dismissed):

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
```diff
@@ -15,65 +15,113 @@
...
+    }
+
+    /**
+     * Increments the retained counter of this buffer.
+     */
     @Override
     public void retain()
     {
-        while (true)
+        if (references.getAndIncrement() == 0)
```
gregw: I think the following would avoid temporarily incrementing to 1, which other threads might see.
```suggestion
        if (references.getAndUpdate(c -> c == 0 ? 0 : c + 1) == 0)
```

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
```diff
@@ -15,65 +15,113 @@
...
+
+    /**
+     * Increments the retained counter of this buffer. It must be done internally by
+     * the pool right after creation and after each un-pooling.
+     * The reason why this method exists on top of {@link #retain()} is to be able to
+     * have some safety checks that must know why the ref counter is being incremented.
+     */
+    void acquire()
+    {
+        if (references.getAndIncrement() != 0)
```
gregw: Avoid modification if it is not already 0.  Not so important here, but see below
```suggestion
        if (references.getAndUpdate(c -> c == 0 ? 1 : c) != 0)
```

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
```diff
@@ -15,65 +15,113 @@
...
+     * Decrements the retained counter of this buffer.
+     * @return true if the buffer was re-pooled, false otherwise.
+     */
+    public boolean release()
     {
         int ref = references.decrementAndGet();
-        if (ref == 0)
-            pool.release(buffer);
-        else if (ref < 0)
+        if (ref < 0)
```
gregw: ditto - use an update to avoid changing value when you should not.

jetty-io/src/main/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPool.java
```diff
@@ -0,0 +1,314 @@
...
+import org.eclipse.jetty.util.BufferUtil;
+import org.eclipse.jetty.util.Pool;
+import org.eclipse.jetty.util.annotation.ManagedAttribute;
+import org.eclipse.jetty.util.annotation.ManagedObject;
+import org.eclipse.jetty.util.annotation.ManagedOperation;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@ManagedObject
+public class DefaultRetainableByteBufferPool implements RetainableByteBufferPool
```
gregw: previously we have named our pools for their implementation and not for their usage.  This really is an `ArrayRetainableByteBufferPool` and if in future we want to switch to a Mapped pool, then the name doesn't make sense.

```suggestion
public class ArrayRetainableByteBufferPool implements RetainableByteBufferPool
```

jetty-io/src/main/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPool.java
```diff
@@ -0,0 +1,314 @@
...
+                {
+                    BufferUtil.clear(byteBuffer);
+                    reservedEntry.release();
+                });
+                reservedEntry.enable(buffer, true);
+                if (direct)
+                    _currentDirectMemory.addAndGet(buffer.capacity());
+                else
+                    _currentHeapMemory.addAndGet(buffer.capacity());
+                releaseExcessMemory(direct);
```
gregw: The release should be done before allocating the new buffer, otherwise for a short period of time we are exceeding the memory capacity.  If multiple threads acquire at the same time, this could be a significant excess.

I guess the problem with releasing prior to allocating is how to release enough space if multiple threads are acquiring.... I think we need a way to atomically reduce the capacity for a period.

Hmmm or am I worried about nothing, because if the pool is full, we still allocate a buffer.... ????

jetty-io/src/main/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPool.java
```diff
@@ -0,0 +1,314 @@
...
+    private void evict(boolean direct, long excess)
+    {
+        if (LOG.isDebugEnabled())
+            LOG.debug("evicting {} bytes from {} pools", excess, (direct ? "direct" : "heap"));
+        long now = System.nanoTime();
+        long totalClearedCapacity = 0L;
+
+        Pool<RetainableByteBuffer>[] buckets = direct ? _direct : _indirect;
+
+        while (totalClearedCapacity < excess)
```
gregw: If multiple threads acquire on a full pool, will multiple threads end up iterating here?   Is there a way to ensure only a single thread does evict? or is it not worth worrying about?

gregw (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPool.java
lorban: Ok, I'm happy with the new name.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
lorban: Good point, I'll do this change.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
lorban: Good point, I'll do this change.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
lorban: Good point, I'll do this change.

jetty-io/src/main/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPool.java
lorban: I think you're worrying too much. Strictly respecting the bound would require a global lock that would completely kill concurrency; so barring this option we can choose between never overshooting but potentially refusing to serve an available buffer or potentially overshooting.

In this case, the overshoot could be at most 1 buffer per concurrent thread, which doesn't sound unreasonable as a healthy server should always have some spare memory that could absorb a spike. Plus, as you noted, we allocate when the pool is full so we already have the potential to overshoot, even if the pool strictly respected its bounds.

jetty-io/src/main/java/org/eclipse/jetty/io/DefaultRetainableByteBufferPool.java
lorban: We discussed this with @sbordet and out conclusion was that it's preferable to have concurrent threads iterating rather than leaving that job to a single one.

The reasoning was that if we end up in a situation where a few threads have to evict concurrently every other day, the impact is going to be negligible no matter how inefficient the eviction may be. But if we are in a situation where eviction happens frequently, this is probably a sign that the pool is undersized and the best thing to do is probably to generate some form of back pressure by making all threads "waste" cycles in the eviction code.

sbordet (dismissed):

gregw (approved):
This looks good.
However, I'd still like to see an alternate version (or perhaps an extension) where the bucket sizes are non-linear.  Instead of buckets for 1k, 2k, 3k, 4k....64k  I think log buckets could be lot more efficient at avoiding lots of unused buckets.  So the bucket sizes should be 1k, 2k, 4k, 8k,16k,32k,64k.

gregw (changes requested):
Consider change from #6513

lorban:
This PR still needs to be rebased and have its history rewritten.

gregw (dismissed):
My PR isn't yet ready, so let's not hold this one up.

sbordet (approved):

gregw:
don't wait for my PR to merge

joakime:
@lorban I think you should merge this soon.
Both @gregw and @sbordet have approved it.

PR #6334, opened 2021-05-28
Fixes #6323 - HttpClient requests with redirects gets stuck/never cal…

### LINKED ISSUE #6323: HttpClient gets stuck/never calls onComplete() when multiple requests with timeouts are sent
**Jetty version**
v9.4.41
v11.0.3 (does not get stuck, but exhibits unexpected behavior, more on that below)

11.0.7+8-LTS

**OS type/version**
OSX 10.15.4

**Description**
I'm using HttpClient to send multiple HEAD requests with multiple threads with followRedirects=true and a timeout of 5 seconds for each request.  I've observed that with sufficient time between requests (~200ms or more), some of the requests will never timeout and their respective `onComplete` listeners never fire.  This causes our application to wait indefinitely on `FuturePromise.get()` calls for these requests to return.  

This behavior did not happen in version 9.4.11.  We've recently upgraded to 9.4.11+ (first tried 9.4.40 and then 9.4.41) in order to get official TLS 1.3 support.  

In version 11.0.3, the `onComplete` listeners fire, but there seems to be a 20-25 second delay after the requests get "stuck" before the on complete listeners fire.  Obviously this is better than the requests never firing the on complete listeners at all, but still problematic.

One thing I've noticed is that the test URL goes through a series of redirects.  There is a second test URL at the top of the test classes (attached below) that does not run into the same issue but it doesn't go through any redirects.

With debug logging enabled in Jetty 9.4.41 for the `TimeoutCompleteListener`, I noticed that the requests get "stuck" when no call to `TimeoutCompleteListener.schedule()` is made for some of them.  Example, I'd test 20 requests and see some number of `TimeoutCompleteListener.schedule()` calls that was less than 20 even though all of the requests should have been scheduled to time out.  I'm not super familiar with the Jetty codebase, so this might not be on the right trail, but hope that helps.

Attached are two test classes for reproducing the behaviors in v9.4.41 and v11.0.3.
[test-classes.zip](https://github.com/eclipse/jetty.project/files/6541858/test-classes.zip)

Thanks!

sbordet:
…ls onComplete()

* Reworked the total timeout handling.
* Now a CyclicTimeouts handles the exchanges in each HttpDestination,
and a CyclicTimeouts handles the exchanges in each HttpConnection
(rather than in HttpChannel).
* Now adjusting the total timeout for copied requests generated by
redirects and authentication.

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/AuthenticationProtocolHandler.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/AuthenticationProtocolHandler.java
@@ -231,7 +231,7 @@ public abstract class AuthenticationProtocolHandler implements ProtocolHandler
                     }
                     else
                     {
-                        TimeoutException failure = new TimeoutException("Total timeout " + request.getTimeout() + " ms elapsed");
+                        TimeoutException failure = new TimeoutException("Total timeout " + request.getConversation().getTimeout() + " ms elapsed");
                         forwardFailureComplete(request, failure, response, failure);
                         return;
                     }
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/AuthenticationProtocolHandler.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/AuthenticationProtocolHandler.java
@@ -24,6 +24,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -218,10 +219,23 @@ public abstract class AuthenticationProtocolHandler implements ProtocolHandler
                 }
                 Request newRequest = client.copyRequest(request, requestURI);
 
-                // Adjust the timeout.
+                // Adjust the timeout of the new request, taking into account the
+                // timeout of the previous request and the time already elapsed.
                 long timeoutAt = request.getTimeoutAt();
                 if (timeoutAt < Long.MAX_VALUE)
-                    newRequest.timeout(timeoutAt - System.nanoTime(), TimeUnit.MILLISECONDS);
+                {
+                    long newTimeout = timeoutAt - System.nanoTime();
+                    if (newTimeout > 0)
+                    {
+                        newRequest.timeout(newTimeout, TimeUnit.NANOSECONDS);
+                    }
+                    else
+                    {
+                        TimeoutException failure = new TimeoutException("Total timeout " + request.getTimeout() + " ms elapsed");
+                        forwardFailureComplete(request, failure, response, failure);
+                        return;
+                    }
+                }
 
                 if (path != null)
                     newRequest.path(path);
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/AuthenticationProtocolHandler.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/AuthenticationProtocolHandler.java
@@ -217,8 +217,12 @@ public abstract class AuthenticationProtocolHandler implements ProtocolHandler
                     path = request.getPath();
                 }
                 Request newRequest = client.copyRequest(request, requestURI);
-                // Disable the timeout so that only the one from the initial request applies.
-                newRequest.timeout(0, TimeUnit.MILLISECONDS);
+
+                // Adjust the timeout.
+                long timeoutAt = request.getTimeoutAt();
+                if (timeoutAt < Long.MAX_VALUE)
+                    newRequest.timeout(timeoutAt - System.nanoTime(), TimeUnit.MILLISECONDS);
+
                 if (path != null)
                     newRequest.path(path);
 
--- a/jetty-fcgi/fcgi-client/src/main/java/org/eclipse/jetty/fcgi/client/http/HttpConnectionOverFCGI.java
+++ b/jetty-fcgi/fcgi-client/src/main/java/org/eclipse/jetty/fcgi/client/http/HttpConnectionOverFCGI.java
@@ -519,4 +519,32 @@ public class HttpConnectionOverFCGI extends AbstractConnection implements Connec
                 LOG.debug("Channel not found for request {}", request);
         }
     }
+
+    private static final class IteratorWrapper<T> implements Iterator<T>
+    {
+        private final Iterator<? extends T> iterator;
+
+        private IteratorWrapper(Iterator<? extends T> iterator)
+        {
+            this.iterator = iterator;
+        }
+
+        @Override
+        public boolean hasNext()
+        {
+            return iterator.hasNext();
+        }
+
+        @Override
+        public T next()
+        {
+            return iterator.next();
+        }
+
+        @Override
+        public void remove()
+        {
+            iterator.remove();
+        }
+    }
 }
--- a/jetty-fcgi/fcgi-client/src/main/java/org/eclipse/jetty/fcgi/client/http/HttpConnectionOverFCGI.java
+++ b/jetty-fcgi/fcgi-client/src/main/java/org/eclipse/jetty/fcgi/client/http/HttpConnectionOverFCGI.java
@@ -21,6 +21,7 @@ package org.eclipse.jetty.fcgi.client.http;
 import java.io.EOFException;
 import java.nio.ByteBuffer;
 import java.nio.channels.AsynchronousCloseException;
+import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.Map;
 import java.util.Queue;
@@ -374,6 +375,12 @@ public class HttpConnectionOverFCGI extends AbstractConnection implements Connec
         }
 
         @Override
+        protected Iterator<HttpChannel> getHttpChannels()
+        {
+            return new IteratorWrapper<>(activeChannels.values().iterator());
+        }
+
+        @Override
         protected SendFailure send(HttpExchange exchange)
         {
             Request request = exchange.getRequest();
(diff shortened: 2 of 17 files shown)


jetty-fcgi/fcgi-client/src/main/java/org/eclipse/jetty/fcgi/client/http/HttpConnectionOverFCGI.java
```diff
@@ -373,6 +374,12 @@ private Delegate(HttpDestination destination)
             super(destination);
         }
 
+        @Override
+        protected Iterator<HttpChannel> getHttpChannels()
+        {
+            return new IteratorWrapper<>(activeChannels.values().iterator());
```
joakime, line 380: Why do you need a IteratorWrapper here?
Seems like you could just return that iterator and be done with it.

sbordet: No, because `activeChannels.values().iterator()` has a type that is a subclass of `HttpChannel`, so the code won't compile.

This is a compile error because the caller obtains a "container" of types `HttpChannel` that actually contains `HttpChannelOverFCGI`, but the caller would be entitled to add to this "container" a different subclass, say `HttpChannelOverHTTP2`, which would break FCGI.

The fact that `Iterator` is a read-only data structure is unfortunately not taken into consideration -- I was surprised too, but that's just how generics work in Java, so the wrapper is required.

jetty-client/src/main/java/org/eclipse/jetty/client/AuthenticationProtocolHandler.java
```diff
@@ -217,8 +217,12 @@ public void onComplete(Result result)
                     path = request.getPath();
                 }
                 Request newRequest = client.copyRequest(request, requestURI);
-                // Disable the timeout so that only the one from the initial request applies.
-                newRequest.timeout(0, TimeUnit.MILLISECONDS);
+
+                // Adjust the timeout.
```
gregw: a bit more of a comment saying how/why the timeout is adjusted would be good

jetty-client/src/main/java/org/eclipse/jetty/client/AuthenticationProtocolHandler.java
```diff
@@ -217,8 +217,12 @@ public void onComplete(Result result)
...
+                long timeoutAt = request.getTimeoutAt();
+                if (timeoutAt < Long.MAX_VALUE)
+                    newRequest.timeout(timeoutAt - System.nanoTime(), TimeUnit.MILLISECONDS);
```
gregw: what happens if the timeout has expired between the time the previous request completed and here?  ie can we pass a negative value without throwing?

jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeouts.java
```diff
@@ -0,0 +1,167 @@
...
+        // When the timeout expires, scan the entities for the next
+        // earliest entity that may expire, and reschedule a new timeout.
+        long prevEarliest = earliestTimeout.getAndUpdate(t -> Math.min(t, expiresAt));
+        if (expiresAt < prevEarliest)
+        {
+            // A new entity expires earlier than previous entities, schedule it.
+            long delay = Math.max(0, expiresAt - System.nanoTime());
+            if (LOG.isDebugEnabled())
+                LOG.debug("{} scheduling timeout in {} ms", this, TimeUnit.NANOSECONDS.toMillis(delay));
+            cyclicTimeout.schedule(delay, TimeUnit.NANOSECONDS);
```
gregw: I think we have a race here.   Imagine if we are expiring at 100 and a thread calls schedule(50), we see this as `<100`  so we enter this block and start calculating delay... but we are preempted.  Then another t hread calls schedule(25), we see that as `<50` so we enter this block, calculate delay and schedule the timeout for 25.  The prempted thread then wakes up and schedules the timeout for 50.

jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeouts.java
```diff
@@ -0,0 +1,167 @@
...
+     * <p>Invoked during the iteration when the given entity is expired.</p>
+     *
+     * @param expirable the entity that is expired
+     * @return whether the entity should be removed from the iterator via {@link Iterator#remove()}
+     */
+    protected abstract boolean onExpired(T expirable);
+
+    private void onTimeoutExpired()
+    {
+        if (LOG.isDebugEnabled())
```
gregw, line 80: Do we need to protect against reentrancy?     If `onTimeoutExpired` is called, it may take some time to do the iteration.  Meanwhile another thread may call `schedule(short_time)` and that could actually expire, resulting in another call entering `onTimeoutExpired`.  See comment below...

jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeouts.java
```diff
@@ -0,0 +1,167 @@
...
+        if (expiresAt < Long.MAX_VALUE)
+            schedule(expiresAt);
+    }
+
+    private void schedule(long expiresAt)
+    {
+        // Schedule a timeout for the earliest entity that may expire.
+        // When the timeout expires, scan the entities for the next
+        // earliest entity that may expire, and reschedule a new timeout.
+        long prevEarliest = earliestTimeout.getAndUpdate(t -> Math.min(t, expiresAt));
```
gregw, line 141: Should we check prevEarliest for a flag value that indicates `onTimeoutExpired()` is currently running and thus we should not do anything else here?

gregw (changes requested):
I have a few concerns.... which probably affect jetty-10 as well.  We may need to hangout.

jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeouts.java
```diff
@@ -0,0 +1,167 @@
...
+ *   <li>may expire at a future time; the iteration records the earliest expiration time among
+ *   all non-expired entities</li>
+ * </ul>
+ * <p>When the iteration is complete, this instance is re-scheduled with the earliest expiration time
+ * calculated during the iteration.</p>
+ *
+ * @param <T> the {@link Expirable} entity type
+ * @see CyclicTimeout
+ */
+public abstract class CyclicTimeouts<T extends CyclicTimeouts.Expirable> implements Destroyable
```
gregw, line 50: We should have some tests for this class

jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeouts.java
gregw: I think the following should resolve this race... but it's a bit seat-of-the-pants
```java
    private void schedule(long expiresAt)
    {
        // Schedule a timeout for the earliest entity that may expire.
        // When the timeout expires, scan the entities for the next
        // earliest entity that may expire, and reschedule a new timeout.
        long prevEarliest = earliestTimeout.getAndUpdate(t -> Math.min(t, expiresAt));

        long expires = expiresAt;
        while (expires < prevEarliest)
        {
            // A new entity expires earlier than previous entities, schedule it.
            long delay = Math.max(0, expires - System.nanoTime());
            if (LOG.isDebugEnabled())
                LOG.debug("{} scheduling timeout in {} ms", this, TimeUnit.NANOSECONDS.toMillis(delay));
            cyclicTimeout.schedule(delay, TimeUnit.NANOSECONDS);

            // If we lost a race and overwrote a subsequent schedule, then the earliestTimeout value will have reduced,
            // in which case we will loop and set it again ourselves.
            prevEarliest = expires;
            expires = earliestTimeout.get();
        }
    }
```

gregw (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/CyclicTimeouts.java
gregw: I think with the race "fix" I propose below, this reentrancy is no longer a problem.  The race "fix" will always schedule an earlier timeout so I think we are good... but need to do some more thought experiments to be sure.

jetty-client/src/main/java/org/eclipse/jetty/client/AuthenticationProtocolHandler.java
sbordet: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/AuthenticationProtocolHandler.java
sbordet: Good catch, the timeout must be positive. Fixed.

jetty-fcgi/fcgi-client/src/main/java/org/eclipse/jetty/fcgi/client/http/HttpConnectionOverFCGI.java
joakime: Shame that the method you are overriding couldn't be `protected Iterator<? extends HttpChannel> getHttpChannels()` instead?

sbordet: @joakime returning wildcard types is a no-no in every book.

gregw (approved):
It would be good to get this tested as widely as possible before staging a branch.
Also the fix and tests for CyclicTimeouts needs to be forward ported to 10

PR #6335, opened 2021-05-30
Fixes #5931 - SslConnection should implement getBytesIn()/getBytesOut().

### LINKED ISSUE #5931: SslConnection should implement getBytesIn()/getBytesOut()
**Jetty version**
9.4.x

**Description**
`SslConnection` does not implement `getBytesIn()/getBytesOut()` but it should, so that load test reports are accurate in showing how many bytes traveled through the network.

Using `IncludeExcludeConnectionStatistics` would allow to select what connection needs to be tracked to avoid double-reporting the bytes.

sbordet:

### DIFF
--- a/jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTLSTest.java
+++ b/jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTLSTest.java
@@ -48,6 +48,7 @@ import org.eclipse.jetty.http.HttpScheme;
 import org.eclipse.jetty.http.HttpStatus;
 import org.eclipse.jetty.io.ByteBufferPool;
 import org.eclipse.jetty.io.ClientConnectionFactory;
+import org.eclipse.jetty.io.Connection;
 import org.eclipse.jetty.io.ConnectionStatistics;
 import org.eclipse.jetty.io.EndPoint;
 import org.eclipse.jetty.io.ssl.SslClientConnectionFactory;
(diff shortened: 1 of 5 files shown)


jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
```diff
@@ -104,7 +105,10 @@
         WAIT_FOR_FILL // Waiting for a fill to happen
     }
 
+    private final AtomicReference<HandshakeState> _handshake = new AtomicReference<>(HandshakeState.INITIAL);
     private final List<SslHandshakeListener> handshakeListeners = new ArrayList<>();
+    private final AtomicLong _bytesIn = new AtomicLong();
```
gregw, line 110: We have been changing to LongAdder in other stats classes, so perhaps we should use here as well?

gregw (dismissed):
Just a niggle regarding the type of the counter, otherwise OK.

Although part of me really hates counting stats all the time, but I think the complexity of optionally counting them is not worth it.

jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
sbordet: I think `LongAdder` is more suited when it's contended, but here is not so I went for `AtomicLong`.

lachlan-roberts:
@sbordet is there any issue that existing usages of `ConnectionStatistics` will be now doubly counting bytes from both `HttpConnection` and `SslConnection`. Someone already using `ConnectionStatistics` to monitor their connections won't be expecting this sudden increase.

gregw:
@lachlan-roberts will the double counting happen?   Doesn't `ConnectionStatistics` just see the outermost connection?

jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTLSTest.java
```diff
@@ -1003,4 +1006,49 @@ public void testForcedNonDomainSNI() throws Exception
             assertEquals(HttpStatus.OK_200, response3.getStatus());
         }
     }
+
+    @Test
+    public void testBytesInBytesOut() throws Exception
+    {
+        // Two connections will be closed: SslConnection and HttpConnection.
+        // Two on the server, two on the client.
```
lachlan-roberts, line 1014: @gregw if you look at the test here it seems to indicate that both `SslConnection` and `HttpConnection` are recorded by `ConnectionStatistics`.

`ConnectionStatistics` works as a `Connection.Listener` so it would apply to `SslConnection` as well.

gregw: Hmmm is this different on the Server-side?  As I don't see how a Connection.Listener would see any connections wrapped by SslConnection?

lachlan-roberts: The `Connection.Listener` is added as a bean to all connectors on the server, and any `Connection.Listener` beans on the Connector are added to each new connection that is created. So the listener will automatically be added to the `SslConnection`.

gregw: Yeah, so the listener is on SslConnection. Is it also inherited by the Wrapped HttpConnection? If so, then i agree the double counting is wrong.

gregw:
@sbordet you forgot to request re-review! I've been waiting for y our checkin!

jetty-io/src/main/java/org/eclipse/jetty/io/ConnectionStatistics.java
```diff
@@ -78,159 +65,312 @@ public void onOpened(Connection connection)
     {
         if (!isStarted())
             return;
-
-        _connections.increment();
+        _stats.incrementCount();
+        _statsMap.computeIfAbsent(connection.getClass().getName(), Stats::new).incrementCount();
```
gregw, line 84: perhaps delegate the computeIfAbsent to a protected method, so that if somebody really wanted to pick and choose which types get counted then they could override that method and only return a Stats if they wanted that specific connection counted and it might be on a different criteria than className.

jetty-io/src/main/java/org/eclipse/jetty/io/ConnectionStatistics.java
```diff
@@ -78,159 +65,312 @@ public void onOpened(Connection connection)
...
     }
 
     @Override
     public void onClosed(Connection connection)
     {
         if (!isStarted())
             return;
+        onClosed(_stats, connection);
+        Stats stats = _statsMap.get(connection.getClass().getName());
```
gregw, line 103: Ditto. Put this get call into a protected method so selection can be done on criteria other than class name.

gregw (dismissed):
Other than a niggle for extensibility, this LGTM

gregw (approved):

PR #6338, opened 2021-06-01
Add comment on why the test ServerWithJMXTest is disabled

olamy:


demos/embedded/src/test/java/org/eclipse/jetty/demos/ServerWithJMXTest.java
```diff
@@ -27,7 +27,7 @@
 
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-@Disabled
+@Disabled(value = "ServerWithJMX try to start/restart a registry on port 1999 this can cause failure as the port is already used and not freed quickly enough by the OS")
```
janbartel:
```suggestion
@Disabled(value = "ServerWithJMX uses registry on fixed port 1999 which can be already in-use")
```

janbartel (dismissed):
A little more succinct message as a suggestion for you.

janbartel (approved):

PR #6342, opened 2021-06-01
Explain EatWhatYouKill naming

gregw:
Some concern has been raised at the EatWhatYouKill strategy name (OK @sbordet you get I told you so rights).
This PR updates the javadoc to be clear about the reason for using that name in that the strategy is named after 
a [hunting proverb](https://www.urbandictionary.com/define.php?term=eat%20what%20you%20kill) in the sense 
that one should only kill to eat. The use of this phrase is not an endorsement of hunting nor killing of wildlife for 
food or sport.


sbordet:
@gregw at the end it's a _dynamic_ `ExecutionStrategy` that changes its mode based on `TryExecutor.tryExecute()`.

So perhaps it's time to call it `DynamicExecutionStrategy` and deprecate `EatWhatYouKill`?

sbordet (approved):
Additional text is good.
However, it's probably time to change the name as I specify in the comments.

gregw:
@sbordet I think you are right - the EWYK is now only one of several strategies that can be applied, and then when compared with PC & PEC it is clearer to describe it as EPC.

I'll do the rename, but only in 10 and leave the EWYK as a deprecated superclass of it.

sbordet:
@gregw EWYK should be a deprecated _subclass_ of `DynamicExecutionStrategy`.

PR #6348, opened 2021-06-02
Issue #6329 Fix xml config problems

### LINKED ISSUE #6329: Regression on graceful shutdown default in Jetty 10
The method Servlet.destroy() is not being called when Jetty is stopped when using the default configuration.

It is called if this line is added to the configuration:

jetty.server.stopAtShutdown=true

But it seems to me that the default behavior should be the opposite, because calling Servlet.destroy() is what one expects from the servlet container, and not calling it looks like an optional optimization.

### LINKED ISSUE #6392: Review accidental xml config changes
**jetty-10/jetty-11**

This supersedes issue #6329. Analysing #6329 (jetty 10/11 does not stop on shutdown by default when running in the distro) revealed that commit https://github.com/eclipse/jetty.project/commit/343cf732b9a5aacd4c84fcb937a4b1624a1a0afd had inadvertently changed a number of default values in jetty xml config files. We need to review the config changes, and where necessary revert/fix incorrect changes.

janbartel:
Closes #6392 

The original problem that was reported in #6329 was the configuration for `Server.setStopAtShutdown` had been mistakenly  changed from a `default` of `true` to `false`, prompting a full review of the changes from the relevant commit (https://github.com/eclipse/jetty.project/commit/343cf732b9a5aacd4c84fcb937a4b1624a1a0afd). This PR puts back some other changes that look like mistakes, but can reviewers @sbordet @gregw and @joakime check carefully if these changes are indeed needed. I've simplified the session xml configs to put the defaults into code rather than xml where possible.

### DIFF
--- a/jetty-gcloud/jetty-gcloud-session-manager/src/main/config-template/etc/sessions/gcloud/session-store.xml
+++ b/jetty-gcloud/jetty-gcloud-session-manager/src/main/config-template/etc/sessions/gcloud/session-store.xml
@@ -12,9 +12,9 @@
     <New id="sessionDataStoreFactory" class="org.eclipse.jetty.gcloud.session.GCloudSessionDataStoreFactory">
        <Set name="gracePeriodSec" property="jetty.session.gracePeriod.seconds"/>
        <Set name="savePeriodSec" property="jetty.session.savePeriod.seconds"/>
-       <Set name="maxRetries"><Property name="jetty.session.gcloud.maxRetries" default="5"/></Set>
-       <Set name="backoffMs"><Property name="jetty.session.gcloud.backoffMs" default="1000"/></Set>
-       <Set name="namespace"><Property name="jetty.session.gcloud.namespace" default=""/></Set>  
+       <Set name="maxRetries" property="jetty.session.gcloud.maxRetries"/>
+       <Set name="backoffMs" property="jetty.session.gcloud.backoffMs"/>
+       <Set name="namespace" property="jetty.session.gcloud.namespace"/>  
        <Set name="entityDataModel">
           <New class="org.eclipse.jetty.gcloud.session.GCloudSessionDataStore$EntityDataModel">
               <Set name="kind">
--- a/jetty-gcloud/jetty-gcloud-session-manager/src/main/config-template/etc/sessions/gcloud/session-store.xml
+++ b/jetty-gcloud/jetty-gcloud-session-manager/src/main/config-template/etc/sessions/gcloud/session-store.xml
@@ -17,45 +17,19 @@
        <Set name="namespace" property="jetty.session.gcloud.namespace"/>  
        <Set name="entityDataModel">
           <New class="org.eclipse.jetty.gcloud.session.GCloudSessionDataStore$EntityDataModel">
-              <Set name="kind">
-                <Property name="jetty.session.gcloud.model.kind" default="GCloudSession" />
-              </Set>
-              <Set name="id">
-                <Property name="jetty.session.gcloud.model.id" default="id" />
-              </Set>
-             <Set name="contextPath">
-                <Property name="jetty.session.gcloud.model.contextPath" default="contextPath" />
-              </Set>
-              <Set name="vhost">
-                <Property name="jetty.session.gcloud.model.vhost" default="vhost" />
-              </Set>
-              <Set name="accessed">
-                <Property name="jetty.session.gcloud.model.accessed" default="accessed" />
-              </Set>
-              <Set name="lastAccessed">
-                <Property name="jetty.session.gcloud.model.lastAccessed" default="lastAccessed" />
-              </Set>
-              <Set name="createTime">
-                <Property name="jetty.session.gcloud.model.createTime" default="createTime" />
-              </Set>
-              <Set name="cookieSetTime">
-                <Property name="jetty.session.gcloud.model.cookieSetTime" default="cookieSetTime" />
-              </Set>
-              <Set name="lastNode">
-                <Property name="jetty.session.gcloud.model.lastNode" default="lastNode" />
-              </Set>
-              <Set name="lastSaved">
-                <Property name="jetty.session.gcloud.model.lastSaved" default="lastSaved" />
-              </Set>
-              <Set name="expiry">
-                <Property name="jetty.session.gcloud.model.expiry" default="expiry" />
-              </Set>
-              <Set name="maxInactive">
-                <Property name="jetty.session.gcloud.model.maxInactive" default="maxInactive" />
-              </Set>
-              <Set name="attributes">
-                <Property name="jetty.session.gcloud.model.attributes" default="attributes" />
-              </Set>
+              <Set name="kind" property="jetty.session.gcloud.model.kind"/>
+              <Set name="id" property="jetty.session.gcloud.model.id"/>
+              <Set name="contextPath" property="jetty.session.gcloud.model.contextPath"/>
+              <Set name="vhost" property="jetty.session.gcloud.model.vhost"/>
+              <Set name="accessed" property="jetty.session.gcloud.model.accessed"/>
+              <Set name="lastAccessed" property="jetty.session.gcloud.model.lastAccessed"/>
+              <Set name="createTime" property="jetty.session.gcloud.model.createTime"/>
+              <Set name="cookieSetTime" property="jetty.session.gcloud.model.cookieSetTime"/>
+              <Set name="lastNode" property="jetty.session.gcloud.model.lastNode"/>
+              <Set name="lastSaved" property="jetty.session.gcloud.model.lastSaved"/>
+              <Set name="expiry" property="jetty.session.gcloud.model.expiry"/>
+              <Set name="maxInactive" property="jetty.session.gcloud.model.maxInactive"/>
+              <Set name="attributes" property="jetty.session.gcloud.model.attributes"/>
           </New>
        </Set>
     </New>
--- a/jetty-gcloud/jetty-gcloud-session-manager/src/main/config-template/etc/sessions/gcloud/session-store.xml
+++ b/jetty-gcloud/jetty-gcloud-session-manager/src/main/config-template/etc/sessions/gcloud/session-store.xml
@@ -10,11 +10,11 @@
   <Call name="addBean">
    <Arg>
     <New id="sessionDataStoreFactory" class="org.eclipse.jetty.gcloud.session.GCloudSessionDataStoreFactory">
-       <Set name="gracePeriodSec"><Property name="jetty.session.gracePeriod.seconds" default="3600" /></Set>
+       <Set name="gracePeriodSec" property="jetty.session.gracePeriod.seconds"/>
        <Set name="savePeriodSec" property="jetty.session.savePeriod.seconds"/>
        <Set name="maxRetries"><Property name="jetty.session.gcloud.maxRetries" default="5"/></Set>
-       <Set name="backoffMs" property="jetty.session.gcloud.backoffMs"/>
-       <Set name="namespace" property="jetty.session.gcloud.namespace"/>
+       <Set name="backoffMs"><Property name="jetty.session.gcloud.backoffMs" default="1000"/></Set>
+       <Set name="namespace"><Property name="jetty.session.gcloud.namespace" default=""/></Set>  
        <Set name="entityDataModel">
           <New class="org.eclipse.jetty.gcloud.session.GCloudSessionDataStore$EntityDataModel">
               <Set name="kind">
@@ -43,6 +43,9 @@
               </Set>
               <Set name="lastNode">
                 <Property name="jetty.session.gcloud.model.lastNode" default="lastNode" />
+              </Set>
+              <Set name="lastSaved">
+                <Property name="jetty.session.gcloud.model.lastSaved" default="lastSaved" />
               </Set>
               <Set name="expiry">
                 <Property name="jetty.session.gcloud.model.expiry" default="expiry" />
--- a/jetty-gcloud/jetty-gcloud-session-manager/src/main/java/org/eclipse/jetty/gcloud/session/GCloudSessionDataStore.java
+++ b/jetty-gcloud/jetty-gcloud-session-manager/src/main/java/org/eclipse/jetty/gcloud/session/GCloudSessionDataStore.java
@@ -55,6 +55,7 @@ public class GCloudSessionDataStore extends AbstractSessionDataStore
     public static final int DEFAULT_MAX_QUERY_RESULTS = 100;
     public static final int DEFAULT_MAX_RETRIES = 5;
     public static final int DEFAULT_BACKOFF_MS = 1000;
+    public static final String DEFAULT_NAMESPACE = "";
 
     protected Datastore _datastore;
     protected KeyFactory _keyFactory;
@@ -65,7 +66,7 @@ public class GCloudSessionDataStore extends AbstractSessionDataStore
     protected boolean _indexesPresent = false;
     protected EntityDataModel _model;
     protected boolean _modelProvided;
-    private String _namespace;
+    private String _namespace = DEFAULT_NAMESPACE;
 
     /**
      * EntityDataModel
@@ -431,7 +432,7 @@ public class GCloudSessionDataStore extends AbstractSessionDataStore
         _namespace = namespace;
     }
 
-    @ManagedAttribute(value = "gclound namespace", readonly = true)
+    @ManagedAttribute(value = "gcloud namespace", readonly = true)
     public String getNamespace()
     {
         return _namespace;
--- a/jetty-gcloud/jetty-gcloud-session-manager/src/main/java/org/eclipse/jetty/gcloud/session/GCloudSessionDataStoreFactory.java
+++ b/jetty-gcloud/jetty-gcloud-session-manager/src/main/java/org/eclipse/jetty/gcloud/session/GCloudSessionDataStoreFactory.java
@@ -22,9 +22,9 @@ import org.eclipse.jetty.server.session.SessionHandler;
  */
 public class GCloudSessionDataStoreFactory extends AbstractSessionDataStoreFactory
 {
-    private String _namespace;
-    private int _maxRetries;
-    private int _backoffMs;
+    private String _namespace = GCloudSessionDataStore.DEFAULT_NAMESPACE;
+    private int _maxRetries = GCloudSessionDataStore.DEFAULT_MAX_RETRIES;
+    private int _backoffMs = GCloudSessionDataStore.DEFAULT_BACKOFF_MS;
     private GCloudSessionDataStore.EntityDataModel _model;
 
     public GCloudSessionDataStore.EntityDataModel getEntityDataModel()
@@ -80,8 +80,9 @@ public class GCloudSessionDataStoreFactory extends AbstractSessionDataStoreFacto
         ds.setBackoffMs(getBackoffMs());
         ds.setMaxRetries(getMaxRetries());
         ds.setGracePeriodSec(getGracePeriodSec());
-        ds.setNamespace(_namespace);
+        ds.setNamespace(getNamespace());
         ds.setSavePeriodSec(getSavePeriodSec());
+        ds.setEntityDataModel(getEntityDataModel());
         return ds;
     }
 }
--- a/jetty-hazelcast/src/main/config/etc/sessions/hazelcast/remote.xml
+++ b/jetty-hazelcast/src/main/config/etc/sessions/hazelcast/remote.xml
@@ -10,21 +10,11 @@
   <Call name="addBean">
     <Arg>
       <New id="sessionDataStoreFactory" class="org.eclipse.jetty.hazelcast.session.HazelcastSessionDataStoreFactory">
-        <Set name="mapName">
-          <Property name="jetty.session.hazelcast.mapName" default="jetty-distributed-session-map" />
-        </Set>
-        <Set name="hazelcastInstanceName">
-          <Property name="jetty.session.hazelcast.hazelcastInstanceName" default="JETTY_DISTRIBUTED_SESSION_INSTANCE" />
-        </Set>
-        <Set name="useQueries">
-          <Property name="jetty.session.hazelcast.useQueries" default="false" />
-        </Set>
-        <Set name="gracePeriodSec">
-          <Property name="jetty.session.gracePeriod.seconds" default="3600" />
-        </Set>
-        <Set name="savePeriodSec">
-          <Property name="jetty.session.savePeriod.seconds" default="0" />
-        </Set>
+        <Set name="mapName" property="jetty.session.hazelcast.mapName"/>
+        <Set name="hazelcastInstanceName" property="jetty.session.hazelcast.hazelcastInstanceName"/>
+        <Set name="useQueries" property="jetty.session.hazelcast.useQueries"/>
+        <Set name="gracePeriodSec" property="jetty.session.gracePeriod.seconds"/>
+        <Set name="savePeriodSec" property="jetty.session.savePeriod.seconds"/>
         <Set name="onlyClient">
           <Property name="jetty.session.hazelcast.onlyClient" default="true" />
         </Set>
--- a/jetty-http2/http2-server/src/main/config/etc/jetty-http2c.xml
+++ b/jetty-http2/http2-server/src/main/config/etc/jetty-http2c.xml
@@ -6,10 +6,10 @@
     <Arg>
       <New class="org.eclipse.jetty.http2.server.HTTP2CServerConnectionFactory">
         <Arg name="config"><Ref refid="httpConfig"/></Arg>
-        <Set name="maxConcurrentStreams"><Property name="jetty.http2c.maxConcurrentStreams" default="1024"/></Set>
-        <Set name="initialStreamRecvWindow"><Property name="jetty.http2c.initialStreamRecvWindow" default="65535"/></Set>
+        <Set name="maxConcurrentStreams" property="jetty.http2c.maxConcurrentStreams"/>
+        <Set name="initialStreamRecvWindow" property="jetty.http2c.initialStreamRecvWindow"/>
         <Set name="initialSessionRecvWindow" property="jetty.http2.initialSessionRecvWindow"/>
-        <Set name="maxSettingsKeys"><Property name="jetty.http2.maxSettingsKeys" default="64"/></Set>
+        <Set name="maxSettingsKeys" property="jetty.http2.maxSettingsKeys"/>
         <Set name="rateControlFactory">
           <New class="org.eclipse.jetty.http2.parser.WindowRateControl$Factory">
             <Arg type="int"><Property name="jetty.http2.rateControl.maxEventsPerSecond" default="20"/></Arg>
--- a/jetty-http2/http2-server/src/main/config/etc/jetty-http2c.xml
+++ b/jetty-http2/http2-server/src/main/config/etc/jetty-http2c.xml
@@ -6,8 +6,8 @@
     <Arg>
       <New class="org.eclipse.jetty.http2.server.HTTP2CServerConnectionFactory">
         <Arg name="config"><Ref refid="httpConfig"/></Arg>
-        <Set name="maxConcurrentStreams" property="jetty.http2c.maxConcurrentStreams"/>
-        <Set name="initialStreamRecvWindow" property="jetty.http2c.initialStreamRecvWindow"/>
+        <Set name="maxConcurrentStreams"><Property name="jetty.http2c.maxConcurrentStreams" default="1024"/></Set>
+        <Set name="initialStreamRecvWindow"><Property name="jetty.http2c.initialStreamRecvWindow" default="65535"/></Set>
         <Set name="initialSessionRecvWindow" property="jetty.http2.initialSessionRecvWindow"/>
         <Set name="maxSettingsKeys"><Property name="jetty.http2.maxSettingsKeys" default="64"/></Set>
         <Set name="rateControlFactory">
--- a/jetty-rewrite/src/main/config/etc/jetty-rewrite-customizer.xml
+++ b/jetty-rewrite/src/main/config/etc/jetty-rewrite-customizer.xml
@@ -9,7 +9,7 @@
       <New id="Rewrite" class="org.eclipse.jetty.rewrite.RewriteCustomizer">
         <Set name="rewriteRequestURI" property="jetty.rewrite.rewriteRequestURI"/>
         <Set name="rewritePathInfo" property="jetty.rewrite.rewritePathInfo"/>
-        <Set name="originalPathAttribute" property="jetty.rewrite.originalPathAttribute"/>
+        <Set name="originalPathAttribute"><Property name="jetty.rewrite.originalPathAttribute" default="requestedPath"/></Set>
       </New>
     </Arg>
   </Call>  
--- a/jetty-rewrite/src/main/config/etc/jetty-rewrite.xml
+++ b/jetty-rewrite/src/main/config/etc/jetty-rewrite.xml
@@ -10,7 +10,7 @@
       <New class="org.eclipse.jetty.rewrite.handler.RewriteHandler">
         <Set name="rewriteRequestURI" property="jetty.rewrite.rewriteRequestURI" />
         <Set name="rewritePathInfo" property="jetty.rewrite.rewritePathInfo" />
-        <Set name="originalPathAttribute" property="jetty.rewrite.originalPathAttribute" />
+        <Set name="originalPathAttribute"><Property name="jetty.rewrite.originalPathAttribute" default="requestedPath"/></Set>
 
         <!-- Set DispatcherTypes  -->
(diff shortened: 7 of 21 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/Server.java
```diff
@@ -78,7 +78,7 @@
     private final ThreadPool _threadPool;
     private final List<Connector> _connectors = new CopyOnWriteArrayList<>();
     private SessionIdManager _sessionIdManager;
-    private boolean _stopAtShutdown;
+    private boolean _stopAtShutdown = true;
```
joakime: This will be a regression for embedded Jetty users that expect it to be false.

The true setting is only relevant for users of jetty-home / jetty-maven-plugin.
This default should be declared in the `server.mod` and code for jetty-maven-plugin.

joakime (changes requested):
The default for Server.stopAtShutdown should remain false (for embedded users).

jetty-server/src/main/java/org/eclipse/jetty/server/Server.java
janbartel: I really don't like the default being different .... but I understand that would be a change for embedded users. I'll make the default back to `true` for xml.

jetty-server/src/main/config/etc/jetty.xml
```diff
@@ -66,8 +66,8 @@
       <Set name="requestHeaderSize" property="jetty.httpConfig.requestHeaderSize"/>
       <Set name="responseHeaderSize" property="jetty.httpConfig.responseHeaderSize"/>
       <Set name="sendServerVersion" property="jetty.httpConfig.sendServerVersion"/>
-      <Set name="sendDateHeader" property="jetty.httpConfig.sendDateHeader"/>
-      <Set name="headerCacheSize" property="jetty.httpConfig.headerCacheSize"/>
+      <Set name="sendDateHeader"><Property name="jetty.httpConfig.sendDateHeader" default="false"/></Set>
+      <Set name="headerCacheSize"><Property name="jetty.httpConfig.headerCacheSize" default="4096"/></Set>
```
gregw: I think we can change the embedded default for headerCacheSize.
I'm tempted to say the same for sendDateHeader.... but I guess we'd better not.

jetty-http2/http2-server/src/main/config/etc/jetty-http2c.xml
```diff
@@ -6,8 +6,8 @@
     <Arg>
       <New class="org.eclipse.jetty.http2.server.HTTP2CServerConnectionFactory">
         <Arg name="config"><Ref refid="httpConfig"/></Arg>
-        <Set name="maxConcurrentStreams" property="jetty.http2c.maxConcurrentStreams"/>
-        <Set name="initialStreamRecvWindow" property="jetty.http2c.initialStreamRecvWindow"/>
+        <Set name="maxConcurrentStreams"><Property name="jetty.http2c.maxConcurrentStreams" default="1024"/></Set>
+        <Set name="initialStreamRecvWindow"><Property name="jetty.http2c.initialStreamRecvWindow" default="65535"/></Set>
```
gregw: @sbordet Can we change the defaults in the code so we don't need defaults in the XML as well?

jetty-gcloud/jetty-gcloud-session-manager/src/main/config-template/etc/sessions/gcloud/session-store.xml
```diff
@@ -10,11 +10,11 @@
   <Call name="addBean">
    <Arg>
     <New id="sessionDataStoreFactory" class="org.eclipse.jetty.gcloud.session.GCloudSessionDataStoreFactory">
-       <Set name="gracePeriodSec"><Property name="jetty.session.gracePeriod.seconds" default="3600" /></Set>
+       <Set name="gracePeriodSec" property="jetty.session.gracePeriod.seconds"/>
        <Set name="savePeriodSec" property="jetty.session.savePeriod.seconds"/>
        <Set name="maxRetries"><Property name="jetty.session.gcloud.maxRetries" default="5"/></Set>
-       <Set name="backoffMs" property="jetty.session.gcloud.backoffMs"/>
-       <Set name="namespace" property="jetty.session.gcloud.namespace"/>
+       <Set name="backoffMs"><Property name="jetty.session.gcloud.backoffMs" default="1000"/></Set>
```
gregw: Can we change this default in the code?  I think changing numeric defaults is OK, as it doesn't change behaviour, just when the behaviour happens.

jetty-gcloud/jetty-gcloud-session-manager/src/main/config-template/etc/sessions/gcloud/session-store.xml
```diff
@@ -10,11 +10,11 @@
   <Call name="addBean">
    <Arg>
     <New id="sessionDataStoreFactory" class="org.eclipse.jetty.gcloud.session.GCloudSessionDataStoreFactory">
-       <Set name="gracePeriodSec"><Property name="jetty.session.gracePeriod.seconds" default="3600" /></Set>
+       <Set name="gracePeriodSec" property="jetty.session.gracePeriod.seconds"/>
        <Set name="savePeriodSec" property="jetty.session.savePeriod.seconds"/>
        <Set name="maxRetries"><Property name="jetty.session.gcloud.maxRetries" default="5"/></Set>
```
gregw: Can we default this in the code as well?

jetty-unixsocket/jetty-unixsocket-server/src/main/config-template/etc/jetty-unixsocket-http2c.xml
```diff
@@ -6,8 +6,8 @@
     <Arg>
       <New class="org.eclipse.jetty.http2.server.HTTP2CServerConnectionFactory">
         <Arg name="config"><Ref refid="unixSocketHttpConfig"/></Arg>
-        <Set name="maxConcurrentStreams" property="jetty.http2c.maxConcurrentStreams"/>
-        <Set name="initialStreamRecvWindow" property="jetty.http2c.initialStreamRecvWindow"/>
+        <Set name="maxConcurrentStreams"><Property name="jetty.http2c.maxConcurrentStreams" default="1024"/></Set>
+        <Set name="initialStreamRecvWindow"><Property name="jetty.http2c.initialStreamRecvWindow" default="65535"/></Set>
```
gregw: @sbordet ditto - can these be defaulted just in the code?

gregw (changes requested):

sbordet:
@janbartel every change in XML config files requires likely a change in the `[ini-template] section of the correspondent `*.mod` file.

sbordet (changes requested):
Every change in XML config files requires likely a change in the `[ini-template]` section of the correspondent `*.mod` file.

jetty-gcloud/jetty-gcloud-session-manager/src/main/config-template/etc/sessions/gcloud/session-store.xml
janbartel: Actually I did default it in the code, I just forgot to remove it from the xml. I'll do that.

janbartel:
@sbordet the changes I have made do not require any changes in the corresponding `.mod` files - I checked them all and if a value is specified in the `.xml`, it is the same as the (commented out!) value in the `.mod` file and the pre https://github.com/eclipse/jetty.project/commit/343cf732b9a5aacd4c84fcb937a4b1624a1a0afd commit values.

In many cases, the commented out  value in the `.mod ` file or the value supplied in the `.xml` file is _different_ to the defaults in the code. Can you and @gregw please give some guidance on which values can be changed in the code vs which need to remain defaulted only in the `.xml`?

gregw:
@janbartel I'll defer my re-review until @sbordet comments

joakime (dismissed):

jetty-gcloud/jetty-gcloud-session-manager/src/main/java/org/eclipse/jetty/gcloud/session/GCloudSessionDataStore.java
```diff
@@ -65,7 +66,7 @@
     protected boolean _indexesPresent = false;
     protected EntityDataModel _model;
     protected boolean _modelProvided;
-    private String _namespace;
+    private String _namespace = DEFAULT_NAMESPACE;
```
joakime, line 69: This is an embedded mode change, it used to return null here, now it returns empty string.

jetty-gcloud/jetty-gcloud-session-manager/src/main/java/org/eclipse/jetty/gcloud/session/GCloudSessionDataStoreFactory.java
```diff
@@ -22,9 +22,9 @@
  */
 public class GCloudSessionDataStoreFactory extends AbstractSessionDataStoreFactory
 {
-    private String _namespace;
-    private int _maxRetries;
-    private int _backoffMs;
+    private String _namespace = GCloudSessionDataStore.DEFAULT_NAMESPACE;
+    private int _maxRetries = GCloudSessionDataStore.DEFAULT_MAX_RETRIES;
+    private int _backoffMs = GCloudSessionDataStore.DEFAULT_BACKOFF_MS;
```
joakime, line 27: These 3 are embedded mode default changes.
These used to be (in embedded mode) null, 0, 0.

jetty-server/src/main/java/org/eclipse/jetty/server/session/AbstractSessionDataStoreFactory.java
```diff
@@ -19,8 +19,8 @@
 public abstract class AbstractSessionDataStoreFactory implements SessionDataStoreFactory
 {
 
-    int _gracePeriodSec;
-    int _savePeriodSec;
+    int _gracePeriodSec = AbstractSessionDataStore.DEFAULT_GRACE_PERIOD_SEC;
+    int _savePeriodSec = AbstractSessionDataStore.DEFAULT_SAVE_PERIOD_SEC;
```
joakime, line 23: These are embedded mode changes.
It used to be 0 for each.

joakime (commented):
There are still changes for embedded mode users.

jetty-gcloud/jetty-gcloud-session-manager/src/main/java/org/eclipse/jetty/gcloud/session/GCloudSessionDataStore.java
janbartel: No, if the namespace is null at startup the GCloudDataStore uses "".

jetty-gcloud/jetty-gcloud-session-manager/src/main/java/org/eclipse/jetty/gcloud/session/GCloudSessionDataStoreFactory.java
janbartel: See above

jetty-server/src/main/java/org/eclipse/jetty/server/session/AbstractSessionDataStoreFactory.java
janbartel: See above

jetty-http2/http2-server/src/main/config/etc/jetty-http2c.xml
sbordet: @gregw we should change these default to match those in the code, as these defaults are quite wrong.

jetty-unixsocket/jetty-unixsocket-server/src/main/config-template/etc/jetty-unixsocket-http2c.xml
sbordet: Same.

jetty-server/src/main/config/etc/jetty.xml
```diff
@@ -105,7 +105,7 @@
     <!-- =========================================================== -->
     <!-- extra server options                                        -->
     <!-- =========================================================== -->
-    <Set name="stopAtShutdown" property="jetty.server.stopAtShutdown"/>
+    <Set name="stopAtShutdown"><Property name="jetty.server.stopAtShutdown" default="true"/></Set>
```
sbordet, line 108: Does not this require a test with the `JettyHomeTester`?

sbordet (changes requested):

jetty-http2/http2-server/src/main/config/etc/jetty-http2c.xml
gregw: @sbordet I think the idea is to not have defaults in the XML if at all possible.  If the defaults in the code are correct, then we should just change these to the new form of the XML that only sets if the property is set.

jetty-server/src/main/config/etc/jetty.xml
janbartel: I tried, but it's just not possible with the current setup. When `destroy() `is called on the remote jetty process the IO pipes to that process seem to be immediately closed, so we never see the appropriate `"Stopped Server"` log message (although it is generated, as running the distro at the command line attests). I think this is a much bigger issue, namely how do we test each and every module to ensure correctness, so I've opened a separate issue: #6373

jetty-http2/http2-server/src/main/config/etc/jetty-http2c.xml
janbartel: Removed defaults from xml in favour of using code defaults instead.

jetty-server/src/main/config/etc/jetty.xml
janbartel: Removed default for headerCacheSize, used code defaults instead.

jetty-unixsocket/jetty-unixsocket-server/src/main/config-template/etc/jetty-unixsocket-http2c.xml
janbartel: Removed xml defaults in favour of using code defaults instead.

janbartel:
@gregw @sbordet  I have removed the defaults set in xml in favour of using the defaults in the code as discussed. Where appropriate I have changed the .mod files to also use the same values as the code defaults. Please rereview.

gregw (dismissed):

jetty-gcloud/jetty-gcloud-session-manager/src/main/config-template/etc/sessions/gcloud/session-store.xml
```diff
@@ -44,6 +44,9 @@
               <Set name="lastNode">
                 <Property name="jetty.session.gcloud.model.lastNode" default="lastNode" />
               </Set>
+              <Set name="lastSaved">
+                <Property name="jetty.session.gcloud.model.lastSaved" default="lastSaved" />
+              </Set>
```
sbordet: What's the rationale to have some as `property` attribute and some as `Property` elements?
The code and the `*.mod` `[ini-template]` are consistent, so I guess they can all be `property` attributes?

jetty-hazelcast/src/main/config/etc/sessions/hazelcast/remote.xml
```diff
@@ -10,21 +10,11 @@
...
-          <Property name="jetty.session.savePeriod.seconds" default="0" />
-        </Set>
+        <Set name="mapName" property="jetty.session.hazelcast.mapName"/>
+        <Set name="hazelcastInstanceName" property="jetty.session.hazelcast.hazelcastInstanceName"/>
+        <Set name="useQueries" property="jetty.session.hazelcast.useQueries"/>
+        <Set name="gracePeriodSec" property="jetty.session.gracePeriod.seconds"/>
+        <Set name="savePeriodSec" property="jetty.session.savePeriod.seconds"/>
         <Set name="onlyClient">
           <Property name="jetty.session.hazelcast.onlyClient" default="true" />
         </Set>
```
sbordet, line 20: Why was this kept as `Property`? Is it correct that embedded sees `false`, but standalone server sees `true`?

jetty-rewrite/src/main/config/etc/jetty-rewrite-customizer.xml
```diff
@@ -9,7 +9,7 @@
       <New id="Rewrite" class="org.eclipse.jetty.rewrite.RewriteCustomizer">
         <Set name="rewriteRequestURI" property="jetty.rewrite.rewriteRequestURI"/>
         <Set name="rewritePathInfo" property="jetty.rewrite.rewritePathInfo"/>
-        <Set name="originalPathAttribute" property="jetty.rewrite.originalPathAttribute"/>
+        <Set name="originalPathAttribute"><Property name="jetty.rewrite.originalPathAttribute" default="requestedPath"/></Set>
```
sbordet, line 12: Seems like a breaking change?

jetty-rewrite/src/main/config/etc/jetty-rewrite.xml
```diff
@@ -10,7 +10,7 @@
       <New class="org.eclipse.jetty.rewrite.handler.RewriteHandler">
         <Set name="rewriteRequestURI" property="jetty.rewrite.rewriteRequestURI" />
         <Set name="rewritePathInfo" property="jetty.rewrite.rewritePathInfo" />
-        <Set name="originalPathAttribute" property="jetty.rewrite.originalPathAttribute" />
+        <Set name="originalPathAttribute"><Property name="jetty.rewrite.originalPathAttribute" default="requestedPath"/></Set>
```
sbordet, line 13: Ditto.

jetty-server/src/main/config/etc/jetty-threadlimit.xml
```diff
@@ -10,7 +10,7 @@
     <Arg>
       <New id="ThreadLimitHandler" class="org.eclipse.jetty.server.handler.ThreadLimitHandler">
         <Arg name="forwardedHeader"><Property name="jetty.threadlimit.forwardedHeader"/></Arg> 
-        <Set name="enabled" property="jetty.threadlimit.enabled"/> 
+        <Set name="enabled"><Property name="jetty.threadlimit.enabled" default="true"/></Set>
```
sbordet, line 13: Was not enabled before, why was changed to enabled?

jetty-server/src/main/config/etc/jetty.xml
```diff
@@ -66,7 +66,7 @@
       <Set name="requestHeaderSize" property="jetty.httpConfig.requestHeaderSize"/>
       <Set name="responseHeaderSize" property="jetty.httpConfig.responseHeaderSize"/>
       <Set name="sendServerVersion" property="jetty.httpConfig.sendServerVersion"/>
-      <Set name="sendDateHeader" property="jetty.httpConfig.sendDateHeader"/>
+      <Set name="sendDateHeader"><Property name="jetty.httpConfig.sendDateHeader" default="false"/></Set>
```
sbordet, line 69: Another behavior change? Why?

jetty-server/src/main/config/etc/jetty-gzip.xml
```diff
@@ -13,11 +13,7 @@
     <Arg>
       <New id="GzipHandler" class="org.eclipse.jetty.server.handler.gzip.GzipHandler">
         <Set name="minGzipSize" property="jetty.gzip.minGzipSize"/>
-        <Set name="checkGzExists" property="jetty.gzip.checkGzExists"/>
-        <Set name="compressionLevel" property="jetty.gzip.compressionLevel"/>
         <Set name="inflateBufferSize" property="jetty.gzip.inflateBufferSize"/>
-        <Set name="inflaterPoolCapacity" property="jetty.gzip.inflaterPoolCapacity"/>
-        <Set name="deflaterPoolCapacity" property="jetty.gzip.deflaterPoolCapacity"/>
```
sbordet, line 20: I think this XML is out-of-date. It should be changed to allocate the `InflaterPool` and `DeflaterPool` and the relevant properties set on the pools (rather than on the handler as it is done here).

jetty-server/src/main/config/modules/server.mod
```diff
@@ -63,7 +63,7 @@ etc/jetty.xml
 # jetty.httpConfig.compliance=RFC7230
 
 ## URI Compliance: SAFE, STRICT
-# jetty.httpConfig.uriCompliance=SAFE
+# jetty.httpConfig.uriCompliance=DEFAULT
```
sbordet, line 66: Update comment with possible values, as this may be included in documentation.

sbordet (changes requested):
There seems to be a lot of breaking changes in XML files that seem to be out of scope for this PR, if the scope was initially to fix `stopOnShutdown` behavior.

Also, feels like many/most properties should be `property` attributes so that embedded and standalone have the same configuration.
However, some `property` attribute was changed to `Property` element and viceversa, but could not find a clear logic for that?

jetty-gcloud/jetty-gcloud-session-manager/src/main/config-template/etc/sessions/gcloud/session-store.xml
janbartel: The `Property` element allows a default value to be set, whereas the `property` attribute does not.

jetty-hazelcast/src/main/config/etc/sessions/hazelcast/remote.xml
janbartel: This is a `Property` because we are setting the `default` value. As the same HazelcastSessionDataStoreFactory is used in both the `remote.xml `and `default.xml`, we cannot default to the value in the class, but must set this field explicitly to match either the remote or embedded case.  The other alternative is to make it a `property`, and remove the default and trust that the value in the .`mod/.ini `file is not commented out/correct.

jetty-rewrite/src/main/config/etc/jetty-rewrite-customizer.xml
janbartel: It was this way before the commit that changed the xml that kicked off this PR: https://github.com/eclipse/jetty.project/commit/343cf732b9a5aacd4c84fcb937a4b1624a1a0afd

jetty-rewrite/src/main/config/etc/jetty-rewrite.xml
janbartel: See above comment.

jetty-server/src/main/config/etc/jetty-threadlimit.xml
janbartel: Because if its not enabled it's not doing anything, and the value previously was `true`, so I've reinstated it. The value in the class is `false`. Again, see commit https://github.com/eclipse/jetty.project/commit/343cf732b9a5aacd4c84fcb937a4b1624a1a0afd

jetty-server/src/main/config/etc/jetty.xml
janbartel: @sbordet I don't think you've understood the fundamental of this PR :)  The whole point is that commit https://github.com/eclipse/jetty.project/commit/343cf732b9a5aacd4c84fcb937a4b1624a1a0afd accidentally changed the configuration by removing the `default` values, so this PR is putting them back the way they were.

jetty-server/src/main/config/etc/jetty-gzip.xml
janbartel: Is it fine to take a all the default values for the `InflaterPool` and `DeflaterPool`?

jetty-server/src/main/config/modules/server.mod
janbartel: Done.

jetty-gcloud/jetty-gcloud-session-manager/src/main/config-template/etc/sessions/gcloud/session-store.xml
gregw: @sbordet specifically, sometime we "want" to have different defaults for embedded usage (default in the code) vs deployed usage (default in the XML).    I'm not sure we should create any more situations where the default is different depending on the run mode, but we do have several historic cases.     We did originally convert some of them to `property` elements (ie use only the default from the code), but that caused complaints.   So we need to use `Property` element whenever we want the xml to always set a value, even if the property is not set.

jetty-rewrite/src/main/config/etc/jetty-rewrite-customizer.xml
gregw: I think in this case  we can make this one a `property` as I see no reason for the embedded default to be different from the deployed default.

jetty-rewrite/src/main/config/etc/jetty-rewrite.xml
gregw: ditto

jetty-server/src/main/config/etc/jetty-gzip.xml
gregw: I think yes.

Note that I am going to open another issue.  We need a way to check that `property` attributes actually exist, else we can change the code and not notice the XML is wrong unless the property is actually set.     I think we should always validate that there is at least one matching setter, even if the property is not set.

jetty-gcloud/jetty-gcloud-session-manager/src/main/config-template/etc/sessions/gcloud/session-store.xml
janbartel: I've converted all the `<Property>` elements to `property` attributes as the `defaults` are the same as the code (and the corresponding `.mod` file).

jetty-rewrite/src/main/config/etc/jetty-rewrite-customizer.xml
janbartel: So to be crystal clear on this:   `RewriteCustomer` has a default of `null` for `_originalPathAttribute`. The `rewrite-customizer.mod` file has a _commented out_ value for `jetty.rewrite.originalPathAttribute=requestedPath`. Before commit https://github.com/eclipse/jetty.project/commit/343cf732b9a5aacd4c84fcb937a4b1624a1a0afd the xml defaulted it to `requestedPath`. So this PR is simply putting things back the way they were. @gregw and @sbordet do either of you believe the value should now be defaulted _differently_???? If so, to what?

jetty-rewrite/src/main/config/etc/jetty-rewrite.xml
janbartel: Both @sbordet and @gregw see above comment.

jetty-server/src/main/config/etc/jetty-gzip.xml
janbartel: @sbordet  I have configured an `InflaterPool` and a `DeflaterPool`.  Note that I have had to change the names of the properties in the `.mod` file as they no longer make sense - I've put them in the `jetty-rewrite.xml ` as `deprecated`. I've used the default values that are present in the code.

joakime (dismissed):

sbordet:
I think it's a mistake to try to revert 343cf73.

All releases of Jetty 10, since 10.0.0-alpha0 included, had the 343cf73 behavior.
Which means that no official release of Jetty 10 had any different behavior.

This PR is trying to revert that commit and instate many new behaviors, breaking who was relying on the behavior of basically all the Jetty 10 releases.

@janbartel if you look at the issue that generated this PR (#6329), it was about `jetty.server.stopAtShutdown` only.
That it was caused by 343cf73 does not mean that we have to change many other behaviors.

I'm fine to change `<Property>` elements into attributes or viceversa, as long as the behavior is maintained or it's a blatant bug like `stopAtShutdown`.

If there are wrong behaviors, let's review them in different issues so we can enumerate the behaviors that were changed one by one in the release notes, and at least users will know what behavior changed and why.

jetty-server/src/main/config/etc/jetty-gzip.xml
```diff
@@ -26,6 +22,19 @@
...
+        <Set name="inflaterPool">
+            <New class="org.eclipse.jetty.util.compression.InflaterPool">
+                <Arg type="int"><Property name="jetty.gzip.inflaterPool.capacity" deprecated="jetty.gzip.inflaterPoolCapacity" default="1024"/></Arg>
+                <Arg type="boolean"><Property name="jetty.gzip.inflaterPool.noWrap" default="true"/></Arg>
+            </New>
+        </Set>
+        <Set name="deflaterPool">
+            <New class="org.eclipse.jetty.util.compression.DeflaterPool">
+                <Arg type="int"><Property name="jetty.gzip.deflaterPool.capacity" deprecated="jetty.gzip.deflaterPoolCapacity" default="1024"/></Arg>
+                <Arg type="int"><Property name="jetty.gzip.deflaterPool.compression" deprecated="jetty.gzip.compression" default="-1"/></Arg>
```
sbordet: The `deprecated` attribute should be `jetty.gzip.compressionLevel`.
I would use `jetty.gzip.deflaterPool.compression*Level*` for the new name, otherwise looks like a boolean.

jetty-server/src/main/config/etc/jetty-gzip.xml
sbordet: You did not put them in `jetty-rewrite.xml`, did you?

jetty-rewrite/src/main/config/etc/jetty-rewrite-customizer.xml
sbordet: This should be restored to the same behavior it had for all the Jetty 10 versions.
The default value of `null` means to not store the originalPath as a request attribute, which is a valid behavior and has been so for all Jetty 10 versions.

jetty-server/src/main/config/etc/jetty-threadlimit.xml
sbordet: Again, breaks the behavior of all Jetty 10 releases.
If it's a wrong value, let's open a new issue rather than doing it here where it gets lost in other changes.

jetty-server/src/main/config/modules/gzip.mod
```diff
@@ -21,17 +21,23 @@ etc/jetty-gzip.xml
...
+#jetty.gzip.inflaterPool.noWrap=true
+
 ## Deflater pool max size (-1 for unlimited, 0 for no pooling)
-# jetty.gzip.deflaterPoolCapacity=-1
+# jetty.gzip.deflaterPool.capacity=1024
 
-## Inflater pool max size (-1 for unlimited, 0 for no pooling)
-# jetty.gzip.inflaterPoolCapacity=-1
+## Gzip compression level (-1 for default)
+# jetty.gzip.deflaterPool.compressionLevel=-1
```
sbordet, line 37: This is the right property name, but the correspondent XML was using the wrong name.

sbordet (changes requested):

gregw:
@sbordet you can't both argue that this is an issue that can be punted from a release train AND that the behavior should be kept as the changed one because of previous releases.

I think version 1 if this PR did just fix the missing attribute issue until @joakime pointed out the inadvertent behavior change.

I see no reason to keep any accidental behavior changes

jetty-server/src/main/config/etc/jetty-gzip.xml
janbartel: OK, fixed.

jetty-server/src/main/config/modules/gzip.mod
janbartel: Fixed.

jetty-server/src/main/config/etc/jetty-gzip.xml
janbartel: Sorry, working on too many things at once. They are of course in` jetty-gzip.xml`.

janbartel:
> I think it's a mistake to try to revert 343cf73.
> 
> All releases of Jetty 10, since 10.0.0-alpha0 included, had the 343cf73 behavior.
> Which means that no official release of Jetty 10 had any different behavior.

Which means that all official releases of jetty-10 are in fact incorrect/broken.
Just because we have had n broken releases is not a reason to make n+1.

> 
> This PR is trying to revert that commit and instate many new behaviors, breaking who was relying on the behavior of basically all the Jetty 10 releases.
> 
> @janbartel if you look at the issue that generated this PR (#6329), it was about jetty.server.stopAtShutdown only.
> That it was caused by 343cf73 does not mean that we have to change many other behaviors.
> 
> I'm fine to change <Property> elements into attributes or viceversa, as long as the behavior is maintained or it's a blatant bug like stopAtShutdown.
Well, the `InflaterPool/DeflaterPool `is a blatant bug. If anyone had actually configured a property value, jetty would not have been able to start.

The jetty-threadlimit.xml is also a blatant bug: if you enabled the threadlimit module, you would be forgiven for assuming it was actually limiting, but without this change it is not!

The only nice-to-haves is the clean-up of the session modules default values, the rest you can easily argue are blatant bugs, or rotten `.mod` files.

> If there are wrong behaviors, let's review them in different issues so we can enumerate the behaviors that were changed one by one in the release notes, and at least users will know what behavior changed and why.

sbordet (approved):

gregw (approved):

PR #6349, opened 2021-06-02
Issue #6347 Fix session-store-gcloud,logging-jcl-capture,logging-log4j1-capture

### LINKED ISSUE #6347: session-store-gcloud module broken logging dependency
jetty-10 head

> [589] java -jar ../jetty-home/start.jar --add-modules=session-store-gcloud
> ERROR : Module logging-jul provides logging, which is already provided by logging-jetty enabled in [transitive provider of logging for threadpool, transitive provider of logging for logging-jcl-capture]
> 
> Usage: java -jar $JETTY_HOME/start.jar [options] [properties] [configs]
>        java -jar $JETTY_HOME/start.jar --help  # for more information
> 

Our documentation says:

> As GCloud requires certain Java Commons Logging features to work correctly, Jetty routes these through SLF4J by transitively enabling the jcl-slf4j module during installation. Therefore, you will also need to enable one of the SLF4J implementation modules. You can either choose one ahead of time and enable it at the same time as the session-store-gcloud module, or you can just enable session-store-gcloud module and it will print out a list of available SLF4J implementations. You can then choose one and enable it.

Either the module or the documentation or both are broken.

janbartel:
closes #6347 

gcloud-datastore.mod depended on wrong jul logging module: was logging-jul but should be logging-jul-capture.

Along the way I discovered that the logging-jul-capture and loggin-jcl-capture were both broken, so fixed those.

### DIFF
--- a/documentation/jetty-documentation/src/main/asciidoc/operations-guide/sessions/session-gcloud.adoc
+++ b/documentation/jetty-documentation/src/main/asciidoc/operations-guide/sessions/session-gcloud.adoc
@@ -62,10 +62,8 @@ The `session-store-gcloud` module provides GCloud support for storing session da
 
 Because the Google Cloud DataStore is not a technology provided by the Eclipse Foundation, when enabling the module you will be prompted to assent to the licenses of the external vendor.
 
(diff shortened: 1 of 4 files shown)


joakime (approved):
LGTM

PR #6351, opened 2021-06-03
Fixes #5931 - SslConnection should implement getBytesIn()/getBytesOut…

### LINKED ISSUE #5931: SslConnection should implement getBytesIn()/getBytesOut()

sbordet:
…(). (#6335)

Updated ConnectionStatistics to report both the stats of all connections,
and the stats grouped by connection class.

(cherry picked from commit f902d12fe8b7a69d7f697db0ce8ff4495c9e2141)


sbordet:
Forward port of #6335.

gregw (approved):

PR #6353, opened 2021-06-03
Rename EWYK The AdaptiveExecutionStrategy

### LINKED ISSUE #6391: Rename EatWhatYouKill to AdaptiveExecutionStrategy
**Target Jetty version(s)**
10, 11
**Enhancement Description**

Rename EatWhatYouKill to AdaptiveExecutionStrategy, which better represents the nature of the current algorith,
Improve the javadoc and code clarity.

gregw:
Rename EWYK to AdaptiveExecutionStrategy, which better represents the nature of the strategy.
#6391 


jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -32,27 +32,44 @@
...
+ * This sub-strategy is called ProduceExecuteConsume (PEC).</p>
+ *
+ * <p>If the producing thread may block, then for produced task that also may block,
+ * the strategy may attempts to dispatch another producing thread via a {@link TryExecutor}.
+ * If either there is already a producer thread pending or the call
+ * to {@link TryExecutor#tryExecute(Runnable)} succeeds, then the task
+ * is executed directly and the pending producer will take over production.
+ * This sub-strategy is called Execute Produce Consume (EPC), but was previously
+ * named EatWhatYouKill (EWYK) after a hunting proverb, in the sense that one
+ * should kill(produce) only to eat(consume).</p>
```
sbordet: No need to mention EWYK anymore.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -32,27 +32,44 @@
...
- * <p>However, if the task produced uses the {@link Invocable} API to indicate that
- * it will not block, then the strategy will run it directly, regardless of the
- * presence of a pending producer thread and then resume production after the
- * task has completed. When operating in this pattern, the sub-strategy is called
- * ProduceConsume (PC).</p>
- * <p>If there is no pending producer thread available and if the task has not
- * indicated it is non-blocking, then this strategy will dispatch the execution of
- * the task and immediately continue production. When operating in this pattern, the
- * sub-strategy is called ProduceExecuteConsume (PEC).</p>
+ * <p>If a produced task has used the {@link Invocable} API to indicate that
```
sbordet: Link exact API: `{@link Invocable#invokeNonBlocking(Runnable)`

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -32,27 +32,44 @@
...
+ * If either there is already a producer thread pending or the call
+ * to {@link TryExecutor#tryExecute(Runnable)} succeeds, then the task
+ * is executed directly and the pending producer will take over production.
+ * This sub-strategy is called Execute Produce Consume (EPC), but was previously
+ * named EatWhatYouKill (EWYK) after a hunting proverb, in the sense that one
+ * should kill(produce) only to eat(consume).</p>
+ *
+ * <p>If there is no pending producer thread available and the produce task
+ * is {@link Invocable.InvocationType#EITHER}, then the PIC sub-strategy is
+ * used. Otherwise the PEC sub-strategy is used.</p>
```
sbordet: IMHO the whole javadoc can be improved in this way:

Opening paragraph is good.

Then mention that a task can use `invokeNonBlocking()`, therefore marking the _thread_ as non-blocking.
Then mention that strategies may be cascaded, so the next can know if it has been invoked from a non-blocking thread.

Then mention the various cases, using bullet lists or `dl`, `dd`, `dt` to enumerate the various cases with less prose.
Something like:

* Invoked from non-blocking thread; task produced is:
** BLOCKING: foo...
** NON_BLOCKING: bar...
** EITHER: baz...

etc.

sbordet (changes requested):

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
gregw: I disagree.  We have put considerable effort into publicizing EWYK as a smart strategy.  I don't think we should dissappear it entirely.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
gregw: Agreed.   I rewrote it in that style and ultimately it came out a lot simpler.... so much so that it indicated that the actual code itself was too complex when selecting the sub-strategy.
So I rewrote the code to follow the documented substrategy choice.  I'm 99% sure it is exactly the same algorithm, just simpler.  But if there is a doubt, then let's just change documentation and leave the code.  @lachlan-roberts @lorban extra eyes would be handy.

lachlan-roberts: @gregw The algorithm looks the same to me, only difference I can see is you were previously using the lock for some of the logic and you aren't using it anymore. Was this intentional?

gregw: Doh! Good catch!

lachlan-roberts (dismissed):
LGTM

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -0,0 +1,465 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * <p>An adaptive execution strategy that uses the {@link Invocable} status
+ * of both the task and the current thread to select an optimal strategy
+ * that prioritizes executing the task immediately in the current
+ * producing thread if it can be done so without thread starvation issues.</p>
+ *
+ * This strategy selects between the following sub-strategies:
```
sbordet: Surround with `<p>`.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -0,0 +1,465 @@
...
+ *     and then continues to produce.</dd>
+ *     <dt>ProduceExecuteConsume(PEC)</dt>
+ *     <dd>The producing thread dispatches the task to a thread pool to be executed and then immediately resumes
+ *     producing.</dd>
+ *     <dt>ExecuteProduceConsume(EPC)</dt>
+ *     <dd>The producing thread consumes the task by executing it directly (as in PC mode) but then races with
+ *     a pending producer thread to take over production.
+ *     </dd>
+ * </dl>
+ * The sub-strategy is selected as follows:
```
sbordet: Surround with `<p>`.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -0,0 +1,465 @@
...
+ * <dl>
```
sbordet, line 59: Put `<dd>` on new lines for clarity.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -0,0 +1,465 @@
...
+ *     <dt>PC</dt><dd>If the produced task has been invoked with {@link Invocable#invokeNonBlocking(Runnable)}
+ *     to indicate that it is {@link Invocable.InvocationType#NON_BLOCKING}.</dd>
```
sbordet:
```suggestion
 *     <dt>PC</dt>
 *     <dd>When the producing thread is non-blocking, and the produced task is {@link Invocable.InvocationType#NON_BLOCKING}.</dd>
```

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -0,0 +1,465 @@
...
+ *     <dt>EPC</dt><dd>If the producing thread is not {@link Invocable.InvocationType#NON_BLOCKING}
+ *     and a pending producer thread is available, either because there is already a pending producer
+ *     or one is successfully started with {@link TryExecutor#tryExecute(Runnable)}.</dd>
```
sbordet:
```suggestion
 *     <dt>EPC</dt>
 *    <dd>When the producing thread is not marked as non-blocking, and a producer thread is available,
 *    since {@link TryExecutor#tryExecute(Runnable)} returned {@code true}.</dd>
```

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -0,0 +1,465 @@
...
+ *     <dt>PIC</dt><dd>If the produced task has used the {@link Invocable#getInvocationType()} API to
+ *     indicate that it is {@link Invocable.InvocationType#EITHER}.</dd>
```
sbordet:
```suggestion
 *     <dt>PIC</dt>
 *     <dd>When the producing thread is not marked as non-blocking, and a producer thread is not available,
 *     and the produced task is {@link Invocable.InvocationType#EITHER}.</dd>
```

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -0,0 +1,465 @@
...
+ *     <dt>PEC</dt><dd>Otherwise.</dd>
```
sbordet:
```suggestion
 *     <dt>PEC</dt>
 *     <dd>Otherwise.</dd>
```

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -0,0 +1,465 @@
...
+ * </dl>
+ *
+ * <p>Because of the preference for {@code PC} mode, on a multicore machine with many
+ * many {@link Invocable.InvocationType#NON_BLOCKING} tasks, multiple instances of the strategy may be
+ * required to keep all CPUs on the system busy.</p>
+ *
+ * <p>Since the producing thread may be invoked with {@link Invocable#invokeNonBlocking(Runnable)}
+ * this allows {@link AdaptiveExecutionStrategy}s to be efficiently and safely chained: so that a task
```
sbordet: Remove "so that".

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -0,0 +1,465 @@
...
+ * produced by one execution strategy may become itself be a producer in a second execution strategy
+ * (e.g. an IO selector may use an execution strategy to handle multiple connections and each
+ * connection may use a execution strategy to handle multiplexed channels/streams within the connection).
+ * A task containing another {@link AdaptiveExecutionStrategy} should identify as
```
sbordet: Start another `<p>`.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -0,0 +1,465 @@
...
+ * {@link Invocable.InvocationType#EITHER} so when there are no pending producers threads available to
+ * the first strategy, then it may invoke the second as {@link Invocable.InvocationType#NON_BLOCKING}.
+ * This avoids starvation as the production on the second strategy can always be executed,
+ * but without the risk that it may block the last available producer for the first strategy.</p>
+ *
+ * <p>This strategy was previously named EatWhatYouKill (EWYK) because its preference to for
+ * a producer to directly consume a task was similar to the hunting proverb, in the sense that one
+ * should eat(consume) what they kill(produce).</p>
```
sbordet:
```suggestion
 * <p>This strategy was previously named EatWhatYouKill (EWYK) because it prefers that
 * a producer directly consumes the task it produces, similarly to the hunting proverb that says that one
 * should eat (i.e. consume) what is killed (i.e. produced).</p>
```

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -0,0 +1,465 @@
...
+                        : Mode.PRODUCE_EXECUTE_CONSUME;
+                }
+            }
+        }
+        else
+        {
+            mode = taskType == Invocable.InvocationType.EITHER
+                ? Mode.PRODUCE_INVOKE_CONSUME
+                : Mode.PRODUCE_EXECUTE_CONSUME;
+        }
```
sbordet: I think this block should be reverted back to its original code, for 2 reasons:
1. This is a 3-way if/else with the first `if` having a different condition than the second if/else. The previous version was clearer in that it only had a if/else on the thread blocking/non-blocking and was so much easier to read.
2. The previous version had a lot of good comments that have been all stripped out but they are very important to keep.

The cost is exactly the same, so I prefer the previous version because it allows better reasoning by splitting all the cases, rather than this version that coalesces multiple cases.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/EatWhatYouKill.java
```diff
@@ -13,469 +13,16 @@
...
- * ProduceConsume (PC).</p>
- * <p>If there is no pending producer thread available and if the task has not
- * indicated it is non-blocking, then this strategy will dispatch the execution of
- * the task and immediately continue production. When operating in this pattern, the
- * sub-strategy is called ProduceExecuteConsume (PEC).</p>
+ * @deprecated This class has been renamed to {@link AdaptiveExecutionStrategy}
  */
-@ManagedObject("eat what you kill execution strategy")
-public class EatWhatYouKill extends ContainerLifeCycle implements ExecutionStrategy, Runnable
+@Deprecated
```
sbordet: Perhaps add `forRemoval=true`?

sbordet (changes requested):

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
gregw: I'd prefer to be explicit, as just saying "is non-blocking" is not clear.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
gregw: ditto

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
gregw: which previous version?  The version prior to this PR or the version prior to the last commit?

The problem with the version prior to the last commit was that it didn't lock.  To add locking means that it either needs to take the lock when it doesn't need to, or some code needs to be reproduced, or we need to add a null check.  

Let me try some of those alternatives...

gregw: Oh you must have meant the version before this PR.  Nah!!!! the comments were clear, but the code was impenatrable and too long.  It took me some time to extract the simple description of the algorithm for the javadoc.    I have pushed another version of the refactored algorithm, this time with good comments and a little "hack" on the lock to avoid code duplication and/or null checks.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/AutoLock.java
```diff
@@ -34,6 +34,38 @@
...
+        boolean isLocked()
+        {
+            return false;
+        }
+
+        @Override
+        public void close()
+        {
+        }
+    };
```
sbordet: @gregw this does not make sense.

This is only necessary because of a refactoring in the strategy, only to _try_ to simplify a perfectly readable algorithm into something that is now completely unreadable that probably also confuses the JIT.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -0,0 +1,482 @@
...
+            mode = Mode.PRODUCE_CONSUME;
+        }
+        else
+        {
+            // The produced task might block. Is this producing thread allowed to block?
+            boolean mayBlock = !nonBlocking;
+
+            // only if this thread may block do we need to take the lock so we can
+            // atomically check/mutate the state and pending status.
+            try (AutoLock l = mayBlock ? _lock.lock() : AutoLock.NO_LOCK)
```
sbordet: Please no.
Can we just revert the algorithm as it was before? 
Much simpler, much more readable, each case had its own block and it was easy to reason.

sbordet (changes requested):

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
gregw: I don't agree. In the previous algorithm, each Sun strategy is selected multiple times in multiple locations over several pages of code, in a store that didn't at all match the documentation.

I'll revert to the version without the lock back, but dinner duplication of code... but with comments

gregw:
@sbordet I have pushed a version without the lock hack that well matches the documented algorithm.  Note also with this forumulation there is one less test the common case for a non-blocking task, so it will take a much simpler path.

The cost is that for the case of an potentially blocking task when there are no pending producers available, then the taskType is checked twice. This double check can be removed, but only by breaking out the EITHER and BLOCKING cases and thus duplicating the synchronized block and changing the flow from the documentation... but that may also be OK.... I'll add it as a comment to compare.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -0,0 +1,490 @@
...
+                    // Use PIC to directly consume the task and then resume production.
+                    mode = Mode.PRODUCE_INVOKE_CONSUME;
+                    break;
+                }
+
+                // Otherwise we use PEC: this thread continues to produce and the task is
+                // consumed by the executor
+                mode = Mode.PRODUCE_EXECUTE_CONSUME;
+                break;
+        }
```
gregw: This is an alternative with some code duplication, but one less test in one case:
```suggestion
        Mode mode;
        // Switch on the invocation type of the produced task
        Invocable.InvocationType taskType = Invocable.getInvocationType(task);
        switch (taskType)
        {
            case NON_BLOCKING:
                // The produced task will never block, so we directly consume the task with PC
                // and then resume production.
                mode = Mode.PRODUCE_CONSUME;
                break;

            case BLOCKING:
                // if this thread may block then we need to take the lock so we can
                // atomically check/mutate the state and pending status.
                if (!nonBlocking)
                {
                    try (AutoLock l = _lock.lock())
                    {
                        if (_pending || _tryExecutor.tryExecute(this))
                        {
                            // This producing thread may block and there is a pending producer available
                            // so we use EPC: the producer will directly consume the task and then
                            // race with the pending producer to take over production.
                            _pending = true;
                            _state = State.IDLE;
                            mode = Mode.EXECUTE_PRODUCE_CONSUME;
                            break;
                        }
                    }
                }

                // otherwise we use PEC: this thread continues to produce and the task is consumed by the executor
                mode = Mode.PRODUCE_EXECUTE_CONSUME;
                break;

            case EITHER:
                // if this thread may block then we need to take the lock so we can
                // atomically check/mutate the state and pending status.
                if (!nonBlocking)
                {
                    try (AutoLock l = _lock.lock())
                    {
                        if (_pending || _tryExecutor.tryExecute(this))
                        {
                            // This producing thread may block and there is a pending producer available
                            // so we use EPC: the producer will directly consume the task and then
                            // race with the pending producer to take over production.
                            _pending = true;
                            _state = State.IDLE;
                            mode = Mode.EXECUTE_PRODUCE_CONSUME;
                            break;
                        }
                    }
                }

                // otherwise use PIC to directly consume the task and then resume production.
                mode = Mode.PRODUCE_INVOKE_CONSUME;
                break;

            default:
                throw new IllegalStateException();
        }
```

gregw: I've moved on past this alternative.

gregw:
@sbordet So I have  re-re-re-reworked this class so:
 + The documentation is in the format you asked for
 + The actual code now actually matches the algorithm as described in the documentation
 + It has lots of good comments everywhere, but hopefully none that don't add value
 + the decomposition of the algorithm into methods has changed bit, giving smaller more meaningful methods (eg `selectSubStrategy`). Hopefully these will inlined to the same result anyway.
 + The only substantial change to the algorithm is that by first checking the taskType, then we can avoid a needless check on the thread type for PC tasks.

Considering the effort expended on a "simple" rename, please give this version fair consideration before just saying: NAH let's keep the old one

Note that I think there is some scope to manually unwind the production loop so that there is a blocking and a nonblocking version.    I did give that a try, but code clarity suffered a lot and I think there would only be very marginal gains if any.   The evidence is that the majority of productions are done with blocking invocations, so that path should be well optimized by the JIT without manual unwinding.

lachlan-roberts (dismissed):

gregw:
@sbordet nudge... let's get this merged long before next release train

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -0,0 +1,564 @@
...
+
+            switch (_state)
+            {
+                case IDLE:
+                    // The strategy was IDLE, so this thread can become the producer.
+                    _state = State.PRODUCING;
+                    break;
+
+                case PRODUCING:
+                    // The strategy is already producing, so another thread must be the producer
```
sbordet: Nit: add period at the end.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -0,0 +1,564 @@
...
+                LOG.warn("Unable to produce", th);
+            }
+        }
+    }
+
+    /**
+     * Select the execution strategy.
+     * @param task The task to select the strategy for.
+     * @param nonBlocking True if the producing thread cannot block.
+     * @return The sub-strategy mode to use for the task.
```
sbordet: Remove "mode".

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -0,0 +1,564 @@
...
+
+                // Otherwise use PEC: the task is consumed by the executor and this thread continues to produce.
+                return SubStrategy.PRODUCE_EXECUTE_CONSUME;
+
+            default:
+                throw new IllegalStateException(String.format("taskType=%s %s", taskType, this));
+        }
+    }
+
+    /** Consume a task with a sub-strategy.
```
sbordet: Start javadoc on the next line like the other javadocs.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -0,0 +1,564 @@
...
+     * @param task The task to consume.
+     * @param subStrategy The execution sub-strategy mode to use to consume it.
+     * @return True if the sub-strategy requires the caller to continue to produce tasks.
+     */
+    private boolean consumeTask(Runnable task, SubStrategy subStrategy)
+    {
+        // Consume and/or execute task according to the selected mode.
+        if (LOG.isDebugEnabled())
+            LOG.debug("{} m={} t={}/{}", this, subStrategy, task, Invocable.getInvocationType(task));
```
sbordet: `m=` should be `ss=` to be consistent with the `IllegalStateException` message.
As `this` may be quite long, I would put it as last, so the log line reads the subStrategy, the task, and then the long `this.toString()`.

sbordet (dismissed):
Looks good apart niggles in the comment/javadocs.

Since many javadocs were added, a general rule for javadocs is that they should be written in 3rd person, therefore "Selects a sub-strategy" rather than "Select a sub-strategy", "Runs" rather than "Run", etc.
See:
https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#styleguide
https://blog.joda.org/2012/11/javadoc-coding-standards.html

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
gregw: All the other debugs in this class start with logging this, so keeping it the same makes it easier to see differences.

lachlan-roberts (dismissed):

sbordet (dismissed):
Not all javadocs were put in 3rd person, it's just to have a little more consistency, but I don't want to annoy too much. LGTM.

gregw:
@sbordet nudge

sbordet:
@gregw I made further changes to the javadocs, see my commit.

sbordet (approved):

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -345,16 +348,17 @@ private SubStrategy selectSubStrategy(Runnable task, boolean nonBlocking)
...
-     * @param subStrategy The execution sub-strategy mode to use to consume it.
+     * @param subStrategy The execution sub-strategy to use to consume the task.
      * @return True if the sub-strategy requires the caller to continue to produce tasks.
      */
     private boolean consumeTask(Runnable task, SubStrategy subStrategy)
     {
         // Consume and/or execute task according to the selected mode.
         if (LOG.isDebugEnabled())
-            LOG.debug("{} ss={} t={}/{} {}", this, subStrategy, task, Invocable.getInvocationType(task));
+            LOG.debug("ss={} t={}/{} {}", subStrategy, task, Invocable.getInvocationType(task), this);
```
gregw, line 361: @sbordet this is now inconsistent with the other debugs in the class.... but I guess that is not so important.

PR #6355, opened 2021-06-03
Fixes #6323 - HttpClient requests with redirects gets stuck/never cal…

### LINKED ISSUE #6323: HttpClient gets stuck/never calls onComplete() when multiple requests with timeouts are sent

sbordet:
…ls onComplete()

* Reworked the total timeout handling.
* Now a CyclicTimeouts handles the exchanges in each HttpDestination,
and a CyclicTimeouts handles the exchanges in each HttpConnection
(rather than in HttpChannel).
* Now adjusting the total timeout for copied requests generated by
redirects and authentication.

(cherry picked from commit 2e7d17400f44097b95f2110ba5b595ec90fbe03f)

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/AuthenticationProtocolHandler.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/AuthenticationProtocolHandler.java
@@ -19,6 +19,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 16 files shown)


sbordet:
Forward port of #6334.

gregw (approved):

PR #6356, opened 2021-06-03
Fixes #6354 - allow jetty bundles to use 1.7 slf4j

### LINKED ISSUE #6354: org.slf4j dependency imports osgi packages at 2.0
**Jetty version**
10.0.x

All

**OS type/version**
All

**Description**
The jetty JARs are packaged as OSGi bundles.  The ones that require slf4j have an Import-Package header in their META-INF/MANIFEST.MF that have the following specified:

```
Import-Package: ... org.slf4j;version="[2.0.0,3)"
```

When using such JARs as bundles in an OSGi framework this forces the installation of a the slf4j 2.0 (which currently is only available as alpha) in order to resolve the jetty bundles.  I've been told jetty should work with the current stable released versions of slf4j (version 1.x).  If this is true then the imports should be updated to include the minimum version jetty must have.

Also, if slf4j is entirely optional then the imports should be declared as optional (`resolution:=optional`)

tjwatson:


joakime:
Seems pretty straight forward, I'm going to wait to see if the osgi build/test portions of CI pass.

joakime (approved):
CI is happy.
And some local testing of this PR shows that this change seems to propagate to child modules nicely. (manually cracked open a few select jars and looked at the generated manifest)

PR #6357, opened 2021-06-04
Issue #4772 - support partial messages for Jetty WS API annotations

### LINKED ISSUE #4772: Jetty WebSocket API onMessage annotation does not support partial messages.
**Jetty version**
10.0.x

**Description**
The Jetty WebSocket API `@OnWebSocketMessage` annotation does not support partial messages, while the Javax API does support this. To use partial messages you must implement `WebSocketPartialListener`.

We should consider adding support for partial messages for the `OnWebSocketMessage` annotation in the jetty ws API.

lachlan-roberts:
## Closes #4772.

Adds support for partial messages to the Jetty WebSocket API `@OnWebSocketMessage` annotation.

### DIFF
--- a/jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/RemoteEndpoint.java
+++ b/jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/RemoteEndpoint.java
@@ -103,8 +103,9 @@ public interface RemoteEndpoint
      * @param fragment the text being sent
      * @param isLast true if this is the last piece of the partial bytes
      * @param callback callback to notify of success or failure of the write operation
+     * @throws IOException this never throws IOException, it was a mistake to have this in the signature.
      */
-    void sendPartialString(String fragment, boolean isLast, WriteCallback callback);
+    void sendPartialString(String fragment, boolean isLast, WriteCallback callback) throws IOException;
 
     /**
      * Send a Ping message containing the given application data to the remote endpoint, blocking until all of the
--- a/jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/RemoteEndpoint.java
+++ b/jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/RemoteEndpoint.java
@@ -104,7 +104,7 @@ public interface RemoteEndpoint
      * @param isLast true if this is the last piece of the partial bytes
      * @param callback callback to notify of success or failure of the write operation
      */
-    void sendPartialString(String fragment, boolean isLast, WriteCallback callback) throws IOException;
+    void sendPartialString(String fragment, boolean isLast, WriteCallback callback);
 
     /**
      * Send a Ping message containing the given application data to the remote endpoint, blocking until all of the
--- a/jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/annotations/OnWebSocketMessage.java
+++ b/jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/annotations/OnWebSocketMessage.java
@@ -30,30 +30,28 @@ import org.eclipse.jetty.websocket.api.Session;
  * <p>
  * <u>Text Message Versions</u>
  * <ol>
- * <li><code>public void methodName(String text)</code></li>
- * <li><code>public void methodName({@link Session} session, String text)</code></li>
- * <li><code>public void methodName(Reader reader)</code></li>
- * <li><code>public void methodName({@link Session} session, Reader reader)</code></li>
+ * <li>{@code public void methodName(String text)}</li>
+ * <li>{@code public void methodName(Session session, String text)}</li>
+ * <li>{@code public void methodName(Reader reader)}</li>
+ * <li>{@code public void methodName(Session session, Reader reader)}</li>
  * </ol>
  * <p>Note: that the {@link Reader} in this case will always use UTF-8 encoding/charset (this is dictated by the RFC 6455 spec for Text Messages. If you need to
- * use a non-UTF-8 encoding/charset, you are instructed to use the binary messaging techniques.</p><br>
+ * use a non-UTF-8 encoding/charset, you are instructed to use the binary messaging techniques.</p>
  * <u>Binary Message Versions</u>
  * <ol>
- * <li><code>public void methodName(ByteBuffer message)</code></li>
- * <li><code>public void methodName({@link Session} session, ByteBuffer message)</code></li>
- * <li><code>public void methodName(byte buf[], int offset, int length)</code></li>
- * <li><code>public void methodName({@link Session} session, byte buf[], int offset, int length)</code></li>
- * <li><code>public void methodName(InputStream stream)</code></li>
- * <li><code>public void methodName({@link Session} session, InputStream stream)</code></li>
+ * <li>{@code public void methodName(ByteBuffer message)}</li>
+ * <li>{@code public void methodName(Session session, ByteBuffer message)}</li>
+ * <li>{@code public void methodName(byte[] buf, int offset, int length)}</li>
+ * <li>{@code public void methodName(Session session, byte[] buf, int offset, int length)}</li>
+ * <li>{@code public void methodName(InputStream stream)}</li>
+ * <li>{@code public void methodName(Session session, InputStream stream)}</li>
  * </ol>
  * <u>Partial Message Variations</u>
  * <p>These are used to receive partial messages without aggregating them into a complete WebSocket message. Instead the a boolean
- * argument is supplied to indicate whether this is the last segment of data of the message. This can be done most efficiently
- * for the {@link java.nio.ByteBuffer} type as no copying needs to occur to get the data into this format.</p>
+ * argument is supplied to indicate whether this is the last segment of data of the message.</p>
  * <ol>
- * <li><code>public void methodName(ByteBuffer payload, boolean last)</code></li>
- * <li><code>public void methodName(byte payload, boolean last)</code></li>
- * <li><code>public void methodName(String payload, boolean last)</code></li>
+ * <li>{@code public void methodName(ByteBuffer payload, boolean last)}</li>
+ * <li>{@code public void methodName(String payload, boolean last)}</li>
  * </ol>
  * <p>Note: Similar to the signatures above these can all be used with an optional first {@link Session} parameter.</p>
  */
--- a/jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/annotations/OnWebSocketMessage.java
+++ b/jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/annotations/OnWebSocketMessage.java
@@ -35,9 +35,8 @@ import org.eclipse.jetty.websocket.api.Session;
  * <li><code>public void methodName(Reader reader)</code></li>
  * <li><code>public void methodName({@link Session} session, Reader reader)</code></li>
  * </ol>
- * Note: that the {@link Reader} in this case will always use UTF-8 encoding/charset (this is dictated by the RFC 6455 spec for Text Messages. If you need to
- * use a non-UTF-8 encoding/charset, you are instructed to use the binary messaging techniques.
- * <p>
+ * <p>Note: that the {@link Reader} in this case will always use UTF-8 encoding/charset (this is dictated by the RFC 6455 spec for Text Messages. If you need to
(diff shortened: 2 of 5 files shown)


jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/RemoteEndpoint.java
```diff
@@ -104,7 +104,7 @@
      * @param isLast true if this is the last piece of the partial bytes
      * @param callback callback to notify of success or failure of the write operation
      */
-    void sendPartialString(String fragment, boolean isLast, WriteCallback callback) throws IOException;
+    void sendPartialString(String fragment, boolean isLast, WriteCallback callback);
```
lachlan-roberts: This was wrong to have the `throws IOException` as this takes a callback, but I'm not sure if there will be any issues with changing it now that its already there.

sbordet: @lachlan-roberts no that will break applications that wrapped `sendPartialString()` into a `try/catch`.
You have to put `throws IOException` back. Perhaps add javadocs that say that it was a mistake and it's never thrown.

jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/annotations/OnWebSocketMessage.java
```diff
@@ -35,9 +35,8 @@
  * <li><code>public void methodName(Reader reader)</code></li>
  * <li><code>public void methodName({@link Session} session, Reader reader)</code></li>
  * </ol>
- * Note: that the {@link Reader} in this case will always use UTF-8 encoding/charset (this is dictated by the RFC 6455 spec for Text Messages. If you need to
- * use a non-UTF-8 encoding/charset, you are instructed to use the binary messaging techniques.
- * <p>
+ * <p>Note: that the {@link Reader} in this case will always use UTF-8 encoding/charset (this is dictated by the RFC 6455 spec for Text Messages. If you need to
+ * use a non-UTF-8 encoding/charset, you are instructed to use the binary messaging techniques.</p><br>
```
sbordet: Remove `<br>` at the end of line.

jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/annotations/OnWebSocketMessage.java
```diff
@@ -47,6 +46,16 @@
  * <li><code>public void methodName(InputStream stream)</code></li>
  * <li><code>public void methodName({@link Session} session, InputStream stream)</code></li>
  * </ol>
+ * <u>Partial Message Variations</u>
+ * <p>These are used to receive partial messages without aggregating them into a complete WebSocket message. Instead the a boolean
+ * argument is supplied to indicate whether this is the last segment of data of the message. This can be done most efficiently
+ * for the {@link java.nio.ByteBuffer} type as no copying needs to occur to get the data into this format.</p>
+ * <ol>
+ * <li><code>public void methodName(ByteBuffer payload, boolean last)</code></li>
+ * <li><code>public void methodName(byte payload, boolean last)</code></li>
```
sbordet: Isn't this `byte[]` in the signature?
Frankly, I would not offer a `byte[]` variant.

Also, what do you do with the `ByteBuffer` version if it is retained by the application, but not consumed? I think the signature needs a `Callback` to signal when the `ByteBuffer` can be released, no?

If you choose for no `Callback`, then you have to remove from the javadocs above "as no copying needs to occur to get the data", and specify precisely that applications *must* consume the buffer before returning from the method.

jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/annotations/OnWebSocketMessage.java
```diff
@@ -35,9 +35,8 @@
  * <li><code>public void methodName(Reader reader)</code></li>
```
sbordet: Use `{@code ...}` instead of `<code>`.

jetty-websocket/websocket-jetty-common/src/main/java/org/eclipse/jetty/websocket/common/JettyWebSocketFrameHandlerFactory.java
```diff
@@ -78,6 +79,51 @@
...
+        new InvokerUtils.Arg(ByteBuffer.class).required(),
+        new InvokerUtils.Arg(boolean.class).required()
+    };
+
+    private static final InvokerUtils.Arg[] binaryPartialArrayCallingArgs = new InvokerUtils.Arg[]{
+        new InvokerUtils.Arg(Session.class),
+        new InvokerUtils.Arg(byte[].class).required(),
+        new InvokerUtils.Arg(boolean.class).required()
+    };
+
```
sbordet, line 125: I'd drop the `byte[]` ones.

jetty-websocket/websocket-jetty-tests/src/test/java/org/eclipse/jetty/websocket/tests/AnnotatedPartialListenerTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+        segment = Objects.requireNonNull(endpoint.messages.poll(5, TimeUnit.SECONDS));
+        assertThat(segment.bytes, is("o w".getBytes()));
+        assertThat(segment.last, is(false));
+
+        segment = Objects.requireNonNull(endpoint.messages.poll(5, TimeUnit.SECONDS));
+        assertThat(segment.bytes, is("orld".getBytes()));
+        assertThat(segment.last, is(true));
+    }
+}
```
sbordet, line 240: Please add a test that fails deployment if there are 2 `@OnWebSocketMessage`, one for full message, one for partial message.
It's probably covered elsewhere, but since this is a new feature, it's worth making sure the new feature is covered, so maybe you need to add/expand the tests where multiple annotations are tested.

sbordet (changes requested):

jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/annotations/OnWebSocketMessage.java
lachlan-roberts: Sure, but I added this because its harder to read without it because there's no gap between this note and the "Binary Message Versions" header.

jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/annotations/OnWebSocketMessage.java
lachlan-roberts: > Also, what do you do with the ByteBuffer version if it is retained by the application, but not consumed? I think the signature needs a Callback to signal when the ByteBuffer can be released, no?

The websocket-core API has this feature but for the Jetty and Javax APIs they don't take callbacks, so you only own the `ByteBuffer` only until the method returns. If you wish to retain the `ByteBuffer` you must copy it.

> If you choose for no Callback, then you have to remove from the javadocs above "as no copying needs to occur to get the data"

By this I meant that `ByteBuffer` can be passed into the method directly and doesn't need to be copied by the implementation, but this is not the case for `byte[]` or `String` and there will be some copying involved to get it into that format. I didn't mean that you can retain the `ByteBuffer` indefinitely without copying it.

> and specify precisely that applications must consume the buffer before returning from the method.

They don't have to consume the buffer, if they want to ignore the rest of the message they don't need to consume it. And this is no different to the whole message signatures.

I have removed the `byte[]` option and this part of the Javadoc so it is less confusing.

jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/annotations/OnWebSocketMessage.java
lachlan-roberts: I can do this but I will need to remove the linking to the session with `{@link Session}`.

jetty-websocket/websocket-jetty-common/src/main/java/org/eclipse/jetty/websocket/common/JettyWebSocketFrameHandlerFactory.java
lachlan-roberts: Done.

jetty-websocket/websocket-jetty-tests/src/test/java/org/eclipse/jetty/websocket/tests/AnnotatedPartialListenerTest.java
lachlan-roberts: Done.

jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/annotations/OnWebSocketMessage.java
sbordet: > By this I meant that ByteBuffer can be passed into the method directly and doesn't need to be copied by the implementation, but this is not the case for byte[] or String and there will be some copying involved to get it into that format. I didn't mean that you can retain the ByteBuffer indefinitely without copying it.

This is the javadoc for users -- the implementation details should be left out.
As a user, though, I must know if I have to copy the buffer (assuming interest in the content) before returning from the method.

jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/annotations/OnWebSocketMessage.java
```diff
@@ -30,23 +30,30 @@
...
+ * <li>{@code public void methodName(Session session, InputStream stream)}</li>
+ * </ol>
+ * <u>Partial Message Variations</u>
+ * <p>These are used to receive partial messages without aggregating them into a complete WebSocket message. Instead the a boolean
+ * argument is supplied to indicate whether this is the last segment of data of the message.</p>
+ * <ol>
+ * <li>{@code public void methodName(ByteBuffer payload, boolean last)}</li>
+ * <li>{@code public void methodName(String payload, boolean last)}</li>
  * </ol>
+ * <p>Note: Similar to the signatures above these can all be used with an optional first {@link Session} parameter.</p>
```
sbordet, line 58: I think you should add a note about split UTF-8 characters... I'm guessing the `payload` always contains full UTF-8 sequences and you keep split UTF-8 bytes around for the next call?

jetty-websocket/websocket-jetty-tests/src/test/java/org/eclipse/jetty/websocket/tests/AnnotatedPartialListenerTest.java
```diff
@@ -0,0 +1,240 @@
...
+    @Test
+    public void testDoubleOnMessageAnnotation()
+    {
+        InvalidDoubleBinaryListener doubleBinaryListener = new InvalidDoubleBinaryListener();
+        assertThrows(InvalidWebSocketException.class, () -> client.connect(doubleBinaryListener, serverUri));
+
+        InvalidDoubleTextListener doubleTextListener = new InvalidDoubleTextListener();
+        assertThrows(InvalidWebSocketException.class, () -> client.connect(doubleTextListener, serverUri));
+    }
+}
```
sbordet, line 240: Can you please add one more test for split UTF-8 bytes?

sbordet (changes requested):

jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/annotations/OnWebSocketMessage.java
joakime: Also, what happens if the message becomes bad during a partial message delivery.
Eg: first 2 calls to `methodName(String, boolean)` are fine, but the 3rd call cannot be made as the ws-core found the message to be in violation of the websocket messaging rules.

jetty-websocket/websocket-jetty-tests/src/test/java/org/eclipse/jetty/websocket/tests/AnnotatedPartialListenerTest.java
joakime: and late bad message / utf-8 as well.

1. good partial sequence
2. good partial sequence
3. bad partial utf-8 sequence

sbordet (approved):

PR #6362, opened 2021-06-04
Fixes #6330 - CustomRequestLog is missing HTTP version format option.

### LINKED ISSUE #6330: CustomRequestLog is missing HTTP version format option
**Jetty version**
9.4.x

**Description**
If someone wants to move away from CustomRequestLog NCSA format to one that is slightly tweaked but still includes HTTP version (as seen in the NCSA format), then they have to drop the HTTP version portion.

To explain, this started with the stackoverflow question at https://stackoverflow.com/questions/67635098/request-param-is-logged-in-access-log-with-embedded-jetty-server-of-spring-boot

They were using the old-school extended NCSA requestlog and format.
Which outputted the first line in the request `POST /myAPI/v2/customer/message?myID=123&messageText=hello HTTP/1.0`, but they wanted to strip away the query string.

The default format for Extended NCSA in CustomRequestLog is `%{client}a - %u %t "%r" %s %O  "%{Referer}i" "%{User-Agent}i"`, of which the `"%r"` portion logs the first line of the request, if they wanted to simulate the Extended NCSA format, but without the query string, they cannot, as HTTP Version is not an available format option, the best they can do is `"%m %U"` without the HTTP Version portion.

sbordet:
Simplified javadoc.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/CustomRequestLog.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/CustomRequestLog.java
@@ -152,9 +152,7 @@ import static java.lang.invoke.MethodType.methodType;
  *
  * <tr>
  * <td valign="top">%H</td>
- * <td>Returns the name and version of the protocol the request uses in the form
- * protocol/majorVersion.minorVersion, for example, HTTP/1.1. For HTTP servlets,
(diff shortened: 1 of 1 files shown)


sbordet:
Avoid to use text copied from other sources.

PR #6380, opened 2021-06-09
9.4.x: Improve ByteBufferPool implementations

### LINKED ISSUE #6379: Reduce contention in all `ByteBufferPool` implementations
**Target Jetty version(s)**
9.4.x, 10.0.x, 11.0.x

**Enhancement Description**

All `ByteBufferPool` implementations use `ByteBufferPool.Bucket`s to track `ByteBuffer`s. The current implementation of `Bucket` stores buffers in a `ConcurrentLinkedDeque`, and calls `poll()` to acquire a pooled buffer and `offerFirst()` to release a buffer.

This means that all the contention is done at the head of the queue for no particular reason. An easy way to reduce that contention would be to release buffers at the tail of the queue by calling `offer()` instead of `offerFirst()`, this way all the acquisition contention would happen at the head while all the releasing contention would happen at the tail, basically halving the contention.

Replacing the `ConcurrentLinkedDeque` with a `ConcurrentLinkedQueue` could also improve performance an extra bit too.

lorban:
I also added a simple benchmark that shows the results of this improvement.

Original code:

```
Benchmark                                         Mode  Cnt        Score        Error  Units
ArrayByteBufferPoolBenchmark.testAcquireRelease  thrpt    3  2527727,245 ± 389428,504  ops/s
```

Just replacing `offerFirst` with `offer`, keeping the `ConcurrentLinkedDeque`:

```
Benchmark                                         Mode  Cnt        Score       Error  Units
ArrayByteBufferPoolBenchmark.testAcquireRelease  thrpt    3  3401758,446 ± 66958,678  ops/s
```

Replacing `offerFirst` with `offer` and replacing the `ConcurrentLinkedDeque` with a `ConcurrentLinkedQueue`:

```
Benchmark                                         Mode  Cnt        Score        Error  Units
ArrayByteBufferPoolBenchmark.testAcquireRelease  thrpt    3  4302078,532 ± 108538,319  ops/s
```

### DIFF
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferPool.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferPool.java
@@ -20,9 +20,9 @@ package org.eclipse.jetty.io;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
-import java.util.Deque;
 import java.util.List;
-import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Consumer;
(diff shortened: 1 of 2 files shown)


sbordet (dismissed):

gregw (dismissed):
Perhaps say a bit more about performance improvements.... or link to blog:)

sbordet (approved):

PR #6384, opened 2021-06-10
Release process issue template

gregw:
Removed most per release tick boxes
Delegated testing to project tick-list

### DIFF
--- a/.github/ISSUE_TEMPLATE/release-template.md
+++ b/.github/ISSUE_TEMPLATE/release-template.md
@@ -19,14 +19,15 @@ This release process will produce releases:
 - [ ] Link this issue to the target [project(s)](https://github.com/eclipse/jetty.project/projects).
 - [ ] Assign this issue to a "release manager".
 - [ ] Review [draft security advisories](https://github.com/eclipse/jetty.project/security/advisories). Ensure that issues are created and assigned to GitHub Projects to capture any advisories that will be announced.
-- [ ] Review their issues/PRs assigned to the target [project(s)](https://github.com/eclipse/jetty.project/projects).  Any PRs that are moved to subsequent releases should be commented on so their authors are informed.
+- [ ] Review the issues/PRs assigned to the target [project(s)](https://github.com/eclipse/jetty.project/projects).  Any PRs that are moved to subsequent releases should be commented on so their authors are informed.
 - [ ] Freeze the target [project(s)](https://github.com/eclipse/jetty.project/projects) by editing their names to "Jetty X.Y.Z FROZEN"
 - [ ] Wait 24 hours from last change to the issues/PRs included in a FROZEN GitHub Project.
 - [ ] Verify target [project(s)](https://github.com/eclipse/jetty.project/projects) are complete.
 - [ ] Verify that branch `jetty-10.0.x` is merged to branch `jetty-11.0.x`.
 - [ ] Assign issue to "build manager", who will stage the releases.
    + [ ] Ensure `VERSION.txt` additions for each release will be meaningful, descriptive, correct text.
-   + [ ] Stage Release(s).
+   + [ ] Stage 9.4 release with Java 11.
+   + [ ] Stage 10/11 release(s) with Java 16. 
    + [ ] Edit a draft release (for each Jetty release) in GitHub (https://github.com/eclipse/jetty.project/releases). Content is generated with the "changelog tool".
 - [ ] Assign issue to "test manager", who will oversee the testing of the staged releases.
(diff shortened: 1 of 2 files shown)


gregw:
I've recreated the issue template using @sbordet most recent tick list as a basis.  However, I have tried to take on @joakime 's concerns by spinning out testing at least to a per-project tick list. 

thoughts?

Note that I would like to get something like this merged sooner rather than later, ie it is better to have some process rather than forever debate the perfect process.

.github/ISSUE_TEMPLATE/release-template.md
```diff
@@ -0,0 +1,49 @@
...
+**Target Date:**
+
+**Tasks:**
+- [x] Create the release(s) issue.
+- [ ] Update the target Jetty version(s) in the issue.  
+- [ ] Update the target release date in the issue.
+- [ ] Link this issue to the target [project(s)](https://github.com/eclipse/jetty.project/projects).
+- [ ] Assign this issue to a "release manager".
+- [ ] Review [draft security advisories](https://github.com/eclipse/jetty.project/security/advisories). Ensure that issues are created and assigned to GitHub Projects to capture any advisories that will be announced.
+- [ ] Review their issues/PRs assigned to the target [project(s)](https://github.com/eclipse/jetty.project/projects).  Any PRs that are moved to subsequent releases should be commented on so their authors are informed.
```
sbordet: Typo; `their` -> `the`.

.github/ISSUE_TEMPLATE/release-template.md
```diff
@@ -0,0 +1,49 @@
...
+- [ ] Freeze the target [project(s)](https://github.com/eclipse/jetty.project/projects) by editing their names to "Jetty X.Y.Z FROZEN"
+- [ ] Wait 24 hours from last change to the issues/PRs included in a FROZEN GitHub Project.
+- [ ] Verify target [project(s)](https://github.com/eclipse/jetty.project/projects) are complete.
+- [ ] Verify that branch `jetty-10.0.x` is merged to branch `jetty-11.0.x`.
+- [ ] Assign issue to "build manager", who will stage the releases.
+   + [ ] Ensure `VERSION.txt` additions for each release will be meaningful, descriptive, correct text.
+   + [ ] Stage Release(s).
```
sbordet: This misses the bit that Jetty 9.4.x should be built with Java 11. Why that must be so, I have no idea. @joakime?

.github/ISSUE_TEMPLATE/release-template.md
```diff
@@ -0,0 +1,49 @@
...
+   + [ ] Edit a draft release (for each Jetty release) in GitHub (https://github.com/eclipse/jetty.project/releases). Content is generated with the "changelog tool".
+- [ ] Assign issue to "test manager", who will oversee the testing of the staged releases.
+   + [ ] Add [testing tick list](https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/.github/test-ticklist.md) to the target [project(s)](https://github.com/eclipse/jetty.project/projects).
```
sbordet, line 34: The testing tick list is the same for all Jetty versions. I fail to see why needs to be split into different projects.
It's the same as "Stage releases" which is here, not split per-project.
IMHO there should be just 1 tick list to keep things simpler -- one place to look only.

sbordet (changes requested):

janbartel:
@sbordet can you add your feedback after using this template for the last release?

@gregw I don't see anything in here about updating the documentation?

janbartel (commented):
Shouldn't there be a step in here about updating the documentation?

Does the eclipse downloads page also need to be manually edited or not?

gregw:
I added sub points to the update website bullet

.github/ISSUE_TEMPLATE/release-template.md
gregw: I'm on the fence with this one.   I can see @sbordet's point that it makes us have to look in multiple places for status.  I can also see that if we had to reroll just one release it would be easier to track iterations.   But then how likely is a single release re-roll?  Typically we will need to reroll all the releases so a single testing summary checklist would work.   Others need to comment else I think we will revert to a single list.

sbordet:
Having a separate bullet list for the project is IMHO not valuable.

Once all the issues have been assigned to the project and in state "done" the project is not looked at anymore. This happens early in the bullet list, while testing is much later.

Having to do back and forth between 2 web UI items for the same, conceptually, list of things to do brings IMHO unnecessary complexity.

PR #6388, opened 2021-06-10
Fixes #6382 - HttpClient TimeoutException message reports transient v…

### LINKED ISSUE #6382: HttpClient TimeoutException message reports transient values
**Jetty version(s)**
9.4.42

Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.7+8-LTS, mixed mode)

**OS type/version**
OSX 10.15.4

**Description**
When using the HttpClient to make asynchronous requests with a specific timeout, if the request has follow redirects enabled and the request ends up timing out; the error reported includes a timeout value that is different from the one specified on the request itself.  This did not happen in older versions of Jetty.  Example, you set a timeout of 5000ms on a request that ends up getting redirected and ultimately times out.  The error message you'll receive will look like:
```
Total timeout 3808 ms elapsed
```

rather than:
```
Total timeout 5000 ms elapsed
```

**How to reproduce?**
You can run the test classes attached to #6323 and add a call to `result.getFailure().getMessage()` in the `onComplete()` handler near the bottom of the test class.

sbordet:
…alues.

Now using consistently HttpConversation.getTimeout() to report the accurate value.

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/HttpConnection.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/HttpConnection.java
@@ -331,7 +331,7 @@ public abstract class HttpConnection implements Connection, Attachable
             if (exchange != null)
             {
                 HttpRequest request = exchange.getRequest();
-                request.abort(new TimeoutException("Total timeout " + request.getTimeout() + " ms elapsed"));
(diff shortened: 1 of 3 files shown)


lorban (approved):
LGTM

gregw:
comment to wake up CI???

Issue #6396, opened 2021-06-11
Threads created by Jetty no longer run as the Subject that Jetty was started as


stoty:
**Jetty version**
9.4.37.v20210219 and later 9.4.x versions

**Java version/vendor**
openjdk version "1.8.0_292"
OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_292-b10)
OpenJDK 64-Bit Server VM (AdoptOpenJDK)(build 25.292-b10, mixed mode)

**OS type/version**
Darwin 20.5.0 Darwin Kernel Version 20.5.0: Sat May  8 05:10:33 PDT 2021; root:xnu-7195.121.3~9/RELEASE_X86_64

**Description**
The thread creation change in #5859 has caused a regression.
The threads created by Jetty no longer run as the Subject that Jetty was started as.
This was discovered when trying to build Avatica with Jetty 9.4.37.v20210219 
See https://issues.apache.org/jira/browse/CALCITE-4646

**How to reproduce**
* Check out calclite-avatica HEAD
* Update jetty.version to 9.4.37.v20210219 or later in gradle.properties
* run `./gradlew clean test`

janbartel:
We will look into this, but in the meanwhile is there any other information you can supply? Is the thread a request handling-handling thread or has another thread been spun up during request handling? Are you using jetty's SPNEGO integration?

janbartel:
Also, please can you provide some links as to which of your tests that are failing as we're not familiar with your codebase so some navigation help would be useful, as would a stack trace or two.

janbartel:
Just followed the repro steps (I used jetty-9.4.42) and whilst there is a lot of test output, I can't see anything that is an obvious error and no output that mentions SPNEGO.

**Edit:**  Ah! Sorry, I downloaded calcite-master instead of calcite-avatica-master!

stoty:
Yes, this is a request handling thread. Avatica server does not create threads on its own.

Avatica SPNEGO implementation is using Jetty's SPNEGO integration, but the some of the Jetty classes are inherited and extended to work around some bugs/limitations in Jetty's SPENGO handling. 

This is where SPNEGO gets configured in Avatica:
https://github.com/apache/calcite-avatica/blob/e4f22c33c5a67257056ee4428486b900cc139038/server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java#L334

The failing tests are:
[...]

PR #6397, opened 2021-06-11
Jetty 9.4.x 6396 Threads created by Jetty no longer run as the Subject that Jetty was started as #6396

stoty:


jetty-util/src/main/java/org/eclipse/jetty/util/thread/PrivilegedThreadFactory.java
```diff
@@ -44,12 +45,24 @@
      */
     static <T extends Thread> T newThread(Supplier<T> newThreadSupplier)
     {
+        // doPriviliged resets the subject
+        // We need to save the subject, and run a subjext.doAs from inside
+        // doPrivileged to restore it.
+        // see https://www.ibm.com/docs/en/was-zos/8.5.5?topic=service-java-authentication-authorization-authorization
+        Subject subject = Subject.getSubject(AccessController.getContext());
```
janbartel, line 52: I think this is actually a leak of the `Subject` from the calling thread, no? In your particular case you actually _do_ want to inherit the `Subject` because you have one `Subject` that represents the `server`, but in the general case I don't think we want jetty to inherit anything from from the calling thread.  Let me know if my assumption here is flawed.

I was thinking more along the lines of allowing a ThreadFactory to be set on the `QueuedThreadPool`, rather than always using the `PrivilegedThreadFactory`, in that way you can customize your thread inheritance settings as you like.

janbartel (changes requested):

jetty-util/src/main/java/org/eclipse/jetty/util/thread/PrivilegedThreadFactory.java
stoty: In my mind the previous behaviour, where the Jetty threads inherited the active subject is the intuitve one, I'd expect the web server threads to run with the same Subject that it was started as.
Others may have other expectations and requirements.

I have tried overriding QueuedThreadPool, but it is not possible because of the private fields, so I ended up copying and renaming the whole class.
Defining a new QueuedThreadPool and using the modified ThreadFactory in it did solve the problems, the tests are running fine.

Being able to set the threadFactory on QueuedThreadPool is a good solution (probably for other purposes too), and it would be a clean and safe way to solve our problem.

stoty:
Thanks for very quick response and the useful advice @janbartel

stoty:
This can be worked around buy using a custom ThreadFactory as suggested.

PR #6398, opened 2021-06-11
Fixed #6375 XmlConfiguration always checks for setter method

### LINKED ISSUE #6375: Always check XML `Set` elements with `property` attribute
**Target Jetty version(s)**

>= 10.0.x

**Enhancement Description**

Since jetty-10, the XML `Set` element could have a property attribute that only does the set if the property is set.  The problem with this is that the code may change and we do not detect that the XML is now wrong unless somebody attempts to set that property.   The `XmlConfiguration` class should be change to always check that at least one setter exists for a `Set` element with a `property` attribute.

gregw:
Fix #6375 by making XmlConfiguration Set handling always check for a matching setter, even if the property attribute is given but not set.

### DIFF
--- a/jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlConfiguration.java
+++ b/jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlConfiguration.java
@@ -515,8 +515,8 @@ public class XmlConfiguration
          */
         private void set(Object obj, XmlParser.Node node) throws Exception
         {
-            String attr = node.getAttribute("name");
-            String name = "set" + attr.substring(0, 1).toUpperCase(Locale.ENGLISH) + attr.substring(1);
+            String name = node.getAttribute("name");
+            String setter = "set" + name.substring(0, 1).toUpperCase(Locale.ENGLISH) + name.substring(1);
             String id = node.getAttribute("id");
             String property = node.getAttribute("property");
             String propertyValue = null;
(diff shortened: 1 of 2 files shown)


jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlConfiguration.java
```diff
@@ -516,31 +516,38 @@ public void configure(Object obj, XmlParser.Node cfg, int i) throws Exception
...
             {
                 Map<String, String> properties = _configuration.getProperties();
                 propertyValue = properties.get(property);
                 // If no property value, then do not set
                 if (propertyValue == null)
+                {
+                    // check that there is at least one setter that could have matched
+                    if (Arrays.stream(oClass.getMethods()).noneMatch(m -> m.getName().equals(name)))
+                        throw new NoSuchMethodException(String.format("No '%s' on %s", name, oClass.getName()));
+                    // otherwise it is a noop
```
janbartel, line 546: When this method is actually going to set a value, it checks for an exact setter, then a native method, then a field, and then finally other methods that might match.  That being the case, failing if there's no exact setter, even when no value is being set, misses the cases where it could have been a native or field match.

janbartel (changes requested):

jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlConfiguration.java
gregw: I've added a check for a public field.  All the different setter type are covered by the same name, so that is OK.

jetty-xml/src/test/java/org/eclipse/jetty/xml/XmlConfigurationTest.java
```diff
@@ -363,6 +363,32 @@ public void testSetWithNullPropertyAndValue() throws Exception
...
+        tc.setTestString("default");
+
+        NoSuchMethodException e = assertThrows(NoSuchMethodException.class, () -> configuration.configure(tc));
+        assertThat(e.getMessage(), containsString("setWrongName"));
+        assertEquals("default", tc.getTestString());
+    }
+    @Test
+
+    public void testSetWithWrongNameAndNullProperty() throws Exception
+    {
```
gregw:
```suggestion
    }
    
    @Test
    public void testSetWithWrongNameAndNullProperty() throws Exception
    {
```

janbartel (approved):
LGTM

PR #6409, opened 2021-06-15
Fix #6400 QueuedThreadPool always interrupts threads in stop

### LINKED ISSUE #6400: QueuedThreadPool interrupts pool threads when stopped with zero timeout
**Jetty version(s)**

9.4.42

openjdk version "1.8.0_282"

**OS type/version**

macOS 10.15

**Description**

When `QueuedThreadPool` is configured with a stop timeout of zero it skips trying to stop its threads. This leaves them running until they eventually end after 60s when their idle timeout elapses.

**How to reproduce?**

Running the following code will result in the JVM taking ~60 seconds to exit:

```java
import org.eclipse.jetty.util.thread.QueuedThreadPool;

public class QueuedThreadPoolProblem {

		QueuedThreadPool threadPool = new QueuedThreadPool(2);
		threadPool.setStopTimeout(0);
		threadPool.start();
		threadPool.stop();
	}

}
```

Increasing the timeout to 1ms allows the JVM to exit nigh-on immediately:

```java
import org.eclipse.jetty.util.thread.QueuedThreadPool;

public class QueuedThreadPoolProblem {

		QueuedThreadPool threadPool = new QueuedThreadPool(2);
		threadPool.setStopTimeout(1);
		threadPool.start();
		threadPool.stop();
	}

}
```

gregw:
Fixes #6400 Even if there is no timeout, always interrupt pool threads to attempt to stop them.

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
@@ -978,25 +978,19 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
             if (LOG.isDebugEnabled())
                 LOG.debug("Runner started for {}", QueuedThreadPool.this);
 
+            // We always start idle so the idle count is not adjust when we enter the loop below
+            // with no job.  The actual thread and idle counts have already been adjusted accounting
+            // for this newly started thread in #execute(Runnable)
             boolean idle = true;
             try
             {
                 Runnable job = null;
                 while (true)
                 {
-                    // If we had a job,
-                    if (job != null)
-                    {
-                        // signal that we are idle again
-                        if (!addCounts(0, 1))
-                            break;
-                        idle = true;
-                    }
-                    // else check we are still running
-                    else if (_counts.getHi() == Integer.MIN_VALUE)
-                    {
+                    // Adjust the idle count and check we are still running
+                    if (!addCounts(0, job == null ? 0 : 1))
                         break;
-                    }
+                    idle = true;
 
                     try
                     {
@@ -1027,6 +1021,9 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
                                 continue;
                         }
 
+                        // The actual idle count was already adjusted in #execute(Runnable)
+                        // Here we just remember we are not idle so that we don't adjust again
+                        // in finally below if we exit loop after running this job
                         idle = false;
 
                         // run job
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
@@ -234,30 +234,29 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
         // If stop timeout try to gracefully stop
         long timeout = getStopTimeout();
         BlockingQueue<Runnable> jobs = getQueue();
+
+        // Fill the job queue with noop jobs to wakeup idle threads.
+        for (int i = 0; i < threads; ++i)
+            jobs.offer(NOOP);
+
+        // If we have a timeout, try to let jobs complete naturally for half our stop time
+        if (timeout > 0)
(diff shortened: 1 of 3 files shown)


gregw:
I think it was an oversite that we don't at least interrupt threads.   We really wanted to stop threads in that situation, but that API is deprecated.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -234,30 +234,29 @@ protected void doStop() throws Exception
...
-            for (int i = 0; i < threads; ++i)
-            {
-                jobs.offer(NOOP);
-            }
 
-            // try to let jobs complete naturally for half our stop time
-            joinThreads(System.nanoTime() + TimeUnit.MILLISECONDS.toNanos(timeout) / 2);
+        // Fill the job queue with noop jobs to wakeup idle threads.
+        for (int i = 0; i < threads; ++i)
+            jobs.offer(NOOP);
```
sbordet, line 240: Since we are reviewing this, I don't think this loop guarantees that all threads are woken up.
On a single core, it may be that a single thread wakes up and consumes all the NOOP jobs in the queue.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -234,30 +234,29 @@ protected void doStop() throws Exception
...
-                thread.interrupt();
-            }
+        // If we still have threads running, get a bit more aggressive
+        // interrupt remaining threads
+        for (Thread thread : _threads)
+        {
+            if (LOG.isDebugEnabled())
+                LOG.debug("Interrupting {}", thread);
+            thread.interrupt();
+        }
```
sbordet, line 254: Now it may be possible that we interrupt threads immediately after submitting the NOOP jobs (or normal jobs) -- I'm not sure we want that if we want to gracefully stop (by letting the current tasks finish).

Also, interrupting is the right thing to do, but e.g. `ReservedThreadExecutor` does not handle it well: it just loops around and goes again into `take()/poll()`.

sbordet:
@gregw why does not QTP implements `Graceful`?

Mmmh. I know I have added review comments and I would like to improve, but I've been running in circles for a while now, and nothing seems better than what we have here 😃

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
gregw: I think you are right.  The Runner used to have a `while(isRunning)` loop, so once woken it would check that state.  Now it is only checking the equivalent state only if it doesn't have a job.   So maybe we should always check the running state again?

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
gregw: Ah that is a problem. The `ReservedThread#run` method does do a `while(isRunning)`, but there is an inner loop in `ReservedThread#reservedWait` that doesn't.    I'll fix that....

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
gregw: Ah no, the running state is already always tested - either in addCounts or explicitly if there is no jobs.  It might be clearer to always call addCounts.... let me see...

gregw: @sbordet I've pushed a refactored version that I think makes this clearer... and I am 95% sure is exactly the same logic.
@lachlan-roberts can you have a look and triple check the latest checking doesn't actually change the idle thread count logic.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
gregw: last check in fixed the inner loop of the ReservedThread

sbordet (approved):

PR #6411, opened 2021-06-15
Issue #6407 - Fix URI validation for WebSocket ClientUpgradeRequest

### LINKED ISSUE #6407: Malformed scheme logical expression check in WebSocket ClientUpgradeRequest
**Jetty version(s)**  11.0.5

**Java version/vendor** `(use: java -version)`  15.0.1

**OS type/version** Windows 10

**Description**
in the class ClientUpgradeRequest there is a malformed check in the constructor:
`if (!HttpScheme.WS.is(scheme) || !HttpScheme.WSS.is(scheme))`

That should obviously be
`if (!HttpScheme.WS.is(scheme) && !HttpScheme.WSS.is(scheme))`

**How to reproduce?**
Perform a websocket upgrade request

lachlan-roberts:
## Closes #6407

- Fix logic for URI validation in the constructor of `ClientUpgradeRequest`.
- Deprecate that constructor as the `URI` is not used and the validation is repeated after calling the `WebSocketClient.connect()` method.

### DIFF
--- a/jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/util/WSURI.java
+++ b/jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/util/WSURI.java
@@ -103,6 +103,9 @@ public final class WSURI
     {
         Objects.requireNonNull(inputUri, "Input URI must not be null");
         String httpScheme = inputUri.getScheme();
+        if (httpScheme == null)
+            throw new URISyntaxException(inputUri.toString(), "Undefined HTTP scheme");
+
         if ("ws".equalsIgnoreCase(httpScheme) || "wss".equalsIgnoreCase(httpScheme))
             return inputUri;
 
--- a/jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/util/WSURI.java
+++ b/jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/util/WSURI.java
@@ -104,22 +104,13 @@ public final class WSURI
         Objects.requireNonNull(inputUri, "Input URI must not be null");
         String httpScheme = inputUri.getScheme();
         if ("ws".equalsIgnoreCase(httpScheme) || "wss".equalsIgnoreCase(httpScheme))
-        {
-            // keep as-is
             return inputUri;
-        }
 
+        String afterScheme = inputUri.toString().substring(httpScheme.length());
         if ("http".equalsIgnoreCase(httpScheme))
-        {
-            // convert to ws
-            return new URI("ws" + inputUri.toString().substring(httpScheme.length()));
-        }
-
+            return new URI("ws" + afterScheme);
         if ("https".equalsIgnoreCase(httpScheme))
-        {
-            // convert to wss
-            return new URI("wss" + inputUri.toString().substring(httpScheme.length()));
-        }
+            return new URI("wss" + afterScheme);
 
         throw new URISyntaxException(inputUri.toString(), "Unrecognized HTTP scheme");
     }
--- a/jetty-websocket/websocket-jetty-client/src/main/java/org/eclipse/jetty/websocket/client/ClientUpgradeRequest.java
+++ b/jetty-websocket/websocket-jetty-client/src/main/java/org/eclipse/jetty/websocket/client/ClientUpgradeRequest.java
@@ -50,11 +50,12 @@ public final class ClientUpgradeRequest implements UpgradeRequest
         this.host = null;
     }
 
+    @Deprecated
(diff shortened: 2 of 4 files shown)


jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/util/WSURI.java
```diff
@@ -104,22 +104,13 @@ public static URI toWebsocket(final URI inputUri) throws URISyntaxException
         Objects.requireNonNull(inputUri, "Input URI must not be null");
         String httpScheme = inputUri.getScheme();
         if ("ws".equalsIgnoreCase(httpScheme) || "wss".equalsIgnoreCase(httpScheme))
```
sbordet, line 109: Replace this line too using `HttpScheme.WS[S].is(...)`

jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/util/WSURI.java
```diff
@@ -104,22 +104,13 @@ public static URI toWebsocket(final URI inputUri) throws URISyntaxException
...
-        {
-            // keep as-is
             return inputUri;
-        }
 
+        String afterScheme = inputUri.toString().substring(httpScheme.length());
```
sbordet, line 112: `httpScheme` may be `null`, you need a guard like in `Negotiated.toWebSocket()`.

Is it possible to implement this method once and call it from different places?

jetty-websocket/websocket-jetty-client/src/main/java/org/eclipse/jetty/websocket/client/ClientUpgradeRequest.java
```diff
@@ -50,11 +50,12 @@ public ClientUpgradeRequest()
         this.host = null;
     }
 
+    @Deprecated
```
sbordet, line 56: Add javadoc `@deprecated` that explains what to use instead.

sbordet (changes requested):

jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/util/WSURI.java
lachlan-roberts: We can't really do this in the Jetty WebSocket API, we don't want to add a dependency on `jetty-http`.

jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/util/WSURI.java
lachlan-roberts: > Is it possible to implement this method once and call it from different places?

Not really as the API does not have dependencies on anything and where it is used in `websocket-core` cannot depend on `websocket-jetty-api`.

jetty-websocket/websocket-jetty-client/src/main/java/org/eclipse/jetty/websocket/client/ClientUpgradeRequest.java
```diff
@@ -50,11 +50,16 @@ public ClientUpgradeRequest()
         this.host = null;
     }
 
+    /**
+     * @deprecated use the no-args constructor instead.
+     * @see ClientUpgradeRequest#ClientUpgradeRequest()
```
sbordet:
```suggestion
     * @deprecated use {@link #ClientUpgradeRequest()} instead.
```

sbordet (changes requested):
One last nit.

sbordet (approved):

jetty-websocket/websocket-jetty-client/src/main/java/org/eclipse/jetty/websocket/client/ClientUpgradeRequest.java
joakime: Should we just deprecate all of the ClientUpgradeRequest.requestURI behaviors too? (as in `.getRequestURI()` should probably be deprecated and always return null too.

joakime (dismissed):
I think we should remove all of requestURI internals from ClientUpgradeRequest, and deprecate the constructor and getters.

jetty-websocket/websocket-jetty-client/src/main/java/org/eclipse/jetty/websocket/client/ClientUpgradeRequest.java
lachlan-roberts: @joakime it implements these methods from the `UpgradeRequest` interface, and these methods need to be used on other implementations of this interface so they can't be deprecated.

PR #6412, opened 2021-06-15
Issue #6406 - port jetty-jaspi to use Jakarta Authentication 2.x

### LINKED ISSUE #6406: Jetty Jaspi module not compatible with Jakarta EE 9 (Jakarta Authentication)
**Jetty version(s)**
11.0.3

16.0.1
**OS type/version**
Windows 10

**Description**

jetty-jaspi module in 11.0.3 relies on old geronimo-jaspi 2.0.0 and javax.xml.bind:jaxb-api:2.3.1
This relies on looking up the javax.security.auth.message.config.AuthConfigFactory instead of the JakartaEE 9 Jakarta Authentication jakarta.security.auth.message.config.AuthConfigFactory, and therefore fails to load any Jakarta Authentication modules (ServerAuthModule)

**How to reproduce?**
Configure and  run a Jakarta Authentication EE 9 compatible ServerAuthModule.

frode-carlsen:
* removed test dependency on old Geronomi-jaspi and created basic AuthConfFactory which can be initialized programmatically


jetty-jaspi/pom.xml
```diff
@@ -63,38 +63,28 @@
...
-      <artifactId>javax.security.auth.message</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>javax.xml.bind</groupId>
-      <artifactId>jaxb-api</artifactId>
-      <version>2.3.1</version>
-      <scope>test</scope>
+      <groupId>jakarta.authentication</groupId>
+      <artifactId>jakarta.authentication-api</artifactId>
+      <version>2.0.0</version>
```
janbartel: This needs to be defined in the top level `pom.xml`, and the existing declaration for `org.eclipse.jetty.orbit|javax.security.auth.message `removed.
Also the `jetty-home/pom.xml `also mentions the old `javax.security.auth.message` and `javax.activation` artifacts and should also be removed and replaced with the new jakarta jars instead.
The `jetty-maven-plugin/src/it/jetty-start-war-mojo-it/pom.xml` also is relying on `javax.activation` and needs to be upgraded - this probably depends on upgrading to a newer version of the glassfish `bean-validation-webapp`.

jetty-jaspi/pom.xml
```diff
@@ -63,38 +63,28 @@
...
-      <version>2.0.0</version>
-      <exclusions>
-        <exclusion>
-          <groupId>org.apache.geronimo.specs</groupId>
-          <artifactId>geronimo-jaspic_1.0_spec</artifactId>
-        </exclusion>
-      </exclusions>
+      <groupId>jakarta.activation</groupId>
+      <artifactId>jakarta.activation-api</artifactId>
+      <version>2.0.1</version>
```
janbartel: This should also be defined in the top level `pom.xml`.

jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/JaspiAuthConfigProvider.java
```diff
@@ -0,0 +1,135 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** 
+ * Support for JASPI 2.0 Servlet Container Profile. 
+ */
+public class JaspiAuthConfigProvider implements AuthConfigProvider
+{
+
+    private static final Logger log = LoggerFactory.getLogger(JaspiAuthConfigProvider.class);
```
janbartel: The jetty convention is to use `LOG` as it's a static.

jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/JaspiAuthConfigProvider.java
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * Convenience constructor.
+     * 
+     * @param properties properties 
+     * @param serverAuthModuleClassName Name of class implementing ServerAuthModule ifc
+     */
+    public JaspiAuthConfigProvider(Map<String, String> properties, String serverAuthModuleClassName)
+    {
+        Objects.requireNonNull(serverAuthModuleClassName);
+        log.trace("Instantiated with: {}", serverAuthModuleClassName);
```
janbartel: Pretty sure we don't use `trace` in jetty, just `debug`.

jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/SimpleAuthConfig.java
```diff
@@ -15,59 +15,67 @@
...
     {
-        return true;
+        // NOOP
     }
 
     @Override
-    public void refresh()
+    public boolean isProtected()
     {
+        return false;
```
janbartel: This was `true`, are you sure the change to `false` is correct?

janbartel (changes requested):
Thanks for your time and effort in upgrading jetty to work with the latest jaspi version!

I've left some comments in the code of things that need to be addressed - if you have any questions please just ask.

jetty-jaspi/src/test/java/org/eclipse/jetty/security/jaspi/JaspiTest.java
```diff
@@ -86,10 +89,31 @@ protected UserPrincipal loadUserInfo(String username)
         }
     }
 
+    @BeforeAll
+    public static void beforeAll() throws Exception
+    {
+        AuthConfigFactory factory = new BasicAuthConfigFactory();
+
```
janbartel: We also need to come up with a way of configuring this in a jetty xml file, as this is how jetty modules are enabled. Take a look at the `jetty-jaspi/src/main/config/modules/jaspi.mod` file - it references an xml that will be put on jetty's execution path to configure jaspi. Have a read of https://www.eclipse.org/jetty/documentation/jetty-11/operations-guide/index.html#og-modules to find out about jetty modules.

frode-carlsen: Ok, the previous implementation relied on the old Geronimo implementation (so they'd both need to be set up) and the xml file reflects that. I'm not sure that's appropriate to replicate into the jetty project, presumably (unless geronimo upgrades at some point) people would pick the glassfish one instead (which has a different configuration format (not-xml), or another implementation.  Alternatively a new configuration format would need to be implemented in jetty if one wants to configure it through xml or what not.

jetty-jaspi/pom.xml
frode-carlsen: might be more appropriate to import the jakartaee bom for the jakarta 9 dependency management? Or do you want individual specs still listed ?

jetty-jaspi/pom.xml
frode-carlsen: * removing org.eclipse.jetty.orbit|javax.security.auth.message
* removing RC3 activation-api version referenced in jetty-osgi/test-jetty-osgi

* The glassfish bean-validation-webapp seems to have been ditched 3 years ago so that's going to be hard.
Seems to have happened when the jersey repository moved from https://github.com/javaee/jersey/tree/master/examples to https://github.com/eclipse-ee4j/jersey, 
d (https://github.com/eclipse-ee4j/jersey/commit/065d21270092d6bed7b2dabbe1f65e0933372b80)
Would suggest this would be better left to a separate issue. Probably isn't a dependency one ought to have here

jetty-jaspi/src/test/java/org/eclipse/jetty/security/jaspi/JaspiTest.java
frode-carlsen: Perhaps an alternative would be to provide an example of a "configure" xml file? https://www.eclipse.org/jetty/documentation/jetty-11/operations-guide/index.html#og-modules-directive-xml

jetty-jaspi/pom.xml
janbartel: @olamy should we update the bean-validation-webapp that the jetty-maven-plugin tests are using???

jetty-jaspi/src/test/java/org/eclipse/jetty/security/jaspi/JaspiTest.java
janbartel: I was thinking we could provide an xml file for the `BasicAuthConfigFactory`, and parameterize the necessary args (classnames, map values etc).  

To allow users to provide their own `AuthConfigFactory` implementation (reasonable?) maybe it would be a` jaspi.mod` to get the jars onto the classpath and maybe a `jaspi-basic.mod` to get the `BasicAuthConfigFactory` xml config file onto the classpath?

frode-carlsen: this can be a list of ServerAuthModules though, as the JaspiTest shows, it's 'possible' to have different ServerAuthModules for different context paths (and possibly by different logical servers). Though I'd imagine it would be rare - at least in a microservice world.  

If we settle for a single ServerAuthModule supported by default in a module we could parameterize classname of ServerAuthModule impl and context path (possibly also logical server, but I don't see it used so far).  Maybe support realm, but that's jetty specific configuration

janbartel: Maybe the jaspi-basic.xml file would be as simple as creating the BasicAuthConfigFactory, and assigning it an `id` so it could be referenced in user-supplied xml files to configure the various ServerAuthModules. As you said previously, maybe we need to provide an example, commented out? Although, I kinda hate to do that, as nothing rots as quickly as commented out code/xml ;)

frode-carlsen: ok, I'm not familiar with the modules configuration so will need to look into it and give it a shot, probably later tonight

jetty-home/pom.xml
```diff
@@ -402,33 +402,6 @@
               <outputDirectory>${source-assembly-directory}/lib/glassfish-jstl</outputDirectory>
             </configuration>
           </execution>
-          <execution>
```
janbartel: We still need to copy the jaspi dependency jars into ${assembly-directory/lib/jaspi.

jetty-home/pom.xml
```diff
@@ -534,11 +507,6 @@
   </build>
   <dependencies>
     <!-- Orbit Deps -->
```
janbartel: At this point there aren't any more orbit deps, so we can get rid of the comment ;)

jetty-home/pom.xml
frode-carlsen: this seems to only include  the orbit jars that are no longer needed?

jetty-jaspi/src/test/java/org/eclipse/jetty/security/jaspi/JaspiTest.java
frode-carlsen: I'm out of my depth here, but there are several things that need to be done.  I've tried to illustrate using the following mock jaspi-basic.xml file

```
<?xml version="1.0"?>
<!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "http://www.eclipse.org/jetty/configure_10_0.dtd">

<Configure id="context" class="org.eclipse.jetty.webapp.WebAppContext">

  <!-- Step 1: init AuthConfigFactory for jakarta.   -->
  <New id="jaspiAuthConfigFactory" class="org.eclipse.jetty.jaspi.BasicAuthConfigFactory" />
  <Call class="jakarta.security.auth.message.config.AuthConfigFactory" name="setFactory">
    <Arg><Ref refid="jaspiAuthConfigFactory" /></Arg>
  </Call>
  
  <!-- Step 2: configure AuthConfigProvider with appropriate ServerAuthModules -->
  <Ref refid="jaspiAuthConfigFactory">
	<Call name="registerConfigProvider">
          <Arg type="String">org.eclipse.jetty.security.jaspi.BasicAuthModule</Arg>
          <Arg>
              <Map>
                  <Entry>
                      <Item>AppContextID</Item>
                      <Item>server /other</Item>
                  </Entry>
                  <Entry>
                      <Item>org.eclipse.jetty.security.jaspi.modules.RealmName</Item>
                      <Item>TestRealm</Item>
                  </Entry>
              </Map>
          </Arg>
          <Arg type="String">HttpServlet</Arg>
          <Arg type="String">server /ctx</Arg>
          <Arg type="String">a test provider</Arg>
	</Call>
	<Call name="registerConfigProvider">
            <Arg type="String">org.eclipse.jetty.security.jaspi.HttpHeaderAuthModule</Arg>
            <Arg>
                <Map>
                    <Entry>
                        <Item>AppContextID</Item>
                        <Item>server /other</Item>
                    </Entry>
                </Map>
            </Arg>
            <Arg type="String">HttpServlet</Arg>
            <Arg type="String">server /other</Arg>
            <Arg type="String">another test provider</Arg>
	</Call>
  </Ref>
  
  <!-- Step 3: connect JASPI AuthConfigFactory to WebAppContext -->
  <Set name="securityHandler">
    <New class="org.eclipse.jetty.security.ConstraintSecurityHandler">
      <Set name="loginService">
        <New class="org.eclipse.jetty.plus.jaas.JAASLoginService">
          <Set name="name">JAASRealm</Set>
          <Set name="loginModuleName">jaas</Set>
        </New>
      </Set>
  
      <Set name="authenticatorFactory">
        <New class="org.eclipse.jetty.security.jaspi.JaspiAuthenticatorFactory" />
      </Set>
    </New>
  </Set>
  
</Configure>
```


**Step 1**:  This sets up the BasicAuthConfigFactory to be de default AuthConfigFactory.  Can alternatively be done through system property  "authconfigprovider.factory" defined by jakarta.security.auth.message.config.AuthConfigFactory#DEFAULT_FACTORY_SECURITY_PROPERTY and referenced from jakarta...AuthConfigFactory#getFactory

Should one wish to use a different AuthConfigFactory (e.g. glassfish or another impl), then it might be better to provide this through a system property (with a default value if not set otherwise)

**Step 2**:  This is custom per app, but is usable across different AuthConfigFatory implementations (though most provide another means of configuration as well - such as the old geronimo jaspi script).  I've emulated the test Server AuthModule for good measure, but other portable ServerAuthModules could be configured here depending on need

**Step 3**: Finally, it all needs to be connected back to Jetty WebAppContext.  I don't see webappcontext being a default id in the module system, but "context" seems to have been used in the examples.  Not sure how this is all wired together

jetty-home/pom.xml
frode-carlsen: I'm out of my depth here, but there are several things that need to be done.  I've tried to illustrate using the following mock jaspi-basic.xml file

```
<?xml version="1.0"?>
<!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "http://www.eclipse.org/jetty/configure_10_0.dtd">

<Configure id="context" class="org.eclipse.jetty.webapp.WebAppContext">

  <!-- Step 1: init AuthConfigFactory for jakarta. 
       TBD: Could alternatively be set through system property "authconfigprovider.factory" defined by jakarta.security.auth.message.config.AuthConfigFactory#DEFAULT_FACTORY_SECURITY_PROPERTY and referenced from AuthConfigFactory#getFactory
  -->
  <New id="jaspiAuthConfigFactory" class="org.eclipse.jetty.jaspi.BasicAuthConfigFactory" />
  <Call class="jakarta.security.auth.message.config.AuthConfigFactory" name="setFactory">
    <Arg><Ref refid="jaspiAuthConfigFactory" /></Arg>
  </Call>
  
  <!-- Step 2: configure AuthConfigProvider with appropriate ServerAuthModules -->
  <Ref refid="jaspiAuthConfigFactory">
	<Call name="registerConfigProvider">
		<Arg type="String">org.eclipse.jetty.security.jaspi.BasicAuthModule</Arg>
	    <Arg>
            <Map>
              <Entry>
                <Item>AppContextID</Item>
                <Item>server /other</Item>
              </Entry>
			  <Entry>
				<Item>org.eclipse.jetty.security.jaspi.modules.RealmName</Item>
				<Item>TestRealm</Item>
			  </Entry>
            </Map>
		</Arg>
		<Arg type="String">HttpServlet</Arg>
		<Arg type="String">server /ctx</Arg>
		<Arg type="String">a test provider</Arg>
	</Call>
	<Call name="registerConfigProvider">
		<Arg type="String">org.eclipse.jetty.security.jaspi.HttpHeaderAuthModule</Arg>
		<Arg>
            <Map>
              <Entry>
                <Item>AppContextID</Item>
                <Item>server /other</Item>
              </Entry>
            </Map>
		</Arg>
		<Arg type="String">HttpServlet</Arg>
		<Arg type="String">server /other</Arg>
		<Arg type="String">another test provider</Arg>
	</Call>
  </Ref>
  
  <!-- Step 3: connect JASPI AuthConfigFactory to WebAppContext -->
  <Set name="securityHandler">
    <New class="org.eclipse.jetty.security.ConstraintSecurityHandler">
      <Set name="loginService">
        <New class="org.eclipse.jetty.plus.jaas.JAASLoginService">
          <Set name="name">JAASRealm</Set>
          <Set name="loginModuleName">jaas</Set>
        </New>
      </Set>
  
      <Set name="authenticatorFactory">
        <New class="org.eclipse.jetty.security.jaspi.JaspiAuthenticatorFactory" />
      </Set>
    </New>
  </Set>
  
</Configure>

```

Step 1:  This sets up the BasicAuthConfigFactory to be de default AuthConfigFactory.  Can alternatively be done through system property

jetty-jaspi/src/test/java/org/eclipse/jetty/security/jaspi/JaspiTest.java
frode-carlsen: My question is then, how would this best be separated into resuable/configurable components?

janbartel: OK, so the `jaspi-basic.xml` file needs to configure stuff at the `Server` level, because when it runs, the `WebAppContext` won't exist yet. IIUC the `BasicAuthConfigFactory` can be a shared resource amongst all contexts, right?  If so, then the `jaspi-basic.xml` file would look like:

```xml
<?xml version="1.0"?>
<!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "https://www.eclipse.org/jetty/configure_10_0.dtd">

<Configure id="Server" class="org.eclipse.jetty.server.Server">

  <!-- ===================================================================== -->
  <!-- Configure a factory for Jaspi                                         -->
  <!-- ===================================================================== -->
  <Call name="addBean">
   <Arg>
        <New id="jaspiAuthConfigFactory" class="org.eclipse.jetty.jaspi.BasicAuthConfigFactory" />
   </Arg>
  </Call>
  <Call class="jakarta.security.auth.message.config.AuthConfigFactory" name="setFactory">
      <Arg><Ref refid="jaspiAuthConfigFactory" /></Arg>
  </Call>
</Configure>
```

You can remove all of **Step 3**.  The `ConstraintSecurityHandler` is the default that will be used; the `SecurityHandler` will find whatever `LoginService` has been configured (or the user can define their own in their own context xml file - that's a documented part of jetty). As for the `JaspiAuthenticatorFactory`, you don't need to configure it for each webapp. Instead, you put a `META-INF/services/` definition file in `jetty-jaspi/src/main/resources` for the `JaspiAuthenticatorFactory` and jetty will find it and hook it up to the `SecurityHandler` when it starts. 

So, AFAICT it is just **Step 2** that needs to be done per `WebAppContext` in a user-defined context xml file. So in that case, we could do the same as the `openid.mod` does, and copy a commented-out (urgh) example xml file into the `$jetty.base/etc` directory that the user can modify, see: https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-openid/src/main/config/modules/openid.mod#L15  and

jetty-home/pom.xml
janbartel: Sure but don't we need to copy in the _new_ jaspi dependency jars into `${assembly-directory/lib/jaspi`?

frode-carlsen: ok, adding jakarta.authentication-api here (only spec needed, no other deps)

jetty-jaspi/src/test/java/org/eclipse/jetty/security/jaspi/JaspiTest.java
frode-carlsen: Ok, sounds good. 
->Step 1: this means that if anyone wants to use a different provider (atm only glassfish has one compatible), they'd have to create a new module definition to match this part (replacing BasicAuthConfigFactory).  Perhaps the classname should be parameterized somehow?
->Step 2: added example code, ( note this must always be tweaked to match chosen ServerAuthModules,as the two implementations given are only availlable in unittest scope)

janbartel:
@frode-carlsen I've fixed and tweaked the jetty `.mod `and `.etc` files, and attached them to this post, can you check them in please?

[files.zip](https://github.com/eclipse/jetty.project/files/6684233/files.zip)

jetty-jaspi/src/main/config/etc/jaspi-basic.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0"?>
```
janbartel: This file should be in `src/main/config/etc/jaspi/jaspi-basic.xml` to agree with the location mentioned in the `jaspi-basic-auth-config-factory.mod` file .

jetty-jaspi/pom.xml
frode-carlsen: I presume the issue of the bean-validation-webapp should be opened as a separate issue, marking conversation resolved

jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/BasicAuthConfigFactory.java
```diff
@@ -0,0 +1,111 @@
...
+import javax.security.auth.AuthPermission;
+
+import jakarta.security.auth.message.config.AuthConfigFactory;
+import jakarta.security.auth.message.config.AuthConfigProvider;
+import jakarta.security.auth.message.config.RegistrationListener;
+
+/** 
+ * A very basic {@link AuthConfigFactory} that allows for registering providers programmatically.
+ */
+public class BasicAuthConfigFactory extends AuthConfigFactory
```
lachlan-roberts: I'm not a fan of this naming as it can be confused to have some relation to Basic authentication. Can this be renamed to something like `DefaultAuthConfigFactory` (along with the .xml file, and the .mod file)?

frode-carlsen: renamed

jetty-jaspi/src/main/config/modules/jaspi/jaspi-authmoduleconfig.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0"?>
+<!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "https://www.eclipse.org/jetty/configure_10_0.dtd">
+<Configure>
+  <!-- 
+       Example code to configure ServerAuthModules to be used by Jakarta Authentication 
+       using the programmatic Authentication API (registerConfigProvider)
+  -->
+  
+  <!--
```
janbartel: I think we should _not_ have commented out xml: it can't be verified and just rots.  So, my suggestion is that you move the `BasicAuthModule` up into `src/main/java/org/eclipse/jetty/security/jaspi/modules`  so it can actually be used.  Then, we can change the contents of this file to something like:

```suggestion
    <Ref refid="jaspiAuthConfigFactory">
	<Call name="registerConfigProvider">
          <Arg type="String">org.eclipse.jetty.security.jaspi.modules.BasicAuthModule</Arg>
          <Arg>
              <Map>
                  <Entry>
                      <Item>AppContextID</Item>
                      <Item>server /demo</Item>
                  </Entry>
                  <Entry>
                      <Item>org.eclipse.jetty.security.jaspi.modules.RealmName</Item>
                      <Item>TestRealm</Item>
                  </Entry>
              </Map>
          </Arg>
          <Arg type="String">HttpServlet</Arg>
          <Arg type="String">server /ctx</Arg>
          <Arg type="String">a test provider using BASIC authentication</Arg>
	</Call>
  </Ref>
</Configure>
```

We should put comments next to each arg/item to explain what they are.  Eg what is the arg `server /ctx`?? Is this a jaspi constant, or is it supposed to be the context path, or something else???

jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/modules/BaseAuthModule.java
```diff
@@ -22,15 +22,15 @@
...
-import javax.security.auth.message.AuthException;
-import javax.security.auth.message.AuthStatus;
-import javax.security.auth.message.MessageInfo;
-import javax.security.auth.message.MessagePolicy;
-import javax.security.auth.message.callback.CallerPrincipalCallback;
-import javax.security.auth.message.callback.GroupPrincipalCallback;
-import javax.security.auth.message.config.ServerAuthContext;
-import javax.security.auth.message.module.ServerAuthModule;
 
+import jakarta.security.auth.message.AuthException;
```
janbartel, line 26: I think the name of this class should be 'AbstractAuthModule` so we don't get confused with BASIC authentication. It should also be `abstract`.

janbartel (changes requested):
Could you also add more javadoc to the classes, to explain what they do.

We are also going to need some jetty documentation for this too :)  I can help you with that, but will need your input as the jaspi expert.

jetty-jaspi/src/main/config/modules/jaspi/jaspi-authmoduleconfig.xml
frode-carlsen: method invoked is part of the AuthConfigFactory interface, which has javadoc., so not sure it needs anything else her?  e.g. "server /ctx" is example of a (logical) hostname + context path expalined in the spec.  "HttpServlet" is a constant in the spec (atm only profile supported here - the other, SOAP is deprecated and not supported here)

frode-carlsen: I can't vouch for the quality of the Basic authentication auth module, but it looks Ok. Suggest this is marked as example

jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/modules/BaseAuthModule.java
frode-carlsen: Most of the classes are simple implementations of the required Jakarta Authentication interfaces, so method doc should be inherited. 
The way Jakarta Authentication works is has a number of abstract-wrapper-factory-wrapper-layers which requires a bit of bridging code to jetty.  Since this does not support dynamically registering auth modules via a web admin gui, or any other profiles but the Servlet Container Profiles, most of these are only for briding jetty security to jakarta authentication.

It basically boils down to being able to call the AuthConfigFactory#registerConfigProvider (as jetty has no admin gui to manipulate ServerAuthModules)

In Jakarta Authentication 3.0 (Jakara EE 10) this will be further simplified by allowing one to register a ServerAuthModule directly on the AuthConfigFactory - avoiding some layers of wrapping (other than to support the more advanced (legacy) wrapper-wrapper mode of registering auth modules with different message layer profiles).

jetty-jaspi/src/main/config/modules/jaspi/jaspi-authmoduleconfig.xml
frode-carlsen: Added some basic documentation and included the module setup above , but presumably most using Jakarta Authentication (JASPI) would be familiar with the specification.

documentation/jetty-documentation/src/main/asciidoc/old_docs/security/authentication.adoc
```diff
@@ -511,3 +511,56 @@ Below is an example which, like the one above, sets up a server with a `HashLogi
...
 ----
+
+==== Jakarta Authentication (JASPI)
+
+Jetty can utilize portable authentication modules that implements the Jakara Authentication specification. This requires the jetty-jaspi module.
+
+Only modules conforming to the ServerAuthModule interface in the https://jakarta.ee/specifications/authentication/2.0/jakarta-authentication-spec-2.0.pdf[JakartaAuthentication] are supported. These modules must be configured before start-up.  
+
+The following illustrates a jetty module setting up HTTP Basic Authentication using a Jakarta Authentication module that comes packaged with the jetty-jaspi module: `org.eclipse.jetty.security.jaspi.modules.BasicAuthenticationAuthModule`
+
```
janbartel, line 522: We need to add some documentation about the acceptable values for the `appContext`: in particular the prefix `server` and how you could supply different values for that.

documentation/jetty-documentation/src/main/asciidoc/old_docs/security/authentication.adoc
```diff
@@ -511,3 +511,56 @@ Below is an example which, like the one above, sets up a server with a `HashLogi
...
+[source, xml, subs="{sub-order}"]
+----
```
janbartel, line 524: I think there's a way we can put markup in an .xml file that can be parsed and included in the documentation, I'll get back to you on that.

jetty-jaspi/src/main/config/modules/jaspi/jaspi-authmoduleconfig.xml
```diff
@@ -0,0 +1,45 @@
...
+                      <Item>org.eclipse.jetty.security.jaspi.modules.RealmName</Item>
+                      <Item>TestRealm</Item>
+                  </Entry>
+              </Map>
+          </Arg>
+          <!-- Message Layer Identifier as per spec chapter 3.1  -->
+          <Arg type="String">HttpServlet</Arg>
+          <!-- 
+            Application Context Identifier as per spec chapter 3.2.  
+            Should be same as the one configured for the provided BasicAuthModule above.
```
janbartel:
```suggestion
            Should be same as the one configured for the provided BasicAuthenticationAuthModule above.  
```

jetty-jaspi/src/main/config/modules/jaspi/jaspi-authmoduleconfig.xml
```diff
@@ -0,0 +1,45 @@
...
+              <Map>
+                  <Entry>
+                      <!-- 
+                        Application Context Identifier as per spec chapter 3.2 (property key=AppContextID) 
+                       -->
+                      <Item>AppContextID</Item>
+                      <!-- 
+                        Should be of the form "logical-hostname<space>context-path"                       
+                       -->
+                      <Item>server /ctx</Item>
```
janbartel: It would be ideal if the `server /ctx` didn't have to be repeated here and at line 38.

janbartel (changes requested):

jetty-jaspi/src/main/config/modules/jaspi/jaspi-authmoduleconfig.xml
frode-carlsen: will remove first entry - leftover from geronimo example

documentation/jetty-documentation/src/main/asciidoc/old_docs/security/authentication.adoc
frode-carlsen: Ok, added some more from the specification. It should match the ServletContainer#getVirtualServerName, but in reality isn't necessary other than for compliance with the api in here.

janbartel:
@frode-carlsen I've harmonised your documentation with the new jetty documentation, and included the jaspi-authmoduleconfig.xml automatically into the doco. Here's the tar ball for you.
[changes.zip](https://github.com/eclipse/jetty.project/files/6782029/changes.zip)

frode-carlsen:
> [changes.zip](https://github.com/eclipse/jetty.project/files/6782029/changes.zip)

@janbartel  The operations-guide/jaspi folder is empty. Was that intentional?

frode-carlsen:
Added a operations-guide/jaspi/chapter.adoc

janbartel:
oops, here's the missing chapter.adoc

[jaspi-chapter.zip](https://github.com/eclipse/jetty.project/files/6783986/jaspi-chapter.zip)

documentation/jetty-documentation/src/main/asciidoc/operations-guide/jaspi/chapter.adoc
```diff
@@ -0,0 +1,45 @@
...
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+[[og-jaspi]]
+=== JASPI
+
+Enabling this module allows Jetty to utilize authentication modules that implement the Jakara Authentication (JASPI) specification. JASPI provides an SPI (Service Provider Interface) for pluggable, portable, and standardized authentication modules. Compatible modules are portable between servers that support the JASPI specification. This module provides a bridge from Jakarta Authentication to the Jetty Security framework.  
+
+Only modules conforming to the "Servlet Container Profile" with the ServerAuthModule interface within the https://jakarta.ee/specifications/authentication/2.0/jakarta-authentication-spec-2.0.pdf[JakartaAuthentication] are supported her. These modules must be configured before start-up.  Operations for runtime registering or de-registering authentication modules are not supported.
```
janbartel: Remove `her.` on line 19.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/jaspi/chapter.adoc
```diff
@@ -0,0 +1,45 @@
...
+When you enable the `jaspi` module, the example configuration file `jaspi-authmoduleconfig.xml` is copied into your `${jetty.base}/etc` directory:
+
+----
+include::{JETTY_HOME}/modules/jaspi/jaspi-authmoduleconfig.xml[tags=documentation]
+----
+
+It illustrates setting up HTTP Basic Authentication using a Jakarta Authentication module that comes packaged with jetty: `org.eclipse.jetty.security.jaspi.modules.BasicAuthenticationAuthModule`, and applies it for a context named `/ctx`. Other custom or 3rd party modules that are compatible with the `ServerAuthModule` interface in JASPI may be imported here.
+
+
+
```
janbartel, line 57: You need to add a description of how to finish the configuration setup for a webapp, which involves editing the context xml file for the webapp, and adding this line:  
```xml
  <Get name="securityHandler">
      <Set name="authenticatorFactory"><New class="org.eclipse.jetty.security.jaspi.JaspiAuthenticatorFactory"/></Set>
  </Get>
```

janbartel (changes requested):

janbartel:
It's looking good!  The programmer manual doco needs to be done, but if you're happy to leave it to me, I'll have a go at it after this PR is merged.

frode-carlsen:
👍 yes, that would be great

janbartel:
@lachlan-roberts please review this PR to have another pair of eyes on it: especially as I've contributed to it :)

documentation/jetty-documentation/src/main/asciidoc/old_docs/security/authentication.adoc
```diff
@@ -511,3 +511,52 @@ Below is an example which, like the one above, sets up a server with a `HashLogi
...
+          </Arg>
+          <!-- Message Layer Identifier as per spec chapter 3.1  -->
+          <Arg type="String">HttpServlet</Arg>
+          <!-- 
+            Application Context Identifier as per spec chapter 3.2:
+            
+            AppContextID ::= hostname blank context-path
+            
+            Hostname may be the physical or logical server name.
+            Per the specification this should match the value returned from getVirtualServerName
```
janbartel: See comments below in the code for the JaspiAuthenticationFactory.

documentation/jetty-documentation/src/main/asciidoc/old_docs/security/authentication.adoc
```diff
@@ -511,3 +511,52 @@ Below is an example which, like the one above, sets up a server with a `HashLogi
...
+<Configure>
+ <Ref refid="jaspiAuthConfigFactory">
+	<Call name="registerConfigProvider">
+          <Arg type="String">org.eclipse.jetty.security.jaspi.modules.BasicAuthenticationAuthModule</Arg>
+          <Arg>
+              <Map>
+                  <Entry>
+                      <!-- Realm as utilised by Jetty Security  -->
+                      <Item>org.eclipse.jetty.security.jaspi.modules.RealmName</Item>
+                      <Item>TestRealm</Item>
```
janbartel: The realm name is `Test Realm` (note the space!).

documentation/jetty-documentation/src/main/asciidoc/operations-guide/jaspi/chapter.adoc
```diff
@@ -0,0 +1,56 @@
...
+When you enable the `jaspi` module, the example configuration file `jaspi-authmoduleconfig.xml` is copied into your `${jetty.base}/etc` directory:
+
+----
+include::{JETTY_HOME}/modules/jaspi/jaspi-authmoduleconfig.xml[tags=documentation]
+----
+
+It illustrates setting up HTTP Basic Authentication using a Jakarta Authentication module that comes packaged with jetty: `org.eclipse.jetty.security.jaspi.modules.BasicAuthenticationAuthModule`, and applies it for a context named `/ctx`. Other custom or 3rd party modules that are compatible with the `ServerAuthModule` interface in JASPI may be imported here.
+
+Once the appropriate `ServerAuthModule` is configured, the configuration must be completed by linking it to the webapp. Open the webapp's context.xml and add the following:
```
janbartel: So sorry, actually this step is not necessary! (Thanks @lachlan-roberts  for pointing that out!). I have re-tested and it is definitely not necessary, so we can remove these instructions about the context xml.

jetty-jaspi/src/main/config/modules/jaspi/jaspi-authmoduleconfig.xml
```diff
@@ -0,0 +1,45 @@
...
+  -->
+  <Ref refid="jaspiAuthConfigFactory">
+	<Call name="registerConfigProvider">
+      <Arg type="String">org.eclipse.jetty.security.jaspi.modules.BasicAuthenticationAuthModule</Arg>
+      <Arg>
+        <Map>
+            <Entry>
+                <!-- Realm as utilised by Jetty Security  -->
+                <Item>org.eclipse.jetty.security.jaspi.modules.RealmName</Item>
+                <Item>TestRealm</Item>
```
janbartel: See previous comment, needs a space in the realm name.

jetty-jaspi/src/main/config/modules/jaspi/jaspi-authmoduleconfig.xml
```diff
@@ -0,0 +1,45 @@
...
+        </Map>
+      </Arg>
+      <!-- Message Layer Identifier as per spec chapter 3.1  -->
+      <Arg type="String">HttpServlet</Arg>
+      <!-- 
+        Application Context Identifier as per spec chapter 3.2:
+        
+        AppContextID ::= hostname blank context-path
+        
+        Hostname may be the physical or logical server name.
```
janbartel: See comments in JaspiAuthenticationFactory.

jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/JaspiAuthenticatorFactory.java
```diff
@@ -140,13 +157,13 @@ protected Subject findServiceSubject(Server server)
...
-     * use the name of the a principal in the service subject.
-     * If not found, return "server".
+     * Find a servername. If {@link #setServerName(String)} has not been called,
+     * then use the name of the a principal in the service subject. If not found,
+     * return "server".
      *
      * @param server the server to find the name of
-     * @return the server name from the service Subject (or default value if not found in subject or principals)
+     * @return the server name from the service Subject (or default value if not
+     *         found in subject or principals)
```
janbartel, line 134: According to the spec, this should be the value returned from `ServletContext.getVirtualServerName()`.  In jetty's case, this will be the first `virtual host` for the context, or `null` if there aren't any.  So I think you should ask the context for it's virtualServerName, and if you get `null`, fall back to these other ways of obtaining it, finally defaulting to the string `server`.

NB you will also have to update the jaspi-authmoduleconfig.xml file comment to reflect the above algorithm.

janbartel (changes requested):

jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/JaspiAuthenticatorFactory.java
frode-carlsen: the logic in this class hasn't been changed, but I've thrown in a call to the ServletContext.getVirtualServerName to try that if the _serverName member variable hasn't been configured.

I'm not entirely sure if the existing logic of using the first Subject's first principal name in the original logic is appropriate, but kept it as it is current behaviour.

lachlan-roberts:
@frode-carlsen @janbartel I have suggested some further changes to this PR by creating a PR into this branch (see https://github.com/frode-carlsen/jetty.project/pull/1). Can you please review these changes and they can be merged into this PR if acceptable.

frode-carlsen:
@lachlan-roberts I've merged in your PR to make things easier. 
Got a couple of comments on it that still may need to be addressed

frode-carlsen:
No idea why this fails the eclipsefdn/eca check, both emails seem to be covered. Perhaps I shouldn't have merge it myself.

lachlan-roberts:
@frode-carlsen thanks! I have replied to the comments.

> No idea why this fails the eclipsefdn/eca check, both emails seem to be covered. Perhaps I shouldn't have merge it myself.

I don't think that has anything to do with ECA failure, last time that happened I think there was some outage on eclipse validation servers.

lachlan-roberts (approved):
LGTM

janbartel:
@frode-carlsen you might need to rebase or merge from jetty-11.0.x branch, looks like there were some changes which were breaking the build, but according to CI it should be fixed now.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/jaspi/chapter.adoc
```diff
@@ -0,0 +1,43 @@
...
+Enable the `jaspi` module:
+
+----
+include::{JETTY_HOME}/modules/jaspi.mod[]
+----
+
+[[og-jaspi-xml]]
+===== Configure Jaspi
+
+When you enable the `jaspi` module, the example configuration file `jaspi-authmoduleconfig.xml` is copied into your `${jetty.base}/etc` directory:
```
janbartel: This is wrong now - the jaspi-authmoduleconfig.xml file is not copied into $jetty.base, it seems to be the jaspi-demo.xml file instead?

documentation/jetty-documentation/src/main/asciidoc/operations-guide/jaspi/chapter.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+[source, xml, subs="{sub-order}"]
+----
+include::{SRCDIR}/jetty-jaspi/src/main/config/etc/jaspi/jaspi-demo.xml[tags=documentation]
+----
+
+It illustrates setting up HTTP Basic Authentication using a Jakarta Authentication module that comes packaged with jetty: `org.eclipse.jetty.security.jaspi.modules.BasicAuthenticationAuthModule`, and applies it for a context named `/ctx`. Other custom or 3rd party modules that are compatible with the `ServerAuthModule` interface in JASPI may be imported here.
```
janbartel: @lachlan-roberts I thought we were going to configure the standard jetty test webapp at context path `/test`?

jetty-http-spi/pom.xml
```diff
@@ -48,12 +47,11 @@
     <dependency>
       <groupId>jakarta.ws.rs</groupId>
       <artifactId>jakarta.ws.rs-api</artifactId>
-      <version>3.0.0</version>
     </dependency>
     <dependency>
       <groupId>com.sun.xml.ws</groupId>
       <artifactId>jaxws-rt</artifactId>
-      <version>3.0.0-M5</version>
```
janbartel: This change should not be part of this PR.

jetty-jaspi/src/main/config/etc/jaspi/jaspi-authmoduleconfig.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "https://www.eclipse.org/jetty/configure_10_0.dtd">
```
janbartel, line 2: I don't think we need this file, now that we have the jaspi-demo module that copies an example config file into jetty.base/etc.

jetty-jaspi/src/main/config/etc/jaspi/jaspi-demo.xml
```diff
@@ -0,0 +1,58 @@
...
+        The algorithm applied here will use the
+        _serverName on the configured JaspiAuthenticatorFactory (if set) and try to match it
+        against the "server" part (in the "server /ctx" example below).
+        Next it will try to match the ServletContext#getVirtualServerName to the "server" part.
+        If neither are set, it will then try to match the first Subject's principal name, and finally fall back to
+        the default value "server" if none are available.
+
+        The context-path should match the context path where this applies.
+      -->
+      <Arg type="String">server /ctx</Arg>
```
janbartel: @lachlan-roberts I thought we were going to configure the demo test webapp at `/test`?

jetty-jaspi/src/main/config/modules/jaspi-demo.mod
```diff
@@ -0,0 +1,13 @@
...
+Enables JASPI basic authentication the /ctx context path.
+
+[tags]
+security
+
+[depend]
+jaspi
+
+[files]
+basehome:etc/jaspi/jaspi-demo.xml|etc/jaspi/jaspi-authmoduleconfig.xml
```
janbartel: Copy it as the name `jaspi-demo.xml`.

jetty-jaspi/src/main/config/modules/jaspi.mod
```diff
@@ -3,9 +3,20 @@
...
+
 [depend]
 security
+auth-config-factory
 
 [lib]
 lib/jetty-jaspi-${jetty.version}.jar
 lib/jaspi/*.jar
+
+[xml]
```
janbartel, line 17: I don't think we should copy over this file. If users want to see how to make a  jetty jaspi config file that is what the jaspi-demo module is for.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/jaspi/chapter.adoc
janbartel: Need to document the jaspi-demo module and explain how it can be used.

jetty-osgi/test-jetty-osgi/pom.xml
```diff
@@ -228,7 +228,6 @@
     <dependency>
       <groupId>jakarta.activation</groupId>
       <artifactId>jakarta.activation-api</artifactId>
-      <version>2.0.0-RC3</version>
```
janbartel: Why is this change in here?

jetty-security/src/main/java/org/eclipse/jetty/security/EmptyLoginService.java
```diff
@@ -0,0 +1,54 @@
...
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.security;
+
+import jakarta.servlet.ServletRequest;
+import org.eclipse.jetty.server.UserIdentity;
+
+public class EmptyLoginService implements LoginService
```
janbartel, line 22: Comments? Why do we need this class?

jetty-security/src/main/java/org/eclipse/jetty/security/SecurityHandler.java
```diff
@@ -572,6 +572,11 @@ else if (authentication instanceof Authentication.Deferred)
                             authenticator.secureResponse(request, response, isAuthMandatory, null);
```
janbartel: @lachlan-roberts I thought we'd agreed you'd do any changes in the SecurityHandler in a separate PR that can be reviewed and applied separately to this one?

jetty-security/src/main/java/org/eclipse/jetty/security/WrappedAuthConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+// ========================================================================
+//
+
+package org.eclipse.jetty.security;
+
+import java.util.Set;
+
+import org.eclipse.jetty.security.Authenticator.AuthConfiguration;
+
+public class WrappedAuthConfiguration implements AuthConfiguration
```
janbartel, line 24: Comments, purpose of  class?

jetty-webapp/src/main/java/org/eclipse/jetty/webapp/JaspiConfiguration.java
```diff
@@ -0,0 +1,53 @@
...
+public class JaspiConfiguration extends AbstractConfiguration
+{
+    private static final Logger LOG = LoggerFactory.getLogger(JaspiConfiguration.class);
+
+    public JaspiConfiguration()
+    {
+        addDependencies(WebXmlConfiguration.class, MetaInfConfiguration.class, WebInfConfiguration.class, FragmentConfiguration.class);
+        addDependents(WebAppConfiguration.class);
+
+        hide("jakarta.security.auth.message.");
```
janbartel, line 34: I think you can achieve this in the `jaspi.mod` file by adding the `jakarta.security.auth.message.` package to serverclasses, something like:

jetty.webapp.addServerClasses+=,jakarta.security.auth.message.

jetty-webapp/src/main/resources/META-INF/services/org.eclipse.jetty.webapp.Configuration
```diff
@@ -1,6 +1,7 @@
 org.eclipse.jetty.webapp.FragmentConfiguration
 org.eclipse.jetty.webapp.JettyWebXmlConfiguration
 org.eclipse.jetty.webapp.JaasConfiguration
+org.eclipse.jetty.webapp.JaspiConfiguration
```
janbartel, line 4: Not needed if you do the serverclasses in the jaspi.mod file.

pom.xml
```diff
@@ -1047,16 +1047,36 @@
         <artifactId>jetty-jakarta-websocket-api</artifactId>
         <version>${websocket.api.version}</version>
       </dependency>
+      <dependency>
```
janbartel: How is the jakarta.ws.rs and jakarta.activation  jars relevant here?

jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/DefaultAuthConfigFactory.java
```diff
@@ -0,0 +1,248 @@
...
+import jakarta.security.auth.message.config.AuthConfigFactory;
+import jakarta.security.auth.message.config.AuthConfigProvider;
+import jakarta.security.auth.message.config.RegistrationListener;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** 
+ * A very basic {@link AuthConfigFactory} that allows for registering providers programmatically.
+ */
+public class DefaultAuthConfigFactory extends AuthConfigFactory
```
janbartel, line 31: Need a unit test for this class.

jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/JaspiAuthenticator.java
```diff
@@ -15,72 +15,119 @@
...
+        this._authConfig = authConfig;
     }
 
     @Override
     public void setConfiguration(AuthConfiguration configuration)
     {
+        LoginService loginService = configuration.getLoginService();
+        if (loginService == null)
+        {
+            // Add an empty login service so we can use JASPI without tying into Jetty auth mechanisms.
```
janbartel, line 96: I don't see how having a fake LoginService allows a webapp to do it's own jaspi authentication?

I think if there's no LoginService, isn't this an error?

janbartel (changes requested):

pom.xml
frode-carlsen: ok, reverting jakarta version cleanup, can be handled on different PR

jetty-http-spi/pom.xml
frode-carlsen: ok, reverting jakarta version cleanup, can be handled on different PR

jetty-webapp/src/main/resources/META-INF/services/org.eclipse.jetty.webapp.Configuration
frode-carlsen: @lachlan-roberts ?

jetty-webapp/src/main/java/org/eclipse/jetty/webapp/JaspiConfiguration.java
frode-carlsen: @lachlan-roberts ?

jetty-osgi/test-jetty-osgi/pom.xml
frode-carlsen: it's a conflicting version with main pom? I can take out, but should be addressed then in another PR

frode-carlsen: seems to have been addressed already, resolving conf

jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/DefaultAuthConfigFactory.java
frode-carlsen: added

frode-carlsen:
@lachlan-roberts I presume the rest of the comments above were meant for the changes you added, don't think I've got anything else to contribute here.

lachlan-roberts:
@frode-carlsen no problem, I will discuss these remaining comments with Jan, and will provide another PR to your branch if necessary.

frode-carlsen:
updated with latest from jetty-11.0.x branch

frode-carlsen:
merged chnages from @lachlan-roberts, presume this is now ready to go?

janbartel (approved):

Issue #2134, opened 2018-01-17
UnixSocketEndPoint should return local and remote SocketAddress


joakime:
Currently, our UnixSocketEndPoint has null's for the local/remote endpoints.

https://github.com/eclipse/jetty.project/blob/jetty-9.4.8.v20171121/jetty-unixsocket/src/main/java/org/eclipse/jetty/unixsocket/UnixSocketEndPoint.java#L46-L56

If we can change those methods to return a java.net.SocketAddress then the
existing jnr.unixsocket.UnixSocketChannel can be used to obtain the local/remote UnixSocketAddress (which implements java.net.SocketAddress)

Hopefully these local/remote addresses can then be exposed in the HttpServletRequest local/remote address APIs.

sbordet:
This is trivial to implement, as we have just have to delegate to the `UnixSocketChannel` passed to the constructor.

joakime:
Seems like it, but it has a hitch ...

``` java
    @Override
    public InetSocketAddress getLocalAddress()
    {
        return _channel.getLocalAddress(); // fails, as this is just a SocketAddress, not a InetSocketAddress
    }
```

joakime:
That's because a UnixSocket's `SocketAddress` is a file, and has nothing to do with Inet anything.

sbordet:
Duplicated by #6410.

PR #6414, opened 2021-06-15
Fixes #6410 - Use SocketAddress instead of InetSocketAddress.

### LINKED ISSUE #6410: Ensure Jetty IO uses SocketAddress instead of InetSocketAddress
**Target Jetty version(s)**
10.0.x

**Enhancement Description**
In preparation for support of Unix Domain Sockets, introduced in Java 16 and handled in #6043, many Jetty I/O classes should be changed to return/use `SocketAddress` instead of `InetSocketAddress`.

As this is a quite large change that is better done separately.

sbordet:
Removed usages of InetSocketAddress in method signatures where possible.
Deprecated old methods, and added new methods with SocketAddress.


joakime (approved):
The changes in ByteArrayEndPoint had me confused for a hot minute, but i'm happy with these changes.

BTW, You fixed Issue #2134 with this PR as well.

gregw (approved):

PR #6452, opened 2021-06-21
Issue #6451 - Request.getServletPath not following spec 3.5

### LINKED ISSUE #6451: Request#getServletPath() returns null for ROOT mapping
**Jetty version(s)**
10.0.5

Java 11

**OS type/version**
Windows 10 x64 21H1

**Description**
Using:
```
 <servlet-mapping>
    <servlet-name>rwtServlet</servlet-name>
    <url-pattern></url-pattern>
  </servlet-mapping>
```
with empty "url-pattern" will register the servlet at root path. In this case Request#getServletPath() should return empty string like in Jetty 9.4.x., but with Jetty 10.x it returns null. Probably the issue is introduced by the new class ServletPathMapping. When pathInContext is null the _servletPath is set to null as well.

joakime:
The change in Jetty 10 for supporting ServletPathMappings caused a regression in behavior for the `HttpServletRequest.getServletPath()` method.

This adds a testcase and introduces the special case logic that Servlet spec 3.5 specifies.

Link to section 3.5 in Servlet Spec 5.0 (couldn't find the 4.0 spec doc online)

https://github.com/eclipse-ee4j/servlet-api/blob/5.0.0-RELEASE/spec/src/main/asciidoc/servlet-spec-body.adoc#35-request-path-elements

> Servlet Path: The path section that directly corresponds to the mapping which activated this request. This path starts with a `"/"` character except in the case where the request is matched with the `"/*"` or `""` pattern, in which case it is an empty string.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
@@ -1386,7 +1386,11 @@ public class Request implements HttpServletRequest
         // INCLUDE dispatch, in which case this method returns the servletPath of the source servlet,
         // which we recover from the IncludeAttributes wrapper.
         ServletPathMapping mapping = findServletPathMapping();
-        return mapping == null ? "" : mapping.getServletPath();
+        // per Servlet spec 3.5, mapping servlet paths of "/*", "/", and "" should return ""
+        if ((mapping == null) || mapping.getServletPath() == null || (mapping.getServletPath().equals("")) || mapping.getServletPath().equals("/") || (mapping.getServletPath().equals("/*")))
+            return "";
+        else
+            return mapping.getServletPath();
     }
 
     public ServletResponse getServletResponse()
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
@@ -1386,11 +1386,7 @@ public class Request implements HttpServletRequest
         // INCLUDE dispatch, in which case this method returns the servletPath of the source servlet,
         // which we recover from the IncludeAttributes wrapper.
         ServletPathMapping mapping = findServletPathMapping();
-        // per Servlet spec 3.5, mapping servlet paths of "/*", "/", and "" should return ""
-        if ((mapping == null) || mapping.getServletPath() == null || (mapping.getServletPath().equals("")) || mapping.getServletPath().equals("/") || (mapping.getServletPath().equals("/*")))
-            return "";
-        else
-            return mapping.getServletPath();
+        return mapping == null ? "" : mapping.getServletPath();
     }
 
     public ServletResponse getServletResponse()
--- a/jetty-servlet/src/main/java/org/eclipse/jetty/servlet/ServletHandler.java
+++ b/jetty-servlet/src/main/java/org/eclipse/jetty/servlet/ServletHandler.java
@@ -1522,8 +1522,10 @@ public class ServletHandler extends ScopedHandler
                 switch (pathSpec.getGroup())
                 {
                     case EXACT:
-                    case ROOT:
                         _servletPathMapping = new ServletPathMapping(_pathSpec, _servletHolder.getName(), _pathSpec.getPrefix());
+                        break;
+                    case ROOT:
+                        _servletPathMapping = new ServletPathMapping(_pathSpec, _servletHolder.getName(), "/");
                         break;
                     default:
                         _servletPathMapping = null;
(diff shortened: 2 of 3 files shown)


jetty-servlet/src/test/java/org/eclipse/jetty/servlet/ServletContextHandlerTest.java
```diff
@@ -724,6 +724,39 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IO
         assertEquals("getContextPath()=[]", response.getContent(), "response content");
     }
 
+    /**
+     * Address spec "3.5. Request Path Elements" with respect to Servlet Path.
+     */
+    @ParameterizedTest
+    @ValueSource(strings = {"/*", "/", ""})
+    public void testGetServletPathEmpty(String inputServletPath) throws Exception
```
gregw:
```suggestion
    public void testGetServletPathEmpty(String pathSpec) throws Exception
```

jetty-servlet/src/test/java/org/eclipse/jetty/servlet/ServletContextHandlerTest.java
```diff
@@ -724,6 +724,39 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IO
...
+
+        StringBuilder rawRequest = new StringBuilder();
+        rawRequest.append("GET / HTTP/1.1\r\n");
+        rawRequest.append("Host: local\r\n");
+        rawRequest.append("Connection: close\r\n");
+        rawRequest.append("\r\n");
+        String rawResponse = _connector.getResponse(rawRequest.toString());
+        HttpTester.Response response = HttpTester.parseResponse(rawResponse);
+        assertEquals(200, response.getStatus(), "response status");
+        assertEquals("getServletPath()=[]", response.getContent(), "response content");
```
gregw, line 757: This test is not correct for pathSpec "/", which is the default match and the servletPath should be the entire path, which in this case is "/".
As per the linked document:
>  in the case where the request is matched with the "/*" or "" pattern, in which case it is an empty string
So it is only those patterns that should pass this test.

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -1386,7 +1386,11 @@ public String getServletPath()
         // INCLUDE dispatch, in which case this method returns the servletPath of the source servlet,
         // which we recover from the IncludeAttributes wrapper.
         ServletPathMapping mapping = findServletPathMapping();
-        return mapping == null ? "" : mapping.getServletPath();
+        // per Servlet spec 3.5, mapping servlet paths of "/*", "/", and "" should return ""
+        if ((mapping == null) || mapping.getServletPath() == null || (mapping.getServletPath().equals("")) || mapping.getServletPath().equals("/") || (mapping.getServletPath().equals("/*")))
+            return "";
+        else
+            return mapping.getServletPath();
```
gregw: I don't think this is the correct fix.  The servletPath returned from the mapping should be "" for a root match.  We can see the code [intends to make it so](https://github.com/eclipse/jetty.project/blob/jetty-10.0.5/jetty-server/src/main/java/org/eclipse/jetty/server/ServletPathMapping.java#L51-L56).  

I'm seeing the mapping for that as:
```
ServletPathMapping{matchValue=, pattern=, servletName=org.eclipse.jetty.servlet.ServletContextHandlerTest$3-2bdd8394, mappingMatch=null, servletPath=null, pathInfo=null}
``` 
So the question is, why is this not seen as a ROOT match?  Why is the mappingMatch null?

gregw: Ah, as the OP said, it looks to be an issue with the pathInContext being null when constructing the ServletPathMapping???

gregw: Oh! We are treating EXACT and ROOT matches the same.... but ROOT matches have no prefix. 
So I can see a fix... but not sure I like it..... standby I'll update branch if it passed tests....

gregw (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
gregw: @joakime  I've pushed an alternate fix that corrects the ROOT match so the mapping servletPath is also correct.
It works... but something feels off about it still???  I don't like forcing the path.  Maybe the prefix of a root pathspec should be "" ?

jetty-servlet/src/main/java/org/eclipse/jetty/servlet/ServletHandler.java
```diff
@@ -1522,9 +1522,11 @@ void destroyListener(EventListener listener)
                 switch (pathSpec.getGroup())
                 {
                     case EXACT:
-                    case ROOT:
                         _servletPathMapping = new ServletPathMapping(_pathSpec, _servletHolder.getName(), _pathSpec.getPrefix());
                         break;
+                    case ROOT:
+                        _servletPathMapping = new ServletPathMapping(_pathSpec, _servletHolder.getName(), "/");
```
joakime, line 1528: The tests seem to be happy with pathInContext being `""` too (instead of `"/"`)

gregw (approved):

Issue #6447, opened 2021-06-21
Deprecate support for UTF16 encoding in URIs


gregw:
**Target Jetty version(s)**
9, 10, 11

**Enhancement Description**

The support for unicode encoded characters in URIs (eg %u2192 for the →character) have never been standardized and are little used.  Unicode (or more precisely UTF-16) encoding was [rejected](https://www.w3.org/International/iri-edit/draft-duerst-iri.html) by the W3C.

Deprecate their usage and require a compliance mode.

lachlan-roberts:
UTF16 encoding has been deprecated and is now disabled in 10.0.x/11.0.x by default. It can be re-enabled by using a `UriCompliance` mode to enable `Violation.UTF16_ENCODINGS`.

firoj07:
I will strictly follow and promote the code of conduct Thank you and connecting numbers

guusdk:
Apologies for reviving an old issue. I'm struggling with understanding in what versions exactly UTF-16 URL-encoding has been deprecated/disabled. This issue, as well as the [changelog](https://github.com/eclipse/jetty.project/releases/tag/jetty-9.4.43.v20210629) suggests that it has been removed in 9.4.43 as well as 10.0.x and 11.0.x. However, the default configuration of  a 9.4.43 does appear to retain UTF-16 URL encoding. This matches with @lachlan-roberts 's comment, which mentions deprecating the feature only in 10.0.x and 11.0.x (and does not mention 9.4.43). Was support deprecated in 9.4.43?

gregw:
UTF-16 is still supported in the latest 9.4. release (9.4.51), and is supported through to jetty-12.
What has changed, is if support is on/off by default or not.  If you want UTF-16 support, then it is available with the right compliance mode.

The deprecation was just moving support from being always on to a compliance mode.

guusdk:
Thanks Greg! I worded it awkwardly, but that is the gist of what I had assumed: Instead of having UTF-16 URL encoding support enabled by default, it now is disabled by default.

What is confusing me is that the changelog suggests that this change is introduced in 9.4.43. However, when using 9.4.43 (without any explicit HttpCompliance configured, in other words, "using the default") we find that Jetty is _not rejecting_ UTF-16 URL encoding (just like in earlier versions, but unlike version 10.0.6 for example).

gregw:
I believe we left the defaults as there were in jetty-9  (I'm not looking in detail, as jetty-9 is end-of-life for support).

[...]

PR #6457, opened 2021-06-22
Issue #6447 - Deprecate support for UTF16 encoding in URIs

lachlan-roberts:
## Issue #6447

Deprecate support for UTF16 encoding in URIs and add compliance mode to allow UTF16 encodings.


jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -73,9 +73,16 @@
...
-        PARAM
+        PARAM,
+
+        /**
+         * Contains UTF16 encodings
+         */
+        UTF16
     }
 
+    EnumSet<Violation> AMBIGUOUS = EnumSet.complementOf(EnumSet.of(Violation.UTF16));
```
joakime: Is the name `AMBIGUOUS` here still valid?

joakime: In fact, is this EnumSet even used? (I don't see it)

jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -470,8 +470,7 @@ public static String decodePath(String path, int offset, int length)
                             char u = path.charAt(i + 1);
                             if (u == 'u')
                             {
-                                // TODO remove %u support in jetty-10
-                                // this is wrong. This is a codepoint not a char
+                                // UTF16 encoding is only supported with UriCompliance.Violation.UTF16_ENCODINGS
```
joakime: What about the comment about removing %u support on line 560 in this file too?

https://github.com/eclipse/jetty.project/blob/a09d9e94b5cc542b8436f9ef3171da42cfc3b13b/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java#L560-L562

joakime (changes requested):
Minor nits about unused fields and javadoc comments.

jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: This comment doesn't seem to apply here, as `UriCompliance.Violation.UTF16_ENCODINGS` are not even used in this URIUtil class?

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
gregw: It is still the right name, as it is the subset that is ambiguous.  It was used in a version... but not now... Unless there is an efficient way to get the intersection of 2 enum sets

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -1033,8 +1080,9 @@ private void parse(State state, final String uri)
                                 break;
                             case '%':
                                 // must have be in an encoded path
```
janbartel:
```suggestion
                                // must have been in an encoded path
```

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -938,9 +982,11 @@ private void parse(State state, final String uri)
             int mark = 0; // the start of the current section being parsed
             int pathMark = 0; // the start of the path section
             int segment = 0; // the start of the current segment within the path
-            boolean encoded = false; // set to true if the path contains % encoded characters
+            boolean encodedPath = false; // set to true if the path contains % encoded characters
+            boolean encodedUtf16 = false; // Is the current encoding for UTF16?
+            int encodedCharacters = 0; // partial state of parsing a % encoded character<x>
+            int encodedValue = 0; // the partial encoded value
             boolean dot = false; // set to true if the path containers . or .. segments
```
janbartel:
```suggestion
            boolean dot = false; // set to true if the path contains . or .. segments
```

jetty-http/src/test/java/org/eclipse/jetty/http/HttpURITest.java
```diff
@@ -323,109 +327,134 @@ public void testCanonicalDecoded()
         return Arrays.stream(new Object[][]
             {
                 // Simple path example
-                {"http://host/path/info", "/path/info", EnumSet.noneOf(Ambiguous.class)},
-                {"//host/path/info", "/path/info", EnumSet.noneOf(Ambiguous.class)},
-                {"/path/info", "/path/info", EnumSet.noneOf(Ambiguous.class)},
+                {"http://host/path/info", "/path/info", EnumSet.noneOf(Violation.class)},
```
janbartel, line 330: Unrelated to UTF16, but we don't seem to have any tests where the scheme name contains a ".",  which is a legal character.
Or percent encoded host names?

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
lachlan-roberts: To work out if we have an ambiguous violation we do
```java
!_violations.isEmpty() && !(_violations.size() == 1 && _violations.contains(Violation.UTF16))
```

So we never use this `AMBIGUOUS` except in testing. Should I inline its usage and remove it?

jetty-http/src/test/java/org/eclipse/jetty/http/HttpURITest.java
lachlan-roberts: I have added some extra tests for this.

joakime (approved):

PR #6467, opened 2021-06-24
Issue #6447 - Deprecate support for UTF16 encoding in URIs

lachlan-roberts:
## Issue #6447

- Merge from PR #6457.
- Also brought some other ComplianceModes back to disable ambiguous empty segments, and ambiguous encodings.


gregw (approved):

gregw:
I'm a bit wary about changing compliance defaults, but let's put some clear instructions in the release announcement

lachlan-roberts:
> I'm a bit wary about changing compliance defaults, but let's put some clear instructions in the release announcement

This does not include new compliance defaults, it excludes the new compliance options from the current defaults so it does not change behaviour.

gregw:
I'll merge this since the ECA check is broken.

PR #6471, opened 2021-06-24
9.4.x: Wait for handshake to be done before closing the client socket

### LINKED ISSUE #6443: Flakey test SslBytesServerTest.testHandshakeWithResumedSessionThenClose
**Jetty 11.0.x**

This test is failing in my branch for PR #6428 - needless to say that PR does not change anything related to this test.

lorban:
Fixes #6443 in branch 9.4.x.

### DIFF
--- a/jetty-client/src/test/java/org/eclipse/jetty/client/ssl/SslBytesServerTest.java
+++ b/jetty-client/src/test/java/org/eclipse/jetty/client/ssl/SslBytesServerTest.java
@@ -334,7 +334,7 @@ public class SslBytesServerTest extends SslBytesTest
 
         final SSLSocket client2 = newClient(proxy);
 
-        threadPool.submit(() ->
+        Future<Object> handshakeFuture = threadPool.submit(() ->
         {
             client2.startHandshake();
             return null;
@@ -366,6 +366,9 @@ public class SslBytesServerTest extends SslBytesTest
         // Client Done
         TLSRecord doneRecord = proxy.readFromClient();
         assertNotNull(doneRecord);
+
+        // Wait for socket to be done with handshake
+        handshakeFuture.get(5, TimeUnit.SECONDS);
         // Close
         client2.close();
         TLSRecord closeRecord = proxy.readFromClient();


sbordet:
I don't think this is correct. When the client closes it must produce a close_notify and send it to the server. It's the reply to this close_notify from the server that is optional, although we want to try hard and reply correctly. Also I would double check if this is still true in RFC 8446.

lorban:
@sbordet you're right; I mis-read the RFC that still says:
```
   Unless some other fatal alert has been transmitted, each party is
   required to send a close_notify alert before closing the write side
   of the connection.
```

So here's my next theory: the client's socket is closed after the server read the `doneRecord`. But it could be that the client code hasn't yet returned from its blocking write and still needs to set a flag to mark the handshake as done.

Before the thread that performs the handshake has a chance to set this "handshake done" flag, the other thread could call `close` on the `SSLSocket` and interpret that as a close during the handshake, hence not requiring the alert packet to be sent.

I've added a `handshakeFuture.get()` before the call to close and can't reproduce the problem since.

WDYT?

jetty-client/src/test/java/org/eclipse/jetty/client/ssl/SslBytesServerTest.java
```diff
@@ -366,6 +366,9 @@ public void testHandshakeWithResumedSessionThenClose() throws Exception
         // Client Done
         TLSRecord doneRecord = proxy.readFromClient();
         assertNotNull(doneRecord);
+
+        // Wait for socket to be done with handshake
+        handshakeFuture.get();
```
sbordet: Don't wait forever.

sbordet (changes requested):

jetty-client/src/test/java/org/eclipse/jetty/client/ssl/SslBytesServerTest.java
lorban: Sure, I've added a 5 seconds timeout.

sbordet (dismissed):

sbordet (approved):

PR #6474, opened 2021-06-26
Jetty 9.4.x #6473 normalize only once

### LINKED ISSUE #6473: Improve alias checking in PathResource
**Target Jetty version(s)**
9.10,11

**Enhancement Description**

Currently multiple calls to canonicalPath are done, which is inefficient repeat scanning. Include normalization in alias checks, as was previously done in FileResource, so that only a single normalization pass is needed.

gregw:
Fix #6473 Reduce multiple canonicalPath calls with single alias check in PathResource

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
@@ -36,22 +36,45 @@ import org.eclipse.jetty.util.UrlEncoded;
 /**
  * Http URI.
  * Parse an HTTP URI from a string or byte array.  Given a URI
- * <code>http://user@host:port/path/info;param?query#fragment</code>
- * this class will split it into the following undecoded optional elements:<ul>
+ * <code>http://user@host:port/path;param1/%2e/info;param2?query#fragment</code>
+ * this class will split it into the following optional elements:<ul>
  * <li>{@link #getScheme()} - http:</li>
  * <li>{@link #getAuthority()} - //name@host:port</li>
  * <li>{@link #getHost()} - host</li>
  * <li>{@link #getPort()} - port</li>
- * <li>{@link #getPath()} - /path/info</li>
- * <li>{@link #getParam()} - param</li>
+ * <li>{@link #getPath()} - /path;param1/%2e/info;param2</li>
+ * <li>{@link #getDecodedPath()} - /path/info</li>
+ * <li>{@link #getParam()} - param2</li>
  * <li>{@link #getQuery()} - query</li>
  * <li>{@link #getFragment()} - fragment</li>
  * </ul>
  *
- * <p>Any parameters will be returned from {@link #getPath()}, but are excluded from the
- * return value of {@link #getDecodedPath()}.   If there are multiple parameters, the
- * {@link #getParam()} method returns only the last one.
- */
+ * <p>The path part of the URI is provided in both raw form ({@link #getPath()}) and
+ * decoded form ({@link #getDecodedPath}), which has: path parameters removed,
+ * percent encoded characters expanded and relative segments resolved.  This approach
+ * is somewhat contrary to <a href="https://tools.ietf.org/html/rfc3986#section-3.3">RFC3986</a>
+ * which no longer defines path parameters (removed after
+ * <a href="https://tools.ietf.org/html/rfc2396#section-3.3">RFC2396</a>) and specifies
+ * that relative segment normalization should take place before percent encoded character
+ * expansion. A literal interpretation of the RFC can result in URI paths with ambiguities
+ * when viewed as strings. For example, a URI of {@code /foo%2f..%2fbar} is technically a single
+ * segment of "/foo/../bar", but could easily be misinterpreted as 3 segments resolving to "/bar"
+ * by a file system.
+ * </p>
+ * <p>
+ * Thus this class avoid and/or detects such ambiguities. Furthermore, by decoding characters and
+ * removing parameters before relative path normalization, ambiguous paths will be resolved in such
+ * a way to be non-standard-but-non-ambiguous to down stream interpretation of the decoded path string.
+ * The violations are recorded and available by API such as {@link #hasViolation(Violation)} so that requests
+ * containing them may be rejected in case the non-standard-but-non-ambiguous interpretations
+ * are not satisfactory for a given compliance configuration.   Implementations that wish to
+ * process ambiguous URI paths must configure the compliance modes to accept them and then perform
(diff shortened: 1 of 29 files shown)


lachlan-roberts (dismissed):

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -36,22 +36,42 @@
...
+ * this class will split it into the following optional elements:<ul>
  * <li>{@link #getScheme()} - http:</li>
  * <li>{@link #getAuthority()} - //name@host:port</li>
  * <li>{@link #getHost()} - host</li>
  * <li>{@link #getPort()} - port</li>
- * <li>{@link #getPath()} - /path/info</li>
- * <li>{@link #getParam()} - param</li>
+ * <li>{@link #getPath()} - /path;param1/%2e/info;param2</li>
+ * <li>{@link #getDecodedPath()} - /path/info</li>
+ * <li>{@link #getParam()} - param2</li>
```
janbartel, line 47: Needs an explanation of why we don't return `param1`.

jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -780,143 +781,128 @@ public static String parentPath(String p)
...
      * </p>
      *
-     * @param path the path to convert, decoded, with path separators '/' and no queries.
+     * @param path the encoded URI from the path onwards, which may contain query strings and/or fragments
      * @return the canonical path, or null if path traversal above root.
+     * @see #canonicalPath(String)
+     * @see #canonicalURI(String)
      */
-    public static String canonicalPath(String path)
+    public static String canonicalURI(String path)
```
janbartel: Is the param a path or a uri?  If it's a uri, then the param should be called `uri`. Also, there are so many decodeds, encodeds flying around, that it might be clearer to name the method, rather than just relying on the javadoc of the param.  Suggestion:  `canonicalencodedURI`.

jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -780,143 +781,128 @@ public static String parentPath(String p)
...
      * </p>
      *
-     * @param path the path to convert (expects URI/URL form, encoded, and with path separators '/')
+     * @param path the decoded URI path to convert. Any special characters (e.g. '?', "#") are assumed to be part of
+     * the path segments.
      * @return the canonical path, or null if path traversal above root.
+     * @see #canonicalURI(String)
      */
-    public static String canonicalEncodedPath(String path)
+    public static String canonicalPath(String path)
```
janbartel, line 905: As above, it would be clearer to call this method `canonicalDecodedPath` - or to keep the api the same, add this method and make `canonicalPath` deprecated and refer to `canonicalDecodedPath`.

jetty-util/src/main/java/org/eclipse/jetty/util/resource/URLResource.java
```diff
@@ -272,9 +272,7 @@ public Resource addPath(String path)
         throws IOException
     {
         if (path == null)
-            return null;
-
-        path = URIUtil.canonicalPath(path);
+            throw new MalformedURLException("null path");
```
janbartel, line 275: Is this change from returning null  to MalformedURLException tested?

jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -300,7 +300,8 @@ public Resource addPath(String path) throws IOException
         {
             return new ResourceCollection(resources.toArray(new Resource[0]));
         }
-        return null;
+
+        throw new MalformedURLException();
```
janbartel, line 304: Is this tested?

janbartel (dismissed):
Some comments and suggestions, but largely looks ok: difficult to review so many changes.

jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
gregw: No, because I can't see how this can ever happen, as either `resource` or `resources` will be non null.
Thus throwing rather than returning null just make sure it is seen as an error if it error does somehow get there

jetty-util/src/main/java/org/eclipse/jetty/util/resource/URLResource.java
gregw: yes - The main resource classes do not return null and few calls to addPath ever expect a null return.
So this is just making it more uniform.

jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
gregw: These are different methods that do different things.  This is definitely a URI which is never decoded.
I will rename the argument and improve the javadoc

jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
gregw: It is a uri. I will rename

jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
gregw: I'm not keen to rename this method.  There is now only `canonicalPath` and `canonicalUri`, with javadoc making it clear what they are.

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
gregw: added

janbartel (dismissed):

jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -93,9 +92,11 @@ private final Path checkAliasPath()
         }
 
         if (!abs.isAbsolute())
-        {
             abs = path.toAbsolutePath();
-        }
+
+        Path normal = path.normalize();
+        if (!abs.equals(normal))
+            return normal;
```
lachlan-roberts, line 99: What happens if the normalized path is to a symlink? Do we want to do the code below to resolve the target of the symlink?

janbartel (dismissed):

lachlan-roberts (approved):

PR #6477, opened 2021-06-28
Issue #6473 - Improve alias checking in PathResource.

sbordet:
## Issue #6473

* Reverted %-escape handling for URI query parts.
* Performing canonicalization in ServletContext.getResource(),
  and improving alias checking in ContextHandler.getResource().
* Performing canonicalization checks in Resource.addPath() to avoid
  navigation above of the root.
* Test added and fixed.
* Various cleanups.

Signed-off-by: Simone Bordet <simone.bordet@gmail.com>

PR #6478, opened 2021-06-29
Issue #6473 - canonicalPath refactor & fix alias check in PathResource (Jetty-10)

lachlan-roberts:
## Issue #6473

Merge of PR #6474 and PR #6477 for Jetty-10.

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
@@ -34,7 +34,7 @@ import org.eclipse.jetty.util.UrlEncoded;
  *
  * A URI such as
  * {@code http://user@host:port/path;param1/%2e/info;param2?query#fragment}
- * are split into the following optional elements:<ul>
+ * is split into the following optional elements:<ul>
  * <li>{@link #getScheme()} - http:</li>
  * <li>{@link #getAuthority()} - //name@host:port</li>
  * <li>{@link #getHost()} - host</li>
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
@@ -15,7 +15,10 @@ package org.eclipse.jetty.http;
 
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.EnumSet;
+import java.util.Set;
 
 import org.eclipse.jetty.util.HostPort;
 import org.eclipse.jetty.util.Index;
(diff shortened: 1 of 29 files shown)


jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -31,54 +33,94 @@
  * via the static methods such as {@link #build()} and {@link #from(String)}.
  *
  * A URI such as
- * <code>http://user@host:port/path;ignored/info;param?query#ignored</code>
- * is split into the following undecoded elements:<ul>
+ * {@code http://user@host:port/path;param1/%2e/info;param2?query#fragment}
+ * this class will split it into the following optional elements:<ul>
```
gregw: The grammar of this sentence is split between old and new version:

gregw (changes requested):

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -34,7 +34,7 @@
  *
  * A URI such as
  * {@code http://user@host:port/path;param1/%2e/info;param2?query#fragment}
- * this class will split it into the following optional elements:<ul>
+ * are split into the following optional elements:<ul>
```
gregw:
```suggestion
 * is split into the following optional elements:<ul>
```

gregw (changes requested):
sorry to be pendantic

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -31,54 +33,94 @@
...
+ * <p>
+ * If there are multiple path parameters, only the last one is returned by {@link #getParam()}.
+ * </p>
+ **/
 public interface HttpURI
 {
+    /**
+     * Violations of safe URI interpretations
+     */
     enum Violation
```
sbordet: This class was called `Ambiguous` so this is a breaking change.

Also, why does it not implement `ComplianceViolation`, which seems to summarize the violation concept?

On a closer look, I think this class here is a complete duplicate of `UriCompliance.Violation`.
It's here because in 9.4.x we don't have `UriCompliance` (so it's a merge side-effect).
Should it not be completely removed, so that in 10 we only use `UriCompliance.Violation`?

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -166,15 +213,26 @@ static Immutable from(String scheme, String host, int port, String pathQuery)
...
+     * @param violation the violation to check.
+     * @return true if the URI has the passed violation.
+     */
+    boolean hasViolation(Violation violation);
+
+    /**
+     * @return Set of violations in the URI.
+     */
+    Collection<Violation> getViolations();
+
```
sbordet, line 188: See comment above, I think methods that expose `Violation` should be removed or changed to use `UriCompliance.Violation`.

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -1216,6 +1298,10 @@ else if (c == '/')
                             {
                                 switch (encodedValue)
                                 {
+                                    case 0:
+                                    // Byte 0 cannot be present in a UTF-8 sequence in any position
+                                    // other than as the NUL ASCII byte which we do not wish to allow.
```
sbordet: Fix indentation.

jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
```diff
@@ -231,6 +227,11 @@ public static UriCompliance from(String spec)
             boolean exclude = element.startsWith("-");
             if (exclude)
                 element = element.substring(1);
+
+            // Ignore removed name. TODO: remove in future release.
+            if (element.equals("NON_CANONICAL_AMBIGUOUS_PATHS"))
+                continue;
+
```
sbordet, line 232: This block should be removed, because the constant `NON_CANONICAL_AMBIGUOUS_PATHS` was only present in PRs, never made into actual releases.

jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
```diff
@@ -330,4 +331,47 @@ public String toString()
...
+    public static String checkUriCompliance(UriCompliance compliance, HttpURI uri)
+    {
+        for (HttpURI.Violation violation : HttpURI.Violation.values())
+        {
+            if (uri.hasViolation(violation) && (compliance == null || !compliance.allows(__uriViolations.get(violation))))
+                return violation.getMessage();
+        }
+        return null;
+    }
```
sbordet, line 341: See previous comment. This mapping from `HttpURI.Violation` to `UriCompliance.Violation` should not be necessary.
`HttpURI` should just use `UriCompliance.Violation`.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -2305,6 +2308,8 @@ public InputStream getResourceAsStream(String path)
         @Override
         public Set<String> getResourcePaths(String path)
         {
+            if (path == null)
+                return null;
```
sbordet, line 2321: Should not the logic here be similar to `getResource()`? I think we should canonicalize here too.

And also in `getRealPath()`.

I think we need to review the `ServletContext` implementation more carefully.

jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
```diff
@@ -1713,15 +1713,6 @@ public void testAmbiguousPaths() throws Exception
...
             containsString("pathInfo=/path/info")));
-
-        _connector.getBean(HttpConnectionFactory.class).getHttpConfiguration().setUriCompliance(UriCompliance.from(EnumSet.of(
-            UriCompliance.Violation.AMBIGUOUS_PATH_SEPARATOR,
-            UriCompliance.Violation.AMBIGUOUS_PATH_SEGMENT,
-            UriCompliance.Violation.AMBIGUOUS_PATH_PARAMETER,
-            UriCompliance.Violation.NON_CANONICAL_AMBIGUOUS_PATHS)));
-        assertThat(_connector.getResponse(request), Matchers.allOf(
-            startsWith("HTTP/1.1 200"),
-            containsString("pathInfo=/path/ambiguous/.././info")));
```
sbordet, line 1724: Seems like a legit test. Why was it removed?

jetty-servlet/src/main/java/org/eclipse/jetty/servlet/DefaultServlet.java
```diff
@@ -420,8 +419,7 @@ else if (_servletContext instanceof ContextHandler.Context)
             }
             else
             {
-                URL u = _servletContext.getResource(pathInContext);
-                r = _contextHandler.newResource(u);
```
sbordet, line 424: Why was this changed? Can `_servletContext` never be other than `ContextHandler.Context`?

jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -471,6 +471,7 @@ public static String decodePath(String path, int offset, int length)
                             if (u == 'u')
                             {
                                 // UTF16 encoding is only supported with UriCompliance.Violation.UTF16_ENCODINGS.
+                                // This is wrong. This is a codepoint not a char
```
sbordet, line 474: This needs to be fixed then? 9.4.x too?

sbordet (changes requested):

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
gregw: Ambiguous was public for a few releases, but not used in any interfaces at all, so I think it is OK to change.
But worth investigating if we can actually use UriCompliance.Violation here....

jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
gregw: are we sure... if so, then yes

jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
gregw: becaues the compliance mode it was testing no longer exists.

jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
gregw: Nope - this is just how we implemented %u.  It is not strictly correct, but let's not improve a feature we really want to remove.

jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
sbordet: Doh! I was wrong, keep this code, as the constant was in 10.0.5.

PR #6484, opened 2021-06-30
Update release-template.md

gregw:
Add a notification step in testing phase of the release issue process.

### DIFF
--- a/.github/ISSUE_TEMPLATE/release-template.md
+++ b/.github/ISSUE_TEMPLATE/release-template.md
@@ -33,6 +33,7 @@ This release process will produce releases:
 - [ ] Assign issue to "test manager", who will oversee the testing of the staged releases.
    + [ ] Add [testing tick list](https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/.github/test-ticklist.md) to the target [project(s)](https://github.com/eclipse/jetty.project/projects).
    + [ ] Testing tick lists complete in all target project(s).
+   + [ ] Notify interested parties and invite testing of the staged release(s).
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/release-template.md
```diff
@@ -33,6 +33,7 @@ This release process will produce releases:
 - [ ] Assign issue to "test manager", who will oversee the testing of the staged releases.
    + [ ] Add [testing tick list](https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/.github/test-ticklist.md) to the target [project(s)](https://github.com/eclipse/jetty.project/projects).
    + [ ] Testing tick lists complete in all target project(s).
+   + [ ] Notify interested parties and invite testing of the staged release(s).
```
sbordet, line 36: Should not this be before the "Testing tick lists complete..." item?

I think it's another reason I liked the tick list to _not_ be separate -- now you have to look at another place to see if that "Notify interested parties..." was not already specified in the testing tick list.

sbordet (changes requested):
The addition

.github/ISSUE_TEMPLATE/release-template.md
gregw: I think we should do our own testing before asking others to test.  But then we don't need to complete all our testing.  It's kind of in parallel.

I agree about 1 list rather than 2, but that's another issue. Feel free to raise another PR for that.

WalkerWatch (approved):

sbordet (approved):

PR #6485, opened 2021-06-30
Issue #6470 - remove ByteBufferPool usage from MessageInputStream

lachlan-roberts:
#6470 
Signed-off-by: Lachlan Roberts <lachlan@webtide.com>

Issue #6470, opened 2021-06-24
java.nio.ReadOnlyBufferException


Horcrux7:
**Jetty version(s)**: 9.4.42.v20210604

**Java version/vendor** `(use: java -version)`
openjdk version "11.0.11" 2021-04-20
OpenJDK Runtime Environment AdoptOpenJDK-11.0.11+9 (build 11.0.11+9)
OpenJDK 64-Bit Server VM AdoptOpenJDK-11.0.11+9 (build 11.0.11+9, mixed mode)

**OS type/version**: Alpine 3.11 in a Docker container

**Description**
It looks like that after update from 9.4.40.v20210413 to version 9.4.42.v20210604 we get the follow exception on every second request. It occur also on other OS like Windows and OSX. Any idea?

```
java.nio.ReadOnlyBufferException
	at java.base/java.nio.HeapByteBufferR.put(Unknown Source)
	at org.eclipse.jetty.util.BufferUtil.putCRLF(BufferUtil.java:1251)
	at org.eclipse.jetty.http.HttpGenerator.prepareChunk(HttpGenerator.java:501)
	at org.eclipse.jetty.http.HttpGenerator.committed(HttpGenerator.java:317)
	at org.eclipse.jetty.http.HttpGenerator.generateResponse(HttpGenerator.java:469)
	at org.eclipse.jetty.server.HttpConnection$SendCallback.process(HttpConnection.java:744)
	at org.eclipse.jetty.util.IteratingCallback.processing(IteratingCallback.java:241)
	at org.eclipse.jetty.util.IteratingCallback.iterate(IteratingCallback.java:223)
	at org.eclipse.jetty.server.HttpConnection.send(HttpConnection.java:550)
	at org.eclipse.jetty.server.HttpChannel.sendResponse(HttpChannel.java:915)
	at org.eclipse.jetty.server.HttpChannel.write(HttpChannel.java:987)
	at org.eclipse.jetty.server.handler.gzip.GzipHttpOutputInterceptor$GzipBufferCB.process(GzipHttpOutputInterceptor.java:416)
	at org.eclipse.jetty.util.IteratingCallback.processing(IteratingCallback.java:241)
	at org.eclipse.jetty.util.IteratingCallback.iterate(IteratingCallback.java:223)
	at org.eclipse.jetty.server.handler.gzip.GzipHttpOutputInterceptor.gzip(GzipHttpOutputInterceptor.java:139)
[...]

PR #6486, opened 2021-06-30
Issue #6470 - prevent EOF being released back into pool in MessageInputStream

lachlan-roberts:
## Issue #6470 

Prevent the static readOnly buffer `EOF` from being released back into the `ByteBufferPool`.

Alternative to #6485


gregw (approved):
That looks better!

PR #6490, opened 2021-07-01
Issue #6489 - fix RFC3986_UNAMBIGUOUS and UNSAFE UriCompliance modes.

### LINKED ISSUE #6489: Some URI valid compliance modes cannot be set in .ini file.
**Jetty version(s)**
10.0.x - 11.0.x

The `RFC3986_UNAMBIGUOUS` and `UNSAFE` URI compliance modes are not able to be set with `jetty.httpConfig.uriCompliance` in the `server.ini` file.

lachlan-roberts:
closes #6489

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
@@ -159,7 +159,7 @@ public final class UriCompliance implements ComplianceViolation.Mode
 
     public static UriCompliance valueOf(String name)
     {
-        for (UriCompliance compliance : Arrays.asList(DEFAULT, LEGACY, RFC3986, STRICT, SAFE))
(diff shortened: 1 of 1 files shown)


sbordet (approved):

jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
```diff
@@ -159,7 +159,7 @@ public String getDescription()
 
     public static UriCompliance valueOf(String name)
     {
-        for (UriCompliance compliance : Arrays.asList(DEFAULT, LEGACY, RFC3986, STRICT, SAFE))
+        for (UriCompliance compliance : Arrays.asList(DEFAULT, LEGACY, RFC3986, RFC3986_UNAMBIGUOUS, UNSAFE, SAFE, STRICT))
```
sbordet: Other `*Compliance` classes have a static `KNOWN_MODES` list, and I think for consistency it should be introduced in this class too.

Nit: try to use `List.of()` instead of `Arrays.asList()`.

sbordet (changes requested):
Can you please

sbordet (approved):

PR #6494, opened 2021-07-05
Reduce header cache memory usage on non persistent requests

### LINKED ISSUE #6493: HttpParser.parsedHeader consume too much memory
**Jetty version(s)**

10.0.5

openjdk version "11.0.2"

**OS type/version**

NAME="CentOS Linux"
VERSION="8 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="8"
PLATFORM_ID="platform:el8"
PRETTY_NAME="CentOS Linux 8 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:8"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

**Description**

HttpParser.parsedHeader consume too much memory than jetty 9.4

**How to reproduce?**

jetty 10.0.5

image

more details see attachment
[admin-20210702-01.zip](https://github.com/eclipse/jetty.project/files/6761409/admin-20210702-01.zip)

gregw:
Fix #6493. Delay creating a header cache until a second request on a parser.

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
@@ -218,6 +218,7 @@ public class HttpParser
     private final int _maxHeaderBytes;
     private final HttpCompliance _complianceMode;
     private final Utf8StringBuilder _uri = new Utf8StringBuilder(INITIAL_URI_LENGTH);
+    private final FieldCache _fieldCache = new FieldCache();
     private HttpField _field;
     private HttpHeader _header;
     private String _headerString;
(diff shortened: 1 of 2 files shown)


lorban (changes requested):
I think the logic is sound, but I don't like all those null checks of the two cache fields being scattered all around this class. It's probably worth extracting some of that logic into another `FieldCache` class IMHO.

lorban (approved):
This looks a million times better!

jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
```diff
@@ -2007,4 +2004,87 @@ private IllegalCharacterException(State state, HttpTokens.Token token, ByteBuffe
...
+        {
+            if (_cache == null && _cacheableFields != null)
+            {
+                _cache = Index.buildMutableVisibleAsciiAlphabet(_caseSensitive, _size);
+                for (HttpField f : _cacheableFields)
+                {
+                    if (!_cache.put(f))
+                        break;
+                }
+                _cacheableFields.clear();
```
lorban, line 2085: This call to `clear` is useless as you're nulling out the ref on the next line.

gregw: no harm helping out the GC.

PR #6513, opened 2021-07-12
Extensible RetainableByteBufferPool  #6322

### LINKED ISSUE #6322: Use `RetainableByteBuffer` in `HttpConnection`

gregw:
Improves on #6332 for #6322

Add a Log2 bucket size imple for ArrayRetainableByteBufferPool

Even if we don't actually add/use the `LogBucket` class, I think the changes made to make the ArrayRetainableByteBufferPool extensible are good ones to have.


jetty-io/src/main/java/org/eclipse/jetty/io/ArrayRetainableByteBufferPool.java
```diff
@@ -311,4 +334,45 @@ private void evict(boolean direct, long excess)
...
+
+        public LogBuckets(int minCapacity, int maxCapacity, int maxBucketSize, long maxHeapMemory, long maxDirectMemory)
+        {
+            super(minCapacity,
+                -1,
+                maxCapacity,
+                maxBucketSize,
+                maxHeapMemory,
+                maxDirectMemory,
+                TypeUtil::log2NextPowerOf2,
```
lorban: I would not change the existing buckets-sizing algorithm without measurements. Heck, I'm not even convinced making it configurable (even only in code via an internal API) is a good idea.

We currently have little visibility over how buckets are sized in the wild so I'm worried that changing this may lead to adverse effects.

gregw:
@lorban The intention is not to change the existing behaviour.  I don't think I have and the CI failure appears to be exactly that.... a CI failure.  Will redo the CI tests to see.

But I'm a strong advocate of making this class extensible, even if we don't add  the badly named `LogBuckets`.  Unless the class is extensible then we won't ever be able to do any comparisons.
Also I'll include in this making the pool dumpable so that we can see a lot more detail of how the buckets are being used (regardless of allocation).  We can then get some dumps from real live servers and do a calculation to see how alternate bucket strategies might affect the pattern we see.

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayRetainableByteBufferPool.java
gregw: We need to make changes like this PR so we can see how buckets are sized in the wild.     I'm definitely not proposing that `LogBuckets` be used until we see the data, but making the pool Dumpable, sampleable and extensible are all worthwhile things.

gregw:
@sbordet @lorban ignoring the `LogBucket` class itself, I think this PR has some good improvements for:
- code clarity with the `Bucket` sub class 
- setting defaults (not duplicated in constructors)
- dumpability so we can see the distribution in buckets and how buffers are being used
- extensibility, we can have a `LogBuckets` if it proves necessary.

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayRetainableByteBufferPool.java
sbordet: I'm not against the power of 2 idea, but perhaps it should be stripped out to maintain, as you say, dumpability and statistics.

sbordet (changes requested):

gregw:
@sbordet The problem with stripping it out, is that we then have no way of actually testing that the class is really extensible.    There were a lot of little assumptions in the code that depended on linear buckets. This PR fixes them, but if we took out  LogBuckets, it would be easy for a change to revert to having such a assumptions.    I could move LogBuckets to test???  But I really don't see any harm having it available, as it will allow us to test with it in real world deployments.

jetty-util/src/main/java/org/eclipse/jetty/util/TypeUtil.java
```diff
@@ -800,4 +800,42 @@ public T next()
     {
         return StreamSupport.stream(new ServiceLoaderSpliterator<>(serviceLoader), false);
     }
+
+    /**
+     * Round up to the next power of 2.
+     * @param v An integer &gt; 0 and &lt;= half of {@link Integer#MAX_VALUE}
+     * @return The next power of two that is equal too or larger than the passed integer.
+     */
+    public static int nextPowerOf2(int v)
```
sbordet: Can you please check whether you can do this with `Long.highestOneBit()`?

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayRetainableByteBufferPool.java
```diff
@@ -311,4 +359,51 @@ private void evict(boolean direct, long excess)
...
+            return String.format("%s{capacity=%d}", super.toString(), _capacity);
+        }
+    }
+
+    /**
+     * A variant of the {@link ArrayRetainableByteBufferPool} that
+     * uses buckets of buffers that increase in size by a power of
+     * 2 (eg 1k, 2k, 4k, 8k, etc.).
+     */
+    public static class LogBuckets extends ArrayRetainableByteBufferPool
```
sbordet: I like better `PowerOfTwoStrategy` or `ExponentialStrategy`.

sbordet (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
```diff
@@ -55,4 +60,56 @@ static RetainableByteBufferPool findOrAdapt(Container container, ByteBufferPool
...
+     * A Pool wrapper that collects sample statistics on the acquired
+     * buffers.
+     * @see SampleStatistic
+     */
+    @ManagedObject
+    class SampledPool implements RetainableByteBufferPool, Dumpable
+    {
+        private final RetainableByteBufferPool _pool;
+        private final SampleStatistic _heap = new SampleStatistic();
+        private final SampleStatistic _direct = new SampleStatistic();
```
lorban: I think those two counters are going to be a bit meaningless at the pool layer. IMHO it would be a lot more informative to move them to the `Bucket` layer, maybe by creating a `DumpableBucket` class.

lorban (changes requested):
I quite like the idea behind this change. More visibility and extensibility is always good to have. I just think that the buckets should be sampled instead of the pool.

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayRetainableByteBufferPool.java
```diff
@@ -52,48 +57,74 @@ public ArrayRetainableByteBufferPool(int minCapacity, int factor, int maxCapacit
...
             throw new IllegalArgumentException("The capacity factor must be a divisor of maxCapacity");
 
-        int length = maxCapacity / _factor;
+        if (bucketIndexFor == null)
+            bucketIndexFor = c -> (c - 1) / f;
+        if (bucketCapacity == null)
+            bucketCapacity = i -> (i + 1) * f;
 
+        int length = bucketIndexFor.apply(maxCapacity) + 1;
         @SuppressWarnings("unchecked")
```
sbordet: Unnecessary suppression.

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayRetainableByteBufferPool.java
```diff
@@ -52,48 +57,74 @@ public ArrayRetainableByteBufferPool(int minCapacity, int factor, int maxCapacit
...
-        Pool<RetainableByteBuffer>[] directArray = new Pool[length];
+        Bucket[] directArray = new Bucket[length];
         @SuppressWarnings("unchecked")
```
sbordet: Unnecessary suppression.

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayRetainableByteBufferPool.java
```diff
@@ -52,48 +57,74 @@ public ArrayRetainableByteBufferPool(int minCapacity, int factor, int maxCapacit
...
+                                            Function<Integer, Integer> bucketIndexFor, Function<Integer, Integer> bucketCapacity)
+    {
         if (minCapacity <= 0)
             minCapacity = 0;
-        _minCapacity = minCapacity;
         if (maxCapacity <= 0)
             maxCapacity = 64 * 1024;
-        if ((maxCapacity % _factor) != 0 || _factor >= maxCapacity)
+
+        final int f = factor <= 0 ? 1024 : factor;
```
sbordet: Drop `final`, not necessary.

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayRetainableByteBufferPool.java
```diff
@@ -311,4 +358,51 @@ private void evict(boolean direct, long excess)
...
+            return String.format("%s{capacity=%d}", super.toString(), _capacity);
+        }
+    }
+
+    /**
+     * A variant of the {@link ArrayRetainableByteBufferPool} that
+     * uses buckets of buffers that increase in size by a power of
+     * 2 (eg 1k, 2k, 4k, 8k, etc.).
+     */
+    public static class ExponentialPool extends ArrayRetainableByteBufferPool
```
sbordet: Not sure we want to make this `public`, as it binds us to maintain it, etc.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
```diff
@@ -55,4 +60,56 @@ static RetainableByteBufferPool findOrAdapt(Container container, ByteBufferPool
...
         return retainableByteBufferPool;
     }
+
+    /**
+     * A Pool wrapper that collects sample statistics on the acquired
+     * buffers.
+     * @see SampleStatistic
+     */
+    @ManagedObject
+    class SampledPool implements RetainableByteBufferPool, Dumpable
```
sbordet: This class is implicitly `public`, which I don't think it's a good idea.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
```diff
@@ -55,4 +60,56 @@ static RetainableByteBufferPool findOrAdapt(Container container, ByteBufferPool
...
+        private final SampleStatistic _heap = new SampleStatistic();
+        private final SampleStatistic _direct = new SampleStatistic();
+
+        public SampledPool(RetainableByteBufferPool pool)
+        {
+            _pool = pool;
+        }
+
+        @ManagedAttribute("Heap acquire samples")
+        public SampleStatistic getHeapSample()
```
sbordet: Unfortunately `SampleStatistic` is not manageable, so this JMX exposure results in no data being exposed.
I think we need either more specific getters, or MXBean-ify `SampleStatistic` which may not be easy.

I would make this class package private in `ArrayRBBP` for now, and eventually open it up later.

jetty-io/src/test/java/org/eclipse/jetty/io/ArrayRetainableByteBufferPoolTest.java
```diff
@@ -313,4 +315,33 @@ public void testAcquireRelease()
...
+            if (capacity >= ((ArrayRetainableByteBufferPool)pool.getPool()).getMaxCapacity())
+                break;
+
+            b = pool.acquire(capacity + 1, false);
+            assertThat(b.capacity(), Matchers.is(capacity * 2));
+
+            capacity = capacity * 2;
+        }
+
+        System.err.println(pool.dump());
```
sbordet: Remove this?

sbordet (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
gregw: Hmmm good test for extensibility to have SampledBuckets.    Will try....

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayRetainableByteBufferPool.java
gregw: I want to make it public so we can easily try using it in real deployments.    
But as there appears to be resistance to this idea, I'll move to test and we can used the sampled one to gather data to see if `ExponentialPool` is worthwhile or not.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
gregw: We want at least this one to be public, so we can run real deployments and capture data about the pool usage.   You can't simultaneously argue that `ExponentialPool` is not justified by any data and then prevent anything that allows the data to be collected!

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
gregw: If it is private, then we can't use it to gather any info.
Anyway, I'm going to look at sampling on buckets rather than on acquire.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
gregw: Actaully,   thinking about it, we can just run with a small factor and then we will effectively get a histogram of the acquired sizes.    So maybe this class is not needed after all.

gregw:
@lorban @sbordet I have removed `SampledPool` as we get a reasonable histogram of acquire size by using the normal linear pool and dumping the bucket sizes.
I have added a bit more info into the bucket dumps, but as the sampling is in a race, it may be of marginal value on a server underload, but will be OK if dumped after a load test.

I have moved the ExponentialPool to be an inner class of the test.   So process from here is to release this.... ask some real clients do give us some server dumps; analyse the "histograms"; if an ExponentialPool looks like it will work, then re-release with that class public and test.

gregw:
@lorban @sbordet nudge

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayRetainableByteBufferPool.java
```diff
@@ -311,4 +356,53 @@ private void evict(boolean direct, long excess)
...
+                    if (pos > 0 || lim < buffer.capacity())
+                    {
+                        // If limit is not capacity, then flush mode, otherwise fill mode
+                        used += (lim < buffer.capacity()) ? lim : pos;
+                        capacity += buffer.capacity();
+                    }
+                }
+            }
+
+            return String.format("%s{capacity=%d,inuse=%d(%d%%),used=%d%%}",
```
lorban: I'm not sure the `used` calculation will report anything meaningful, a more useful metric would be to calculate the mean 'waste' (i.e.: the difference between the requested size and the returned buffer's capacity) but that would require some extra processing.

I think we should leave this as-is and eventually revisit that later on.

lorban (approved):
Besides the checkstyle failure, LGTM.

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayRetainableByteBufferPool.java
gregw: Agreed. We are kind of thwarted by the fact that we have no idea if a buffer is in fill or flush mode.
At least with this metric, it is sampling buffers that have a pointer between 0 and capacity, which is at least a hint of partial usage.

gregw:
@sbordet nudge!!!

gregw:
Closed in favour of #6538

PR #6516, opened 2021-07-13
cherry-pick hazelcast remote distribution test from 9.4.x (#6068)

olamy:
* Add distribution test using a remote hazelcast only client true and false, false is failing with CNFE

### DIFF
--- a/jetty-hazelcast/src/main/config/modules/session-store-hazelcast-remote.mod
+++ b/jetty-hazelcast/src/main/config/modules/session-store-hazelcast-remote.mod
@@ -24,8 +24,6 @@ lib/hazelcast/hazelcast-${hazelcast.version}.jar
 
 [ini]
 hazelcast.version?=4.1
-jetty.webapp.addSystemClasses+=,org.eclipse.jetty.hazelcast.
-jetty.webapp.addServerClasses+=,-org.eclipse.jetty.hazelcast.
 
 [license]
 Hazelcast is an open source project hosted on Github and released under the Apache 2.0 license.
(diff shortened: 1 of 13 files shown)


jetty-hazelcast/src/main/config/modules/session-store-hazelcast-remote.mod
```diff
@@ -14,18 +14,18 @@ sessions
...
 maven://com.hazelcast/hazelcast/${hazelcast.version}|lib/hazelcast/hazelcast-${hazelcast.version}.jar
-maven://com.hazelcast/hazelcast-client/${hazelcast.version}|lib/hazelcast/hazelcast-client-${hazelcast.version}.jar
 
 [xml]
 etc/sessions/hazelcast/remote.xml
 
 [lib]
 lib/jetty-hazelcast-${jetty.version}.jar
 lib/hazelcast/hazelcast-${hazelcast.version}.jar
-lib/hazelcast/hazelcast-client-${hazelcast.version}.jar
```
olamy, line 25: interesting to find a bug we have here with this test ;)

olamy:
@janbartel nudge :)

jetty-hazelcast/src/main/config/modules/session-store-hazelcast-remote.mod
```diff
@@ -14,18 +14,18 @@ sessions
...
 
 [ini]
 hazelcast.version?=4.1
+jetty.webapp.addSystemClasses+=,org.eclipse.jetty.hazelcast.
+jetty.webapp.addServerClasses+=,-org.eclipse.jetty.hazelcast.
```
janbartel: I'm not sure this is totally necessary as org.eclipse.jetty package is a server class by default, isn't it?

janbartel (dismissed):
I'm a little cautious about making these system classes, just because we haven't made any of the other session integrations system classes, but we _have_ made most of their dependencies server classes.

jetty-hazelcast/src/main/config/modules/session-store-hazelcast-remote.mod
olamy: ah yup good catch. not needed for 10.

olamy:
@janbartel you need to approve again (sorry for that but that's the github process...)

janbartel (approved):

PR #6517, opened 2021-07-14
Issue #6515 - Adding Test to ensure that ClasspathPattern doesn't need prepend anymore

### LINKED ISSUE #6515: multiple version of same jars with setParentLoaderPriority(false) in Embedded Jetty causing issue
**Jetty version(s)** **9.4.40v20210413**

**Java version/vendor** `(use: java -version)` **openjdk version "11.0.9"**

**OS type/version** **AIX 7.1.5**

**Description**: We are using embedded jetty and we have several vendors deploying war files to this jetty webserver. We are seeing some class loading issues. The webapp.setParentLoaderPriority(false); Which means the classes contained within WEB-INF/lib or WEB-INF/classes have priority over classes on the parent classloader. But we have a scenario where we have 2 version of same jar file say xx.jar ( xx.v2jar and xxv1.jar). The xxv1.jar is bundled within the war file (i.e. is present in WEB-INF/lib). The other version (xxv2.jar, a higher version) is present in the class path of the webserver. But while deploying the war file the loader looks for the some dependencies present in the classpath while the dependency is already present in WEB-INF/lib which causes a conflict and war file is not deployed successfully. I couldn't understand this behavior. As per the documentation shouldn't is pick up all it dependencies from WEB-INF/lib. When I remove the dependency(xxv2.jar) from the webserver classpath the was picks up all the dependencies from WEB-INF/lib and the war file is deployed successfully. Am I missing something. Or is there any gap in my understanding.

**How to reproduce?**

joakime:
Just a simple testcase for #6515 

### DIFF
--- a/jetty-webapp/src/test/java/org/eclipse/jetty/webapp/ClasspathPatternTest.java
+++ b/jetty-webapp/src/test/java/org/eclipse/jetty/webapp/ClasspathPatternTest.java
@@ -127,6 +127,17 @@ public class ClasspathPatternTest
         assertTrue(_pattern.match("org.example.Anything$Else"));
     }
 
+    @Test
+    public void testMatchFundamentalExcludeSpecific()
+    {
+        _pattern.clear();
+        _pattern.add("javax.");
+        _pattern.add("-javax.ws.rs.", "-javax.inject.");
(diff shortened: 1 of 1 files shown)


gregw (approved):
needs to be in 10 also

PR #6518, opened 2021-07-15
Work in progress for Resource cleanup

joakime:
Responsibilities:

* Resource.addPath(String) should enforce base / child behaviors
* Accessing Resources via HTTP should support proper URI encoding.


jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -418,11 +418,24 @@ public abstract boolean renameTo(Resource dest)
...
+     *         <li>The path provided will not be URI decoded by the Resource implementation.</li>
+     *         <li>URI canonicalization will be attempted before being used against the Resource.
+     *             This can result in a normal Resource, perhaps with it being an alias to a different Resource,
+     *             perhaps even rejected with an Exception if there is an attempt to navigate above this Resource.</li>
+     *         <li>The path provided will be given to the Resource implementation for interpretation on how it
+     *             will behave (such as using "a\\b.txt" on a Path based in linux vs windows).</li>
+     *         <li>The path provided is ensured to be relative and belongs within the Resource, attempts to reference a path
+     *             outside of the Resource will result in an exception.</li>
+     *     </ul>
+     * </p>
```
joakime, line 434: @gregw @lachlan-roberts I could use feedback on this clarification on addPath

jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -418,11 +418,24 @@ public abstract boolean renameTo(Resource dest)
     /**
      * Returns the resource contained inside the current resource with the
      * given name, which may or may not exist.
+     * <p>
+     *     Constraints:
+     *     <ul>
+     *         <li>The path is expected to be in URI path-like format.</li>
```
gregw, line 424: This "URI path-like" format is the ambiguous decoded path space that is returned from `getPathInfo` etc.
I think it is this space that needs to be better defined, which is part of the conversation at: https://github.com/eclipse-ee4j/servlet-api/issues/18#issuecomment-879769757

The path is "URI-like" in that it is `/` separated segments, but it is not a URI as it cannot contain encoded characters, specifically it cannot contain an encoded `/`, so there are some paths that cannot be specified.

jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -418,11 +418,24 @@ public abstract boolean renameTo(Resource dest)
...
+     *         <li>The path provided will not be URI decoded by the Resource implementation.</li>
```
gregw, line 425: I think this is a bit of a convoluated way of saying that the path is decoded.  Ie any % characters will not result in further decoding, but will themselves be encoded

jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -418,11 +418,24 @@ public abstract boolean renameTo(Resource dest)
...
+     *         <li>URI canonicalization will be attempted before being used against the Resource.
+     *             This can result in a normal Resource, perhaps with it being an alias to a different Resource,
+     *             perhaps even rejected with an Exception if there is an attempt to navigate above this Resource.</li>
```
gregw, line 428: I don't think this is exactly correct. Rather the path is checked to see if conanicalized, would it navigate above the base and if so an MUE us thrown.  Otherwise the possibly non-canonical path is used, which may result in a non-normal Resource which will return true from isAlias

jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -418,11 +418,24 @@ public abstract boolean renameTo(Resource dest)
...
+     *         <li>The path provided will be given to the Resource implementation for interpretation on how it
+     *             will behave (such as using "a\\b.txt" on a Path based in linux vs windows).</li>
```
gregw, line 430: Hmmm probably true... but really not very desirable.
Ideally `\` would never be treated as a separator as the separator is '/' in this space.

jetty-util/src/test/java/org/eclipse/jetty/util/resource/ResourceAddPathTest.java
```diff
@@ -0,0 +1,231 @@
...
+import org.eclipse.jetty.util.TypeUtil;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+public class ResourceAddPathTest
```
gregw, line 37: I'd much rather test cases be added to the existing `ResourceTest` rather than create a special purpose unit test.    By creating multiple special purpose test that overlap with other test cases it just becomes too complex to see what is tested and then gaps emerge.   We have already gone down this path to some extent as we have `ResourceTest`, `FileSystemResourceTest`, `PathResourceTest`, `ResourceAliasTest`.  

We need to simplify and rationalize the testing rather than duplicate tests, so that `ResourceTest` does all the tests it can over all the different implementations of `Resource`.  Then we can have a `PathResourceTest` iff there are some tests that can only be done on PathResources and not other resources in general.    We should strive to get rid of  `ResourceAliasTest`, `FileSystemResourceTest` and definitely not add `ResourceAddPathTest`

jetty-util/src/main/java/org/eclipse/jetty/util/PathUtil.java
```diff
@@ -0,0 +1,157 @@
...
+
+package org.eclipse.jetty.util;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+/**
+ * Collection of Path encoding / decoding utilities for URI, Http Headers, and HTML
+ */
+public class PathUtil
```
gregw, line 22: I don't think this is well named.   It reads like it is for utility methods for the `Path` class.

Instead it has methods that work on the strange decoded URI path space that is ill defined by the servlet specification.    Currently we have such methods on URIUtil, which I guess is similarly badly named as they are not URI paths either.... well I guess they are decoded URI paths, so it's a little bit better.

Eitherway, I think we should strive to not add extra classes like this, unless their purpose is clearly defined.

jetty-util/src/main/java/org/eclipse/jetty/util/PathUtil.java
```diff
@@ -0,0 +1,157 @@
...
+        URI_ENCODED_CHARS = new String[127];
+        Arrays.fill(URI_ENCODED_CHARS, null);
+
+        // URI Reserved Chars
+        String uriReservedGenDelims = ":?#[]@"; // intentionally missing "/"
+        String uriReservedSubDelims = "!$&'()*+,;=";
+        // Extra Reserved Chars (specified by Jetty)
+        String jettyReserved = "%\"<> \\^`{}|";
+
+        String reserved = uriReservedGenDelims + uriReservedSubDelims + jettyReserved;
```
gregw, line 40: I like this array approach, but I think we should switch around how we create it to defining what characters will not be encoded rather than which characters will be encoded.

We should not encode `/` and the unreserved characters from the URI spec:
 - ALPHA: `A` - `Z` + `a` - `z`
 - DIGIT: `0` - `9`
 - `-` / `.` / `_` / `~`

I think we should encode everything else.

jetty-util/src/main/java/org/eclipse/jetty/util/PathUtil.java
```diff
@@ -0,0 +1,157 @@
...
+                // all of these characters are 8-bit and above (unicode) and should be encoded.
+                buf.append('%');
+                buf.append(HEXES[(c & 0xF0) >> 4]);
+                buf.append(HEXES[(c & 0x0F)]);
+            }
+        }
+        return buf.toString();
+    }
+
+    public static String encodePathDelayAlloc(String rawpath)
```
gregw, line 81: I think the JMH has indicated enough that this is a better algorithm than the `URIUtil.encodePath`.   Just update that method and don't need to check in and maintain the jmh or alternatives.

If we do want to check in a JMH, it should be a general comparison between doing a static array lookup vs `switch(char)`

gregw (changes requested):

jetty-util/src/main/java/org/eclipse/jetty/util/PathUtil.java
gregw: It would be interesting to jmh compare the approach used in HttpParser, which looks up characters in an array of `Token[256]` and then does a switch on the enum obtained from `token.getType`.

Issue #6372, opened 2021-06-07
Review socket options configuration


sbordet:
**Jetty version**
9.4.x

**Description**
`HTTP2Client` hard-codes the TCP_NODELAY setting, but it should be configurable.

In Jetty 10, `ClientConnector` also hard-codes it, but again should be configurable.

Look also at client-side REUSEADDR, useful when binding on the client and then restarting.

Look also at other options that may have been introduced in Java 11.

PR #6522, opened 2021-07-16
Fixes #6043 - Reimplement UnixSocket support based on Java 16.

### LINKED ISSUE #6043: Reimplement UnixSocket support based on Java 16
**Jetty version**
10.0.x

**Description**
Java 16 provides an implementation of Unix Sockets ([JEP 380](https://openjdk.java.net/jeps/380)).

We should provide an implementation based on that rather than the more unstable JNR.

sbordet:
* Introduced new module "jetty-server-unixdomain".
It uses reflection to access the Java 16 Unix-Domain classes to keep compatibility with the other modules and the build.
* Added Jetty module with only HTTP/1.1 support for now (requires review of the modules to reuse them with various connectors).
* Updated documentation to mention UnixDomainServerConnector.
* Updated client libraries to support Unix-Domain.
* Updated PROXY protocol implementation to support Unix-Domain.


jetty-unixdomain-server/src/main/config/etc/jetty-unixdomain.xml
```diff
@@ -0,0 +1,34 @@
...
+  <Call name="addConnector">
+    <Arg>
+      <New id="unixDomainConnector" class="org.eclipse.jetty.unixdomain.server.UnixDomainServerConnector">
+        <Arg><Ref refid="Server" /></Arg>
+        <Arg type="int"><Property name="jetty.unixdomain.acceptors" default="1"/></Arg>
+        <Arg type="int"><Property name="jetty.unixdomain.selectors" default="-1"/></Arg>
+        <Arg>
+          <Array type="org.eclipse.jetty.server.ConnectionFactory">
+            <Item>
+              <New class="org.eclipse.jetty.server.HttpConnectionFactory">
```
gregw: Is `HTTP` the only protocol that we will deliver over UnixDomain sockets?
Can we do `SSL`, `Proxy`, `HTTP/2` ? 

At the very least, this file should be named `jetty-unixdomain-http.xml`

jetty-unixdomain-server/src/main/config/modules/unixdomain.mod
```diff
@@ -0,0 +1,36 @@
+[description]
+Enables support for Java 16 Unix-Domain server sockets.
```
gregw:
```suggestion
Enables support for HTTP over Java 16 Unix-Domain server sockets.
```
rename module to `unixdomain-http` or `http-unixdomain`

gregw (changes requested):
I'm concerned with hard coding (in config xml) that unix domain is only for raw HTTP.   I would expect that other protocols may be carried (eg PROXY from haproxy), so we should set up to be a bit more flexible.

pom.xml
```diff
@@ -150,6 +150,7 @@
     <module>jetty-bom</module>
     <module>documentation</module>
     <module>jetty-keystore</module>
+    <module>jetty-unixdomain-server</module>
```
olamy, line 153: this module can be included only within a jdk16 maven profile.

olamy (changes requested):
the profile <id>eclipse-release</id> must be updated as well to enforce usage of jdk16 when releasing

pom.xml
sbordet: It does not require Java 16 to _build_, only to be used, so there is no need for profiles, not in the build nor in the release.

jetty-unixdomain-server/src/main/config/etc/jetty-unixdomain.xml
sbordet: @gregw I agree, but the current status for the other implementation is that we have 7 modules for all possible permutations.

I wanted to explore the possibility to separate the "carrier", i.e. IP vs Unix-Domain, from the protocols you attach to it.
Something along the lines of `--add-modules=http2` would implicitly use IP (i.e. `ServerConnector`), while `--add-modules=unix-domain,http2` would use Unix-Domain (i.e. `UnixDomainServerConnector`).

I thought that this could be worked out in a different PR.

pom.xml
olamy: in this case you need to remove this https://github.com/eclipse/jetty.project/pull/6522/files#diff-723a454092742efde002c05bbdc642cd7c1a3a4b1d069bffe9ab63fdb1c10bfbR24
```
        <configuration>
          <source>16</source>
          <target>16</target>
          <release>16</release>
        </configuration>
```
because this doesn't build with jdk11.
```
release version 16 not supported
```

sbordet: Ah, good catch!

jetty-unixdomain-server/src/main/config/etc/jetty-unixdomain.xml
gregw: Working out the full topology can be punted to a different PR, but you need to at least start with the right name.    This should be `unixdomain-http` with initially can be complete in itself, but in future might depend on a `unix-domain` module.      If you start with the non protocol module name, then we have less flexibility in future

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java
```diff
@@ -165,12 +167,13 @@ public void configureConnector() throws Exception
         // Create a ServerConnector instance.
         ServerConnector connector = new ServerConnector(server, 1, 1, new HttpConnectionFactory());
 
-        // Configure TCP parameters.
+        // Configure IP parameters.
 
-        // The TCP port to listen to.
+        // The IP port to listen to.
```
lorban: The port still relates to `TCP`.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java
```diff
@@ -179,6 +182,33 @@ public void configureConnector() throws Exception
...
+        Server server = new Server();
+
+        // The number of acceptor threads.
+        int acceptors = 1;
+
+        // The number of selectors.
+        int selectors = 1;
+
+        // Create a ServerConnector instance.
+        UnixDomainServerConnector connector = new UnixDomainServerConnector(server, 1, 1, new HttpConnectionFactory());
```
lorban: Aren't you supposed to use the `acceptors` and `selectors` variables here?

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
```diff
@@ -549,24 +549,25 @@ protected boolean removeDestination(HttpDestination destination)
...
     {
         HttpDestination destination = (HttpDestination)resolveDestination(request);
         destination.send(request, listeners);
     }
 
-    protected void newConnection(final HttpDestination destination, final Promise<Connection> promise)
+    protected void newConnection(HttpDestination destination, Promise<Connection> promise)
     {
+        // Multiple threads may access the map, especially with DEBUG logging enabled.
+        Map<String, Object> context = new ConcurrentHashMap<>();
```
lorban, line 561: Is this a read-only map or a modifiable one? In the former case, an unmodifiable `HashMap` would be preferable.

jetty-client/src/main/java/org/eclipse/jetty/client/http/HttpClientTransportOverHTTP.java
```diff
@@ -26,22 +26,24 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(HttpClientTransportOverHTTP.class);
 
     private final ClientConnectionFactory factory = new HttpClientConnectionFactory();
     private int headerCacheSize = 1024;
     private boolean headerCacheCaseSensitive;
 
     public HttpClientTransportOverHTTP()
     {
-        this(Math.max(1, ProcessorUtils.availableProcessors() / 2));
+        this(1);
```
lorban: Are you sure you want to change this behavior in this PR?

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
```diff
@@ -221,20 +238,16 @@ public void connect(SocketAddress address, Map<String, Object> context)
...
-                    LOG.debug("Binding to {} to connect to {}{}", bindAddress, address, (reuseAddress ? " reusing address" : ""));
-                channel.setOption(StandardSocketOptions.SO_REUSEADDR, reuseAddress);
-                channel.bind(bindAddress);
-            }
+                bind(channel, bindAddress);
             configure(channel);
 
             boolean connected = true;
-            boolean blocking = isConnectBlocking();
+            boolean blocking = isConnectBlocking() && address instanceof InetSocketAddress;
```
lorban, line 255: Why is `isConnectBlocking` ignored for non `InetSocketAddress` instances?

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
```diff
@@ -288,9 +301,34 @@ public void accept(SocketChannel channel, Map<String, Object> context)
...
+        {
+            boolean reuseAddress = getReuseAddress();
+            if (LOG.isDebugEnabled())
+                LOG.debug("Binding to {} reusing address {}", bindAddress, reuseAddress);
+            channel.setOption(StandardSocketOptions.SO_REUSEADDR, reuseAddress);
+            channel.bind(bindAddress);
+        }
+        catch (Throwable x)
+        {
+            if (LOG.isDebugEnabled())
```
lorban, line 321: I'm not certain silently ignoring the bind address when bind fails is a sensible thing to do. Why do you want to change the existing behavior of throwing when bind fails?

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
```diff
@@ -288,9 +301,34 @@ public void accept(SocketChannel channel, Map<String, Object> context)
...
+                LOG.debug("Could not bind {}", channel);
+        }
+    }
+
     protected void configure(SocketChannel channel) throws IOException
     {
-        channel.socket().setTcpNoDelay(true);
+        try
+        {
+            channel.setOption(StandardSocketOptions.TCP_NODELAY, true);
```
lorban, line 330: Why forcing this TCP option by default?

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
```diff
@@ -351,4 +389,78 @@ protected void connectionFailed(SelectableChannel channel, Throwable failure, Ob
...
+             * derived from the input socket address.</p>
+             * <p>The input socket address represents the destination socket address to
+             * connect to, as it is typically specified by a URI authority, for example
+             * {@code localhost:8080} if the URI is {@code http://localhost:8080/path}.</p>
+             * <p>However, the returned socket address may be different as the implementation
+             * may use a Unix-Domain socket address to physically connect to the virtual
+             * destination socket address given as input.</p>
+             * <p>The return type is a pair/record holding the socket channel and the
+             * socket address, with the socket channel not yet connected.
+             * The implementation of this methos must not call
```
lorban: Typo: implementation of this **methos** -> **method**

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
```diff
@@ -351,4 +389,78 @@ protected void connectionFailed(SelectableChannel channel, Throwable failure, Ob
...
+
+        public SocketAddress getSocketAddress()
+        {
+            return address;
+        }
+
+        /**
+         * <p>A factory for {@link SocketChannelWithAddress} instances.</p>
+         */
+        public interface Factory
```
lorban: This mechanism won't work for QUIC as it works with `DatagramChannel` instead of `SocketChannel`.

The `SocketChannelWithAddress` class could be modified to hold a `SelectableChannel` instead of `SocketChannel` but that would not solve the API problem that `ClientConnector` already exposes `SocketChannel` via `configure(SocketChannel channel)` for instance.

This class is sitting between OSI Layer 4 (TCP) and Layer 5 (Sockets, both TCP and Unix Domain), or maybe even overlapping both so I'm not sure I would expose it as an API. Something doesn't look right about it.

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
```diff
@@ -351,4 +389,78 @@ protected void connectionFailed(SelectableChannel channel, Throwable failure, Ob
...
+        {
+            public static Factory forUnixDomain(Path path)
+            {
+                return (address, context) ->
+                {
+                    try
+                    {
+                        ProtocolFamily family = Enum.valueOf(StandardProtocolFamily.class, "UNIX");
+                        Class<?> channelClass = Class.forName("java.nio.channels.SocketChannel");
+                        SocketChannel socketChannel = (SocketChannel)channelClass.getMethod("open", ProtocolFamily.class).invoke(null, family);
```
lorban: How about `SocketChannel.class.getMethod("open", ProtocolFamily.class).invoke(null, family);`?

jetty-server/src/main/java/org/eclipse/jetty/server/ProxyConnectionFactory.java
```diff
@@ -957,4 +970,47 @@ public void write(Callback callback, ByteBuffer... buffers) throws WritePendingE
...
+                    return (SocketAddress)unixDomainSocketAddress.getMethod("of", String.class).invoke(null, path);
+                return null;
+            }
+            catch (Throwable ignored)
+            {
+                return null;
+            }
+        }
+
+        private static String toPath(byte[] bytes)
```
lorban, line 1004: This method looks unused.

lorban (changes requested):

jetty-unixdomain-server/src/main/config/etc/jetty-unixdomain.xml
sbordet: Done.

jetty-unixdomain-server/src/main/config/modules/unixdomain.mod
sbordet: Done.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java
sbordet: Good catch!

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
sbordet: It is a modifiable one, as different `ClientConnectionFactory` may enrich it with additional entries.

jetty-client/src/main/java/org/eclipse/jetty/client/http/HttpClientTransportOverHTTP.java
sbordet: Good point, reverted.

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
sbordet: Because in the code below I cannot call `channel.socket()` if it is a Unix-Domain socket address (it throws).
I would do `channel instanceof SocketChannel`, but that is true for both network and Unix-Domain, despite the fact that calling `socket()` on a `SocketChannel` may throw if it is Unix-Domain. 🤷🏼‍♂️

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
sbordet: Again because for Unix-Domain, binding makes no sense, so if `ClientConnector` is misconfigured, we ignore the binding attempt.

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
sbordet: You are right that this should be configurable, but it's covered by #6372.

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
sbordet: Much better indeed.

jetty-server/src/main/java/org/eclipse/jetty/server/ProxyConnectionFactory.java
sbordet: It's used at lines 614 and 616.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java
```diff
@@ -179,6 +182,33 @@ public void configureConnector() throws Exception
...
+
+        // Create a ServerConnector instance.
+        UnixDomainServerConnector connector = new UnixDomainServerConnector(server, acceptors, selectors, new HttpConnectionFactory());
+
+        // Configure Unix-Domain parameters.
+
+        // The Unix-Domain path to listen to.
+        connector.setUnixDomainPath(Path.of("/tmp/jetty.sock"));
+
+        // The TCP accept queue size.
```
lorban, line 204: This is unrelated to TCP.

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
joakime: This should probably be a new issue.

sbordet:
@lorban I made the mechanism to create a Unix-Domain `ClientConnector` more private, as discussed.

I also updated `FastCGIProxyServlet` to use Unix-Domain when configured so.

lorban (dismissed):

sbordet:
@gregw @olamy nudge

gregw (dismissed):
better in that out at this stage!

jetty-fcgi/fcgi-server/src/test/java/org/eclipse/jetty/fcgi/server/proxy/FastCGIProxyServletTest.java
```diff
@@ -216,4 +230,34 @@ public void handle(String target, org.eclipse.jetty.server.Request baseRequest,
...
     }
+
+    @Test
+    @EnabledForJreRange(min = JRE.JAVA_16)
+    public void testUnixDomain() throws Exception
+    {
+        int maxUnixDomainPathLength = 108;
+        Path path = Files.createTempFile("unix", ".sock");
+        if (path.normalize().toAbsolutePath().toString().length() > maxUnixDomainPathLength)
+            path = Files.createTempFile(Path.of("/tmp"), "unix", ".sock");
```
olamy, line 241: can be a problem on windows or osx? this test should not run on windows `@DisabledOnOs(OS.WINDOWS)`?

jetty-unixdomain-server/src/test/java/org/eclipse/jetty/unixdomain/server/UnixDomainTest.java
```diff
@@ -0,0 +1,267 @@
...
+import org.junit.jupiter.api.condition.JRE;
+
+import static org.eclipse.jetty.client.ProxyProtocolClientConnectionFactory.V1;
+import static org.eclipse.jetty.client.ProxyProtocolClientConnectionFactory.V2;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
```
olamy, line 57: maybe need `@DisabledOnOs(OS.WINDOWS)`

tests/test-distribution/src/test/java/org/eclipse/jetty/tests/distribution/DistributionTests.java
```diff
@@ -906,4 +908,37 @@ public void testDefaultLoggingProviderNotActiveWhenExplicitProviderIsPresent() t
             assertFalse(Files.exists(jettyBase.resolve("resources/jetty-logging.properties")));
         }
     }
+
+    @Test
+    @EnabledForJreRange(min = JRE.JAVA_16)
```
olamy, line 903: `@DisabledOnOs(OS.WINDOWS)`

olamy (changes requested):
might need to disable some tests for windows OS

joakime (approved):
Tested on Windows 10 laptop, works as expected.

sbordet:
@olamy this PR now runs fine on Windows too.
There is a new profile to set the Unix-Domain directory for Windows.
Property `unix.socket.tmp` has been removed in favor of `jetty.unixdomain.dir`.

olamy (approved):

PR #6527, opened 2021-07-19
Adding testcase to show Proxy with optional SSL

joakime:

### DIFF
--- a/jetty-server/src/test/java/org/eclipse/jetty/server/ProxyProtocolTest.java
+++ b/jetty-server/src/test/java/org/eclipse/jetty/server/ProxyProtocolTest.java
@@ -29,11 +29,15 @@ import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import org.eclipse.jetty.http.HttpTester;
 import org.eclipse.jetty.server.handler.AbstractHandler;
 import org.eclipse.jetty.util.TypeUtil;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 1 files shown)


jetty-server/src/test/java/org/eclipse/jetty/server/ProxyProtocolTest.java
```diff
@@ -57,6 +62,45 @@ public void destroy() throws Exception
             server.stop();
     }
 
+    @Test
+    public void testNoProxyJustHttp() throws Exception
```
sbordet, line 66: If the goal of this test is to test whether the absence of the PROXY protocol lands on the next protocol, then why wrapping TLS into a detector?

Otherwise wrapping TLS is tested and documented already (at least in 10).
See `DetectorConnectionTest`.

If it is an example, it is "wrong" as the `SslConnectionFactory` is not configured with a KeyStore so in case of TLS it will never work.

sbordet (changes requested):

jetty-server/src/test/java/org/eclipse/jetty/server/ProxyProtocolTest.java
joakime: The point with that specific testcase is to show that proxy and ssl are optional connection factories.
I'll be happy to document that on that specific testcase.

Do you want another testcase showing that ssl works too?

sbordet: @joakime I don't understand.
How's this test different from `DetectorConnectionTest.testOptionalSsl()`?

joakime:
Closing, as this is handled in other test cases.

PR #6529, opened 2021-07-21
Issue #6373 testing modules, starting with session storage

### LINKED ISSUE #6373: Need test infrastructure to enable testing of all jetty modules
Currently we have a few tests that exercise the distro in` tests/test-distribution` but I think we need to either augment it, or create a separate project whereby a CI job could just endlessly enable and test each of our jetty modules. That would be the only way to check that a module is actually working as expected. We would need to check:
* syntax of `.xml `files and `.mod` files 
* that each module can actually be enabled
* setting of each property in a `.ini` file
* function of properties set in a `.ini` file

olamy:


jetty-gcloud/jetty-gcloud-session-manager/src/main/java/org/eclipse/jetty/gcloud/session/GCloudSessionDataStore.java
```diff
@@ -455,15 +457,61 @@ public int getMaxRetries()
...
+    {
+        return _projectId;
+    }
+
     @Override
     protected void doStart() throws Exception
     {
         if (!_dsProvided)
         {
+            boolean defaultInstance = true;
```
janbartel: I'm not fond of the extra boolean for `defaultInstance`. I'm a bit on the fence for configuring passwords in here as well. Isn't there a way to provide a Datastore that is already configured for the test?

janbartel (changes requested):

jetty-gcloud/jetty-gcloud-session-manager/src/main/java/org/eclipse/jetty/gcloud/session/GCloudSessionDataStore.java
olamy: we do not need to configure password here, the issue with configuring user/password is with infinispan :) 
anyway I removed the boolean.

jetty-gcloud/jetty-gcloud-session-manager/src/main/java/org/eclipse/jetty/gcloud/session/GCloudSessionDataStore.java
```diff
@@ -455,15 +457,53 @@ public int getMaxRetries()
...
+    {
+        return _projectId;
+    }
+
     @Override
     protected void doStart() throws Exception
     {
         if (!_dsProvided)
         {
+            DatastoreOptions.Builder builder = DatastoreOptions.newBuilder();
```
janbartel:
```suggestion
            boolean customNamespace = StringUtil.isBlank(getNamespace());
            boolean customHost = StringUtil.isBlank(getHost());
            boolean customProjectId = StringUtil.isBlank(getProjectId());
            if (!customNamespace && !customHost && !customProjectId)
                 _datastore = DatastoreOptions.getDefaultInstance().getService();
            else
            {
                DatastoreOptions.Builder builder = DatastoreOptions.newBuilder();
                if (customNamespace)
                    builder.setNamespace(getNamespace());
                if (customHost)
                    builder.setHost(getHost());
                if (customProjectId)
                    builder.setProjectId(getProjectId());
                _datastore = builder.build().getService();
            }
```

janbartel:
Looking pretty good - although this is till marked as Draft.

olamy:
@janbartel yup because still wip for memcached and mongodb

jetty-home/src/main/resources/modules/hawtio.mod
```diff
@@ -26,6 +26,9 @@ http://hawt.io/
 http://github.com/hawtio/hawtio
 http://www.apache.org/licenses/LICENSE-2.0.html
 
+[ini]
+hawtio.version?=2.0.2
```
olamy: @janbartel @joakime  maybe last 2.13.5? (I have simply put the first version working correctly with jdk11)

tests/test-distribution/src/test/java/org/eclipse/jetty/tests/distribution/session/MongodbSessionDistributionTests.java
```diff
@@ -0,0 +1,80 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
+
+/**
+ *  This simulate the onlyClient option which means the JVM running Jetty is only an Hazelcast client and not part
+ *  of the cluster
+ */
```
janbartel, line 28: Delete this comment - this is the mongo test, not the hazelcast test.

janbartel (changes requested):
Looking good, a couple of things in the comments.

jetty-gcloud/jetty-gcloud-session-manager/src/main/java/org/eclipse/jetty/gcloud/session/GCloudSessionDataStore.java
olamy: uhm it means custom namespace or projectId are never used except when everything has been setup.
I guess users must be able to setup custom projectId or namespace or host without configuring all of those 3 parameters?
btw the current PR doesn't take into account neither (that's why I was using a boolean in a previous commits)

jetty-home/src/main/resources/modules/hawtio.mod
olamy: change default to last 2.13.5

janbartel (approved):

PR #6530, opened 2021-07-21
ReservedThreadExecutor ReservedThread.offer should never block indefinitely

lorban:
If something awry happens in `ReservedThread.run()` after pushing the thread onto the stack (i.e.: if `reservedWait` throws OOME for instance) then the thread calling `ReservedThread.tryExecute()` could block forever while that method's contract strictly forbids that.

So let's not use any blocking call in there.

See also #6496

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
@@ -256,6 +256,7 @@ public class ReservedThreadExecutor extends AbstractLifeCycle implements TryExec
         }
         catch (RejectedExecutionException e)
         {
+            _pending.decrementAndGet();
             LOG.ignore(e);
         }
     }
@@ -283,8 +284,7 @@ public class ReservedThreadExecutor extends AbstractLifeCycle implements TryExec
 
             try
             {
-                _task.put(task);
-                return true;
+                return _task.offer(task, 1, TimeUnit.SECONDS);
             }
             catch (Throwable e)
             {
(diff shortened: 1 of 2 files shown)


jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -336,63 +336,75 @@ private Runnable reservedWait()
...
+                if (LOG.isDebugEnabled())
+                    LOG.debug("{} Entering", this);
 
-                // Are we stopped?
-                if (size < 0)
-                    return;
-
-                // Are we surplus to capacity?
-                if (size >= _capacity)
+                while (isRunning())
```
gregw, line 344: This needs to become `while(true)` otherwise there is a race with `doStop`

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -336,63 +336,75 @@ private Runnable reservedWait()
...
+                        LOG.warn(e);
+                    }
                 }
             }
-
-            if (LOG.isDebugEnabled())
-                LOG.debug("{} Exited", this);
+            catch (Throwable x)
+            {
+                LOG.warn(x);
```
gregw, line 401: This doesn't do what it says in the PR description. If `reservedWait` throws, the reserved thread is left on the stack?

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -336,63 +336,75 @@ private Runnable reservedWait()
...
                 {
-                    if (LOG.isDebugEnabled())
-                        LOG.debug("{} size {} > capacity", this, size, _capacity);
-                    if (_starting)
-                        _pending.decrementAndGet();
-                    return;
-                }
+                    // test and increment size BEFORE decrementing pending,
+                    // so that we don't have a race starting new pending.
+                    int size = _size.get();
```
gregw, line 348: my preferred change here is to make size and pending an `AtomicBiInteger`.  I have already done that work in a branch against jetty-10, but have not put it forward as I wanted to see if we can work out the other issues first.    But if we are going to make changes, then I'd prefer to make the logic clearly atomic with a BiInteger rather than optimisticly atomic.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -283,8 +284,7 @@ public boolean offer(Runnable task)
 
             try
             {
-                _task.put(task);
-                return true;
+                return _task.offer(task, 1, TimeUnit.SECONDS);
```
gregw, line 287: If this times out, then we don't put the reserved thread back onto the stack, so the reserved thread may then wait forever.  We need to at least flag that it has to die.

gregw (changes requested):
I do agree that if we are to merge something into HEAD, that it has to be a lot simpler than #6496.  However, I think this is too simple and misses a few fixes.

I'll do another iteration on #6496 and separate out what I think a simple defensive changes from the speculative debugging changes.  I might also merge in my BiInteger changes, as once you see the maths is not atomic, it can't be unseen... even though we "know" it should be OK.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
lorban: We had a lengthy discussion on the subject yesterday with @sbordet. Our conclusion was that the `pending` counter seems useless if not for reporting, and the `size` counter could be replaced with `_stack.size()`.

There's always going to be some non-atomicity between the counters and the stack, so I'm not sure what making both counters atomic between themselves would bring.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
lorban: You're correct. Yet another overlook.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
lorban: Maybe we should never use a blocking-without-timeout read call to the `SynchronousQueue`?

lorban (commented):
We agree that `ReservedThreadExecutor` needs some improvements, but we don't yet know exactly what they should be. This needs to be discussed in more details IMHO.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
gregw: You can't do that.  It used to be like that at it was a race.    See my PR as I think it can be done without any race.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutorWithPool.java
```diff
@@ -0,0 +1,184 @@
...
+            {
+                // TODO handle
+            }
+        }
+
+        public boolean offer(Runnable task)
+        {
+            try
+            {
+                exchanger.exchange(task, 1, TimeUnit.SECONDS);
```
gregw, line 130: This is just using a different data structure for the offer, when we don't know if the problem was with the `SynchronousQueue` or not.
You'd have to think that `Exchanger` are internally going to be moderately similar, but `Exchanger` is likely to be more complex because it is unnecessarily bi-directional.

Besides, having the offering thread block (or even yield in an internal spin) here defeats the whole purpose of EPC.   We can't give an opportunity for the calling thread to be moved to another CPU core.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutorWithPool.java
```diff
@@ -0,0 +1,184 @@
...
+import org.eclipse.jetty.util.component.AbstractLifeCycle;
+import org.eclipse.jetty.util.log.Log;
+import org.eclipse.jetty.util.log.Logger;
+
+@ManagedObject("An alternate pool for reserved threads")
+public class ReservedThreadExecutorWithPool extends AbstractLifeCycle implements TryExecutor
+{
+    private static final Logger LOG = Log.getLogger(ReservedThreadExecutorWithPool.class);
+
+    private final Pool<ReservedThread> stack;
```
gregw, line 40: Using the `Pool` for the stack is a big change in semantics.  Currently one a reserved thread is given a task, it is no longer counted as a reserved thread and does not hold any of the capacity, thus it does not prevent other threads becoming reserved.   Using a pool will put an absolute limit on the numer of tasks that can be reserved plus can be executed by something that used to be reserved.

I am also concerned that the underlying COW list might not be the right data structure if the reserved thread collection is moderately volatile.

gregw (changes requested):
oh I was initially sooooo much in favour of using the `Pool` class here... until I thought about it....
Doh!

PR #6533, opened 2021-07-22
9.4.x: Add defensive check in HttpConnection content production

### LINKED ISSUE #6491: onDataAvailable() not called when HttpParser is closed prematurely
**Jetty version(s)**
9.4.42

**Description**
If the `HttpParser` is closed prematurely (state = `CLOSE`), then trying to read content asynchronously fail because `HttpConnection.fillAndParseForContent()` tests whether the parser is in content state.

lorban:
Currently, if the parser gets closed and the connection isn't aborted, the code enters a select/wakeup/fillAndParseForContent/wait infinite loop.

The above should never happen unless a bug gets introduced, but if such bug would happen at least the situation would be detected and acted upon.

See #6491.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
@@ -316,6 +316,11 @@ public class HttpConnection extends AbstractConnection implements Runnable, Http
      */
     protected boolean fillAndParseForContent()
     {
+        // Defensive check to avoid an infinite select/wakeup/fillAndParseForContent/wait loop
+        // in case the parser was mistakenly closed and the connection was not aborted.
+        if (_parser.isClose() || _parser.isClosed())
+            throw new IllegalStateException("Parser is closed: " + _parser);
+
         boolean handled = false;
         while (_parser.inContentState())
         {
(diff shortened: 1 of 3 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -316,6 +316,11 @@ else if (filled < 0)
      */
     protected boolean fillAndParseForContent()
     {
+        // Defensive check to avoid an infinite select/wakeup/fillAndParseForContent/wait loop
+        // in case the parser was mistakenly closed and the connection was not aborted.
+        if (_parser.isClose() || _parser.isClosed())
```
gregw: Didn't we add a `isXxx` that tested both of these states?  If not, let's do so as this double test is done at least one other location.

gregw (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
lorban: Sure, if @sbordet agrees.

sbordet: @lorban I agree. I don't see the method in 9.4.x, but I would call it `isTerminated()` rather than `isPastEnd()` or similar.
The javadoc for the new method should be enough to explain the state.

sbordet (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
lorban: Okay, I introduced `isTerminated()` by copying the mechanism that was already used to implement `isIdle()` and `isComplete()`.

sbordet (dismissed):

gregw (approved):

PR #6534, opened 2021-07-22
Update Build to work with JDK 17

joakime:
Updates to Jenkinsfile
Updates to jdk17 profile to disabled known broken plugins for JDK 17
Updates to weld for jdk17 support
Fixes to javadoc because of tightened rules

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,8 +40,8 @@
     <junit.version>5.7.0</junit.version>
     <maven.version>3.6.3</maven.version>
     <maven.resolver.version>1.6.1</maven.resolver.version>
-    <weld.version>3.1.5.Final</weld.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -1465,6 +1465,17 @@
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>jdk17</id>
+      <activation>
+        <jdk>[1.7,)</jdk>
```
sbordet: You meant `17`, not `1.7` here, right?

sbordet (changes requested):

pom.xml
joakime: Yeah.  good catch!

sbordet (approved):

olamy (approved):

PR #6535, opened 2021-07-23
Non blocking ReservedThreadExecutor

gregw:
This is yet another alternative fix for #6495 replacing #6496
The stack is removed entirely and all the reserved threads poll in parallel on the same synchronous queue.  
The `tryExecute` semantic is simply to do a non-blocking offer to the queue which will either succeed or fail, and no further action is required.

The `size` and `pending` counts are important to keep the capacity correct.
The reserved threads are added as beans, but only for the purposes of dumping.
The reserved threads track their state to make the dumps more meaningful.

The only issue with losing the stack is that it may prevent reserved threads idling out as load might be shared over all the reserved threads rather than just the most recent.
However, there are other strategies that could be applied to deal with that (eg looking at recent minimum size before re-polling).


jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -266,139 +254,136 @@ public String toString()
...
+        STOPPED
+    }
 
-        public void stop()
-        {
-            offer(STOP);
-        }
+    private class ReservedThread implements Runnable
+    {
+        private final AtomicReference<State> _state = new AtomicReference<>(State.PENDING);
```
lorban: `_state` could be a simple `volatile` as only `toString()` makes it visible to other threads.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -266,139 +254,136 @@ public String toString()
...
+                    // check the state
+                    State state = _state.get();
+
+                    // reduce pending if this thread was pending
+                    int pending = AtomicBiInteger.getHi(count) - (state == State.PENDING ? 1 : 0);
+                    int size = AtomicBiInteger.getLo(count);
+
+                    // Exit if excess capacity or it has been too long since we hit zero reserved threads?
+                    if (size >= _capacity ||
+                        size > 0 && _idleTime > 0 && _idleTimeUnit.toNanos(_idleTime) < (_lastZero - System.nanoTime()) && _state.compareAndSet(state, State.IDLED_OUT))
```
lorban: 1. What is that `_lastZero` logic about? On a busy server, I expect the `size` counter to never reach 0 and I wouldn't want to have to re-create the `ReservedThread`s every minute or so, even if that would be fairly cheap, I fail to see the usefulness of such logic.

2. The state change to `IDLED_OUT` here will make the state change to `POLLING` a few lines below always fail because you don't refresh the `state` local variable.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -212,51 +207,44 @@ public boolean tryExecute(Runnable task)
...
-                    return;
-                if (_pending.compareAndSet(pending, pending + 1))
-                {
-                    if (LOG.isDebugEnabled())
-                        LOG.debug("{} startReservedThread p={}", this, pending + 1);
-                    _executor.execute(new ReservedThread());
-                    return;
-                }
+                ReservedThread thread = new ReservedThread();
+                addBean(thread);
```
lorban: `addBean` ends up modifying a COW collection, that's going to generate a lot of garbage on a busy server where lots of reserved threads are created.

lorban (commented):
I like the simplification a lot!

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
gregw: probably... but essentially AtomicReference is just a simple volatile with some conveniance methods :)
I'll have a play and see if it reads better.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
gregw: 1.  If we have not been to 0 reserved threads for a period longer than the idle timeout, then we timeout.
2. Exactly!

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
gregw: It is a worry, but this is not the fine grained reserved thread add/create, only the courser grained lifecycle.   A thread can become reserved, run a task, become reserved, run a task, etc. many times without touching the COW collection.    Note that means the collection contains all the reserved threads, plus all the threads that were once reserved but are not RUNNING.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
gregw: I'll look to see if this logic can be made clearer...

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
lorban: When there are many more threads calling tryExecute than there are reserved threads in the pool, most reserved threads are thrown away after they're done with their execution because the pending counter is constantly pegged at the maximum.

I'm not sure how realistic such load is, but it's very easy to write a synthetic test that triggers this behavior.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
lorban: 1. Is that useful?
2. `IllegalStateException("Bad State")` is thrown when this happens.

gregw: 1. @lorban The issue I'm trying so solve here is that if we have multiple reserved threads waiting on the SQ, then there is no guarantee of the order in which they wake up.  Let's say wake up is fair and longest waiting wakes up first, then this is bad for idling out reserved threads.   If we have 30 reserved threads, 30s idle timeout and 1 request per second, then each of the 30 reserved threads will receive 1 request every 30 seconds and thus never time out.

2. will fix

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
gregw: I'm not sure that is true.  If there are many more calls to tryExecute than the capacity of the reserved thread pool, then most calls return false and nothing is done.   A reserved thread is only started if there is capacity for it.  It is only stopped if it goes idle.   So modifications should not be too frequent.... but I'll have a look at the idle mechanism and make sure we don't kill all our reserved threads at once.

gregw: I'm using a concurrent queue now for holding the threads for dumping

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
gregw: fixed the ordering and clarified the algorithm a bit.  Also slowed the shrinkage rate

gregw:
@lorban updated

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -62,13 +69,14 @@ public String toString()
 
     private final Executor _executor;
     private final int _capacity;
-    private final ConcurrentLinkedDeque<ReservedThread> _stack;
-    private final AtomicInteger _size = new AtomicInteger();
-    private final AtomicInteger _pending = new AtomicInteger();
+    private final Queue<ReservedThread> _threads = new ConcurrentLinkedQueue<>();
```
lorban: A concurrent `Set` is both more natural for this case and faster. I'd go for:

```java
Set<ReservedThread> _threads = ConcurrentHashMap.newKeySet();
```

jetty-util/src/test/java/org/eclipse/jetty/util/thread/ReservedThreadExecutorTest.java
```diff
@@ -332,6 +359,6 @@ public void run()
         pool.stop();
 
         assertThat(usedReserved.get() + usedPool.get(), is(LOOPS));
```
lorban, line 367: My original stress test repeated the above processing a few times because the `usedReserved` counter would stick to 0 after the bug triggered, so I also tested that `usedReserved` must be > 0.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -266,139 +265,144 @@ public String toString()
...
 
-        public void stop()
-        {
-            offer(STOP);
-        }
+    private class ReservedThread implements Runnable
+    {
+        // The state and thread are kept only for dumping
+        private volatile State _state = State.PENDING;
+        private Thread _thread;
```
lorban: `_thread` should be marked `volatile` as `toString()` will be called from concurrent threads.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -266,139 +265,144 @@ public String toString()
...
-                        _pending.decrementAndGet();
-                    return;
-                }
+                    State next;
+                    if (size < 0 || size >= _capacity)
+                    {
+                        // The executor has stopped or this thread is excess to capacity
+                        next = State.STOPPED;
+                    }
+                    else if (size > 0 && _idleTime > 0 && _idleTimeUnit.toNanos(_idleTime) < (System.nanoTime() - _lastZero))
```
lorban: Okay, you're kinda sorta reproducing the previous "stack" effect by expiring reserved threads. Got it.

lorban (changes requested):

gregw:
@sbordet nudge?

jetty-util/src/test/java/org/eclipse/jetty/util/thread/ReservedThreadExecutorTest.java
lorban: This does not check that `tryExecute` doesn't stop working after some time. You need to reset the counters and latch after those assertions and loop over to resubmit the tasks.

jetty-util/src/test/java/org/eclipse/jetty/util/thread/ReservedThreadExecutorTest.java
```diff
@@ -277,7 +305,7 @@ public void stressTest() throws Exception
         reserved.setIdleTimeout(0, null);
         reserved.start();
 
-        final int LOOPS = 1000000;
+        final int LOOPS = 2000000;
```
lorban: This seems too large to me, I don't think 2 mil. executions will show anything more than 2k.

lorban (changes requested):

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -126,16 +133,16 @@ public int getCapacity()
...
 
     @ManagedAttribute(value = "pending reserved threads", readonly = true)
     public int getPending()
     {
-        return _pending.get();
+        return _count.getHi();
     }
 
-    @ManagedAttribute(value = "idletimeout in MS", readonly = true)
+    @ManagedAttribute(value = "idle timeout in MS", readonly = true)
```
sbordet: while you're at it, please change "MS" to "ms".

jetty-util/src/test/java/org/eclipse/jetty/util/thread/ReservedThreadExecutorTest.java
gregw: reduced a bit.  I was really just wanting the test to run long enough that the JIT would have done some changes

gregw:
@sbordet nudge

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -172,27 +179,10 @@ public void doStop() throws Exception
...
-                continue;
-            }
-
-            _size.decrementAndGet();
-            thread.stop();
-        }
+        _count.set(0, -1);
+        for (int i = _threads.size(); i-- > 0;)
+            _queue.offer(STOP);
+        _threads.clear();
```
sbordet, line 196: This is still prone to leaving threads inside `_threads`, as the `offer()` may fail (no threads to take it yet), but a thread is starting. Not sure how to fix it though, I think it's really difficult unless grabbing a coarse lock.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -212,52 +202,61 @@ public boolean tryExecute(Runnable task)
     {
         if (LOG.isDebugEnabled())
             LOG.debug("{} tryExecute {}", this, task);
-
         if (task == null)
             return false;
 
-        ReservedThread thread = _stack.pollFirst();
-        if (thread == null)
+        boolean offered = _queue.offer(task);
```
sbordet, line 221: I think we should return early like we did before, _without_ a check on the size.
We can have `size > 0` and fail the offer and still want to start a thread. See below.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -212,52 +202,61 @@ public boolean tryExecute(Runnable task)
...
+                break;
+            if (_count.compareAndSet(count, getHi(count), --size))
+                break;
         }
-
-        int size = _size.decrementAndGet();
-        if (!thread.offer(task))
-            return false;
-
         if (size == 0 && task != STOP)
```
sbordet, line 229: I don't think this is correct now, as we want to start threads even when the `size > 0`.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -212,52 +202,61 @@ public boolean tryExecute(Runnable task)
...
-                if (_pending.compareAndSet(pending, pending + 1))
-                {
-                    if (LOG.isDebugEnabled())
-                        LOG.debug("{} startReservedThread p={}", this, pending + 1);
-                    _executor.execute(new ReservedThread());
-                    return;
-                }
+                ReservedThread thread = new ReservedThread();
+                _executor.execute(thread);
+                _threads.add(thread);
```
sbordet, line 254: I think these 2 lines should be swapped, as we want to record the thread before we submit it for execution.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -62,13 +69,14 @@ public String toString()
...
-    private final AtomicInteger _size = new AtomicInteger();
-    private final AtomicInteger _pending = new AtomicInteger();
+    private final Set<ReservedThread> _threads = ConcurrentHashMap.newKeySet();
+    private final SynchronousQueue<Runnable> _queue = new SynchronousQueue<>(false);
+    private final AtomicBiInteger _count = new AtomicBiInteger(); // hi=pending; lo=size;
 
     private ThreadPoolBudget.Lease _lease;
     private long _idleTime = 1L;
     private TimeUnit _idleTimeUnit = TimeUnit.MINUTES;
+    private volatile long _lastZero = Long.MAX_VALUE;
```
sbordet: I prefer `_lastEmpty` or `_lastEmptyTime`.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -212,52 +202,61 @@ public boolean tryExecute(Runnable task)
...
-                    if (LOG.isDebugEnabled())
-                        LOG.debug("{} startReservedThread p={}", this, pending + 1);
-                    _executor.execute(new ReservedThread());
-                    return;
-                }
+                ReservedThread thread = new ReservedThread();
+                _executor.execute(thread);
+                _threads.add(thread);
             }
+            catch (RejectedExecutionException e)
```
sbordet: I think this should be a `catch(Thowable)` as the executor may fail to start a thread (e.g. too many for the OS configuration, or out of memory).

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -62,13 +69,14 @@ public String toString()
...
-    private final ConcurrentLinkedDeque<ReservedThread> _stack;
-    private final AtomicInteger _size = new AtomicInteger();
-    private final AtomicInteger _pending = new AtomicInteger();
+    private final Set<ReservedThread> _threads = ConcurrentHashMap.newKeySet();
+    private final SynchronousQueue<Runnable> _queue = new SynchronousQueue<>(false);
+    private final AtomicBiInteger _count = new AtomicBiInteger(); // hi=pending; lo=size;
 
     private ThreadPoolBudget.Lease _lease;
     private long _idleTime = 1L;
     private TimeUnit _idleTimeUnit = TimeUnit.MINUTES;
```
sbordet: I would convert these 2 fields into 1 (in nanos), makes it so much easier to use it in the rest of the code.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
sbordet: I would prefer this update to be atomic using `AtomicLong.updateAndGet()`, otherwise many threads can see the expression to be true and idle themselves.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -266,139 +265,144 @@ public String toString()
...
-                if (task == STOP)
-                    // return on STOP poison pill
-                    break;
+                    // Is the task the STOP poison pill?
+                    if (task == STOP)
+                    {
+                        if (_state != State.IDLE)
+                            _state = State.STOPPED;
+                        break;
+                    }
```
sbordet: This block has logic that should be in `reservedWait()`.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -266,139 +265,144 @@ public String toString()
...
-                LOG.debug("{} waiting", this);
+                LOG.debug("{} waiting {}", this, ReservedThreadExecutor.this);
 
             while (true)
             {
                 try
                 {
-                    Runnable task = _idleTime <= 0 ? _task.take() : _task.poll(_idleTime, _idleTimeUnit);
+                    // Always poll at some period as safety to ensure we don't poll forever.
+                    Runnable task = _idleTime <= 0 ? _queue.poll(30, TimeUnit.SECONDS) : _queue.poll(_idleTime, _idleTimeUnit);
```
sbordet: The hardcoded 30 seconds should be either the default idle timeout (1 minute) or a much shorter time like 1 second to speed up stopping.

If we stop while entering this method, this thread is going to wait 30 seconds (the offer of the STOP task will fail, as the thread is still not polling), which seems like a long time.
Waiting just 1 second will make the stop edge case faster, and if the server is idle, it's a loop around every second.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -266,139 +265,144 @@ public String toString()
...
                     if (LOG.isDebugEnabled())
-                        LOG.debug("{} task={}", this, task);
+                        LOG.debug("{} task={} {}", this, task, ReservedThreadExecutor.this);
                     if (task != null)
                         return task;
 
-                    if (_stack.remove(this))
+                    // Has the RTE stopped?
+                    if (_count.getLo() < 0)
+                        return STOP;
```
sbordet: This should also change the state, like the block below does, so that the caller does not have to differentiate between the 2 cases where it can get back a STOP task.

sbordet (changes requested):

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
gregw: A thread that is starting will see the -1 either as it starts or after a timeout

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
gregw: I don't agree. I don't think we want to start a thread unless we are size 0, even if the offer fails.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
gregw: This PR doesn't change the basic logic.  This is still true.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
gregw: I made it ms as that is a better resolution.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
gregw: I've changed to offer size to the queue.   I originally was spinning here waiting for zero, but we just risk blocking in stop.
I think this is a reasonable compromise.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
gregw: More precisely, if we fail in the offer here, it is not because of insufficient size.  We have reserved thread(s), they've just not arrived yet. starting another is not helpful.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
gregw: OK it changes it a little. Previously if we failed to pick a thread from the stack, it assumed that was because size==0.
This PR now always checks the size, because if it missed the thread because it was late to the rendezvous, then starting another doesn't help and just delays the producing thread more.
So I like this improvement in the logic.  @sbordet

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
gregw: resolving to discuss below

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
gregw: I've simplified the loop a little bit.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -43,7 +52,7 @@
  * whenever it has been idle for that period.
```
sbordet: Javadoc should be updated to minimally say that they will idle timeout.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -62,13 +71,14 @@ public String toString()
 
     private final Executor _executor;
     private final int _capacity;
-    private final ConcurrentLinkedDeque<ReservedThread> _stack;
-    private final AtomicInteger _size = new AtomicInteger();
-    private final AtomicInteger _pending = new AtomicInteger();
+    private final Set<ReservedThread> _threads = ConcurrentHashMap.newKeySet();
+    private final SynchronousQueue<Runnable> _queue = new SynchronousQueue<>(false);
+    private final AtomicBiInteger _count = new AtomicBiInteger(); // hi=pending; lo=size;
+    private final AtomicLong _lastEmptyTime = new AtomicLong(Long.MAX_VALUE);
```
sbordet: This should be initialized to `System.nanoTime()`.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -62,13 +71,14 @@ public String toString()
...
 
     private ThreadPoolBudget.Lease _lease;
-    private long _idleTime = 1L;
-    private TimeUnit _idleTimeUnit = TimeUnit.MINUTES;
+    private static final long DEFAULT_IDLE_TIMEOUT = TimeUnit.MINUTES.toNanos(1);
```
sbordet: Move it up with the `static`s.

sbordet (changes requested):

sbordet (approved):

lorban (approved):

PR #6540, opened 2021-07-26
Issue #6497 - Replace the Alias checkers with new implementation. (Jetty-9.4)

### LINKED ISSUE #6497: Replace SameFileAliasChecker
**Target Jetty version(s)**
9.4.x+

**Enhancement Description**
The current SameFileAliasChecker is almost a no-op based on changes before it is called.

We should address the 2 common usages.

1. Case Insensitive aliases - for windows requests (ala `/index.html` vs `/INDEX.HTML`)
2. FileSystem Normalization - for OSX and ZFS based filessytems that treat UTF-8 encoding differently on request vs storage.

lachlan-roberts:
## Issue #6497

Deprecate and replace the `SameFileAliasChecker` with the `AllowedResourceAliasChecker` which does some additional safety checks.

This new AliasChecker is extended to also replace the `AllowSymLinkAliasChecker` with `SymlinkAllowedResourceAliasChecker`.


jetty-server/src/main/java/org/eclipse/jetty/server/SymlinkAllowedResourceAliasChecker.java
```diff
@@ -0,0 +1,72 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import org.eclipse.jetty.server.handler.ContextHandler;
+import org.eclipse.jetty.util.log.Log;
+import org.eclipse.jetty.util.log.Logger;
+import org.eclipse.jetty.util.resource.Resource;
+
+/**
+ * An extension of {@link AllowedResourceAliasChecker} which only allows resources if they are symlinks.
```
gregw:
```suggestion
 * An extension of {@link AllowedResourceAliasChecker} which only allows aliased resources if they are symlinks.
```

jetty-server/src/main/java/org/eclipse/jetty/server/SymlinkAllowedResourceAliasChecker.java
```diff
@@ -0,0 +1,72 @@
...
+    @Override
+    public boolean check(String uri, Resource resource)
+    {
+        try
+        {
+            // Check the resource is allowed to be accessed.
+            if (!super.check(uri, resource))
+                return false;
+
+            // Only approve resource if it is accessed by a symbolic link.
```
gregw:
```suggestion
            // Approve resource if it is a symbolic link.
```

jetty-server/src/main/java/org/eclipse/jetty/server/SymlinkAllowedResourceAliasChecker.java
```diff
@@ -0,0 +1,72 @@
...
+            Path resourcePath = resource.getFile().toPath();
+            if (Files.isSymbolicLink(resourcePath))
+                return true;
+
+            // TODO: If base resource contains symlink then this will always return true.
+            //  But we don't want to deny all paths if the resource base is symbolically linked.
```
gregw:
```suggestion
            // or it has a symlink in its path to the resource base
```

jetty-server/src/main/java/org/eclipse/jetty/server/SymlinkAllowedResourceAliasChecker.java
```diff
@@ -0,0 +1,72 @@
...
+            if (super.hasSymbolicLink(resourcePath))
```
gregw: Call a varient of this method that only checks up to the resourceBase for a symbolicLink

jetty-server/src/main/java/org/eclipse/jetty/server/handler/AllowSymLinkAliasChecker.java
```diff
@@ -33,7 +33,9 @@
  * to check resources that are aliased to other locations.   The checker uses the
  * Java {@link Files#readSymbolicLink(Path)} and {@link Path#toRealPath(java.nio.file.LinkOption...)}
  * APIs to check if a file is aliased with symbolic links.</p>
+ * @deprecated use {@link org.eclipse.jetty.server.SymlinkAllowedResourceAliasChecker} instead.
  */
+@Deprecated
```
gregw, line 38: Perhaps replace the actual implementation with `SymlinkAllowedResourceAliasChecker`

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
```diff
@@ -0,0 +1,164 @@
...
+            return true;
+
+        // Lets try each path segment
+        Path base = path.getRoot();
+        for (Path segment : path)
+        {
+            base = base.resolve(segment);
+            if (Files.isSymbolicLink(base))
+                return true;
+        }
```
gregw: How about instead walk up the `getParent()`s and stop once you  get to a path that is same as baseResource?

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
```diff
@@ -0,0 +1,164 @@
...
+        catch (Throwable t)
+        {
+            LOG.warn(t);
+            return false;
+        }
+
+        return true;
+    }
+
+    protected boolean isProtectedPath(Path resourcePath, LinkOption[] linkOptions) throws IOException
```
gregw, line 144: javadoc as users might extend this method

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
```diff
@@ -0,0 +1,164 @@
...
+        String[] protectedTargets = _contextHandler.getProtectedTargets();
+        if (protectedTargets != null)
+        {
+            for (String s : protectedTargets)
+            {
+                // TODO: we are always following links for the base resource.
+                //  We cannot use toRealPath(linkOptions) here as it throws if file does not exist,
+                //  and the protected targets do not have to always exist.
+                String protectedTarget = new File(basePath, s).getCanonicalPath();
+                if (StringUtil.startsWithIgnoreCase(targetPath, protectedTarget))
```
gregw: This feels fragile.
It might be better to iterate over `getParent` to the basePath and test each parent in turn to see if it is protected.
If you don't reach the basePath, you get the main check for free.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
```diff
@@ -0,0 +1,164 @@
...
+
+    protected boolean isProtectedPath(Path resourcePath, LinkOption[] linkOptions) throws IOException
+    {
+        String basePath = Objects.requireNonNull(getPath(_contextHandler.getBaseResource())).toRealPath(linkOptions).toString();
+        String targetPath = resourcePath.toRealPath(linkOptions).toString();
+
+        if (!targetPath.startsWith(basePath))
+            return true;
+
+        String[] protectedTargets = _contextHandler.getProtectedTargets();
```
gregw: Ultimately we should make the a lifecycle and in doStart build a caseInsensitive Trie that we can lookup into.

gregw (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/SymlinkAllowedResourceAliasChecker.java
lachlan-roberts: That would cause the issue I described in the comment. If we do this and the base resource contains a symlink we will deny all files under the base resource, even though the only reason they are an alias is because the base resource contains symlink.

So doing that would break the capability of symlinking the base resource directory.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/AllowSymLinkAliasChecker.java
lachlan-roberts: They have different constructors. I could possibly combine them and add a new constructor to this class and deprecate the old one.

jetty-server/src/main/java/org/eclipse/jetty/server/SymlinkAllowedResourceAliasChecker.java
gregw: The purpose of this checker is to allow aliases if they are symlinks within to scope of the webapp.   
So in the case of `/var/lib/jetty-home/webapps/mycontext/foo/bar/bob.txt` we would walk up the parent tree and if any of the following:
 - `/var/lib/jetty-home/webapps/mycontext/foo/bar/bob.txt`
 - `/var/lib/jetty-home/webapps/mycontext/foo/bar`
 - `/var/lib/jetty-home/webapps/mycontext/foo`

Are symlinks, then we allow the alias.   If the alias is above that point, then we don't even look.  It is not allowed.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/AllowSymLinkAliasChecker.java
gregw: Ah yes.... would have to make it a lifecycle and look for the threadload context in doStart.  Possible, but not worth it.

jetty-server/src/main/java/org/eclipse/jetty/server/SymlinkAllowedResourceAliasChecker.java
lachlan-roberts: I am fine with this change, but it is worth noting that it is a possible breaking change for people who are already relying of that behaviour from `AllowSymLinkAliasChecker`.

So if they run into this issue they can be directed to configure `AllowedResourceAliasChecker` which will allow this use case.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
lachlan-roberts: done

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
lachlan-roberts: done

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
```diff
@@ -0,0 +1,217 @@
...
+        _basePath = getPath(_contextHandler.getBaseResource());
+        if (_basePath == null)
+            throw new IllegalStateException("Could not obtain base resource path");
+
+        String[] protectedTargets = _contextHandler.getProtectedTargets();
+        if (protectedTargets != null)
+        {
+            for (String s : protectedTargets)
+            {
+                _protectedPaths.add(new File(_basePath.toFile(), s).toPath());
```
lachlan-roberts: This is a weird way to add the paths together, but doing `path.resolve(...)` does not work because the protectedTarget can start with "/".

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
```diff
@@ -0,0 +1,217 @@
...
+    private final HashSet<Path> _protectedPaths = new HashSet<>();
+    private Path _basePath;
+
+    /**
+     * @param contextHandler the context handler to use.
+     * @param checkSymlinkTargets true to check that the target of the symlink is an allowed resource.
+     */
+    public AllowedResourceAliasChecker(ContextHandler contextHandler, boolean checkSymlinkTargets)
+    {
+        _contextHandler = contextHandler;
```
lachlan-roberts, line 62: @gregw I could not do this with the `ContextHandler.getContext.getContextHandler()` because it is not always defined when `doStart` is called.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
```diff
@@ -0,0 +1,217 @@
...
+
+        for (Path protectedPath : _protectedPaths)
+        {
+            String protect;
+            if (Files.exists(protectedPath, linkOptions))
+                protect = protectedPath.toRealPath(linkOptions).toString();
+            else if (linkOptions == NO_FOLLOW_LINKS)
+                protect = protectedPath.normalize().toAbsolutePath().toString();
+            else
+                protect = protectedPath.toFile().getCanonicalPath();
```
lachlan-roberts, line 166: I don't like this code but I cant think of a better alternative. The call to `toRealPath` requires the file to exist otherwise it throws, but we must check this anyway, because `/web-inf` might not exist but `/WEB-INF` could exist and so the `Files.exists` call would return false.

lachlan-roberts:
@gregw can you re-review, I had to make some additional changes since last hangout, I have explained why in the comments.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
gregw: How can that be?   Can you explain?

lachlan-roberts: If an alias checker is added after the server is started (which we currently allow), then `ContextHandler.getCurrentContext().getContextHandler()` will throw NPE.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
gregw: why not use `URIUtil.addPaths` ?

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
gregw: I think we have made a rod for our backs by specifying the protected targets as lowercase `/web-inf`.
However, I very much doubt that anybody other than us adds any protected targets,  so why don't we just fix this!
We can change our protected targets to use the correct cases and throw an ISE if anybody calls it with any other protected targets (or if the set a protected target that doesn't exist).

Then in doStart we can find the actual `Path`s of all the protected targets (that exist) and replace all this with the logic I prefer... walking up the parent links looking for any path that is sameFile to any of the protected targets

gregw (changes requested):
I'm torn! 
See my comments about why we shouldn't just fix the protected targets?

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
gregw: Then is it worthwhile adding a doStart at all?  I don't think so if we are just doing the string comparisons.
If we look for the real `Path`s of the protected files, then it is probably worthwhile still.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
lachlan-roberts: It has the signature `(String, String)` instead of `(Path, String)`, and it returns a `String` instead of a `Path`.
And we need the Path so we can use the linkOptions to canonicalize it.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
lachlan-roberts: Its probably worthwhile as we don't have to repeat the work to resolve the basePath and protectedTaget paths every time `check()` is called.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
lachlan-roberts: I don't think its a good idea, we might not think anyone is calling that method, but that's just a guess.
If someone is using it and we change it, then it will expose directories that were previously protected.

Doesn't seem like something we want going into a minor release, especially for `9.4`.

lachlan-roberts: Actually maybe we still wouldn't expose it because its still protected by `ContextHander.isProtectedTarget` which is case insensitive.

Still doesn't feel like a good idea. Maybe open another issue to consider changing protectedTargets to be case sensitive. We could still change the logic of this later on, but for now it seems like its doing the job of not allowing access to protected targets.

gregw: I think we can check if somebody sets a new protected target that does not exist, so I'm not that worried about breaking any usage of that.

The check that we want to make is that the resource is in the baseResource directory, but is not in any of the protected resource directories.    We are trying to do that in a FS independent ways  with case insensitive string-starts checks, but they are ultimately just hacks.

If we can precisely identify the `Path` of the base resource and the `Path`s of the protected resources.  We can then walk the parent linkage of the resource and check that none of it's parents are sameFile as the protected resource and that ultimately one of the parents is sameFile as base resource - we then have an absolutely precise check that does exactly what we want.

gregw: Maybe we do this only in 10? ... but if we are adding new recommended checkers in 9 and deprecating the existing ones, we may as well add ones that we have 100% confidence in.

gregw:
@lachlan-roberts So I'm thinking that perhaps we do go forward with the string version for jetty-9, but for jetty-10/11 we use #6567 as a basis and then have a version that is based entirely on `Path`s
If so, I think we should work on the 10/11 version first, before committing this PR, to make sure any lessons learnt there can be echo'ed here to ensure a smooth transition.

sbordet:
@lachlan-roberts still failing tests, can you make this PR pass the tests?

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
sbordet: It's really strange that we have a `Path`, we call `toFile()`, create a `File` and then back to `toPath()`.
I'm surprised there is not a way to just use `Path` without using the `File` APIs.
If the `File` APIs are necessary, needs a comment.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
```diff
@@ -0,0 +1,218 @@
...
+    {
+        return _basePath;
+    }
+
+    @Override
+    protected void doStart() throws Exception
+    {
+        _basePath = getPath(_contextHandler.getBaseResource());
+        if (_basePath == null)
+            _basePath = Paths.get("/").toAbsolutePath();
```
sbordet, line 81: Why calling `toAbsolutePath()`? Isn't just a single `/`?
Is it because of Windows? If so, needs a comment.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
sbordet: Why the `LinkOption[]`? Why not `LinkOption...` instead?

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
```diff
@@ -0,0 +1,218 @@
...
+                // Check that the target prefix really is a path segment.
+                if (target.charAt(protect.length()) == File.separatorChar)
+                    return true;
+            }
+        }
+
+        return false;
+    }
+
+    protected boolean hasSymbolicLink(Path path)
```
sbordet, line 183: Shouldn't this be called `isSymbolicLink()`?

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
```diff
@@ -0,0 +1,218 @@
...
+        Path resourcePath = getPath(resource);
+        if (resourcePath == null)
+            return false;
+
+        try
+        {
+            if (isProtectedPath(resourcePath, NO_FOLLOW_LINKS))
+                return false;
+
+            if (_checkSymlinkTargets && hasSymbolicLink(resourcePath))
```
sbordet, line 117: Why is this check here? Should it not be in the `SymlinkAllowedResourceAliasChecker` subclass?
I mean that seems weird that this base class does a bit of symlink but not all... Can't all symlink handling and configuration fields be pushed to the subclass?

jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -264,9 +267,8 @@ private ContextHandler(Context context, HandlerContainer parent, String contextP
         _scontext = context == null ? new Context() : context;
         _attributes = new AttributesMap();
         _initParams = new HashMap<>();
-        addAliasCheck(new ApproveNonExistentDirectoryAliases());
         if (File.separatorChar == '/')
-            addAliasCheck(new AllowSymLinkAliasChecker());
+            addAliasCheck(new SymlinkAllowedResourceAliasChecker(this));
```
sbordet, line 271: Why only add this for `/`?
Also, the class javadoc seems to be in contrast with what is configured here in code.

sbordet (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
joakime: This won't do what you expect on Windows.

If you are running from a Windows Share for example, take `\\machine\sharename\jetty` your code will result in a Path that is no longer in the share location.
Is that what you want?

The Path has a FileSystem object associated with it, you can ask for its root, or even for all roots that the FileSystem layer supports.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
joakime: A Path can have 0..n symbolic links.
I like "hasSymbolicLink".

jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
joakime: This is tricky.

Not all FileSystems on Linux or OSX support symlinks.
And some FileSystems on Windows support symlinks.

Basing this default off of File.separatorChar is not wholly accurate.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
joakime: The `FileSystem.getPath(String first, String ... more)` API supports this call structure.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
sbordet: But that's not what the method does. It does not check if there is some other path that links to this path.
It checks whether this path (or one of its parents) _is_ a symbolic link.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    protected boolean hasSymbolicLink(Path path)
+    {
+        return hasSymbolicLink(path.getRoot(), path);
+    }
+
+    protected boolean hasSymbolicLink(Path base, Path path)
+    {
+        for (Path p = path; (p != null) && !p.equals(base); p = p.getParent())
```
sbordet: Also, I'm not sure the for condition `!p.equals(base)` works, unless it is guaranteed that base is a normalized parent path  of `path`: `Path.of("/home/simon/..").equals("/home") => false`.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
lachlan-roberts: It should not be called `isSymbolicLink`, within this method we are calling `Files.isSymbolicLink()` which has a different semantic. The `isSymbolicLink` method only returns true if the file itself is a symlink, and doesn't check the parent directories in the path.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
lachlan-roberts: This is a different check to what is done in the subclass, the subclass does not even use this logic, it sets `_checkSymlinkTargets`  to false.

This is an optional check to ensure the target of the symlink is also an allowed resource.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
lachlan-roberts: This is pre-existing behaviour which we are maintaining to not break existing deployments.

gregw: This should be fine, because if a FS does not support symlinks, then this checker will never see a symlink and thus will be a noop

gregw:
@lachlan-roberts @sbordet @joakime   I think the work being done in this PR is important, but I'm concerned that we are not converging on a solution that is easily understandable.

So I think it is worthwhile to take a step back a re-describe what is being attempted here, then discuss the possible algorithms to achieve that.   There are two semantics we are trying to achieve here:
 + `AllowedResourceAliasChecker` it checking that real file of the targeted resource is within the real file of the base resource and not within the real file directory of any protected resources.  This will allow a user to support soft aliasing (eg case insensitive files and 8.3 names), but at the risk of bypassing some security constraints.
 + `SymlinkAllowedResourceAliasChecker` is an extension of that will allow a real that is outside of base or a protected resource, but only if it is simlinked from a file that passes the `AllowedResoureAliasChecker`.   It is for simlinks within a webapp that point outside of base or to a protected resource, which  I don't  think this is a big use-case.   The original `AllowSymLinkAliasChecker` was added really to support symlinks to a jetty-base, webapps or webapp war/directory.  This is a use-case that I think is much better dealt with by #6567 which has been merged to 10, but could be back ported to 9.  It prevents the baseResource from itself being an alias and thus making the alias checkers fire on every request.

I think backporting #6567 to 9 is a bit big for a dot release, but it is probably better than continuing with the mostly broken `AllowSymLinkAliasChecker`, but then I think the algorithms below work without it anyway.. they will just be triggered less often.

My preferred algorithm for an  `AllowedResourceAliasChecker` is that at `doStart` we determine the real path (following links) of the base resource and all protected resources (this is made a little difficult as we assume case insensitivity... see below).   The to check if a aliased resource is OK we determine the real path (following links) of the resource and then:
1. if the resource is null or does not exist return false
2. if the resource is the sameFile as any protected resource return false
3. if the resource is the sameFile as the base resource return true
4. assign resource = resource.getParent loop back to 1.

I think this is a portable (somewhat expensive) check that will precisely determine if resource references a file within the real base resource that is not protected.

My preferred algorithm for `SymlinkAllowedResourceAliasChecker` is much the same, but we start the loop with the realpath (NOT following links) of the resource, so we check precisely that the link itself is in the base resource and not protected.

To achieve this, we will need to stop defining protected paths in a case insensitive way. We currently define "web-inf" and "meta-inf", which will not find the real protected targets on linux (hence this PR tries to get around it with case insensitive string checks). We can change this to "WEB-INF" and "META-INF", but if any users are use the API to add their own protected targets, then we will have a hard time telling they don't exist or are just case insensitive.   I think we can work around this by: deprecating the public API and making it internal; failing to deploy if there are any user defined protected targets that do not exist in the base resource when we start.

PR #6542, opened 2021-07-26
Jetty 9.4.x improve blocking array queue test perf

### LINKED ISSUE #6541: Improve BlockingArrayQueueTest to remove @DisabledIfSystemProperty markers
Improve the performance of `testTake()` and `testConcurrentAccess()` such as they can always be enabled.

Relates to #5684 and #6327

lorban:
See #6541.

`testTake()` was modified to not sleep. Execution is now as fast as the CPU goes.

`testConcurrentAccess()` was cleaned up and modified to introduce some execution randomness with sleeps but cutting them as short as they can usefully be. Execution is now down to < 1s instead of 6+s.

### DIFF
--- a/jetty-util/src/test/java/org/eclipse/jetty/util/BlockingArrayQueueTest.java
+++ b/jetty-util/src/test/java/org/eclipse/jetty/util/BlockingArrayQueueTest.java
@@ -32,9 +32,11 @@ import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.DisabledIfSystemProperty;
 
+import static org.awaitility.Awaitility.await;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNull;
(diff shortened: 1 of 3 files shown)


sbordet (dismissed):

jetty-util/src/test/java/org/eclipse/jetty/util/BlockingArrayQueueTest.java
```diff
@@ -196,56 +198,55 @@ public void run()
 
         thread.start();
 
-        Thread.sleep(1000);
+        // Wait until the spawned thread is blocked in queue.take().
+        await().atMost(5, TimeUnit.SECONDS).until(() -> thread.getState() == Thread.State.WAITING);
```
gregw, line 205: This is all very cool and all.... but do we have CQ to allow us to bring in this dependency?  Do we need CQs for tests???
Isn't it easier to write the 5 lines of code to do this spin and test rather than answer the CQ questions?

gregw (changes requested):

jetty-util/src/test/java/org/eclipse/jetty/util/BlockingArrayQueueTest.java
lorban: I agree because I only used the most basic functionality of awaitility, so it was easy enough to substitute it with some helper.

gregw (approved):

PR #6543, opened 2021-07-26
10.0.x: Add defensive check in HttpConnection content production

### LINKED ISSUE #6491: onDataAvailable() not called when HttpParser is closed prematurely

lorban:
Port of #6533 to 10.0.x branch.

Currently, if the parser gets closed and the connection isn't aborted, the code enters a select/wakeup/fillAndParseForContent/wait infinite loop.

The above should never happen unless a bug gets introduced, but if such bug would happen at least the situation would be detected and acted upon.

See #6491.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannelOverHttp.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannelOverHttp.java
@@ -90,14 +90,7 @@ public class HttpChannelOverHttp extends HttpChannel implements HttpParser.Reque
                 LOG.debug("needContent has content immediately available: {}", _content);
             return true;
         }
-        try
-        {
-            _httpConnection.parseAndFillForContent();
-        }
-        catch (Throwable x)
-        {
-            _content = new HttpInput.ErrorContent(x);
-        }
+        parseAndFillForContent();
         if (_content != null)
         {
             if (LOG.isDebugEnabled())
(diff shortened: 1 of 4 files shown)


sbordet (dismissed):

sbordet:
@lorban please check why the tests are failing?

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannelOverHttp.java
```diff
@@ -111,7 +118,14 @@ public boolean needContent()
...
                 LOG.debug("produceContent has no content, parsing and filling");
-            _httpConnection.parseAndFillForContent();
+            try
+            {
+                _httpConnection.parseAndFillForContent();
+            }
+            catch (Throwable x)
+            {
+                _content = new HttpInput.ErrorContent(x);
+            }
```
sbordet: Please refactor into a common method the two occurrences.

sbordet (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannelOverHttp.java
lorban: Done.

sbordet (approved):

PR #6557, opened 2021-07-28
Issue #6556 Ensure context classloader set when operating on memcache.

### LINKED ISSUE #6556: MemcachedSessionDataMap needs to set the context classloader before serialization/deseriazliation.
Because the MemcachedSessionDataMap does not use ContextHandler.handle(Runnable), the context classloader is not set correctly when the SessionData is serialized into memcached.

janbartel:
Closes #6556


jetty-memcached/jetty-memcached-sessions/src/main/java/org/eclipse/jetty/memcached/session/MemcachedSessionDataMap.java
```diff
@@ -140,9 +142,13 @@ public void setHeartbeats(boolean heartbeats)
...
     public void initialize(SessionContext context)
     {
+        if (isStarted())
+            throw new IllegalStateException("Context set after MemcachedSessionDataMap started");
+        
         try
         {
-            _builder.setTranscoder(new SessionDataTranscoder());
+            _context = context;
+            _builder.setTranscoder(new SessionDataTranscoder(_context.getContext().getClassLoader()));
```
olamy: there is no such constructor. typo?

olamy (changes requested):

jetty-memcached/jetty-memcached-sessions/src/main/java/org/eclipse/jetty/memcached/session/MemcachedSessionDataMap.java
janbartel: ooops, I didn't realize that constructor was one you added in your jetty-10.0.x-distribution-session-tests branch ;) Fixed now.

olamy (dismissed):

olamy (dismissed):

jetty-util/src/main/java/org/eclipse/jetty/util/FuturePromise.java
```diff
@@ -118,6 +118,27 @@ public C get() throws InterruptedException, ExecutionException
             throw (CancellationException)new CancellationException().initCause(_cause);
         throw new ExecutionException(_cause);
     }
+    
+    /**
+     * Return the result if completed successfully
+     * or rethrow the Exception
```
gregw: Need to be clearer here that the `Exception` thrown is not wrapped with `ExecutionException` like it is with `get()`.

jetty-util/src/main/java/org/eclipse/jetty/util/FuturePromise.java
```diff
@@ -118,6 +118,27 @@ public C get() throws InterruptedException, ExecutionException
...
+     * 
+     * @return the computed result
+     * @throws Exception
```
gregw: Note that it can rethrow an `Error` as well as an `Exception`

jetty-memcached/jetty-memcached-sessions/src/main/java/org/eclipse/jetty/memcached/session/MemcachedSessionDataMap.java
```diff
@@ -155,14 +161,48 @@ public void initialize(SessionContext context)
...
+            {
+                result.succeeded(_client.get(id));
+            }
+            catch (Exception e)
+            {
+                result.failed(e);
+            }
+        };
+
+        _context.run(r);
```
gregw, line 181: Elsewhere in the codebase we have done something like:
```suggestion
        if (_context == null)
            r.run();
        else
            _context.run(r);
```
I'm not sure if sessions can work without contexts???

gregw (changes requested):

jetty-util/src/main/java/org/eclipse/jetty/util/FuturePromise.java
janbartel: Done.

jetty-util/src/main/java/org/eclipse/jetty/util/FuturePromise.java
janbartel: Done.

jetty-memcached/jetty-memcached-sessions/src/main/java/org/eclipse/jetty/memcached/session/MemcachedSessionDataMap.java
janbartel: The `_context` here is not a `ContextHandler` but a `SessionContext` that has a `run()` method that implements this very recipe, see https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-server/src/main/java/org/eclipse/jetty/server/session/SessionContext.java#L84

gregw (approved):

PR #6563, opened 2021-07-30
Fix #6562 last written bytebuffer

### LINKED ISSUE #6562: HttpOutput.write(ByteBuffer buffer)
**Jetty version** 9.4.43

**Java version** 1.8.0_261

**Question**
I am using the methods `HttpOutput.write(ByteBuffer buffer)` and `HttpOutput.write(byte[] b, int off, int len)` to send the response. Also I am using GzipHandler. I am testing with the same response data. When I use the `write(byte[] b, int off, int len)` method, the _Content-Length_ header is set in the response. When I use the `write(ByteBuffer buffer)` method, the _Transfer-Encoding: chunked_ header is set in the response. I only call the method once. I suspect the `last` variable is not being set correctly. Could this be just a misprint?

`HttpOutput.write(int b)`:
https://github.com/eclipse/jetty.project/blob/735e97d5c7e41a39e759a5d76e45f0b5c1226209/jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java#L1002
`HttpOutput.write(byte[] b, int off, int len)`:
https://github.com/eclipse/jetty.project/blob/735e97d5c7e41a39e759a5d76e45f0b5c1226209/jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java#L796
`HttpOutput.write(ByteBuffer buffer)`:
https://github.com/eclipse/jetty.project/blob/735e97d5c7e41a39e759a5d76e45f0b5c1226209/jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java#L926

gregw:
Fixes #6562 the last written bytebuffer calculation.
Also fixed an associated issue with unnecessary flush of an empty when last calculation already signalled last.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
@@ -863,12 +863,10 @@ public class HttpOutput extends ServletOutputStream implements Runnable
         // Blocking write
         try
         {
-            boolean complete = false;
             // flush any content from the aggregate
             if (BufferUtil.hasContent(_aggregate))
             {
-                complete = last && len == 0;
-                channelWrite(_aggregate, complete);
+                channelWrite(_aggregate, last && len == 0);
 
                 // should we fill aggregate again from the buffer?
                 if (len > 0 && !last && len <= _commitSize && len <= maximizeAggregateSpace())
@@ -897,10 +895,6 @@ public class HttpOutput extends ServletOutputStream implements Runnable
                     len -= getBufferSize();
                 }
                 channelWrite(view, last);
-            }
-            else if (last && !complete)
-            {
-                channelWrite(BufferUtil.EMPTY_BUFFER, true);
             }
 
             onWriteComplete(last, null);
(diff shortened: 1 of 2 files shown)


gregw:
@lorban @lachlan-roberts In an attempt to get 100% code coverage, I could not find a path to get to the final last emptybuffer write. So I have removed it and I think it is good.  But give me extra good review on that deletion.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
```diff
@@ -896,10 +896,6 @@ public void write(byte[] b, int off, int len) throws IOException
                 }
                 channelWrite(view, last);
             }
-            else if (last)
-            {
-                channelWrite(BufferUtil.EMPTY_BUFFER, true);
```
lorban, line 901: Serving a 0 length `byte[]` goes through this branch. This is why `GzipContentLengthTest.testServletTypeLengthStreamWrite()` fails for scenario `empty.txt`.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
```diff
@@ -973,8 +969,6 @@ public void write(ByteBuffer buffer) throws IOException
                 // write any remaining content in the buffer directly
                 if (len > 0)
                     channelWrite(buffer, last);
-                else if (last)
```
lorban, line 976: Serving a 0 length `byte[]` goes through this branch. This is why `GzipContentLengthTest.testHttpOutputWrite()` fails for scenario `empty.txt`.

lorban (changes requested):
You should probably add some tests to `HttpOutputTest` to assert that writing 0 length buffers does work.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
gregw: Ah that's how it can be called!!! Restored it and added test cases in `HttpOutputTest` so it is now covered!

lorban (approved):

Issue #6327, opened 2021-05-27
Review disabled tests in Jetty 10.x


janbartel:
See #5684 which is the list of disabled tests in jetty-9.4.x. This issue is tracking the list of disabled tests in jetty-10.0.x instead.

- [x] [demos/demo-jetty-webapp/src/test/java/org/eclipse/jetty/TestServer.java#L51](https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/demos/demo-jetty-webapp/src/test/java/org/eclipse/jetty/TestServer.java#L51):  ` @Disabled("Not a test case")`
- [x] [demos/embedded/src/test/java/org/eclipse/jetty/demos/ServerWithJMXTest.java#L30](https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/demos/embedded/src/test/java/org/eclipse/jetty/demos/ServerWithJMXTest.java#L30):  ` @Disabled`
- [ ] [jetty-alpn/jetty-alpn-conscrypt-client/src/test/java/org/eclipse/jetty/alpn/java/client/ConscryptHTTP2ClientTest.java#L52](https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-alpn/jetty-alpn-conscrypt-client/src/test/java/org/eclipse/jetty/alpn/java/client/ConscryptHTTP2ClientTest.java#L52):     ` @Disabled("issue google/conscrypt#667")`
- [ ] [jetty-client/src/test/java/org/eclipse/jetty/client/api/Usage.java#L46](https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-client/src/test/java/org/eclipse/jetty/client/api/Usage.java#L46):  ` @Disabled`
- [ ] [jetty-client/src/test/java/org/eclipse/jetty/client/ExternalSiteTest.java#L34](https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-client/src/test/java/org/eclipse/jetty/client/ExternalSiteTest.java#L34):  ` @Disabled`
- [x] [jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientRedirectTest.java#L290](https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientRedirectTest.java#L290):     ` @Disabled`
[...]

PR #6564, opened 2021-07-30
Re-enable disabled tests

lorban:
See #6327.

Re-enable another set of tests disabled because they weren't reliable.

Besides concentrating on the tests themselves, I'd like to collect opinions about the style I use to fix those tests, i.e.: remove any dependency on timing and use awaitility's wait-for-some-time-until-some-state-changed mechanism.

This requires exposing more of the internals to tests, sometimes with ugly helper classes to access package-protected methods like I do in this example, but has the benefits of being strictly correct and requiring as little time as needed, hence shortening our test runs.


jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
```diff
@@ -1510,6 +1510,12 @@ public void run()
         }
     }
 
+    // For testing
+    ApiState getApiState()
+    {
+        return _apiState;
+    }
```
joakime: I personally do not have a problem with exposing internals for testing reasons.

Especially in these kinds of deep classes where extending them is more complicated than it's worth.

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/AsyncIOServletTest.java
```diff
@@ -1854,4 +1866,35 @@ public void stopServer() throws Exception
             super.stopServer();
         }
     }
+
+    static class Await
```
joakime: Should this go in some kind of central test jar for all of our modules to use?

Perhaps we can extend build-resources for something like this.

Or have a `<classifier>tests</classifier>` specific artifact on jetty-util?

joakime: I also have no problem with making a new (independently versioned) jar in jetty-toolchain for these kinds of things.

gregw: Since we already have the jetty-toolchain used in our testing, I'd prefer adding these kind of simple utility classes to that rather than bringing in a new dependency.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
gregw: The state is provided in the toString output, so a test that checks for it as a substring is probably as robust as this and doesn't require an extra method.

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/AsyncIOServletTest.java
janbartel: I like the Await class. Might be able to use it in some of the session tests. Should it just go into jetty-util?

janbartel (changes requested):
I like it. I agree with @gregw comments about using the HttpOutput.toString output instead of exposing the ApiState. I like the Await class very much.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
lorban: Okay, I got rid of the extra method and extracted the api state from `toString`.

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/AsyncIOServletTest.java
lorban: Since there seems to be some interest in what the `Await` class does, I've decided to get rid of it and add a test dependency on awaitility as `Await` was just a cut-down version of that lib that was so far used only once.

lorban:
Since there seems to be a generally favorable opinion of what the `Await` class does, I've replaced it with its original version which is the awaitility testing lib as I'm planning to use it in more disabled tests.

joakime:
> Since there seems to be a generally favorable opinion of what the `Await` class does, I've replaced it with its original version which is the awaitility testing lib as I'm planning to use it in more disabled tests.

Don't forget to file the CQ for it.

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/AsyncIOServletTest.java
```diff
@@ -431,7 +428,19 @@ public void onWritePossible() throws IOException
 
                         // Wait for the failure to arrive to
                         // the server while we are about to write.
-                        sleep(2000);
+                        try
+                        {
+                            await().atMost(5, TimeUnit.SECONDS).until(() ->
+                            {
+                                out.write(new byte[0]);
```
sbordet: Why the `write()`?

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/HttpClientContinueTest.java
```diff
@@ -319,12 +319,10 @@ public void onComplete(Result result)
 
     @ParameterizedTest
     @ArgumentsSource(TransportProvider.class)
-    @Tag("Slow")
-    @DisabledIfSystemProperty(named = "env", matches = "ci") // TODO: SLOW, needs review
     public void testExpect100ContinueWithContentWithResponseFailureBefore100Continue(Transport transport) throws Exception
     {
         init(transport);
-        long idleTimeout = 1000;
+        long idleTimeout = 100;
```
sbordet: I'm not sure I like this, as 100 ms may be a too short time for the request to even arrive to the server due to class loading or GC on the client.

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/HttpClientContinueTest.java
```diff
@@ -366,12 +364,10 @@ public void onComplete(Result result)
 
     @ParameterizedTest
     @ArgumentsSource(TransportProvider.class)
-    @Tag("Slow")
-    @DisabledIfSystemProperty(named = "env", matches = "ci") // TODO: SLOW, needs review
     public void testExpect100ContinueWithContentWithResponseFailureAfter100Continue(Transport transport) throws Exception
     {
         init(transport);
-        long idleTimeout = 1000;
+        long idleTimeout = 100;
```
sbordet: Ditto.

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/HttpClientContinueTest.java
```diff
@@ -507,35 +512,40 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
...
+        await().atMost(5, TimeUnit.SECONDS).until(() ->
+        {
+            Thread thread = handlerThread.get();
+            return thread != null && thread.getState() == Thread.State.WAITING;
+        });
 
         content.offer(ByteBuffer.wrap(chunk1));
 
-        Thread.sleep(1000);
+        assertTrue(demandLatch.await(5, TimeUnit.SECONDS));
```
sbordet: Before, the 1 second sleep kinda made sure the first chunk arrived to the server.
Now, the 2 chunks may arrive to the server in one chunk, as the wait has been practically zeroed, so the test is now different from before.

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/HttpClientStreamTest.java
```diff
@@ -680,20 +678,29 @@ public void handle(String target, org.eclipse.jetty.server.Request baseRequest,
...
                 .send(result ->
                 {
                     if (result.isSucceeded() && result.getResponse().getStatus() == 200)
-                        latch.countDown();
+                        requestLatch.countDown();
                 });
 
             // Make sure we provide the content *after* the request has been "sent".
-            Thread.sleep(1000);
+            assertTrue(demandLatch.await(5, TimeUnit.SECONDS));
```
sbordet: Again, the latch could be counted down really quickly, so the content will be available before the request is sent, contrary to the purpose of the test.

sbordet (changes requested):

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/HttpClientContinueTest.java
sbordet: * await on client until read() blocks.
* read exactly chunk1.length bytes
* now the next read on server will block, count down a latch to tell main thread to await again
* block on read() on server
* await until read() blocks
* offer second chunk.
* read second chunk on server
* wait for successful response on client

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/HttpClientContinueTest.java
sbordet: * use AtomicReference to hold `Request`.
* send request
* on server get client-side request and abort it instead of using client-side idle timeout as a mean to cause a failure.
* on server, wait for the request to be completed on the client, then exit `handle()`

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/HttpClientContinueTest.java
sbordet: Similar steps as above.

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/AsyncIOServletTest.java
sbordet: * cast to `HttpOutput` and call a direct API that does not aggregate to write some content
* wait until a write fails -- the until() loop should be `write()+isReady()` and check for some "failed" state (because write() does not throw)
* data to write can just be an array of zeroes
* remove `write()` outside until() loop.

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/HttpClientStreamTest.java
lorban: Indeed. Counting down the latch in a `onRequestCommit` listener would solve that problem, wouldn't it?

lorban:
CQ 23637 filed.

jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTest.java
```diff
@@ -690,10 +689,9 @@ public void onComplete(Result result)
 
     @ParameterizedTest
     @ArgumentsSource(ScenarioProvider.class)
-    @DisabledIfSystemProperty(named = "env", matches = "ci") // TODO: SLOW, needs review
     public void testRequestIdleTimeout(Scenario scenario) throws Exception
     {
-        long idleTimeout = 1000;
+        long idleTimeout = 100;
```
sbordet: Too short as discussed.

jetty-fcgi/fcgi-server/src/test/java/org/eclipse/jetty/fcgi/server/HttpClientTest.java
```diff
@@ -437,7 +435,7 @@ public void handle(String target, org.eclipse.jetty.server.Request baseRequest,
         assertThrows(TimeoutException.class, () ->
             client.newRequest(host, port)
                 .scheme(scheme)
-                .idleTimeout(idleTimeout, TimeUnit.MILLISECONDS)
+                .idleTimeout(idleTimeout * 90 / 100, TimeUnit.MILLISECONDS)
```
sbordet: This yields a difference between client and server of just 100 ms, which could make the test flaky.

jetty-fcgi/fcgi-server/src/test/java/org/eclipse/jetty/fcgi/server/HttpClientTest.java
```diff
@@ -437,7 +435,7 @@ public void handle(String target, org.eclipse.jetty.server.Request baseRequest,
...
                 .timeout(3 * idleTimeout, TimeUnit.MILLISECONDS)
```
sbordet: In this test there are a number of places that refer to `3 * idleTimeout` which made sense when the server was waiting `2 * idleTimeout`, but now raise the question "why wait 3x"?

I'm not particularly opposed to reduce the `idleTimeout = 500` to speed up the test, but I think this is one test we don't want to speed up -- it is testing idle timeouts and it has to wait.

jetty-io/src/test/java/org/eclipse/jetty/io/SelectorManagerTest.java
```diff
@@ -117,8 +121,7 @@ protected void connectionFailed(SelectableChannel channel, Throwable ex, Object
             SocketChannel client1 = SocketChannel.open();
             client1.configureBlocking(false);
             client1.connect(address);
-            long timeout = connectTimeout * 2;
-            timeoutConnection.set(timeout);
+            timeoutConnection.set(connectTimeout * 110 / 100);
```
sbordet: Again, 100 ms difference is too small.
You would need to wait until `latch1` is counted down, then release the thread in `doFinishConnect()`.
Also, `3 * idleTimeout` should probably be changed (see previous comment).

sbordet (changes requested):

jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTest.java
lorban: Ah yes. This one I forgot to rework it, but since it's almost identical to the FCGI one below, let's only discuss that other one and I'll make them both work the same way.

jetty-fcgi/fcgi-server/src/test/java/org/eclipse/jetty/fcgi/server/HttpClientTest.java
lorban: Not in this case. Since this test is about triggering an idle timeout, as long as the server waits more than what the client's idle timeout, this test does its job. It does not matter if the response was too late because of a GC pause or because of the sleep as long as we can guarantee that the server will take more time than the client is willing to wait. So if we can make sure the server will wait at least 100ms more, that's good enough.

But maybe I should make the server wait `idleTimeout * 110 / 100` and make the client use `idleTimeout = 1000` as its timeout to improve clarity a bit.

jetty-fcgi/fcgi-server/src/test/java/org/eclipse/jetty/fcgi/server/HttpClientTest.java
lorban: The 3X timeout does not have an impact on the time the test takes: it's just the upper time bound that would make the test fail. I could hardcode it to 3s or 5s but that doesn't matter too much as long as the test can entirely run during that time.

jetty-io/src/test/java/org/eclipse/jetty/io/SelectorManagerTest.java
lorban: It's the same problem as above: it doesn't matter if it's 100ms more or 1s more as long as the connection delay is higher than the timeout.

jetty-io/src/test/java/org/eclipse/jetty/io/SelectorManagerTest.java
```diff
@@ -120,39 +114,36 @@ protected void connectionFailed(SelectableChannel channel, Throwable ex, Object
...
             {
                 @Override
                 public void failed(Throwable x)
                 {
-                    latch1.countDown();
+                    failedConnectionLatch.countDown();
                 }
             });
-            assertTrue(latch1.await(connectTimeout * 3, TimeUnit.MILLISECONDS));
+            assertTrue(failedConnectionLatch.await(connectTimeout * 2, TimeUnit.MILLISECONDS));
```
sbordet, line 126: `failedConnectionLatch` has count=1, but it is awaited twice, which does not seem correct.

jetty-server/src/test/java/org/eclipse/jetty/server/ConnectionOpenCloseTest.java
```diff
@@ -153,15 +147,13 @@ public void onClosed(Connection connection)
             assertTrue(closeLatch.await(5, TimeUnit.SECONDS));
 
             // Wait some time to see if the callbacks are called too many times
-            TimeUnit.SECONDS.sleep(1);
+            TimeUnit.MILLISECONDS.sleep(200);
```
sbordet, line 150: I feel again this is a too short time compared to before.

jetty-server/src/test/java/org/eclipse/jetty/server/ConnectionOpenCloseTest.java
```diff
@@ -223,7 +215,7 @@ public void onClosed(Connection connection)
         assertTrue(closeLatch.await(5, TimeUnit.SECONDS));
 
         // Wait some time to see if the callbacks are called too many times
-        TimeUnit.SECONDS.sleep(1);
+        TimeUnit.MILLISECONDS.sleep(200);
```
sbordet, line 218: Ditto.

sbordet (changes requested):

jetty-server/src/test/java/org/eclipse/jetty/server/ConnectionOpenCloseTest.java
lorban: There is a 3rd test in the same class (`testOpenClose`) that uses a 200ms time to count the exact same events. I think the three delays should be unified, and I don't think waiting too long will make the test more reliable; either the code is correct and the delay won't change anything or we introduce some bug and eventually, and only eventually, those tests will trip on those assertions. Making the delay longer will only make the eventuality more common.

jetty-server/src/test/java/org/eclipse/jetty/server/ConnectionOpenCloseTest.java
lorban: See above.

jetty-io/src/test/java/org/eclipse/jetty/io/SelectorManagerTest.java
sbordet: My bad, the code is fine.

sbordet (approved):

lorban:
@janbartel nudge

janbartel (approved):

PR #6567, opened 2021-08-02
Fix #6565 Deploy Symlinked applications

### LINKED ISSUE #6565: Improve deployment of symlinked webapplications
**Target Jetty version(s)**
9, 10

**Enhancement Description**

After #6114 we can now deploy symlinked war files with the correct context path. eg.
```
   webapps
      + contextA.war -> /var/lib/myapp.war
      + contextB.war -> /var/lib/myapp.war
```
But the following does not work correctly:
```
   webapps
      + contextA -> /var/lib/myapp/
      + contextB -> /var/lib/myapp/
```

gregw:
Fix #6565 Deploy Symlinked applications by treating extracting context name (which becomes the default context path) from the base resource and then following aliases, so that base resource will not be an alias.   

Added warning in ContextHandler if the base resource is an alias that we may not support this in future releases.

### DIFF
--- a/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/WebAppProvider.java
+++ b/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/WebAppProvider.java
@@ -22,6 +22,7 @@ import org.eclipse.jetty.deploy.ConfigurationManager;
 import org.eclipse.jetty.deploy.util.FileID;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.handler.ContextHandler;
+import org.eclipse.jetty.util.StringUtil;
 import org.eclipse.jetty.util.URIUtil;
 import org.eclipse.jetty.util.annotation.ManagedAttribute;
 import org.eclipse.jetty.util.annotation.ManagedObject;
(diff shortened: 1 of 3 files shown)


jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/WebAppProvider.java
```diff
@@ -265,8 +266,18 @@ public ContextHandler createContextHandler(final App app) throws Exception
...
             throw new IllegalStateException("App resource does not exist " + resource);
 
-        String context = file.getName();
+        final String contextName = file.getName();
 
+        // Resource aliases (after getting name) to ensure baseResource is not an alias
+        if (resource.isAlias())
+        {
+            resource = Resource.newResource(resource.getAlias());
+            file = resource.getFile().toPath().toRealPath().toFile();
```
lachlan-roberts: Do we really need the `toRealPath()` here, shouldn't the `resource.getAlias()` already give you this?

gregw: Maybe not, but I wanted to protect against changes to the implemention of Resource and also to deal with things like symlinks to symlinks.   Better to be safe than sorry.

lachlan-roberts: Maybe then you should do something like:
```java
file = new File(resource.getAlias()).toPath().toRealPath().toFile();
resource = Resource.newResource(file);
```

so that `file` and `resource.getFile()` are the same.

gregw: very good point!

lachlan-roberts (approved):
LGTM

janbartel (approved):
I don't really like having different behaviour between distro and embedded (we go to a lot of lengths in other cases to ensure that they are the same), but I don't feel strongly enough to give it the thumbs down ;)

gregw:
@janbartel I agree, but I'd like to go with the warning first and see what feedback we get.  If nobody comments, then we can update embedded behaviour in a later PR.

Issue #6553, opened 2021-07-28
Review usage of Authentication.UNAUTHENTICATED in SecurityHandler


lachlan-roberts:
Review usage of `Authentication.UNAUTHENTICATED` in `SecurityHandler`.

The `JaspiAuthenticator` may return this in the case that it did not receive a `Principal` to use and returning `UNAUTHENTICATED` the request to continue through the `SecurityHandler` even if authentication was mandatory.

I think we may need to add a case in `SecurityHandler.handle` for:
```java
else if ((authentication == Authentication.UNAUTHENTICATED) && isAuthMandatory)
{
    response.sendError(HttpServletResponse.SC_FORBIDDEN, "unauthenticated");
    baseRequest.setHandled(true);
}
```

janbartel:
Looks good - PR?

PR #6568, opened 2021-08-02
Issue #6553 - give 401 response if UNAUTHENTICATED and auth is mandatory

lachlan-roberts:
## Issue #6553

If `Authenticator` returns `UNAUTHENTICATED` and auth is mandatory then we should return 403 response.


jetty-security/src/main/java/org/eclipse/jetty/security/SecurityHandler.java
```diff
@@ -572,6 +572,11 @@ else if (authentication instanceof Authentication.Deferred)
                             authenticator.secureResponse(request, response, isAuthMandatory, null);
                     }
                 }
+                else if ((authentication == Authentication.UNAUTHENTICATED) && isAuthMandatory)
+                {
+                    response.sendError(HttpServletResponse.SC_FORBIDDEN, "unauthenticated");
+                    baseRequest.setHandled(true);
+                }
```
sbordet, line 579: Test case?

sbordet (dismissed):

jetty-security/src/main/java/org/eclipse/jetty/security/SecurityHandler.java
```diff
@@ -572,6 +572,11 @@ else if (authentication instanceof Authentication.Deferred)
                             authenticator.secureResponse(request, response, isAuthMandatory, null);
                     }
                 }
+                else if ((authentication == Authentication.UNAUTHENTICATED) && isAuthMandatory)
```
gregw: Isn't this just:
```suggestion
                else if (isAuthMandatory)
```

ie. if auth is mandatory, then we really need either `ResponseSent` or `User` auth types.

jetty-security/src/main/java/org/eclipse/jetty/security/SecurityHandler.java
```diff
@@ -572,6 +572,11 @@ else if (authentication instanceof Authentication.Deferred)
                             authenticator.secureResponse(request, response, isAuthMandatory, null);
                     }
                 }
```
gregw, line 574: Should we also check for mandatory auth here, as I don't think we should defer mandatory auth.

gregw (changes requested):

jetty-security/src/main/java/org/eclipse/jetty/security/SecurityHandler.java
lachlan-roberts: I tried enforcing `isAuthMandatory` for `DeferredAuthentication` as well but it broke things, apparently some of the authenticators are using DeferredAuthentication to bypass the security constraints on purpose, for example FORM auth does this to allow access to the form page.

jetty-security/src/main/java/org/eclipse/jetty/security/SecurityHandler.java
lachlan-roberts: done

jetty-security/src/main/java/org/eclipse/jetty/security/SecurityHandler.java
gregw: Deferred auth is the exact opposite of mandatory. If a request has credentials but his a none mandatory URI, then we used deferred auth to avoid the cost of checking the credentials unless somebody calls a auth method

lachlan-roberts: I understand that. 

I am just saying that if we change the code to not allow deferred auth for isMandatory==true we will break how some of our existing Authenticators work.

gregw: I'm not saying that.  The spirit of this PR is to make the code here more robust in the face of unexpected returns from `validateRequest`, so  I'm saying that if `validateRequest` above somehow returned a `Deferred` auth then in this section we should check for `isMandatory==true` and if so resolve the authentication.

Actually we should probably do that above, so the results of that authentication can be fed into this if block.

gregw: OK - I now understand that our own authenticators are returning deferred auth to indicate a temporary OK for login pages and similar.   So I think this needs to be cleaned up... but not in this PR.

gregw (approved):

lachlan-roberts:
@gregw do you think this should be brought back to 9.4?

gregw:
@lachlan-roberts if we updated the JASPI auth in 9.4 to be usable, then yes this needs to be backported.   But if the recent jaspi changes were 10 only, then I say leave 9.4 as is.

lachlan-roberts:
@gregw this is not related to the JASPI updates. Looks like the existing JASPI implementation in 9.4 to can also return `UNAUTHENTICATED` in which case the request will be allowed past the security constraint.

gregw:
OK then I think backport is needed.

Issue #6554, opened 2021-07-28
Allow creation of DefaultIdentityService without realmName.


lachlan-roberts:
`SecurityHandler.doStart()` has the following logic:
```java
if (_identityService == null)
{
    if (_loginService != null)
        setIdentityService(_loginService.getIdentityService());

    if (_identityService == null)
        setIdentityService(findIdentityService());

    if (_identityService == null)
    {
        if (_realmName != null)
        {
            setIdentityService(new DefaultIdentityService());
            manage(_identityService);
        }
    }
    else
        unmanage(_identityService);
}
```

I am not sure we need the check for `if (_realmName != null)` to create the `DefaultIdentityService`.

This has caused previous issues with some people trying to configure `jetty-openid` from xml before. It would also make it easier to configure and use JASPI without this check.

janbartel:
Not checking the `_realmName` would make sense to me: we've already tried to find the `LoginService`, and if there was one, we would have asked it for it's `IdentityService`. As we didn't find one, it seems to be irrelevant to check the `_realmName`. In fact, it maybe should be an error for either of the 2 following conditions:
1.  we get this far with a `realmName` configured, but we didn't find any  `LoginServices`
2. we have a `realmName` but it didn't match any of of the `LoginServices`

@gregw  thoughts?

lachlan-roberts:
@janbartel This change will not currently work because the `DefaultAuthenticatorFactory` will return a `BasicAuthenticator` even if the AuthMethod is null. So if you create a `DefaultIdentiyService` it will always create a `BasicAuthenticator`.

I am not sure why we are doing this in `DefaultAuthenticationFactory`, doesn't seem like we should need to. There is a test towards the end of `SecurityHandler.doStart` which will throw if there is a `realmName` but no `Authenticator`.

 I will open a PR with these changes as well to see what it looks like.

PR #6569, opened 2021-08-02
Issue #6554 - create the DefaultIdentityService even if no realmName is provided

lachlan-roberts:
## Issue #6554 

- create the `DefaultIdentityService` even if no realmName is provided
- `DefaultAuthenticatorFactory` should not create `BasicAuthenticator` for null AuthMethod

### DIFF
--- a/jetty-security/src/main/java/org/eclipse/jetty/security/DefaultAuthenticatorFactory.java
+++ b/jetty-security/src/main/java/org/eclipse/jetty/security/DefaultAuthenticatorFactory.java
@@ -62,7 +62,7 @@ public class DefaultAuthenticatorFactory implements Authenticator.Factory
         String auth = configuration.getAuthMethod();
         Authenticator authenticator = null;
 
-        if (auth == null || Constraint.__BASIC_AUTH.equalsIgnoreCase(auth))
+        if (Constraint.__BASIC_AUTH.equalsIgnoreCase(auth))
             authenticator = new BasicAuthenticator();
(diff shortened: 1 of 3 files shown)


jetty-security/src/main/java/org/eclipse/jetty/security/DefaultAuthenticatorFactory.java
```diff
@@ -62,7 +62,7 @@ public Authenticator getAuthenticator(Server server, ServletContext context, Aut
         String auth = configuration.getAuthMethod();
         Authenticator authenticator = null;
 
-        if (auth == null || Constraint.__BASIC_AUTH.equalsIgnoreCase(auth))
+        if (Constraint.__BASIC_AUTH.equalsIgnoreCase(auth))
```
janbartel, line 65: Is there a test for this case?

janbartel (changes requested):

jetty-security/src/main/java/org/eclipse/jetty/security/DefaultAuthenticatorFactory.java
gregw: @lachlan-roberts I agree that we need at least some simple test that checks a DIS is created if there is no realm.  Ie not necessarily testing this specific line, but testing the PR in general.

gregw (changes requested):

gregw (approved):

janbartel (approved):

PR #6571, opened 2021-08-02
Issue #6558 - improved json array converter

### LINKED ISSUE #6558: Allow to configure return type in JSON array parsing
**Target Jetty version(s)**
9.4.x

**Enhancement Description**
Currently, `JSON.parse("[1]")` returns an `Object[]`, while `AsyncJSON` returns `List<Object>`.

This is a bit inconsistent, although there is no standard, so that many different implementation return different types (e.g. Jackson vs Gson vs Jetty), so applications need to be aware of these differences and "sniff" the type before casting.

Would be great if at least within Jetty we can make this consistent, so for example provide `AsyncJSON` of an option such as `parseArrayAsList=true|false` so that it is consistent with the `JSON` class.

And perhaps the same to the JSON class.

For example, in CometD using the `AsyncJSONTransport` uses `AsyncJSON` and therefore arrays are parsed as lists, while `WebSocketTransport` uses `JSON` and so arrays are parsed as arrays.
Trying to deserialize that JSON into an object whose class has a `setProperty(Object[] values)`, for example using `JSONPojoConvertor`, will work with one transport, but fail with the other, which is inconsistent.

sbordet:


jetty-util-ajax/src/test/java/org/eclipse/jetty/util/ajax/AsyncJSONTest.java
```diff
@@ -525,4 +529,47 @@ public void testEncodedCaching()
...
+        Object result = extractor.apply(async.complete());
+        // AsyncJSON historically defaults to list.
+        assertThat(result, Matchers.instanceOf(List.class));
+        // JSON historically defaults to array.
+        result = extractor.apply(sync.parse(new JSON.StringSource(json)));
+        assertNotNull(result);
+        assertTrue(result.getClass().isArray(), json + " -> " + result);
+
+        // Configure AsyncJSON to return arrays.
+        factory.setArrayConverter(List::toArray);
```
joakime, line 564: Ah, now that `setArrayConverter` method and signature makes sense to me.

joakime (approved):

olamy (approved):

PR #6572, opened 2021-08-02
Issue #5684 - windows test overhaul

joakime:
Overhaul the `@DisabledOnOs(WINDOWS)` tests

Reviewed all Windows specific test behaviors and adjusted to avoid `@Disabled...` if possible.


joakime:
Two tests are particularly troublesome and would need more work to identify what's going on.

`jetty-server/src/test/java/org/eclipse/jetty/server/ssl/SniSslConnectionFactoryTest.java `

This one has exposed a peculiar TLSv1.3 behavior difference on Windows vs Linux.
A SSL client on Windows sees Socket closures and exceptions related to it, while on Linux we see the SSLHandshakeException (which is what we expect).

If the test is forced to TLSv1.2 only, then it's reliable and consistent with Linux behavior.

`http2-client/src/test/java/org/eclipse/jetty/http2/client/ConcurrentStreamCreationTest.java`

This one is unreliable on slower systems, and there does not seem to be a way to reliably run it.
The stream count is never reached, and will result in excessive logging output eventually resulting in a total build/test timeout.

sbordet (approved):

PR #6579, opened 2021-08-03
Fix #6562 last written bytebuffer (#6563)

### LINKED ISSUE #6562: HttpOutput.write(ByteBuffer buffer)

gregw:
Cherry pick of fix for #6562 from jetty-9 to jetty-10 for the last written `ByteBuffer` calculation.
Also fixed an associated issue with unnecessary flush of an empty when last calculation already signalled last.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpOutput.java
@@ -847,10 +847,12 @@ public class HttpOutput extends ServletOutputStream implements Runnable
         // Blocking write
         try
         {
+            boolean complete = false;
             // flush any content from the aggregate
             if (BufferUtil.hasContent(_aggregate))
             {
-                channelWrite(_aggregate, last && len == 0);
+                complete = last && len == 0;
+                channelWrite(_aggregate, complete);
 
                 // should we fill aggregate again from the buffer?
                 if (len > 0 && !last && len <= _commitSize && len <= maximizeAggregateSpace())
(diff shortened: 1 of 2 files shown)


sbordet:
@gregw can you fix the test failures?

gregw:
@sbordet hmmm somehow the test is now failing in both 9 and this PR.   Note sure why it was working when it was merged in to 9???

gregw:
@sbordet ah a racy test. Fixed here and will backport to 9

lorban (approved):

Issue #6552, opened 2021-07-27
FileBufferedInterceptor.dispose not working due to locked file


joakime:
**Jetty version(s)**
9.4.x - HEAD

**Java version/vendor** `(use: java -version)`
Java 8u292 and 11.0.1

**OS type/version**
Windows 10

**Description**
The FileBufferedInterceptor.dispose() attempts to clean up files it has recently used.
However, this is not possible currently on windows, as it will throw a AccessDeniedException due to the fact that the file is currently locked.

We need to verify that we properly close the various file handles and resources for these files before we call dispose.

```
2021-07-27 16:03:45.670:WARN:oejsh.FileBufferedResponseHandler:qtp2144353034-5103: Could not delete file C:\Users\jenkins\workspace\nightlies\jetty-9.4.x-windows-nightly\jetty-server\target\tests\test-org.eclipse.jetty.server.handler.FileBufferedResponseHandlerTest\BufferedResponse15180402607277205671
java.nio.file.AccessDeniedException: C:\Users\jenkins\workspace\nightlies\jetty-9.4.x-windows-nightly\jetty-server\target\tests\test-org.eclipse.jetty.server.handler.FileBufferedResponseHandlerTest\BufferedResponse15180402607277205671
	at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:89)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
	at java.base/sun.nio.fs.WindowsFileSystemProvider.implDelete(WindowsFileSystemProvider.java:274)
	at java.base/sun.nio.fs.AbstractFileSystemProvider.delete(AbstractFileSystemProvider.java:105)
	at java.base/java.nio.file.Files.delete(Files.java:1141)
	at org.eclipse.jetty.server.handler.FileBufferedResponseHandler$FileBufferedInterceptor.dispose(FileBufferedResponseHandler.java:124)
```

**How to reproduce?**
Run the build / test on windows 10.

joakime:
After a conversation with @sbordet we are going to use the Java 9+ JVM feature `java.lang.ref.Cleaner` on Jetty 10+ to cleanup temp files in `FileBufferedHandler` properly for all OS's, including Linux.

lachlan-roberts:
[...]

PR #6581, opened 2021-08-03
Issue #5684 - Windows test overhaul

joakime:
**Port of PR #6572 for Jetty 10.0.x**

+ Migrate from @DisabledOnOs(WINDOWS) to assumptions on capabilities instead.
+ Fix other outstanding windows testing issues.
+ Cleanup FileBufferedResponseHandlerTest expectations on Windows.
+ PathWatcher scan interval is variable on windows
+ If unable to start testcase based on assumption,
  the stop shouldn't fail testcase
+ Increase various wait timeouts
+ Make tests less strict due to system speed issues
+ Disable Sni tests due to TLS behaviors differences in Windows
  + Windows TLSv1.3 seems to introduce this difference
  + If we restrict to TLSv1.2 this passes.
  + On Linux TLSv.13 on client side will always return a
  + javax.net.ssl.SSLHandshakeException in those test cases that expect it.
  + However, on Windows, Only the TLSv1.2 implementation will return a javax.net.ssl.SSLHandshakeException,
  + All other TLS versions on Windows will result in a
  + javax.net.ssl.SSLException: Software caused connection abort: recv failed
+ Disable ConcurrentStreamCreationTest
  + Not possible to create all of these streams.
+ Fixing DeploymentTempDirTest
  + Using unique workdir per testcase.
  + Don't expect to delete files / directories between tests
    (not supported on windows due to file locking anyway)


jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
```diff
@@ -100,21 +100,38 @@ public void resetBuffer()
             BufferedInterceptor.super.resetBuffer();
         }
 
+        private void closeFileOutput()
+        {
+            if (_fileOutputStream != null)
+            {
+                try
+                {
+                    _fileOutputStream.flush();
```
janbartel: Are you sure that this is a good thing to do? I see that `dispose()` is often called when some error condition has happened, so is it a good idea to try and flush content in that situation?

jetty-server/src/test/java/org/eclipse/jetty/server/ssl/SniSslConnectionFactoryTest.java
```diff
@@ -110,6 +110,18 @@ private void start(BiConsumer<SslContextFactory.Server, SecureRequestCustomizer>
...
+        {
+            // Restrict behavior in this testcase to how TLSv1.2 operates.
+            // This is because of behavior differences in TLS between Linux and Windows.
+            // On Linux TLS on client side will always return a javax.net.ssl.SSLHandshakeException
+            // in those test cases that expect it.
+            // However, on Windows, there are differences between using OpenJDK 8 and OpenJDK 11.
+            // Only the TLSv1.2 implementation will return a javax.net.ssl.SSLHandshakeException,
+            // all other TLS versions will result in a
+            // javax.net.ssl.SSLException: Software caused connection abort: recv failed
+            // sslContextFactory.setIncludeProtocols("TLSv1.2");
```
janbartel: There's no code inside this if?

jetty-servlets/pom.xml
```diff
@@ -39,7 +39,11 @@
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-webapp</artifactId>
       <version>${project.version}</version>
-      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-server</artifactId>
+      <version>${project.version}</version>
```
janbartel: This shouldn't be necessary because `jetty-server` is a transitive dependency of `jetty-webapp`

jetty-servlets/src/test/java/org/eclipse/jetty/servlets/WelcomeFilterTest.java
```diff
@@ -80,12 +80,14 @@ public void prepareServer() throws Exception
...
 
         WebAppContext context = new WebAppContext(server, directoryPath.toString(), "/");
+        context.setDefaultsDescriptor(MavenTestingUtils.getTestResourceFile("simple-default.xml").toString());
         server.setHandler(context);
         String concatPath = "/*";
 
         FilterHolder filterHolder = new FilterHolder(new WelcomeFilter());
         filterHolder.setInitParameter("welcome", "welcome.html");
         context.addFilter(filterHolder, concatPath, EnumSet.of(DispatcherType.REQUEST));
+        server.setDumpAfterStart(true);
```
janbartel: Probably don't want this amount of output during a test.

jetty-servlets/src/test/java/org/eclipse/jetty/servlets/WelcomeFilterTest.java
```diff
@@ -80,12 +80,14 @@ public void prepareServer() throws Exception
         }
 
         WebAppContext context = new WebAppContext(server, directoryPath.toString(), "/");
+        context.setDefaultsDescriptor(MavenTestingUtils.getTestResourceFile("simple-default.xml").toString());
```
janbartel: What's the difference between the simple-default.xml and the normal webdefault.xml?

jetty-util/src/test/java/org/eclipse/jetty/util/resource/FileSystemResourceTest.java
```diff
@@ -658,7 +656,7 @@ public void testSymlink(Class<PathResource> resourceClass) throws Exception
         {
             // if unable to create symlink, no point testing the rest
             // this is the path that Microsoft Windows takes.
-            assumeTrue(true, "Not supported");
+            assumeTrue(false, "Not supported");
```
janbartel: Won't  the `assumeTrue` throw an Exception that makes the test fail, rather than skipping the test? Same for the other uses of `assumeTrue(false)` in this test.

janbartel (dismissed):

jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
joakime: These changes in this FileBufferedResponseHandler need to be rolled back, they were from early attempts at resolving the locked file issue (which is also filed as Issue #6552 )

jetty-server/src/test/java/org/eclipse/jetty/server/ssl/SniSslConnectionFactoryTest.java
joakime: This is from the 9.4.x commits - 60bfc75ce4e4289e924c42cea4f555ca215b1e54

I had a review of the TLS differences in windows with @sbordet and it turns out while the issue appears to be solved with just using TLSv1.2, the problem is more deeply rooted.
I need to file a new issue to review this discovery.

jetty-servlets/pom.xml
joakime: Yet the `WelcomeFilterTest` fails to execute under windows without it.

I'll dig into this more.

jetty-servlets/src/test/java/org/eclipse/jetty/servlets/WelcomeFilterTest.java
joakime: I'll change the name of that new file and add a comment here.
It's only difference is that it disabled complex file serving and serves using simple I/O techniques.
Namely it disables memory mapped features in DefaultServlet and ResourceService.

jetty-util/src/test/java/org/eclipse/jetty/util/resource/FileSystemResourceTest.java
joakime: This is an "assume" and it was written incorrectly, as `assumeTrue(true, "foo")` never does anything. It's the equiv of `assertTrue(true, "bar")`.
We were protected from this mistake because of the `@DisabledOnOs(WINDOWS)` that used to be here.

It might make more sense if you see the "long form" version of this same assumption ...

``` java
boolean symlinkSupported;
try
{
  Files.createSymbolicLink(bar, foo);
  symlinkSupported = true;
}
catch (UnsupportedOperationException | FileSystemException e)
{
  symlinkSupported = false;
}
assumeTrue(symlinkSupported); // break out of test if symlink isn't supported
doTestThatRequiresSymlinkSupport();
```

jetty-server/src/test/java/org/eclipse/jetty/server/ssl/SniSslConnectionFactoryTest.java
joakime: I opened #6609 to resolve/solve this in a later PR.

jetty-servlets/pom.xml
joakime: If I remove this, `WelcomeFilterTest` fails to run with a "java.lang.NoClassDefFoundError: org/eclipse/jetty/server/Connector"

The jetty-server is just not present in the test classpath when surefire executes. *boggle*

jetty-servlets/src/test/java/org/eclipse/jetty/servlets/WelcomeFilterTest.java
joakime: I'm growing paranoid.  Even this dump shows no `jetty-server` artifact on the server classloader or the webapp classloader?
How did I get this far to call **server**.setDumpAfterStart(true) and get a result telling me that server isn't there??

jetty-servlets/pom.xml
joakime: I rolled back this change.

jetty-servlets/src/test/java/org/eclipse/jetty/servlets/WelcomeFilterTest.java
joakime: I removed this addition.

jetty-servlets/src/test/java/org/eclipse/jetty/servlets/WelcomeFilterTest.java
joakime: The name is fixed to be more clear, and there is now a comment explaining the why it exists.

joakime:
@sbordet this PR is ready for review.

tests/test-distribution/src/test/java/org/eclipse/jetty/tests/distribution/session/FileSessionDistributionTests.java
```diff
@@ -16,9 +16,7 @@
 import java.util.Collections;
 import java.util.List;
 
-/**
- *
- */
+@Testcontainers(disabledWithoutDocker = false)
```
joakime, line 24: Oddly, this test doesn't fail on Windows, and it doesn't have docker.

joakime:
Docker on windows is quite complicated.  It seems to want to run WSL and only serve Linux containers?  Not sure if our tests can handle that kind of mix (where the test is windows, and the server/backend is linux)

tests/test-distribution/src/test/java/org/eclipse/jetty/tests/distribution/session/FileSessionDistributionTests.java
olamy: yup because this test is not using docker :)  it's only session storage in files

olamy:
not sure we want to try install/setup docker on the windows box...

joakime:
> not sure we want to try install/setup docker on the windows box...

I agree, it's a mess, and seems to also only want to run if you have an active login, seems like it's tied to an actual UI element somehow.
Can't get it to run headless, but is it because of this perceived UI requirement, or some kind of WSL requirement?

jetty-http2/http2-client/src/test/java/org/eclipse/jetty/http2/client/AbstractTest.java
```diff
@@ -45,6 +47,28 @@
...
+            }
+            catch (IOException e)
+            {
+                // fall back to localhost if we cannot
+                serverHost = "localhost";
+            }
+        }
+
+        return serverHost;
+    }
```
sbordet: Please revert and just use "localhost".

jetty-http2/http2-client/src/test/java/org/eclipse/jetty/http2/client/SmallThreadPoolLoadTest.java
```diff
@@ -138,8 +137,7 @@ private boolean test(Session session, CountDownLatch latch) throws Exception
         boolean download = random.nextBoolean();
         HttpMethod method = download ? HttpMethod.GET : HttpMethod.POST;
 
-        int maxContentLength = 128 * 1024;
-        int contentLength = random.nextInt(maxContentLength) + 1;
+        int contentLength = (random.nextInt(5) * random.nextInt(1024) * random.nextInt(1024)) + 1;
```
sbordet: Please restore back to original.

jetty-http2/http2-client/src/test/java/org/eclipse/jetty/http2/client/SmallThreadPoolLoadTest.java
```diff
@@ -202,7 +199,11 @@ protected void service(HttpServletRequest request, HttpServletResponse response)
                 {
                     int contentLength = request.getIntHeader("X-Download");
                     if (contentLength > 0)
-                        response.getOutputStream().write(new byte[contentLength]);
+                    {
+                        byte[] buf = new byte[contentLength];
+                        ThreadLocalRandom.current().nextBytes(buf);
+                        response.getOutputStream().write(buf);
+                    }
```
sbordet: Please restore back to original.

jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -297,6 +298,11 @@ public boolean doGet(HttpServletRequest request, HttpServletResponse response)
             // Send the data
             releaseContent = sendData(request, response, included, content, reqRanges);
         }
+        catch (InvalidPathException e)
+        {
+            notFound(request, response);
+            return response.isCommitted();
```
sbordet, line 309: Please add at least a `LOG.debug()` to log what's the invalid path.
Also, at line 249 may be worth to throws `InvalidPathException` since the code lines are the same.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
joakime: These changes have been rolled back.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
```diff
@@ -114,7 +114,8 @@ protected void dispose()
                 }
                 catch (Throwable t)
                 {
-                    LOG.warn("Could not delete file {}", _filePath, t);
+                    LOG.debug("Could not immediately delete file (delaying to jvm exit) {}", _filePath, t);
```
sbordet: Guard with "if debug enabled" pattern.

jetty-server/src/test/java/org/eclipse/jetty/server/GracefulStopTest.java
```diff
@@ -414,7 +414,7 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
                     }
                 }
 
-                response.getWriter().printf("read %d/%d%n", c, contentLength);
+                response.getWriter().printf("read %d/%d\n", c, contentLength);
```
sbordet: Stick with `%n` and don't test for line endings, or not even use any line endings.

jetty-server/src/test/java/org/eclipse/jetty/server/ssl/SniSslConnectionFactoryTest.java
```diff
@@ -235,8 +247,8 @@ public void testBadSNIConnect() throws Exception
         assertThat(response, Matchers.containsString("Invalid SNI"));
     }
 
+    @DisabledOnOs(value = OS.WINDOWS, disabledReason = "TLSv1.3 behavior differences between Linux and Windows")
```
sbordet: Maybe add the issue # to `disabledReason` and delete the comment at line 113?

jetty-servlets/src/test/java/org/eclipse/jetty/servlets/AbstractDoSFilterTest.java
```diff
@@ -289,7 +290,8 @@ public void testMultipleSessionTracking() throws Exception
         String responses = doRequests(request1 + request2 + request1 + request2 + request1, 2, 1100, 1100, last);
 
         assertEquals(11, count(responses, "HTTP/1.1 200 OK"));
-        assertEquals(0, count(responses, "DoSFilter: delayed"));
+        // This test is system speed dependent, so allow some (20%-ish) requests to be delayed, but not more.
+        assertThat("delayed count", count(responses, "DoSFilter: delayed"), lessThan(2));
```
sbordet, line 294: Can you use awaitility to wait for the count to be 0?

jetty-servlets/src/test/java/org/eclipse/jetty/servlets/WelcomeFilterTest.java
```diff
@@ -80,6 +80,8 @@ public void prepareServer() throws Exception
         }
 
         WebAppContext context = new WebAppContext(server, directoryPath.toString(), "/");
+        // Turn off memory-mapped behavior in DefaultServlet for Windows testing reasons.
+        context.setDefaultsDescriptor(MavenTestingUtils.getTestResourceFile("no-memory-mapped-default-servlet.xml").toString());
```
sbordet: Is it possible to use the override descriptor?
Or maybe something like `context.getDefaultsDescriptor().setInitParam(...)` to change just that value?

jetty-util/src/test/java/org/eclipse/jetty/util/PathWatcherTest.java
```diff
@@ -378,7 +378,7 @@ public void testSequence() throws Exception
             capture.finishedLatch.await(LONG_TIME, TimeUnit.MILLISECONDS);
             end = System.nanoTime();
             capture.assertEvents(expected);
-            assertThat(end - start, greaterThan(TimeUnit.MILLISECONDS.toNanos(2 * QUIET_TIME)));
+            assertThat(end - start, greaterThan(TimeUnit.MILLISECONDS.toNanos(QUIET_TIME)));
```
sbordet: The `assert` was probably failing due to timer resolution (e.g. the values being equal and not strictly greater), so the test should be reviewed for other occurrences of the same pattern.

jetty-util/src/test/java/org/eclipse/jetty/util/ScannerTest.java
```diff
@@ -33,35 +34,32 @@
...
 
 public class ScannerTest
 {
-    static File _directory;
+    static Path _directory;
     static Scanner _scanner;
     static BlockingQueue<Event> _queue = new LinkedBlockingQueue<>();
     static BlockingQueue<Set<String>> _bulk = new LinkedBlockingQueue<>();
 
     @BeforeAll
```
sbordet: Don't like `@BeforeAll` with all the statics, especially for files...
Maybe use a per-test-method `Path` and `@BeforeEach`?

jetty-util/src/test/java/org/eclipse/jetty/util/resource/FileSystemResourceTest.java
sbordet: Maybe change comment to `Symlink not supported`?

jetty-util/src/test/java/org/eclipse/jetty/util/resource/FileSystemResourceTest.java
```diff
@@ -930,14 +927,13 @@ public void testNTFSFileEncodedDataStreamAlias(Class<PathResource> resourceClass
             catch (InvalidPathException e)
             {
                 // NTFS filesystem streams are unsupported on some platforms.
-                assumeTrue(true, "Not supported on this OS");
+                assumeTrue(false, "NFTS Dats streams not supported");
```
sbordet: Fix typo. Maybe update the comment on the line above?

jetty-util/src/test/java/org/eclipse/jetty/util/resource/ResourceTest.java
```diff
@@ -287,15 +288,23 @@ public void testResourceContent(Scenario data)
...
+        {
+            String globReference = testDir.toAbsolutePath() + File.separator + '*';
+            Resource globResource = Resource.newResource(globReference);
+            assertNotNull(globResource, "Should have produced a Resource");
+        }
+        catch (InvalidPathException e)
+        {
+            // if unable to reference the glob file, no point testing the rest.
+            // this is the path that Microsoft Windows takes.
+            assumeTrue(true, "Not supported on this OS");
```
sbordet: First parameter should be `false`?

jetty-webapp/src/test/java/org/eclipse/jetty/webapp/WebAppDefaultServletTest.java
```diff
@@ -48,7 +52,7 @@ public void prepareServer() throws Exception
         connector.getConnectionFactory(HttpConnectionFactory.class).getHttpConfiguration().setUriCompliance(UriCompliance.RFC3986);
         server.addConnector(connector);
 
-        Path directoryPath = MavenTestingUtils.getTargetTestingDir().toPath();
+        Path directoryPath = workDir.getEmptyPathDir();
         IO.delete(directoryPath.toFile());
         Files.createDirectories(directoryPath);
```
sbordet: These 2 lines above are not necessary?

jetty-websocket/websocket-jetty-tests/src/test/java/org/eclipse/jetty/websocket/tests/WebSocketOverHTTP2Test.java
```diff
@@ -246,8 +246,17 @@ public void testWebSocketConnectPortDoesNotExist() throws Exception
             wsClient.connect(wsEndPoint, uri).get(5, TimeUnit.SECONDS));
 
         Throwable cause = failure.getCause();
-        assertThat(cause, instanceOf(ConnectException.class));
-        assertThat(cause.getMessage(), containsStringIgnoringCase("Connection refused"));
+        cause.printStackTrace();
+        if (cause instanceof ClosedChannelException)
```
sbordet: This should be investigated. Connecting to a non-listening port should always report `ConnectException`.
Is it the JVM or our code creating the `ClosedChannelException`?

tests/test-integration/src/test/java/org/eclipse/jetty/test/rfcs/RFC2616BaseTest.java
```diff
@@ -95,8 +95,8 @@ public static void setUpServer(XmlBasedJettyServer testableserver, Class<?> test
 
         server = testableserver;
         server.load();
+        // server.getServer().setDumpAfterStart(true);
```
sbordet: Just delete the line.

tests/test-integration/src/test/resources/NIOHttps.xml
```diff
@@ -1,36 +1,34 @@
...
+            </Item>
+            <Item>
+              <New class="org.eclipse.jetty.server.HttpConnectionFactory">
+                <Arg name="config"><Ref refid="sslHttpConfig"/></Arg>
+              </New>
+            </Item>
+          </Array>
+        </Arg>
+        <Set name="host" property="jetty.http.host" />
+        <Set name="port" property="jetty.http.port" />
```
sbordet: Should not these property be `jetty.ssl.*`?

sbordet (changes requested):

jetty-websocket/websocket-jetty-tests/src/test/java/org/eclipse/jetty/websocket/tests/WebSocketOverHTTP2Test.java
joakime: Opened Issue #6660 to cover this case.
I'll revert these changes in this branch in the meantime.

jetty-servlets/src/test/java/org/eclipse/jetty/servlets/AbstractDoSFilterTest.java
joakime: this would never reach 0 on a slow machine, as the existence of `DosFilter: delayed` means a response actually got delayed, it's too late to await.

jetty-http2/http2-client/src/test/java/org/eclipse/jetty/http2/client/SmallThreadPoolLoadTest.java
joakime: Restored

jetty-http2/http2-client/src/test/java/org/eclipse/jetty/http2/client/SmallThreadPoolLoadTest.java
joakime: Restored

jetty-http2/http2-client/src/test/java/org/eclipse/jetty/http2/client/AbstractTest.java
joakime: Reverted

jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
joakime: Reworked.

jetty-server/src/test/java/org/eclipse/jetty/server/GracefulStopTest.java
joakime: Reworked

jetty-server/src/test/java/org/eclipse/jetty/server/ssl/SniSslConnectionFactoryTest.java
joakime: Reason updated everywhere

jetty-util/src/test/java/org/eclipse/jetty/util/ScannerTest.java
joakime: Reworked

jetty-util/src/test/java/org/eclipse/jetty/util/resource/FileSystemResourceTest.java
joakime: All assume methods have updated and appropriate descriptions.

jetty-util/src/test/java/org/eclipse/jetty/util/resource/ResourceTest.java
joakime: Fixed

jetty-webapp/src/test/java/org/eclipse/jetty/webapp/WebAppDefaultServletTest.java
joakime: Removed.

tests/test-integration/src/test/resources/NIOHttps.xml
joakime: Just removed them. not necessary.

jetty-servlets/src/test/java/org/eclipse/jetty/servlets/WelcomeFilterTest.java
joakime: I got rid of this defaults descriptor, found a simpler way.

sbordet (approved):

sbordet (dismissed):

sbordet (approved):

PR #6583, opened 2021-08-03
Fixes #6558 - Allow configuring return type in JSON array parsing.

### LINKED ISSUE #6558: Allow to configure return type in JSON array parsing

sbordet:
Introduced `arrayConverter` in both JSON and AsyncJSON.Factory.

(cherry picked from commit 342396c7ee33a4fa51e3d6b1facc5337942dc354)

### DIFF
--- a/jetty-util-ajax/src/main/java/org/eclipse/jetty/util/ajax/AsyncJSON.java
+++ b/jetty-util-ajax/src/main/java/org/eclipse/jetty/util/ajax/AsyncJSON.java
@@ -19,7 +19,9 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
 
 import org.eclipse.jetty.util.BufferUtil;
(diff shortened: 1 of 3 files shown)


sbordet:
Forward port of #6571 to Jetty 10.

joakime (approved):

PR #6584, opened 2021-08-04
Fixed racy test

gregw:
Fixed racy test from #6563 

### DIFF
--- a/jetty-server/src/test/java/org/eclipse/jetty/server/HttpOutputTest.java
+++ b/jetty-server/src/test/java/org/eclipse/jetty/server/HttpOutputTest.java
@@ -767,7 +767,7 @@ public class HttpOutputTest
     @Test
     public void testEmptyArray() throws Exception
     {
-        AtomicBoolean committed = new AtomicBoolean();
+        FuturePromise<Boolean> committed = new FuturePromise<>();
         AbstractHandler handler = new AbstractHandler()
         {
             @Override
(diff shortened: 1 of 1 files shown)


olamy (approved):

jetty-server/src/test/java/org/eclipse/jetty/server/HttpOutputTest.java
```diff
@@ -767,16 +767,23 @@ public void setNext(Interceptor interceptor)
     @Test
     public void testEmptyArray() throws Exception
     {
-        AtomicBoolean committed = new AtomicBoolean();
+        FuturePromise<Boolean> committed = new FuturePromise<>();
```
lachlan-roberts, line 770: If you're using `FuturePromise` now, maybe you should add a timeout for the `get()` call.

gregw: oops I missed this comment.  I'm planning for success and the test will pass!        If we break something, then a test that hangs forever will get just as much attention than one that times out.

lachlan-roberts: Yeah but its not ideal, sometimes its hard to find where it timed out and why. Much better to just fail after a few seconds.

gregw: I've fixed this in the 10 PR which has not yet been merged.

PR #6585, opened 2021-08-04
Issue #6566 - add executor to WebSocketComponents & use for Dispatched Messages

### LINKED ISSUE #6566: High CPU use due to new thread created on every websocket message in DispatchedMessageSink
In websocket-jetty-server 11.0.6, the class "org.eclipse.jetty.websocket.core.internal.messages.DispatchedMessageSink" accept method starts a new Thread (NOT using a thread pool).  In my app this is causing a new thread creation for every web socket message (not connection).  As a result, I am seeing about 4x total app CPU usage.  As a workaround, I updated the code to use a Thread Pool and now see low CPU usage again.

lachlan-roberts:
## Issue #6566 

Use an executor to dispatch new threads for `InputStream` or `Reader` type messages.

This doesn't fix #6566 as that issue was also reproduced using the `onMessage(ByteBuffer)` style methods. But this should still offer some improvement for dispatched messages.

### DIFF
--- a/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/WebSocketComponents.java
+++ b/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/WebSocketComponents.java
@@ -31,12 +31,12 @@ import org.eclipse.jetty.util.thread.QueuedThreadPool;
  */
 public class WebSocketComponents extends ContainerLifeCycle
 {
-    private final DecoratedObjectFactory objectFactory;
-    private final WebSocketExtensionRegistry extensionRegistry;
-    private final Executor executor;
-    private final ByteBufferPool bufferPool;
-    private final InflaterPool inflaterPool;
-    private final DeflaterPool deflaterPool;
+    private final DecoratedObjectFactory _objectFactory;
+    private final WebSocketExtensionRegistry _extensionRegistry;
+    private final Executor _executor;
+    private final ByteBufferPool _bufferPool;
+    private final InflaterPool _inflaterPool;
+    private final DeflaterPool _deflaterPool;
 
     public WebSocketComponents()
     {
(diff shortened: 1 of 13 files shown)


jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/WebSocketComponents.java
```diff
@@ -43,41 +46,54 @@ public WebSocketComponents()
...
+
+    public WebSocketComponents(WebSocketExtensionRegistry extensionRegistry, DecoratedObjectFactory objectFactory,
+                               ByteBufferPool bufferPool, InflaterPool inflaterPool, DeflaterPool deflaterPool, Executor executor)
+    {
+        _extensionRegistry = (extensionRegistry == null) ? new WebSocketExtensionRegistry() : extensionRegistry;
+        _objectFactory = (objectFactory == null) ? new DecoratedObjectFactory() : objectFactory;
+        _bufferPool = (bufferPool == null) ? new MappedByteBufferPool() : bufferPool;
+        _inflaterPool = (inflaterPool == null) ? new InflaterPool(CompressionPool.DEFAULT_CAPACITY, true) : inflaterPool;
+        _deflaterPool = (deflaterPool == null) ? new DeflaterPool(CompressionPool.DEFAULT_CAPACITY, Deflater.DEFAULT_COMPRESSION, true) : deflaterPool;
+        _executor = (executor == null) ? new QueuedThreadPool() : executor;
```
gregw: rather than create a whole new pool here, I think it would be better to use an `Executor` that just starts a new thread for each call to `execute` - so if you don't pass in a thread pool then you don't get a thread pool.

I know we do new QTP in other places... but I'm questioning if that is still the right thing to do.

jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/WebSocketServerComponents.java
```diff
@@ -92,6 +97,8 @@ public static WebSocketComponents ensureWebSocketComponents(Server server, Servl
             serverComponents.unmanage(deflaterPool);
         if (server.contains(bufferPool))
             serverComponents.unmanage(bufferPool);
+        if (executor != null)
+            serverComponents.unmanage(executor);
```
gregw, line 101: It's a little funny to create the WSSC with executor as auto and then externally force to unmanaged.   Kind of feels like the WSSC should be making these decisions for itself internally.    But if this is the only place the WSSCs are instantiated, then I guess it is OK?

gregw (changes requested):

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/WebSocketComponents.java
lachlan-roberts: If you start a `WebSocketClient` by default you will use this `executor` for the `HttpClient`. I don't think we'd really want this strategy of starting a new thread every time for the client.

gregw: OK

jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/WebSocketServerComponents.java
lachlan-roberts: I'm following the same pattern as t he other components here.

Although I think it may not be necessary as I think the threadPool should always be started by the time we reach here, so it will not be managed by the WebSocketComponents. But I don't think there is any harm to have this line just to be sure.

gregw (dismissed):

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/WebSocketComponents.java
```diff
@@ -43,41 +46,64 @@ public WebSocketComponents()
...
+        if (executor == null)
+        {
+            QueuedThreadPool threadPool = new QueuedThreadPool();
+            threadPool.setName("WebSocket@" + hashCode());
+            _executor = threadPool;
+        }
+        else
+        {
+            _executor = executor;
+        }
```
gregw, line 70: If you are breaking out like this, then you may as well do:

```suggestion
        if (executor == null)
        {
            QueuedThreadPool threadPool = new QueuedThreadPool();
            threadPool.setName("WebSocket@" + hashCode());
            _executor = threadPool;
            addBean(_executor, true);
        }
        else
        {
            _executor = executor;
            addBean(_executor, false);
        }
```

gregw: Hmmm but then all the other beans are managed differently.....  I still don't like how the management decision is made external to this class... but I'll live with it in this PR.   But do think about fixing it sometime.

gregw (approved):

PR #6588, opened 2021-08-05
Issue #6552 - Windows Cleaner+FileDisposer for FileBufferedResponseHandler

joakime:
Proposal of using `java.lang.ref.Cleaner` with `FileBufferedResponseHandler`

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
@@ -189,7 +189,7 @@ public class FileBufferedResponseHandler extends BufferedResponseHandler
             // Create an iterating callback to do the writing
             IteratingCallback icb = new IteratingCallback()
             {
-                private final long fileLength = _filePath.toFile().length();
+                private final long _fileLength = _filePath.toFile().length();
                 private long _pos = 0;
                 private boolean _last = false;
 
@@ -199,8 +199,8 @@ public class FileBufferedResponseHandler extends BufferedResponseHandler
                     if (_last)
                         return Action.SUCCEEDED;
 
-                    long len = Math.min(MAX_MAPPED_BUFFER_SIZE, fileLength - _pos);
-                    _last = (_pos + len == fileLength);
+                    long len = Math.min(MAX_MAPPED_BUFFER_SIZE, _fileLength - _pos);
+                    _last = (_pos + len == _fileLength);
                     ByteBuffer buffer = BufferUtil.toMappedBuffer(_filePath, _pos, len);
                     getNextInterceptor().write(buffer, _last, this);
                     _pos += len;
@@ -224,8 +224,8 @@ public class FileBufferedResponseHandler extends BufferedResponseHandler
                 @Override
                 public String toString()
                 {
-                    return String.format("FileBufferedInterceptor@%X[fileLength=%d,pos=%d,last=%b]",
-                        this.hashCode(), fileLength, _pos, _last
+                    return String.format("FileBufferedInterceptor@%x[fileLength=%d,pos=%d,last=%b]",
+                        hashCode(), _fileLength, _pos, _last
                     );
                 }
             };
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
@@ -220,14 +220,6 @@ public class FileBufferedResponseHandler extends BufferedResponseHandler
                     dispose();
                     callback.failed(cause);
                 }
-
-                @Override
-                public String toString()
-                {
-                    return String.format("FileBufferedInterceptor@%x[fileLength=%d,pos=%d,last=%b]",
-                        hashCode(), _fileLength, _pos, _last
-                    );
-                }
             };
             icb.iterate();
         }
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
@@ -28,6 +28,7 @@ import org.eclipse.jetty.util.BufferUtil;
 import org.eclipse.jetty.util.Callback;
 import org.eclipse.jetty.util.IO;
 import org.eclipse.jetty.util.IteratingCallback;
+import org.eclipse.jetty.util.PhantomCleaner;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -106,18 +107,8 @@ public class FileBufferedResponseHandler extends BufferedResponseHandler
             _fileOutputStream = null;
             _aggregating = null;
 
-            if (_filePath != null)
-            {
-                try
-                {
-                    Files.delete(_filePath);
-                }
-                catch (Throwable t)
-                {
-                    LOG.warn("Could not delete file {}", _filePath, t);
-                }
-                _filePath = null;
-            }
+
+            _filePath = null;
         }
 
         @Override
@@ -226,6 +217,9 @@ public class FileBufferedResponseHandler extends BufferedResponseHandler
                     callback.failed(cause);
                 }
             };
+            // once icb is GC'd, cleanup the File.
+            if (_filePath != null)
+                PhantomCleaner.register(icb, () -> new PhantomCleaner.FileDispose(_filePath));
             icb.iterate();
         }
     }
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
@@ -216,10 +216,21 @@ public class FileBufferedResponseHandler extends BufferedResponseHandler
                     dispose();
                     callback.failed(cause);
                 }
+
+                @Override
+                public String toString()
+                {
+                    final StringBuilder sb = new StringBuilder("FileBufferedInterceptor[");
+                    sb.append("fileLength=").append(fileLength);
+                    sb.append(", pos=").append(_pos);
+                    sb.append(", last=").append(_last);
+                    sb.append(']');
+                    return sb.toString();
+                }
             };
             // once icb is GC'd, cleanup the File.
             if (_filePath != null)
-                PhantomCleaner.register(icb, () -> new PhantomCleaner.FileDispose(_filePath));
+                PhantomCleaner.register(icb, new PhantomCleaner.FileDispose(_filePath));
             icb.iterate();
         }
     }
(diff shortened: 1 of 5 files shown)


jetty-util/src/main/java/org/eclipse/jetty/util/PhantomCleaner.java
```diff
@@ -0,0 +1,71 @@
...
+import java.io.IOException;
+import java.lang.ref.Cleaner;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class PhantomCleaner
```
gregw: I know it the JVM class uses "Phantom" as a concept, but I don't like the name here.   `FileGcCleaner` would be better.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
```diff
@@ -225,7 +216,21 @@ protected void onCompleteFailure(Throwable cause)
...
+                    sb.append("fileLength=").append(fileLength);
+                    sb.append(", pos=").append(_pos);
+                    sb.append(", last=").append(_last);
+                    sb.append(']');
+                    return sb.toString();
+                }
             };
+            // once icb is GC'd, cleanup the File.
+            if (_filePath != null)
+                PhantomCleaner.register(icb, new PhantomCleaner.FileDispose(_filePath));
```
gregw: Do we only want to do delayed deletion on windows? or do we want to use the same mechanism everywhere?

Perhaps we could try to delete here and if that fails then register the cleaner?

gregw: Also is the icb really the right object?  Surely it should be the buffer?   icb doesn't even have a reference to the buffer.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
```diff
@@ -225,7 +216,21 @@ protected void onCompleteFailure(Throwable cause)
...
+
+                @Override
+                public String toString()
+                {
+                    final StringBuilder sb = new StringBuilder("FileBufferedInterceptor[");
+                    sb.append("fileLength=").append(fileLength);
+                    sb.append(", pos=").append(_pos);
+                    sb.append(", last=").append(_last);
+                    sb.append(']');
+                    return sb.toString();
```
sbordet: Use `String.format()` and `getClass().getSimpleName()`.

jetty-util/src/main/java/org/eclipse/jetty/util/PhantomCleaner.java
sbordet: This is a generic mechanism, not only for files, so `JVMCleaner` or `ReferenceCleaner` or just `Cleaner`.

jetty-util/src/main/java/org/eclipse/jetty/util/PhantomCleaner.java
```diff
@@ -0,0 +1,71 @@
...
+            }
+        }
+
+        @Override
+        public String toString()
+        {
+            final StringBuilder sb = new StringBuilder("FileDispose[");
+            sb.append("filename='").append(filename).append('\'');
+            sb.append(']');
+            return sb.toString();
```
sbordet: Please use `String.format()` and `getClass().getSimpleName()`.

jetty-util/src/main/java/org/eclipse/jetty/util/PhantomCleaner.java
```diff
@@ -0,0 +1,71 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(PhantomCleaner.class);
+    private static final Cleaner CLEANER = Cleaner.create();
+
+    public static void register(Object obj, Runnable action)
+    {
+        LOG.info("register({}, {})", obj, action);
+        CLEANER.register(obj, action);
+    }
+
+    public static class FileDispose implements Runnable
```
sbordet: This is really a file *delete*, "dispose" is too generic. I would rename to `DeleteFile`.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
sbordet: @gregw it cannot be the buffer because we may iterate and have multiple buffers. Using the ICB guarantees that when the ICB is done, then all the buffers are also done and we register with the cleaner only once.
It is also critical that we register a reference to an object that is not strongly referenced, and the ICB reference satisfies this condition.

This mechanism will delay the file deletion in all cases, not just Windows.
Trying to delete and if fails use the cleaner will introduce a race as the cleaner may run at any moment... but `Files.deleteIfExists()` should take care of that.

jetty-util/src/main/java/org/eclipse/jetty/util/PhantomCleaner.java
```diff
@@ -0,0 +1,71 @@
...
+import org.slf4j.LoggerFactory;
+
+public final class PhantomCleaner
+{
+    private static final Logger LOG = LoggerFactory.getLogger(PhantomCleaner.class);
+    private static final Cleaner CLEANER = Cleaner.create();
+
+    public static void register(Object obj, Runnable action)
+    {
+        LOG.info("register({}, {})", obj, action);
```
sbordet: `Log.debug()`, not `info()`.

jetty-util/src/main/java/org/eclipse/jetty/util/PhantomCleaner.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public static class FileDispose implements Runnable
+    {
+        private static final Logger LOG = LoggerFactory.getLogger(FileDispose.class);
+        private final String filename;
+
+        public FileDispose(Path file)
+        {
+            // storing this reference as a String to allow GC to work.
+            this.filename = file.toAbsolutePath().toString();
```
sbordet: There is no need to store the `Path` as a string, the comment is wrong as this is just the action.

jetty-util/src/main/java/org/eclipse/jetty/util/PhantomCleaner.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        @Override
+        public void run()
+        {
+            Path file = Paths.get(filename);
+            try
+            {
+                LOG.info("Cleaning {}", file);
```
sbordet: `LOG.debug()`, not `info()`.

sbordet (changes requested):

jetty-server/src/test/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandlerTest.java
```diff
@@ -119,6 +125,14 @@ public void after() throws Exception
         _server.stop();
     }
 
+    @AfterAll
+    public static void cleanup() throws InterruptedException
+    {
+        System.gc();
+        Thread.sleep(1000);
+        System.gc();
+    }
```
joakime: This temporary, and is only used to demonstrate a wart.
It will be removed.

jetty-server/src/test/resources/jetty-logging.properties
```diff
@@ -4,3 +4,4 @@
 #org.eclipse.jetty.servlet.LEVEL=DEBUG
 #org.eclipse.jetty.server.ConnectionLimit.LEVEL=DEBUG
 #org.eclipse.jetty.server.AcceptRateLimit.LEVEL=DEBUG
+org.eclipse.jetty.util.ReferenceCleaner.LEVEL=DEBUG
```
sbordet: Comment this out.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
```diff
@@ -225,6 +220,14 @@ protected void onCompleteFailure(Throwable cause)
                     dispose();
                     callback.failed(cause);
                 }
+
+                @Override
+                public String toString()
+                {
+                    return String.format("FileBufferedInterceptor@%X[fileLength=%d,pos=%d,last=%b]",
```
sbordet: Don't hardcode the class name, use `getClass().getSimpleName()`.
Also, the hashCode should be `%x`, not `%X`.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
```diff
@@ -225,6 +220,14 @@ protected void onCompleteFailure(Throwable cause)
...
+                        this.hashCode(), fileLength, _pos, _last
```
sbordet: Drop `this.`, unnecessary.
Also, field `fileLength` should be prefixed with `_` for consistency.

jetty-util/src/main/java/org/eclipse/jetty/util/ReferenceCleaner.java
```diff
@@ -0,0 +1,92 @@
...
+        @Override
+        public void run()
+        {
+            delete(file);
+        }
+
+        @Override
+        public String toString()
+        {
+            return String.format("%s@%x[file='%s']", this.getClass().getSimpleName(), this.hashCode(), file);
```
sbordet: Drop `this.`, unnecessary.

sbordet (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
joakime: This is an anonymous class.

joakime: I'll just remove this one, as we moved the register from this anonymous class to the outer interceptor anyway.

sbordet (approved):
LGTM.
However, if we decide to not use mapping for Windows, then this PR is moot as it is not necessary for Linux.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/FileBufferedResponseHandler.java
```diff
@@ -106,18 +109,9 @@ protected void dispose()
...
-                {
-                    Files.delete(_filePath);
-                }
-                catch (Throwable t)
-                {
-                    LOG.warn("Could not delete file {}", _filePath, t);
-                }
-                _filePath = null;
-            }
+            ReferenceCleaner.DeleteFile.delete(_filePath);
```
lachlan-roberts, line 112: Maybe we should only register the file with the Cleaner if this delete fails. This way we should not have anything registered with the Cleaner except on windows.

joakime: I tried that, the .reset() and .resetBuffer() use cases made that approach difficult.
Besides, the line you flagged just deletes the file directly, no register step.

lachlan-roberts: > Besides, the line you flagged just deletes the file directly, no register step.

I meant to only register here if this line fails to delete the file, and to not register it with the Cleaner when the file is created.


> I tried that, the .reset() and .resetBuffer() use cases made that approach difficult.

What do you mean by the .reset() and .resetBuffer() cases?

gregw:
@joakime can you resolve conflicts and get a clean CI build

joakime:
I'll need to merge/rebase this PR.

PR #6597, opened 2021-08-10
Issue #6476 - warn on exec

### LINKED ISSUE #6476: Show message if JVM args are present but new JVM is spawned based on active modules
**Target Jetty version(s)**

10.0+

**Enhancement Description**

We recently upgraded to Jetty 10 and consequently used some more out-of-the-box modules that ship with jetty. One of the modules used an `[exec]` block which causes the creating of a new JVM. 

It would be nice if some message will be printed (info or warn) saying that a new JVM will be created. With our old configuration (Jetty 9.x) we used to start Jetty with `java $JAVA_OPTIONS -jar $JETTY_HOME/start.jar` with all kinds of java parameters in `$JAVA_OPTIONS`. This stopped working with our new configuration in Jetty 10, but it wasn't immediately obvious why it stopped working.

Ideally if you detect that Jetty has been launched with JVM args, e.g. `java --add-opens <module> -Xmx 2GB -jar ./start.jar` and with `exec` mode being active, then log a warning that JVM parameters will not have any effect for the newly spawned JVM. Alternatively you might be able to transparently use the provided JVM args for the newly spawned JVM as well. To get the JVM args you could use `ManagementFactory.getRuntimeMXBean().getInputArguments()`.

sbordet:


sbordet:
@gregw see if you want to modify `jetty.sh`.

joakime:
The output from `--list-config` should also indicate if a JVM will be forked or not.

sbordet:
> The output from `--list-config` should also indicate if a JVM will be forked or not.

You mean that it should be added, or that it is already indicated and should be documented?

joakime:
> > The output from `--list-config` should also indicate if a JVM will be forked or not.
> 
> You mean that it should be added, or that it is already indicated and should be documented?

I mean that it should be added (if it isn't there already).

jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
```diff
@@ -467,6 +468,17 @@ else if (args.isCreateFiles() || !args.getStartModules().isEmpty())
...
+
+            List<String> execModules = args.getEnabledModules().stream()
+                .map(name -> args.getAllModules().get(name))
+                // Keep only the forking modules.
+                .filter(module -> !module.getJvmArgs().isEmpty())
+                .map(Module::getName)
+                .collect(Collectors.toList());
+            StartLog.warn("Forking second JVM due to forking module(s): %s.", execModules);
+            StartLog.warn("Consider using option --dry-run to generate the command line:");
+            StartLog.warn("$(java -jar $JETTY_HOME/start.jar --dry-run)");
```
gregw: 3 lines of warning is a bit extreme!
```suggestion
            StartLog.warn("Forking JVM due to exec in module(s): %s.  Use --dry-run to generate non forking command line.", execModules);
```

gregw (changes requested):
Doco is good.  But 3 lines of warning is too much.

joakime:
> Doco is good. But 3 lines of warning is too much.

I completely disagree with this statement about the number of lines of warning.

This warning is important, it is more important than a normal warning, but not suitable for the next level of severity (eg: error or fatal).

There's TONS of examples of projects with multiline important warnings for a single topic/concept. (docker, k8s, osgi, testcontainers, dropwizard, spring, apt, log4j startup warnings, slf4j init warnings, logback init warnings, node, groovy, maven, etc)  I'll be happy to demonstrate each of them if you want.

The reason multiline warnings exists, is that single line warnings is on the wrong side of the "is this important or just noise" question that people have.
We want to cross that threshold with this PR and make sure we report this important information.

joakime (dismissed):
We need a `StartLog.important()` level of information conveyance to the user.

gregw:
We have certainly had complaints previously about over verbose logging and there is nothing said in the 3 lines that can't be said in 1.

Asking users to ignore multiple warnings is not good.

Better have an INFO level

sbordet:
@gregw @joakime I have updated the PR with your suggestions, nudge?

gregw (approved):

PR #6598, opened 2021-08-11
Fix #6476 Use dry run in jetty.sh

### LINKED ISSUE #6476: Show message if JVM args are present but new JVM is spawned based on active modules

gregw:
#6476  Use a --dry-run in jetty.sh to pre-expand the java arguments and thus avoid having two JVMs running in the case of exec.

Also made a small change to allow script to check the current directory for JETTY_BASE, as that allows testing and runs in the same style as direct calls to start.jar

### DIFF
--- a/jetty-home/src/main/resources/bin/jetty.sh
+++ b/jetty-home/src/main/resources/bin/jetty.sh
@@ -66,7 +66,8 @@ NAME=$(echo $(basename $0) | sed -e 's/^[SK][0-9]*//' -e 's/\.sh$//')
 #    <Arg><Property name="jetty.home" default="."/>/webapps/jetty.war</Arg>
 #
 # JETTY_BASE
-#   Where your Jetty base directory is.  If not set, the value from
+#   Where your Jetty base directory is.  If not set, then the currently
+#   directory is checked, otherwise the value from
 #   $JETTY_HOME will be used.
 #
 # JETTY_RUN
@@ -238,7 +239,6 @@ then
   fi
 fi
 
-
 ##################################################
 # No JETTY_HOME yet? We're out of luck!
 ##################################################
@@ -247,20 +247,23 @@ if [ -z "$JETTY_HOME" ]; then
   exit 1
 fi
 
+RUN_DIR=$PWD
 cd "$JETTY_HOME"
 JETTY_HOME=$PWD
-
 
 ##################################################
 # Set JETTY_BASE
 ##################################################
+export JETTY_BASE
 if [ -z "$JETTY_BASE" ]; then
-  JETTY_BASE=$JETTY_HOME
+  if [ -d "$RUN_DIR/start.d" -o -f "$RUN_DIR/start.ini" ]; then
+    JETTY_BASE=$RUN_DIR
+  else
+    JETTY_BASE=$JETTY_HOME
+  fi
 fi
-
 cd "$JETTY_BASE"
 JETTY_BASE=$PWD
-
 
 #####################################################
 # Check that jetty is where we think it is
@@ -430,7 +433,7 @@ case "`uname`" in
 CYGWIN*) JETTY_START="`cygpath -w $JETTY_START`";;
 esac
 
-RUN_ARGS=(${JAVA_OPTIONS[@]} -jar "$JETTY_START" ${JETTY_ARGS[*]})
+RUN_ARGS=$("$JAVA" ${JAVA_OPTIONS[@]} -jar "$JETTY_START" --dry-run=opts,path,main,args ${JETTY_ARGS[*]})
 RUN_CMD=("$JAVA" ${RUN_ARGS[@]})
 
 #####################################################
--- a/jetty-home/src/main/resources/bin/jetty.sh
+++ b/jetty-home/src/main/resources/bin/jetty.sh
@@ -433,8 +433,8 @@ case "`uname`" in
 CYGWIN*) JETTY_START="`cygpath -w $JETTY_START`";;
 esac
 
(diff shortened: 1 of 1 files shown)


jetty-home/src/main/resources/bin/jetty.sh
```diff
@@ -430,7 +433,7 @@ case "`uname`" in
 CYGWIN*) JETTY_START="`cygpath -w $JETTY_START`";;
 esac
 
-RUN_ARGS=(${JAVA_OPTIONS[@]} -jar "$JETTY_START" ${JETTY_ARGS[*]})
+RUN_ARGS=$("$JAVA" ${JAVA_OPTIONS[@]} -jar "$JETTY_START" --dry-run=opts,path,main,args ${JETTY_ARGS[*]})
 RUN_CMD=("$JAVA" ${RUN_ARGS[@]})
```
sbordet: If I define `JAVA_OPTIONS="--show-version"` then `RUN_ARGS` is set to the wrong string, which fails starting.

I think this is more correct:

```suggestion
RUN_ARGS=$("$JAVA" -jar "$JETTY_START" --dry-run=opts,path,main,args ${JETTY_ARGS[*]})
RUN_CMD=("$JAVA" ${JAVA_OPTIONS[@]} ${RUN_ARGS[@]})
```

sbordet (changes requested):

jetty-home/src/main/resources/bin/jetty.sh
gregw: @sbordet Good point, but I think the options need to end up in RUN_ARGS as they are sometimes used separately. We also need to protect again any options from module files that might make the output multiline.

jetty-home/src/main/resources/bin/jetty.sh
```diff
@@ -247,21 +247,24 @@ if [ -z "$JETTY_HOME" ]; then
   exit 1
 fi
 
+RUN_DIR=$PWD
```
joakime: Note that `$PWD` is a bash feature.
OSX is deprecating bash and moving to zsh. 
https://support.apple.com/en-us/HT208050
Basically, Apple is stuck with Bash 3.2 as it's GPLv2, but newer versions of bash are GPLv3.
Apple will not upgrade/ship and/or use anything with GPLv3 license.
They want to get rid of bash entirely simply as their version is very out of date, and have made it clear that it will go away in the future.

Perhaps we should move to `RUN_DIR=$(pwd)` instead? (it's more compatible)

joakime: Just to be clear, there is zero urgency on this change.
Feel free to leave it as it is for now.

gregw: ack

jetty-home/src/main/resources/bin/jetty.sh
```diff
@@ -430,7 +433,7 @@ case "`uname`" in
 CYGWIN*) JETTY_START="`cygpath -w $JETTY_START`";;
 esac
 
-RUN_ARGS=(${JAVA_OPTIONS[@]} -jar "$JETTY_START" ${JETTY_ARGS[*]})
+RUN_ARGS=$(echo $JAVA_OPTIONS ; "$JAVA" -jar "$JETTY_START" --dry-run=opts,path,main,args ${JETTY_ARGS[*]} | tail -1 )
```
sbordet: I don't like the `tail -1` as in "why do we take only the last line?" and "surely if there are multiple lines we don't want to take only the last one and discard the previous ones".
I'd rather fail to start if people puts newlines into `JAVA_OPTIONS`.

sbordet (changes requested):

jetty-home/src/main/resources/bin/jetty.sh
gregw: We have no control over what JAVA_OPTIONS users set. We have no control over what output those options generate or indeed if the JVM generates any other output.
If fact I think we should do more, something like:
```
 java --show-version -jar ../jetty-home/start.jar --dry-run=opts,path,main,args __jetty__=__args__ 2>/dev/null | grep __jetty__=__args__ | sed 's/ __jetty__=__args__ / /'
```

gregw: I think this is simpler!

sbordet: > We have no control over what JAVA_OPTIONS users set.

That's exactly why we should do exactly nothing to try to format it. We know our `--dry-run` generates a single line, and if they set JAVA_OPTIONS right, then there is nothing else to do, no `tail`, no `sed`, etc.

We don't do `tail -1` on the other variables in this file that can be set by users, so why do it here?
Am I missing something?

sbordet (approved):

PR #6606, opened 2021-08-12
Fix flaky HttpOutputTest.testEmptyBufferKnown

### LINKED ISSUE #6605: Flaky test: HttpOutputTest.testEmptyBufferKnown
**Jetty version(s)**
9.4.x

**Description**
`HttpOutputTest.testEmptyBufferKnown` fails about 1 of every 20 times on the following assertion:

```java
assertThat(committed.get(), is(false));
```

lorban:
Fixes #6605 

Make sure the assertion does not happen before the committed flag is read.

### DIFF
--- a/jetty-server/src/test/java/org/eclipse/jetty/server/HttpOutputTest.java
+++ b/jetty-server/src/test/java/org/eclipse/jetty/server/HttpOutputTest.java
@@ -27,6 +27,7 @@ import java.nio.ByteBuffer;
 import java.nio.channels.ReadableByteChannel;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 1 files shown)


sbordet (approved):

jetty-server/src/test/java/org/eclipse/jetty/server/HttpOutputTest.java
```diff
@@ -852,7 +853,7 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
     @Test
     public void testEmptyBufferKnown() throws Exception
```
sbordet: The current test name is weird.
Can you rename the test to be `testResponseIsCommittedWithZeroContentLength()` or similar?

sbordet (changes requested):

jetty-server/src/test/java/org/eclipse/jetty/server/HttpOutputTest.java
lorban: This test is mostly about testing what's going on with an empty buffer, so we should keep the `testEmptyBuffer` prefix.

WDYT of `testEmptyBufferWithZeroContentLength`?

sbordet: `testResponseCommittedWithEmptyBufferAndZeroContentLength()`?

sbordet (approved):

PR #6610, opened 2021-08-12
Fixes #6372 - Review socket options configuration

### LINKED ISSUE #6372: Review socket options configuration

sbordet:
Introduced in ClientConnector:
* tcpNoDelay
* reusePort
* receiveBufferSize
* sendBufferSize

Reworked configuration of socket options in ClientConnector.
JMX-ified ClientConnector.

Introduced reusePort in ServerConnector.
Updated server modules with the new reusePort property.

### DIFF
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
@@ -18,6 +18,7 @@ import java.net.InetSocketAddress;
 import java.net.ProtocolFamily;
 import java.net.SocketAddress;
 import java.net.SocketException;
+import java.net.SocketOption;
 import java.net.StandardProtocolFamily;
 import java.net.StandardSocketOptions;
 import java.nio.channels.SelectableChannel;
@@ -33,6 +34,8 @@ import java.util.concurrent.Executor;
 import org.eclipse.jetty.util.IO;
 import org.eclipse.jetty.util.JavaVersion;
 import org.eclipse.jetty.util.Promise;
+import org.eclipse.jetty.util.annotation.ManagedAttribute;
+import org.eclipse.jetty.util.annotation.ManagedObject;
 import org.eclipse.jetty.util.component.ContainerLifeCycle;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 import org.eclipse.jetty.util.thread.QueuedThreadPool;
@@ -41,6 +44,10 @@ import org.eclipse.jetty.util.thread.Scheduler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * <p>The client-side component that connects to server sockets.</p>
+ */
+@ManagedObject
 public class ClientConnector extends ContainerLifeCycle
 {
     public static final String CLIENT_CONNECTOR_CONTEXT_KEY = "org.eclipse.jetty.client.connector";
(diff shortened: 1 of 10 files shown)


jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
```diff
@@ -173,26 +206,120 @@ public void setIdleTimeout(Duration idleTimeout)
...
     public void setBindAddress(SocketAddress bindAddress)
     {
         this.bindAddress = bindAddress;
     }
 
+    /**
+     * @return whether small TCP packets are sent without delay
+     */
+    @ManagedAttribute("Whether small TCP packets are sent without delay")
+    public boolean isTCPNoDelay()
```
gregw, line 262: Should we go with camel case conventions for these new methods: `isTcpNoDelay`?

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
```diff
@@ -41,6 +44,10 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * <p>The client-side component that connects to server sockets.</p>
```
gregw, line 48: Maybe an example or an @see here to show how to get/use this class

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
```diff
@@ -173,26 +206,120 @@ public void setIdleTimeout(Duration idleTimeout)
...
+    public void setReusePort(boolean reusePort)
+    {
+        this.reusePort = reusePort;
+    }
+
+    /**
+     * @return the receive buffer size in bytes, or -1 for the default value
+     */
+    @ManagedAttribute("The receive buffer size in bytes")
+    public int getReceiveBufferSize()
```
gregw, line 321: There appears no test for these new methods?

gregw (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
sbordet: No, TCP is an initialism, not a word. See for example `java.net.URLEncoder`, not camel-cased.

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
sbordet: Some property is tested, but some are trivially passed in as options.
Furthermore, some of these are just "hints" to the OS, so may or may not have effect.
I thought they were simple enough and deserved no testing.

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
sbordet: @gregw I expanded the javadocs.
`@see` references are not possible because we can't really reference anything outside `jetty-io`.
The documentation is already plenty about how to use `ClientConnector`.

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
gregw: That's an old class. The style now is to lower case even acronyms (see https://google.github.io/styleguide/javaguide.html#s5-naming)

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
```diff
@@ -41,6 +44,32 @@
...
+ * <pre>
+ * // Create a ClientConnector instance.
+ * ClientConnector connector = new ClientConnector();
+ *
+ * // Configure the ClientConnector.
+ * connector.setSelectors(1);
+ * connector.setSslContextFactory(new SslContextFactory.Client());
+ *
+ * // Pass it to the HttpClient transport.
+ * HttpClientTransport transport = new HttpClientTransportDynamic(clientConnector);
```
gregw: The `Connector` is not called `clientConnector` above.

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
gregw: I was more looking for an example of the code that you just put into the javadoc at the start of this class (which has an error in it), but I see now that ClientConnector is already tested in a few places... but would be nice to see at least some tests call the setters for buffer size perhaps... not a blocker.

gregw (changes requested):

Issue #6618, opened 2021-08-14
ID token `azp` claim should not be required if `aud` is single value array


agavrilov76:
**Jetty version(s)**
9,10,11

**Java version/vendor** `(use: java -version)`
N/A

**OS type/version**
N/A

**Description**
If audience of an ID token is an array containing a single value, the `azp` claim should not be required.
https://openid.net/specs/openid-connect-core-1_0.html#IDToken

**How to reproduce?**

A sample ID token: 

`eyJraWQiOiIxIiwiYWxnIjoiUlMyNTYifQ.eyJhY3IiOiIxIiwic3ViIjoiNTg5ODQyMjA1MzgyMzg0ODQ0OCIsImF1ZCI6WyJ0ZWxlbm9yZGlnaXRhbC1kb2Jwb3J0YWx0ZWxlbm9yaWQtd2ViIl0sImF1dGhfdGltZSI6MTYyODk0ODA2NywiYW1yIjpbIlNTTyJdLCJpc3MiOiJodHRwczpcL1wvc2lnbmluLnRlbGVub3JpZC1zdGFnaW5nLmNvbVwvb2F1dGgiLCJ0ZF9zbHMiOmZhbHNlLCJleHAiOjE2Mjg5NTI5MzgsImlhdCI6MTYyODk0OTAzOH0.hKSYTSxJj3vzqKrvjDGqu45hrYnrnKTRrh9BmoGUADO48OGkh0GC9H2Qm529qi8iQKO5PXK6qc1lnDpfX8xv0zmwYSPJHwRrtFEXjNLrEIsuTdtrK59Tf7t7LXdxduZI95YfWtEqbPQ7EO-ykDCQev44pE0ry4IvrsFRYUyf7HU
`

causes the following exception:

```
WARN  [2021-08-14 14:23:01,889] org.eclipse.jetty.security.openid.OpenIdLoginService: 
! org.eclipse.jetty.security.openid.OpenIdCredentials$AuthenticationException: A multi-audience ID token needs to contain an azp claim
! at org.eclipse.jetty.security.openid.OpenIdCredentials.validateAudience(OpenIdCredentials.java:150)
! at org.eclipse.jetty.security.openid.OpenIdCredentials.validateClaims(OpenIdCredentials.java:120)
! at org.eclipse.jetty.security.openid.OpenIdCredentials.redeemAuthCode(OpenIdCredentials.java:103)
[...]

PR #6620, opened 2021-08-16
Issue #6618 - azp claim should not be required for single value aud array

lachlan-roberts:
## Issue #6618

It currently incorrectly assumed that any array for the `aud` claim will contain more than 1 value.

We should only enforce an `azp` claim if the array for `aud` is longer than 1.

### DIFF
--- a/jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdCredentials.java
+++ b/jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdCredentials.java
@@ -46,6 +46,14 @@ public class OpenIdCredentials implements Serializable
     private String authCode;
     private Map<String, Object> response;
     private Map<String, Object> claims;
+    private boolean verified = false;
+
+    public OpenIdCredentials(Map<String, Object> claims)
+    {
+        this.redirectUri = null;
+        this.authCode = null;
+        this.claims = claims;
+    }
 
     public OpenIdCredentials(String authCode, String redirectUri)
     {
@@ -96,7 +104,6 @@ public class OpenIdCredentials implements Serializable
                 claims = JwtDecoder.decode(idToken);
                 if (LOG.isDebugEnabled())
                     LOG.debug("claims {}", claims);
-                validateClaims(claims, configuration);
             }
             finally
             {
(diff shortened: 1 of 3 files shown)


jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdCredentials.java
```diff
@@ -105,14 +106,14 @@ public void redeemAuthCode(OpenIdConfiguration configuration) throws Exception
         }
     }
 
-    private void validateClaims(OpenIdConfiguration configuration) throws Exception
+    static void validateClaims(Map<String, Object> claims, OpenIdConfiguration configuration) throws Exception
```
gregw: It's a little strange to have this as a static method and then need to pass in claims.  Why not just make it a normal method?

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdCredentials.java
```diff
@@ -126,7 +127,7 @@ private void validateClaims(OpenIdConfiguration configuration) throws Exception
             throw new AuthenticationException("ID Token has expired");
     }
 
-    private void validateAudience(OpenIdConfiguration configuration) throws AuthenticationException
+    private static void validateAudience(Map<String, Object> claims, OpenIdConfiguration configuration) throws AuthenticationException
```
gregw: why static and passing in claims? why not keep as a non-static?  If the answer is for testing.... I'd rather keep the design correct... perhaps with a special constructor.

gregw (changes requested):

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdCredentials.java
lachlan-roberts: I don't really think it makes the design incorrect, it makes sense to have a method which validates claims given a configuration. And this doesn't impact anything because these methods are not public. I could add a constructor for `OpenIdCredentials(Map<String, Object> claims)` but that makes less sense to me.

I'll give it a try and see if it makes it any better.

gregw (approved):
well that certainly makes it simpler to read!

sbordet (approved):

Issue #6624, opened 2021-08-16
org.eclipse.jetty.client.HttpClientTLSTest#testForcedNonDomainSNI fails on java17


gregw:
**Jetty version(s)**
9, 10, 11

**Java version/vendor** `(use: java -version)`
```
openjdk version "17-ea" 2021-09-14
```
also
```
openjdk version "17" 2021-09-14
OpenJDK Runtime Environment (build 17+35-2724)
OpenJDK 64-Bit Server VM (build 17+35-2724, mixed mode, sharing)
```
**OS type/version**

linux

**Description**

Test org.eclipse.jetty.client.HttpClientTLSTest#testForcedNonDomainSNI fails on java17:
```
2021-08-17 09:29:13.894:WARN:oeji.ManagedSelector:client-31: java.lang.IllegalArgumentException: The encoded server name value is invalid
2021-08-17 09:29:13.903:INFO:oejs.AbstractConnector:main: Stopped ServerConnector@6279cee3{SSL, (ssl, http/1.1)}{0.0.0.0:0}

java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: The encoded server name value is invalid

	at org.eclipse.jetty.client.util.FutureResponseListener.getResult(FutureResponseListener.java:118)
	at org.eclipse.jetty.client.util.FutureResponseListener.get(FutureResponseListener.java:101)
	at org.eclipse.jetty.client.HttpRequest.send(HttpRequest.java:730)
	at org.eclipse.jetty.client.HttpClientTLSTest.testForcedNonDomainSNI(HttpClientTLSTest.java:1022)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:78)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:688)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
[...]

PR #6625, opened 2021-08-16
Disable ipv6 test for #6624

gregw:
Temp disable of ipv6 test for #6624,  that is breaking the build.

### DIFF
--- a/jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTLSTest.java
+++ b/jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTLSTest.java
@@ -63,7 +63,6 @@ import org.eclipse.jetty.server.SecureRequestCustomizer;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
 import org.eclipse.jetty.server.SslConnectionFactory;
-import org.eclipse.jetty.toolchain.test.Net;
 import org.eclipse.jetty.util.StringUtil;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 import org.eclipse.jetty.util.thread.ExecutorThreadPool;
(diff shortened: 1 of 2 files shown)


gregw:
CI failure is unrelated flakey test. See #6626

jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTLSTest.java
```diff
@@ -1014,6 +1013,7 @@ public void testForcedNonDomainSNI() throws Exception
             .send();
         assertEquals(HttpStatus.OK_200, response2.getStatus());
 
+        /* TODO Fix. See #6624
```
olamy, line 1016: what about marking the test method with `@DisabledOnJre({JRE.JAVA_17})` 
easier to search for common annotation `@Disabled*`

olamy (changes requested):
I would prefer usage of the annotation `@Disabled`

jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTLSTest.java
gregw: Because the test is parameterized and tests other stuff.  So the test is not disabled, just the IPv6 part of it is.

lachlan-roberts: The problem is its hard to keep track of TODOs because we have so many of them in the code. So it might just get lost and never be fixed, maybe if you at least commented in the issue as a reminder so it is not forgotten.

lachlan-roberts (approved):

olamy (approved):

jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTLSTest.java
gregw: Will do.   This PR does not close the issue, so the issue will remain open.

PR #6628, opened 2021-08-17
Fix flaky test from #6562/#6563

gregw:
Fix flaky test from #6562/#6563


jetty-server/src/test/java/org/eclipse/jetty/server/HttpOutputTest.java
```diff
@@ -864,7 +863,7 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
                 response.setContentLength(0);
                 ((HttpOutput)response.getOutputStream()).write(ByteBuffer.wrap(new byte[0]));
                 assertThat(response.isCommitted(), is(true));
-                latch.countDown();
+                committed.succeeded(response.isCommitted());
```
janbartel, line 864: response.isCommitted() can never be false here, otherwise the assert would fail.

janbartel (dismissed):

lachlan-roberts (approved):

olamy (approved):

PR #6635, opened 2021-08-18
Issue #6566 - use the demand API for the websocket MessageSinks

lachlan-roberts:
## Issue #6566

By using the demand API in the websocket `MessageSink`s we can demand all the frames of a websocket message without succeeding the callback and releasing control of the frame payload buffers. This allows the `MessageSink`s to not need to copy the payload of each buffer and just store the buffer/callback pair. Now only one copy must be done which is to combine the frame payloads directly before it is delivered to the application.


jetty-io/src/main/java/org/eclipse/jetty/io/BufferCallbackAccumulator.java
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * Get the amount of bytes which have been accumulated.
+     * This will add up the remaining of each buffer in the accumulator.
+     * @return the total length of the content in the accumulator.
+     */
+    public int getLength()
+    {
+        int length = 0;
+        for (Entry entry : _entries)
+            length = Math.addExact(length, entry.buffer.remaining());
```
gregw: Surely there is some unintelligble stream based way of doing this:
```suggestion
        length = _entries.stream().map(Entry::getBuffer).map(ByteBuffer::remaining).sum();
```

jetty-io/src/main/java/org/eclipse/jetty/io/BufferCallbackAccumulator.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * Get the amount of bytes which have been accumulated.
+     * This will add up the remaining of each buffer in the accumulator.
+     * @return the total length of the content in the accumulator.
+     */
+    public int getLength()
+    {
+        int length = 0;
```
gregw: could length be `> Integer.MAX_VALUE` ?

jetty-io/src/main/java/org/eclipse/jetty/io/BufferCallbackAccumulator.java
```diff
@@ -0,0 +1,93 @@
...
+        Entry(ByteBuffer buffer, Callback callback)
+        {
+            this.buffer = buffer;
+            this.callback = callback;
+        }
+    }
+
+    public void addEntry(ByteBuffer buffer, Callback callback)
+    {
+        _entries.add(new Entry(buffer, callback));
```
gregw: Since we know that the length will ultimately be needed to take the content, we may as well keep a running total of the length here.  Then if it exceeds Integer.MAX_VALUE (or some other max), we can throw here rather than continue.

jetty-io/src/main/java/org/eclipse/jetty/io/BufferCallbackAccumulator.java
```diff
@@ -0,0 +1,93 @@
...
+        int length = 0;
+        for (Entry entry : _entries)
+            length = Math.addExact(length, entry.buffer.remaining());
+        return length;
+    }
+
+    /**
+     * @return a newly allocated byte array containing all content written into the accumulator.
+     */
+    public byte[] toByteArray()
```
gregw: I prefer `take` than `to` to make it clear that it clears the accumulator.
```suggestion
    public byte[] takeByteArray()
```

jetty-io/src/main/java/org/eclipse/jetty/io/BufferCallbackAccumulator.java
gregw: I see the max size is protected in where this is used... but this class is in io utility package.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ByteArrayMessageSink.java
```diff
@@ -55,8 +53,8 @@ public void accept(Frame frame, Callback callback)
             long maxBinaryMessageSize = session.getMaxBinaryMessageSize();
             if (maxBinaryMessageSize > 0 && size > maxBinaryMessageSize)
             {
-                throw new MessageTooLargeException(String.format("Binary message too large: (actual) %,d > (configured max binary message size) %,d",
-                    size, maxBinaryMessageSize));
+                callback.failed(new MessageTooLargeException(
```
gregw: Instead of recalculating the size here (which is also done by the accumulator), why not delegate the max size to the accumulator itself?
At the very least, let the accumulator calculate a running total of the size and ask it here rather than have 2 counts.

gregw: Also, don't all the previous callbacks need to be failed as well?

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ByteArrayMessageSink.java
```diff
@@ -55,8 +53,8 @@ public void accept(Frame frame, Callback callback)
...
+                    String.format("Binary message too large: (actual) %,d > (configured max binary message size) %,d", size, maxBinaryMessageSize)));
```
gregw: don't we need a return here?   
Or perhaps a throw so the finally block below is used???

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ByteArrayMessageSink.java
```diff
@@ -97,14 +102,14 @@ public void accept(Frame frame, Callback callback)
         }
     }
 
-    private void aggregatePayload(Frame frame) throws IOException
+    private void aggregatePayload(Frame frame, Callback callback)
```
gregw: this method feels verbose and like it is double checking stuff.   I think if you inline it then the hasPayload stuff will already have been done.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ByteBufferMessageSink.java
```diff
@@ -68,41 +67,58 @@ public void accept(Frame frame, Callback callback)
...
         {
             ByteBuffer payload = frame.getPayload();
-
             if (out == null)
-                out = new ByteArrayOutputStream(BUFFER_SIZE);
-
-            BufferUtil.writeTo(payload, out);
-            payload.position(payload.limit()); // consume buffer
+                out = new BufferCallbackAccumulator();
+            out.addEntry(payload, callback);
```
gregw: ditto about the size calculations.  Don't do it twice.

gregw (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/BufferCallbackAccumulator.java
lachlan-roberts: I'll replace this with a counter as they are added.

jetty-io/src/main/java/org/eclipse/jetty/io/BufferCallbackAccumulator.java
lachlan-roberts: No because we use addExact which checks for integer overflow.

We don't want to support lengths over `Integer.MAX_VALUE` anyway because it won't be able to be accumulated into a single ByteBuffer.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ByteArrayMessageSink.java
lachlan-roberts: Yeah I thought I reverted this change, we should throw here so it will be caught and will fail all the callbacks.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ByteArrayMessageSink.java
lachlan-roberts: It's not double checking anything but I can inline it.

sbordet:
@lachlan-roberts can you fix the test failures?

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ByteArrayMessageSink.java
```diff
@@ -55,8 +53,8 @@ public void accept(Frame frame, Callback callback)
             long maxBinaryMessageSize = session.getMaxBinaryMessageSize();
             if (maxBinaryMessageSize > 0 && size > maxBinaryMessageSize)
             {
-                throw new MessageTooLargeException(String.format("Binary message too large: (actual) %,d > (configured max binary message size) %,d",
-                    size, maxBinaryMessageSize));
+                throw new MessageTooLargeException(
+                    String.format("Binary message too large: (actual) %,d > (configured max binary message size) %,d", size, maxBinaryMessageSize));
```
gregw, line 56: The `size` above is still a double calculation of the length.  Can't it be replaced with `out == null ? 0 : out.getLength()`

gregw (changes requested):

gregw (dismissed):

jetty-io/src/main/java/org/eclipse/jetty/io/BufferCallbackAccumulator.java
```diff
@@ -0,0 +1,99 @@
...
+
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+import org.eclipse.jetty.util.BufferUtil;
+import org.eclipse.jetty.util.Callback;
+
+public class BufferCallbackAccumulator
```
sbordet: Should be renamed to `ByteBufferCallbackAccumulator`.

jetty-io/src/main/java/org/eclipse/jetty/io/BufferCallbackAccumulator.java
```diff
@@ -0,0 +1,99 @@
...
+
+    public void addEntry(ByteBuffer buffer, Callback callback)
+    {
+        _entries.add(new Entry(buffer, callback));
+        _length = Math.addExact(_length, buffer.remaining());
+    }
+
+    /**
+     * Get the amount of bytes which have been accumulated.
+     * This will add up the remaining of each buffer in the accumulator.
```
sbordet: This javadocs is weird. I would just leave the `@return` because it should be enough to explain what this method does: it returns the number of bytes accumulated so far.

jetty-io/src/main/java/org/eclipse/jetty/io/BufferCallbackAccumulator.java
sbordet: Please add a minimal javadoc.

jetty-io/src/main/java/org/eclipse/jetty/io/BufferCallbackAccumulator.java
```diff
@@ -0,0 +1,99 @@
...
+            _length -= entry.buffer.remaining();
+            buffer.put(entry.buffer);
+            iterator.remove();
+            entry.callback.succeeded();
+        }
+
+        if (!_entries.isEmpty())
+            throw new IllegalStateException("remaining entries: " + _entries.size());
+        if (_length != 0)
+            throw new IllegalStateException("non-zero length: " + _length);
```
sbordet: Are these checks done just as paranoid checks in case of concurrent usage?
I'm not sure they are worth it, as the class is not designed to be concurrent at all.

jetty-io/src/main/java/org/eclipse/jetty/io/BufferCallbackAccumulator.java
```diff
@@ -0,0 +1,99 @@
...
+            return new byte[0];
+
+        byte[] bytes = new byte[length];
+        ByteBuffer buffer = BufferUtil.toBuffer(bytes);
+        BufferUtil.clearToFill(buffer);
+        writeTo(buffer);
+        return bytes;
+    }
+
+    public void writeTo(ByteBuffer buffer)
```
sbordet: Method is `public` so anyone can call it with possibly a buffer of the wrong capacity.
I would add a check.

jetty-util/src/main/java/org/eclipse/jetty/util/Callback.java
```diff
@@ -133,7 +133,7 @@ public void failed(Throwable x)
     }
 
     /**
-     * Creaste a callback that runs completed when it succeeds or fails
+     * Create a callback that runs completed when it succeeds or fails
```
sbordet: Javadocs should be written in 3rd person, so "Creates".

jetty-util/src/main/java/org/eclipse/jetty/util/Callback.java
```diff
@@ -396,6 +396,32 @@ public InvocationType getInvocationType()
...
     {
+        /**
+         * Create a completable future given a callback.
+         *
+         * @param callback The nested callback.
+         * @return a new Completable which will succeed this callback when completed.
+         */
+        public static Completable from(Callback callback)
+        {
+            return new Completable()
```
sbordet: Pass `callback.getInvocationType()` to the constructor.

jetty-io/src/main/java/org/eclipse/jetty/io/BufferCallbackAccumulator.java
```diff
@@ -0,0 +1,99 @@
...
+    {
+        for (Iterator<Entry> iterator = _entries.iterator(); iterator.hasNext();)
+        {
+            Entry entry = iterator.next();
+            _length -= entry.buffer.remaining();
+            buffer.put(entry.buffer);
+            iterator.remove();
+            entry.callback.succeeded();
+        }
```
sbordet: This `for` loop seems too complicated. How about this:

```suggestion
        for (Entry entry : _entries)
        {
            buffer.put(entry.buffer);
            entry.callback.succeeded();
        }
        _entries.clear();
        _length = 0;
```

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/DispatchedMessageSink.java
```diff
@@ -135,22 +135,28 @@ public void accept(Frame frame, final Callback callback)
...
+                    session.demand(1);
             });
         }
+        else
+        {
+            frameCallback = Callback.from(() ->
+            {
+                callback.succeeded();
+                session.demand(1);
+            }, callback::failed);
```
sbordet: We want to preserve the `InvocationType` of `callback`, so do this instead:

```suggestion
            frameCallback = new Callback.Nested(callback)
            {
                @Override
                public void succeeded()
                {
                    super.succeeded();
                    session.demand(1);
                }
            };
```

jetty-websocket/websocket-jetty-common/src/main/java/org/eclipse/jetty/websocket/common/JettyWebSocketFrameHandler.java
```diff
@@ -226,16 +228,7 @@ public void onFrame(Frame frame, Callback callback)
...
-                {
-                    demand();
-                }
-                catch (Throwable t)
-                {
-                    callback.failed(t);
-                    return;
-                }
-
+                demand();
```
sbordet: I compare this code with the correspondent in `JavaxWebSocketFrameHandler` and they are too different.
There, each `on*Frame()` method was explicitly demanding; here, we wrap the callback, but then only pass it to certain frame types, which raises the question why we do it, but only for certain methods?
Can't we implement this like it was done for `JavaxWebSocketFrameHandler`, which is clearer?

sbordet (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/BufferCallbackAccumulator.java
lachlan-roberts: They are done just as a sanity check. If someone has changed the content of the buffer since it was accumulated the length could be different to when it was originally added. So the `_length` field after this method could potentially be a positive or negative non-zero value.

They could be removed if you don't think its necessary, I just thought it was safer to include them.

jetty-io/src/main/java/org/eclipse/jetty/io/BufferCallbackAccumulator.java
lachlan-roberts: I don't think it matters, it will throw if the buffer does not have enough space, and will only succeed the callbacks up to the point at which it failed.

jetty-util/src/main/java/org/eclipse/jetty/util/Callback.java
lachlan-roberts: The majority of the javadocs in this class use "Create" instead of "Creates". All the ones written by Greg use "Create" and all the ones by you use "Creates".

I'll just change all of them to "Creates" then.

jetty-io/src/main/java/org/eclipse/jetty/io/BufferCallbackAccumulator.java
lachlan-roberts: The problem is that if one of the puts fails, or one of the succeeded throws, then you will both succeed and fail some of the callbacks in the list. By removing them as we iterate we can be sure we only succeed or fail once for each callback.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/DispatchedMessageSink.java
lachlan-roberts: I have changed this usage, but this probably needs a larger review of general `Callback` usage, there is a ton of usages of `from(Runnable success, Consumer<Throwable> failure)` that wrap another callback and don't preserve the `InvocationType`.

jetty-websocket/websocket-jetty-common/src/main/java/org/eclipse/jetty/websocket/common/JettyWebSocketFrameHandler.java
lachlan-roberts: done

jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferCallbackAccumulator.java
```diff
@@ -0,0 +1,101 @@
...
+            _length -= entry.buffer.remaining();
+            buffer.put(entry.buffer);
+            iterator.remove();
+            entry.callback.succeeded();
+        }
+
+        if (!_entries.isEmpty())
+            throw new IllegalStateException("remaining entries: " + _entries.size());
+        if (_length != 0)
+            throw new IllegalStateException("non-zero length: " + _length);
```
sbordet: 1. Add check for `buffer.remaining() >= _length` at beginning of method.
1. Simplify loop with `for (entry : entries)` assume `put()` and `succeeded()` don't throw.
1. Drop these extra checks.
1. Reset _entries and _length at the end of the iteration.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/DispatchedMessageSink.java
```diff
@@ -135,22 +135,32 @@ public void accept(Frame frame, final Callback callback)
...
         }
+        else
+        {
+            frameCallback = new Callback.Nested(callback)
+            {
+                @Override
+                public void succeeded()
+                {
+                    session.demand(1);
+                    super.succeeded();
```
sbordet, line 159: Should not these 2 lines be inverted?

jetty-websocket/websocket-jetty-common/src/main/java/org/eclipse/jetty/websocket/common/JettyWebSocketFrameHandler.java
```diff
@@ -342,6 +326,7 @@ private void acceptMessage(Frame frame, Callback callback)
         if (activeMessageSink == null)
         {
             callback.succeeded();
+            coreSession.demand(1);
```
sbordet: Call `demand()` instead?

sbordet (dismissed):

sbordet (commented):
LGTM, but there are test failures.

jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferCallbackAccumulator.java
```diff
@@ -73,7 +73,7 @@ public int getLength()
 
     public void writeTo(ByteBuffer buffer)
```
gregw, line 74: This param should be called `bufferInFillMode` to make it clear it is deviating from our standard policy of keeping buffers in flush state.  It also needs javadoc saying as much.

.... or better yet, do the flip to fill within this method so anybody calling it does not need to do the flip and flipback themselves

gregw (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferCallbackAccumulator.java
lachlan-roberts: I have implemented your second suggestion.

gregw (approved):

PR #6639, opened 2021-08-18
Fixes #6603 - HTTP/2 max local stream count exceeded

### LINKED ISSUE #6603: HTTP/2 max local stream count exceeded
**Jetty version(s)** 9.4.41.v20210516

**Java version/vendor** 1.8.0_271

**OS type/version**: Linux and Windows 10 machine

**Description** 
HTTP/2 max local stream count exceeded

We were using jetty 9.4.35 version  and we encountered this "HTTP/2 max local stream count 1 exceeded" error ,Simon suggested us to upgrade jetty version to 9.4.41 , so we did jetty jar upgrade to 9.4.41 version , but still we see the error .

Below is the code snippet we have written to create http2 client to post requests to destination host.

     HttpClient httpClient = null;
     HTTP2Client http2Client=null;
        try {
            http2Client = new HTTP2Client();
            http2Client.start();
           SslContextFactory sslContextFactory = new SslContextFactory.Client(true);
           final KeyStore sslKeyStore = loadKeystore(); //loadKeystore method loads keystire
          if (sslKeyStore != null) {
             sslContextFactory.setKeyStore(sslKeyStore);
           }
            httpClient = new HttpClient(new HttpClientTransportOverHTTP2(http2Client),
                    sslContextFactory);
            httpClient.setMaxConnectionsPerDestination(100);
            httpClient.setMaxRequestsQueuedPerDestination(250000);
            httpClient.start();
            return httpClient;
        }
        catch (Exception exp) {
       }

**How to reproduce?**
Set the MaxConnectionsPerDestination value to 100 and try to post more than 10K requests.

sbordet:
Made MAX_CONCURRENT_STREAMS setting work on a per-connection basis.

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/ConnectionPool.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/ConnectionPool.java
@@ -109,13 +109,22 @@ public interface ConnectionPool extends Closeable
     interface Multiplexable
     {
         /**
-         * @return the max number of requests multiplexable on a single connection
+         * @return the default max number of requests multiplexable on a connection
          */
         int getMaxMultiplex();
 
         /**
-         * @param maxMultiplex the max number of requests multiplexable on a single connection
+         * @param maxMultiplex the default max number of requests multiplexable on a connection
          */
         void setMaxMultiplex(int maxMultiplex);
+
+        /**
+         * @param connection the multiplexed connection
+         * @param maxMultiplex the max number of requests multiplexable on the given connection
+         */
+        default void setMaxMultiplex(Connection connection, int maxMultiplex)
+        {
+            setMaxMultiplex(maxMultiplex);
+        }
     }
 }
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/ConnectionPool.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/ConnectionPool.java
@@ -104,27 +104,18 @@ public interface ConnectionPool extends Closeable
     }
 
     /**
-     * Marks a connection pool as supporting multiplexed connections.
+     * Marks a connection as supporting multiplexed requests.
      */
     interface Multiplexable
     {
         /**
-         * @return the default max number of requests multiplexable on a connection
+         * @return the max number of requests that can be multiplexed on a connection
          */
         int getMaxMultiplex();
 
         /**
-         * @param maxMultiplex the default max number of requests multiplexable on a connection
+         * @param maxMultiplex the max number of requests that can be multiplexed on a connection
          */
         void setMaxMultiplex(int maxMultiplex);
-
-        /**
-         * @param connection the multiplexed connection
-         * @param maxMultiplex the max number of requests multiplexable on the given connection
-         */
-        default void setMaxMultiplex(Connection connection, int maxMultiplex)
-        {
-            setMaxMultiplex(maxMultiplex);
-        }
     }
 }
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/ConnectionPool.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/ConnectionPool.java
@@ -102,20 +102,4 @@ public interface ConnectionPool extends Closeable
          */
         ConnectionPool newConnectionPool(HttpDestination destination);
     }
-
-    /**
-     * Marks a connection as supporting multiplexed requests.
-     */
-    interface Multiplexable
-    {
-        /**
-         * @return the max number of requests that can be multiplexed on a connection
-         */
-        int getMaxMultiplex();
-
-        /**
-         * @param maxMultiplex the max number of requests that can be multiplexed on a connection
-         */
-        void setMaxMultiplex(int maxMultiplex);
-    }
 }
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/MultiplexConnectionPool.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/MultiplexConnectionPool.java
@@ -34,12 +34,21 @@ public class MultiplexConnectionPool extends AbstractConnectionPool
 
     public MultiplexConnectionPool(HttpDestination destination, int maxConnections, boolean cache, Callback requester, int maxMultiplex)
     {
-        this(destination, new Pool<>(Pool.StrategyType.FIRST, maxConnections, cache), requester, maxMultiplex);
+        this(destination, Pool.StrategyType.FIRST, maxConnections, cache, requester, maxMultiplex);
     }
 
-    public MultiplexConnectionPool(HttpDestination destination, Pool<Connection> pool, Callback requester, int maxMultiplex)
+    public MultiplexConnectionPool(HttpDestination destination, Pool.StrategyType strategy, int maxConnections, boolean cache, Callback requester, int maxMultiplex)
     {
-        super(destination, pool, requester);
+        super(destination, new Pool<Connection>(strategy, maxConnections, cache)
+        {
+            @Override
+            protected int getMaxMultiplex(Connection connection)
+            {
+                if (connection instanceof Multiplexable)
+                    return ((Multiplexable)connection).getMaxMultiplex();
+                return super.getMaxMultiplex(connection);
+            }
+        }, requester);
         setMaxMultiplex(maxMultiplex);
     }
 
@@ -67,5 +76,16 @@ public class MultiplexConnectionPool extends AbstractConnectionPool
     public void setMaxUsageCount(int maxUsageCount)
     {
         super.setMaxUsageCount(maxUsageCount);
+    }
+
+    /**
+     * Marks a connection as supporting multiplexed requests.
+     */
+    public interface Multiplexable extends Connection
+    {
+        /**
+         * @return the max number of requests that can be multiplexed on a connection
+         */
+        int getMaxMultiplex();
     }
 }
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/MultiplexConnectionPool.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/MultiplexConnectionPool.java
@@ -25,7 +25,7 @@ import org.eclipse.jetty.util.annotation.ManagedAttribute;
 import org.eclipse.jetty.util.annotation.ManagedObject;
 
 @ManagedObject
-public class MultiplexConnectionPool extends AbstractConnectionPool implements ConnectionPool.Multiplexable
+public class MultiplexConnectionPool extends AbstractConnectionPool
 {
     public MultiplexConnectionPool(HttpDestination destination, int maxConnections, Callback requester, int maxMultiplex)
     {
@@ -54,12 +54,6 @@ public class MultiplexConnectionPool extends AbstractConnectionPool implements C
     public void setMaxMultiplex(int maxMultiplex)
     {
         super.setMaxMultiplex(maxMultiplex);
-    }
-
-    @Override
-    public void setMaxMultiplex(Connection connection, int maxMultiplex)
-    {
-        super.setMaxMultiplex(connection, maxMultiplex);
     }
 
     @Override
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/MultiplexConnectionPool.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/MultiplexConnectionPool.java
@@ -57,6 +57,12 @@ public class MultiplexConnectionPool extends AbstractConnectionPool implements C
     }
 
     @Override
+    public void setMaxMultiplex(Connection connection, int maxMultiplex)
+    {
+        super.setMaxMultiplex(connection, maxMultiplex);
+    }
+
+    @Override
     @ManagedAttribute(value = "The maximum amount of times a connection is used before it gets closed")
     public int getMaxUsageCount()
     {
--- a/jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Session.java
+++ b/jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Session.java
@@ -738,7 +738,10 @@ public abstract class HTTP2Session extends ContainerLifeCycle implements ISessio
             int maxCount = getMaxLocalStreams();
             if (maxCount >= 0 && localCount >= maxCount)
             {
-                promise.failed(new IllegalStateException("Max local stream count " + maxCount + " exceeded"));
+                IllegalStateException failure = new IllegalStateException("Max local stream count " + maxCount + " exceeded");
+                if (LOG.isDebugEnabled())
+                    LOG.debug("Could not create local stream #{} for {}", streamId, this, failure);
+                promise.failed(failure);
                 return null;
             }
             if (localStreamCount.compareAndSet(localCount, localCount + 1))
@@ -751,7 +754,7 @@ public abstract class HTTP2Session extends ContainerLifeCycle implements ISessio
             stream.setIdleTimeout(getStreamIdleTimeout());
             flowControl.onStreamCreated(stream);
             if (LOG.isDebugEnabled())
-                LOG.debug("Created local {}", stream);
+                LOG.debug("Created local {} for {}", stream, this);
             return stream;
         }
         else
@@ -786,6 +789,8 @@ public abstract class HTTP2Session extends ContainerLifeCycle implements ISessio
             int maxCount = getMaxRemoteStreams();
             if (maxCount >= 0 && remoteCount - remoteClosing >= maxCount)
             {
+                if (LOG.isDebugEnabled())
+                    LOG.debug("Could not create remote stream #{} for {}", streamId, this);
                 reset(null, new ResetFrame(streamId, ErrorCode.REFUSED_STREAM_ERROR.code), Callback.from(() -> onStreamDestroyed(streamId)));
                 return null;
             }
@@ -799,7 +804,7 @@ public abstract class HTTP2Session extends ContainerLifeCycle implements ISessio
             stream.setIdleTimeout(getStreamIdleTimeout());
             flowControl.onStreamCreated(stream);
             if (LOG.isDebugEnabled())
-                LOG.debug("Created remote {}", stream);
+                LOG.debug("Created remote {} for {}", stream, this);
             return stream;
         }
         else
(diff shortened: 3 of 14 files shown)


jetty-client/src/main/java/org/eclipse/jetty/client/MultiplexConnectionPool.java
```diff
@@ -56,6 +56,12 @@ public void setMaxMultiplex(int maxMultiplex)
         super.setMaxMultiplex(maxMultiplex);
     }
 
+    @Override
+    public void setMaxMultiplex(Connection connection, int maxMultiplex)
+    {
+        super.setMaxMultiplex(connection, maxMultiplex);
+    }
```
gregw: Not needed

jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Session.java
```diff
@@ -785,6 +788,8 @@ protected IStream createRemoteStream(int streamId)
             int maxCount = getMaxRemoteStreams();
             if (maxCount >= 0 && remoteCount - remoteClosing >= maxCount)
             {
+                if (LOG.isDebugEnabled())
+                    LOG.debug("Could not create remote stream #{} for {}", streamId, this);
```
gregw: If  you are keeping this debug, you may as well include `maxCount`, `remoteCount` & `remoteClosing`

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
```diff
@@ -169,6 +169,14 @@ public final void setMaxMultiplex(int maxMultiplex)
         if (maxMultiplex < 1)
             throw new IllegalArgumentException("Max multiplex must be >= 1");
         this.maxMultiplex = maxMultiplex;
+
+        try (Locker.Lock l = locker.lock())
+        {
+            if (closed)
+                return;
+
+            entries.forEach(entry -> entry.setMaxMultiplex(maxMultiplex));
```
gregw: Should this really set the max for each connection?    Perhaps it should be set to `Math.min(maxMultiplex, entry.getMaxMultiplex())`?

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
```diff
@@ -169,6 +169,14 @@ public final void setMaxMultiplex(int maxMultiplex)
         if (maxMultiplex < 1)
             throw new IllegalArgumentException("Max multiplex must be >= 1");
         this.maxMultiplex = maxMultiplex;
```
gregw: Should we rename the private field to make it clear it is the default:
```suggestion
        this.defaultMaxMultiplex = maxMultiplex;
```

Eitherway, this class now needs more javadoc describing how the maxMultiplex works

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
```diff
@@ -507,15 +515,28 @@ public String toString()
...
         // The pooled item.  This is not volatile as it is set once and then never changed.
         // Other threads accessing must check the state field above first, so a good before/after
         // relationship exists to make a memory barrier.
         private T pooled;
+        private volatile int maxMultiplex;
 
         Entry()
         {
             this.state = new AtomicBiInteger(Integer.MIN_VALUE, 0);
+            this.maxMultiplex = Pool.this.maxMultiplex;
```
gregw:
```suggestion
            this.maxMultiplex = Pool.this.defaultMaxMultiplex;
```

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
```diff
@@ -507,15 +515,28 @@ public String toString()
         // hi: positive=open/maxUsage counter; negative=closed; MIN_VALUE pending
         // lo: multiplexing counter
         private final AtomicBiInteger state;
-
         // The pooled item.  This is not volatile as it is set once and then never changed.
         // Other threads accessing must check the state field above first, so a good before/after
         // relationship exists to make a memory barrier.
         private T pooled;
+        private volatile int maxMultiplex;
```
gregw: I'm still wondering if we should store 2 values here:
```suggestion
        private volatile int remoteMaxMultiplex;
        private volatile int localMaxMultiplex;
```

We set localMaxMultiplex when we send a settings frame, we set remoteMaxMultiplex when we receive a settings frame.   If we change settings, local changes first and then remote changes when we get the ack, otherwise remote changes first and local changes second.

It will always be a hard error to have current greater than the localMaxMultiplex. 

If the localMaxMultiplex is less than the remoteMaxMultiplex then that doesn't matter, it is still and error have current greater than the local.

If the localMaxMultiplex is greater than the remoteMaxMultiplex then we have to be more nuanced.  It is an error for an remote created stream to exceed the remoteMaxMultiplex, but a locally created stream is allowed to let current exceed remoteMaxMultiplex but not localMaxMultiplex.

I think we really need to deal with this type of dynamic max streams adjustment. We have already seen one example in the wild... others will follow even if it is a bad idea.

gregw: Also this kind of exceeds my threshold for unused features in a base utility class.  Perhaps we need to split `Pool` into `Pool` and `MultiplexPool`?

gregw (changes requested):
I've lots of questions... but at the very least we need more javadoc in `Pool` class

jetty-client/src/main/java/org/eclipse/jetty/client/MultiplexConnectionPool.java
sbordet: It is needed. The `super` method is `protected`, this is `public`.

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
sbordet: We don't have the word "default" for a lot of other cases, e.g. `ServerConnector.idleTimeout` and `EndPoint.idleTimeout`, or the flow control windows got HTTP/2 `ConnectionFactory`s, so I would not add "default" here.
It is just the `maxMultiplex` of the `Pool`.

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
sbordet: > We set localMaxMultiplex when we send a settings frame

No, we don't. I don't think what you say is correct, although I understand the underlying idea of handling the case where a second incoming SETTINGS frame has a smaller `max_concurrent_streams` value.
However, I would handle this latter case in a different PR.

jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Session.java
sbordet: Done.

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
sbordet: I removed this.

It was there because the first request of a new connection is first reserving an entry, which would inherit the initial `maxMultiplex=1024`. However, when the connection is established, the SETTINGS frame with `max_concurrent_streams` sets `Pool.maxMultiplex`, but did not change that of the reserved entry, so I thought this was a good place to set it.
However, a better place to do it is in `AbstractConnectionPool.FutureConnection.succeeded()` when we `enable()` the reserved `Entry`.

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
gregw: Let's pair program on this tonight. I think it can be fixed in this PR.

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
gregw: But in other places we don't have the same fieldname on a subclass in the same file, but with a different meaning.

gregw: Also at the very least we need javadoc

jetty-client/src/main/java/org/eclipse/jetty/client/MultiplexConnectionPool.java
gregw: Hmmmmm kind of proves my point that we need to separate out `Pool` and `MultiplexedPool`.   It's not good to have "hidden" features on all pools that can be made public.   I'd rather cut and paste the Pool class and not have all our other pools carrying around unused multiplex complexity.

jetty-client/src/main/java/org/eclipse/jetty/client/ConnectionPool.java
```diff
@@ -109,13 +109,22 @@
...
          */
         void setMaxMultiplex(int maxMultiplex);
+
+        /**
+         * @param connection the multiplexed connection
+         * @param maxMultiplex the max number of requests multiplexable on the given connection
+         */
+        default void setMaxMultiplex(Connection connection, int maxMultiplex)
+        {
+            setMaxMultiplex(maxMultiplex);
```
gregw: This implementation is wrong.  Setting the max multiplex for a connection is not the same as setting the default for the pool

jetty-http2/http2-http-client-transport/src/main/java/org/eclipse/jetty/http2/client/http/HttpConnectionOverHTTP2.java
```diff
@@ -52,6 +53,7 @@
     private final AtomicInteger sweeps = new AtomicInteger();
     private final Session session;
     private boolean recycleHttpChannels = true;
+    private int maxMultiplex = 1;
```
gregw: This is duplicated state... which is almost always bad.  See below:

jetty-http2/http2-http-client-transport/src/main/java/org/eclipse/jetty/http2/client/http/HttpConnectionOverHTTP2.java
```diff
@@ -74,6 +76,16 @@ public void setRecycleHttpChannels(boolean recycleHttpChannels)
         this.recycleHttpChannels = recycleHttpChannels;
     }
 
+    public int getMaxMultiplex()
+    {
+        return maxMultiplex;
```
gregw: Avoid duplicate state:
```suggestion
        Object attachment = getAttachment();
        return (attachment instanceof ConnectionPool.Multiplexable)
            ? ((ConnectionPool.Multiplexable)attachment).getMaxMultiplex()
            : 0;
```

jetty-http2/http2-http-client-transport/src/main/java/org/eclipse/jetty/http2/client/http/HttpConnectionOverHTTP2.java
gregw: Argh!!! it id worse!!! This is triple state as it is also stored in [Http2Session](https://github.com/eclipse/jetty.project/blob/jetty-10.0.6/jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Session.java#L101-L102), but at least there is is kept correctly as the local and remote value.

gregw:
I think the fundamental problem here is that we are duplicating state.  The Http2Session correctly maintains the local and remote max streams, but we then duplicate this as a single value in the Pool abstraction.

I think we need to abstract the multiplexing support from the Pool so that it can access the state correctly held in Http2Session.

jetty-http2/http2-http-client-transport/src/main/java/org/eclipse/jetty/http2/client/http/HttpConnectionOverHTTP2.java
gregw: @sbordet I think this could at least remove some state duplication.... but might not be needed if we can delegate all the way down to HTTP2Session.

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
gregw: @sbordet  As we discussed on hangout, we don't need to fix all the problems in this PR, but we should not add wrong things to the Pool.   We should externalize the multiplexing checks as per my draft PR

lorban: @gregw I second this goal. I'd be very happy if we managed to externalize the multiplexing out of the low-level `Pool` as IMHO it was a design mistake.

That design also has adverse effects on performance as the delegation of the multiplexing accounting is delegated to a lower-level that the only thing it can do is to track it by using an atomic counter that's naturally going to be highly contended, i.e.: the HTTP2 pool is much slower when connections are multiplexed than when they're not because there's a ton of contention on the CAS loops around the multiplexing counters. If the multiplexing counter was tracked at the connection pool layer, we could add some contention-mitigating strategies.

sbordet:
@gregw, @lorban and I gave it a go at trying to move the multiplex feature out of `Pool` but that means that we need to move out the `Pool.StrategyType` too, and also provide (again) a way to access the `Pool` by index (as now the strategy of what `Pool.Entry` to access is driven from outside).

FTR, moving out `usageCount` would be trivial, and tracked by #6669.

I think the best solution would be to keep multiplexing in `Pool`, as the cost of moving it out is not compensating the advantages, which _may_ amount in just some less contention on an atomic counter (at the cost of dropping some other feature like the strategies), but very likely will result in the same contention, just elsewhere (if we want to keep the strategies).

About the thought bubble of having the multiplex check externalized, it may be a good idea, but A) it is not necessary to fix this particular issue, and B) I don't think it will solve the general problem anyway, as I can still see a race where a thread passes the check, but by the time it arrives to create a local stream representing the request to send, the value may have been changed with the arrival of a `SETTINGS` frame.

I opened #6670 about the decrement of `max_concurrent_streams`.

I think this PR is good as is, so I refreshed the review status.

gregw:
@sbordet Externalizing the multiplex check does work.   The race you describe definitely does exist, but that's the second issue that I said we did not want to fix in this PR.    At least if the multiplex check is externalized then that race can be fixed by putting all the logic/state around what is the max into a single state that is aware of settings received, sent and messages queued (ie the session).  

That race will never EVER be solved inside the pool so moving more multiplexing logic inside the pool is a step in the wrong direction.

As @lorban says, there is a cost in multiplexing support that we are having on all our pool usages.  So perhaps we should just split `Pool` into `Pool` and `MultiPool` (it would be nice if the later extended the former, but whilst I think that is possible, it is not necessary).     But even then, I would argue that the max streams logic should be external to the pool as per my proposed PR, as it is only in the Session class that the race of changing settings can be solved (but  not necessarily by this PR).

sbordet:
@gregw @lorban I have incorporate the changes from #6648 and rebased so the changes in this PR are now only fixing the original issue. Please re-review.

The only glitch is that it also deprecates `maxUsage`, which should be addressed in a different issue.
It can be easily reverted, although its deprecation is quite similar to that of `maxMultiplex`.

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
```diff
@@ -264,17 +347,26 @@ public Entry reserve()
...
             entries.add(entry);
             return entry;
         }
     }
 
+    protected Entry newEntry()
+    {
+        if (maxMultiplex >= 0 || maxUsage >= 0)
+            return new MultiEntry();
+        return new MonoEntry();
```
gregw, line 362: Having this as a dynamic decision does have a small cost (hopefully which the JIT can optimize out).    But it is necessary to maintain the current API.    Ultimately once we remove the deprecated methods from the base pool, then this decision can be made in a overriden method in the class that also provides the MultiEntry.    So I think this is a good compromise with the current API.

@lorban can you do some performance tests to see if the `NonoEntry` based pool is faster than the previous for things like `ByteBufferPool`.   I'm not expecting much, as it still hits an atomic int about as much as the previous atomic long... but it does avoid accessing the volatile max fields, so that could be a small win.

jetty-http2/http2-http-client-transport/src/main/java/org/eclipse/jetty/http2/client/http/HttpClientTransportOverHTTP2.java
```diff
@@ -60,7 +60,7 @@ public HttpClientTransportOverHTTP2(HTTP2Client client)
         setConnectionPoolFactory(destination ->
         {
             HttpClient httpClient = getHttpClient();
-            return new MultiplexConnectionPool(destination, httpClient.getMaxConnectionsPerDestination(), destination, httpClient.getMaxRequestsQueuedPerDestination());
+            return new MultiplexConnectionPool(destination, httpClient.getMaxConnectionsPerDestination(), destination, 1);
```
gregw, line 65: I'm assuming the 1 is there until overridden by the settings frame?   Eitherway a comment would be good explaining why 1.

gregw (commented):
LGTM, but then I had a hand in much of it... so I won't approve.

jetty-http2/http2-http-client-transport/src/main/java/org/eclipse/jetty/http2/client/http/HttpClientTransportOverHTTP2.java
sbordet: Done.

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
lorban: I can try benchmarking `Pool` again to make sure how this change affects perf, but I already have some concerns with this `newEntry` method.

Now that there can be multiple implementations of `Entry`, all `try*` calls have become virtual. They used to be monomorphic, this implementation makes them bi-morphic as there are by default 2 classes extending from `Entry`. But anyone extending `Pool` to override this method will make those `try*` methods mega-morphic.

Since `tryAcquire` is on the fast path (it's in the hot loop of `acquire`) this change introduces an easy way to silently harm performance as there can now be a mega-morphic call on the fast path. IIRC this is the reason why the pool's strategy was written the way it was because the original proposal was using multiple implementations and was measurably slower.

lorban (changes requested):
The `newEntry` method should be made `private` and commented to make sure we never mistakenly add a 3rd `Entry` implementation.

lorban (approved):

Issue #6595, opened 2021-08-10
Re-enable distribution tests for third parties webapp not migrated to Jakarta namespace


olamy:
**Jetty version(s)**
11.0.x

Tests disabled here: https://github.com/eclipse/jetty.project/pull/6589

PR #6644, opened 2021-08-19
[WIP] Issue #6595 dynamic javax classloader

olamy:
TODO mapping of file `META-INF/services/javax.servlet.ServletContainerInitializer`

### DIFF
--- a/jetty-javax-transformation/src/main/java/org/eclipse/jetty/transformer/JavaxWebappClassloader.java
+++ b/jetty-javax-transformation/src/main/java/org/eclipse/jetty/transformer/JavaxWebappClassloader.java
@@ -23,6 +23,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import org.eclipse.jetty.webapp.WebAppClassLoader;
 import org.objectweb.asm.ClassReader;
@@ -96,7 +97,7 @@ public class JavaxWebappClassloader extends WebAppClassLoader
 
                 return cw.toByteArray();
             }
-            catch (IOException e)
+            catch (Exception e)
             {
                 IllegalClassFormatException icfe = new IllegalClassFormatException(e.getMessage());
                 icfe.initCause(e);
@@ -136,9 +137,31 @@ public class JavaxWebappClassloader extends WebAppClassLoader
         @Override
         public String[] mapTypes(String[] internalNames)
         {
-            List<String> names = Arrays.asList(internalNames);
-            return super.mapTypes(internalNames);
+            List<String> names = Arrays.stream(internalNames).map(
+                    s ->
+                    {
+                        if (s.startsWith(JAVAX_SERVLET))
+                        {
+                            return JAKARTA_SERVLET + s.substring(JAVAX_SERVLET.length());
+                        }
+                        return s;
+                    }
+            ).collect(Collectors.toList());
+            return names.toArray(new String[0]);
         }
+
+        @Override
+        public String mapSignature(String signature, boolean typeSignature)
+        {
+            return super.mapSignature(signature, typeSignature);
+        }
+
+        @Override
+        public String mapPackageName(String name)
+        {
+            return super.mapPackageName(name);
+        }
+
     }
 
 }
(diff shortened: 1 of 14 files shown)


gregw:
interesting!

jetty-javax-transformation/src/main/java/org/eclipse/jetty/transformer/JavaxWebappClassloader.java
```diff
@@ -0,0 +1,92 @@
...
+            }
+        }
+    }
+
+    static class JavaxRemapper extends Remapper
+    {
+
+        private static final String JAVAX_SERVLET = "javax/servlet/";
+
+        private static final String JAKARTA_SERVLET = "jakarta/servlet/";
```
joakime, line 116: We'll need a more robust mapping eventually.

Here's all of the `javax.*` to `jakarta.*` imports in Jetty itself ... https://gist.github.com/joakime/88909e94d1be776a8bdecc0698fd55eb

joakime:
What happens to attributes?
Like the ones on `RequestDispatcher`, that we put in the `HttpServletRequest` attributes during an include/exclude?

olamy:
> What happens to attributes?
> Like the ones on `RequestDispatcher`, that we put in the `HttpServletRequest` attributes during an include/exclude?

Oh yes a lot to do!
Do not hesitate to add more ;)
The idea is to have comments on is it eventually a good way to do this and what is missing

jetty-javax-transformation/src/main/java/org/eclipse/jetty/transformer/JavaxWebappClassloader.java
joakime: Do we have to worry about version mapping?
Like, Javax Servlet 2.5 to Jakarta Servlet 5.0? (it will be missing many methods and whatnot)

It might make more sense to only go backwards (from most recent version of jakarta back to javax namespace)

gregw: Isn't that what this PR is attempting?   It is not giving a 5.0 request renamed to a 2.5 request to a 2.5 app, rather it is renaming the 2.5 app to be a 5.0 app.

gregw:
For request attributes, we could  try to write a special `Attributes` impl that will do the name conversion on the fly, so any getAttribute("javax.servlet.Foo") will actually get "jakarta.servlet.Foo".  But the problem here is that attributes holders are handled as a stack and this converter might be written over instead.   Thus we probably need to wrap the request itself and do the translation in the `getAttribute` method itself above the stack if `Attributes` maps.

Edit: alternately we could have a mode in the normal jetty-11 that supports the old names.

gregw:
With regards to "mapping of file META-INF/services/javax.servlet.ServletContainerInitializer", @lorban has already had a bit of a look at doing manipulation of that dynamically and I think it is very difficult.
Thus we might have to do that at deployment time... although as this is not a generic manipulation, we still may get away with a fixed translate on getResource APIs?

olamy:
my idea about Attributes is to modify some classes via ASM to intercept method such `setAttribute` to modify the `name` from `javax.servlet` to `jakarta.servlet`

gregw:
@olamy That amounts to the same thing.  But ASM would have to know to only modify the `ServletRequest.getAttribute(String)` method for some contexts, which could be tough as `ServletRequest` is a server class and not loaded by the context class loader.     Wrapping requests as they enter the translated context will probably be simpler.

olamy:
oh you mean something such `JavaxWebAppContext extends WebAppContext`? sounds like a good idea. 
it can make user configuration a bit more complicated but looks easier to implement.

joakime:
We should add tests for dynamic servlet / filter / listener registration added in Servlet 3.0 as well.

The following methods could prove ... uh ... "unique" ...

```
jakarta.servlet.ServletContext
  // listeners
  public void addListener(String className);
  public <T extends EventListener> void addListener(T t);
  public void addListener(Class<? extends EventListener> listenerClass);
  public <T extends EventListener> T createListener(Class<T> clazz) throws ServletException;

  // servlets
  public ServletRegistration.Dynamic addServlet(String servletName, String className);
  public ServletRegistration.Dynamic addServlet(String servletName, Servlet servlet);
  public ServletRegistration.Dynamic addServlet(String servletName, Class<? extends Servlet> servletClass);
  public <T extends Servlet> T createServlet(Class<T> clazz) throws ServletException;
  public ServletRegistration getServletRegistration(String servletName);
  public Map<String, ? extends ServletRegistration> getServletRegistrations();

  // jsp files
  public ServletRegistration.Dynamic addJspFile(String servletName, String jspFile);

  // filters
  public FilterRegistration.Dynamic addFilter(String filterName, String className);
  public FilterRegistration.Dynamic addFilter(String filterName, Filter filter);
  public FilterRegistration.Dynamic addFilter(String filterName, Class<? extends Filter> filterClass);
  public <T extends Filter> T createFilter(Class<T> clazz) throws ServletException;
  public FilterRegistration getFilterRegistration(String filterName);
  public Map<String, ? extends FilterRegistration> getFilterRegistrations();
```

PR #6648, opened 2021-08-20
Thought bubble on abstracting out multiplexing from the pool

gregw:
Abstract out multiplexing check from the pool, so a multiplexing pool can add a per connection check.
For Http2, this should use the existing state kept in HTTP2Session for local and remote maxSessions

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
@@ -68,7 +68,10 @@ public class Pool<T> implements AutoCloseable, Dumpable
     private final ThreadLocal<Entry> cache;
     private final AtomicInteger nextIndex;
     private volatile boolean closed;
+
+    @Deprecated
     private volatile int maxUsage = -1;
+    @Deprecated
     private volatile int maxMultiplex = -1;
 
     /**
@@ -175,6 +178,7 @@ public class Pool<T> implements AutoCloseable, Dumpable
      * @return the maximum number of entries
      */
     @ManagedAttribute("The maximum number of entries")
+    @Deprecated
     public int getMaxEntries()
     {
         return maxEntries;
@@ -184,11 +188,13 @@ public class Pool<T> implements AutoCloseable, Dumpable
      * @return the default maximum in-use count of entries
      */
     @ManagedAttribute("The default maximum multiplex count of entries")
+    @Deprecated
     public int getMaxMultiplex()
     {
         return maxMultiplex == -1 ? 1 : maxMultiplex;
     }
 
+    @Deprecated
     protected int getMaxMultiplex(T item)
     {
         return getMaxMultiplex();
@@ -199,6 +205,7 @@ public class Pool<T> implements AutoCloseable, Dumpable
      *
      * @param maxMultiplex the default maximum multiplex count of entries
      */
+    @Deprecated
     public final void setMaxMultiplex(int maxMultiplex)
     {
         if (maxMultiplex < 1)
@@ -222,11 +229,13 @@ public class Pool<T> implements AutoCloseable, Dumpable
      * @return the default maximum usage count of entries
      */
     @ManagedAttribute("The default maximum usage count of entries")
+    @Deprecated
     public int getMaxUsageCount()
     {
         return maxUsage;
     }
 
+    @Deprecated
     protected int getMaxUsageCount(T item)
     {
         return getMaxUsageCount();
@@ -239,6 +248,7 @@ public class Pool<T> implements AutoCloseable, Dumpable
      *
      * @param maxUsageCount the default maximum usage count of entries
      */
+    @Deprecated
     public final void setMaxUsageCount(int maxUsageCount)
     {
         if (maxUsageCount == 0)
@@ -648,6 +658,7 @@ public class Pool<T> implements AutoCloseable, Dumpable
             return false;
         }
 
+        @Deprecated
         public int getUsageCount()
         {
             return isIdle() ? 0 : 1;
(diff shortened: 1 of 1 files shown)


gregw:
Hmmm I don't think the remote/local maxStreams are correctly kept upto date with SETTINGs frames???

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
```diff
@@ -667,8 +672,8 @@ boolean tryAcquire()
                 int usageCount = AtomicBiInteger.getHi(encoded);
                 boolean closed = usageCount < 0;
                 int inUseCount = AtomicBiInteger.getLo(encoded);
-                int maxUsageCount = Pool.this.maxUsageCount;
-                if (closed || !checkInUse.apply(pooled, inUseCount) || (maxUsageCount > 0 && usageCount >= maxUsageCount))
+                int maxUsageCount = Pool.this.maxUsage;
+                if (closed || !available.test(pooled, inUseCount) || (maxUsageCount > 0 && usageCount >= maxUsageCount))
```
lorban: Ouch! `inUseCount` is going to go through `int -> Integer -> int` transformations just for this call to happen... on the fast path! My guts tell me this change alone is easily going to cost ~20% throughput.

If we only take perf into account, making `available` a simple `Predicate` would be enough as we could delegate the counter to the predicate itself. But I'm not sure it can be done without breaking backward compatibility.

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
```diff
@@ -548,34 +574,17 @@ public String toString()
      */
     public class Entry
     {
-        // hi: positive=open/maxUsage counter; negative=closed; MIN_VALUE pending
-        // lo: multiplexing counter
+        // hi: negative==closed else total-usage counter; negative=closed; MIN_VALUE pending
```
lorban: `negative==closed` was already documented on the same line.

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
```diff
@@ -270,7 +296,7 @@ public Entry reserve(int allotment)
             if (space <= 0)
                 return null;
 
-            if (allotment >= 0 && (getReservedCount() * getMaxMultiplex()) >= allotment)
+            if (allotment >= 0 && getReservedCount() >= allotment)
```
lorban: I would not touch the implementation of this method at all. We already had a discussion about the `allotment` parameter and concluded that it was a bad idea so we shifted away from this method that we only kept for backward compatibility. IMHO there's no point in modifying the implementation of something we know is fundamentally wrong at the high-level.

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
```diff
@@ -212,7 +238,7 @@ public final void setMaxMultiplex(int maxMultiplex)
     @ManagedAttribute("The default maximum usage count of entries")
     public int getMaxUsageCount()
     {
-        return maxUsageCount;
+        return maxUsage;
```
lorban: It's already been concluded that max usage should not have been implemented in the low-level `Pool` but rather in the higher level `ConnectionPool`, like what we did for the max-time-to-live. If you're considering deprecating the max multiplex, this other property should be deprecated too.

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
```diff
@@ -128,11 +131,32 @@ public Pool(StrategyType strategyType, int maxEntries)
...
+    public Pool(StrategyType strategyType, int maxEntries, boolean cache, BiPredicate<T, Integer> available)
     {
         this.maxEntries = maxEntries;
         this.strategyType = strategyType;
         this.cache = cache ? new ThreadLocal<>() : null;
         this.nextIndex = strategyType == StrategyType.ROUND_ROBIN ? new AtomicInteger() : null;
+        this.available = available == null ? this::isAvailable : available;
+    }
+
+    private boolean isAvailable(T item, Integer inUse)
```
lorban: I'd rather inline this method as it only is used as a lambda.

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
```diff
@@ -548,34 +574,17 @@ public String toString()
      */
     public class Entry
     {
-        // hi: positive=open/maxUsage counter; negative=closed; MIN_VALUE pending
-        // lo: multiplexing counter
+        // hi: negative==closed else total-usage counter; negative=closed; MIN_VALUE pending
+        // lo: in-use counter
         private final AtomicBiInteger state;
```
lorban: If we got rid of the two counters this `AtomicBiInteger` keeps track of, we could simplify it down to a tri-state `PENDING / OPEN / CLOSED` that would measurably boost performance of highly contended pools.

jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
gregw: So we need an interface that takes int rather than use BiPredicate

gregw:
@sbordet @lorban   I've checked in a big refactor (work in progress) of this throught bubble along the lines of a discussion I had with simone this morning.  Please read these notes before looking at the code.

+ The main change to `Pool` is that  the Entry is now extensible.   The idea being that for non-connection pool stuff we have a really simple single use Entry, but that connection pool can have it's own more complex entry that supports multiplexing and max usage.
+ However FOR NOW, I have left multiplexing ability in the util `Pool`, this is because of all the existing methods and because `ConnectionPool` has `Pool` in it's API.   So as a first step, the Pool has two `Entry` impls: `MultiUseEntry` and `SingleUseEntry` and if the setMaxXxxx methods have not been used, then it uses the `SingleUseEntry`
+ I've gone through the connection pools and where possible stopped them creating their own pool instance and instead mostly pass around the pool strategy they want to use.
+ The HTTP2 connection types extends the new `Pool.getMaxMultiplex(T item)` method to look up the max in the HTTP2Session, so per connection state is not duplicated.

I've run out of time, so I have no idea how close this is to working.

PR #6653, opened 2021-08-23
Fixes #6652 - Improve ReservedThreadExecutor dump.

### LINKED ISSUE #6652: Improve ReservedThreadExecutor dump
**Target Jetty version(s)**
9.4.x

**Enhancement Description**
`ReservedThreadExecutor.dump()` can yield un-intuitive output; for example with capacity 1, this may happen:

```
+= ReservedThreadExecutor@26c2e1dc{s=1/1,p=1} - STARTED
  |  +> reserved size=3
  |     +> ReservedThread@a6fccfc{RESERVED,thread=Thread[server-2188,5,main]}
  |     +> ReservedThread@645f60a4{PENDING,thread=null}
  |     +> ReservedThread@4647ec4a{RUNNING,thread=Thread[server-2191,5,main]}
```

which is un-intuitive because there are 3 threads when the capacity is 1.

The state of `ReservedThreadExecutor` is correct because there is only 1 thread that is `RESERVED`. 
One is `RUNNING` and the one `PENDING` will be checked against the capacity and returned to the pool -- it was "caught" by the dump in a transient moment.

sbordet:
Now adding/removing ReservedThread instances to _threads only in reservedWait().
In this way the window of time in which they can be caught for a dump() is shorter, producing more accurate dumps.

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
@@ -179,14 +179,19 @@ public class ReservedThreadExecutor extends AbstractLifeCycle implements TryExec
 
         super.doStop();
 
-        // Offer STOP task to all waiting reserved threads.
-        for (int i = _count.getAndSetLo(-1); i-- > 0;)
+        // Mark this instance as stopped.
+        int size = _count.getAndSetLo(-1);
+
+        // Offer the STOP task to all waiting reserved threads.
+        for (int i = 0; i < size; ++i)
         {
-            // yield to wait for any reserved threads that have incremented the size but not yet polled
+            // Yield to wait for any reserved threads that
+            // have incremented the size but not yet polled.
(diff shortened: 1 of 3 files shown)


jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
```diff
@@ -301,38 +305,50 @@ private Runnable reservedWait()
...
-                {
-                    LOG.ignore(e);
-                }
+                _state = State.STOPPED;
+                return STOP;
+            }
+            finally
+            {
+                // No longer a reserved thread.
+                _threads.remove(this);
```
gregw: I don't like this extra work here.  We want minimum time and effort between allocating a reserved thread and it acting on the task.
Making it modifyf the `_threads` collection here just wastes time.  See my alternative cleanup of dump in #6658

gregw (changes requested):

gregw (approved):

PR #6655, opened 2021-08-23
Issue #6654 - Fix NPEs from bad Servlet API use

### LINKED ISSUE #6654: ServerUpgradeRequest.getCookies() can throws NullPointerException
https://github.com/eclipse/jetty.project/blob/b2c420a124392b3e79a6abf263ec00f51e387409/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/ServerUpgradeRequest.java#L93

Hi,
request.getCookies() can be null and this code can throws a NullPointerException

joakime:

### DIFF
--- a/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/ServerUpgradeRequest.java
+++ b/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/ServerUpgradeRequest.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import java.util.TreeMap;
 import java.util.stream.Collectors;
 import javax.servlet.ServletRequest;
+import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpSession;
 
@@ -90,9 +91,13 @@ public class ServerUpgradeRequest
     {
         if (cookies == null)
         {
-            cookies = Arrays.stream(request.getCookies())
-                .map(c -> new HttpCookie(c.getName(), c.getValue()))
-                .collect(Collectors.toList());
+            Cookie[] reqCookies = request.getCookies();
+            if (reqCookies != null)
+            {
+                cookies = Arrays.stream(reqCookies)
+                    .map(c -> new HttpCookie(c.getName(), c.getValue()))
+                    .collect(Collectors.toList());
+            }
         }
 
         return cookies;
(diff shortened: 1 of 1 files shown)


jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/ServerUpgradeRequest.java
```diff
@@ -135,10 +140,13 @@ public int getHeaderInt(String name)
     {
         Map<String, List<String>> headers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
         Enumeration<String> headerNames = request.getHeaderNames();
-        while (headerNames.hasMoreElements())
+        if (headerNames != null)
```
joakime, line 147: The servlet spec indicates that this can also be null.

jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/ServerUpgradeRequest.java
```diff
@@ -90,9 +91,13 @@ public ServerUpgradeRequest(WebSocketNegotiation negotiation) throws BadMessageE
...
-            cookies = Arrays.stream(request.getCookies())
-                .map(c -> new HttpCookie(c.getName(), c.getValue()))
-                .collect(Collectors.toList());
+            Cookie[] reqCookies = request.getCookies();
+            if (reqCookies != null)
+            {
+                cookies = Arrays.stream(reqCookies)
+                    .map(c -> new HttpCookie(c.getName(), c.getValue()))
+                    .collect(Collectors.toList());
+            }
```
sbordet, line 100:
```suggestion
            }
            else
            {
                cookies = List.of();
            }
```

sbordet (changes requested):
I don't like returning `null` if there are no cookies, as it is prone to errors 😃

jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/ServerUpgradeRequest.java
joakime: done.

sbordet (dismissed):

jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/ServerUpgradeRequest.java
```diff
@@ -90,9 +91,17 @@ public ServerUpgradeRequest(WebSocketNegotiation negotiation) throws BadMessageE
...
+            else
+            {
+                cookies = List.of();
```
lachlan-roberts: Shouldn't we also return null here if the request had no cookies?

gregw: We had this discussion already recently and determined that `emptyList` was both more readable and more effficient.

```suggestion
                cookies = Collections.emptyList();
```

gregw (dismissed):
one tiny niggle....

joakime:
> one tiny niggle....

The call to `List.of()` is actually a call to `ImmutableCollections.emptyList()` (jvm package private class) already, which seems more correct than `Collections.emptyList()`.
All uses of `List.of(...)` result in immutable lists.

sbordet (approved):

PR #6657, opened 2021-08-24
Issue #6642 - WebSocket handling of Connection: upgrade,close.

### LINKED ISSUE #6642: WebSocket handling of Connection: upgrade,close.
**Jetty version(s)**
Jetty 10+

**Description**
In Jetty 9.4.x, there was a test for bad client `Connection` headers.

https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-websocket/websocket-server/src/test/java/org/eclipse/jetty/websocket/server/FirefoxTest.java

This concept needs to be restored in a new testcase on the Jetty WebSocket rewrite.

Two test cases need to be tested and ensured that they work.

The "Firefox" case. of `Connection: keep-alive, Upgrade`

```
GET /socket/ HTTP/1.1
Host: example.com:8080
Upgrade: websocket
Connection: keep-alive, Upgrade
Sec-WebSocket-Key: 9Y7WzchJwKR72PHdKrf7Mg==
Sec-WebSocket-Protocol: chat
Sec-WebSocket-Version: 13
```

And the "iPhone" and "Proxy/Load Balancer" case of `Connection: close, Upgrade`

```
GET /socket/ HTTP/1.1
Host: example.com:8080
Upgrade: websocket
Connection: close, Upgrade
Sec-WebSocket-Key: 9Y7WzchJwKR72PHdKrf7Mg==
Sec-WebSocket-Protocol: chat
Sec-WebSocket-Version: 13
```

lachlan-roberts:
## Closes #6642 

add tests for WebSocket Connection Headers.

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
@@ -344,6 +344,7 @@ public class HttpGenerator
             _endOfContent = EndOfContent.UNKNOWN_CONTENT;
             return Result.FLUSH;
         }
+        _state = State.END;
 
         // If this is an upgrade then we don't want to close the connection.
         if (_info.isResponse() && ((MetaData.Response)_info).getStatus() == HttpStatus.SWITCHING_PROTOCOLS_101)
@@ -356,8 +357,6 @@ public class HttpGenerator
             if (connectionHeader != null && connectionHeader.contains(HttpHeaderValue.UPGRADE.asString()))
                 return Result.DONE;
         }
-
-        _state = State.END;
         return Boolean.TRUE.equals(_persistent) ? Result.DONE : Result.SHUTDOWN_OUT;
     }
 
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
@@ -345,6 +345,18 @@ public class HttpGenerator
             return Result.FLUSH;
         }
 
+        // If this is an upgrade then we don't want to close the connection.
+        if (_info.isResponse() && ((MetaData.Response)_info).getStatus() == HttpStatus.SWITCHING_PROTOCOLS_101)
+        {
+            return Result.DONE;
+        }
+        else if (_info.isRequest())
+        {
+            HttpField connectionHeader = _info.getFields().getField(HttpHeader.CONNECTION);
+            if (connectionHeader != null && connectionHeader.contains(HttpHeaderValue.UPGRADE.asString()))
+                return Result.DONE;
+        }
+
         _state = State.END;
         return Boolean.TRUE.equals(_persistent) ? Result.DONE : Result.SHUTDOWN_OUT;
     }
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
@@ -345,15 +345,6 @@ public class HttpGenerator
             return Result.FLUSH;
         }
         _state = State.END;
-
-        // If this is a request, don't close the connection until the server responds.
-        if (_info.isRequest())
-            return Result.DONE;
-
-        // If successfully upgraded it is responsibility of the next protocol to close the connection.
-        if (_info.isResponse() && ((MetaData.Response)_info).getStatus() == HttpStatus.SWITCHING_PROTOCOLS_101)
-            return Result.DONE;
-
         return Boolean.TRUE.equals(_persistent) ? Result.DONE : Result.SHUTDOWN_OUT;
     }
 
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
@@ -346,17 +346,14 @@ public class HttpGenerator
         }
         _state = State.END;
 
-        // If this is an upgrade then we don't want to close the connection.
-        if (_info.isResponse() && ((MetaData.Response)_info).getStatus() == HttpStatus.SWITCHING_PROTOCOLS_101)
-        {
+        // If this is a request, don't close the connection until the server responds.
+        if (_info.isRequest())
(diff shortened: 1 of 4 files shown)


joakime:
Here's one websocket lib making the same discovery we just did with the "close" token.

https://github.com/warmcat/libwebsockets/issues/1435#issuecomment-511244454

lachlan-roberts:
@joakime @gregw I have added back the test case for `Connection: close, Upgrade` and have added some code in `HttpGenerator` which prevents the connection from being closed.

joakime:
Should we offer a configuration for websocket to satisfy behaviors of bad websocket clients?

What if the naive websocket client sends a normal `Connection: upgrade`, but the upgrade doesn't happen.
We respond with a normal HTTP response, and now the connection is kept open on the server side.

I wonder if we should have a configuration that can force a `Connection: close` on the response of bad upgrades to help poor quality websocket clients?
Is this even a concern?

joakime:
Looks like `org.eclipse.jetty.http.HttpGeneratorServerTest.testResponseUpgrade` is failing due to this change.

jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
```diff
@@ -344,8 +344,19 @@ private Result completing(ByteBuffer chunk, ByteBuffer content)
...
+        if (_info.isResponse() && ((MetaData.Response)_info).getStatus() == HttpStatus.SWITCHING_PROTOCOLS_101)
+        {
+            return Result.DONE;
+        }
+        else if (_info.isRequest())
+        {
+            HttpField connectionHeader = _info.getFields().getField(HttpHeader.CONNECTION);
+            if (connectionHeader != null && connectionHeader.contains(HttpHeaderValue.UPGRADE.asString()))
+                return Result.DONE;
+        }
```
gregw: I'm a bit worried about this case.  Just because you send an upgrade doesn't mean that it will be accepted and that we can avoid shutting down the output.   I think we have to ponder this one a bit more.

lachlan-roberts: I don't think we really even need this case because our `WebSocketClient` will never send a `Connection: close, Upgrade` header. But it makes the test easier to write because we can use the `WebSocketClient`.

But also because we never send the `close, Upgrade` header this code will never return here, so I don't think it is much of an issue. But I could try re-write the test without the `WebSocketClient` if you want.

gregw: I think leave out this if branch.   getField is a linear search that will iterate over all fields if it is not there, so it is a waste of CPU to do on every single request.

sorry if that makes the testing a bit more difficult.

gregw (changes requested):

jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
```diff
@@ -344,8 +344,19 @@ private Result completing(ByteBuffer chunk, ByteBuffer content)
...
         return Boolean.TRUE.equals(_persistent) ? Result.DONE : Result.SHUTDOWN_OUT;
```
gregw, line 348: @lachlan-roberts  As discussed on the hangout,  I'm now thinking that for generating requests, it is a bit of a cheat to `shutdown_out` as soon as the request is sent.  The RFC says the output should be shutdown after the response is received.
So we could change here to only return DONE for request and then add logic to the clients response completion to check for a non persistent request and a non-101 response so it could do the shutdown.

Another option is to keep this code as is and send the shutdown for requests, but in HttpSenderOverHTTP just remember the shutdown rather that doing it immediately and then apply it once a non 101 response is complete.

Need to discuss this with @sbordet

gregw: @lachlan-roberts have a look at HttpReceiverOverHTTP.  There is already a check for 101 response in an `if(complete)`.  So if not a 101 we could very simply do a  `this.getHttpChannel().getHttpSender()` and query the sender to see if is not persistent and then shutdown the output here.
It can either ask the sender if it is persistent, or we could do the change above to make it just remember the shutdown rather than actually doing the shutdown.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -885,8 +886,10 @@ private void releaseChunk()
         @Override
         protected void onCompleteSuccess()
         {
+            boolean upgrading = _info != null && _info.getStatus() == HttpStatus.SWITCHING_PROTOCOLS_101;
```
gregw: I'm not sure this is enough now.  Don't we have to also deal with the 200 OK for `CONNECT` requests?   See below:

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -885,8 +886,10 @@ private void releaseChunk()
...
             release().succeeded();
-            if (_shutdownOut)
+            // If successfully upgraded it is responsibility of the next protocol to close the connection.
+            if (_shutdownOut && !upgrading)
```
gregw, line 891: Perhaps this covers both cases:
```suggestion
            // If successfully upgraded it is responsibility of the next protocol to close the connection.
            if (_shutdownOut && _channel.getRequest().getAttribute(UPGRADE_CONNECTION_ATTRIBUTE) == null)
)
```

gregw (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
lachlan-roberts: That wasn't really in the scope of this PR, @sbordet said he was going to build on this in a subsequent PR to get the `CONNECT` working.

gregw: Perhaps, but the change I suggest below should also work for the purposes of this PR and at least has a chance of being right for a subsequent `CONNECT` PR.    I'm not saying add tests for `CONNECT` in this PR, as you can leave that to @sbordet's subsequent PR.

gregw (approved):

PR #6663, opened 2021-08-24
Issue #6661 - Make ServerConnector socket options setting optional

### LINKED ISSUE #6661: Some SocketOptions not supported on Windows
**Jetty version(s)**
10.0.x

11.0.12

**OS type/version**
Windows 10

**Description**

The recent work on Issue #6372 is not compatible with Windows 10.

```
Caused by: java.lang.UnsupportedOperationException: 'SO_REUSEPORT' not supported
	at java.base/sun.nio.ch.ServerSocketChannelImpl.setOption(ServerSocketChannelImpl.java:151)
	at org.eclipse.jetty.server.ServerConnector.openAcceptChannel(ServerConnector.java:338)
```

joakime:
Closes #6661

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/ServerConnector.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/ServerConnector.java
@@ -19,6 +19,7 @@ import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 import java.net.Socket;
 import java.net.SocketException;
+import java.net.SocketOption;
 import java.net.StandardSocketOptions;
 import java.nio.channels.Channel;
 import java.nio.channels.SelectableChannel;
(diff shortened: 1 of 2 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/ServerConnector.java
```diff
@@ -348,6 +349,18 @@ protected ServerSocketChannel openAcceptChannel() throws IOException
...
 
+    private <T> void setSocketOption(ServerSocketChannel channel, SocketOption<T> name, T value)
+    {
+        try
+        {
+            channel.setOption(name, value);
+        }
+        catch (Throwable t)
+        {
+            LOG.warn("Unable to set socket option {} to {}", name, value, t);
```
sbordet: Don't want to warn here, see `ClientConnector.setSocketOption(...)`, just use the same debug that's there.

sbordet (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/ServerConnector.java
joakime: Updated.

sbordet (approved):

joakime:
Going to wait for green from CI before merging.

PR #6674, opened 2021-08-27
Jetty 10.0.x - dependency updates

joakime:


joakime:
I reverted the infinispan version updates here, as it more complicated then it seems.
Opened Issue #6687 to address that in a different PR.

Issue #6601, opened 2021-08-11
jetty-websocket-core not usable standalone, only with websocket-javax-server or websocket-jetty-server


martinmeeser:
**Jetty version(s)**
Jetty 10.x, Jetty 11.x

**Java version/vendor** `(use: java -version)`
not java version dependent
 
**OS type/version**
not os related

**Description**
I am using an embedded jetty. I am NOT using ServletContext handler.
I was using Jetty 10 beta.

I directly used a WebsocketUpgradeHandler from jetty-websocket-core, basically following the examples from the tests (ChatServer).
In the non-betas, this is not  possible anymore.

The reason is the following:
In class
WebSocketExtensionRegistry lines 82-90:
```
        try
        {
            Extension ext = components.getObjectFactory().createInstance(extClass);
            ext.init(config, components);

            return ext;
        }
        catch (Throwable t)
        {
            throw new BadMessageException("Cannot instantiate extension: " + extClass, t);
        }
```
This will lead to
``T o = clazz.getDeclaredConstructor().newInstance();
in in utils object factory.

Now this leads to an exception. I am assuming this is because the class is not exported in module-info.java (only to the websocket-javax-server or websocket-jetty-server).

So there is negotiation going on, the deflate extension is found, but it can not be instantiated, because the module system prevents it

The test should run fine since they can access the internal classes.
But any other usage of this module is not possible.

**How to reproduce?**
Use WebsocketUpgradeHandler in your own module.

- Exporting all classes from the internal packages will solve this, please do

joakime:
Just to be clear, are you sure you are using `jetty-websocket-core`, which doesn't exist in Jetty 10+?

All of the websocket artifacts have the `websocket-<impl>-<feature>` name.
Where `<impl>` is one of: `core`, `javax`, `jakarta`, or `jetty`.
and `<feature>` is one of: `common`, `client`, `server`, `api`, `test`

The `WebSocketUpgradeHandler` is part of the `websocket-core-server` artifact.

The `WebSocketExtensionRegistry` does exist, and it's part of the `websocket-core-common` artifact.

The requirement on `components.getObjectFactory().createIntance(extClass)` mean that the Extension you have registered must have a public no-args constructor and return an implementation of the `Extension` class.

If you decide to use this, along with JPMS, then it's up to you to declare your usages in your own `module-info.java`.

if you declared a new extension like `org.meeser.app.MeeserExtension`, then your entries would be something like this ...

``` java
module org.meeser.app {
[...]

PR #6680, opened 2021-08-30
Issue #6601 - fix JPMS issues for websocket-core-common

lachlan-roberts:
## Issue #6601

- Add module to test websocket-core with JPMS
- export websocket-core-common to jetty-util for the `DecoratedObjectFactory`


tests/test-jpms/test-websocket-core/src/test/java/WebSocketCoreJpmsTest.java
```diff
@@ -0,0 +1,72 @@
...
+import org.eclipse.jetty.websocket.core.client.CoreClientUpgradeRequest;
+import org.eclipse.jetty.websocket.core.client.WebSocketCoreClient;
+import org.eclipse.jetty.websocket.core.server.WebSocketNegotiator;
+import org.eclipse.jetty.websocket.core.server.WebSocketUpgradeHandler;
+import org.example.websocket.MyFrameHandler;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class WebSocketCoreJpmsTest
```
sbordet: I prefer JPMS all upper case.

tests/test-jpms/test-websocket-core/src/main/java/org/example/websocket/MyFrameHandler.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.eclipse.jetty.util.Callback;
+import org.eclipse.jetty.websocket.core.CloseStatus;
+import org.eclipse.jetty.websocket.core.CoreSession;
+import org.eclipse.jetty.websocket.core.Frame;
+import org.eclipse.jetty.websocket.core.FrameHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class MyFrameHandler implements FrameHandler
```
sbordet: Can this class be moved to `src/test/java`?
If possible, I would leave `src/main/java` empty, apart for the `module-info.java`.

tests/test-jpms/test-websocket-core/src/main/java/module-info.java
```diff
@@ -0,0 +1,20 @@
...
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+module org.eclipse.jetty.websocket.core.tests
```
sbordet: Please add javadoc explaining that this module descriptor is there so that the tests are run in JPMS mode, therefore testing the JPMS module descriptors of the dependencies involved.

sbordet (changes requested):
Typical Maven structure for child modules is to repeat the parent directory name, or drop it entirely, so:

```
test-jpms
`- test-jpms-websocket-core
```

Right now the child directory is `test-websocket-core` so I would add `jpms`.

Also make sure you have a newline at EOF.

tests/test-jpms/test-websocket-core/src/test/java/WebSocketCoreJpmsTest.java
gregw: Says the man who created `HttpReceiverOverHTTP` :)
Our coding standards don't  say anything one way or the other about such cases....  but searching for `Jpms` and `JPMS` shows that the later is used more often, so we should be consistant.

sbordet (approved):

Issue #6100, opened 2021-03-25
Fix Jetty-10 failing tests on Windows.


lachlan-roberts:
**Jetty version**
10.0.x

**Description**
Multiple tests are failing when the jetty-10.0.x tests are run on windows, these should be reviewed and fixed.

lachlan-roberts:
@joakime do you know if this can be closed after the #6581 PR?

lachlan-roberts:
Looks like the windows run on jetty-10.0.x seems to be passing on jenkins.
https://jenkins.webtide.net/job/nightlies/job/jetty-10.0.x-windows-nightly/

Although there is some failures for jdk-17
https://jenkins.webtide.net/job/nightlies/job/jetty-10.0.x-windows-nightly-jdk17/

PR #6681, opened 2021-08-30
Jetty 10.0.x 6497 alias checkers alt

### LINKED ISSUE #6497: Replace SameFileAliasChecker

gregw:
An alternative fix for #6497 that avoids using string prefix comparison and instead uses explicit File and Path methods

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
@@ -21,6 +21,8 @@ import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
 
 import org.eclipse.jetty.server.handler.ContextHandler;
 import org.eclipse.jetty.util.component.AbstractLifeCycle;
@@ -100,6 +102,7 @@ public class AllowedResourceAliasChecker extends AbstractLifeCycle implements Co
             Path link = path.toRealPath(NO_FOLLOW_LINKS);
             Path real = path.toRealPath(FOLLOW_LINKS);
 
+            // Allowed IFF the real file is allowed and either it is not linked or the link file itself is also allowed.
             return isAllowed(real) && (link.equals(real) || isAllowed(link));
         }
         catch (Throwable t)
@@ -114,19 +117,22 @@ public class AllowedResourceAliasChecker extends AbstractLifeCycle implements Co
         // If the resource doesn't exist we cannot determine whether it is protected so we assume it is.
         if (Files.exists(path))
         {
+            // resolve the protected paths now, as they may have changed since start
+            List<Path> protect = _protected.stream()
+                .map(AllowedResourceAliasChecker::getRealPath)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());
+
+            // Walk the path parent links looking for the base resource, but failing if any steps are protected
             while (path != null)
             {
                 if (Files.isSameFile(path, _base))
                     return true;
 
-                for (Path protectedPath : _protected)
+                for (Path p : protect)
                 {
-                    if (Files.exists(protectedPath))
-                    {
-                        protectedPath = protectedPath.toRealPath(FOLLOW_LINKS);
-                        if (Files.exists(protectedPath) && Files.isSameFile(path, protectedPath))
-                            return false;
-                    }
+                    if (Files.isSameFile(path, p))
+                        return false;
                 }
 
                 path = path.getParent();
@@ -134,6 +140,24 @@ public class AllowedResourceAliasChecker extends AbstractLifeCycle implements Co
         }
 
         return false;
+    }
+
+    private static Path getRealPath(Path path)
+    {
+        if (path == null || !Files.exists(path))
+            return null;
+        try
+        {
+            path = path.toRealPath(FOLLOW_LINKS);
+            if (Files.exists(path))
+                return path;
+        }
+        catch (IOException e)
+        {
+            if (LOG.isDebugEnabled())
+                LOG.debug("No real path for {}", path, e);
+        }
+        return null;
     }
 
     protected Path getPath(Resource resource)
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
@@ -13,7 +13,6 @@
 
 package org.eclipse.jetty.server;
 
-import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.LinkOption;
@@ -89,10 +88,6 @@ public class AllowedResourceAliasChecker extends AbstractLifeCycle implements Co
     {
         try
         {
-            // do not allow any file separation characters in the URI
-            if (File.separatorChar != '/' && pathInContext.indexOf(File.separatorChar) >= 0)
-                return false;
-
             // The existence check resolves the symlinks.
             if (!resource.exists())
                 return false;
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
@@ -85,7 +85,7 @@ public class AllowedResourceAliasChecker extends AbstractLifeCycle implements Co
     }
 
     @Override
-    public boolean check(String uri, Resource resource)
+    public boolean check(String pathInContext, Resource resource)
     {
         // The existence check resolves the symlinks.
         if (!resource.exists())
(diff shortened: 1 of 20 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
```diff
@@ -40,172 +40,103 @@
...
-                return false;
+            Path link = path.toRealPath(NO_FOLLOW_LINKS);
+            Path real = path.toRealPath(FOLLOW_LINKS);
 
-            if (_checkSymlinkTargets && hasSymbolicLink(resourcePath))
-            {
-                if (isProtectedPath(resourcePath, FOLLOW_LINKS))
-                    return false;
-            }
+            return isAllowed(real) && (link.equals(real) || isAllowed(link));
```
lachlan-roberts: Its logically the same but it should read the other way around.
```java
isAllowed(link) && (link.equals(real) || isAllowed(real))
```

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
```diff
@@ -40,172 +40,103 @@
...
+        Path path = getPath(resource);
+        if (path == null)
             return false;
 
         try
         {
-            if (isProtectedPath(resourcePath, NO_FOLLOW_LINKS))
-                return false;
+            Path link = path.toRealPath(NO_FOLLOW_LINKS);
+            Path real = path.toRealPath(FOLLOW_LINKS);
```
lachlan-roberts: If the target of the symlink does not exist then this will throw.
Which is probably fine but then you are always warning, is this what we want to do?

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
```diff
@@ -40,172 +40,103 @@
...
                     return true;
 
-                // Check that the target prefix really is a path segment.
-                if (target.charAt(protect.length()) == File.separatorChar)
-                    return true;
+                for (Path protectedPath : _protected)
+                {
+                    if (Files.exists(protectedPath))
+                    {
+                        protectedPath = protectedPath.toRealPath(FOLLOW_LINKS);
```
lachlan-roberts: What is the point of this line, if you always wanted to use the RealPath for this then you could do this in `doStart`, or if you assume that the base dir is never symlinked then you wouldn't do it at all.

But either way, doing this breaks the use case of the symlinked base dir from an embedded configuration.

jetty-server/src/main/java/org/eclipse/jetty/server/SymlinkAllowedResourceAliasChecker.java
```diff
@@ -34,33 +33,42 @@
...
+
+        Path path = getPath(resource);
+        if (path == null)
+            return false;
 
-            // Approve if path is a symbolic link.
-            Path resourcePath = resource.getFile().toPath();
-            if (Files.isSymbolicLink(resourcePath))
-                return true;
+        String[] segments = pathInContext.split("/");
```
lachlan-roberts: We really can't do this. This is exactly what we're trying to change the alias checkers to avoid. Use the Path or File APIs. Why not do a loop similar to what is done in `AllowedResourceAliasChecker` with `Path.getParent()`?

jetty-server/src/main/java/org/eclipse/jetty/server/SymlinkAllowedResourceAliasChecker.java
```diff
@@ -34,33 +33,42 @@
...
+        {
+            for (String segment : segments)
+            {
+                fromBase = fromBase.resolve(segment);
+                if (!Files.exists(fromBase))
+                    return false;
+                if (Files.isSymbolicLink(fromBase))
+                    return true;
+                if (!isAllowed(fromBase))
+                    return false;
```
lachlan-roberts, line 70: What is the point of the `isAllowed()` check here? Why is it after the `isSymbolicLink()` check where we would return true?

lachlan-roberts (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
gregw: I'm a bit dubious we want to warn below anyway.   I'm inclined to change that to a debug.
Eitherway, if they have a symlink that points nowhere, then I'm OK with throwing here.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
gregw: I don't think so.  The real always has to be allowed.   For symlinks we have the additional condition that the link itself also has to be allowed.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
gregw: I'm concerned that META-INF and/or WEB-INF directories could be added after start... unlikely but possible.     So I resolve the protected targets here.

I don't think it does break the case of a symlinked base dir in embedded.   We resolve both the base and the resource using FOLLOW_LINKS, then for this checker no more symlinks are allowed.  So the request resource must be within the base resource.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
```diff
@@ -223,6 +154,10 @@ private Path getPath(Resource resource)
...
     public String toString()
     {
-        return String.format("%s@%x{checkSymlinkTargets=%s}", AllowedResourceAliasChecker.class.getSimpleName(), hashCode(), _checkSymlinkTargets);
+        return String.format("%s@%x{base=%s,protected=%s}",
+            this.getClass().getSimpleName(),
+            hashCode(),
+            _base,
+            Arrays.asList(_contextHandler.getProtectedTargets()));
     }
 }
```
gregw: eol

jetty-server/src/main/java/org/eclipse/jetty/server/SymlinkAllowedResourceAliasChecker.java
gregw: Because we need to walk down from the top, not up from the bottom.
Why can't we split the path on '/'? We know it is a pathInContext that is always separated by '/' and there is no escape for it (because of the stupid decoding done converting URI to a path).      What is the case that this doesn't work?

jetty-server/src/main/java/org/eclipse/jetty/server/SymlinkAllowedResourceAliasChecker.java
gregw: We are walking down from the top looking for a symbolic link, which we will allow.  But anything about that must be allowed.    This is to protect against the "wEb-InF/someSymLinkNotForPublicConsumption.xml", which without this check would be allowed just because it is a symlink.  With this test, it is not allowed because the first segment "wEb-InF" is a protected resource.

gregw:
@lachlan-roberts let me add some comments and javadoc to clarify my answers above....

gregw:
@lachlan-roberts new version pushed.     I'm now handling symlinks by constructing the real URI to them and checking that with `ContextHandler.isProtectedTarget(String)`.

gregw:
@joakime I think @lachlan-roberts and I have closed our positions somewhat on this one.   Would be good to get your review now.

lachlan-roberts (dismissed):

gregw:
I've made this a direct PR to 10.0.x to make reviewing simpler for others.  This has changes from both @lachlan-roberts and myself in it

gregw:
@joakime @lachlan-roberts nudge

lachlan-roberts (approved):

gregw:
@joakime your thoughts?

jetty-server/src/main/java/org/eclipse/jetty/server/SymlinkAllowedResourceAliasChecker.java
```diff
@@ -0,0 +1,86 @@
...
+    public SymlinkAllowedResourceAliasChecker(ContextHandler contextHandler)
+    {
+        super(contextHandler);
+    }
+
+    @Override
+    protected boolean check(String pathInContext, Path path)
+    {
+        // do not allow any file separation characters in the URI, as we need to know exactly what are the segments
+        if (File.separatorChar != '/' && pathInContext.indexOf(File.separatorChar) >= 0)
```
lachlan-roberts, line 44: shouldn't this be `pathInContext.indexOf(File.separatorChar) != -1`?

gregw: isn't that the same thing?

gregw:
@joakime regardless of if  this has been run on windows, do you have any comments on inspecting the code?  Is this approach what you were advocating about using the `Path` class and associates more?

lachlan-roberts:
The AliasChecker tests passed when I ran them on windows.

I ran a full build with tests and had a whole bunch of failures which seem to be unrelated. I've never had a windows build pass all tests locally. But I will take a more detailed look at the windows failures soon for issue #6100.

gregw:
Ok, I'll merge tonight unless I hear otherwise.

PR #6692, opened 2021-09-02
Update HttpInput puml diagrams

lorban:
Close the gap between the diagrams and the actual code.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannelState.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannelState.java
@@ -1286,6 +1286,10 @@ public class HttpChannelState
         return woken;
     }
 
+    /**
+     * Called to indicate that some content was produced and is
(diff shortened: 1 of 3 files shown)


gregw (changes requested):
Looks good.... but one thing I found surprising was the `onReadIdle` call as that name didn't really make sense.     I looked at the actual method, but there is no javadoc on it.  Can you add a little javadoc in this PR as well.

lorban:
@gregw I looked at `onContentAdded`, `onReadUnready` and `onReadIdle` and I find the three method names to be unclear.

`onContentProduced`, `onContentDemanded` and `onContentConsume` sound better to me as their really reflect what the FSM (i.e.: the ContentProducer) is doing rather than what the caller (i.e.: HttpInput) wants to do.

I kept the existing method names, but I've changed their javadoc to reflect the above: produced, demanded and consuming.

gregw (approved):

PR #6694, opened 2021-09-02
Fixes #6693 - FastCGI review

### LINKED ISSUE #6693: FastCGI review
**Target Jetty version(s)**
10.0.x

**Enhancement Description**
In Jetty 10, FastCGI has been minimally updated to drop the support for multiplexing (#2095), as the protocol itself does not support it right, and in practice this feature is never used.

However, a lot of code that was handling multiplexing was left there. 
The client is more inefficient and the server produces race conditions such as those reported in e.g. #6217.

sbordet:
- Removed code supporting multiplex in the client.
- Removed code supporting multiplex in the server.
- Reworked the server-side processing of a request, now more similar to HTTP/1.1.
- Improved javadocs.

### DIFF
--- a/jetty-fcgi/fcgi-client/src/main/java/org/eclipse/jetty/fcgi/parser/Parser.java
+++ b/jetty-fcgi/fcgi-client/src/main/java/org/eclipse/jetty/fcgi/parser/Parser.java
@@ -106,11 +106,8 @@ public abstract class Parser
                             // parsing; the async operation will eventually resume parsing.
                             return true;
                         }
-                        else
-                        {
-                            padding = headerParser.getPaddingLength();
-                            state = State.PADDING;
-                        }
+                        padding = headerParser.getPaddingLength();
+                        state = State.PADDING;
                     }
                     break;
                 }
--- a/jetty-fcgi/fcgi-client/src/main/java/org/eclipse/jetty/fcgi/parser/Parser.java
+++ b/jetty-fcgi/fcgi-client/src/main/java/org/eclipse/jetty/fcgi/parser/Parser.java
@@ -36,6 +36,20 @@ import org.slf4j.LoggerFactory;
  * </pre>
  * <p>Depending on the {@code type}, the content may have a different format,
  * so there are specialized content parsers.</p>
+ * <p>A typical exchange is:</p>
+ * <pre>
+ * BEGIN_REQUEST
+ * PARAMS (length > 0)
+ * PARAMS (length == 0 to signal end of PARAMS frames)
+ * [STDIN (length > 0 in case of request content)]
+ * STDIN (length == 0 to signal end of STDIN frames and end of request)
+ * ...
+ * STDOUT (length > 0 with HTTP headers and HTTP content)
+ * STDOUT (length == 0 to signal end of STDOUT frames)
+ * [STDERR (length > 0)]
+ * [STDERR (length == 0 to signal end of STDERR frames)]
+ * END_REQUEST
+ * </pre>
  *
  * @see HeaderParser
  * @see ContentParser
@@ -70,7 +84,10 @@ public abstract class Parser
                     ContentParser contentParser = findContentParser(headerParser.getFrameType());
                     if (headerParser.getContentLength() == 0)
                     {
-                        contentParser.noContent();
+                        padding = headerParser.getPaddingLength();
+                        state = State.PADDING;
+                        if (contentParser.noContent())
+                            return true;
                     }
                     else
                     {
@@ -89,9 +106,12 @@ public abstract class Parser
                             // parsing; the async operation will eventually resume parsing.
                             return true;
                         }
+                        else
+                        {
+                            padding = headerParser.getPaddingLength();
+                            state = State.PADDING;
+                        }
                     }
-                    padding = headerParser.getPaddingLength();
-                    state = State.PADDING;
                     break;
                 }
                 case PADDING:
@@ -176,7 +196,6 @@ public abstract class Parser
             @Override
             public void onFailure(int request, Throwable failure)
             {
-
             }
         }
     }
--- a/jetty-fcgi/fcgi-client/src/main/java/org/eclipse/jetty/fcgi/parser/Parser.java
+++ b/jetty-fcgi/fcgi-client/src/main/java/org/eclipse/jetty/fcgi/parser/Parser.java
@@ -39,14 +39,14 @@ import org.slf4j.LoggerFactory;
  * <p>A typical exchange is:</p>
  * <pre>
  * BEGIN_REQUEST
- * PARAMS (length > 0)
+ * PARAMS (length &gt; 0)
  * PARAMS (length == 0 to signal end of PARAMS frames)
- * [STDIN (length > 0 in case of request content)]
+ * [STDIN (length &gt; 0 in case of request content)]
  * STDIN (length == 0 to signal end of STDIN frames and end of request)
  * ...
- * STDOUT (length > 0 with HTTP headers and HTTP content)
+ * STDOUT (length &gt; 0 with HTTP headers and HTTP content)
  * STDOUT (length == 0 to signal end of STDOUT frames)
- * [STDERR (length > 0)]
+ * [STDERR (length &gt; 0)]
  * [STDERR (length == 0 to signal end of STDERR frames)]
  * END_REQUEST
  * </pre>
--- a/jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/HttpChannelOverFCGI.java
+++ b/jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/HttpChannelOverFCGI.java
@@ -13,9 +13,7 @@
 
 package org.eclipse.jetty.fcgi.server;
 
-import java.util.ArrayList;
 import java.util.LinkedList;
-import java.util.List;
 import java.util.Locale;
 import java.util.Queue;
 import java.util.concurrent.Executor;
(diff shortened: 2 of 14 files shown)


jetty-fcgi/fcgi-client/src/main/java/org/eclipse/jetty/fcgi/parser/Parser.java
```diff
@@ -89,9 +106,12 @@ public boolean parse(ByteBuffer buffer)
                             // parsing; the async operation will eventually resume parsing.
                             return true;
                         }
+                        else
```
lorban: There is no need for an `else` block here as that branch is the fall-though one.

jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/HttpChannelOverFCGI.java
```diff
@@ -45,104 +43,121 @@
...
-        try (AutoLock l = _lock.lock())
-        {
-            failure = _specialContent == null ? null : _specialContent.getError();
-            if (failure == null)
-                _contentQueue.offer(content);
-        }
+        Throwable failure = _specialContent == null ? null : _specialContent.getError();
+        if (failure == null)
+            _contentQueue.offer(content);
         if (failure != null)
```
lorban: This `if` could be an `else`.

jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/HttpChannelOverFCGI.java
```diff
@@ -45,104 +43,121 @@
...
+        if (needContent("immediate"))
+            return true;
+        parseAndFill();
+        if (needContent("parsed"))
+            return true;
+        connection.getEndPoint().tryFillInterested(_asyncFillCallback);
+        return false;
+    }
+
+    private boolean needContent(String context)
```
lorban: I find this method name slightly confusing as it doesn't "demand" content like all other `needContent` methods do. How about calling it `hasContent`?

jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/HttpChannelOverFCGI.java
```diff
@@ -45,104 +43,121 @@
...
-                LOG.debug("needContent has content? {}", hasContent);
-            return hasContent;
+                LOG.debug("needContent has {} content {}", context, this);
+            return true;
         }
+        return false;
     }
 
     @Override
     public HttpInput.Content produceContent()
```
lorban, line 113: The following method is surprisingly cryptic. I'd be happy if it could be simplified a bit, like for instance:
```java
if (!hasContent())
  parseAndFill();
if (!hasContent())
  return null;
Content content = _contentQueue.poll();
if (content == null)
  content = _specialContent;
```

jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/HttpChannelOverFCGI.java
```diff
@@ -152,10 +167,7 @@ protected boolean eof()
     {
         if (LOG.isDebugEnabled())
             LOG.debug("received EOF");
-        try (AutoLock l = _lock.lock())
-        {
-            _specialContent = new HttpInput.EofContent();
-        }
+        _specialContent = new HttpInput.EofContent();
```
lorban: `new HttpInput.EofContent()` could be a constant.

jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/HttpChannelOverFCGI.java
```diff
@@ -260,13 +263,45 @@ private boolean doOnIdleTimeout(Throwable x)
     @Override
     public void recycle()
     {
-        try (AutoLock l = _lock.lock())
+        if (!_contentQueue.isEmpty())
+            throw new AssertionError("unconsumed content: " + _contentQueue);
+        _specialContent = null;
+        super.recycle();
```
lorban, line 283: `super.recycle()` should be the first call, otherwise there is a risk that demanding content could be left pending. Of course, this bug does not exist for FCGI and HTTP/1 but that was a real HTTP/2 bug so I'd prefer that we keep this pattern in sync across all protocols.

jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/ServerFCGIConnection.java
```diff
@@ -117,51 +118,109 @@ else if (read == 0)
...
+    /**
+     * This is just a "consume" method, so it must not call
+     * fillInterested(), but just consume what's in the network
+     * for the current request.
+     */
+    void parseAndFill()
+    {
+        if (LOG.isDebugEnabled())
+            LOG.debug("parseAndFill {}", this);
+        while (channel != null)
```
lorban, line 138: I think it would be worthwhile to add a comment stating that this is the "committee loop", that it matches what HTTP/1 is doing and that it shouldn't be touched lightly despite the fact that it's counter-intuitive.

lorban (dismissed):
There are a few minor cleanups that could be done, but otherwise LGTM.

jetty-fcgi/fcgi-client/src/main/java/org/eclipse/jetty/fcgi/parser/Parser.java
sbordet: Done.

jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/HttpChannelOverFCGI.java
sbordet: Done.

jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/HttpChannelOverFCGI.java
sbordet: Done.

jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/HttpChannelOverFCGI.java
sbordet: Done.

jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/HttpChannelOverFCGI.java
sbordet: Done.

jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/HttpChannelOverFCGI.java
sbordet: Done. Will not change HTTP/1.1 though.

jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/ServerFCGIConnection.java
sbordet: Done.

jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/HttpChannelOverFCGI.java
```diff
@@ -35,114 +33,141 @@
...
+    private static final HttpInput.Content EOF_CONTENT = new HttpInput.EofContent();
 
     private final Queue<HttpInput.Content> _contentQueue = new LinkedList<>();
-    private final AutoLock _lock = new AutoLock();
-    private HttpInput.Content _specialContent;
+    private final Callback _asyncFillCallback = new AsyncFillCallback();
+    private final ServerFCGIConnection connection;
     private final HttpFields.Mutable fields = HttpFields.build();
     private final Dispatcher dispatcher;
+    private HttpInput.Content _specialContent;
```
gregw: other fields here are a bit of a mix of  _ and no _.  Can you correct them all to be consistent.

jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/HttpChannelOverFCGI.java
```diff
@@ -35,114 +33,141 @@
...
     {
-        HttpInput.Content content;
-        try (AutoLock l = _lock.lock())
+        if (!hasContent())
+            parseAndFill();
+
+        if (!hasContent())
+            return null;
+
+        HttpInput.Content content = _contentQueue.poll();
```
gregw: Do we still need a queue here?  onContent now always returns true to ensure back pressure, so we should not get more than one content.
Also should we rename the action in the parse from ASYNC to IN_CONTENT?

lorban (dismissed):

sbordet:
@lorban @gregw please re-review.

gregw (approved):

lorban (approved):

Issue #6696, opened 2021-09-05
High WebSocket memory usage in Jetty 10


SerCeMan:
**Jetty version(s)**

`10.0.6`

**Java version/vendor** `(use: java -version)`

```
$ java -version
openjdk version "11.0.12" 2021-07-20 LTS
OpenJDK Runtime Environment Zulu11.50+19-CA (build 11.0.12+7-LTS)
OpenJDK 64-Bit Server VM Zulu11.50+19-CA (build 11.0.12+7-LTS, mixed mode)
```

**OS type/version**

Mac/Ubuntu

**Description**

Hi, Jetty maintainers!

I got back to trying to reproduce https://github.com/eclipse/jetty.project/issues/6328. While I wasn't able to reproduce the referenced issue yet, I seem to have found the main issue that we experienced with the upgrade to Jetty 10, which was related to memory usage. 

Compared to Jetty 9, the memory usage per open WebSocket connection has increased significantly, making it problematic for using when servers maintain a large number of connections. I've put together a minimal reproducer project here: [jetty10-memory-test](https://github.com/SerCeMan/jetty10-memory-test), which implements two minimal WebSocket echo servers on Jetty 9, and Jetty 10, both run with `Xmx1G` and a test that opens 10k connections. 

When running the tests against a bare-bones jetty 9 server, the memory usage does not increase much. 
<img width="1208" alt="Screen Shot 2021-09-05 at 12 13 56 pm" src="https://user-images.githubusercontent.com/1780970/132112642-a396582f-3281-4260-b016-6504d2f42f55.png">

However, when running the tests against Jetty 10, the memory usage grows significantly, and the server dies with an out of memory error after reaching ~9k connections. 
<img width="1202" alt="Screen Shot 2021-09-05 at 12 15 31 pm" src="https://user-images.githubusercontent.com/1780970/132112674-520f920c-a243-4d6f-8c54-a74b02503718.png">

The OOM heapdump seems to match the heapdump from the load tests which discovered the issue, and it hints at `org.eclipse.jetty.websocket.core.server.internal.RFC6455Negotiation`.
<img width="1062" alt="Screen Shot 2021-09-05 at 12 19 25 pm" src="https://user-images.githubusercontent.com/1780970/132112733-3ea2d623-a8d1-4248-a89d-6992308111bc.png">

[...]

PR #6698, opened 2021-09-06
Issue #6696 - don't keep Request object in the WebSocketNegotiation after upgrade

lachlan-roberts:
Issue #6696 

null out the baseRequest field in `WebSocketNegotiation` after the websocket upgrade is complete.

### DIFF
--- a/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/internal/AbstractHandshaker.java
+++ b/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/internal/AbstractHandshaker.java
@@ -167,7 +167,7 @@ public abstract class AbstractHandshaker implements Handshaker
         // Save state from request/response and remove reference to the base request/response.
         upgradeRequest.upgrade();
         upgradeResponse.upgrade();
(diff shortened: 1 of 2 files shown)


jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/internal/AbstractHandshaker.java
```diff
@@ -167,6 +167,7 @@ public boolean upgradeRequest(WebSocketNegotiator negotiator, HttpServletRequest
         // Save state from request/response and remove reference to the base request/response.
         upgradeRequest.upgrade();
         upgradeResponse.upgrade();
+        negotiation.setBaseRequest(null);
```
sbordet: The previous 2 lines that call `upgrade()` actually null-out the references to the request and response, respectively.
I would call the method on `WebSocketNegotiation` as `upgrade()` as well, just in case in future we want to null-out some other field, or copy something from the `Request` object to save it for later use, like the other `upgrade()` methods do.

sbordet (changes requested):

sbordet (approved):

gregw:
@lachlan-roberts please merge

PR #6702, opened 2021-09-07
Cleaning up jetty-start Usecase testing.

joakime:
Changes made to jetty-start to make managing usecases easier.
This change is also to prepare for the `[ini]` resolution order issue PR next.

+ Changes are isolate to `src/test/java`
+ No changes to `src/main/java`
+ Each test is now self contained entirely within  java code
+ Removed `*.prepare.txt`, `*.cmdline.txt`, and `*.assert.txt` files
+ Removed all `src/test/resources/usecases` content


gregw:
@joakime can y ou motivate the change a bit more?
I accept that the current "interpretted" test setup is a bit clunky, but I'm not sure that writing individual tests is really a better approach... maybe it is... but I'll have to think about it.
But more importantly, I think any such refactor should just be in 10 rather than 9.... hmmm maybe 9 is OK because this is only tests..... I'm cautious.

Anyway, some more explanation would help to understand the why as well as the what/where/how/etc.

jetty-start/src/test/java/org/eclipse/jetty/start/CollectionAsserts.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.util.Collection;
+import java.util.List;
+
+import org.hamcrest.Matchers;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class CollectionAsserts
```
joakime: This is a cleaned up and renamed ConfigurationAsserts from before.  
See: https://github.com/eclipse/jetty.project/blob/jetty-9.4.43.v20210629/jetty-start/src/test/java/org/eclipse/jetty/start/ConfigurationAssert.java

It's not a new class here, but it is renamed and significantly smaller.
The old ConfigurationAssert had no reason to exist anymore as a result of this PR's cleanup.

jetty-start/src/test/java/org/eclipse/jetty/start/usecases/AbstractUseCase.java
```diff
@@ -0,0 +1,265 @@
...
+import org.eclipse.jetty.start.StartArgs;
+import org.eclipse.jetty.start.StartLog;
+import org.eclipse.jetty.toolchain.test.FS;
+import org.eclipse.jetty.toolchain.test.jupiter.WorkDir;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public abstract class AbstractUseCase
```
sbordet, line 51: We always called base test classes `AbstractTest` or similar, so I would keep the convention, also because all subclasses are called `*Test`.

sbordet (approved):
Small renaming if you want to do it, otherwise LGTM.

PR #6734, opened 2021-09-08
Issue #6072 - backport of minimum fix for jetty-9.3.x

joakime:
Backport of #6072 (Large TLS Record Fix) to branch `jetty-9.3.x`

Co-authored-by: Olivier Lamy <oliver.lamy@gmail.com>

### DIFF
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
@@ -743,7 +743,7 @@ public class SslConnection extends AbstractConnection
                                     case BUFFER_UNDERFLOW:
                                         // Continue if we can compact?
                                         if (BufferUtil.compact(_encryptedInput))
-                                            continue;
+                                            break decryption; // to cause fill again.
 
                                         // Are we out of space?
                                         if (BufferUtil.space(_encryptedInput) == 0)
@@ -753,6 +753,7 @@ public class SslConnection extends AbstractConnection
                                         }
 
                                         _underFlown = true;
+                                        // Fall through to OK case.
                                     case OK:
                                     {
                                         if (unwrapHandshakeStatus == HandshakeStatus.FINISHED)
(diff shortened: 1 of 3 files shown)


jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
```diff
@@ -741,6 +741,18 @@ else if (_decryptedInput == null)
...
+                                        if (BufferUtil.compact(_encryptedInput))
+                                            continue;
+
+                                        // Are we out of space?
+                                        if (BufferUtil.space(_encryptedInput) == 0)
+                                        {
+                                            BufferUtil.clear(_encryptedInput);
+                                            throw new SSLHandshakeException("Encrypted buffer max length exceeded");
+                                        }
+
```
gregw, line 754: I suspect we should also have:
```java
                                    // if we just filled some
                                    if (netFilled > 0)
                                        continue; // try filling some more
```

jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
```diff
@@ -741,6 +741,18 @@ else if (_decryptedInput == null)
...
+                                        _underFlown = true;
```
gregw, line 755: I'm concerned about the fall through to OK here.  I'm wondering if we should not just return like:
```suggestion
                                        _underFlown = true;
                                        break decryption;
```
This is the probably outcome from falling through anyway.

At the very least we need a comment saying that we intend to fall through and why.

gregw (changes requested):
I would not mind walking through this one with @sbordet

jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
joakime: This is handled in two ways already, up at line 700, and case NEED_UNWRAP around line 813.

And it should be `break decryption;` to force a read/fill again on 9.3.x

jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
joakime: Added the comment.

joakime:
@gregw i had a review from @sbordet and made the last set of changes I think this backport needs.

gregw (approved):

sbordet (approved):

Issue #6756, opened 2021-09-09
Deprecate `/jetty-spring/` artifact in `jetty-9.4.x` releases.


joakime:
**Jetty version(s)**
9.4.x

**Java version/vendor** `(use: java -version)`
All

**Description**
The `/jetty-spring/` artifact in `jetty-10.0.x` has already been dropped.

But the `/jetty-spring/` artifact on `jetty-9.4.x` is still using Spring 3.2.8.RELEASE (which was declared EOL in [Dec 2016](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Versions#supported-versions))

This was only noticed because of the new dependabot configuration.

joakime:
Also of note, Spring 3.2.x has a supported JDK version range of 1.4 thru 1.7. (it does not support JDK 8 or JDK 11, and is known to not work on JDK 11 properly)
We should have dropped `/jetty-spring/` back when Jetty 9.x was started.

joakime:
Jetty Spring removed in jetty-10.0.x onwards

PR #6757, opened 2021-09-09
Issue #6756 - Drop jetty-spring module/artifact

joakime:
The `/jetty-spring/` artifact in `jetty-10.0.x` has already been dropped.

But the `/jetty-spring/` artifact on `jetty-9.4.x` is still using Spring 3.2.8.RELEASE (which was declared EOL in [Dec 2016](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Versions#supported-versions))

This was only noticed because of the new dependabot configuration.

Also of note, Spring 3.2.x has a supported JDK version range of 1.4 thru 1.7. (it does not support JDK 8 or JDK 11, and is known to not work on JDK 11 properly)
We should have dropped `/jetty-spring/` back when Jetty 9.4.x was started. (it was EOL then)

_(wow, [Jetty 9.x has been around since 2013](https://github.com/eclipse/jetty.project/releases/tag/jetty-9.0.0.v20130308)!)_


gregw (commented):
I'm not understanding?  We've dropped the module in 10, but as it has been a part of jetty 9.4 from the beginning, why are we dropping it now?   

We probably CAN drop it because I don't think anybody uses it, but that doesn't mean we should drop a whole module in a dot release.   Just let it sit with whatever ancient version of spring it works with.

PR #6758, opened 2021-09-09
Issue #6752 - Extensible DefaultSessionCache map implementation

### LINKED ISSUE #6752: DefaultSessionCache more extensible using ConcurrentMap

joakime:
Alternate approach to PR #6753 from @prenagha 

Closes #6752 

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/session/DefaultSessionCache.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/session/DefaultSessionCache.java
@@ -13,6 +13,7 @@
 
 package org.eclipse.jetty.server.session;
 
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.function.Function;
@@ -52,12 +53,12 @@ public class DefaultSessionCache extends AbstractSessionCache
 
     /**
      * @param manager The SessionHandler related to this SessionCache
-     * @param sessionMap The session map implementation to use
+     * @param sessions The session map implementation to use
      */
-    public DefaultSessionCache(SessionHandler manager, ConcurrentMap<String, Session> sessionMap)
+    public DefaultSessionCache(SessionHandler manager, ConcurrentMap<String, Session> sessions)
     {
         super(manager);
-        this._sessions = sessionMap;
+        _sessions = Objects.requireNonNull(sessions, "Session Map may not be null");
     }
 
     /**
@@ -193,5 +194,12 @@ public class DefaultSessionCache extends AbstractSessionCache
     public boolean doReplace(String id, Session oldValue, Session newValue)
     {
         return _sessions.replace(id, oldValue, newValue);
+    }
+
+    @Override
+    protected void doStart() throws Exception
+    {
+        Objects.requireNonNull(_sessions, "Session Map may not be null");
+        super.doStart();
     }
 }
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/session/DefaultSessionCache.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/session/DefaultSessionCache.java
@@ -195,11 +195,4 @@ public class DefaultSessionCache extends AbstractSessionCache
     {
         return _sessions.replace(id, oldValue, newValue);
     }
-
-    @Override
-    protected void doStart() throws Exception
-    {
-        Objects.requireNonNull(_sessions, "Session Map may not be null");
-        super.doStart();
-    }
 }
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/session/DefaultSessionCache.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/session/DefaultSessionCache.java
@@ -28,7 +28,7 @@ import org.slf4j.LoggerFactory;
 /**
  * DefaultSessionCache
  *
- * A session store that keeps its sessions in memory in a concurrent map
+ * A session store that keeps its sessions in memory within a concurrent map
  */
 @ManagedObject
 public class DefaultSessionCache extends AbstractSessionCache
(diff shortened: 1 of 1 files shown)


joakime:
@prenagha an additional constructor to this PR could be added to allow for some late construction of the map.  But I'm not sure if it's worth the effort.

``` java
    /**
     * @param manager The SessionHandler related to this SessionCache
     * @param sessionMapSupplier The supplier of the session map implementation to use (to allow for late binding)
     */
    public DefaultSessionCache(SessionHandler manager, java.util.function.Supplier<ConcurrentMap<String,Session>> sessionMapSupplier)
    {
        this(manager, sessionMapSupplier.get());
    }
```

which would allow your example to work like ...

``` java
public class SessionExpiringNearCache extends DefaultSessionCache {
  public SessionNearCache(SessionHandler manager) {
    super(manager, () -> Caffeine.newBuilder()
        .expireAfterAccess(15, TimeUnit.MINUTES)
        .build());
  }
}
```

or just use the DefaultSessionCache directly, no extends ...

``` java
DefaultSessionCache sessionCache = new DefaultSessionCache(sessionHandler,
   () -> Caffeine.newBuilder().expireAfterAccess(15, TimeUnit.MINUTES).build());
```

or (if using jgroups for example)

``` java
DefaultSessionCache sessionCache = new DefaultSessionCache(sessionHandler, new ReplicatedHashMap<>(jgroupsChannel));
```

prenagha:
@joakime Constructor approach looks good. Not sure if you all ever do null checking, but might be a good idea to add to the constructor that takes the map

```
Objects.requireNonNull(sessionMap, "Session Map is required");
```

jetty-server/src/main/java/org/eclipse/jetty/server/session/DefaultSessionCache.java
```diff
@@ -37,16 +38,26 @@
...
+    }
+
+    /**
+     * @param manager The SessionHandler related to this SessionCache
+     * @param sessionMap The session map implementation to use
+     */
+    public DefaultSessionCache(SessionHandler manager, ConcurrentMap<String, Session> sessionMap)
     {
         super(manager);
+        this._sessions = sessionMap;
```
janbartel:
```suggestion
        _sessions = sessionMap;
```

janbartel (changes requested):
If you get rid of the "this", I'll approve this.

jetty-server/src/main/java/org/eclipse/jetty/server/session/DefaultSessionCache.java
gregw: so whilst we are picking nits... the field and argument names should both be the same.  either `_sessions = sessions` or `_sessionMap = sessionMap`.    Since the field is protected, it is kind of part of the class contract, so I guess it can't be renamed.

janbartel:
@joakime I'm definitely not interested in making the constructor any more complicated than just simply passing in an arg of type ConcurrentMap because it is just not worth it. But I _am_ interested in _either_ the addition of the Objects.requireNonNull check in the constructor, or a null check in the doStart() method (good catch @prenagha).

jetty-server/src/main/java/org/eclipse/jetty/server/session/DefaultSessionCache.java
```diff
@@ -37,16 +38,26 @@
...
     public DefaultSessionCache(SessionHandler manager)
+    {
+        this(manager, new ConcurrentHashMap<>());
+    }
+
+    /**
+     * @param manager The SessionHandler related to this SessionCache
+     * @param sessionMap The session map implementation to use
+     */
+    public DefaultSessionCache(SessionHandler manager, ConcurrentMap<String, Session> sessionMap)
```
gregw: other option to avoid null issues is:
```suggestion
    {
        this(manager, null);
    }

    /**
     * @param manager The SessionHandler related to this SessionCache
     * @param sessionMap The session map implementation to use
     */
    protected DefaultSessionCache(SessionHandler manager, ConcurrentMap<String, Session> sessionMap)
    {
        super(manager);
        _sessions = sessionMap == null ? new ConcurrentHashMap() : sessionMap;
```

gregw: Also make this constructor protected

janbartel:
@joakime is this ready for re-review yet?

joakime:
how is this still conflicting??

joakime:
@janbartel this PR is ready for a review

jetty-server/src/main/java/org/eclipse/jetty/server/session/DefaultSessionCache.java
```diff
@@ -184,4 +195,11 @@ public boolean doReplace(String id, Session oldValue, Session newValue)
     {
         return _sessions.replace(id, oldValue, newValue);
     }
+
+    @Override
+    protected void doStart() throws Exception
+    {
+        Objects.requireNonNull(_sessions, "Session Map may not be null");
```
janbartel: This check shouldn't be necessary:  `_sessions` is  `final` && the constructor at line 51 ensures a non-null map && the other constructor at line 61 rejects nulls.

janbartel (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/session/DefaultSessionCache.java
joakime: I just removed the entire `doStart()`

janbartel (approved):

PR #6765, opened 2021-09-10
Jetty 10.0.x quic http3

### LINKED ISSUE #6728: QUIC and HTTP/3
**Target Jetty version(s)**
10.0.x

**Enhancement Description**
This is the umbrella issue for the implementation of QUIC and HTTP/3.

QUIC: [RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000)
HTTP/3: https://datatracker.ietf.org/doc/html/draft-ietf-quic-http

gregw:
This PR is a draft that I created against the h3 development branch so we can use github tools to review/discuss the code


jetty-quic/quic-common/src/main/java/org/eclipse/jetty/quic/common/ProtocolQuicSession.java
```diff
@@ -0,0 +1,56 @@
...
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.quic.common;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+public class ProtocolQuicSession
```
gregw: @sbordet @lorban We should do a 'Http1ProtocolQuicSession` so we can quickly validate the basic lifecycle with working code.

gregw:
Hmmmm this PR might be a bit big to be useful... but we will ultimately need to review, so starting early???

However, I'm wondering if we can reduce verbosity a little bit by squashing the many commits (yeah I know I normally argue against squashes in branches, but rules always have exceptions)?

jetty-http3/http3-server/src/main/java/org/eclipse/jetty/http3/server/HTTP3QuicSession.java
```diff
@@ -0,0 +1,49 @@
...
+import org.eclipse.jetty.http3.qpack.Instruction;
+import org.eclipse.jetty.http3.qpack.QpackDecoder;
+import org.eclipse.jetty.http3.qpack.QpackException;
+import org.eclipse.jetty.quic.common.ProtocolQuicSession;
+import org.eclipse.jetty.quic.common.QuicSession;
+
+public class HTTP3QuicSession extends ProtocolQuicSession
+{
+    private final QpackDecoder decoder;
+    private final Instruction.Handler decoderHandler = new QpackDecoderInstructionHandler();
```
gregw: So it is expected that there will be an encoder here as well.
Potentially also parsers/generators, Flushers and queues of frames????

jetty-http3/http3-qpack/src/main/java/org/eclipse/jetty/http3/qpack/QpackDecoder.java
```diff
@@ -0,0 +1,276 @@
...
+    public static final Logger LOG = LoggerFactory.getLogger(QpackDecoder.class);
+
+    private final List<Instruction> _instructions = new ArrayList<>();
+    private final Instruction.Handler _handler;
+    private final QpackContext _context;
+    private final DecoderInstructionParser _parser;
+    private final List<EncodedFieldSection> _encodedFieldSections = new ArrayList<>();
+    private final NBitIntegerParser _integerDecoder = new NBitIntegerParser();
+    private final int _maxHeaderSize;
+    private MetaData _metaData;
```
gregw: @lachlan-roberts 
A `Decoder` can decode `MetaData` for multiple streams, so why is there a single `_metaData` field?  I'm guessing this is the current meta data field being built?  But are we sure that we can't be called like:
```java
    decoder.decode(streamA, bufferWithFirstHalfOfMetaDataA, myHandler);
    decoder.decode(streamB, bufferWithFirstHalfOfMetaDataB, myHandler);
    decoder.decode(streamA, bufferWithSecondHalfOfMetaDataA, myHandler);
    decoder.decode(streamB, bufferWithSecondHalfOfMetaDataB, myHandler);
```
If this can't happen, then at the very least this needs to be documented, but I'd also rather we remembered the streamId in progress and threw an ISE if we are called with another stream when we have not finished the first.

jetty-http3/http3-qpack/src/main/java/org/eclipse/jetty/http3/qpack/QpackDecoder.java
```diff
@@ -0,0 +1,276 @@
...
+    {
+        return _context;
+    }
+
+    public interface Handler
+    {
+        void onMetaData(long streamId, MetaData metadata);
+    }
+
+    public boolean decode(long streamId, ByteBuffer buffer, Handler handler) throws QpackException
```
gregw, line 107: @lachlan-roberts I've given this feedback before.... but I'll try again as I think I'm getting my objections a bit better formed.

I don't like the API of `QpackDecoder` as it looks like an active object rather than a passive state machine (like SslEngine or quiche).    This is important because the caller of `decode` is likely to be in a very specific invocation context, specifically needing to know if it is BLOCKING, NON_BLOCKING or EITHER, but with this API it is not possible to say that `decode` is a non blocking call or not, since it depends entirely on the implementations of the handlers that the decoder has.   Now it may be possible for the user of the decoder to know completely the invocation behaviours of those call backs, but even so that relationship may not be clear to somebody reading the code.

A simple example is that if the `Handler.onMetaData` callback goes on the actually invoke the application with the request that is represented by that metadata, then a thread may be still inside the decoder and blocked (or active) in the application when another thread needs to call `decode` either for other streams or perhaps for trailers on the same stream.   So sure, we can say that the handler should not be implemented that way... but it would be better to have an API that couldn't be implemently badly.

To take control of the invocation, the user of `Decoder` will need to:
 + construct the `Decoder` with an `Instruction.Handler` that never blocks, so it probably needs to just add them to a queue.
 + call the `decode` method with a `Handler` that never blocks and just remembers the `MetaData`
 + reacts to a true return from `decode` by looking for a remembered `MetaData`  and acting on it.
 + ensures all remembered `Instructions` are handled after the `decode` is returned but not delayed by any metadata handling.

Why not just make that explict with `decode` return a `Result` that contains both  `MetaData` and a collection of `Instructions`.  Or alternately just an action enumeration saying that there is a `MetaData` and/or collection of `Instructions` available, which can then be pulled from the decoder.

jetty-http3/http3-qpack/src/main/java/org/eclipse/jetty/http3/qpack/QpackDecoder.java
```diff
@@ -0,0 +1,276 @@
...
+
+        MetaData metaData = _metaData;
+        _metaData = null;
+        if (metaData != null)
+            handler.onMetaData(streamId, metaData);
+
+        return metaData != null;
+    }
+
+    void parseInstruction(ByteBuffer buffer) throws QpackException
```
gregw: @lachlan-roberts is this public API?  is this the only way that recieved instructions make it into the instruction parser?

jetty-http3/http3-qpack/src/main/java/org/eclipse/jetty/http3/qpack/QpackDecoder.java
```diff
@@ -0,0 +1,276 @@
...
+    @Override
+    public String toString()
+    {
+        return String.format("QpackDecoder@%x{%s}", hashCode(), _context);
+    }
+
+    /**
+     * This delivers notifications from the DecoderInstruction parser directly into the Decoder.
+     */
+    class DecoderAdapter implements DecoderInstructionParser.Handler
```
gregw: @lachlan-roberts  
I don't think `Adaptor` is the right name here.   That implies it is an abstract implementation that is intended to be further extended.
Rather this is the implementation of the parser handler that integrates into the `Decoder`, so perhaps just call it `InstructionHandler`

jetty-quic/quic-server/src/main/java/org/eclipse/jetty/quic/server/H3Server.puml
```diff
@@ -0,0 +1,65 @@
+@startuml
+title "H3 Quic Server"
```
gregw, line 2: @sbordet @lachlan-roberts @lorban I know it is early days, but I do think puml diagrams like this will help us work out details.
I'm happy to mostly maintain it, but if you can review and comment.

gregw:
@sbordet @lorban @lachlan-roberts On the call last week I struggled to make my concerns clear about the scheduling of h3.     I've put more thought into it, so let me have another try.

First, let's go over how the execution strategy works in other places:
 + For the selector manager, we don't have a thread do a select and then iterator over all the selected keys.  Instead we have a producer model that when asked to produce, the producer may look at the results from the previous select and produce a task that is associated with one of the selected keys.  The producer doesn't know what protocol is being spoken and a selected key could be an acceptor, h1, h2, ssl, or even just an update task could be produced.
 + At the h2 level, we have another execution strategy where the producer does the reading/parsing of the frames.  It produces a single tasks even though there may be more reading able to be done or more frames in the data already read.   This is key because we want the execution strategy to have the best change for using the same thread that did the reading & parsing to als odo the handling.

So in both cases, we have a producer that unwinds a natural loop (over selected keys, or over fill&parsed frames) and produces a single task that is hot in the CPU cache that can be fed to the execution strategy to potentially be directly executed without dispatch.

For quic+h3 there is an equivalent to the selector selecting a set of keys.   It is quiche producing a set of readable streams and writable streams.   We need a producer at that level that will produce the next tasks from that list - which may be to pick a single readable stream, read the data, pick the correct parser for that (based on streamID it could be settings stream, an instruction stream or a H3 stream(which itself could parse header frames or data frames).  A single task should be returned from that production so that the same thread can go onto handle that task with the frame/parsing hot in its cache.    The next producing will then produce the next task - which may be from the same read/parse (eg there may be multiple metadata's produced from a single instruction), or it may go to the next readable streamId to do a read/parse etc.

Ultimately, a produce will find nothing to do and return, leaving that execution strategy idle, to be woken up when it is next selected.   Note that this should allows the execution strategy itself to do the "1 and only 1 thread" logic that @sbordet has currently written around AtomicInt.  The ExecutionStrategy already only ever allows a single producer, so that will protect our qpack tables and parsers from having multiple threads - we don't need to do anything extra.

OK, so it  is a little more complex that the h2 execution stragegy, in that we don't have a single parser that the producer can feed.  But I'm pretty sure we can do it similar to the selector, where the specifics of how to act  are in the key attachment.  For quiche/h3 it will be the streamId that will allows a sub-producer to be selected by the selector, that will then do what ever it needs to do to produce the next task (read & parse,  or produce task from a previous parse, or wake up a writer, or callback the application to say it can write, etc. etc.)

The key think is that the execution strategy should be thought of as having a producer at the QuicSession level and that it pulls tasks out of the multiple streams that may be indicated as active by quiche.    Currently we are thinking that the execution strategy is after the stream layers and that QuicSession pushes to the streams and those streams push to the producer of a subsequent execution strategy.   I don't think that can work... and if it can, it is too complex.   I'm sure we can "push into a producer", but I think it will involve handoffs.  Much better to have a producer that pulls tasks out of the various statemachines and parsers.

@lachlan-roberts, so to fit this model, that is why I like a Decoder that you pull from rather than it actively calls handlers.     There will be a producer that will ask the Decoder: "do you have any thing I can act on?" and the Decoder can return an Instruction to send or an emitted MetaData.  The producer can act on that task and arrange for another producer to call the Decoder again.  Only once the Decoder says: "Nothing more I can do", does the producer look to see if the instruction stream and/or normal stream is able to be read from, and the resulting data parsed into frames and fed into the Decoder, which is then asked agin: "do you have anything I can act on?".  Doing this way, will unwind the loops that currently exist in the Decoder and allow the execution strategy to be involved.  More over, whilst there is probably some queue of metadata & instances that will be produced, this queue can be internal, final and reused.   A new list will not need to be created on every parse.

jetty-quic/quic-common/src/main/java/org/eclipse/jetty/quic/common/QuicSession.java
```diff
@@ -0,0 +1,415 @@
...
+
+    public void process(SocketAddress remoteAddress, ByteBuffer cipherBufferIn) throws IOException
+    {
+        // While the connection ID remains the same,
+        // the remote address may change so store it again.
+        this.remoteAddress = remoteAddress;
+
+        quicheConnection.feedCipherText(cipherBufferIn, remoteAddress);
+
+        if (quicheConnection.isConnectionEstablished())
```
gregw: maybe first check for `protocolSession!=null` as we don't know if/when the established state might revert to false during a shutdown?

jetty-io/src/main/java/org/eclipse/jetty/io/EndPoint.java
```diff
@@ -203,7 +203,15 @@ default void close()
...
-    int fill(ByteBuffer buffer) throws IOException;
+    default int fill(ByteBuffer buffer) throws IOException
+    {
+        throw new UnsupportedOperationException();
+    }
+
+    default SocketAddress receive(ByteBuffer buffer) throws IOException
+    {
+        throw new UnsupportedOperationException();
+    }
```
gregw: I'm not convinced it is correct to add this method to our abstract EndPoint.  OK for the short term, but perhaps we should consider refactoring so that `EndPoint` has neither `fill` nor `receive`.  Instead these signatures should be moved to `DatagramEndPoint` and `StreamEndPoint`?   Ditto for `flush`,  `write` and `send`.

But let's leave it this way for now and just be aware that it may need to change.

jetty-quic/quic-common/src/main/java/org/eclipse/jetty/quic/common/QuicSession.java
```diff
@@ -0,0 +1,434 @@
...
+ * <p>Represents a logical connection with a remote peer, identified by a QUIC connection ID.</p>
+ * <p>Each QuicSession maintains a number of QUIC streams, identified by their QUIC stream ID;
+ * Each QUIC stream is wrapped in an {@link EndPoint}, namely {@link QuicStreamEndPoint}.</p>
+ * <p>Bytes received from a {@link QuicConnection} in {@link #process(SocketAddress, ByteBuffer)}
+ * are passed to Quiche for processing; in turn, Quiche produces a list of QUIC stream IDs that
+ * have pending I/O events, either read-ready or write-ready.</p>
+ * <p>On the receive side, a QuicSession <em>fans-out</em> to multiple {@link QuicStreamEndPoint}s.</p>
+ * <p>On the send side, many {@link QuicStreamEndPoint}s <em>fan-in</em> to a QuicSession.</p>
+ */
+public abstract class QuicSession
```
gregw: @sbordet @lorban  I find the naming very confusion.   A `QuicSession` containers a `QuicheConnection` and the `QuicConnection` is not a connection at all, but more part of the `Connector` or maybe the `ConnectionFactory`.

I think it would be far simpler to rename `QuicSession` to `QuicConnection`, as that is what it is.  The current `QuicConnection` is only named that way because it is being shoehorned into the current SelectorManager and AbstractConnection.  It kind of fits, but the wrong name shows that there is probably a better way.

gregw:
Some more musings on execution strategy.... this time mostly for my own consumption....

Currently the reading of h3 datagrams is looking like the following pseudo code:
```java
SocketAddress peer = DatagramChannel.receive(netbuf)
QuicheConnectionId cId = QuicheConnectionId.fromPacket(netbuf)
QuicheConnection quicheConnection = map.get(cId);
if (quicheConnection==null)
{
    QuicheConnection quicheConnection = QuicheConnection.tryAccept(config, new SimpleTokenValidator(peer), netBuf, peer);
    if (quicheConnection==null)
    {
        if (QuicheConnection.negotiate(new SimpleTokenMinter(peer), netBuf, negBuf))
            datagramChannel.send(negBuf, peer);
        return null;
    }
}
quicheConnection.feedCipherText(netBuf, peer); // TODO what if not all consumed?

// At this point current code mutually excludes and execute the following:
for (streamId : quicheConnection.readableStreamIds())
{
    quicheConnection.drainClearTextForStream(streamId, streamBuf);
    H3Parser parser = pickParserBy(streamId); // In reality done by picking a ProtoQuicStream instance
    H3Frame frame = parser.parse(streamId, streamBuf)

    if (frame instanceof HeaderFrame)
    {
        decoder.decode(frame.getContent());
        while (qpacDecoder.hasMetaData())
        {
            request.setMetData(qpacDecoder.takeMetaData();
            if (httpChannel.onRequest(request))
                execute(httpChannel);
        }
    }
    else if (frame instanceof DataFrame)
    {
        if (httpChannel.onContent(frame.getContent()))
            execute(httpChannel);
    }
    else if (frame instanceof Instruction)
    {
        decoder.parseInstruction(frame.getContent());
        while (qpacDecoder.hasMetaData())
        {
            request.setMetData(qpacDecoder.takeMetaData();
            if (httpChannel.onRequest(request))
                execute(httpChannel);
        }
    }
    else if (frame instanceof Settings)
    {
    }
}
```
This is probably triple executed: 1) the default onFillable mechanism assumes blocking so the selectorManger strategy is effectively an execute; 2) the execute before iterating over the readable streams; 3) the HttpChannel will need to be executed.

With regards to parallelism:
 + the `QuicConnection.onFillable` will only be called by a single thread, so only a single `feedCipherText` call will happen at once.  Ideally we should be able to do `feedCipherText` for different connectionIds in parallel
 + once fed the cipherText, the iterating over readable streams is executed, so that iteration can be done for multiple connections at once, but each stream within a connection is processed serially until the HttpChannel is executed.  This means that the channel will never be handled by the same thread that did the reading and parsing.

So I think this is suffering from being a square peg shoved into the round hole that is our SelectorManager->EndPoint->Connection abstraction.   It looks like it might work, but I don't think it will ever work efficiently.   So I'm going to think about how this could/should look if we started from scratch and tried to directly apply our execution strategy producer/consumer abstraction.
 
We need to well match our execution strategies with the 2 fan outs: multiple connections; and  multiple streams within a connection:
 + The initial `receive` and any `tryAcquire` calls should be done in the producer of the first level ExecutionStrategy.
+ The `quicheConnection.feedCipherText` should be done withing a consumer of the first level ExecutionStrategy, so they can happen in parallel for different connections.   The trick will be what to do if two tasks for the same connection are produced quickly? Should the second one block and wait for the first, or just queue its data and return?  I'm favouring the later as the consumer will probably just be the second level execution strategy, so the task becomes `{ connection.addDatagram(buf); connection.produce();}`. Actually the add can probably be done in the first level produce so the consume may just be to call the 2nd level produce().
+ There may need to be a queue of encrypted data buffers in each connection.  Maybe there is a queue inside quiche, but the call to `feedCipherText`  may not be fully consuming, so we'd need to store the unconsumed buffer anyway.
+ The 2nd level producer will do `feedCipherText`, frame parsing and header decoding for a connection. Since producers are single threaded, this means only a single thread accessing the decoder and other associated state at once.
+ The 2nd level consumer tasks will be the HttpChannel, so they can run in parallel if there are many, but by the same producing thread if there is only 1 active.

I think it is possible to have the same thread: receive; feedCipherText; parse; decode; HttpChannel.run();   2 levels of producers will allow other threads to take over producing at the receive step and/or the feedCipeherText step.

gregw:
@sbordet @lachlan-roberts @lorban Oh oh!  I think we have a problem with back pressure.

If I understand quic/quiche correctly (and there is a good chance that I don't) then back pressure on the read side is done by not calling `quicheConnection.drainClearTextForStream`.  Ie. once you have  called drain, then those bytes are considered consumed and flow control magic will happen to allow more bytes to arrive.

Thus our current approach has been wrong - we are reading those bytes on the selector callback, parsing them and then looking to use an invocation strategy to invoke the application.   But that can't work because what if the application has not yet done a read?     I think this will have to be more like h1 and that the call to drain will be done within the application thread calling `read`, which will call `produceContent` which will call `drainClearTextForStream` and then parse those bytes for data (or trailers!!!).     This means that `needContent` will need to register something, so that when the stream is next seen in the `readableStreamIds` then the only action taken is to call `onDataAvailable` and allow the application to again call `read` which will ultimately call `drainClearTextForStream`

gregw:
@sbordet can you help me fill out the following breakdown of all the layers involved and the basics of how they communicate?:

 +  SelectorManager <-> DatagramChannelEndPoint: 
    * -> onSelected()
    * <- submit(selectorUpdate)
 +  DatagramChannelEndPoint <-> QuicConnection:
    * -> onFillable()
    * <- fillInterested(), receive(buffer), send(address, buffers), write(callback, address, buffers)
 + QuicConnection <-> QuicSession:
    * -> process(address, buffer)
    * <- write(callback, address, buffers)
 + QuicSession <-> [Client|Server]HTTP3Session (ProtocolSession)
    * -> process();
    * <- getReadableStreamIds(), getWritableStreamIds()
 +  [Client|Server]HTTP3Session <-> QuicStreamEndPoint:
    * -> onReadable(), onWritable()
    * <- fill(streamId, buffer), flush(streamId, buffer)
 + QuicStreamEndPoint <-> HTTP3StreamConnection:
    * -> onFillable()
    * <- fillInterested(), fill(buffer), flush(buffer), write(callback, buffer)
 +  HTTP3StreamConnection <-> HTTP3Session:
    * -> parser.parse(buffer) -> onXYZ() parser events
    * <- writeFrame(streamId, frame, callback)
 +  HTTP3Session <-> HTTP3Stream: 
    * -> onXYZ(Stream) events (e.g. onRequest(), onResponse(), onTrailer(), onDataAvailable(), etc.)
    * <- writeFrame(streamId, frame, callback)
 +  HTTP3Stream <-> HttpChannelOverHTTP3: 
    * -> onXYZ(Stream) events
    * <- readData(), demand(), writeFrame(streamId, frame, callback)
 + HttpChannelOverHTTP3 <-> Servlet:
    * -> Handler.handle(), onContentProducible(),  sendCallback.succeeded();
    * <- needContent(), produceContent(), transport.send(request, response, buffer, last, callback);
 + Servlet <-> App:
    * -> service(req,resp), onDataAvailable(), onWritePossible()
    * <- in.isReady(), in.read(byte[]), out.isReady(), out.write(byte[])

sbordet:
@gregw updated the comment above.

sbordet:
@gregw things to review more carefully:

* HttpClientTransportDynamic.findConnector(): Avoid to always allocate a default `ClientConnector` and searching deep in the hierarchy, as the `ClientConnector` may now be part of a lower level component.
* `MultiplexConnectionPool`: now also `getMaxUsageCount()` is overridden to ask the connection, if it implements `MaxUsable` to return the count. This is necessary because in HTTP/3 a connection can only open N streams (differently from HTTP/1 or HTTP/2 where a connection could be reused almost infinitely, HTTP/3 limits the number of requests you can make on a connection, and then the client must "open" another connection -- or rather, allocate a new connection ID).
* `PreEncodedHttpField`: removed the hardcoded number of encoders.
* `AsyncContentProducer`: added `isUnready()`, used in `BlockingContentProducer`.

The rest is either new code or tests modified to run HTTP/3 as well.

jetty-http3/http3-server/src/main/config/modules/http3.mod
```diff
@@ -0,0 +1,50 @@
+[description]
+Enables the support for the HTTP/3 protocol.
```
gregw:
```suggestion
Enables experimental support for the HTTP/3 protocol.
```

jetty-http3/http3-server/src/main/java/org/eclipse/jetty/http3/server/HTTP3ServerConnector.java
```diff
@@ -0,0 +1,61 @@
...
+        getQuicConfiguration().setMaxBidirectionalRemoteStreams(128);
+        // HTTP/3 requires a few mandatory unidirectional streams.
+        getQuicConfiguration().setMaxUnidirectionalRemoteStreams(8);
+        getQuicConfiguration().setUnidirectionalStreamRecvWindow(1024 * 1024);
+    }
+
+    @Override
+    protected void doStart() throws Exception
+    {
+        super.doStart();
```
gregw, line 61:
```suggestion
        LOG.info("H3/Quic support is experimental");
        super.doStart();
```

jetty-http3/http3-server/src/main/java/org/eclipse/jetty/http3/server/HTTP3ServerConnector.java
```diff
@@ -0,0 +1,61 @@
...
+import org.eclipse.jetty.quic.server.QuicServerConnector;
+import org.eclipse.jetty.server.ConnectionFactory;
+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.util.ssl.SslContextFactory;
+import org.eclipse.jetty.util.thread.Scheduler;
+
+/**
+ * <p>A HTTP/3 specific {@link QuicServerConnector} that configures QUIC parameters according to HTTP/3 requirements.</p>
+ */
+public class HTTP3ServerConnector extends QuicServerConnector
```
gregw, line 36: I think having this class is not correct long term.   Just like an SslConnector can be h1 or h2, a QuicServerConnector can be h1, h2 or h3.
So don't we want to just have the QuicServerConnector and use code/xml to add the supported protocols to it? 

OK for it to stay as an expediency for the experimental h3, but we should review the need for it.

gregw:
So long as this is clearly marked as experimental, then I think it is ok to merge. 
We need to have an architecture review of the approach after we have settled on some more ideas in jetty-12

sbordet:
@gregw I have incorporated your suggestions and added a log on the client too.

gregw:
@sbordet so since this is "my" PR, I think you can review it and we can merge.  The only permanent thing from this is the maven coords and they are reasonable going forward anyway.

sbordet:
I'll merge - we have done some improvements on the module system but we're waiting for libraries to be in Maven Central.

sbordet (approved):

PR #6766, opened 2021-09-10
Jetty 10.0.x 6687 upgrade infinispan

### LINKED ISSUE #6687: Upgrade Infinispan in all active Jetty branches
**Jetty version(s)**
9.4.x, 10.0.x, and 11.0.x

**Description**

Our usage of infinispan is getting long in the tooth.

We have Infinispan `9.4.8.Final` in all of our branches.
Which uses Hibernate `5.10.3.Final`

Infinispan version 9.4.8.Final is subject to CVE-2019-10158, and [CVE-2019-10174](https://bugzilla.redhat.com/show_bug.cgi?id=CVE-2019-10174).

Infinispan version 12.1.7.Final has been released

We need to upgrade, and should at least skip Infinispan 10.x and start investigating using version 11.x or newer. (There are a few CVEs in the infinispan 10.x series we want to avoid as well)

It should be noted that the Docker Hub for infinispan has also moved.

We are using [`jboss/infinispan-server`](https://hub.docker.com/r/jboss/infinispan-server), but the new location is [`infinispan/server`](https://hub.docker.com/r/infinispan/server)

janbartel:
Upgrade to infinispan 11.0.11

Note that the docker container version for the infinispan server is only at 11.0.9.


janbartel:
@olamy nudge

jetty-infinispan/infinispan-remote-query/pom.xml
```diff
@@ -9,7 +9,7 @@
   <name>Jetty :: Infinispan Session Manager Remote with Querying</name>
   <properties>
     <bundle-symbolic-name>${project.groupId}.infinispan.remote.query</bundle-symbolic-name>
-    <infinispan.docker.image.version>9.4.8.Final</infinispan.docker.image.version>
+    <infinispan.docker.image.version>11.0.9.Final</infinispan.docker.image.version>
```
olamy, line 12: use the property `${infinispan.version}`?

olamy (approved):

javadoc/pom.xml
```diff
@@ -427,7 +433,7 @@
     <dependency>
       <groupId>org.infinispan.protostream</groupId>
       <artifactId>protostream</artifactId>
-      <version>4.2.2.Final</version>
+      <version>4.3.4.Final</version>
```
olamy: we should have this with the version in dptMngt of the parent pom and get rid of version in child pom

olamy (changes requested):

jetty-infinispan/infinispan-remote-query/pom.xml
janbartel: @olamy I can't use the `infinispan.version` property because there is no docker image at that version.

olamy: arghhh :(

olamy (dismissed):

jetty-infinispan/infinispan-remote-query/pom.xml
joakime: The docker image exists as `11.0.9.Final` ...

https://hub.docker.com/layers/infinispan/server/11.0.9.Final/images/sha256-3dfa71e88775f03cd3a6aa9b7f7618d0518ff10f6ad68b4331d5d4c997bea6be

olamy (approved):

PR #6768, opened 2021-09-10
Change dependabot interval to daily

joakime:
+ Only chose to change branches 9.4.x and 10.0.x to daily.
  Leaving 11.0.x at weekly, so that it's updates can trail behind the 10.0.x ones that are merged forward to 11.0.x

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,13 +3,13 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "daily"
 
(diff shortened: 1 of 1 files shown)


sbordet (dismissed):

.github/dependabot.yml
```diff
@@ -3,13 +3,13 @@ updates:
   - package-ecosystem: "github-actions"
```
olamy: we should remove this as we do not use any `github-actions`

olamy (changes requested):

olamy (approved):

PR #6839, opened 2021-09-15
Issue #3514 interpolation of mod files to avoid hardcoding version in mod files

### LINKED ISSUE #3514: Use interpolation of versions from pom in mod files
Currently all mod files have a hardcoded version of libraries we used. So they can be out dated compared from version we use in test because require some manual changes.
We can can use maven interpolation to get the exact same version as used in test and declared in poms

olamy:
to prevent issues with the current usage of `${}` in mod files. Interpolation of values from pom will be based on the delimiter `@....@` such `@log4j2.version@`


jetty-home/src/main/resources/modules/jolokia.mod
```diff
@@ -15,9 +15,12 @@ jmx
 etc/jolokia.xml
 
 [files]
-maven://org.jolokia/jolokia-war/1.3.3/war|lib/jolokia/jolokia.war
+maven://org.jolokia/jolokia-war/${jolokia.version}/war|lib/jolokia/jolokia.war
```
janbartel, line 18: Is this correct? Why is this `${jolokia.version}` here, but `@jolokia.version@` on line 22?

jetty-home/pom.xml
```diff
@@ -865,6 +871,46 @@
       <type>jar</type>
       <optional>true</optional>
     </dependency>
+    <!-- adding those dependencies to ensure they are resolved and even kick upgrade from dependabot -->
+    <dependency>
+      <groupId>io.hawt</groupId>
+      <artifactId>hawtio-default</artifactId>
+      <version>${hawtio.version}</version>
```
janbartel: Does the `version` need to be referenced here? All of the other added dependencies don't use the `version`.

pom.xml
```diff
@@ -479,7 +483,7 @@
             <dependency>
               <groupId>org.eclipse.jetty.toolchain</groupId>
               <artifactId>jetty-assembly-descriptors</artifactId>
-              <version>1.0</version>
+              <version>1.1-SNAPSHOT</version>
```
janbartel: Once the main build is happy, you'll have to make a release of the `jetty-assembly-descriptors`.

jetty-home/src/main/resources/modules/jamon.mod
```diff
@@ -17,8 +17,8 @@ etc/jamon.xml
 
 [files]
 lib/jamon/
-maven://com.jamonapi/jamon/2.81|lib/jamon/jamon-2.81.jar
-maven://com.jamonapi/jamon_war/2.81/war|lib/jamon/jamon.war
+maven://com.jamonapi/jamon/${jamon.version}|lib/jamon/jamon-${jamon.version}.jar
```
janbartel, line 20: Should this be `@jamon.version@`?

janbartel (dismissed):
Need to also do: memcached and mongo.

jetty-home/src/main/resources/modules/jolokia.mod
olamy: as explained in the PR description ;) 
``` 
to prevent issues with the current usage of ${} in mod files. 
Interpolation of values from pom will be based on the delimiter @....@ such @log4j2.version@
```
this file have the following section 
```
[ini]
jolokia.version?=@jolokia.version@
```
this section will be transformed to 
```
[ini]
jolokia.version?=3.5.4 (or whatever the value coming from the pom)
```
The idea is to keep the the option to run jetty using such parameter to test new jolokia version 
```
java -jar start.jar jolokia.version=6.4.4
```

jetty-home/pom.xml
olamy: it's not used elsewhere than in jetty-home module.
The idea is to declare it here (even if it's definitely not needed) to have some upgrade notification coming from our good friend dependabot :)

pom.xml
olamy: oh yup for sure

jetty-home/src/main/resources/modules/jamon.mod
olamy: same reason as jolokia

olamy:
memcached and mongo done

jetty-home/src/main/resources/modules/jolokia.mod
joakime: To me the use of `@blah@` is processed by the resource plugin filtering resulting in a "hardcoded" value at build time.
The use of `${blah}` is left alone.

pom.xml
```diff
@@ -1445,6 +1533,65 @@
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>config-template</id>
+      <activation>
+        <file>
+          <exists>src/main/config-template</exists>
```
joakime, line 1585: nice, only run if we have that kind of setup.

joakime (changes requested):
2 more mod files need conversion.

```
./demos/demo-jndi-webapp/src/main/config/modules/demo-jndi.mod:maven://org.eclipse.jetty.orbit/javax.mail.glassfish/1.4.1.v201005082020/jar|lib/ext/javax.mail.glassfish-1.4.1.v201005082020.jar
./demos/demo-jndi-webapp/src/main/config/modules/demo-jndi.mod:maven://jakarta.transaction/jakarta.transaction-api/1.3.2/jar|lib/ext/jakarta.transaction-api-1.3.2.jar
./jetty-home/src/main/resources/modules/jminix.mod:maven://org.jminix/jminix/1.1.0|lib/jminix/jminix-1.1.0.jar
./jetty-home/src/main/resources/modules/jminix.mod:https://maven.restlet.talend.com/org/restlet/org.restlet/1.1.5/org.restlet-1.1.5.jar|lib/jminix/org.restlet-1.1.5.jar
./jetty-home/src/main/resources/modules/jminix.mod:https://maven.restlet.talend.com/org/restlet/org.restlet.ext.velocity/1.1.5/org.restlet.ext.velocity-1.1.5.jar|lib/jminix/org.restlet.ext.velocity-1.1.5.jar
./jetty-home/src/main/resources/modules/jminix.mod:maven://org.apache.velocity/velocity/1.5|lib/jminix/velocity-1.5.jar
./jetty-home/src/main/resources/modules/jminix.mod:maven://oro/oro/2.0.8|lib/jminix/oro-2.0.8.jar
./jetty-home/src/main/resources/modules/jminix.mod:https://maven.restlet.talend.com/com/noelios/restlet/com.noelios.restlet/1.1.5/com.noelios.restlet-1.1.5.jar|lib/jminix/com.noelios.restlet-1.1.5.jar
./jetty-home/src/main/resources/modules/jminix.mod:https://maven.restlet.talend.com/com/noelios/restlet/com.noelios.restlet.ext.servlet/1.1.5/com.noelios.restlet.ext.servlet-1.1.5.jar|lib/jminix/com.noelios.restlet.ext.servlet-1.1.5.jar
./jetty-home/src/main/resources/modules/jminix.mod:maven://net.sf.json-lib/json-lib/2.2.3/jar/jdk15|lib/jminix/json-lib-2.2.3-jdk15.jar
./jetty-home/src/main/resources/modules/jminix.mod:maven://commons-lang/commons-lang/2.4|lib/jminix/commons-lang-2.4.jar
./jetty-home/src/main/resources/modules/jminix.mod:maven://commons-beanutils/commons-beanutils/1.7.0|lib/jminix/commons-beanutils-1.7.0.jar
./jetty-home/src/main/resources/modules/jminix.mod:maven://commons-collections/commons-collections/3.2|lib/jminix/commons-collections-3.2.jar
./jetty-home/src/main/resources/modules/jminix.mod:maven://net.sf.ezmorph/ezmorph/1.0.6|lib/jminix/ezmorph-1.0.6.jar
./jetty-home/src/main/resources/modules/jminix.mod:maven://org.jgroups/jgroups/2.12.1.3.Final|lib/jminix/jgroups-2.12.1.3.Final.jar
./jetty-home/src/main/resources/modules/jminix.mod:maven://org.jasypt/jasypt/1.8|lib/jminix/jasypt-1.8.jar
```

jetty-home/src/main/resources/modules/jolokia.mod
olamy: yes I didn't want to touch anything about `${bla}` because already used in mod files.

olamy:
@joakime to be honest I would even prefer remove jminix as it's not really maintained and hawtio is providing the same features (and it's maintained)

jetty-home/src/main/resources/modules/jolokia.mod
janbartel: I don't think this difference is documented anywhere? So you're saying that if the substitution happens at runtime, like when the start mechanism processes the .mod/.ini file then you use `${var}`, but if it happens at build time, then you use `@var@`??? That's a pretty subtle difference, think that needs a good explanation somewhere.

jetty-home/pom.xml
janbartel: I'm confused. It's declared in the `dependencyManagement` section of the top level pom, so why does the `version` need to be declared here?

jetty-home/src/main/resources/modules/jolokia.mod
joakime: @janbartel The `@var@` syntax is standard Maven Resource handling with filtering.

* The `@var@` syntax is compile time expanded during the compile/package steps.
* The `${var}` syntax is runtime expanded, and in the specific cases on this PR, only by jetty-start.

jetty-home/src/main/resources/modules/jamon.mod
joakime: We want to allow people at runtime to change jamon.version, we don't expand it at compile time.

jetty-home/src/main/resources/modules/jolokia.mod
joakime: Btw, we use the same `@var@` syntax in our `jetty-maven-plugin/src/it/*` and `jetty-jspc-maven-plugin/src/it/*` test cases.
Also, the `@var@` syntax is the most common form of maven resource filtering (during the various resource phases of the build).

joakime (dismissed):
I'm happy with this as it sits.

I'll go ahead and merge up the conflicts from the infinispan PR into this.

pom.xml
joakime: I'll go ahead and release 1.1 of jetty-assembly-descriptors (as I feel this is done and working)

joakime: Done, it is now on maven central.
https://repo1.maven.org/maven2/org/eclipse/jetty/toolchain/jetty-assembly-descriptors/1.1/

jetty-home/pom.xml
olamy: yup fixed

olamy:
demo-jndi.mod fixed.
not sure we want to fix jminix but deprecate/remove it. (see https://github.com/eclipse/jetty.project/issues/6879)

joakime (approved):

PR #6859, opened 2021-09-16
Cleaning up maven properties

joakime:
+ Ensuring all dependencies and plugins have properties
+ Organized in top level pom (by purpose and then alphabetized)
+ All dependency properties end in ".version"
+ All plugin properties end in ".plugin.version"

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,66 +12,116 @@
   <inceptionYear>1995</inceptionYear>
 
   <properties>
-    <compiler.source>1.8</compiler.source>
-    <compiler.target>1.8</compiler.target>
+    <!-- build -->
+    <maven.compiler.source>8</maven.compiler.source>
+    <maven.compiler.target>8</maven.compiler.target>
     <jetty.url>https://eclipse.org/jetty</jetty.url>
+    <jmhjar.name>benchmarks</jmhjar.name>
+    <jpms-module-name>${bundle-symbolic-name}</jpms-module-name>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -12,66 +12,116 @@
...
+    <versions.plugin.version>2.8.1</versions.plugin.version>
 
     <!-- testing -->
+    <awaitility.version>4.1.0</awaitility.version>
     <it.debug>false</it.debug>
     <jetty.test.version>5.8</jetty.test.version>
-    <spring-boot.version>2.1.1.RELEASE</spring-boot.version>
+    <jetty.testtracker.log>false</jetty.testtracker.log>
     <jsr250-api.version>1.0</jsr250-api.version>
-    <javax.annotation-api>1.3.2</javax.annotation-api>
```
olamy, line 69: this need to updated in some maven plugin its as well

olamy (changes requested):

pom.xml
joakime: It's there in line 44 of the main pom.xml

https://github.com/eclipse/jetty.project/blob/505874110edcd040aebd3c52c9de9f0ec03112b0/pom.xml#L44

olamy:
Yes but annotation.api and not annotation-api ;) which I guess the ITs are
using

joakime:
I'm fixing the `jetty-maven-plugin` it tests now.

olamy (approved):

PR #6861, opened 2021-09-17
Fix #6860 IPv6 Format

### LINKED ISSUE #6860: IPv6 format
**Jetty version(s)**
9.4.x

**Description**

After fix #5128 for #5079, it is no longer possible to configure how an IPv6 address is returned from the servlet-api.  Some applications are expecting the address to be bracketed, others without brackets and yet others might depend on receiving the address exactly how it is received.   This is causing problems on PAAS/SAAS providers that cannot modify the applications.  We need a per context configuration to control IPv6 format.

gregw:
Fix #6860 IPv6 format by adding a per context configuration.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
@@ -86,7 +86,6 @@ import org.eclipse.jetty.server.session.Session;
 import org.eclipse.jetty.server.session.SessionHandler;
 import org.eclipse.jetty.util.Attributes;
 import org.eclipse.jetty.util.AttributesMap;
-import org.eclipse.jetty.util.HostPort;
 import org.eclipse.jetty.util.IO;
 import org.eclipse.jetty.util.MultiMap;
 import org.eclipse.jetty.util.StringUtil;
@@ -998,7 +997,7 @@ public class Request implements HttpServletRequest
                 String name = InetAddress.getLocalHost().getHostAddress();
                 if (StringUtil.ALL_INTERFACES.equals(name))
                     return null;
-                return formatAddrOrHost(name);
+                return _channel.formatAddrOrHost(name);
             }
             catch (UnknownHostException e)
             {
@@ -1014,7 +1013,7 @@ public class Request implements HttpServletRequest
         String result = address == null
             ? local.getHostString()
             : address.getHostAddress();
-        return formatAddrOrHost(result);
+        return _channel.formatAddrOrHost(result);
     }
 
     /*
@@ -1027,7 +1026,7 @@ public class Request implements HttpServletRequest
         {
             InetSocketAddress local = _channel.getLocalAddress();
             if (local != null)
-                return formatAddrOrHost(local.getHostString());
+                return _channel.formatAddrOrHost(local.getHostString());
         }
 
         try
@@ -1035,7 +1034,7 @@ public class Request implements HttpServletRequest
             String name = InetAddress.getLocalHost().getHostName();
             if (StringUtil.ALL_INTERFACES.equals(name))
                 return null;
-            return formatAddrOrHost(name);
+            return _channel.formatAddrOrHost(name);
         }
         catch (UnknownHostException e)
         {
@@ -1260,7 +1259,7 @@ public class Request implements HttpServletRequest
                 ? remote.getHostString()
                 : address.getHostAddress();
 
-        return formatAddrOrHost(result);
+        return _channel.formatAddrOrHost(result);
     }
 
     /*
@@ -1276,7 +1275,7 @@ public class Request implements HttpServletRequest
             return "";
 
         // We want the URI host, so add IPv6 brackets if necessary.
-        return formatAddrOrHost(remote.getHostString());
+        return _channel.formatAddrOrHost(remote.getHostString());
     }
 
     /*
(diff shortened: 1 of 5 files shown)


lachlan-roberts:
@gregw I pushed a fix to this branch, it had about 60 checkstyle failures.
I think you laptop is not configured with the Jetty intellij code style.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -292,6 +300,17 @@ public Context getServletContext()
         return _scontext;
     }
 
+    @ManagedAttribute("The format of IPv6 addresses return from the servlet API")
```
sbordet: "return" -> "returned"

jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -292,6 +300,17 @@ public Context getServletContext()
...
+    public IPV6Format getIpv6Format()
```
sbordet: Since the class is called `IPV6Format` (capital `P`), this method should also be `getIPv6Format()` with capital `P`.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -292,6 +300,17 @@ public Context getServletContext()
...
+    {
+        return _ipv6Format;
+    }
+
+    public void setIpv6Format(IPV6Format ipv6Format)
```
sbordet: Capital `P`.

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -2604,4 +2603,21 @@ else if (oldQuery == null)
...
+            return addr;
+        if (getContext() == null || getContext().getContextHandler() == null)
+            return HostPort.normalizeHost(addr);
+        switch (getContext().getContextHandler().getIpv6Format())
+        {
+            case BRACKETED:
+                return HostPort.normalizeHost(addr);
+            case UNBRACKETED:
+                return HostPort.denormalizeHost(addr);
+            default:
```
sbordet: I'd rather have the explicit case, and a `default` case that throws, in case in future we add other formats.

sbordet (changes requested):

sbordet:
@gregw this PR is missing tests, I fear that the "unchanged" semantic may not work as the field could already be set "normalized".

lachlan-roberts:
@gregw this still doesn't work too well with `ForwardedRequestCustomizer`, the problem is that it goes through `HostPort` first which will bracket the IPv6 address regardless of the `IPv6Format`. The other difference is that 9.3 treats the last segment of an unbracketed address as the port and 9.4 does not. So the behaviour is now:
```
=== Jetty 9.4 ===
UNBRACKETED: 1:2:3:4:5:6:7:8:8080 --> 1:2:3:4:5:6:7:8:8080
UNCHANGED:   1:2:3:4:5:6:7:8:8080 --> [1:2:3:4:5:6:7:8:8080]

=== Jetty 9.3 ===
1:2:3:4:5:6:7:8:8080 --> 1:2:3:4:5:6:7:8
```

gregw:
@lachlan-roberts I'm open to suggestions?  But I'm not sure that 1:2:3:4:5:6:7:8:8080 is really ever valid?

gregw:
So 9.3 just assumed that the last : always indicated a port.   I think this was a bad/wrong/dangerous thing.  So I'm not sure we should be bug for bug compatible with that.

lachlan-roberts:
> So 9.3 just assumed that the last : always indicated a port. I think this was a bad/wrong/dangerous thing. So I'm not sure we should be bug for bug compatible with that.

That's fine. But I think the main issue is that `HostPort` already normalizes the address. This means even using `ForwardedRequestCustomizer` the `IPv6Format.UNCHANGED` acts the same as `IPv6Format.BRACKETED`.

gregw:
Which is why we have the UNBRACKETED option.  I don't think we can do anything else as this is a per webapp setting and the customizer is run before the webapp is known.
I think if we run with either UNBRACKETED or UNCHANGED then the sponsered use-case will work fine.

gregw:
Both @lachlan-roberts and I were not totally happy that UNCHANGED didn't really work as we frequently normalize elsewhere.  So instead of introducing new forever public configuration, this change now is the minimal that we can do to make the formatting in an easily extensible class for the sponsor.  @sbordet please re-review.

lachlan-roberts:
@gregw this PR has failures in `DoSFilterTest`:

```
[INFO] Running org.eclipse.jetty.servlets.DoSFilterTest
Running org.eclipse.jetty.servlets.DoSFilterTest.testRemotePortLoadIdCreationIpv6()
Expected: (is "[::192.9.5.5]:12345" or is "[0:0:0:0:0:0:c009:505]:12345")
     but: was "0:0:0:0:0:0:c009:505:12345"
```

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -2604,4 +2603,9 @@ else if (oldQuery == null)
     {
         throw new ServletException("HttpServletRequest.upgrade() not supported in Jetty");
     }
+
+    private String formatAddrOrHost(String name)
+    {
+        return _channel == null ? HostPort.normalizeHost(name) : _channel.formatAddrOrHost(name);
```
sbordet, line 2609: `_channel` is a `final` field. Can it ever be `null`?
If it is only `null` in tests, you may want to add a comment about that.

sbordet (approved):
LGTM apart a glitch in `Request.formatAddrOrHost()`.

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
gregw: It should only be null in tests, but we have a specific test to check it works with null.

PR #6867, opened 2021-09-17
Cleanup redundant transitive dependency usage/declarations

joakime:
Revert the `maven-enforcer-plugin` to `3.0.0-M3`

Disabled `<requireUpperBoundDeps>` configuration.

Eliminate redundant transitive dependency declarations.

### DIFF
--- a/jetty-jspc-maven-plugin/pom.xml
+++ b/jetty-jspc-maven-plugin/pom.xml
@@ -76,19 +76,23 @@
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-plugin-api</artifactId>
+      <version>${maven.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.maven.plugin-tools</groupId>
       <artifactId>maven-plugin-annotations</artifactId>
+      <version>${maven.plugin-tools.version}</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-artifact</artifactId>
+      <version>${maven.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.maven.plugin-tools</groupId>
       <artifactId>maven-plugin-tools-api</artifactId>
+      <version>${maven.plugin-tools.version}</version>
       <exclusions>
         <exclusion>
           <groupId>junit</groupId>
--- a/jetty-maven-plugin/pom.xml
+++ b/jetty-maven-plugin/pom.xml
@@ -106,27 +106,32 @@
     <dependency>
       <groupId>org.apache.maven.shared</groupId>
       <artifactId>maven-artifact-transfer</artifactId>
-      <version>0.13.1</version>
+      <version>${maven-artifact-transfer.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-plugin-api</artifactId>
+      <version>${maven.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-artifact</artifactId>
+      <version>${maven.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-core</artifactId>
+      <version>${maven.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.maven.plugin-tools</groupId>
       <artifactId>maven-plugin-tools-api</artifactId>
+      <version>${maven.plugin-tools.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.maven.plugin-tools</groupId>
       <artifactId>maven-plugin-annotations</artifactId>
+      <version>${maven.plugin-tools.version}</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
--- a/jetty-unixsocket/pom.xml
+++ b/jetty-unixsocket/pom.xml
@@ -11,6 +11,7 @@
   <properties>
     <bundle-symbolic-name>${project.groupId}.unixsocket</bundle-symbolic-name>
     <spotbugs.onlyAnalyze>org.eclipse.jetty.unixsocket.*</spotbugs.onlyAnalyze>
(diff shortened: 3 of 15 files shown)


janbartel (changes requested):
I am definitely in favour of reverting to `3.0.0-M3` for the `enforcer` plugin.

I don't think we need to remove the `enforceUpperBound` rule, _unless_ that too was added at the same time as the upgrade to `3.0.0`.

I'm not convinced about adding in all these other changes because a) some of them seem to go against the convention of moving every dependency into the `dependencyManagement` section of the pom that I thought we were supposed to be striving for and b) the build of this branch doesn't work and it will be difficult to work out why.

I think we should just do a simple PR to change back the version of the `enforcer` plugin, so we are back to a known working configuration,  then do a subsequent PR with a cleanup of the dependencies.

jetty-jspc-maven-plugin/pom.xml
```diff
@@ -76,19 +76,23 @@
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-plugin-api</artifactId>
+      <version>${maven.version}</version>
```
olamy, line 79: no need of this as it's already declared in dptMngt in top level

jetty-maven-plugin/pom.xml
```diff
@@ -106,27 +106,32 @@
     <dependency>
       <groupId>org.apache.maven.shared</groupId>
       <artifactId>maven-artifact-transfer</artifactId>
-      <version>0.13.1</version>
+      <version>${maven-artifact-transfer.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-plugin-api</artifactId>
+      <version>${maven.version}</version>
```
olamy, line 114: not needed

pom.xml
```diff
@@ -846,26 +849,11 @@
         <artifactId>javax.annotation-api</artifactId>
         <version>${javax.annotation.api.version}</version>
       </dependency>
-      <dependency>
-        <groupId>org.ow2.asm</groupId>
-        <artifactId>asm</artifactId>
-        <version>${asm.version}</version>
```
olamy, line 852: we use it directly and having this here in dptMngt will insure coherence between modules especially for transitive dependencies as it;s the role of dependencyManagement section.
See
https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Management

```
A second, and very important use of the dependency management section is 
to control the versions of artifacts used in transitive dependencies.
```

pom.xml
```diff
@@ -911,53 +899,7 @@
         <version>${javax.transaction.api.version}</version>
         <scope>provided</scope>
       </dependency>
-      <!-- maven deps -->
```
olamy, line 914: we use it directly and having this here in dptMngt will insure coherence between modules

olamy (changes requested):

jetty-unixsocket/pom.xml
```diff
@@ -11,6 +11,7 @@
   <properties>
     <bundle-symbolic-name>${project.groupId}.unixsocket</bundle-symbolic-name>
     <spotbugs.onlyAnalyze>org.eclipse.jetty.unixsocket.*</spotbugs.onlyAnalyze>
+    <jnr-unixsocket.version>0.38.10</jnr-unixsocket.version>
```
olamy, line 14: I would prefer having this in single place (i.e top pom)

pom.xml
```diff
@@ -91,7 +91,8 @@
     <maven.compiler.plugin.version>3.8.1</maven.compiler.plugin.version>
     <maven.dependency.plugin.version>3.2.0</maven.dependency.plugin.version>
     <maven.eclipse.plugin.version>2.10</maven.eclipse.plugin.version>
-    <maven.enforcer.plugin.version>3.0.0</maven.enforcer.plugin.version>
+    <maven.enforcer.plugin.version>3.0.0-M3</maven.enforcer.plugin.version>
```
olamy, line 94: :(

pom.xml
```diff
@@ -220,11 +223,11 @@
                 <versionOsgiRule implementation="org.eclipse.jetty.toolchain.enforcer.rules.RequireOsgiCompatibleVersionRule" />
                 <versionRedhatRule implementation="org.eclipse.jetty.toolchain.enforcer.rules.RequireRedhatCompatibleVersionRule" />
                 <versionDebianRule implementation="org.eclipse.jetty.toolchain.enforcer.rules.RequireDebianCompatibleVersionRule" />
-                <requireUpperBoundDeps>
+                <!--<requireUpperBoundDeps>
```
olamy, line 226: :(

pom.xml
```diff
@@ -1274,6 +1150,7 @@
       <properties>
         <settingsPath>${env.GLOBAL_MVN_SETTINGS}</settingsPath>
         <surefire.rerunFailingTestsCount>3</surefire.rerunFailingTestsCount>
+        <spotbugs.plugin.version>4.3.0</spotbugs.plugin.version>
```
olamy, line 1153: I would prefer having this in single place (top of the pom)

tests/test-sessions/test-infinispan-sessions/pom.xml
```diff
@@ -103,42 +103,6 @@
...
-      <version>${infinispan.version}</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.infinispan</groupId>
-      <artifactId>infinispan-client-hotrod</artifactId>
-      <version>${infinispan.version}</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
```
olamy, line 124: some part of the code use so must be declare

olamy (changes requested):

pom.xml
olamy: part of documentation here https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Management 
```
The dependency management section is a mechanism for centralizing dependency information. 
When you have a set of projects that inherit from a common parent, 
it's possible to put all information about the dependency in the common POM and have simpler references 
to the artifacts in the child POMs.
```

PR #6872, opened 2021-09-19
Fix #6868 Fix copy constructors

### LINKED ISSUE #6868: _uriCompliance attribute has been forgotten in the HttpConfiguration constructor from another configuration
**Jetty version(s)**
10.0.6 and 10.0.7-SNAPSHOT

any

**OS type/version**
any

**Description**
In the constructor `public HttpConfiguration(HttpConfiguration config)`, the `_uriCompliance` is not copied from the original.

**How to reproduce?**
setting the jetty.httpConfig.uriCompliance to LEGACY in start.d/server.ini has no effect on the HttpConfiguration used for https

gregw:
Fix #6868 copy constructors found by static code analysis

### DIFF
--- a/jetty-webapp/src/main/java/org/eclipse/jetty/webapp/ClassMatcher.java
+++ b/jetty-webapp/src/main/java/org/eclipse/jetty/webapp/ClassMatcher.java
@@ -496,10 +496,16 @@ public class ClassMatcher extends AbstractSet<String>
     {
     }
 
-    public ClassMatcher(ClassMatcher patterns)
+    public ClassMatcher(ClassMatcher matcher)
     {
-        if (patterns != null)
-            setAll(patterns.getPatterns());
+        if (matcher != null)
+        {
+            _entries.putAll(matcher._entries);
+            _patterns.include(matcher._patterns.getIncluded().toArray(new Entry[0]));
+            _patterns.exclude(matcher._patterns.getExcluded().toArray(new Entry[0]));
+            _locations.include(matcher._locations.getIncluded().toArray(new Entry[0]));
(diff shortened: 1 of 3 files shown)


jetty-webapp/src/main/java/org/eclipse/jetty/webapp/ClassMatcher.java
```diff
@@ -496,10 +496,16 @@ public ClassMatcher()
...
-        if (patterns != null)
-            setAll(patterns.getPatterns());
+        if (matcher != null)
+        {
+            _entries.putAll(matcher._entries);
+            _patterns.include(matcher._patterns.getIncluded().toArray(new Entry[0]));
+            _patterns.exclude(matcher._patterns.getExcluded().toArray(new Entry[0]));
+            _locations.include(matcher._locations.getIncluded().toArray(new Entry[0]));
+            _locations.exclude(matcher._locations.getExcluded().toArray(new Entry[0]));
+        }
```
lachlan-roberts: The `ClassMatcher.add()` method actually sets `_locations` and `_patterns` anyway.

But I think this is equivalent behaviour, unless someone has a subclass of `ClassMatcher` to override `add(Entry)`.

lachlan-roberts (dismissed):

jetty-webapp/src/main/java/org/eclipse/jetty/webapp/ClassMatcher.java
gregw: yeah, it was more to keep the static analysis quiet so we can get a clean test.
Hmmm but your point about add is valid.   I'll see if I can suppress the analysis.

lachlan-roberts (approved):

PR #6885, opened 2021-09-21
Fix #6860 IPv6 Format (#6861 for jetty-10)

### LINKED ISSUE #6860: IPv6 format

gregw:
Fix #6860 IPv6 format by adding an extensible HttpChannel method

cherry-pick to 10

Co-authored-by: Lachlan Roberts <lachlan@webtide.com>

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
@@ -985,7 +985,7 @@ public class Request implements HttpServletRequest
                 String name = InetAddress.getLocalHost().getHostAddress();
                 if (StringUtil.ALL_INTERFACES.equals(name))
                     return null;
-                return HostPort.normalizeHost(name);
+                return formatAddrOrHost(name);
             }
             catch (UnknownHostException e)
             {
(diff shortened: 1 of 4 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -1322,20 +1321,20 @@ public String getScheme()
     @Override
     public String getServerName()
     {
-        return _uri == null ? findServerName() : _uri.getHost();
+        return _uri == null ? findServerName() : formatAddrOrHost(_uri.getHost());
```
lachlan-roberts, line 1324: We were not normalising before here and now we are. So not sure if this will impact anything.

lachlan-roberts (approved):
One note about a change in `Request.getServerName()`, other than that LGTM.

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
gregw: The previous impl was inconsistent as we normalise in `findServerName()` but not in `_uri.getHost()`, although the later is probably already normalised.  I think this is a low risk change for the better.

PR #6886, opened 2021-09-21
Fix #6883 relative welcome redirect

### LINKED ISSUE #6883: Welcome file redirects do not honor the relativeRedirectAllowed option
**Jetty version(s)**

9.4.43.v20210629

openjdk version "11.0.12" 2021-07-20 LTS

**OS type/version**

macOS Big Sur 11.6 (20G165)

**Description**

Last year #5029 Relative Redirection introduced a new `relativeRedirectAllowed` option. (Thanks!) Unfortunately, at least some of the redirects that Jetty itself initiates do not yet honor that option.

For instance, `org.eclipse.jetty.server.ResourceService.sendWelcome` (the method that redirects to add a missing trailing slash if necessary) builds an absolute redirect URL. But `org.eclipse.jetty.server.Response` only honors `HttpConfiguration.isRelativeRedirectAllowed()` if the redirect URL is relative (includes no scheme).

This behavior is problematic for us because in some environments, we use a reverse proxy or load balancer in front of Jetty for SSL termination. As a result, https: requests look to Jetty like http: requests, so it ends up redirecting some https: requests to http: in order to add a trailing slash.

**How to reproduce?**

1. Allow relative redirects by calling `httpConfiguration.setRelativeRedirectAllowed(true)`.
2. Register a servlet at a path like "/docs/*".
3. Issue a GET /docs request.
4. Notice that the Location header in the response contains an absolute URL (containing a scheme and host) instead of just a relative URL such as just "/docs/".

gregw:
Fix #6883 relative welcome redirect
 + make all redirects able to be relative
 + added test for relative redirection in ResourceService

### DIFF
--- a/jetty-http/src/test/java/org/eclipse/jetty/http/matchers/HttpFieldsContainsHeaderValue.java
+++ b/jetty-http/src/test/java/org/eclipse/jetty/http/matchers/HttpFieldsContainsHeaderValue.java
@@ -18,7 +18,7 @@
 
 package org.eclipse.jetty.http.matchers;
 
-import java.util.Locale;
+import java.util.Objects;
 
 import org.eclipse.jetty.http.HttpField;
 import org.eclipse.jetty.http.HttpFields;
@@ -55,21 +55,7 @@ public class HttpFieldsContainsHeaderValue extends TypeSafeMatcher<HttpFields>
         if (field == null)
             return false;
 
-        // Use HttpField.contains() logic
-        if (field.contains(this.value))
-            return true;
-
-        // Simple equals
-        if (this.value == field.getValue())
-            return true;
-
-        // Try individual value logic
-        String lcValue = this.value.toLowerCase(Locale.ENGLISH);
-        for (String value : field.getValues())
-        {
-            if (value.toLowerCase(Locale.ENGLISH).contains(lcValue))
-                return true;
-        }
-        return false;
+        // exact match or HttpField.contains match (which is not just a simple substring).
+        return Objects.equals(this.value, field.getValue()) || field.contains(this.value);
     }
 }
--- a/jetty-http/src/test/java/org/eclipse/jetty/http/HttpFieldsMatchers.java
+++ b/jetty-http/src/test/java/org/eclipse/jetty/http/HttpFieldsMatchers.java
@@ -20,6 +20,7 @@ package org.eclipse.jetty.http;
 
 import org.eclipse.jetty.http.matchers.HttpFieldsContainsHeaderKey;
 import org.eclipse.jetty.http.matchers.HttpFieldsContainsHeaderValue;
+import org.eclipse.jetty.http.matchers.HttpFieldsHeaderValue;
 import org.hamcrest.Matcher;
 
 public class HttpFieldsMatchers
@@ -32,6 +33,11 @@ public class HttpFieldsMatchers
     public static Matcher<HttpFields> containsHeader(HttpHeader header)
     {
         return new HttpFieldsContainsHeaderKey(header);
+    }
+
+    public static Matcher<HttpFields> headerValue(String keyName, String value)
+    {
+        return new HttpFieldsHeaderValue(keyName, value);
     }
 
     public static Matcher<HttpFields> containsHeaderValue(String keyName, String value)
--- a/jetty-http/src/test/java/org/eclipse/jetty/http/matchers/HttpFieldsHeaderValue.java
+++ b/jetty-http/src/test/java/org/eclipse/jetty/http/matchers/HttpFieldsHeaderValue.java
@@ -18,7 +18,7 @@
 
 package org.eclipse.jetty.http.matchers;
 
-import java.util.Locale;
+import java.util.Objects;
 
 import org.eclipse.jetty.http.HttpField;
 import org.eclipse.jetty.http.HttpFields;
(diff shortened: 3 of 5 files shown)


lachlan-roberts:
@gregw the new tests do not seem to test the original issue. I locally reverted the changes to `ResourceService` and all the new tests in `DefaultServletTest` still passed.

gregw:
@lachlan-roberts good catch.  There is a bug in `HttpFieldsContainsHeaderValue`, it just looks for a substring rather than a proper containment.
I'll see if I can fix that... else I'll need another matcher.

gregw:
@lachlan-roberts I've pushed a fix for that, but there is a good chance it might break some other tests that were relying on the wrong behavior

jetty-http/src/test/java/org/eclipse/jetty/http/matchers/HttpFieldsContainsHeaderValue.java
```diff
@@ -55,21 +55,7 @@ protected boolean matchesSafely(HttpFields fields)
...
-            return true;
-
-        // Try individual value logic
-        String lcValue = this.value.toLowerCase(Locale.ENGLISH);
-        for (String value : field.getValues())
-        {
-            if (value.toLowerCase(Locale.ENGLISH).contains(lcValue))
-                return true;
-        }
-        return false;
```
gregw: @joakime Do you recall why you implemented this?   values are not case insensitive and a string contains is not a good enough check.
For example if you want to check for "xyz", then it should match `xyz`, `foo,xyz,bar` and `"xyz"`, but not `XYZ` or `fooxyzbar`.

joakime: This class was introduced in the junit5 revision - https://github.com/eclipse/jetty.project/commit/a3f1592c50c3af7001a5ab3c773768660eb8d07a
It is being used by multiple classes in that revision.

* [HpackDecoderTest](https://github.com/eclipse/jetty.project/blob/a3f1592c50c3af7001a5ab3c773768660eb8d07a/jetty-http2/http2-hpack/src/test/java/org/eclipse/jetty/http2/hpack/HpackDecoderTest.java)
* [HttpManyWaysToAsyncCommitTest](https://github.com/eclipse/jetty.project/blob/a3f1592c50c3af7001a5ab3c773768660eb8d07a/jetty-server/src/test/java/org/eclipse/jetty/server/HttpManyWaysToAsyncCommitTest.java)
* [DefaultServletTest](https://github.com/eclipse/jetty.project/blob/a3f1592c50c3af7001a5ab3c773768660eb8d07a/jetty-servlet/src/test/java/org/eclipse/jetty/servlet/DefaultServletTest.java)
* [DefaultHandlerTest](https://github.com/eclipse/jetty.project/blob/a3f1592c50c3af7001a5ab3c773768660eb8d07a/tests/test-integration/src/test/java/org/eclipse/jetty/test/DefaultHandlerTest.java)

Lets take a look on the revision just before that junit5 commit and see what the test logic was/is ...

* [HpackDecoderTest](https://github.com/eclipse/jetty.project/blob/0ae0c500f0057d5c2f09e266bf8b15953a83ce22/jetty-http2/http2-hpack/src/test/java/org/eclipse/jetty/http2/hpack/HpackDecoderTest.java)
https://github.com/eclipse/jetty.project/blob/0ae0c500f0057d5c2f09e266bf8b15953a83ce22/jetty-http2/http2-hpack/src/test/java/org/eclipse/jetty/http2/hpack/HpackDecoderTest.java#L182
https://github.com/eclipse/jetty.project/blob/0ae0c500f0057d5c2f09e266bf8b15953a83ce22/jetty-http/src/main/java/org/eclipse/jetty/http/HttpFields.java#L173-L182
* [HttpManyWaysToAsyncCommitTest](https://github.com/eclipse/jetty.project/blob/0ae0c500f0057d5c2f09e266bf8b15953a83ce22/jetty-server/src/test/java/org/eclipse/jetty/server/HttpManyWaysToAsyncCommitTest.java)
https://github.com/eclipse/jetty.project/blob/0ae0c500f0057d5c2f09e266bf8b15953a83ce22/jetty-server/src/test/java/org/eclipse/jetty/server/HttpManyWaysToAsyncCommitTest.java#L334
https://github.com/eclipse/jetty.project/blob/0ae0c500f0057d5c2f09e266bf8b15953a83ce22/jetty-server/src/test/java/org/eclipse/jetty/server/AbstractHttpTest.java#L116-L119
* [DefaultServletTest](https://github.com/eclipse/jetty.project/blob/0ae0c500f0057d5c2f09e266bf8b15953a83ce22/jetty-servlet/src/test/java/org/eclipse/jetty/servlet/DefaultServletTest.java)
https://github.com/eclipse/jetty.project/blob/0ae0c500f0057d5c2f09e266bf8b15953a83ce22/jetty-servlet/src/test/java/org/eclipse/jetty/servlet/DefaultServletTest.java#L930-L934
https://github.com/eclipse/jetty.project/blob/0ae0c500f0057d5c2f09e266bf8b15953a83ce22/jetty-servlet/src/test/java/org/eclipse/jetty/servlet/DefaultServletTest.java#L1445-L1449
* [DefaultHandlerTest](https://github.com/eclipse/jetty.project/blob/0ae0c500f0057d5c2f09e266bf8b15953a83ce22/tests/test-integration/src/test/java/org/eclipse/jetty/test/DefaultHandlerTest.java)
https://github.com/eclipse/jetty.project/blob/0ae0c500f0057d5c2f09e266bf8b15953a83ce22/tests/test-integration/src/test/java/org/eclipse/jetty/test/DefaultHandlerTest.java#L173

Nothing obvious why it's case insensitive ATM.

I think it's because `HttpField.contains(String)` back then did that??

https://github.com/eclipse/jetty.project/blob/0ae0c500f0057d5c2f09e266bf8b15953a83ce22/jetty-http/src/main/java/org/eclipse/jetty/http/HttpField.java#L110

joakime: Yeah, even our current `HttpField.contains(String)` does a the value check in a case insensitive way.

gregw: I've looked at `HttpField.contains(String)` and I have my concerns about that as well. So I just made another Matcher for an exact match.

joakime: > I've looked at `HttpField.contains(String)` and I have my concerns about that as well. So I just made another Matcher for an exact match.

We should open a new Issue to review that then.

joakime (approved):

PR #6887, opened 2021-09-22
Issue #4317 - exclude text/event-stream MIME type from GzipHandler

### LINKED ISSUE #4317: EventSource does not work with GzipHandler
**Jetty version**
9.4.22.v20191022
**Java version**
11
**OS type/version**
Ubuntu
**Description**
When you enable gzip-module, the EventSource-mechanism does not work. Chrome and FF simply don't receive events. Hard to find the problem. As they send 'accept-encoding: gzip', i think it should be possible to make it work with gzip? But for now it would be better to make `text/event-stream` default-excluded in here:
```java
public GzipHandler()
    {
        _methods.include(HttpMethod.GET.asString());
        for (String type : MimeTypes.getKnownMimeTypes())
        {
            if ("image/svg+xml".equals(type))
                _paths.exclude("*.svgz");
            else if (type.startsWith("image/") ||
                type.startsWith("audio/") ||
                type.startsWith("video/"))
                _mimeTypes.exclude(type);
        }
        _mimeTypes.exclude("application/compress");
        _mimeTypes.exclude("application/zip");
        _mimeTypes.exclude("application/gzip");
        _mimeTypes.exclude("application/bzip2");
        _mimeTypes.exclude("application/brotli");
        _mimeTypes.exclude("application/x-xz");
        _mimeTypes.exclude("application/x-rar-compressed");
```

lachlan-roberts:
Issue #4317

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipHandler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipHandler.java
@@ -204,6 +204,9 @@ public class GzipHandler extends HandlerWrapper implements GzipFactory
         _mimeTypes.exclude("application/x-xz");
         _mimeTypes.exclude("application/x-rar-compressed");
 
+        // It is possible to SSE with GzipHandler but you will need to set _synFlush to true which will impact performance.
+        _mimeTypes.exclude("text/event-stream");
+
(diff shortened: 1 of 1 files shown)


lachlan-roberts:
Maybe we should leave this for 9.4 and only make the change in 10/11.

PR #6894, opened 2021-09-22
Fix #6883 relative welcome redirect (#6886)

### LINKED ISSUE #6883: Welcome file redirects do not honor the relativeRedirectAllowed option

gregw:
Cheery pick 9 to 10

Fix #6883 relative welcome redirect
 + make all redirects able to be relative
 + added test for relative redirection in ResourceService

### DIFF
--- a/tests/jetty-http-tools/src/main/java/org/eclipse/jetty/http/tools/matchers/HttpFieldsHeaderValue.java
+++ b/tests/jetty-http-tools/src/main/java/org/eclipse/jetty/http/tools/matchers/HttpFieldsHeaderValue.java
@@ -1,0 +1,60 @@
+//
+//  ========================================================================
+//  Copyright (c) 1995-2021 Mort Bay Consulting Pty Ltd and others.
+//  ------------------------------------------------------------------------
+//  All rights reserved. This program and the accompanying materials
+//  are made available under the terms of the Eclipse Public License v1.0
+//  and Apache License v2.0 which accompanies this distribution.
+//
+//      The Eclipse Public License is available at
+//      http://www.eclipse.org/legal/epl-v10.html
+//
+//      The Apache License v2.0 is available at
+//      http://www.opensource.org/licenses/apache2.0.php
+//
+//  You may elect to redistribute this code under either of these licenses.
+//  ========================================================================
+//
+
+package org.eclipse.jetty.http.matchers;
+
+import java.util.Objects;
+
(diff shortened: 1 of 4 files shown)


joakime:
Need to fix the license header

tests/jetty-http-tools/src/main/java/org/eclipse/jetty/http/tools/matchers/HttpFieldsHeaderValue.java
```diff
@@ -0,0 +1,60 @@
+//
+//  ========================================================================
+//  Copyright (c) 1995-2021 Mort Bay Consulting Pty Ltd and others.
+//  ------------------------------------------------------------------------
+//  All rights reserved. This program and the accompanying materials
+//  are made available under the terms of the Eclipse Public License v1.0
+//  and Apache License v2.0 which accompanies this distribution.
+//
+//      The Eclipse Public License is available at
+//      http://www.eclipse.org/legal/epl-v10.html
```
joakime: This is the wrong license header for Jetty 10.0.x

joakime (changes requested):
License header needs to be fixed

joakime (approved):

PR #6937, opened 2021-09-28
Changed order of entries in module-info.java to be canonical

### LINKED ISSUE #6938: module-info.java file do not use the canonical order for the elements
It seems that the canonical order of the elements inside the module-info.java file in the OpenJDK/src is always the same:

`module some.module.name {
requires ...

requires transitive ...

exports ...

exports ... to
...

opens ... to
...

uses ...

provides ... with
...
}`

The order in our module-info.java files were a bit muddled, with requires sometimes appearing near the end.

Furthermore, the entries are usually alphabetically sorted.

(There is one case in the OpenJDK where requires transitive is out of order, but that was because someone changed it and didn't reorder it. I've sent a pull-request there too, but it's so insignificant I almost doubt it will be accepted.)

See also
https://docs.oracle.com/javase/specs/jls/se17/html/jls-7.html#jls-ModuleDirective

kabutz:
It seems that the canonical order of the elements inside the module-info.java file in the OpenJDK/src is always the same:

`module some.module.name {
requires ...

requires transitive ...

exports ...

exports ... to
...

opens ... to
...

uses ...

provides ... with
...
}`

The order in our module-info.java files were a bit muddled, with requires sometimes appearing near the end.

Furthermore, the entries are usually alphabetically sorted.

(There is one case in the OpenJDK where requires transitive is out of order, but that was because someone changed it and didn't reorder it. I've sent a pull-request there too, but it's so insignificant I almost doubt it will be accepted.)

See also 
https://docs.oracle.com/javase/specs/jls/se17/html/jls-7.html#jls-ModuleDirective


kabutz:
@sbordet this touched a lot of files and there is a distinct possibility that I made a mistake somewhere, despite checking everything several times.

sbordet (approved):

Issue #4275, opened 2019-11-06
Path Normalization/Traversal - Context Matching


venukb:
We have an app with 2 servlet path defined

`/ws/v1`
`/foo`

`/ws/v1` was exposed on edge (ATS). However `/foo` accessed on Jetty using this 

`/ws/v1/..;/..;/foo` - so we had a path exposed on the edge which was not intended to be exposed

- Is this matching done as per servlet spec?
- Is there a configuration to avoid this in Jetty
- I've heard that Undertow for e.g. does not allow the above. 

(Jetty version: 9.4.22 - same behavior seen in older versions as well)

joakime:
The use of `;` in path segments is a URI with path parameters.

https://tools.ietf.org/html/rfc3986#section-3.3

As for how path parameters are to be treated.
That's an open issue on the servlet spec - https://github.com/eclipse-ee4j/servlet-api/issues/18

There's been discussion about how to handle normalization when the URI is complicated with path parameters, [URI templates](https://tools.ietf.org/html/rfc6570) variables, etc.
Right now, that's an undefined part of the servlet spec.

joakime:
Java `URI.normalize()` also doesn't handle path parameters during requests for normalization.

``` java
package uri;

import java.net.URI;

public class URITest
{
    public static void main(String[] args)
    {
        dump(URI.create("http://machine.com/ws/v1/..;/..;/foo"));
        dump(URI.create("http://machine.com/ws/v1/../../foo"));
    }

    public static void dump(URI uri)
    {
        System.out.println("URI = " + uri);
        System.out.println("   .normalize = " + uri.normalize());
    }
}
```

results in ...

``` plain
URI = http://machine.com/ws/v1/..;/..;/foo
   .normalize = http://machine.com/ws/v1/..;/..;/foo
URI = http://machine.com/ws/v1/../../foo
   .normalize = http://machine.com/foo
```

venukb:
Yup agree `;` is valid.
But `/ws/v1/..;/..;/foo` matches a servlet with url pattern `/foo` - 

Seems like this is being misused. Was checking to see if there were any controls/options in Jetty - to not match paths like this

i.e. in org.eclipse.jetty.server.Request

I see **__originalURI_** is `/ws/v1/..;/../foo` in this and **__servletPath_** is `/foo`.
Is this match also as per Servlet Spec?

gregw:
[...]

PR #6939, opened 2021-09-28
Improve #4275 ambiguous URIs

gregw:
#4275
A URI like `/foo/%2e%2e;/bar` should be ambiguous both because of the encoded dots and because of the parameters.  This means that the default setting of jetty-9 is a bit more secure as this path is considered ambiguous if either Violation.SEGMENT or Violation.PARAM is set.


jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -782,15 +782,14 @@ private void checkSegment(String uri, int segment, int end, boolean param)
...
         {
-            // The segment is always ambiguous.
-            _violations.add(Violation.SEGMENT);
-        }
-        else if (param && ambiguous == Boolean.FALSE)
-        {
-            // The segment is ambiguous only when followed by a parameter.
-            _violations.add(Violation.PARAM);
+            // Is the segment intrinsically ambiguous
+            if (ambiguous == Boolean.TRUE)
```
lachlan-roberts: Why do we even have `.` and `..` in the `Trie` if we only ever act on `Boolean.TRUE` returns?

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -782,15 +782,14 @@ private void checkSegment(String uri, int segment, int end, boolean param)
 
         // Look for segment in the ambiguous segment index.
         Boolean ambiguous = __ambiguousSegments.get(uri, segment, end - segment);
-        if (ambiguous == Boolean.TRUE)
+        if (ambiguous != null)
```
lachlan-roberts, line 785: Do we really want the `PARAM` check to be inside this block?
Then paths like `/path/test;param/info` are not ambiguous with `Violation.PARAM` even though they are decoded to `/path/test/info`. 

I can see tests like `{";/bar", "/bar", EnumSet.of(Violation.EMPTY)},` but not sure why we wouldn't classify this as ambiguous.

lachlan-roberts (changes requested):

lachlan-roberts:
@gregw there are test failures.

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
gregw: We act on FALSE as it gives a non-null return.

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
gregw: I don't understand?   `/path/test;param/info` is not ambiguous as it decodes to exactly `/path/test/info`.   Paths like `/path/..;param/info` and `/path/%2e%2e;param/info` are ambiguous because the RFC strictly says  they should not be normalized, but once params are stripped they could be.

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
gregw: Specifically, the logic is:
 + Any ambiguous segment that also has a param are `Violation.PARAM` (ie. they are ambiguous but may need to the parameter to make them so)
 + Only ambiguous segments that return `Boolean.TRUE` are `Violation.SEGMENT` (ie they are ambiguous with or without a parameter).

lachlan-roberts (dismissed):

jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
MacX19: Boolean.TRUE is not enum value. May be better 
`if (Boolean.TRUE.equals(ambiguous))`
https://stackoverflow.com/questions/16437236/boolean-true-myboolean-vs-boolean-true-equalsmyboolean

gregw:
@lachlan-roberts can you re-review.... I've been a little evil and snuck in a few extra fixes replacing == with equals

lachlan-roberts (approved):

PR #6949, opened 2021-09-30
Update CVE releated issues with CVE number

gregw:
It can be hard to find issues associated with CVEs after the fact.  We should update the issue description to include the CVE number once published.


lachlan-roberts:
The advisory is usually published well after all the PRs are merged. It is easy to lose track of what PRs and Issues are associated with a CVE even before the time of publishing. Maybe we should specify that there needs to be a list of the related Issues/PRs on the CVE advisory so we can keep it updated as we go.

gregw:
@sbordet nudge

gregw:
@lachlan-roberts Good idea! we should also modify security advisory template so we can keep track of issue numbers and PR numbers there, but I think we still need this step in the release process to update issues as well.

firoj07:
> It can be hard to find issues associated with CVEs after the fact. We should update the issue description to include the CVE number once published.

update-CVE-issues

sbordet (approved):

PR #6997, opened 2021-10-14
Issue #6973 - Setup Request/Response for success with RequestLog

### LINKED ISSUE #6973: Jetty starts consuming CPU that remains high even without any traffic
**Jetty version(s)**
`9.4.41.v20210516`
`9.4.43.v20210629`

```
$ java -version
openjdk version "1.8.0_292"
OpenJDK Runtime Environment (build 1.8.0_292-8u292-b10-0ubuntu1~18.04-b10)
OpenJDK 64-Bit Server VM (build 25.292-b10, mixed mode)
```
**OS type/version**
Ubuntu 18.04

**Description**
We had an issue with some of the production servers spiking with CPU use.
We do not know what triggered it. When affected servers are removed from load balancer and all traffic to them ceases, the CPU usage still remains high.

About 5 out of 8 servers got affected over a stretch of several hours. The servers belong to different clusters and run different application code on them. The common theme is that they all are Spring apps with embedded Jetty, are on Amazon EC2 and exposed to the internet via load balancers. They are in different AWS accounts, different load balancers and their load pattern is not correlated in any way. Just the AWS region is the same. We saw `9.4.41.v20210516` and `9.4.43.v20210629` affected. Restarting the app (with Jetty) fixes CPU usage.

We tried profiling the running app and building flamegraphs and saw that all the CPU is being eaten by a couple of "qtp" threads with stacktraces similar to:

```
"qtp708090483-69573" #69573 prio=5 os_prio=0 tid=0x00007f6b1e429000 nid=0x311f in Object.wait() [0x00007f6ac36c8000]
   java.lang.Thread.State: WAITING (on object monitor)
        at java.lang.Object.wait(Native Method)
        at java.lang.Object.wait(Object.java:502)
        at org.eclipse.jetty.server.HttpInput.blockForContent(HttpInput.java:584)
        at org.eclipse.jetty.server.HttpInput$1.blockForContent(HttpInput.java:1164)
        at org.eclipse.jetty.server.HttpInput.read(HttpInput.java:330)
        - locked <0x00000000c54d9790> (a java.util.ArrayDeque)
        at org.eclipse.jetty.server.HttpInput.read(HttpInput.java:270)
        at org.eclipse.jetty.util.UrlEncoded.decodeUtf8To(UrlEncoded.java:485)
        - locked <0x00000000c54dea40> (a org.eclipse.jetty.util.MultiMap)
        at org.eclipse.jetty.util.UrlEncoded.decodeTo(UrlEncoded.java:572)
        at org.eclipse.jetty.server.Request.extractFormParameters(Request.java:582)
        at org.eclipse.jetty.server.Request.extractContentParameters(Request.java:531)
        at org.eclipse.jetty.server.Request.getParameters(Request.java:435)
        at org.eclipse.jetty.server.Request.getParameterNames(Request.java:1093)
        at ch.qos.logback.access.spi.AccessEvent.buildRequestParameterMap(AccessEvent.java:338)
        at ch.qos.logback.access.spi.AccessEvent.getRequestParameterMap(AccessEvent.java:351)
        at ch.qos.logback.access.spi.AccessEvent.prepareForDeferredProcessing(AccessEvent.java:584)
        #
        # code from https://jira.qos.ch/browse/LOGBACK-1486 here
        #
        at ch.qos.logback.core.AsyncAppenderBase.append(AsyncAppenderBase.java:160)
        at ch.qos.logback.core.UnsynchronizedAppenderBase.doAppend(UnsynchronizedAppenderBase.java:84)
        at ch.qos.logback.core.spi.AppenderAttachableImpl.appendLoopOnAppenders(AppenderAttachableImpl.java:51)
        at ch.qos.logback.access.jetty.RequestLogImpl.log(RequestLogImpl.java:137)
        at org.eclipse.jetty.server.HttpChannel.onCompleted(HttpChannel.java:812)
        at org.eclipse.jetty.server.HttpChannel.onBadMessage(HttpChannel.java:872)
        at org.eclipse.jetty.server.HttpChannelOverHttp.badMessage(HttpChannelOverHttp.java:283)
        at org.eclipse.jetty.http.HttpParser.badMessage(HttpParser.java:1647)
        at org.eclipse.jetty.http.HttpParser.parseNext(HttpParser.java:1629)
        at org.eclipse.jetty.server.HttpConnection.parseRequestBuffer(HttpConnection.java:376)
        at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:265)
        at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:311)
        at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:105)
        at org.eclipse.jetty.io.ChannelEndPoint$1.run(ChannelEndPoint.java:104)
        at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.runTask(EatWhatYouKill.java:336)
        at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.doProduce(EatWhatYouKill.java:313)
        at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.tryProduce(EatWhatYouKill.java:171)
        at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.run(EatWhatYouKill.java:129)
        at org.eclipse.jetty.util.thread.ReservedThreadExecutor$ReservedThread.run(ReservedThreadExecutor.java:383)
        at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:882)
        at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1036)
        at java.lang.Thread.run(Thread.java:748)
```

Clearly we are at the end of request trying to log it (with `logback-access`) and while preparing request for deferred processing it tries to read some of the parameters from it and it all ends up in `HttpInput.read()`.

Which in turn blocks for content. However judging from how much CPU is used, our suspicion is that no actual sleeping happens, the thread just wakes up immediately, re-checks conditions, goes back to waiting and the story repeats.

I do not know anything about Jetty internals but given this code https://github.com/eclipse/jetty.project/blob/jetty-9.4.41.v20210516/jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java#L1161-L1166 just calls `input.blockForContent()` and always returns `true` (unless throws), then this loop https://github.com/eclipse/jetty.project/blob/jetty-9.4.41.v20210516/jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java#L314-L340 reduces to: 

```java
            while (true)
            {
                Content item = nextContent();
                if (item != null)
                {
                    l = get(item, b, off, len);
                    if (LOG.isDebugEnabled())
                        LOG.debug("{} read {} from {}", this, l, item);

                    // Consume any following poison pills
                    if (item.isEmpty())
                        nextInterceptedContent();
                    break;
                }

                // No content, so should we block?
                input.blockForContent();
            }
```

So in theory as long as `nextContent()` returns `null`, it just spins `input.blockForContent()`.

We tried verifying it by enabling debug logging with
```
<logger name="org.eclipse.jetty.server.HttpInput" level="DEBUG"/>
```
and it resulted in about 10Mb/sec of logs:
```
-----------------------------------------------------------------------------------------------------------------------------------------------
|       @timestamp        | level |       thread       |                                       message                                        |
|-------------------------|-------|--------------------|--------------------------------------------------------------------------------------|
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-71877 | HttpInputOverHTTP@1202b981[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
| 2021-10-08 14:54:35.000 | DEBUG | qtp708090483-69573 | HttpInputOverHTTP@72895282[c=0,q=0,[0]=null,s=STREAM] blocking for content timeout=0 |
-----------------------------------------------------------------------------------------------------------------------------------------------
```
So it does indeed look like `read()` just spins `blockForContent()` endlessly which for some reason wakes up every time but there are no data available (or `read()` would log something on this line https://github.com/eclipse/jetty.project/blob/jetty-9.4.41.v20210516/jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java#L321 )
 
Remember, there is no any load on that server at that moment, no established HTTP connections, just couple of some old requests stuck in this state.

**How to reproduce?**
Unfortunately I have no idea what triggers it. We just saw several servers ended up in this state. Last time we saw it before today was several months ago. Given all affected servers are on AWS `eu-west-1` and behind load balancers, it could be that some region-wide issue with AWS load balancers resulted in some incorrect TCP connection closure or something like that which in turn triggered some edge condition in Jetty.

It also can be the same issue as https://github.com/eclipse/jetty.project/issues/5435 although I am not 100% sure

joakime:
Adds `Request.onRequestLog()` to reset internals of Request/Response to a place where usage of those objects in `RequestLog.log(Request,Response)` don't cause problems or bad information.

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpFields.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpFields.java
@@ -1282,6 +1282,14 @@ public interface HttpFields extends Iterable<HttpField>
             _fields[_size] = null;
         }
 
+        public void setAll(HttpFields httpFields)
+        {
+            for (HttpField field : httpFields)
+            {
+                put(field);
+            }
+        }
+
         public int size()
         {
             return _size;
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpFields.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpFields.java
@@ -1037,19 +1037,6 @@ public interface HttpFields extends Iterable<HttpField>
         }
 
         /**
-         * Set all the fields in one go
-         *
-         * @param httpFields the fields to set
-         */
-        public void putAll(HttpFields httpFields)
-        {
-            for (HttpField field : httpFields)
-            {
-                put(field);
-            }
-        }
-
-        /**
          * Sets the value of a date field.
          *
          * @param name the field name
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpFields.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpFields.java
@@ -1037,6 +1037,19 @@ public interface HttpFields extends Iterable<HttpField>
         }
 
         /**
+         * Set all the fields in one go
+         *
+         * @param httpFields the fields to set
+         */
+        public void putAll(HttpFields httpFields)
+        {
+            for (HttpField field : httpFields)
+            {
+                put(field);
+            }
+        }
+
+        /**
          * Sets the value of a date field.
          *
          * @param name the field name
@@ -1280,14 +1293,6 @@ public interface HttpFields extends Iterable<HttpField>
             _size--;
             System.arraycopy(_fields, i + 1, _fields, i, _size - i);
             _fields[_size] = null;
-        }
-
-        public void setAll(HttpFields httpFields)
-        {
-            for (HttpField field : httpFields)
-            {
-                put(field);
-            }
         }
 
         public int size()
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
@@ -1413,6 +1413,19 @@ public class Request implements HttpServletRequest
         return session.getId();
     }
 
+    protected void onRequestLog()
+    {
+        // Don't allow pulling more parameters
+        _contentParamsExtracted = true;
+
+        // Reset the status code to what was committed
+        MetaData.Response committedResponse = getResponse().getCommittedMetaData();
+        if (committedResponse != null)
+        {
+            getResponse().setStatus(committedResponse.getStatus());
+        }
+    }
+
     /**
      * Called when the request is fully finished being handled.
      * For every session in any context that the session has
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
@@ -1418,18 +1418,8 @@ public class Request implements HttpServletRequest
         // Don't allow pulling more parameters
         _contentParamsExtracted = true;
 
-        // Reset the response to what it was when committed
-        MetaData.Response committedResponse = getResponse().getCommittedMetaData();
-        if (committedResponse != null)
-        {
-            // restore status
-            getResponse().setStatus(committedResponse.getStatus());
-
-            // restore headers
-            HttpFields.Mutable responseFields = getResponse().getHttpFields();
-            responseFields.clear();
-            responseFields.setAll(committedResponse.getFields());
-        }
+        // Restore the response to what it was when committed
+        getResponse().restoreCommittedMetaData();
     }
 
     /**
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
@@ -1418,11 +1418,17 @@ public class Request implements HttpServletRequest
         // Don't allow pulling more parameters
         _contentParamsExtracted = true;
 
-        // Reset the status code to what was committed
+        // Reset the response to what it was when committed
         MetaData.Response committedResponse = getResponse().getCommittedMetaData();
         if (committedResponse != null)
         {
+            // restore status
             getResponse().setStatus(committedResponse.getStatus());
+
+            // restore headers
+            HttpFields.Mutable responseFields = getResponse().getHttpFields();
+            responseFields.clear();
+            responseFields.setAll(committedResponse.getFields());
         }
     }
 
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
@@ -651,19 +651,22 @@ public class Response implements HttpServletResponse
     @Override
     public void setHeader(String name, String value)
     {
-        if (isCommitted() || !isMutable())
-            return;
+        boolean allowSetHeader = !isCommitted();
 
         long biInt = _errorSentAndIncludes.get();
         if (biInt != 0)
         {
             boolean errorSent = AtomicBiInteger.getHi(biInt) != 0;
             boolean including = AtomicBiInteger.getLo(biInt) > 0;
+            allowSetHeader &= including;
             if (!errorSent && including && name.startsWith(SET_INCLUDE_HEADER_PREFIX))
                 name = name.substring(SET_INCLUDE_HEADER_PREFIX.length());
             else
                 return;
         }
+
+        if (!allowSetHeader)
+            return;
 
         if (HttpHeader.CONTENT_TYPE.is(name))
             setContentType(value);
@@ -704,19 +707,22 @@ public class Response implements HttpServletResponse
     @Override
     public void addHeader(String name, String value)
     {
-        if (isCommitted() || !isMutable())
-            return;
+        boolean allowSetHeader = !isCommitted();
 
         long biInt = _errorSentAndIncludes.get();
         if (biInt != 0)
         {
             boolean errorSent = AtomicBiInteger.getHi(biInt) != 0;
             boolean including = AtomicBiInteger.getLo(biInt) > 0;
+            allowSetHeader &= including;
             if (!errorSent && including && name.startsWith(SET_INCLUDE_HEADER_PREFIX))
                 name = name.substring(SET_INCLUDE_HEADER_PREFIX.length());
             else
                 return;
         }
+
+        if (!allowSetHeader)
+            return;
 
         if (HttpHeader.CONTENT_TYPE.is(name))
         {
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
@@ -1305,6 +1305,29 @@ public class Response implements HttpServletResponse
         return meta;
     }
 
+    /**
+     * Restore the CommittedMetaData, overwriting the current
+     * metadata in the process.
+     * <p>
+     *     This is useful when you want to see the state of
+     *     the Response object as it was when it was committed.
+     * </p>
+     */
+    public void restoreCommittedMetaData()
+    {
+        MetaData.Response committedResponse = getCommittedMetaData();
+        if (committedResponse != null)
+        {
+            // restore status
+            setStatus(committedResponse.getStatus());
+
+            // restore headers
+            HttpFields.Mutable responseFields = getHttpFields();
+            responseFields.clear();
+            responseFields.putAll(committedResponse.getFields());
+        }
+    }
+
     @Override
     public boolean isCommitted()
     {
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
@@ -216,6 +216,9 @@ public class Response implements HttpServletResponse
 
     public void addCookie(HttpCookie cookie)
     {
+        if (isCommitted() || !isMutable())
+            return;
+
         if (StringUtil.isBlank(cookie.getName()))
             throw new IllegalArgumentException("Cookie.name cannot be blank/null");
      
@@ -258,30 +261,30 @@ public class Response implements HttpServletResponse
     @Override
     public void addCookie(Cookie cookie)
     {
-        //Servlet Spec 9.3 Include method: cannot set a cookie if handling an include
-        if (isMutable())
-        {
-            if (StringUtil.isBlank(cookie.getName()))
-                throw new IllegalArgumentException("Cookie.name cannot be blank/null");
+        // Servlet Spec 9.3 Include method: cannot set a cookie if handling an include
+        if (isCommitted() || !isMutable())
+            return;
 
-            String comment = cookie.getComment();
-            // HttpOnly was supported as a comment in cookie flags before the java.net.HttpCookie implementation so need to check that
-            boolean httpOnly = cookie.isHttpOnly() || HttpCookie.isHttpOnlyInComment(comment);
-            SameSite sameSite = HttpCookie.getSameSiteFromComment(comment);
-            comment = HttpCookie.getCommentWithoutAttributes(comment);
+        if (StringUtil.isBlank(cookie.getName()))
+            throw new IllegalArgumentException("Cookie.name cannot be blank/null");
 
-            addCookie(new HttpCookie(
-                cookie.getName(),
-                cookie.getValue(),
-                cookie.getDomain(),
-                cookie.getPath(),
-                cookie.getMaxAge(),
-                httpOnly,
-                cookie.getSecure(),
-                comment,
-                cookie.getVersion(),
-                sameSite));
-        }
+        String comment = cookie.getComment();
+        // HttpOnly was supported as a comment in cookie flags before the java.net.HttpCookie implementation so need to check that
+        boolean httpOnly = cookie.isHttpOnly() || HttpCookie.isHttpOnlyInComment(comment);
+        SameSite sameSite = HttpCookie.getSameSiteFromComment(comment);
+        comment = HttpCookie.getCommentWithoutAttributes(comment);
+
+        addCookie(new HttpCookie(
+            cookie.getName(),
+            cookie.getValue(),
+            cookie.getDomain(),
+            cookie.getPath(),
+            cookie.getMaxAge(),
+            httpOnly,
+            cookie.getSecure(),
+            comment,
+            cookie.getVersion(),
+            sameSite));
     }
 
     /**
@@ -292,6 +295,9 @@ public class Response implements HttpServletResponse
      */
     public void replaceCookie(HttpCookie cookie)
     {
+        if (isCommitted() || !isMutable())
+            return;
+
         for (ListIterator<HttpField> i = _fields.listIterator(); i.hasNext(); )
         {
             HttpField field = i.next();
@@ -602,40 +608,42 @@ public class Response implements HttpServletResponse
     @Override
     public void setDateHeader(String name, long date)
     {
-        if (isMutable())
-        {
-            HttpHeader header = HttpHeader.CACHE.get(name);
-            if (header == null)
-                _fields.putDateField(name, date);
-            else
-                _fields.putDateField(header, date);
-        }
+        if (isCommitted() || !isMutable())
+            return;
+
+        HttpHeader header = HttpHeader.CACHE.get(name);
+        if (header == null)
+            _fields.putDateField(name, date);
+        else
+            _fields.putDateField(header, date);
     }
 
     @Override
     public void addDateHeader(String name, long date)
     {
-        if (isMutable())
-            _fields.addDateField(name, date);
+        if (isCommitted() || !isMutable())
+            return;
+
+        _fields.addDateField(name, date);
     }
 
     public void setHeader(HttpHeader name, String value)
     {
-        if (isMutable())
-        {
-            if (HttpHeader.CONTENT_TYPE == name)
-                setContentType(value);
-            else
-            {
-                _fields.put(name, value);
+        if (isCommitted() || !isMutable())
+            return;
 
-                if (HttpHeader.CONTENT_LENGTH == name)
-                {
-                    if (value == null)
-                        _contentLength = -1L;
-                    else
-                        _contentLength = Long.parseLong(value);
-                }
+        if (HttpHeader.CONTENT_TYPE == name)
+            setContentType(value);
+        else
+        {
+            _fields.put(name, value);
+
+            if (HttpHeader.CONTENT_LENGTH == name)
+            {
+                if (value == null)
+                    _contentLength = -1L;
+                else
+                    _contentLength = Long.parseLong(value);
             }
         }
     }
@@ -643,6 +651,9 @@ public class Response implements HttpServletResponse
     @Override
     public void setHeader(String name, String value)
     {
+        if (isCommitted() || !isMutable())
+            return;
+
         long biInt = _errorSentAndIncludes.get();
         if (biInt != 0)
         {
@@ -693,6 +704,9 @@ public class Response implements HttpServletResponse
     @Override
     public void addHeader(String name, String value)
     {
+        if (isCommitted() || !isMutable())
+            return;
+
         long biInt = _errorSentAndIncludes.get();
         if (biInt != 0)
         {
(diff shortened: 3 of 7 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -1413,6 +1413,19 @@ public String changeSessionId()
...
+    {
+        // Don't allow pulling more parameters
+        _contentParamsExtracted = true;
+
+        // Reset the status code to what was committed
+        MetaData.Response committedResponse = getResponse().getCommittedMetaData();
+        if (committedResponse != null)
+        {
+            getResponse().setStatus(committedResponse.getStatus());
+        }
```
gregw: Should we also reset response headers? or are they already the same due to pass by reference.

joakime: Good question.
I'll expand the tests to see what happens.

joakime:
@gregw you were right, the response headers needed reset as well.

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -1413,6 +1413,25 @@ public String changeSessionId()
...
+        {
+            // restore status
+            getResponse().setStatus(committedResponse.getStatus());
+
+            // restore headers
+            HttpFields.Mutable responseFields = getResponse().getHttpFields();
+            responseFields.clear();
+            responseFields.setAll(committedResponse.getFields());
+        }
+    }
```
joakime, line 1423: I almost think a `Response.restoreCommittedState()` would make sense. WDYT?

jetty-http/src/main/java/org/eclipse/jetty/http/HttpFields.java
```diff
@@ -1282,6 +1282,14 @@ private void remove(int i)
             _fields[_size] = null;
         }
 
+        public void setAll(HttpFields httpFields)
```
lachlan-roberts: There are no other set methods here, maybe this should be named `putAll` instead. Or just another overloaded `put(HttpFields)` method.

joakime: Sure, I can do that.
I used to have the `.clear()` in here, which made this name make sense.
`putAll` is better aligned with the rest of the methods here.

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
joakime: The headers are being restored now as well.

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
joakime: Implemented.

jetty-http/src/main/java/org/eclipse/jetty/http/HttpFields.java
joakime: Done, and I moved it to the appropriate place (sort wise).

lachlan-roberts (dismissed):
LGTM

jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -1305,6 +1305,29 @@ public void setTrailerFields(Supplier<Map<String, String>> trailers)
...
+        MetaData.Response committedResponse = getCommittedMetaData();
+        if (committedResponse != null)
+        {
+            // restore status
+            setStatus(committedResponse.getStatus());
+
+            // restore headers
+            HttpFields.Mutable responseFields = getHttpFields();
+            responseFields.clear();
+            responseFields.putAll(committedResponse.getFields());
```
gregw: I'm now worried about the performance impact of this.    On every response we are clearing any copying all the response headers just in case of the very VERY few cases that they are modified after commit.

I'm now not inclined to do this, at least not by default.

@lachlan-roberts  is there anyway the requestLogs could directly access the committed fields without the need to copy them back?

gregw (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
joakime: This is only done if a RequestLog is defined.

joakime: I felt the impact of having a switch on the various .getHeader() methods would be worse (performance wise).
As it would impact 100% of users, even those that don't use RequestLog.
WDYT?

joakime: As for "very few cases", like 100% of all webapps that use spring-security or use jersey?

joakime: Would it be possible to make the various .setHeader() implementations not actually set the header if the response is already committed?

joakime: There's potentially more waste (performance wise) of allowing the header to be set (even a new header) after response commit.

lachlan-roberts: @gregw the request log uses the jetty `Request` and `Response` classes so it does have access to the committed metadata.

joakime: @lachlan-roberts but RequestLog implementations (like logback-access, spring-boot, jersey) have generic Request logging based on the HttpServletRequest and HttpServletResponse.

joakime: Is a check like ... `if (committedMetadata != metadata)` possible without worse performance?
ooh, how about a setting a "dirty" flag on the Response if any response header is set after commit?
We could use that to determine if we need to copy the headers back (or not).
That should cover the "very few cases" aspect, right?   (I think there might be a race condition tho)

gregw: A `committedMetadata.isEqualTo(metadata)` would fail fast, but be a full sweep for the most common case of no changes.
A dirty bit might be better, but still it would either have to be in HttpFields (and cost every other usage), or in Response, adding more complexity in more places.

So I'm probably more inclined to have an option in HttpChannel or HttpConfiguration for "restoreResponseFields", that defaults to off.

Or alternately, once committed, we could make the response headers immutable, so any calls would become noops?

joakime:
@gregw @lachlan-roberts I changed the implementation to rely on isCommitted in Response to not change headers after commit.
The previous commit that copied HttpFields from committed metadata to normal metadata has been reverted.

jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -651,20 +651,23 @@ public void setHeader(HttpHeader name, String value)
     @Override
     public void setHeader(String name, String value)
     {
-        if (isCommitted() || !isMutable())
-            return;
+        boolean allowSetHeader = !isCommitted();
```
gregw, line 654: Hmm it is a bit strange in the previous code that we were both checking isMutable and doing a direct check of the biInt below???
I guess that was for the `SET_INCLUDE_HEADER_PREFIX`.

But we have lots of other methods that just mutate the fields directly after checking only isMutable.
I'd prefer a solution where we continue to call isMutable and just become immutable after commit!

gregw (dismissed):
Also, if we are to prevent changes to the headers after commit, perhaps we should also just prevent changes to the status after commit also - ie fix the problem the same way for both.

Edit: setStatus already checks `isMutable()`, so we just need to become immutable after commit!

gregw:
Hmmm other complication here is that the committedMetaData might be different to the the fields at time of commit, as the container might send a 400 or 500 directly....
Need some more head scratching.....

gregw:
Maybe we just need to be clever in the getters:
```java
    @Override
    public String getHeader(String name)
    {
        return (isCommitted()?getCommittedMetaData().getFields():_fields).get(name);
    }
```

joakime:
@gregw I've made the change for the 3 getHeader* methods.

joakime:
@gregw this needs a fresh review

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -1413,6 +1413,15 @@ public String changeSessionId()
         return session.getId();
     }
 
+    protected void onRequestLog()
+    {
+        // Don't allow pulling more parameters
+        _contentParamsExtracted = true;
```
gregw, line 1419: Can we split this PR into two.
This is the fix that we need to fix the actual hard bug.  The not mutating the response is a nice to have feature.  I'm currently working on some ideas for that in jetty-12, so once those work out, it would be good to step in the same direction.

gregw (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
joakime: To me that means ...

1. The restoring of committed metadata
2. The isMutable checks in Response

That what you mean?

gregw: My understanding of the actual bug we are trying to fix is to stop implementations of requestLog from blocking trying to read parameters.   So the `_contentParamsExtracted = true` change here is the actual fix.
Making sure that the response headers and status are immutable after commit is a second order issue.

joakime: At a minimum the headers and status code need to be copied into place. (what `restoreCommittedMetadata` does)

joakime: Basically, restore the committed state is step 1.

Step 2 is to prevent the altering of state after commit.

gregw: No.   Let's say that I don't really care about if we log incorrect data if the headers/status are changed after commit ( I do a bit, but let's pretend).     Even if I don't care about that, I still care a lot if the 3rd party request logger tries to get a parameter and ends up blocking in extractParameters.    So let's fix that hard bug and then deal with the immutability of committed headers for request logs in a second PR.

joakime: Opened PR #7183 for the first part.

joakime: Not going to open second PR, as I don't like PRs depending on other PRs (the merge gets messy, and it's impossible to not rebase force-push in those scenarios).
Lets get #7183 merged, then we can base part 2 on that from `jetty-10.0.x` (post-merge)

joakime:
Closing this old PR as PR #7183 with a lite version of this change has been merged.

PR #7000, opened 2021-10-15
Issue #1087 - add module for a .well-known handler

### LINKED ISSUE #1087: Support .well-known serving in handlers
The `.well-known` directory is being used for more and more things to automate operations related to websites. 

For example, see #794.
Also, Let's Encrypt uses the `.well-known` directory to automate the issue of certificates.

For "normal" contexts, this is rather easy, since it's enough to add the `.well-known` directory to the `resourceBase` and it's done.

For less "normal" contexts, this is not doable.
For example, a `MovedContextHandler` would need to be able to serve content under `.well-known` and if that fails, fall back to the "moved" behavior.
Same goes for contexts that use a `RedirectHandler`.

The major problem is trying to serve the "/" URI.

Currently, a handler that has a `resourceBase` will try to serve `/` as a directory, or give a 403 if directory listing is not allowed.

What would be needed is a handler with some logic along these lines:
- If URI matches `/.well-known` and descendants => serve it like a `ResourceHandler` would from file system
- If URI matches `/` => forward to next/wrapped handler

So that is basically a `ResourceHandler`, but only for a particular directory (or set of paths).

If there are ways to achieve this, will be great to know.

lachlan-roberts:
## Closes #1087

Adds a `well-known.mod` module which can add a handler to serve files on the `/.well-known` context path.
By default it serves files from`${jetty.base}/.well-known` directory.

### DIFF
--- a/documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-well-known.adoc
+++ b/documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-well-known.adoc
@@ -14,14 +14,16 @@
 [[og-module-well-known]]
 ===== Module `well-known`
 
-The `well-known` creates a `ResourceHandler` deployed at the `/.well-known` context path.
-This `Handler` serves files from a directory created at `${jetty.base}/.well-known`, note that this may be seen as a hidden directory by the filesystem.
+The `well-known` creates a `ResourceHandler` deployed at the `/.well-known` context path which serves files from a directory.
+By default, the directory created at `$JETTY_BASE/.well-known` is used, but it can be configured from `well-known.ini` to anywhere in the filesystem.
+Note that the `.well-known` directory may be seen as a hidden directory by the filesystem.
 
-The concept of well-known URIs has been defined in RFC5785.
-It may be used for things like automation of the renewal for Let's Encrypt certificates.
+The concept of well-known URIs has been defined in link:https://datatracker.ietf.org/doc/html/rfc5785[RFC5785].
+This module can be used for things like the automatic renewal of link:https://letsencrypt.org/[Let's Encrypt] certificates.
+See link:https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml[IANA Well-Known URIs] for more possible examples of how this can be used.
 
-The module file is `$JETTY_HOME/modules/well-known.mod`:
+The module properties are:
 
 ----
-include::{JETTY_HOME}/modules/well-known.mod[]
+include::{JETTY_HOME}/modules/well-known.mod[tags=documentation]
 ----
--- a/documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-well-known.adoc
+++ b/documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-well-known.adoc
@@ -14,7 +14,7 @@
 [[og-module-well-known]]
 ===== Module `well-known`
 
-The `well-known` creates a `ResourceHandler` deployed at the `/.well-known` context path which serves files from a directory.
+The `well-known` Jetty module creates a `ResourceHandler` deployed at the `/.well-known` context path which serves files from a directory.
 By default, the directory created at `$JETTY_BASE/.well-known` is used, but it can be configured from `well-known.ini` to anywhere in the filesystem.
 Note that the `.well-known` directory may be seen as a hidden directory by the filesystem.
 
--- a/documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-well-known.adoc
+++ b/documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-well-known.adoc
@@ -11,22 +11,17 @@
 // ========================================================================
 //
 
-[[og-modules-standard]]
-==== Standard Modules
+[[og-module-well-known]]
+===== Module `well-known`
 
-include::module-alpn.adoc[]
-include::module-bytebufferpool.adoc[]
-include::module-deploy.adoc[]
-include::module-http.adoc[]
-include::module-http2.adoc[]
-include::module-http2c.adoc[]
-include::module-http-forwarded.adoc[]
-include::module-https.adoc[]
-include::module-jmx-remote.adoc[]
-include::module-server.adoc[]
-include::module-ssl.adoc[]
-include::module-ssl-reload.adoc[]
-include::module-test-keystore.adoc[]
-include::module-threadpool.adoc[]
-include::module-console-capture.adoc[]
-include::module-well-known.adoc[]
+The `well-known` creates a `ResourceHandler` deployed at the `/.well-known` context path.
+This `Handler` serves files from a directory created at `${jetty.base}/.well-known`, note that this may be seen as a hidden directory by the filesystem.
+
+The concept of well-known URIs has been defined in RFC5785.
+It may be used for things like automation of the renewal for Let's Encrypt certificates.
+
+The module file is `$JETTY_HOME/modules/well-known.mod`:
+
+----
+include::{JETTY_HOME}/modules/well-known.mod[]
+----
--- a/jetty-server/src/main/config/etc/well-known.xml
+++ b/jetty-server/src/main/config/etc/well-known.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0"?>
+<!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "https://www.eclipse.org/jetty/configure_10_0.dtd">
+
+<Configure id="Contexts" class="org.eclipse.jetty.server.handler.ContextHandlerCollection">
+  <New id="WellKnownHandler" class="org.eclipse.jetty.server.handler.ContextHandler">
+    <Set name="contextPath">/.well-known</Set>
+    <Set name="handler">
+      <New class="org.eclipse.jetty.server.handler.ResourceHandler">
+        <Set name="resourceBase">
+          <Call name="resolvePath" class="org.eclipse.jetty.xml.XmlConfiguration">
+            <Arg><Property name="jetty.base"/></Arg>
+            <Arg><Property name="jetty.wellknown.dir" default=".well-known"/></Arg>
+          </Call>
+        </Set>
+        <Set name="directoriesListed">false</Set>
+      </New>
+    </Set>
+  </New>
+
+  <Call name="addHandler">
+    <Arg><Ref refid="WellKnownHandler"/></Arg>
+  </Call>
+</Configure>
+
--- a/jetty-server/src/main/config/modules/well-known.mod
+++ b/jetty-server/src/main/config/modules/well-known.mod
@@ -1,0 +1,23 @@
+DO NOT EDIT - See: https://www.eclipse.org/jetty/documentation/current/startup-modules.html
+
+[description]
+Serve static files from a directory for the "/.well-known" context path.
+
+[tags]
+handler
+
+[depend]
+server
+
+[xml]
+etc/well-known.xml
+
+[files]
+.well-known/
+
+[ini]
+jetty.wellknown.dir?=.well-known
+
+[ini-template]
+## Well Known Directory (relative to $jetty.base)
+# jetty.wellknown.dir=.well-known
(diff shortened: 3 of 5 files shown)


joakime (dismissed):

olamy (dismissed):

jetty-server/src/main/config/etc/well-known.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "https://www.eclipse.org/jetty/configure_10_0.dtd">
+
+<Configure id="Contexts" class="org.eclipse.jetty.server.handler.ContextHandlerCollection">
+  <New id="WellKnownHandler" class="org.eclipse.jetty.server.handler.ContextHandler">
```
sbordet, line 5: Seems to me that most `id` attribute names start with lowercase letter.

jetty-server/src/main/config/modules/well-known.mod
```diff
@@ -0,0 +1,23 @@
...
+server
+
+[xml]
+etc/well-known.xml
+
+[files]
+.well-known/
+
+[ini]
+jetty.wellknown.dir?=.well-known
```
sbordet: Why this section? I don't think it's needed since the XML has a `default` attribute, no?

sbordet (changes requested):
Please also write a distribution test case, we want to be sure that it works (e.g. no typos in XML, etc.).

sbordet:
@lachlan-roberts also, I would not mind documentation about this.
Please add an entry in `modules-standard.adoc` and a correspondent file `module-well-known.adoc`.

jetty-server/src/main/config/etc/well-known.xml
joakime: I was curious, as I thought it was uppercase mostly.
So ...

``` shell
$ cd jetty-home-10.0.7/etc
# lowercase as "id" declaration
$ grep -ERh ' id="[a-z]' * | wc -l
71
# lowercase as "refid" declaration
$ grep -ERh ' refid="[a-z]' * | wc -l
43
# uppercase as "id" declaration
$ grep -ERh ' id="[A-Z]' * | wc -l
74
# uppercase as "refid" declaration
$ grep -ERh ' refid="[A-Z]' * | wc -l
26
```

So uppercase wins by 3 on "id", but is less used in "refid".
It's pretty much an even split in our XML.

This is the breakdown, of both "id" and "refid" usage.

``` shell
$ cd jetty-home-10.0.7/etc
$ grep -Rh "id=" * | sed -e 's/^.*\(id="[^"]*"\).*/\1/' | sort | uniq -c | sort -rn
     64 id="Server"
     11 id="sessionDataStoreFactory"
      8 id="httpConfig"
      6 id="properties"
      6 id="config"
      6 id="cache"
      5 id="unixSocketHttpConfig"
      5 id="sslContextFactory"
      5 id="sslConnector"
      5 id="queryMgrFactory"
      4 id="unixSocketConnector"
      4 id="DeploymentManager"
      3 id="sslHttpConfig"
      3 id="ServerLog"
      3 id="Rewrite"
      3 id="httpConnector"
      3 id="DebugListener"
      3 id="databaseAdaptor"
      3 id="cacheMgr"
      2 id="webappprovider"
      2 id="ThreadPool"
      2 id="threadPool"
      2 id="sessionDataMapFactory"
      2 id="ServerLogName"
      2 id="remoteCacheManager"
      2 id="queryManager"
      2 id="pkixSoftFail"
      2 id="pkixRevocationChecker"
      2 id="pkixPreferCrls"
      2 id="pkixNoFallback"
      2 id="OpenIdConfiguration"
      2 id="MBeanServer"
      2 id="mapping"
      2 id="jaspiAuthConfigFactory"
      2 id="HttpClient"
      2 id="hotrodMgr"
      2 id="defaultConfig"
      2 id="decoratorAttribute"
      2 id="Contexts"
      2 id="context"
      2 id="byteBufferPool"
      1 id="unixDomainConnector"
      1 id="ThreadLimitHandler"
      1 id="Stop"
      1 id="StatsHandler"
      1 id="RequestLog"
      1 id="proxyHolder"
      1 id="MBeanContainer"
      1 id="lowResourceMonitor"
      1 id="logging"
      1 id="keyStoreScanner"
      1 id="jsp"
      1 id="JMXConnectorServer"
      1 id="idMgr"
      1 id="homeBaseWarning"
      1 id="Handlers"
      1 id="GzipHandler"
      1 id="DefaultHandler"
      1 id="DebugHandler"
      1 id="BaseLoginService"
      1 id="alpn"
      1         <Arg name="server"><Ref refid='Server'/></Arg>
```

lachlan-roberts: The convention for Handlers seems to be capital we have `GzipHandler`, `DefaultHandler`, `DebugHandler`, `StatsHandler`. So I will leave it as `WellKnownHandler`.

jetty-server/src/main/config/modules/well-known.mod
lachlan-roberts: Yeah I don't think it is necessary I will remove.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-well-known.adoc
```diff
@@ -0,0 +1,27 @@
...
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+[[og-module-well-known]]
+===== Module `well-known`
+
+The `well-known` creates a `ResourceHandler` deployed at the `/.well-known` context path.
+This `Handler` serves files from a directory created at `${jetty.base}/.well-known`, note that this may be seen as a hidden directory by the filesystem.
```
sbordet: Please specify that the default is `$JETTY_BASE/.well-known` (note use of env variable style, rather than ${jetty.base}, but it can be configured to be anywhere in the file system.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-well-known.adoc
```diff
@@ -0,0 +1,27 @@
...
+
+The concept of well-known URIs has been defined in RFC5785.
```
sbordet: Link the RFC using `link:`

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-well-known.adoc
```diff
@@ -0,0 +1,27 @@
...
+It may be used for things like automation of the renewal for Let's Encrypt certificates.
```
sbordet: "It" at beginning of sentence seems to refer to the RFC. Specify again "This module".
Link `Let's Encrypt`.
Does the RFC specify another example? I think there is one for HTTP/2.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-well-known.adoc
```diff
@@ -0,0 +1,27 @@
...
+
+The module file is `$JETTY_HOME/modules/well-known.mod`:
```
sbordet: Don't link the whole module, just its properties, which should be commented as if it was documentation, see the `alpn` module.

sbordet (changes requested):

joakime:
Out of curiosity, why did we choose `${jetty.base}/.well-known` instead of ...

* `${jetty.base}/webapps/.well-known` or
* `${jetty.base}/webapps/well-known`

This entire concept could also have been done with an injected `${jetty.base}/webapps/well-known.xml` (like we do for the demo modules)

sbordet:
@joakime injecting `${jetty.base}/webapps/well-known.xml` would require to customize the XML with the actual directory, right, as I don't think it can be referenced from the XML as a module variable?

As for the location inside webapps or not, usages of `.well-known` are probably mostly done outside, as there is a collaboration with some other process that writes in the `.well-known` directory.

joakime:
Wish we had modules that can be added/installed without creating a `start.d/<name>.ini`
We could use them directly with `demo` now, and in the future with this `well-known` module.

It just has to create the `${jetty.base}/webapps/well-known.xml` and if anyone wants to customize it, they just customize the XML directly.

I wish we had separated webapps (or webapp-templates?) out from the module system properly.
Having the module system do webapps is awkward and getting more awkward. (`--add-webapp=<name>`?)
Eg: you cannot upgrade the webapp once installed, going from Jetty 10.0.x `demo` module and then upgrading to Jetty 11.0.x is broken, as webapps are not part of the module system logic.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-well-known.adoc
```diff
@@ -0,0 +1,29 @@
...
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+[[og-module-well-known]]
+===== Module `well-known`
+
+The `well-known` creates a `ResourceHandler` deployed at the `/.well-known` context path which serves files from a directory.
+By default, the directory created at `$JETTY_BASE/.well-known` is used, but it can be configured from `well-known.ini` to anywhere in the filesystem.
+Note that the `.well-known` directory may be seen as a hidden directory by the filesystem.
```
gregw, line 19: Wouldn't it be better to make the default `$JETTY_BASE/well-known` so that it is not hidden?

jetty-server/src/main/config/modules/well-known.mod
```diff
@@ -0,0 +1,22 @@
+DO NOT EDIT - See: https://www.eclipse.org/jetty/documentation/current/startup-modules.html
+
+[description]
+Serve static files from a directory for the "/.well-known" context path.
```
gregw, line 4:
```suggestion
Serve static files from a directory for the "/.well-known" context path from the "$JETTY_HOME/.well-known" directory.
```

gregw (changes requested):

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-well-known.adoc
lachlan-roberts: @gregw the problem with that is that the most common use case for this will probably be lets encrypt, and when you use that you need to specify the "webroot" directory and then it creates its files in the `$WEBROOT/.well-known` directory.

So if you made this by default `$JETTY_BASE/well-known` then it wouldn't work without changing the configuration.

jetty-server/src/main/config/modules/well-known.mod
gregw: @lachlan-roberts ?

lachlan-roberts: The directory can be configured below so its not always served from the `$JETTY_HOME/.well-known` directory.

gregw (dismissed):

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-well-known.adoc
```diff
@@ -0,0 +1,29 @@
...
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+[[og-module-well-known]]
+===== Module `well-known`
+
+The `well-known` creates a `ResourceHandler` deployed at the `/.well-known` context path which serves files from a directory.
```
sbordet:
```suggestion
The `well-known` Jetty module creates a `ResourceHandler` deployed at the `/.well-known` context path which serves files from a directory.
```

jetty-server/src/main/config/modules/well-known.mod
```diff
@@ -0,0 +1,22 @@
...
+
+[xml]
+etc/well-known.xml
+
+[files]
+.well-known/
+
+[ini-template]
+# tag::documentation[]
+## Well Known Directory (relative to $jetty.base)
```
sbordet: Please specify here that if it's a relative path then it is relative to `$JETTY_BASE`, otherwise it's an absolute path.

sbordet (changes requested):

sbordet (approved):

PR #7009, opened 2021-10-18
Issue #7008 - Fix regression in `bin/jetty.sh` on systems using `start-stop-daemon`

### LINKED ISSUE #7008: Problem with jetty.sh start regression 10.0.6 -> 10.0.7 when using JETTY_USER
**Jetty version(s)**
10.0.7

Openjdk 17

**OS type/version**
Ubuntu 21.10

**Description**
When using JETTY_USER,   "jetty.sh start"  fails to startup.

**How to reproduce?**
Setting JETTY_USER and configure as a service /etc/init.d/jetty 
Worked with 10.0.6 (changes https://github.com/eclipse/jetty.project/commit/2b0161e743ac4f536e2017148551487624c367d1#diff-2b9f3399552ca99de76591c09421a925e0d63a4c7cea19757ae7a81551993d24 ) and stll works if replace with 10.0.6 version of jetty.sh only.

joakime:
* Using `start-stop-daemon` options properly
* Do not include `JETTY_SYS_PROPS` in `JAVA_OPTIONS` to avoid duplicates that show up in `RUN_ARGS`.
* Only use `JETTY_SYS_PROPS` in `RUN_CMD`.

### DIFF
--- a/jetty-home/src/main/resources/bin/jetty.sh
+++ b/jetty-home/src/main/resources/bin/jetty.sh
@@ -414,7 +414,7 @@ TMPDIR="`cygpath -w $TMPDIR`"
 ;;
 esac
 
-JAVA_OPTIONS=(${JAVA_OPTIONS[*]} "-Djetty.home=$JETTY_HOME" "-Djetty.base=$JETTY_BASE" "-Djava.io.tmpdir=$TMPDIR")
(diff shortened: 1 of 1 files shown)


joakime:
The issue filer has verified that this works for them. https://github.com/eclipse/jetty.project/issues/7008#issuecomment-945866640

olamy (approved):

PR #7013, opened 2021-10-19
Issue #7012 Replace geronimo spec jars with jakarta equivalents

### LINKED ISSUE #7012: Remove all old geronimo spec jars from jetty-10
All of the geronimo spec jars should be replaced by the jakarta equivalents that target the `javax` namespace.

janbartel:
Closes #7012 

Replace all geronimo spec jars by their jakarta equivalents (using the `javax` namespace for jetty-10)

### DIFF
--- a/jetty-osgi/test-jetty-osgi/src/test/java/org/eclipse/jetty/osgi/test/TestOSGiUtil.java
+++ b/jetty-osgi/test-jetty-osgi/src/test/java/org/eclipse/jetty/osgi/test/TestOSGiUtil.java
@@ -104,22 +104,13 @@ public class TestOSGiUtil
         return options;
     }
 
-    public static List<Option> provisionCoreJetty()
-    {
-        List<Option> res = new ArrayList<>();
-        // get the jetty home config from the osgi boot bundle.
-        res.add(CoreOptions.systemProperty("jetty.home.bundle").value("org.eclipse.jetty.osgi.boot"));
-        res.addAll(coreJettyDependencies());
-        return res;
-    }
-
     public static Option optionalRemoteDebug()
     {
         return CoreOptions.when(Boolean.getBoolean("pax.exam.debug.remote"))
             .useOptions(CoreOptions.vmOption("-Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=5005"));
     }
 
-    public static List<Option> coreJettyDependencies()
+    public static List<Option> coreJettyDependencies(boolean withJsp)
     {
         List<Option> res = new ArrayList<>();
         //enables a dump of the status of all deployed bundles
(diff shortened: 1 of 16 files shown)


jetty-osgi/test-jetty-osgi/src/test/java/org/eclipse/jetty/osgi/test/TestOSGiUtil.java
```diff
@@ -104,22 +104,13 @@ public void stop(BundleContext context) throws Exception
...
-    }
-
     public static Option optionalRemoteDebug()
     {
         return CoreOptions.when(Boolean.getBoolean("pax.exam.debug.remote"))
             .useOptions(CoreOptions.vmOption("-Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=5005"));
     }
 
-    public static List<Option> coreJettyDependencies()
+    public static List<Option> coreJettyDependencies(boolean withJsp)
```
joakime: Can we just have a `coreJspDependencies()` instead of this boolean?

joakime (changes requested):
The older `jspDependencies()` made it easier to understand what was being done and why in the various places it was being used.

The new boolean technique is awkward.
Please either restore `jspDependencies()` or make the parameter something else.

``` java
enum JettyFeatureEnum {
  JSP, CDI
}

public static List<Option> coreJettyDependencies(JettyFeatureEnum ... feature) {
```

and then used like ...

``` java
options.addAll(TestOSGiUtil.coreJettyDependencies(JSP));
```

pom.xml
```diff
@@ -57,8 +52,10 @@
     <jackson-databind.version>2.9.10.7</jackson-databind.version>
     <jakarta.activation.api.version>1.2.2</jakarta.activation.api.version>
     <jakarta.annotation.api.version>1.3.5</jakarta.annotation.api.version>
-    <jakarta.el-api.version>3.0.3</jakarta.el-api.version>
     <jakarta.enterprise.cdi.api.version>2.0.2</jakarta.enterprise.cdi.api.version>
+    <jakarta.inject.api.version>1.0.4</jakarta.inject.api.version>
```
olamy: 1.0.5 looks to be available now see #7016

olamy (changes requested):

pom.xml
janbartel: No fair, I'm in a race with dependabot ;)

jetty-osgi/test-jetty-osgi/src/test/java/org/eclipse/jetty/osgi/test/TestOSGiUtil.java
janbartel: OK, I've implemented it a different way. The problem is that the jakarta.transaction-api jar wants to use the javax.el* package, so we must deploy the jakarta.el-api jar to satisfy it, even though it won't be used. However, if we then subsequently configure the JSP bundles, we want to use the org.mortbay.jasper.apache-el jar instead because it provides both the api and impl. So in the latest solution I've removed the prior configuration of the jakarta.el-api jar.

olamy (approved):

janbartel:
@joakime nudge

janbartel:
@joakime nudge

joakime (approved):

Issue #6974, opened 2021-10-08
Major websocket memory change in 9.4.36


slipcon:
**Jetty version(s)**

We've been using 9.4.18 for a long time, but having trouble updating to the latest 9.4.x.
I've narrowed the problem down to starting in 9.4.36.

**Java version/vendor** `(use: java -version)`

openjdk version "11.0.2" 2019-01-15

**OS type/version**

Seen on multiple OSes - MacOS, Windows & Linux.

**Description**

Our application has a websocket servlet, which creates a class that extends WebSocketAdapter and calls         session.getRemote().sendString( data, this ) to transmit JSON messages to the clients.    The clients send data back to the server over the same websocket, but it is very minimal heartbeat/keepalive JSON objects.     

Starting in 9.4.36,  we start to see significant memory utilization very quickly - specifically in the HeapByteBuffer objects, eventually leading to memory exhaustion and crashes.     

Disabling the permessage-deflate extension on the websocket fixes the issue, but we'd like to keep compression of the JSON messages on.   Our largest JSON messages are on the order of 200Kbytes.   We are not doing any sort of jetty memory tuning - and it has worked great prior to 9.4.36.

I suspect this change may be relevant:  https://github.com/eclipse/jetty.project/issues/5499

Please let me know if there is any more information I can provide or steps to try.

**How to reproduce?**

joakime:
Upgrade.

Jetty 9.4.36 has a buffer aggregation bug fixed in PR #6056 that could be biting you.

Please always upgrade to the newest version in your major version space (you are using 9.4.x, which means [9.4.44.v20210927](https://github.com/eclipse/jetty.project/releases/tag/jetty-9.4.44.v20210927) is your best choice for upgrade)

BTW, version 9.4.36 is subject to several published security advisories - https://www.eclipse.org/jetty/security_reports.php

slipcon:
Sorry I should have been clearer -  all versions 9.4.36 and newer, including 9.4.44,  show the same behavior for us.   The security advisories are why we are trying to move forward from 9.4.18

joakime:
> openjdk version "11.0.2" 2019-01-15

This should be upgraded as well.
Version 11.0.2 [expired on April 16, 2019](https://www.oracle.com/java/technologies/javase/11-0-2-relnotes.html)
[...]

PR #7017, opened 2021-10-20
Issue #6974 - improvements & fixes to ByteBufferPool implementations

lachlan-roberts:
## Issue #6974 - improvements to ByteBufferPool implementations

**ByteBufferPool Changes**

- Fixes bugs and raceconditions related to use of `ArrayByteBufferPool` and `MappedByteBufferPool` with non zero settings for `maxHeapMemory` and `maxDirectMemory`.
- Make a constructor argument of zero for `maxHeapMemory` and `maxDirectMemory` mean that a default is calculated based on the total max memory of the JVM.
- Make the `ArrayByteBufferPool` and `MappedByteBufferPool` extensible to change the bucket scaling.
- Implement the `LogArrayByteBufferPool` which doubles the size of each consecutive bucket instead of scaling linearly.
- Implement `Dumpable` for the ByteBufferPool implementations.

**WebSocket Changes**

- `WebSocketUpgradeFilter` always waits until it is started before it gets the `NativeWebSocketConfiguration` so it can be created when the `ByteBufferPool` is ready.
- WebSocket now uses the servers `ByteBufferPool` instead of creating its own `MappedByteBufferPool` every time.


jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
```diff
@@ -186,31 +197,41 @@ private void clearOldestBucket(boolean direct)
...
-        if (b >= _direct.length)
+        int index = getBucketNumber(capacity) - 1;
+        if (index >= _direct.length)
             return null;
         Bucket[] buckets = bucketsFor(direct);
-        Bucket bucket = buckets[b];
-        if (bucket == null && newBucket != null)
-            buckets[b] = bucket = newBucket.apply(b + 1);
-        return bucket;
+        Bucket bucket = buckets[index];
```
lorban: The way you arranged for thread-safety is incorrect, this pattern is called the _double-checked locking_ and can only be made correct by marking the initialized variable as `volatile`.

I don't think you can easily make an array's entries `volatile` so you may want to have a look at `AtomicReferenceArray` instead.

jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferPool.java
```diff
@@ -160,6 +159,7 @@ public void release(ByteBuffer buffer)
         private final int _maxSize;
         private final AtomicInteger _size;
         private final AtomicLong _lastUpdate = new AtomicLong(System.nanoTime());
+        private AtomicLong _poolSize;
```
lorban: You have to mark this variable as `volatile` to get the memory fencing you need between the thread that's going to write it and the threads that are going to read it.

This seems to indicate a weakness in your concurrency model, you should always try to avoid having to access non-final variables from concurrent code.

jetty-io/src/main/java/org/eclipse/jetty/io/MappedByteBufferPool.java
```diff
@@ -93,33 +99,34 @@ public MappedByteBufferPool(int factor, int maxQueueLength, Function<Integer, Bu
...
         super(factor, maxQueueLength, maxHeapMemory, maxDirectMemory);
-        _newBucket = newBucket != null ? newBucket : this::newBucket;
+        _newBucket = newBucket;
     }
 
-    private Bucket newBucket(int key)
+    private Bucket newBucket(int key, boolean direct)
     {
-        return new Bucket(this, key * getCapacityFactor(), getMaxQueueLength());
+        Bucket bucket = (_newBucket != null) ? _newBucket.apply(key) : new Bucket(this, key * getCapacityFactor(), getMaxQueueLength());
```
lorban: The call to the `_newBucket` function seems to be the only reason why you need `setPoolSizeAtomic()`. AFAICT this mechanism isn't used anywhere but in a test so I'd refactor the code so that the `Bucket` constructor takes the `AtomicLong`

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
```diff
@@ -186,31 +197,41 @@ private void clearOldestBucket(boolean direct)
...
-            // removed, so we need this null guard.
+            // Null guard in case this.clear() is called concurrently.
             if (bucket != null)
-                bucket.clear(this::decrementMemory);
+                bucket.clear();
         }
     }
 
-    private int bucketFor(int capacity)
+    protected int getBucketNumber(int capacity)
```
lorban: FYI, I personally have a preference for only using the `get` prefix for plain simple getters, unlike here where you do some calculation.

It's your call, but at least you know why the method was named `bucketFor` instead of `getBucketFor`.

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
```diff
@@ -95,12 +105,29 @@ public long getHeapMemory()
...
+        return _maxHeapMemory;
+    }
+
     public long getMemory(boolean direct)
     {
         AtomicLong memory = direct ? _directMemory : _heapMemory;
         return memory.get();
     }
 
+    public AtomicLong getSizeAtomic(boolean direct)
```
lorban: You definitely don't want this method to be public at the very least. This getter's existence (even if protected / package-private) even makes me question the idea of referencing the `AtomicLong` from the buckets.

lorban (changes requested):
I think the changes are good, except for the concurrency model which feels broken and should be reviewed.

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
lachlan-roberts: I have changed this to initialize all buckets in the constructor and never change the values in the array.

jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferPool.java
lachlan-roberts: I have now made this final and deprecated previous constructor, as bucket is public class as it is defined in an interface.

jetty-io/src/main/java/org/eclipse/jetty/io/MappedByteBufferPool.java
lachlan-roberts: The `setPoolSizeAtomic()` has been removed and replaced with new constructor on `Bucket`.

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
lachlan-roberts: Changed these methods to `bucketFor` and `capacityFor`.

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
lachlan-roberts: Changed this method to `protected`.

jetty-server/src/main/config/etc/jetty-bytebufferpool.xml
```diff
@@ -5,7 +5,7 @@
     <Arg type="int"><Property name="jetty.byteBufferPool.factor" default="1024"/></Arg>
```
gregw, line 5: Perhaps we should make the class in the `<New>` element a parameter, so the log version can easily be put in.  Not sure how simple it would be to do... but perhaps something like:
```xml
  <New id="byteBufferPool">
    <Class><Property name="jetty.byteBufferPool.class" default="org.eclipse.jetty.io.ArrayByteBufferPool"/></Class>
    <Arg type="int"><Property name="jetty.byteBufferPool.minCapacity" default="0"/></Arg>
    ...
```

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
```diff
@@ -155,26 +171,20 @@ public void clear()
...
+    protected void releaseMemory(boolean direct)
     {
         long oldest = Long.MAX_VALUE;
         int index = -1;
         Bucket[] buckets = bucketsFor(direct);
         for (int i = 0; i < buckets.length; ++i)
         {
             Bucket bucket = buckets[i];
-            if (bucket == null)
+            if (bucket == null || bucket.isEmpty())
```
lorban: This null check is now useless, isn't it?

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
```diff
@@ -98,24 +105,30 @@ public ArrayByteBufferPool(int minCapacity, int factor, int maxCapacity, int max
             maxCapacity = 64 * 1024;
         if ((maxCapacity % factor) != 0 || factor >= maxCapacity)
             throw new IllegalArgumentException("The capacity factor must be a divisor of maxCapacity");
+        _maxCapacity = maxCapacity;
         _minCapacity = minCapacity;
 
-        int length = maxCapacity / factor;
+        // Initialize all buckets in constructor and never modify the array again.
```
lorban, line 111: :+1: This is more in line with what `ArrayRetainableByteBufferPool` does in 10.0.x.

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
```diff
@@ -155,26 +171,20 @@ public void clear()
...
-            if (bucket != null)
-                bucket.clear();
-            _indirect[i] = null;
+            _direct[i].clear();
+            _indirect[i].clear();
         }
     }
 
-    private void clearOldestBucket(boolean direct)
+    protected void releaseMemory(boolean direct)
```
lorban, line 179: FYI in 10.0.x `ArrayRetainableByteBufferPool` has a more sophisticated memory release algorithm (the method is named `releaseExcessMemory`). I'd like your opinion about it, and it would be good that you'd familiarize with it too as your changes will have to be ported to 10.0.x and we certainly want to keep the code aligned everywhere we can.

lorban (dismissed):
LGTM

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
```diff
@@ -95,12 +105,29 @@ public long getHeapMemory()
...
     public long getMemory(boolean direct)
     {
         AtomicLong memory = direct ? _directMemory : _heapMemory;
         return memory.get();
     }
 
+    protected AtomicLong getSizeAtomic(boolean direct)
+    {
+        return (direct) ? _directMemory : _heapMemory;
+    }
```
sbordet, line 130: Uhm, no. Can you have in this base class a function that updates the atomics, called from the buckets? The idea is "tell, don't ask". 
The functions are there, but have been deprecated?

jetty-io/src/main/java/org/eclipse/jetty/io/LogArrayByteBufferPool.java
```diff
@@ -0,0 +1,118 @@
...
+
+package org.eclipse.jetty.io;
+
+/**
+ * Extension of the {@link ArrayByteBufferPool} whose bucket sizes increase exponentially instead of linearly.
+ * Each bucket will be double the size of the previous bucket, this decreases the amounts of buckets required
+ * which can lower total memory usage if buffers are often being acquired of different sizes. However as there are
+ * fewer buckets this will also increase the contention on each bucket.
+ */
+public class LogArrayByteBufferPool extends ArrayByteBufferPool
```
sbordet: Uh, not "Log", but "Logarithmic" -- we don't want to confuse it with a pool that logs to SLF4J.

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
```diff
@@ -32,16 +32,24 @@
...
+     * @param maxDirectMemory the max direct memory in bytes, -1 for unlimited memory or 0 to use default heuristic.
+     */
     protected AbstractByteBufferPool(int factor, int maxQueueLength, long maxHeapMemory, long maxDirectMemory)
     {
         _factor = factor <= 0 ? 1024 : factor;
         _maxQueueLength = maxQueueLength;
-        _maxHeapMemory = maxHeapMemory;
-        _maxDirectMemory = maxDirectMemory;
+        _maxHeapMemory = (maxHeapMemory != 0) ? maxHeapMemory : Runtime.getRuntime().maxMemory() / 4;
+        _maxDirectMemory = (maxDirectMemory != 0) ? maxDirectMemory : Runtime.getRuntime().maxMemory() / 4;
```
sbordet, line 53: Old default was unlimited, I would keep that.

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
```diff
@@ -238,4 +246,36 @@ private long getByteBufferCount(boolean direct)
...
+    @Override
+    public void dump(Appendable out, String indent) throws IOException
+    {
+        List<Object> dump = new ArrayList<>();
+        dump.add(String.format("HeapMemory: %d/%d", getHeapMemory(), getMaxHeapMemory()));
+        dump.add(String.format("DirectMemory: %d/%d", getDirectMemory(), getMaxDirectMemory()));
+
+        List<Bucket> indirect = Arrays.stream(_indirect).filter(b -> !b.isEmpty()).collect(Collectors.toList());
+        List<Bucket> direct = Arrays.stream(_direct).filter(b -> !b.isEmpty()).collect(Collectors.toList());
+        if (_detailedDump)
```
sbordet: Use the getter method rather than the field here.

jetty-io/src/main/java/org/eclipse/jetty/io/LogArrayByteBufferPool.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    protected int capacityFor(int bucket)
+    {
+        return 1 << bucket;
+    }
+
+    @Override
+    protected void releaseMemory(boolean direct)
```
sbordet: This is a total duplicate of the base class.

jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferPool.java
```diff
@@ -160,13 +159,21 @@ public void release(ByteBuffer buffer)
         private final int _maxSize;
         private final AtomicInteger _size;
         private final AtomicLong _lastUpdate = new AtomicLong(System.nanoTime());
+        private final AtomicLong _poolSize;
```
sbordet: Don't like this, see previous comment.

sbordet (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
lachlan-roberts: They need to be updated from Bucket which is a class contained in the ByteBufferPool interface which doesn't have access to methods from AbstractByteBufferPool.

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
lachlan-roberts: The whole point of this is to avoid an unbounded memory commitment, we want a default which will not cause the JVM to run out of memory.

jetty-io/src/main/java/org/eclipse/jetty/io/LogArrayByteBufferPool.java
lachlan-roberts: It actually does a different strategy for releasing the memory. The super method actually clears the whole bucket while this one only clears one buffer from the bucket.

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
sbordet: Pass the functions, not the fields, then.

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
sbordet: Sure but a change in behavior is what we can do in 10, no?

jetty-io/src/main/java/org/eclipse/jetty/io/LogArrayByteBufferPool.java
sbordet: I don't like that the whole bucket is cleared.
If you have a steady 100 acquires/s, you have 100 buffers in that bucket; a small spike would clear the whole bucket, causing a subsequent allocation spike, when the memory was only exceeded by 1 buffer, not all of them.

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
lachlan-roberts: This change is prompted by 9.4 WebSocket using the ByteBufferPool to allocate buffers of variable size and eventually running out of memory. So we do need to do something about this in 9.4 and not just 10.

jetty-io/src/main/java/org/eclipse/jetty/io/LogArrayByteBufferPool.java
lachlan-roberts: Clearing a whole bucket is the previous behaviour for ArrayByteBufferPool that I haven't changed. 

But for the log buckets there will be a lot less buckets so clearing a whole bucket is probably far too aggressive.

sbordet: You are correct that the the bucket was cleared before as well, I misread that.
However, it was removed from the array, becoming a "free floating" object that no other thread could access, so clearing it was guaranteed to be thread safe.
Now the bucket is kept, so a concurrent allocation may add buffers to it, while another thread is trying to clear it, possibly making the clearing longer.

sbordet:
@gregw @lachlan-roberts overall I feel that these changes are too much for 9.4.x.
If the problem is in WebSocket, I don't think we need such massive changes in the buffer pools.
I'm sure I'm missing something, so let's sync on this.

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
lachlan-roberts: We have the same method `releaseExcessMemory()` here in 9.4 but its on the base class `AbstractByteBufferPool`, this method just keeps looping calling a `Consumer<Boolean>` until the pool has less memory than the maxMemorySize. The `ArrayByteBufferPool` uses this new `releaseMemory()` method as its release strategy which it passes to the `releaseExcessMemory()` method. I don't think we can easily change this contract as people may be extending `AbstractByteBufferPool` themselves. But we could make our `releaseMemory()` smarter to avoid things like over releasing too much memory.

The `ArrayRetainableByteBufferPool` release strategy is based on oldest entry which we cannot do in `ByteBufferPool`. And I'm not sure I understand it, why do we bother finding the oldest entry? shouldn't all entries in a specific bucket have the same capacity so it won't matter which one is evicted?

jetty-server/src/main/config/etc/jetty-bytebufferpool.xml
lachlan-roberts: I can try this, only problem is that `factor` is a constructor argument for `ArrayByteBufferPool` but is kind of meaningless for the `LogarithmicByteBufferPool`. But I could add a dummy constructor in `LogarithmicByteBufferPool` for factor so it can be used in this way.

lachlan-roberts: `MappedByteBufferPool` does not have a matching constructor either, so I'm not sure how useful this will be.

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
lachlan-roberts: Yep, fixed.

jetty-io/src/main/java/org/eclipse/jetty/io/LogArrayByteBufferPool.java
lachlan-roberts: done

gregw:
@sbordet @lachlan-roberts let's hangout tonight(tomorrow morning)

gregw:
... but note that whilst I don't like changes in 9.4, I also don't like the unlimited memory commitment that we currently allow with the default pool with no easy way to change it.

sbordet (dismissed):

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
```diff
@@ -95,12 +106,34 @@ public long getHeapMemory()
...
+    IntConsumer updateMemory(boolean direct)
+    {
+        return getSizeAtomic(direct)::addAndGet;
+    }
+
+    private AtomicLong getSizeAtomic(boolean direct)
+    {
+        return (direct) ? _directMemory : _heapMemory;
+    }
+
```
gregw: Perhaps:
```suggestion
    IntConsumer updateMemory(boolean direct)
    {
        return (direct) ? _directMemory::addAndGet : _heapMemory::addAndGet;
    }

```

gregw (changes requested):
Can you provide a `bytebufferpool-log` modules that provides `bytebufferpool` and uses an alternative xml that configures the log byte buffers

gregw (approved):

PR #7032, opened 2021-10-22
Issue #7031 - Fixing ResponseWriter println methods

### LINKED ISSUE #7031: ResponseWriter.println(char) does not print newline
**Jetty version(s)**
10.0.7

openjdk version "11.0.9" 2020-10-20
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.9+11)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.9+11, mixed mode)

**OS type/version**
Windows 10, 64 bit

**Description**
When a single character is passed to `response.getWriter().println`, the newline after the character is not printed. In other words, for characters, `print` and `println` have the same effect.

Other overloads for other parameter types do not seem to have this problem, just character.

**How to reproduce?**

Call a servlet that writes single characters using `response.getWriter().println(char)` and verify the output.

joakime:
+ Improving test coverage on response.getWriter()
  and response.getOutputStream() usage


gregw (approved):
Needs to be some award for longest living bug!

Issue #7042, opened 2021-10-25
Simplify configuration to use different OpenIdConfiguration per webapp.


lachlan-roberts:
Currently to use different OpenID providers for different webapps you must configure everything in the context xml for each webapp and manually add security handler with `Authenticator` and `OpenLoginService` etc.

It would be a lot better if this could be simplified if this could be discovered and set up automatically based on realm name. Similar to how the `LoginService` is  found by the `SecurityHandler` from a bean on the Server.

lachlan-roberts:
Merged #7043.

PR #7043, opened 2021-10-25
Issue #7042 - Allow OpenIdConfiguration to be selected based on realm name

lachlan-roberts:
## Issue #7042

Multiple instances of `OpenIdConfiguration` can be added onto the server as beans, the webapp can choose which one to use by specifying its `realm name` as the `issuer` of the `OpenIdConfiguration` it wishes to use.

This allows you to use multiple different OpenIdConfiguratitons without having to specify any per-webapp jetty xml configuration.


jetty-openid/src/main/config/etc/jetty-openid.xml
```diff
@@ -20,30 +20,29 @@
...
+  <Call name="addBean">
+    <Arg>
+      <New id="OpenIdConfiguration" class="org.eclipse.jetty.security.openid.OpenIdConfiguration">
+        <Arg><Property name="jetty.openid.provider" deprecated="jetty.openid.openIdProvider"/></Arg>
+        <Arg><Property name="jetty.openid.provider.authorizationEndpoint"/></Arg>
+        <Arg><Property name="jetty.openid.provider.tokenEndpoint"/></Arg>
+        <Arg><Property name="jetty.openid.clientId"/></Arg>
+        <Arg><Property name="jetty.openid.clientSecret"/></Arg>
+        <Arg><Property name="jetty.openid.authMethod" default="client_secret_post"/></Arg>
+        <Arg><Ref refid="HttpClient"/></Arg>
```
gregw: Should use named Arg elements here to make the XML more readable and order independent

jetty-openid/src/main/config/etc/jetty-openid.xml
```diff
@@ -20,30 +20,29 @@
...
+        <Set name="authenticateNewUsers">
+          <Property name="jetty.openid.authenticateNewUsers" default="false"/>
+        </Set>
+        <Call name="addScopes">
```
gregw, line 41: should this be an add or a set?

jetty-openid/src/test/java/org/eclipse/jetty/security/openid/OpenIdProvider.java
```diff
@@ -217,10 +326,15 @@ public String getName()
...
+
+        public String getIdToken(String provider, String clientId)
         {
             long expiry = System.currentTimeMillis() + Duration.ofMinutes(1).toMillis();
-            return JwtEncoder.createIdToken(provider, clientId, Long.toString(subject), name, expiry);
+            return JwtEncoder.createIdToken(provider, clientId, subject, name, expiry);
         }
     }
-}
+}
```
gregw: no EOL!!!

gregw (changes requested):

jetty-openid/src/main/config/etc/jetty-openid.xml
lachlan-roberts: The method on `OpenIdConguration` is `addScopes`. It could be a set here but there is no set method currently.

jetty-openid/src/main/config/etc/jetty-openid.xml
lachlan-roberts: Isn't this fragile because you could easily change one of the constructor argument names but forget to change this XML.

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
```diff
@@ -79,38 +81,55 @@
...
     @Override
     public void setConfiguration(AuthConfiguration configuration)
     {
-        super.setConfiguration(configuration);
+        if (_configuration == null)
+        {
+            LoginService loginService = configuration.getLoginService();
+            if (!(loginService instanceof OpenIdLoginService))
+                throw new IllegalArgumentException("invalid LoginService " + loginService);
+            this._configuration = ((OpenIdLoginService)loginService).getConfiguration();
```
janbartel:
```suggestion
            _configuration = ((OpenIdLoginService)loginService).getConfiguration();
```

janbartel: There are too many different variables all named `configuration` or `_configuration` but of different types.

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticatorFactory.java
```diff
@@ -28,7 +28,20 @@ public Authenticator getAuthenticator(Server server, ServletContext context, Aut
     {
         String auth = configuration.getAuthMethod();
         if (Constraint.__OPENID_AUTH.equalsIgnoreCase(auth))
-            return new OpenIdAuthenticator();
+        {
+            // If LoginService is an OpenIdLoginService it already contains the configuration and will be obtained in setConfiguration();
```
janbartel: I don't understand this comment. Why isn't line 34 calling `new OpenIdAuthenticator(loginService.getConfiguration())` then?

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthConfiguration.java
```diff
@@ -0,0 +1,52 @@
...
+
+package org.eclipse.jetty.security.openid;
+
+import org.eclipse.jetty.security.Authenticator;
+import org.eclipse.jetty.security.LoginService;
+import org.eclipse.jetty.security.WrappedAuthConfiguration;
+
+public class OpenIdAuthConfiguration extends WrappedAuthConfiguration
+{
+    public static final String AUTHENTICATE_NEW_USERS_INIT_PARAM = "jetty.openid.authenticateNewUsers";
```
janbartel: Why is this init param necessary when the `OpenIdConfiguration` has a setter for this?

janbartel (changes requested):

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticatorFactory.java
lachlan-roberts: Yep good idea, I think this would be a better way to do it.

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthConfiguration.java
lachlan-roberts: I think this is a per webapp configuration of this setting, the `OpenIdConfiguration` setter is the default applied for all the webapps which use that an `OpenIdConfiguration`.

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
lachlan-roberts: I have renamed some of these to make it clearer.

jetty-openid/src/main/config/etc/jetty-openid.xml
gregw: You would get an error in that case as the name would not be found.

gregw (dismissed):

lachlan-roberts:
@janbartel I made the changes you requested.

janbartel (approved):

PR #7047, opened 2021-10-26
fix resource leak due to Files.newDirtoryStream

lujiefsi:
Files.newDirectoryStream will open stream and as jdk said:

 

If timely disposal of file system resources is required, the
{@code try}-with-resources construct should be used to ensure that the
stream's {@link Stream#close close} method is invoked after the stream
operations are completed.

we sould close them with try catch


olamy:
nice fix! but before merge it we need to have you sign the eclipse eca.

joakime:
Closing, as this change is no longer relevant.
A larger, more comprehensive check on `java.nio.file.Files` usages was done in other commits.

lujiefsi:
> Closing, as this change is no longer relevant. A larger, more comprehensive check on `java.nio.file.Files` usages was done in other commits.

https://github.com/eclipse/jetty.project/pull/7050

PR #7058, opened 2021-10-29
Issue #2504 - expose more WebSocket details in jetty server dump

### LINKED ISSUE #2504: Expose more WebSocket details in JMX and Server Dump
In light of the conversation in #2061 more information should be exposed about WebSocket in JMX and even the server dump.

Some initial thoughts ...

* Each WebSocket Server Container (either Jetty Native or JSR356) should be listed separately.
  * Max Message Size (TEXT / BINARY)
  * Idle Timeouts
  * Registered Potential Endpoints (url-pattern + creator / configurator / endpoint class)
    *  URI Template Variables (if using UriTemplatePathSpec)
* Each Client Container should be listed separately.
* The Sessions tracked within each Container should be accessible.
* Each Session details
  * In use Connection implementatino (jetty-io layer, identify type of connection)
  * Network EndPoints (jetty-io layer stuff, which endpoint impl)
  * WebSocket Endpoint (event handler in use, and user object/instance toString)
  * Negotiation Details
    * Request URL
    * URI Template Variable Values
    * Requested / Negotiated Sub Protocols
    * Requested / Negotiated Extensions
  * Idle Timeouts
  * Registered Decoders (if JSR356)
  * Registered Encoders (if JSR356)
  * Registered MessageHandlers (if JSR356)
  * User Properties (if JSR356)

lachlan-roberts:
closes #2504

### DIFF
--- a/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/Configuration.java
+++ b/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/Configuration.java
@@ -238,5 +238,15 @@ public interface Configuration
             if (maxOutgoingFrames != null)
                 configurable.setMaxOutgoingFrames(maxOutgoingFrames);
         }
+
+        @Override
+        public String toString()
+        {
+            return String.format("%s@%x{idleTimeout=%s, writeTimeout=%s, autoFragment=%s, maxFrameSize=%s, " +
+                    "inputBufferSize=%s, outputBufferSize=%s, maxBinaryMessageSize=%s, maxTextMessageSize=%s, maxOutgoingFrames=%s}",
+                getClass().getSimpleName(), hashCode(),
+                idleTimeout, writeTimeout, autoFragment, maxFrameSize, inputBufferSize, outputBufferSize,
+                maxBinaryMessageSize, maxTextMessageSize, maxOutgoingFrames);
+        }
     }
 }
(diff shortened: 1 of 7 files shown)


jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/Configuration.java
```diff
@@ -238,5 +238,15 @@ public void customize(Configuration configurable)
             if (maxOutgoingFrames != null)
                 configurable.setMaxOutgoingFrames(maxOutgoingFrames);
         }
+
+        @Override
+        public String toString()
+        {
+            return String.format("%s@%x{idleTimeout=%s, writeTimeout=%s, autoFragment=%s, maxFrameSize=%s, " +
+                    "inputBufferSize=%s, outputBufferSize=%s, maxBinaryMessageSize=%s, maxTextMessageSize=%s, maxOutgoingFrames=%s}",
```
joakime, line 246: Why are these numbers formatted with `%s` and not `%d` ?

joakime (changes requested):
Please fix the format type.
Do we have an existing unit test that can trigger the new code in this PR?

lachlan-roberts:
> Do we have an existing unit test that can trigger the new code in this PR?

@joakime I don't think there are currently any websocket tests for the output of the dump, I could write some if you want.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/Configuration.java
lachlan-roberts: I don't think there's any major difference between `%d` and `%s` for integers unless you want some special formatting.
 
These are also class types and not the primitive ones, so I thought `%s` might be better, they are `Duration`,  `Boolean`, `Long`, `Integer` and the values are null if no configuration is set for them.

joakime (approved):

PR #7077, opened 2021-11-04
Cleanup of Attributes class

gregw:
- Added Object returns from remove and set
 - Added simple Mapped type for when lazy creation is not useful
 - Added Layer for use by context persistent vs start/stop attributes
 - removed enumerator as being API specific
 - better unit test with good coverage

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/Attributes.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/Attributes.java
@@ -13,9 +13,14 @@
 
 package org.eclipse.jetty.util;
 
+import java.util.AbstractSet;
 import java.util.Collections;
-import java.util.Enumeration;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Attributes.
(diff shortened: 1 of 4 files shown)


gregw:
Actually, whilst I think this is really cleaver.... I don't think we need the name set to be mutable.  I will simplify....  stand by....

gregw:
Oh, something is really wrong with this branch... sorry.... fixing....

gregw:
@sbordet @lachlan-roberts that's better... fixed the branch to have only the changed Attributes files

jetty-util/src/main/java/org/eclipse/jetty/util/Attributes.java
```diff
@@ -104,15 +128,285 @@ public Object getAttribute(String name)
...
+            Object v = _map.remove(name);
+            return v == REMOVED ? null : v;
+        }
+
+        @Override
+        public Object setAttribute(String name, Object attribute)
+        {
+            if (attribute == null)
+                return removeAttribute(name);
+            Object v = _map.put(name, attribute);
```
lachlan-roberts, line 288: I would have expected that if you set an unrelated attribute it would be persisted onto the nested `Attributes`. For example if you're doing a `INCLUDE` dispatch and then you set a custom attribute you would expect it to be available after the include.

 Or is this `Layer` mechanism not intended to be used for dispatched attributes?

gregw: The use-case for this is the `ContextHandler` which contains a `Context`. Attributes set on the `ContextHandler` persist over multiple start/stop cycles, whilst attributes set on the `Context` are not and can be treated as a layer above the `ContextHandler` attributes.

It may be that `Layer` is only really specific to `ContextHandler` so I could move it there, but I wanted to take it out of `ContextHandler` (which currently has a half hearted layer impl) to reduce the volume of the `ContextHandler` class.

gregw:
Note also that I think there is a use-case for both `Attributed.Mapped` (simple usage with no volatile deref for long lived instances) and `AttributesMap` (short term usage with expectation of no attributes).  However, ultimately I'd like to rename to later to `Attributes.Lazy`.

lachlan-roberts (approved):

gregw:
already merged in another PR

PR #7080, opened 2021-11-04
Add a readOnly mode to HttpFields.Mutable.

gregw:
This is useful for when a response is committed.  
Taking an immutable from the response is good enough to write to the protocol, but the issue is that the mutable is left behind and can be further mutated, making request logging difficult.   

This approach allows the mutations to be ignored once the readonly mode is engaged... and has the benefit of avoiding and array copy as well... at the cost of a boolean check on every mutation.

An alternative would be to have a HttpFields wrapper that breaks the link to the wrapped HttpFields once it is committed, but I think that would be just as expensive and less obvious.

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpFields.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpFields.java
@@ -27,6 +27,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.function.BiFunction;
 import java.util.function.BiPredicate;
+import java.util.function.Function;
 import java.util.function.ToIntFunction;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


lorban:
From a high-level perspective, this change makes me question the existence of the `HttpFields.Immutable` class: if `HttpFields.Mutable` can be made read-only (said differently: immutable) while still being a `HttpFields.Mutable` instance then what's the point of `HttpFields.Immutable` at all?

gregw:
@lorban I kind of agree... at least that we need to ask the question.

However, I am still seeing two styles of usage:

1. You have an Mutable builder, which is reused over multiple cycles.  Each cycles takes an Immutable copy that has the nice efficiency of a final array that is efficiently copied from the mutable version.  In this case the Immutable doesn't even have the signature to be mutated.
2. You have a Mutable builder that is used only over a single cycle. Once it get's to a certain point in the cycle we wish to freeze it for all time (perhaps beyond the end of the cycle if somebody keeps a reference). In this case the Mutable always has the signature to be mutated, but they become noops.

I've tried to change style 1 into style 2, but it was a pain.

So if we accept that both styles exist, then perhaps the issue is just documentation to explain the difference between an `Immutable` and a read-only `Mutable`

An alternative is to accept that style 2 is really only specific to making a server response and solve the problem there by having both a Immutable and Mutable field and switching between them.   However, a similar issue might also exist in creating client Requests?

The way forward is either - commit as is with a TODO to review... OR I can try the alternative for style 2.

gregw:
this was a bad idea

PR #7081, opened 2021-11-05
Issue #7059 - prevent an internal NPE in AllowedResourceAliasChecker doStart (9.4)

lachlan-roberts:
## Issue #7059
Backport of changes from PR #7076 to `9.4.x` branch.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
@@ -84,7 +84,7 @@ public class AllowedResourceAliasChecker extends AbstractLifeCycle implements Co
         }
         catch (IOException e)
         {
-            LOG.warn("Initialization failed", e);
+            LOG.warn("Base resource failure ({} is disabled): {}", this.getClass().getName(), _base, e);
             _base = null;
         }
     }
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
@@ -97,13 +97,13 @@ public class AllowedResourceAliasChecker extends AbstractLifeCycle implements Co
         if (_contextHandler.isStarted())
             initialize();
         else
-            _contextHandler.addEventListener(_listener);
+            _contextHandler.addLifeCycleListener(_listener);
     }
 
     @Override
     protected void doStop() throws Exception
     {
-        _contextHandler.removeEventListener(_listener);
+        _contextHandler.removeLifeCycleListener(_listener);
         _base = null;
         _protected.clear();
     }
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
@@ -22,7 +22,6 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.LinkOption;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -30,6 +29,7 @@ import java.util.Objects;
 
 import org.eclipse.jetty.server.handler.ContextHandler;
 import org.eclipse.jetty.util.component.AbstractLifeCycle;
+import org.eclipse.jetty.util.component.LifeCycle;
 import org.eclipse.jetty.util.log.Log;
 import org.eclipse.jetty.util.log.Logger;
 import org.eclipse.jetty.util.resource.PathResource;
@@ -49,6 +49,7 @@ public class AllowedResourceAliasChecker extends AbstractLifeCycle implements Co
 
     private final ContextHandler _contextHandler;
     private final List<Path> _protected = new ArrayList<>();
+    private final AllowedResourceAliasCheckListener _listener = new AllowedResourceAliasCheckListener();
     protected Path _base;
 
     /**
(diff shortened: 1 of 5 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
```diff
@@ -49,48 +49,71 @@
...
+            {
+                for (String s : protectedTargets)
+                    _protected.add(_base.getFileSystem().getPath(_base.toString(), s));
+            }
+        }
+        catch (IOException e)
         {
-            for (String s : protectedTargets)
-                _protected.add(_base.getFileSystem().getPath(_base.toString(), s));
+            LOG.warn("Initialization failed", e);
```
joakime: Can this be a tad more descriptive?

```
LOG.warn("Base resource failure ({} is disabled): {}", this.getClass().getName(), _base, e);
```

jetty-server/src/main/java/org/eclipse/jetty/server/SymlinkAllowedResourceAliasChecker.java
```diff
@@ -45,6 +45,9 @@ public SymlinkAllowedResourceAliasChecker(ContextHandler contextHandler)
     @Override
     protected boolean check(String pathInContext, Path path)
     {
+        if (_base == null)
+            return false;
```
joakime, line 49: Ditto here, if a user adds this AliasChecker, but it's not being used due to an invalid base, it should indicate as such in the log.warn.

Warning once during initialization should be sufficient to convey this state.

joakime (changes requested):
Lets be more friendly to users.

jetty-server/src/main/java/org/eclipse/jetty/server/SymlinkAllowedResourceAliasChecker.java
lachlan-roberts: I don't think we should warn here. This alias checker is always added by default on a `ContextHandler` if `File.separatorChar == '/'`, I think there are valid uses of contextHandler that don't need a base resource so we don't want it always printing out warnings.

joakime: I'm not saying to warn here.
Warn during initialization (aka doStart())

> I think there are valid uses of contextHandler that don't need a base resource so we don't want it always printing out warnings.

This is a bad nuance, and I feel it is a bad argument too.

There's only 2 things in `ContextHandler` that uses the Base Resource.  

1. the `toString()` implementation.
2. the `getResource()` implementation.

IMO ...
The code path that adds the symlink alias checker should actually be in doStart(), not in the constructor of ContextHandler.
And it shouldn't add the alias checker unless there is a base resource.
And it shouldn't allow changing the base resource after start.

lachlan-roberts: > This is a bad nuance, and I feel it is a bad argument too.

We have implementations of `ContextHandler` that aren't supposed to be used with a base resource. For example `MovedContextHandler`, should you be required to set a base resource to not get a warning with this. Same goes for `HttpSpiContextHandler` and the `WellKnownHandler` (still in PR). As far as I'm aware all of these have no use for baseResource.

> The code path that adds the symlink alias checker should actually be in doStart(), not in the constructor of ContextHandler.

The problem is if they don't want the symlink alias check, previously they could just clear it after the constructor. At the point of `doStart()` we don't know if they wanted the symlink check or not. Did they clear the alias checker list or was the list always empty, did they add an addition to the list or did they clear the list and add a replacement alias check.

```java
ServletContextHandler contextHandler = new ServletContextHandler();
contextHandler.clearAliasChecks();
contextHandler.addAliasCheck(customAliasCheck);
```

Maybe we shouldn't change this behaviour in 9.4.

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
```diff
@@ -198,13 +221,43 @@ protected Path getPath(Resource resource)
         }
     }
 
+    private class AllowedResourceAliasCheckListener implements LifeCycle.Listener
```
gregw, line 224: geeeze we need to add default methods on that listener interface!

gregw (dismissed):

jetty-server/src/main/java/org/eclipse/jetty/server/AllowedResourceAliasChecker.java
joakime: Opened as Issue #7124

jetty-server/src/main/java/org/eclipse/jetty/server/SymlinkAllowedResourceAliasChecker.java
joakime: This is still possible to support with the move of the check to ContextHandler.doStart()

lachlan-roberts:
I have added changes suggested by @joakime.

This introduces a behaviour change. The following code:

```java
ContextHandler context = new ContextHandler();
context.setResourceBase(...);
context.addAliasCheck(customAliasCheck);
context.start();
```

Previously resulted in `aliasCheckers={symlinkAliasCheck, customAliasCheck}`, but will now result in just `aliasCheckers={customAliasCheck}`. So if anyone is relying on the default `SymlinkAllowedResourceAliasChecker` but is **also** adding another alias checker then they will be broken.

jetty-server/src/main/java/org/eclipse/jetty/server/SymlinkAllowedResourceAliasChecker.java
gregw: I think no matter what, changing to adding checkers in doStart is going to be too much of a behaviour change for 9.4.  As @lachlan-roberts says, we can't work out if they added a checker that wasn't the one we would add if they actually want it added or not.

So let's do minimal changes for 9.4.

For 10, 11 or 12, I'm OK with some behaviour changes... but I would prefer those to be that we don't add any aliase checkers by default.

joakime (approved):

PR #7092, opened 2021-11-10
Jetty 12 handler

gregw:
This PR has the new `Channel` and `Handler` architecture working in as much as `ChannelTest` and `HandlerTest` pass.
Other classes of specific interest to review are the handler implementations: `HelloHandler` and `EchoHandler`.


gregw:
I've rerouted the Trailers to be delivered as special content.   I think that works better than a random event delivered at some unspecified time.

jetty-http/src/main/java/org/eclipse/jetty/http/HttpVersion.java
```diff
@@ -23,7 +23,8 @@
     HTTP_0_9("HTTP/0.9", 9),
     HTTP_1_0("HTTP/1.0", 10),
     HTTP_1_1("HTTP/1.1", 11),
-    HTTP_2("HTTP/2.0", 20);
+    HTTP_2("HTTP/2.0", 20),
+    HTTP_3("HTTP/2.0", 30);
```
lorban: Copy/paste error; the string should be `HTTP/3.0`.

jetty-server/src/test/java/org/eclipse/jetty/server/EchoHandler.java
```diff
@@ -0,0 +1,91 @@
...
+import org.eclipse.jetty.http.HttpHeader;
+import org.eclipse.jetty.util.IteratingCallback;
+import org.eclipse.jetty.util.StringUtil;
+
+/**
+ * Dump request handler.
+ * Dumps GET and POST requests.
+ * Useful for testing and debugging.
+ */
+public class EchoHandler extends Handler.Abstract
```
gregw: @sbordet @lorban @lachlan-roberts  This is a good place to start looking jetty-12 Handlers.   This is a fully async handler that echos the content-type, content-length, the content itself and any trailers received.

jetty-server/src/main/java/org/eclipse/jetty/server/ConnectionMetaData.java
```diff
@@ -0,0 +1,106 @@
...
+
+package org.eclipse.jetty.server;
+
+import java.net.SocketAddress;
+
+import org.eclipse.jetty.http.HttpVersion;
+import org.eclipse.jetty.io.Connection;
+import org.eclipse.jetty.util.Attributes;
+
+public interface ConnectionMetaData extends Attributes
```
sbordet, line 22: Not sure this should be an interface. Seems just a Java 17 `record` to me, so it may just be that.
Do we need to act polymorphically on this class?

jetty-server/src/main/java/org/eclipse/jetty/server/ConnectionMetaData.java
```diff
@@ -0,0 +1,106 @@
...
+{
+    String getId();
+
+    HttpVersion getVersion();
+
+    String getProtocol();
+
+    Connection getConnection();
+
+    Connector getConnector();
```
sbordet, line 32: Not sure about returning non-primitive-ish objects.
We don't want applications to mess around with the connection (e.g. call upgrade and such).
Should not this be just a metadata holder?

jetty-jmx/src/main/java/org/eclipse/jetty/jmx/MBeanContainer.java
```diff
@@ -242,7 +242,7 @@ private static MetaData newMetaData(MBeanContainer container, Class<?> klass)
             return null;
         String pName = pkg.getName();
         String cName = klass.getName().substring(pName.isEmpty() ? 0 : pName.length() + 1);
-        String mName = pName + ".jmx." + cName + "MBean";
+        String mName = pName + ".jmx." + cName.replaceAll("\\$", "I") + "MBean";
```
sbordet, line 245: Ugh.

```java
class A {
  class Abstract {
  }
}

class AMBean {
  class AbstractMBean {
  }
}
```

Would it work to replace `A$Abstract` with `AMBean$AbstractMBean`, so `replaceAll("\\$", "MBean$")`?

jetty-util/src/main/java/org/eclipse/jetty/util/MultiMap.java
```diff
@@ -60,14 +60,25 @@ public MultiMap(MultiMap<V> map)
...
+     * Get the first value from a multiple value.
+     *
+     * @param name The entry key.
+     * @return Unmodifieable List of values.
+     */
+    public V getValue(String name)
+    {
+        return getValue(name, 0);
+    }
+
```
sbordet, line 75: Do we still need this class at all? Feels like with modern JDK `Map` with `compute()`, `merge()`, `flatMap()`, etc we can get rid of this class?

jetty-server/src/main/java/org/eclipse/jetty/server/ConnectionMetaData.java
gregw: Once we switch the 17, then I think a record might be fine for it.... although we do already have multiple impls for it.  HttpConnection implements it, then we have a Wrapper impl with several variants of that.

jetty-server/src/main/java/org/eclipse/jetty/server/ConnectionMetaData.java
gregw: Agree that they are kind of out of place and we want to reduce access to them as much as possible.
Kind of just there as place holders to help things working.  Currently only really needed for things like CONNECT and extracting x509 certificates from the connector.   Both of those could probably be abstracted away.  Will TODO removal of them.

jetty-jmx/src/main/java/org/eclipse/jetty/jmx/MBeanContainer.java
gregw: Can't have $ in the classname unless it really is an inner class.

One solution is to just not have so many inner classes.... but I'm finding it useful to have less files to look at whilst doing the refacting.    But ultimately we need to review if we should move many of those inner classes to the top level.

However, either way it would be good if the MBean mechanism could be made to work for inner classes.  

Would "Inner" work better?   So if we are looking for a MBean for `Foo$Bar` we would look for `FooInnerBarMBean`?
Or maybe flip, so we look for `BarInFooMBean` ?

gregw:
@sbordet I've updated this PR with much of the feedback received last night, except for serializing the callback/events on Channel.
I think that is the next goal for development, as I think it will lock down a lot of concepts.... so pushing forward towards that.

gregw:
@sbordet @lorban @lachlan-roberts I've tried for the past few hours to implement the serialization of events that we discussed in the call last night.   I'm struggling to make it work/fit in any of the ways we described.

Firstly, the good news, is that it works well having Channel be the centre of truth  - ie. it holds the core state.   Currently that is implemented by several atomics, but as we discussed that is almost certainly racey, so I will replace with a lock and plain fields soon.

However, other than mutually excluding calls that wish to grab the lock at the same (short) time, the concept of mututally excluding active calls into the application doesn't actually work, nor having a queue of events.   The reason is that the act of calling out to the application has little significance to the actual state of the stream/channel because everything is now fully asynchronous.      The places we call the application are:
 + `Handler.handle`, called from the `Runnable` returned by `onRequest`
 + The `onContentAvailable` call back, passed to the `request.demandContent` method and called from the `Runnable` return from `Channel.onContentAvailable` method
 + The `succeeded` or `failed` callbacks after a `response.write` operation completes, called from the callback passed to `stream.send`
 + Whatever error listener mechanism we come up with.

The issue is that the return of any of these calls is just not significant with regards to the `Channel` state. These calls can be noops, they can block or they can pass off handling to other threads and/or asynchronous processes.  What they will do is entirely unrelated to if they have returned or not.   In fact the only return that is meaningful is the boolean return from `Handler.handle` that idicates that something will be done (or may have already been done). 

If the method `channel.onIdleTimeout` has been called, it does not matter if we are currently calling any of these methods, it only matters if any of these methods have managed to do something stateful just before or just after the idle timeout.  If any of these calls (or threads/async processes they invoked), managed to call `write` or `readContent` the nano second before the idle timeout, then we can ignore the idle timeout.    If they called those methods a nano second after, then too late and we have already timed out and will be in process of delivering that exception to the application.

So I do think we serialize events at the `Channel`, but it's just that calling the application is not a significant event, rather it is more a consequence of a previous significant event and may or may not cause the occurance of a subsequent significant event.

I'll ponder some more and update you all tomorrow.

sbordet:
@gregw I don't think it's about serializing the calls to the "application". It's about serializing the events that happen in `Channel`, as in `onRequest()` should not be called concurrently with `onIdleTimeout()` or `onFailure()`, etc.

Note that it's not enough to use a lock, because we don't want to call the application with a lock held, so we need to serialize the invocation of tasks that modify the `Channel` state.

To be more precise, let's assume we read request bytes but the last byte of the headers is not arrived yet.
Time passes and we have an idle timeout racing with a NIO wakeup to read the last byte of the headers.

What I would like to happen is that the NIO wakeup "enqueues" a `Runnable` action that calls `fillAndParse()`; and the idle timeout also enqueues a `Runnable` action that calls `onIdleTimeout()`.
Let's assume the idle timeout wins, so `onIdleTimeout()` gets executed serially, flips a state (so that subsequent `Runnable` action won't be run) and figures out whether application has already been called: if so, returns a `Runnable` job to pass to the `ExecutionStrategy` to call some application error callback, otherwise sets a terminal state and produces an error response (or just closes the stream).

The same thread is now done handling the idle timeout action looks at the queue and sees there is another action to run.
This is the `fillAndParse()` action which may just check the state, sees that it's all over and has nothing to do (as the stream is closed already).

It is crucial that these action-running threads do not block, but this is nicely guaranteed by the fact that `Runnable` jobs are returned to call the application and passed to the `ExecutionStrategy`.
And thanks to the state in `Channel` each action knows its state and whether to return a `Runnable` job for the `ExecutionStrategy`.

So it's not about serializing `Handler.handle()` with `onContentAvailable()` with `succeeded()`, etc.

It's about serializing, in who calls `Channel` (`Stream`?) the calls to `Channel.onRequest()`, `Channel.onIdleTimeout()`, etc.

gregw:
@sbordet i think we are saying the same thing, but just in different ways. I'll implement my understanding of it and we'll iterate from there.

gregw:
@sbordet OK so now I think I've hit an incompatibility between `ExecutionStrategy` and the idea of serializing the `Runnable`s returned from `Channel`


Let's say we are currently running the `Runnable` returned by a call to  `onContentAvailable()`.    Prior to running it, we check the `InvocationType` and saw that it is `NON_BLOCKING`, so the executor lets the very last selector thread call it, because it's not blocking......   but whilst running it, something else (perhaps this thread or perhaps another) demands more content, which is available, so `onContentAvailable()` is called again.  But we don't want it to return a `Runnable` that will execute in parallel, instead we want to queue that `Runnable`, so that it is called by the thread currently a running the first runnable when it completes.   BUT!! what if the second `Runnable` is blocking???   We can't run it in the original thread, as that was run on the basis of being non-blocking.

So either:

 + we just always assume all these callbacks are non-blocking - Never going to happen as they will eventually call the app.  We can only guarantee non-blocking if we are waking up a app thread only.
 + we always assume they are blocking.  yuck dispatch all the time.
 + we have to remember on what basis we started our execution.  If we were started non-blocking and find a blocking task as we exit, then we will have to execute another thread to continue that work

gregw:
@sbordet @lorban @lachlan-roberts The next code I want to have reviewed is around the new `SerializedInvoker` class I have added, it's test and it's usage.
Once we are happy that is the right semantic, then we can dive into the TODOs inside `Channel.onError`

Note that `SerializedInvokerSafe` is locked queue implementation of the semantic that I think is pretty clear.  The main implementation is `AtomicReference` based and needs careful review.... I think the unit test is reasonable, but hard to test for real races.

sbordet:
> what if the second Runnable is blocking? We can't run it in the original thread, as that was run on the basis of being non-blocking.

If the second `Runnable` is blocking, the `ExecutionStrategy` will queue it into the thread pool and the non-blocking thread will continue to run non-blocking code, possibly going back to `select()`.

gregw:
Ooops I've pushed too many changes, so you can see 500+ files.   This PR is not so great to review now.  Anyhow....

@sbordet @lorban @lachlan-roberts I spent a bit of time cleaning up the state fields in `Channel`, `ChannelRequest` & `ChannelResponse`, to try a be a bit more consistent about them.    I have applied the following principals:

 + The request is immutable and somewhat durable beyond the end of it's lifecycle - ie it a reference to it held long after the request cycle will still see the request.
 + The response is mutable but becomes somewhat immutable and durable when it is committed - ie a reference to it held long after the cycle has completed will still see the response as it was sent.
 + All the state is protected by the single lock in channel
 + I've minimize duplication, so now the channel holds the request which hold the response.  This reduces the number of fields that need to be reset when recycling the channel.

gregw:
@sbordet @lachlan-roberts @lorban @joakime @janbartel I have rebased my Handler work onto the latest `java-12.0.x` branch layout.  To avoid losing git history, I have moved all the files I have not updated out of `jetty-server` into `jetty11-server`, so we can move them out of there as we resurrect them, without losing git history.

The tests in `jetty-server` are passing for me, so it is probably a reasonable basis for a) starting thinking about basic benchmarking; b) starting to work on servlet API wrappers; c) moveing any residual session stuff from `jetty11-server` to their new homes; d) continuing to work on some core features that may belong in `jetty-server` such as resource service and multi-part

gregw:
@sbordet @lorban @lachlan-roberts @joakime @janbartel 
As discussed with Simone, I will soon just merge this work onto 12.0.x

However, I have split it into two commits - the first being changes to jetty-util, jetty-io and jetty-http, and the second is the monster change in jetty-server.    Could you give some light review to the first commit.... just for any show stoppers.  If nothing is heard I will merge tomorrow this time.

jetty-util/src/main/java/org/eclipse/jetty/util/Attributes.java
```diff
@@ -104,15 +128,285 @@ public Object getAttribute(String name)
...
+        @Override
+        public Object removeAttribute(String name)
+        {
+            return _map.remove(name);
+        }
+
+        @Override
+        public Object setAttribute(String name, Object attribute)
+        {
+            if (attribute == null)
```
lorban, line 181: I think it would be more sensible to throw a NPE here.

jetty-util/src/main/java/org/eclipse/jetty/util/Attributes.java
```diff
@@ -104,15 +128,285 @@ public Object getAttribute(String name)
...
+        {
+            _map.clear();
+        }
+
+        public int size()
+        {
+            return _map.size();
+        }
+
+        public Set<Map.Entry<String, Object>> getAttributes()
```
lorban: The return type is... odd. A `Set` of `Map.Entry` actually is a `Map`, isn't it?

So the return type should be `Map<String, Object>` or `Set<Attribute>`, shouldn't it?

In the latter case, I think this method should then be called `getAll()` as it is the sibling of `addAll()`.

jetty-util/src/main/java/org/eclipse/jetty/util/Attributes.java
```diff
@@ -104,15 +128,285 @@ public Object getAttribute(String name)
...
+        }
+    }
+
+    /**
+     * An Attributes implementation backed by another Attributes instance, which is treated as immutable, but with a
+     * ConcurrentHashMap used as a mutable layer over it.
+     */
+    class Layer implements Attributes
+    {
+        private static final Object REMOVED = "REMOVED";
```
lorban:
```java
private static final Object REMOVED = new Object();
```

jetty-util/src/main/java/org/eclipse/jetty/util/Attributes.java
```diff
@@ -104,15 +128,285 @@ public Object getAttribute(String name)
...
+                }
+                return true;
+            }
+            return false;
+        }
+    }
+
+    /**
+     * An Attributes implementation backed by another Attributes instance, which is treated as immutable, but with a
+     * ConcurrentHashMap used as a mutable layer over it.
```
lorban, line 247: I think using a `ConcurrentHashMap` to back the changes is useless and counter-intuitive as most (all?) method implementations aren't atomic.

I'd either change the method implementations to make them atomic (hard) or use a plain old `HashMap` as a backing  store and changing the javadoc to state that this is not thread safe, or use a lock and just state that this impl is thread-safe.

jetty-util/src/main/java/org/eclipse/jetty/util/Attributes.java
```diff
@@ -104,15 +128,285 @@ public Object getAttribute(String name)
...
+            int hash = super.hashCode();
+            for (String name : getAttributeNames())
+                hash = hash ^ name.hashCode() ^ getAttribute(name).hashCode();
+            return hash;
+        }
+
+        @Override
+        public boolean equals(Object o)
+        {
+            if (o instanceof Attributes)
```
lorban, line 333: I would not write `equals` such as it can return `true` when two different implementations are used as it's too easy to break the `equals/hashcode` contract and/or have `a.equals(b) == true` while `b.equals(a) == false`.

This smells brittleness and is too easy to break in the future.

jetty-util/src/main/java/org/eclipse/jetty/util/BlockingCallback.java
```diff
@@ -0,0 +1,82 @@
...
+ *     {
+ *         someAsyncCall(args, blocker);
+ *     }
+ * }
+ * </pre>
+ * @see SharedBlockingCallback
+ */
+public class BlockingCallback implements AutoCloseable, Callback, Runnable
+{
+    private static final Throwable SUCCEEDED = new Throwable();
```
lorban: Maybe override `fillInStackTrace` as the stack trace is never going to be needed?

jetty-util/src/main/java/org/eclipse/jetty/util/BlockingCallback.java
```diff
@@ -0,0 +1,82 @@
...
+    @Override
+    public void succeeded()
+    {
+        _future.complete(SUCCEEDED);
+    }
+
+    @Override
+    public void failed(Throwable x)
+    {
+        _future.complete(x == null ? new Throwable() : x);
```
lorban: A question that popped recently in a discussion I had with @sbordet : should the `Callback.failed` contract explicitly state that it is forbidden to pass a null exception?

jetty-util/src/main/java/org/eclipse/jetty/util/BlockingCallback.java
```diff
@@ -0,0 +1,82 @@
...
+ * {
+ *     try(BlockingCallback blocker = new BlockingCallback())
+ *     {
+ *         someAsyncCall(args, blocker);
+ *     }
+ * }
+ * </pre>
+ * @see SharedBlockingCallback
+ */
+public class BlockingCallback implements AutoCloseable, Callback, Runnable
```
lorban: I don't like the `Runnable` implementation as this is mixing concerns that could be handled elsewhere.

jetty-util/src/main/java/org/eclipse/jetty/util/IO.java
```diff
@@ -273,6 +274,24 @@ public static void copyFile(File from, File to) throws IOException
...
+    public static IOException rethrow(Throwable cause)
+    {
+        if (cause instanceof IOException)
+            return (IOException)cause;
+        if (cause instanceof Error)
+            throw (Error)cause;
+        if (cause instanceof RuntimeException)
+            throw (RuntimeException)cause;
+        if (cause instanceof InterruptedException)
+            return new InterruptedIOException()
```
lorban: Save one class?
```java
return (InterruptedIOException) new InterruptedIOException().initCause(cause);
```

jetty-util/src/main/java/org/eclipse/jetty/util/MultiMap.java
lorban: +1

jetty-util/src/main/java/org/eclipse/jetty/util/SharedBlockingCallback.java
```diff
@@ -111,7 +112,7 @@ protected void notComplete(Blocker blocker)
      * callback do not blocak, rather they wakeup the thread that is blocked
      * in {@link #block()}
      */
-    public class Blocker implements Callback, Closeable
+    public class Blocker implements Callback, Closeable, Runnable
```
lorban, line 116: Same as above: I do not like this mixing of concerns.

jetty-util/src/main/java/org/eclipse/jetty/util/SharedBlockingCallback.java
```diff
@@ -223,7 +230,12 @@ public void block() throws IOException
             catch (final InterruptedException e)
             {
                 _state = e;
-                throw new InterruptedIOException();
+                throw new InterruptedIOException()
```
lorban: Same as above: we could save one class here.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/SerializedExecutor.java
```diff
@@ -34,65 +35,271 @@
...
+        return runnable;
+    }
+
+    /**
+     * Arrange for a task to be executed, mutually excluded from other tasks.
+     * This is equivalent to executing any {@link Runnable} returned from {@link #offer(Runnable)}
+     * @param task The task to invoke
+     */
     @Override
     public void execute(Runnable task)
```
lorban, line 72: I find the `Runnable offer(Runnable)` vs `void execute(Runnable)` contract confusing.

I think the reason to have a `Executor` implementation that also have this extra `offer` contract that does not queue the task should be better documented.

I'm starting to wonder if the whole `Executor` / `ExecutionStrategy` / `Invocable` mechanism shouldn't be refreshed?

jetty-util/src/test/java/org/eclipse/jetty/util/BlockingCallbackTest.java
```diff
@@ -0,0 +1,161 @@
...
+                blocker.failed(ex);
+            }
+            fail("Should have thrown IOException");
+        }
+        catch (IOException e)
+        {
+            start = TimeUnit.NANOSECONDS.toMillis(System.nanoTime());
+            assertEquals(ex, e.getCause());
+        }
+        assertThat(TimeUnit.NANOSECONDS.toMillis(System.nanoTime()) - start, lessThan(100L));
```
lorban: Why the timing assertion?

jetty-util/src/test/java/org/eclipse/jetty/util/BlockingCallbackTest.java
```diff
@@ -0,0 +1,161 @@
...
+                    }
+                    blocker.succeeded();
+                }
+            }).start();
+
+            latch.await();
+            start = TimeUnit.NANOSECONDS.toMillis(System.nanoTime());
+        }
+        assertThat(TimeUnit.NANOSECONDS.toMillis(System.nanoTime()) - start, greaterThan(10L));
+        assertThat(TimeUnit.NANOSECONDS.toMillis(System.nanoTime()) - start, lessThan(1000L));
```
lorban: Why the timing assertions?

jetty-util/src/test/java/org/eclipse/jetty/util/BlockingCallbackTest.java
gregw: How else do you make a test flakey???    These tests are copied from SharedBlockingCallback, so I didn't look that closely.    I think they are just checking that something blocked or didn't block... ie didn't take too much time in this case.    I don't think I've seen SharedBlockingCallbackTest as a flakey test, so I think they are pretty reliable.

jetty-util/src/test/java/org/eclipse/jetty/util/BlockingCallbackTest.java
gregw: see above

jetty-util/src/main/java/org/eclipse/jetty/util/thread/SerializedExecutor.java
gregw: I'm definitely wondering what this serializable executor is worth it for the handler arch.  The actions are all serialized by a lock on HttpChannel now, so I'm not sure it is a big problem if multiple callbacks run at the same time.  So I'm dubious that this class is really needed.

However, assuming that it is.... which is the name you find most confusing?    `offer` or `execute`.   Personally I think `execute` is OK as this is just an executor and so externally that is the right contract and internally the mutual exclusion is just a facet of the particular executor.    @sbordet suggested the `offer` name and I initially thought it was good... but now I can see that it is not exactly right, as the executor doesn't have the option to refuse the offer. The semantic is that the task is being submitted to be run and that is done either by another thread or by the runnable that is returned.  Would `submit` be better?  or just javadoc the existing names?

jetty-util/src/main/java/org/eclipse/jetty/util/Attributes.java
gregw: It is existing behaviour to treat a set of null as a remove and that originates from the servlet spec.  So if we changed, then we'd have to have the same "null is a remove" logic in a wrapper class for servlets,

jetty-util/src/main/java/org/eclipse/jetty/util/Attributes.java
gregw: good point!

jetty-http/src/main/java/org/eclipse/jetty/http/HttpVersion.java
gregw: doh!

jetty-util/src/main/java/org/eclipse/jetty/util/Attributes.java
gregw: The benefit of using a String is that it dumps well.  I could use an Object, but then would need to override the toString so it dumps/debugs well.

jetty-util/src/main/java/org/eclipse/jetty/util/Attributes.java
gregw: Nobody calls this method and it's not in the signature.... removing!

jetty-util/src/main/java/org/eclipse/jetty/util/Attributes.java
gregw: It has to be thread safe as we are encouraging async style in the handlers, so different threads will be coordinating via request attributes.
But I agree that a concurrent losely layered over another concurrent is almost certainly wrong.    Probably need a hard atomic lock to avoid test-then-act issues.
For now, I've put in a big TODO to review.

jetty-util/src/main/java/org/eclipse/jetty/util/Attributes.java
gregw: Well at the very least, the implementation was wrong, also for hashCode!   But I'm generally OK with `Attributes` implementations being equal if they have the same contents.

jetty-util/src/main/java/org/eclipse/jetty/util/BlockingCallback.java
gregw: Good question.   I think yes,  but even if it did, we'd need some defense for it.   So for now I'd like to leave as is and that can be a cleanup in another PR.

jetty-util/src/main/java/org/eclipse/jetty/util/BlockingCallback.java
gregw: We have some async methods that just take a runnable, whilst others take a Callback.   Both need this kind of blocking semantic.
I guess we could have a `BlockingRunnable` that only wraps a countdown latch....   Is that the right name?  For now I'll add a TODO to factor out.

gregw: Note that I've done the same Runnable thing with `SharableBlockingCallback`.    
I think we probably need to refactor all of these classes to be a single Blocker class with inner variations, rather than 3 separate classes.   I've added a TODO.

jetty-util/src/main/java/org/eclipse/jetty/util/SharedBlockingCallback.java
gregw: Punted to a TODO

jetty-util/src/main/java/org/eclipse/jetty/util/MultiMap.java
gregw: Yeah it feels very out of date.   But can we punt to a different PR to consider replacing it.  Would be best to do when everything is ported and compiling again so we can refactor all at once.  I will add a TODO.

jetty-util/src/main/java/org/eclipse/jetty/util/SharedBlockingCallback.java
gregw: See #7270

jetty-util/src/main/java/org/eclipse/jetty/util/BlockingCallback.java
gregw: See #7270

jetty-util/src/main/java/org/eclipse/jetty/util/thread/SerializedExecutor.java
lorban: What I find confusing is the `offer` contract: queue a task which may or may not get executed as a response to this call and optionally return you a `Runnable` that you have to get executed externally is _extremely_ odd for an `Executor`.

I feel like this contract should live in some class that does not implement `Executor` but does delegate to one.

jetty-util/src/main/java/org/eclipse/jetty/util/Attributes.java
lorban: The drawback is that using this special string as a value is exploitable, either or purpose or by accident.

jetty-util/src/main/java/org/eclipse/jetty/util/Attributes.java
lorban: Okay, but that should be clearly documented in the javadoc then.

jetty-util/src/main/java/org/eclipse/jetty/util/Attributes.java
lorban: Since this needs to be thread-safe, I think going the `HashMap` + lock route is the best way to go as it's easy to implement. If this lock ever becomes a bottleneck, I'll be fairly simple to spend the necessary time to get rid of it.

But I'm happy with a TODO for now.

jetty-util/src/test/java/org/eclipse/jetty/util/BlockingCallbackTest.java
lorban: You don't want your tests to ever be flakey, do you? In this case, a simple long GC at the wrong time would make this test fail while there's nothing wrong with the code.

It's just me rambling again each time I see a test that relies on time passing, I think I developed an allergy to them over the years. :-)

jetty-util/src/main/java/org/eclipse/jetty/util/thread/SerializedExecutor.java
gregw: It actually started off that way, without any reference to an `Executor` and was called the `SerialInvoker`, but then there was need for a convenience method that did the execute... so it was added... and then it looked like and executor... which was pretty much the already existing `SerialExecutor` class... so I merged them.

So I agree it's a little it strange.   But can we first ask the big question..... what is it doing for us?   Do we really need it at all?     The original idea was to serialize all the events arriving on `HttpChannel` into a queue, that would not need locks because only a single thread was ever executing the queue (protected by whatever mutex is in this class or equivalent).   But that idea didn't work out because... shit I've forgotten why that didn't work out... we need to write that down.  Anyway, that didn't work out and now we lock on HttpChannel and then serialize the consequences of events.    But do we really need to serialize them?  They can't do any actions significant themselves without first grabbing the lock, so we have a lock and a queue and an lock in the mutex for the queue.... maybe we are the worst of both worlds?

Anyway, I'd like to commit in it's current form and I will remember what the issue was with the fully serialized approach and write that up in another issue for posterity.  We can then consider just removing this class all together (although it is used in one other place).   If we don't remove it, then perhaps rename it and remove the contained executor?

gregw: @lorban actually, I thought about this some more and serializing does do an important job for this architecture: it prevents infinite recursion of a callback calling a method that calls a callback etc.

So for now, I have split out `SerializedInvoker` from `SerializedExecutor` (although the former name is wrong).   I have also removed the anti pattern of pretending we are a scheduler.  A serialized runnable is always a BLOCKING runnable as a blocking task may be queued at any time.

jetty-util/src/test/java/org/eclipse/jetty/util/BlockingCallbackTest.java
gregw: I forgot the "/s" for sarcasm.   I agree that such tests are not the best, but I'm not in the mood to rewrite at the moment.

PR #7096, opened 2021-11-11
Add support for gather writes for HttpConnection.SendCallback in Jetty 12

lachlan-roberts:
The `HttpConnection.SendCallback.reset()` now takes `ByteBuffer[]` content instead of a single `ByteBuffer` so it can do gather writes to the `EndPoint`.


jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -709,7 +712,7 @@ public InvocationType getInvocationType()
             return _callback.getInvocationType();
         }
 
-        private boolean reset(MetaData.Request request, MetaData.Response info, ByteBuffer content, boolean last, Callback callback)
+        private boolean reset(MetaData.Request request, MetaData.Response info, ByteBuffer[] content, boolean last, Callback callback)
```
gregw: As we are changing the signature of this method, let's change it's name also (as I always thought it was not right).  How about `prepare` ?

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -743,13 +746,13 @@ public Action process() throws Exception
...
             {
-                HttpGenerator.Result result = _generator.generateResponse(_info, _head, _header, _chunk, _content, _lastContent);
+                HttpGenerator.Result result = _generator.generateResponse(_info, _head, _header, _chunk, _content, getRemainingContent(), _lastContent);
                 if (LOG.isDebugEnabled())
                     LOG.debug("generate: {} for {} ({},{},{})@{}",
                         result,
                         this,
                         BufferUtil.toSummaryString(_header),
-                        BufferUtil.toSummaryString(_content),
+                        Arrays.toString(_content),
```
gregw: probabably need a better dump than this.  Need a utility to get summary or detail strings of an array of buffers.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -788,7 +791,7 @@ public Action process() throws Exception
                         if (_head || _generator.isNoContent())
                         {
                             BufferUtil.clear(_chunk);
-                            BufferUtil.clear(_content);
+                            Arrays.stream(_content).forEach(BufferUtil::clear);
```
gregw: I always worry about stream API on critical path code..... is this going to allocate an iterator when a simple for loop would not?  @lorban @sbordet ???

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -803,28 +806,29 @@ public Action process() throws Exception
                             gatherWrite += 2;
                             bytes += _chunk.remaining();
                         }
-                        if (BufferUtil.hasContent(_content))
+                        long contentLength = getRemainingContent();
```
gregw: Isn't this going to be the same length as calculated above?

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -803,28 +806,29 @@ public Action process() throws Exception
...
                             gatherWrite += 1;
-                            bytes += _content.remaining();
+                            bytes += contentLength;
                         }
                         HttpConnection.this.bytesOut.add(bytes);
                         switch (gatherWrite)
                         {
                             case 7:
-                                getEndPoint().write(this, _header, _chunk, _content);
+                                getEndPoint().write(this, ArrayUtil.add(new ByteBuffer[]{_header, _chunk}, _content));
```
gregw: This could probably be more efficient, as you are allocating a small array just to be added to the content.
Perhaps a special purpose prependToArray that takes 2 items to be prepended?

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -803,28 +806,29 @@ public Action process() throws Exception
...
                                 break;
                             case 6:
                                 getEndPoint().write(this, _header, _chunk);
                                 break;
                             case 5:
-                                getEndPoint().write(this, _header, _content);
+                                getEndPoint().write(this, ArrayUtil.prependToArray(_header, _content, ByteBuffer.class));
```
gregw, line 825: This is clever.  I had not thought of doing it this way!

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
```diff
@@ -890,6 +894,11 @@ private void releaseChunk()
             _chunk = null;
         }
 
+        private long getRemainingContent()
```
gregw, line 914: Does this need to be calculated in the loop?   Can't we just work out the length once when the content is reset/prepared ?

jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
```diff
@@ -570,13 +571,18 @@ private void generateResponseLine(MetaData.Response response, ByteBuffer header)
     }
 
     private void generateHeaders(ByteBuffer header, ByteBuffer content, boolean last)
+    {
+        generateHeaders(header, content == null ? null : new ByteBuffer[]{content}, BufferUtil.length(content), last);
+    }
+
+    private void generateHeaders(ByteBuffer header, ByteBuffer[] content, long contentSize, boolean last)
```
gregw: I'd very much prefer it if we could avoid these temporary arrays and instead only work with contentLength:

```suggestion
    {
        generateHeaders(header, BufferUtil.length(content), last);
    }

    private void generateHeaders(ByteBuffer header, long contentSize, boolean last)
```

I think the generator only every uses the contentLength and then does the occassional clear.  Is the clear really necessary and could it be done outside of the loop?

gregw (changes requested):
The changes are not as impactful as I had thought.... but there are a few tweaks needed.

I'm slightly concerned at the increased complexity for something that might never really be used. So I think you should fix these issues up, but that we should sit on it a bit to see if we every actually have demand for a gather write via the jetty-12 API.  @sbordet your thoughts on where a gather write might be done?

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
lachlan-roberts: Done, this is a private method anyway is no problem changing it.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
lachlan-roberts: Turns out we already have a `BufferUtil.toDetailString(ByteBuffer[])` method. I will use that.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
lachlan-roberts: The tests fail if this is only calculated once in reset/prepare.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConnection.java
lachlan-roberts: I was getting failures from calculating this only once in the constructor so I changed to calculate wherever used to be safe.

But it seems to pass tests if I just calculate once at the start of `process()`, so I will try doing that instead.

jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
lachlan-roberts: I think I can accomplish the same logic only clearing the buffer externally. The changes will be a bit more extensive and some extra logic will need to be done everywhere the generator is used.

gregw (commented):
Let's make this a draft until we review the need for it.

PR #7097, opened 2021-11-11
Issue #2690 - clarify websocket message ordering in documentation

### LINKED ISSUE #2690: Clarify ordering, concurrency behavior of WebSocketListener and normal HTTP responses
I have some thread-safety questions about the behavior of websocket responses and normal HTTP responses. These do not seem to be answered by the Javadoc or any other documentation I have seen.

1. Are responses guaranteed to be delivered to event listeners in the same order that they are received by the network? I assume yes.
2. If two responses arrive sequentially, but the first event listener has not returned yet, does the second response block until the first response is complete? Or do you fire the event in a second thread?
3. If you fire events in different threads, what guarantee is there that event listeners can record the order in which messages were received which (in the case of websocket responses at least) might be significant?

I'm wondering these questions myself, but I'd also like to see them documented more officially too. Thank you.

lachlan-roberts:
closes #2690

### DIFF
--- a/documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/websocket/server-websocket.adoc
+++ b/documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/websocket/server-websocket.adoc
@@ -41,4 +41,3 @@ If your application needs specific features that are not provided by the standar
 include::server-websocket-standard.adoc[]
 include::server-websocket-jetty.adoc[]
 include::server-websocket-filter.adoc[]
-include::../../websocket.adoc[]
--- a/documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/websocket/server-websocket.adoc
+++ b/documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/websocket/server-websocket.adoc
@@ -41,3 +41,4 @@ If your application needs specific features that are not provided by the standar
 include::server-websocket-standard.adoc[]
 include::server-websocket-jetty.adoc[]
 include::server-websocket-filter.adoc[]
+include::../../websocket.adoc[]
--- a/documentation/jetty-documentation/src/main/asciidoc/programming-guide/websocket.adoc
+++ b/documentation/jetty-documentation/src/main/asciidoc/programming-guide/websocket.adoc
@@ -46,7 +46,7 @@ This event is emitted when the WebSocket communication encounters a fatal error,
 Applications interested in the error event receive a `Throwable` that represent the error.
 * The _message_ event.
 The message event is emitted when a WebSocket message is received.
-Only a single thread will be delivering a message event at any one time, and the next message event will not be triggered until the previous one has exited its onMessage method.
+Only one thread at a time will be delivering a message event to the `onMessage` method; the next message event will not be delivered until the previous call to the `onMessage` method has exited.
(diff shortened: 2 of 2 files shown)


documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/websocket/server-websocket.adoc
```diff
@@ -41,3 +41,4 @@ If your application needs specific features that are not provided by the standar
 include::server-websocket-standard.adoc[]
 include::server-websocket-jetty.adoc[]
 include::server-websocket-filter.adoc[]
+include::../../websocket.adoc[]
```
lachlan-roberts: @sbordet I thought `websocket.adoc` was the common section which was supposed to be in both client and server websocket documentation.  I couldn't see it in the Server side documentation I have added this here.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/websocket.adoc
```diff
@@ -46,6 +46,8 @@ This event is emitted when the WebSocket communication encounters a fatal error,
 Applications interested in the error event receive a `Throwable` that represent the error.
 * The _message_ event.
 The message event is emitted when a WebSocket message is received.
+Only a single thread will be delivering a message event at any one time, and the next message event will not be triggered until the previous one has exited its onMessage method.
```
sbordet:
```suggestion
Only one thread at a time will be delivering a message event to the `@OnMessage` method; the next message event will not be delivered until the previous call to the `@OnMessage` method has exited.
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/websocket/server-websocket.adoc
sbordet: I don't think it's the right place, as it speaks about Jetty-specific annotations and classes, so perhaps it should be an `include` inside `server-websocket-jetty.adoc`.

I could not write a generic WebSocket architecture, since you have to refer to annotation and classes that are specific.

Also, including it in the server will cause duplicate links, so perhaps for the server documentation I would just add a paragraph that says something like "for the Jetty WebSocket architecture read xref:..." and point to the client docs (rather than including it again).

sbordet (changes requested):

documentation/jetty-documentation/src/main/asciidoc/programming-guide/websocket.adoc
lachlan-roberts: I didn't want to use that specific `@OnMessage` annotation because this that annotation is specific to `javax.websocket`. This is talking about a more general message event which could be a method from a listener class or the `@OnWebSocketMessage` annotation from Jetty API.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/websocket/server-websocket.adoc
lachlan-roberts: I have reverted this change, looks like there is already some links to it from the Server documentation. Just seems strange that the common websocket documentation is contained within the client section.

sbordet (approved):

PR #7100, opened 2021-11-11
Issue #6728 - Prepare for QUIC and HTTP/3

sbordet:
Modifications to Jetty to prepare for QUIC/HTTP3.

### DIFF
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/DatagramChannelEndPoint.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/DatagramChannelEndPoint.java
@@ -19,8 +19,10 @@ import java.net.SocketAddress;
 import java.nio.ByteBuffer;
 import java.nio.channels.DatagramChannel;
 import java.nio.channels.SelectionKey;
+import java.nio.channels.WritePendingException;
 
 import org.eclipse.jetty.util.BufferUtil;
+import org.eclipse.jetty.util.Callback;
 import org.eclipse.jetty.util.thread.Scheduler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -44,7 +46,14 @@ public class DatagramChannelEndPoint extends SelectableChannelEndPoint
         return (DatagramChannel)super.getChannel();
     }
 
-    @Override
+    /**
+     * <p>Receives data into the given buffer from the returned address.</p>
+     * <p>This method should be used to receive UDP data.</p>
+     *
+     * @param buffer the buffer to fill with data
+     * @return the peer address that sent the data
+     * @throws IOException if the receive fails
+     */
     public SocketAddress receive(ByteBuffer buffer) throws IOException
     {
         if (isInputShutdown())
@@ -64,7 +73,16 @@ public class DatagramChannelEndPoint extends SelectableChannelEndPoint
         return peer;
     }
 
-    @Override
+    /**
+     * <p>Sends to the given address the data in the given buffers.</p>
+     * <p>This methods should be used to send UDP data.</p>
+     *
+     * @param address the peer address to send data to
+     * @param buffers the buffers containing the data to send
+     * @return true if all the buffers have been consumed
+     * @throws IOException if the send fails
+     * @see #write(Callback, SocketAddress, ByteBuffer...)
+     */
     public boolean send(SocketAddress address, ByteBuffer... buffers) throws IOException
     {
         boolean flushedAll = true;
@@ -95,5 +113,20 @@ public class DatagramChannelEndPoint extends SelectableChannelEndPoint
             notIdle();
 
         return flushedAll;
+    }
+
+    /**
+     * <p>Writes to the given address the data contained in the given buffers, and invokes
+     * the given callback when either all the data has been sent, or a failure occurs.</p>
+     *
+     * @param callback the callback to notify of the success or failure of the write operation
+     * @param address the peer address to send data to
+     * @param buffers the buffers containing the data to send
+     * @throws WritePendingException if a previous write was initiated but was not yet completed
+     * @see #send(SocketAddress, ByteBuffer...)
+     */
+    public void write(Callback callback, SocketAddress address, ByteBuffer... buffers) throws WritePendingException
+    {
+        getWriteFlusher().write(callback, address, buffers);
     }
 }
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/DatagramChannelEndPoint.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/DatagramChannelEndPoint.java
@@ -1,0 +1,99 @@
+//
+// ========================================================================
+// Copyright (c) 1995-2021 Mort Bay Consulting Pty Ltd and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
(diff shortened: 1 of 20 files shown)


sbordet:
@gregw build failures seem unrelated.

The relevant changes are:

* Introduced `SelectableChannelEndPoint` as the base class for both `SocketChannelEndPoint` and `DatagramChannelEndPoint`.
* Added `EndPoint.receive()`, `EndPoint.send()` and `EndPoint.write()` to provide UDP semantic.
* Support for the above output primitives in `WriteFlusher`

The rest is code cleanups, some additions to `jetty-util` classes that will be handy later, etc.

jetty-io/src/main/java/org/eclipse/jetty/io/EndPoint.java
```diff
@@ -203,7 +203,24 @@ default void close()
...
+    /**
+     * <p>Receives data into the given buffer from the returned address.</p>
+     * <p>This method should be used to receive UDP data, in conjunction
+     * with {@link DatagramChannelEndPoint}.</p>
+     *
+     * @param buffer the buffer to fill with data
+     * @return the peer address that sent the data
+     * @throws IOException if the receive fails
+     */
+    default SocketAddress receive(ByteBuffer buffer) throws IOException
```
gregw: So why can't the `receive` and `send` methods be moved to a sub-interface of `EndPoint` and in the few places they are needed then the user downcasts to a `DatagramEndPoint` ?

I know that is clunkier, but as we are likely to change the `EndPoint` abstraction in jetty-12 is seams wrong to extend it only to then change it.

jetty-io/src/main/java/org/eclipse/jetty/io/EndPoint.java
```diff
@@ -215,7 +232,26 @@ default void close()
...
+     * <p>This methods should be used to send UDP data, in conjunction
+     * with {@link DatagramChannelEndPoint}.</p>
+     *
+     * @param address the peer address to send data to
+     * @param buffers the buffers containing the data to send
+     * @return true if all the buffers have been consumed
+     * @throws IOException if the send fails
+     * @see #write(Callback, SocketAddress, ByteBuffer...)
+     */
+    default boolean send(SocketAddress address, ByteBuffer... buffers) throws IOException
```
gregw: I don't really understand this signature.    If `send` returns false and the buffers have only been partially consumed, what has happened and what am I supposed to do?  Have a datagram with the partial content been send?   Should I then call it again but with a null address?   Will the subsequent calls send new datagrams or just complete the first? 

At the very least this needs to be called `flush` and have better javadoc, but can't we just hide it and only use the new `write`?

jetty-io/src/main/java/org/eclipse/jetty/io/ManagedSelector.java
```diff
@@ -990,36 +990,42 @@ private CloseConnections(Set<Closeable> closed)
...
+                            {
+                                _closed.add(closeable);
+                                IO.close(closeable);
+                            }
                         }
                     }
                 }
             }
-            _complete.countDown();
+            finally
```
gregw, line 1025: Is this fixing a known bug or just defensive in case something above throws?
I can't see what could throw above???

jetty-io/src/main/java/org/eclipse/jetty/io/DatagramChannelEndPoint.java
```diff
@@ -0,0 +1,99 @@
...
+
+import org.eclipse.jetty.util.BufferUtil;
+import org.eclipse.jetty.util.thread.Scheduler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * <p>An {@link EndPoint} implementation based on {@link DatagramChannel}.</p>
+ */
+public class DatagramChannelEndPoint extends SelectableChannelEndPoint
```
gregw, line 33: Why do we actually need `receive`, `send` and the new `write` on the `EndPoint` interface?
Can't we just write datagram code that depends on this specific `DatagramChannelEndPoint` class for now?
We can abstract away from this once we have experience and other `DatagramEndPoint` implementations

gregw (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/ManagedSelector.java
sbordet: `IO.close()` can throw.

joakime: `IO.close(Closeable)` consumes `IOException` now, specifically to not throw.
Are you saying some other exception is flowing out?

sbordet: Yes. And if it does, without the `finally` block, we wait forever.

jetty-util/src/main/java/org/eclipse/jetty/util/Callback.java
```diff
@@ -39,6 +39,24 @@ public InvocationType getInvocationType()
...
     };
 
+    /**
+     * <p>Completes this callback with the given {@link CompletableFuture}.</p>
+     * <p>When the CompletableFuture completes normally, this callback is succeeded;
+     * when the CompletableFuture completes exceptionally, this callback is failed.</p>
+     *
+     * @param completable the CompletableFuture that completes this callback
+     */
+    default void completeWith(CompletableFuture<?> completable)
```
gregw, line 49: Should this be `CompletableFuture<Void>`?

jetty-util/src/main/java/org/eclipse/jetty/util/Callback.java
```diff
@@ -168,6 +186,18 @@ public void completed()
         };
     }
 
+    static Callback from(InvocationType invocationType, Runnable completed)
```
gregw, line 197: javadoc

jetty-util/src/main/java/org/eclipse/jetty/util/Callback.java
```diff
@@ -285,6 +315,18 @@ public void failed(Throwable x)
 
     class Completing implements Callback
     {
+        private final InvocationType invocationType;
```
gregw, line 329: needs javadoc

jetty-util/src/main/java/org/eclipse/jetty/util/thread/Invocable.java
```diff
@@ -38,6 +38,35 @@
 
     static ThreadLocal<Boolean> __nonBlocking = new ThreadLocal<>();
 
+    public static Runnable from(InvocationType type, Runnable task)
+    {
+        class RunnableInvocable implements Runnable, Invocable
```
gregw: This might be reusable if moved one level up?

gregw (changes requested):
Just a little more javadoc and I think we are good.... other than one question...

jetty-util/src/main/java/org/eclipse/jetty/util/Callback.java
sbordet: No, because sometimes we pass in a `CompletableFuture<Stream>` as we don't care about the result, but just to complete the `Callback`, so the wildcard is necessary.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/Invocable.java
sbordet: @gregw moved one level up (renamed to `Task`) and reused in other places.

sbordet:
@gregw please re-review.

gregw (approved):

PR #7110, opened 2021-11-12
Issue #7109 - Deprecate UnixSocket JNR support.

### LINKED ISSUE #7109: Deprecate UnixSocket JNR support
**Jetty version(s)**
10

**Description**
Modules `jetty-unixsocket` and children should be deprecated now that there is support for UnixDomain socket based on JEP 380 introduced in Java 16.

sbordet:
Deprecated for removal classes and JPMS modules.

### DIFF
--- a/jetty-unixsocket/jetty-unixsocket-client/src/main/java/module-info.java
+++ b/jetty-unixsocket/jetty-unixsocket-client/src/main/java/module-info.java
@@ -11,6 +11,7 @@
 // ========================================================================
 //
 
+@Deprecated(forRemoval = true)
 module org.eclipse.jetty.unixsocket.client
 {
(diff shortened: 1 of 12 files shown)


jetty-unixsocket/jetty-unixsocket-client/src/main/java/module-info.java
```diff
@@ -11,6 +11,7 @@
 // ========================================================================
 //
 
+@Deprecated(forRemoval = true)
```
joakime, line 14: Perhaps include that we will remove it in Jetty 12?

joakime (dismissed):

gregw (dismissed):

PR #7125, opened 2021-11-16
Issue #7111 - Add support to deprecate jetty-home modules.

### LINKED ISSUE #7111: Add support to deprecate jetty-home modules
**Jetty version(s)**
10.x +

**Description**
We need a feature to allow deprecation of modules.

It should notify on add and use.
It should also allow freeform text to indicate what the suggested action on the deprecation. (such as an alternate module name)

sbordet:

### DIFF
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/Module.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/Module.java
@@ -154,6 +154,11 @@ public class Module implements Comparable<Module>
     private final List<String> _depends = new ArrayList<>();
 
     /**
+     * Text from {@code [deprecated]} section
+     */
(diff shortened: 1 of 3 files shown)


janbartel:
@sbordet If someone forgets to add `deprecated`  to the `[tags]` in the .mod file, is the module still listed for `--list-modules`, but without any indication it is deprecated?

joakime (dismissed):
I like it.

sbordet:
@janbartel I have addressed your concern. Now the list of enabled modules reports if a module is deprecated.
Best if it also has a tag, so all deprecated modules can be grouped together by `--list-modules`.

joakime (approved):
:+1:

PR #7143, opened 2021-11-22
Issue #7124 - Add default methods on LifeCycle.Listener interface

### LINKED ISSUE #7124: Add default methods on LifeCycle.Listener interface
> _Originally posted by @gregw in https://github.com/eclipse/jetty.project/pull/7081#r746153714_
>
> geeeze we need to add default methods on that listener interface!

lachlan-roberts:
closes #7124

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/component/AbstractLifeCycle.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/component/AbstractLifeCycle.java
@@ -251,7 +251,7 @@ public abstract class AbstractLifeCycle implements LifeCycle
     }
 
     /**
-     * @deprecated this class is unnecessary now that {@link LifeCycle.Listener} has default methods.
+     * @deprecated this class is redundant now that {@link LifeCycle.Listener} has default methods.
      */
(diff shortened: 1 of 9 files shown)


jetty-util/src/main/java/org/eclipse/jetty/util/component/AbstractLifeCycle.java
```diff
@@ -250,6 +250,10 @@ public void setStopTimeout(long stopTimeout)
         this._stopTimeout = stopTimeout;
     }
 
+    /**
+     * @deprecated this class is unnecessary now that {@link LifeCycle.Listener} has default methods.
```
joakime: "unnecessary"?

How about "redundant" instead?

janbartel (dismissed):

jetty-util/src/main/java/org/eclipse/jetty/util/component/AbstractLifeCycle.java
lachlan-roberts: done

janbartel (approved):

PR #7148, opened 2021-11-23
Restructuring of jetty 12, change module names & fix build

lachlan-roberts:
### Restructuring of Jetty 12

- Can now build project without errors.
- Add servlet-api dependency to the servlet6 `pom.xml` file.
- Add back dependency of servlet-api back into jetty-server until usage of it has been removed.
- Move jetty-plus under the servlet6 module.
- Modules under servlet6 no longer have artifactIds like `jetty-servlet6` instead they use a groupId of `org.eclipse.jetty.servlet6` with artifactId `jetty-servlet`.
- No other modules should depend on directly on the `jetty-servlet-common`, `jetty-webapp-common` other than the different servlet implementations.

```
jetty-project
  └── jetty-servlet-parent
         ├── servlet6 (with groupId org.eclipse.jetty.servlet6)
         │      ├── jetty-annotations
         │      ├── jetty-plus
         │      ├── jetty-security
         │      ├── jetty-servlet
         │      └── jetty-webapp
         │
         ├── jetty-servlet-common
         ├── jetty-webapp-common
         ├── jetty-security-common
         ├── jetty-annotations-common
         ├── jetty-annotations-common
         │
         └── jetty-session
                ├── jetty-gcloud-session-manager
                ├── jetty-hazelcast
                ├── jetty-infinispan
                ├── jetty-memcached
                └── jetty-nosql
```


jetty-osgi/jetty-osgi-boot/pom.xml
```diff
@@ -82,7 +82,7 @@
               jakarta.mail.internet;version="2.0";resolution:=optional,
               jakarta.mail.search;version="2.0";resolution:=optional,
               jakarta.mail.util;version="2.0";resolution:=optional,
-              jakarta.servlet;version="[$(version;==;${}),$(version;+;${jetty.servlet5.api.version}))",
+              jakarta.servlet;version="[$(version;==;${jetty.servlet5.api.version}),$(version;+;${jetty.servlet5.api.version}))",
```
janbartel: Something seems wrong here: this refers to `jetty.servlet5.api.version`, yet line 22 above refers to `org.eclipse.jetty.servlet6`.

janbartel (changes requested):

jetty-osgi/jetty-osgi-boot/pom.xml
lachlan-roberts: There is no servlet 6 version available here yet, it is still using the Servlet 5 version even in `jetty-webapp` which we are importing on line 22. The versions should be updated to servlet6 when it becomes available, but if we changed this now to `jetty.servlet6.api.version` it would fail.

janbartel: @lachlan-roberts you can locally build the servlet6 api and build against that. That's what the old jetty-12-experimental branch was doing. The servlet api is gearing up for a release of 6 soon, so it would be good to have these things in place. BTW you could just rename `jetty.servlet5.api.version` to `jetty.servlet6.api.version` couldn't you?

lachlan-roberts: Sure, but I think it would be best to do the update to servlet 6 version in a different PR to this. This is more aimed at the restructuring of the things under `jetty-servlet-parent` so that the project would build.

I just fixed this line because it was causing this module to fail, it was brought over from 11 incorrectly.

janbartel: @lachlan-roberts also, I'm not sure that having split packages is a good idea. Eg both `jetty-webapp` and `jetty-webapp-common` are `org.eclipse.jetty.webapp`.

janbartel: @lachlan-roberts I think you should be consistent. If you were just porting servlet5 to jetty12, then you should have called the modules `servlet5` not `servlet6`.  But, you've named things `servlet6`. So I suggest you also rename these properties to `servlet6` and just leave them pointing to servlet 5 jars.

lachlan-roberts: > I'm not sure that having split packages is a good idea. Eg both jetty-webapp and jetty-webapp-common are org.eclipse.jetty.webapp.

The idea is that we have `servlet6/jetty-webapp` which might be able to share some code with `servlet5/jetty-webapp` and the shared code would go in `jetty-webapp-common`. But currently there is nothing in these modules as we don't have multiple implementations yet. All the shared modules are currently empty so I could remove them in this PR, but they'll probably have to be brought back eventually. Maybe the package names in `servlet6/jetty-webapp` will need to be `org.eclipse.jetty.webapp6`.

> I think you should be consistent. If you were just porting servlet5 to jetty12, then you should have called the modules servlet5 not servlet6. But, you've named things servlet6. So I suggest you also rename these properties to servlet6 and just leave them pointing to servlet 5 jars.

I think we're going for a complete servlet 6 implementation before we try to support the multiple versions so thats why I called it servlet6 and not servlet5. I can start renaming properties and things here to servlet6, I just thought they'd be lost within the 500+ file changes and will be hard to review. I'll give it a go to see how much there is to change.

olamy:
Perso I don't like this `jetty-servlet-parent` naming. I find it very ugly. I renamed it on purpose during first changes of structure.

jetty-osgi/jetty-osgi-boot/pom.xml
lachlan-roberts: @janbartel I have updated all the properties to `jetty.servlet6.api.version`.

lachlan-roberts:
@janbartel @olamy can get a re-review?

jetty-cdi/pom.xml
```diff
@@ -19,13 +19,13 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
-      <groupId>org.eclipse.jetty.servlet</groupId>
-      <artifactId>jetty-webapp-common</artifactId>
+      <groupId>org.eclipse.jetty.servlet6</groupId>
+      <artifactId>jetty-webapp6</artifactId>
```
olamy, line 23: looks good for now. but not sure we want this module depending on specific servlet version

jetty-client/pom.xml
```diff
@@ -122,8 +122,8 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.eclipse.jetty.servlet</groupId>
-      <artifactId>jetty-security-common</artifactId>
+      <groupId>org.eclipse.jetty.servlet6</groupId>
+      <artifactId>jetty-security6</artifactId>
```
olamy, line 126: we should depend on something common and not a specific version

jetty-connectors/jetty-fcgi/fcgi-server/pom.xml
```diff
@@ -58,8 +58,8 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.eclipse.jetty.servlet</groupId>
-      <artifactId>jetty-annotations-common</artifactId>
+      <groupId>org.eclipse.jetty.servlet6</groupId>
+      <artifactId>jetty-annotations6</artifactId>
```
olamy, line 62: we should depend on something common and not a specific version

jetty-deploy/pom.xml
```diff
@@ -30,8 +30,8 @@
 
   <dependencies>
     <dependency>
-      <groupId>org.eclipse.jetty.servlet</groupId>
-      <artifactId>jetty-webapp-common</artifactId>
+      <groupId>org.eclipse.jetty.servlet6</groupId>
+      <artifactId>jetty-webapp6</artifactId>
```
olamy, line 34: we should depend on something common and not a specific version

jetty-jaas/pom.xml
```diff
@@ -39,8 +39,8 @@
 
   <dependencies>
     <dependency>
-      <groupId>org.eclipse.jetty.servlet</groupId>
-      <artifactId>jetty-security-common</artifactId>
+      <groupId>org.eclipse.jetty.servlet6</groupId>
+      <artifactId>jetty-security6</artifactId>
```
olamy, line 43: we should depend on something common and not a specific version

jetty-jaspi/pom.xml
```diff
@@ -44,8 +44,8 @@
       <artifactId>jetty-jakarta-servlet-api</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.eclipse.jetty.servlet</groupId>
-      <artifactId>jetty-security-common</artifactId>
+      <groupId>org.eclipse.jetty.servlet6</groupId>
+      <artifactId>jetty-security6</artifactId>
```
olamy, line 48: we should depend on something common and not a specific version

jetty-maven-plugin/pom.xml
```diff
@@ -160,8 +160,8 @@
       <artifactId>jetty-util</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.eclipse.jetty.servlet</groupId>
-      <artifactId>jetty-webapp-common</artifactId>
+      <groupId>org.eclipse.jetty.servlet6</groupId>
+      <artifactId>jetty-webapp6</artifactId>
```
olamy, line 164: we should depend on something common and not a specific version

olamy (commented):
I'm not sure if I like the idea of generic module such cdi jaas dependency on jettty-servlet6 or any specific version. I would prefer a dependency to something such *common*. users will choose at runtime depending on what spec version they want to use.
@lachlan-roberts Or is it temporary to get the build compiling? (hehe temporary usually stays forever :))

janbartel (changes requested):
I don't understand the motivation for having almost everything grouped under jetty-servlet, and every groupid underneath that being `org.eclipse.jetty.servlet`.  

Why can't we just retain `org.eclipse.jetty` as the main group id?

There needs to be some `xyz-common` subprojects and some `xyz6` projects. Eg there should be `jetty-webapp-common` and `jetty-webapp6`;  `jetty-session-common` and `jetty-session6`. 

The packages of these `xyz-common` and `xyz6` projects should not be the same.  An example would be `org.eclipse.jetty.webapp.common`  and `org.eclipse.jetty.webapp6`.

olamy:
> I don't understand the motivation for having almost everything grouped under jetty-servlet, and every groupid underneath that being `org.eclipse.jetty.servlet`.
> 
> Why can't we just retain `org.eclipse.jetty` as the main group id?

perso I find this will be very confusing such flat structure as we will have more artifacts than before and I find that better organised and reflecting the code structure.
but maybe it's just me. 
Jetty-12 will be a big change for users so maybe we can change artifacts structure as well to group them together.

> 
> There needs to be some `xyz-common` subprojects and some `xyz6` projects. Eg there should be `jetty-webapp-common` and `jetty-webapp6`; `jetty-session-common` and `jetty-session6`.
> 
> The packages of these `xyz-common` and `xyz6` projects should not be the same. An example would be `org.eclipse.jetty.webapp.common` and `org.eclipse.jetty.webapp6`.

gregw:
I don't think we need to come up with the absolute definitive final layout yet.   We need something that is workable for development that we can closely review before we commit any artifacts to maven central.

Personnally I don't have a strong preference either way.... but....

We currently have packages like `jetty-servlet` containing `org.eclipse.jetty:jetty-servlet`.  These are never going away and will at least contain the versions of jetty 7,8,9,10,11 for all time into the future.  Thus I think it will be less confusing if we somehow find a way to reuse `org.eclipse.jetty:jetty-servlet` so that there will be a version 12,13,14 of that.   Thus instead of creating a `org.eclipse.jetty:jetty-servlet-common` I'd kind of prefer it 
 if `org.eclipse.jetty:jetty-servlet` just became that common module.  We would then end up with:

+ `org.eclipse.jetty:jetty-servlet`
+ `org.eclipse.jetty:jetty-servlet4`
+ `org.eclipse.jetty:jetty-servlet5`
+ `org.eclipse.jetty:jetty-servlet6`

I'm happy for the disk layout of these to group those either by some parent directory.... I'm just not sure if the parent(s) should be grouping all similar modules of different versions or a parent that groups different modules of the same servlet version.

joakime:
We'll wind up with ee specific bom's as well. (for the users of embedded
jetty and testing with jetty)

   - org.eclipse.jetty.ee:api-layer-ee9:${jetty.version}
   - org.eclipse.jetty.ee:api-layer-ee10:${jetty.version}
   - org.eclipse.jetty.ee:api-layer-ee11:${jetty.version}

This can be leveraged with our build as well. (both for the module system
and testing)
The maven repo structure can also mimic this.

The groupId could easily be the delineator here as well.
eg: org.eclipse.jetty.ee9 / org.eclipse.jetty.ee10 etc ...

Joakim Erdfelt / ***@***.***

janbartel:
> We'll wind up with ee specific bom's as well. (for the users of embedded jetty and testing with jetty) - org.eclipse.jetty.ee:api-layer-ee9:${jetty.version} - org.eclipse.jetty.ee:api-layer-ee10:${jetty.version} - org.eclipse.jetty.ee:api-layer-ee11:${jetty.version} This can be leveraged with our build as well. (both for the module system and testing) The maven repo structure can also mimic this. The groupId could easily be the delineator here as well. eg: org.eclipse.jetty.ee9 / org.eclipse.jetty.ee10 etc ... Joakim Erdfelt / ***@***.***

@joakime I don't understand this. What is this org.eclipse.jetty.ee namespace you speak of? Why is this using ee version numbers instead of servlet version numbers? I think we need to be absolutely consistent - either one or the other.

lachlan-roberts:
I have changed the structure so that everything is not under `jetty-servlet`.
It now looks like this:

- jetty-project
  - jetty-servlet (groupId org.eclipse.jetty.servlet)
    - jetty-servlet6
    - jetty-servlet-common
  - jetty-security (groupId org.eclipse.jetty.security)
    - jetty-security6
    - jetty-security-common
  - jetty-webapp (groupId org.eclipse.jetty.webapp)
    - jetty-webapp6
    - jetty-webapp-common
  - jetty-annotations (groupId org.eclipse.jetty.annotations)
    - jetty-annotations6
    - jetty-annotations-common
  - jetty-plus (groupId org.eclipse.jetty.plus)
    - jetty-plus6
    - jetty-plus-common
  - jetty-session (groupId org.eclipse.jetty.session)
    - jetty-gcloud-session-manager
    - jetty-hazelcast
    - jetty-infinispan
    - jetty-memcached
    - jetty-nosql

lachlan-roberts:
@janbartel @joakime @olamy this is ready for review.

janbartel:
@lachlan-roberts getting closer. Personally I don't like the fact that the `org.eclipse.jetty.servlet:jetty-servlet` artifact is now a pom, and not the actual implementation. I think given our naming history, anything with `jetty-` prefix should remain code, and not a pom. I'd prefer the `servlet-parent` artifact name, as that is consistent with the naming approach we already have for modules that have multiple grouped artifacts. I'm not enamoured of the `servlet-common` artifact, rather than being called the `jetty-servlet` artifact, but I guess it has it's advantages, in that maybe the name is indicative that it's no longer a standalone module.

olamy:
> @lachlan-roberts getting closer. Personally I don't like the fact that the `org.eclipse.jetty.servlet:jetty-servlet` artifact is now a pom, and not the actual implementation. I think given our naming history, anything with `jetty-` prefix should remain code, and not a pom. I'd prefer the `servlet-parent` artifact name, as that is consistent with the naming approach we already have for modules that have multiple grouped artifacts. I'm not enamoured of the `servlet-common` artifact, rather than being called the `jetty-servlet` artifact, but I guess it has it's advantages, in that maybe the name is indicative that it's no longer a standalone module.

why not changing? 
Perso I find all naming such `*-parent` ugly and usually such naming is used when pom sit standalone in a separate repo or for very top level organisation or project pom.
12 will be a big change so why not changing to something more standard usage. The usage of such naming `*-parent` is really tending to disappear now (or for very few cases mentioned above).
`jetty-servlet` will have a new groupId as well.
keeping the same groupId:artifactId to something which will be totally different and doesn't do anything as before except some common code will really confuse users.

janbartel:
@olamy  I am all for the new group id space, eg `org.eclipse.jetty.servlet` instead of having everything at `org.eclipse.jetty`.  As for the `*-parent` nomenclature, what do other projects do when then have deep subproject trees? I know that I find it annoying and time consuming when I deal with a project that changes its groupId, and then I have to sift through loads of artifact names that look like a code artifact, but it turns out to be nothing more than a pom ...

Maybe  if we had:

- jetty-project (gid=org.eclipse.jetty: aid=jetty-project)
  - servlet (gid=org.eclipse.jetty.servlet:  aid=servlet)
     - jetty-servlet6   (gid=org.eclipse.jetty.servlet:  aid=jetty-servlet6)
     - jetty-servlet   (gid=org.eclipse.jetty.servlet:  aid=jetty-servlet)
  - security (gid=org.eclipse.jetty.security: aid=security)
     - jetty-security6 
     - jetty-security
  - webapp (gid=org.eclipse.jetty.webapp: aid=webapp)
    - jetty-webapp6 
    - jetty-webapp
  - annotations (gid=org.eclipse.jetty.annotations: aid=annotations)
    - jetty-annotations6
    - jetty-annotations
  - plus (gid=org.eclipse.jetty.plus: aid=plus)
    - jetty-plus6
    - jetty-plus
  - sessions (gid=org.eclipse.jetty.sessions: aid=sessions)
    - jetty-gcloud-session-manager
    - jetty-hazelcast
    - jetty-infinispan
    - jetty-memcached
    - jetty-nosql
    - jetty-sessions

olamy:
if you want good example (good is subjective I agree :P ), look at spring-boot when it was still using maven: https://github.com/spring-projects/spring-boot/tree/2.2.x (even if I don't like the source tree)

- no use of `*-parent` except the very top parent pom
- artifactId always equals to directory name (nothing such directory `jetty-infinispan` with artifactId `infinispan-parent`
- all artifactId starting with `spring-boot-*`. I'd really like to have all our artifactId starting with `jetty-` because it means when users package their projects they have everything related to Jetty starting with `jetty-` (a bit of marketing?) (I find confusing to have some jars named `infinispan-remote-10.0.1.jar` or `http2-client-10.0.1.jar` as it's hard to know where does come from)

btw I'm happy with the source code tree change we already discussed (the one here https://github.com/eclipse/jetty.project/pull/7148#issuecomment-985251890) (but the naming ;) )

janbartel:
@olamy heh heh. You don't like `*-parent` and I don't like `jetty-*` that aren't code artifacts. So that we can move ahead with this, let's give it a go the way @lachlan-roberts has it in his last commit. We can get some experience with it and make any changes we want/need to before anything is released to maven central.

janbartel (approved):
Let's give it a go like this and get some experience with it before it hits maven central.

olamy:
> @olamy heh heh. You don't like `*-parent` and I don't like `jetty-*` that aren't code artifacts. So that we can move ahead with this, let's give it a go the way @lachlan-roberts has it in his last commit. We can get some experience with it and make any changes we want/need to before anything is released to maven central.

it's not only I don't like it. It hurts my eyes when wearing my maven committer hat... But it's just convention and common usage.. 
And we are definitely totally free to use what we prefer and having our own conventions. (I will not mention we already have strange usage of `_` for class field naming 😜)

but what I find totally not conventional is having directory name different from artifactId.
we have such tree : `jetty-websocket` (as directory name on disk) but artifactId `websocket-parent`. so the top pom says `<module>jetty-websocket</module>` (I just find this totally confusing and not following maven convention neither common usage)
but yeah again it's just convention/common usage and we are totally free to have our conventions.

olamy (approved):
yup approving this as we need to move forward. And with this PR we have `jetty-servlet` `jetty-session` etc as a pom  ;)

Issue #7168, opened 2021-11-26
Jetty rewrites header values breaking AWS signatures


cstamas:
**Jetty version(s)**
Tested 9.4.41 (version s3proxy uses) but locally same behaviour with 9.4.44

**Java version/vendor** `(use: java -version)`
openjdk version "11.0.13" 2021-10-19
OpenJDK Runtime Environment Temurin-11.0.13+8 (build 11.0.13+8)
OpenJDK 64-Bit Server VM Temurin-11.0.13+8 (build 11.0.13+8, mixed mode)

**OS type/version**
Linux Mint 20.2 kernel 5.4.0-90-generic

**Description**
Seems Jetty is replacing value of Content-Type from client HTTP request, thus making header (sent by client) as
```
text/plain; charset=utf-8
```
into
```
text/plain; charset=UTF-8
```
While this does not seem biggie, it breaks AWS v2 and v4 signatures, hence requests using content-type that are added to HttpParser CACHE (well, tested for text/plain only, other s3proxy UTs uses not-so-common content types like audio etc not present in Jetty HttpParser CACHE) are doomed to fail due this: SignatureDoesNotMatch error from s3proxy. 
Reason is simple: AWS client will create canonical headers containing "utf-8" while s3proxy does same but will use "UTF-8" and hence resulting signature will be different.

**How to reproduce?**
Jetty UT https://github.com/eclipse/jetty.project/pull/7169
S3Proxy UT https://github.com/gaul/s3proxy/pull/394

Links
S3Proxy issue https://github.com/gaul/s3proxy/issues/392
UT for S3Proxy reproducing issue https://github.com/gaul/s3proxy/pull/394

cstamas:
Created Jetty UT that demonstrates the issue: https://github.com/eclipse/jetty.project/pull/7169

joakime:
Considering that the https://datatracker.ietf.org/doc/draft-ietf-httpbis-message-signatures/ is still a a draft and every revision on it introduces more and more "but this field is also case insensitive", I bet this is yet another situation just like this. (All mime-type and charset usage in any http field is case-insensitive already per the HTTP/1.x specs)

The translation of the HTTP Fields from HTTP/1.1 to HTTP/2 (done by many intermediaries) would also make signature validation nigh impossible.
Plenty of discussion on these issues on http-bis mailing list.

You could probably disable the HttpParser CACHE, but that would just skyrocket your GC.

cstamas:
[...]

PR #7169, opened 2021-11-26
`HttpParserTest.testHeaderContentTypeValue` for issue #7168

cstamas:
Demonstrates how content-type header is rewritten
by HttpParser.

This UT locally fails for me with this
```
org.opentest4j.AssertionFailedError: 
Expected :text/plain;charset=utf-8
Actual   :text/plain;charset=UTF-8
```

### DIFF
--- a/jetty-http/src/test/java/org/eclipse/jetty/http/HttpParserTest.java
+++ b/jetty-http/src/test/java/org/eclipse/jetty/http/HttpParserTest.java
@@ -891,6 +891,28 @@ public class HttpParserTest
     }
 
     @Test
+    public void testHeaderContentTypeValue()
+    {
+        ByteBuffer buffer = BufferUtil.toBuffer(
+            "GET / HTTP/1.1\r\n" +
+                "Host: localhost\r\n" +
+                "Content-Type: text/plain;charset=utf-8\r\n" +
+                "\n\n");
+
+        HttpParser.RequestHandler handler = new Handler();
+        HttpParser parser = new HttpParser(handler);
+        parseAll(parser, buffer);
+
+        assertEquals("GET", _methodOrVersion);
+        assertEquals("/", _uriOrStatus);
+        assertEquals("HTTP/1.1", _versionOrReason);
+        assertEquals("Host", _hdr[0]);
+        assertEquals("localhost", _val[0]);
+        assertEquals("Content-Type", _hdr[1]);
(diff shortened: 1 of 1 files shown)


joakime:
This is an HttpParser cache, not a rewritten header.
Not sure how you would validate AWS Signatures (or https://datatracker.ietf.org/doc/draft-ietf-httpbis-message-signatures/) using the Servlet API, as to validate you need to know the canonical headers, their order, and their breakdown.

Servlet spec does not retain breakdown, nor does it retain order.

Eg:

```
Host: local
X-Foo: abc
Connection: close
X-Foo: xyz
```

This would result in the Servlet spec reporting 3 headers.

```
Host = "local"
X-Foo = ["abc", "xyz"]
Connection = "close"
```

The order of those 3 headers are not guaranteed to be returned in the order "Host", "X-Foo", then "Connection") by the servlet spec.

You don't have access to the raw headers you would need to perform the AWS Signature handling properly.
To satisfy the Servlet spec Jetty removes whitespace, removes quotes (where needed), caches common headers, merges header fields, etc.

How would you use the Servlet spec to validate signature headers?

Also, I would argue that Section 1.3 of https://datatracker.ietf.org/doc/draft-ietf-httpbis-message-signatures/ would make the field values on specific header fields always lowercase as well. (`Accept`, `Accept-Charset`, `Accept-Encoding`, `Accept-Language`, `Content-Type`, `Content-Encoding`, `Content-Language`, `Content-MD5`, and `Connection` to just name a few that are case insensitive already)

cstamas:
Order of signing is passed in an AWS header (this is AWS S3 client doing), and verifying singature happens in S3Proxy that is a Jetty handler (see https://github.com/gaul/s3proxy/issues/392). So, only if values would correspond to those that client sent, all would be okay.... (again, just look at s3proxy codebase and UT in issue that fails)

cstamas:
Here is signature code of s3proxy: https://github.com/gaul/s3proxy/blob/master/src/main/java/org/gaul/s3proxy/AwsSignature.java

janbartel:
@cstamas see https://github.com/gaul/s3proxy/issues/392#issuecomment-985268414. I think this PR is not needed if https://github.com/gaul/s3proxy/issues/392 follows the advice I've given.

joakime:
@cstamas @janbartel now that https://github.com/gaul/s3proxy/issues/392 has been merged, can we close this PR?

cstamas:
Closing as requested.

PR #7180, opened 2021-11-30
Issue #7160 - Add AMBIGUOUS_PATH_ENCODING to default UriCompliance mode.

### LINKED ISSUE #7160: HttpURI considers %25 to be ambiguous, preventing access to static resources with % in their name
**Jetty version(s)**

11.0.7

openjdk version "17" 2021-09-14
OpenJDK Runtime Environment Temurin-17+35 (build 17+35)
OpenJDK 64-Bit Server VM Temurin-17+35 (build 17+35, mixed mode, sharing

**OS type/version**

macOS

**Description**

This is a follow on from the [discussion in #4033](https://github.com/eclipse/jetty.project/issues/4033#issuecomment-976482266). With Jetty 11.0.7, a static resource with a `%` in its filename cannot be accessed due to `HttpURI` considering it to be ambiguous:

https://github.com/eclipse/jetty.project/blob/0c6acfc1a2f7889e577c9a565aff2f4fa32d855c/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java#L1210-L1212

I have a (rather extreme) test that tries to load static resources from a jar. This jar contains an entry named `META-INF/resources/nested-reserved-!#\$%&()*+,:=?@[]-meta-inf-resource.txt` and the test tries to access it using the URL `/nested-reserved-%21%23%24%25%26%28%29%2A%2B%2C%3A%3D%3F%40%5B%5D-meta-inf-resource.txt`. This works with Jetty 9.4.x but fails with 11.0.7. I haven't tried 10.0.x. FWIW, it also works with Tomcat 9 and 10 and Undertow 2.2

**How to reproduce?**

Hopefully the above is sufficient. I can try to extract the test from Spring Boot's build into something standalone if that would be useful.

lachlan-roberts:
## Closes #7160

Remove `AMBIGUOUS_PATH_ENCODING` from the default `UriCompliance` mode.

This means that URIs containing the encoded percent character (`%25`) will be allowed in the default mode.

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
@@ -104,7 +104,9 @@ public final class UriCompliance implements ComplianceViolation.Mode
      * additional violations to avoid most ambiguous URIs.
      * This mode does allow {@link Violation#AMBIGUOUS_PATH_SEPARATOR}, but disallows all out {@link Violation}s.
      */
-    public static final UriCompliance DEFAULT = new UriCompliance("DEFAULT", of(Violation.AMBIGUOUS_PATH_SEPARATOR));
+    public static final UriCompliance DEFAULT = new UriCompliance("DEFAULT",
+        of(Violation.AMBIGUOUS_PATH_SEPARATOR,
+            Violation.AMBIGUOUS_PATH_ENCODING));
 
     /**
      * LEGACY compliance mode that models Jetty-9.4 behavior by allowing {@link Violation#AMBIGUOUS_PATH_SEGMENT},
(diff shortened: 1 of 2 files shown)


sbordet (approved):

gregw (approved):

jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
```diff
@@ -104,7 +104,9 @@ public String getDescription()
      * additional violations to avoid most ambiguous URIs.
      * This mode does allow {@link Violation#AMBIGUOUS_PATH_SEPARATOR}, but disallows all out {@link Violation}s.
```
joakime, line 105: Whoops we missed this javadoc change.

PR #7184, opened 2021-12-01

### LINKED ISSUE #7182: jetty.sh start process should remove jetty_state whenever deleting the pid
Jetty version(s)
9.4.x
10.x
11.x

In the stop process right after the JETTY_PID file is deleted it will delete the JETTY_STATE file, but this does not happen in the running() routine when it will only delete the PID and not JETTY_STATE if there is an orphaned pid file.

Why this matters: You can end up starting jetty with an outdated JETTY_STATE file. If the contents in JETTY_STATE say "STARTED" it will incorrectly report the JETTY has started before it has. But perhaps worse, if the contents in JETTY_STATE are "STOPPED" or "FAILED" it will report jetty as having failed even if the start is successful.

This is particularly a problem when relying on a service to issue the start command because if jetty.sh reports the start as having failed then it will kill the pid it just started, even though it started fine.

Currently the only way to resolve this situation is to either start then stop with the jetty.sh script to get rid of the jetty_state file or to manually delete the jetty.state file. You cannot fix this relying on the service calling jetty.sh.

My suggestion is a small tweak to jetty.sh. In the case for start it already checks if there's a current PID running using the running() routine. If I'm reading it correctly, when there is a PID file but the PID isn't a running process it will delete the PID file. After this rm statement it should have an additional rm statement to delete the JETTY_STATE file.

With this change in place it should avoid the situation I described. Note that I am not getting into any underlying reasons for why one would be in the situation to trigger this state, but I think it makes sense that if you're deleting the pid file to also delete the state file, to keep things in sync.

Edit: I labeled this an enhancement as I didn't think it was worthy of being a bug, but reflecting on it I think this is probably more like a low priority bug.

ianrifkin:
Signed-off-by: Ian Rifkin <ianrifkin@ianrifkin.com>

### DIFF
--- a/jetty-home/src/main/resources/bin/jetty.sh
+++ b/jetty-home/src/main/resources/bin/jetty.sh
@@ -127,7 +127,6 @@ running()
     return
   fi
   rm -f "$1"
-  rm -f "$2"  
   return 1
 }
 
@@ -137,9 +136,9 @@ started()
   for ((T = 0; T < $(($3 / 4)); T++))
   do
     sleep 4
-    [ -z "$(grep STARTED $1 2>/dev/null)" ] || return 0
-    [ -z "$(grep STOPPED $1 2>/dev/null)" ] || return 1
-    [ -z "$(grep FAILED $1 2>/dev/null)" ] || return 1
+    [ -z "$(tail -1 $1 | grep STARTED 2>/dev/null)" ] || return 0
+    [ -z "$(tail -1 $1 | grep STOPPED 2>/dev/null)" ] || return 1
+    [ -z "$(tail -1 $1 | grep FAILED 2>/dev/null)" ] || return 1
     local PID=$(cat "$2" 2>/dev/null) || return 1
     kill -0 "$PID" 2>/dev/null || return 1
     echo -n ". "
@@ -476,7 +475,7 @@ case "$ACTION" in
 
     else
 
-      if running $JETTY_PID $JETTY_STATE
+      if running $JETTY_PID
       then
         echo "Already Running $(cat $JETTY_PID)!"
         exit 1
@@ -527,7 +526,7 @@ case "$ACTION" in
       start-stop-daemon -K -p"$JETTY_PID" -d"$JETTY_HOME" -a "$JAVA" -s HUP
 
       TIMEOUT=30
-      while running "$JETTY_PID" "$JETTY_STATE"; do
+      while running "$JETTY_PID"; do
         if (( TIMEOUT-- == 0 )); then
           start-stop-daemon -K -p"$JETTY_PID" -d"$JETTY_HOME" -a "$JAVA" -s KILL
         fi
@@ -548,7 +547,7 @@ case "$ACTION" in
       kill "$PID" 2>/dev/null
 
       TIMEOUT=30
-      while running $JETTY_PID $JETTY_STATE; do
+      while running $JETTY_PID; do
         if (( TIMEOUT-- == 0 )); then
           kill -KILL "$PID" 2>/dev/null
         fi
@@ -591,7 +590,7 @@ case "$ACTION" in
   run|demo)
     echo "Running Jetty: "
 
-    if running "$JETTY_PID" "$JETTY_STATE"
+    if running "$JETTY_PID"
     then
       echo Already Running $(cat "$JETTY_PID")!
       exit 1
@@ -601,7 +600,7 @@ case "$ACTION" in
     ;;
 
   check|status)
-    if running "$JETTY_PID" "$JETTY_STATE"
+    if running "$JETTY_PID"
     then
       echo "Jetty running pid=$(< "$JETTY_PID")"
     else
@@ -611,7 +610,7 @@ case "$ACTION" in
     dumpEnv
     echo
 
-    if running "$JETTY_PID" "$JETTY_STATE"
+    if running "$JETTY_PID"
     then
       exit 0
     fi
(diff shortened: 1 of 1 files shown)


ianrifkin:
I noted this in the issue but another approach we could do is just change the grep lines to search only the last line in the state file, which avoids the issue of if the state file wasn't deleted and might be a better approach. Thoughts?

jetty-home/src/main/resources/bin/jetty.sh
```diff
@@ -127,6 +127,7 @@ running()
     return
   fi
   rm -f "$1"
+  rm -f "$2"
```
joakime: Odd, not sure what you are trying to do here.

joakime:
The `JETTY_STATE` file is an appended file, not just 1 line.
I'm not convinced this is the best approach to handle this.
Perhaps we should just look at duplicating the behaviors we see in `restart` and `stop` with how we handle `JETTY_STATE`.
Take those behaviors and apply it to `start` as well.

jetty-home/src/main/resources/bin/jetty.sh
ianrifkin: Okay. The concept -- and maybe this is not the best approach -- is that if you're going to delete the JETTY_PID then maybe you should also delete the JETTY_STATE. If the PID is deleted and the state file is present then it will cause the JETTY_STATE file to have previous lines still present (and since the grep is looking at the whole file and not the most recent line) it will think it failed.

ianrifkin:
> The `JETTY_STATE` file is an appended file, not just 1 line.

Correct. That's basically the problem. Let me try to clarify.

It's fine that it's an appended file but in the cases where it's not properly deleted it should either be deleted and/or the grep changed to just look at the most recent line in the file.

To give a stupid example (but easy to replicate)…

Let's say Jetty is running. Under normal circumstances the state file will have one line that says starting and a second line that says started. Then let's say that for whatever reason the jetty.pid file has the wrong pid or even is just accidentally deleted. In this circumstance then stop jetty via a service (which still knows the pid). In this situation the running process is stopped but jetty.sh will not delete the jetty_state file. Then when you start jetty it will currently not delete the state file and will append to it the starting line…

In that situation the state file will have the lines
STOPPING
STOPPED
STARTING

That's all valid but when jetty uses the started() function it will grep the whole file and see that it's STOPPED even though most recently it's starting (and if left alone will start) but jetty.sh will return the error that it's stopped (when it's actually starting).

The PR here was suggesting to delete the jetty_state file more consistently but I think a simpler approach is just changing the grep's to just look at the last line in the appended file. Thoughts?

ianrifkin:
Let me know if you want me to update the PR. Basically I'm now suggesting reverting these proposed changes of consistently deleting the JETTY_STATE and to change the grep to look at the most recent line instead of searching the whole file (b/c jetty.sh should only care if most recently it is started, stopped, or failed.

current:
`  [ -z "$(grep STARTED $1 2>/dev/null)" ] || return 0
    [ -z "$(grep STOPPED $1 2>/dev/null)" ] || return 1
    [ -z "$(grep FAILED $1 2>/dev/null)" ] || return 1`

proposed:
`  [ -z "$(tail -1 $1 | grep STARTED 2>/dev/null)" ] || return 0
    [ -z "$(tail -1 $1 | grep STOPPED 2>/dev/null)" ] || return 1
    [ -z "$(tail -1 $1 | grep FAILED 2>/dev/null)" ] || return 1`

Thoughts?

ianrifkin:
I'll update the PR with my suggestion and you can tell me if you think it's good or not whenever you get a chance. If it helps, the actual problem I saw in real life is then I tried to stop jetty while I knew it was doing something and for some reason it rm's the state file after it is "stopping" but before it is "stopped" -- this underlying issue I've experienced may not be a jetty problem per se, but the end result is after it has stopped I'm left with a state file that says:

`STOPPED Server@2b48a640{STOPPED}[9.4.40.v20210413]
`
and because the grep lines check anywhere in the file I cannot use the service command to start jetty unless I manually delete the state file.

janbartel:
Seems reasonable to only check the last line in the file, considering that it is appended. @joakime what's the advantage of appending, rather than creating/deleting the `.state` file in tandem with the `.pid` file?

joakime:
> Seems reasonable to only check the last line in the file, considering that it is appended. @joakime what's the advantage of appending, rather than creating/deleting the `.state` file in tandem with the `.pid` file?

The `PID` needs to stick around if it's running, that's a given.

The `.state` file has details on how it transitioned, and that's an important troubleshooting detail that has helped in the past.
That transition has revealed issues in many interesting ways (both in what it says and doesn't say)

We also don't want to delete the state file unless you call `stop` (or associated things like `restart`).

The only nuance that we need to cover is when `start` is specifically is used, and the `.pid` exists and is invalid (not a running process, or not java, or not jetty).
We need to delete that invalid `.pid` and associated invalid `.state` file as in this situation (and this is what I had hoped this PR would be limited to)

If `start` fails, the `.state` and `.pid` file have a useful troubleshooting purpose, as several of the reasons for failure can be before jetty logging has started (or if you have a badly configured start logging or jetty logging)

ianrifkin:
> The `.state` file has details on how it transitioned, and that's an important troubleshooting detail

@joakime Hi again. That makes sense to me which is why the current version of this PR changes nothing with deleting the state file -- with this PR you can even increase the history in the state file, if you wanted to. I am fine with whatever you decide, but changing the greps to look at the last line in the state file like this sounds like a better approach IMHO. What would be the use case for wanting to search every line in the state file to determine the most recent state?

joakime (approved):

PR #7186, opened 2021-12-01

### LINKED ISSUE #7182: jetty.sh start process should remove jetty_state whenever deleting the pid

ianrifkin:
Signed-off-by: Ian Rifkin <ianrifkin@ianrifkin.com>

### DIFF
--- a/jetty-home/src/main/resources/bin/jetty.sh
+++ b/jetty-home/src/main/resources/bin/jetty.sh
@@ -126,6 +126,7 @@ running()
     return
   fi
   rm -f "$1"
+  rm -f "$2"  
   return 1
 }
 
@@ -466,7 +467,7 @@ case "$ACTION" in
 
     else
 
-      if running $JETTY_PID
+      if running $JETTY_PID $JETTY_STATE
       then
         echo "Already Running $(cat $JETTY_PID)!"
         exit 1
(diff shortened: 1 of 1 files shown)


ianrifkin:
closing since I learned I only need to have one PR for this

PR #7189, opened 2021-12-01
interpolate bouncycastle.version in mod file

olamy:

### DIFF
--- a/jetty-keystore/pom.xml
+++ b/jetty-keystore/pom.xml
@@ -13,7 +13,7 @@
 
   <properties>
     <bundle-symbolic-name>${project.groupId}.keystore</bundle-symbolic-name>
-    <bouncycastle.version>1.62</bouncycastle.version>
+    <bouncycastle.version>1.70</bouncycastle.version>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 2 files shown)


sbordet (dismissed):

jetty-keystore/pom.xml
```diff
@@ -13,7 +13,7 @@
 
   <properties>
     <bundle-symbolic-name>${project.groupId}.keystore</bundle-symbolic-name>
-    <bouncycastle.version>1.69</bouncycastle.version>
+    <bouncycastle.version>1.62</bouncycastle.version>
```
sbordet: Don't we have all the properties in the main POM?

olamy: maybe we can move this one up. I'm not sure as it's used only here.
no real strong opinions. (for consistency main pom would be right).
@joakime ?

joakime: No strong opinions here either.
Do what you feel makes the most sense.

lachlan-roberts: @olamy why have you downgraded the version from `1.69` to `1.62`.

olamy: because it doesn't work with 1.69 in the mod file which is hardcoded with 1.62 :) . By chance we have distribution test. before the dynamic stuff.
I will work on the upgrade in another PR.
Or maybe better with this? but I don't like mixing stuff.

olamy: upgrade done as well

sbordet (approved):

PR #7227, opened 2021-12-07
Issue #7063 - Remove logging requirement from Password / Credential

### LINKED ISSUE #7063: Simplify command line use of org.eclipse.jetty.util.Password
**Jetty version(s)**
10+

All

**OS type/version**
All

**Description**
The documentation gives example on using `org.eclipse.jetty.util.Password` like ...

```
 java -cp lib/jetty-util-{VERSION}.jar org.eclipse.jetty.util.security.Password

Usage - java org.eclipse.jetty.util.security.Password [<user>] <password>
If the password is ?, the user will be prompted for the password
```

and

```
$ java -cp ../lib/jetty-util-{VERSION}.jar org.eclipse.jetty.util.security.Password username password
2017-12-13 11:19:27.928:INFO::main: Logging initialized @95ms to org.eclipse.jetty.util.log.StdErrLog
password
OBF:1v2j1uum1xtv1zej1zer1xtn1uvk1v1v
MD5:5f4dcc3b5aa765d61d8327deb882cf99
CRYPT:usjRS48E8ZADM
```

But the `org.eclipse.jetty.util.security.Password` class has logging, so this makes the command line a bit more complex ...

Example:

``` shell
[jetty-home-10.0.7]$ java -cp lib/jetty-util-10.0.7.jar org.eclipse.jetty.util.security.Password HelloWorld
Exception in thread "main" java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory
	at org.eclipse.jetty.util.security.Credential.<clinit>(Credential.java:43)
Caused by: java.lang.ClassNotFoundException: org.slf4j.LoggerFactory
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	... 1 more
```

I see a few options here.

1. The simplest approach would be to just remove the logging from Password and Credential, but that hurts normal usage of those classes.

2. Document how to add slf4j-api and jetty-slf4j-impl to the command line.

3. Introduce a new PasswordCli class that doesn't use Credential or logging at all.

4. Introduce a `start.jar` option that just runs `org.eclipse.jetty.util.security.Password` with the appropriate classloader setup. (`--password` ?)

joakime:
Removed / Replaced Logging requirements in Password / Credential to allow for simplified use of `Password` on command line.

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
@@ -284,7 +284,7 @@ public abstract class Credential implements Serializable
                 // Intentionally not using TypeUtil.toString() here as TypeUtil
                 // introduces a Logging requirement that is not compatible
                 // with Password command line.
-                final char[] HEX = "0123456789ABCDEF".toCharArray();
+                final char[] HEX = "0123456789abcdef".toCharArray();
                 int len = digest.length;
                 char[] out = new char[len * 2];
                 for (int i = 0; i < len; i++)
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
@@ -19,7 +19,6 @@ import java.security.MessageDigest;
 import java.util.List;
 import java.util.ServiceLoader;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import org.eclipse.jetty.util.TypeUtil;
 import org.eclipse.jetty.util.thread.AutoLock;
@@ -42,7 +41,7 @@ public abstract class Credential implements Serializable
     // Intentionally NOT using TypeUtil.serviceProviderStream
     // as that introduces a Logger requirement that command line Password cannot use.
     private static final List<CredentialProvider> CREDENTIAL_PROVIDERS = ServiceLoader.load(CredentialProvider.class).stream()
-        .flatMap((p) -> Stream.of(p.get()))
+        .map(ServiceLoader.Provider::get)
         .collect(Collectors.toList());
 
     /**
@@ -265,7 +264,7 @@ public abstract class Credential implements Serializable
                         catch (Exception e)
                         {
                             System.err.println("Unable to access MD5 message digest");
-                            e.printStackTrace(System.err);
+                            e.printStackTrace();
                             return null;
                         }
                     }
@@ -280,7 +279,7 @@ public abstract class Credential implements Serializable
             catch (Exception e)
             {
                 System.err.println("Message Digest Failure");
-                e.printStackTrace(System.err);
+                e.printStackTrace();
                 return null;
             }
         }
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
@@ -187,7 +187,14 @@ public abstract class Credential implements Serializable
         MD5(String digest)
         {
             digest = digest.startsWith(__TYPE) ? digest.substring(__TYPE.length()) : digest;
-            _digest = TypeUtil.parseBytes(digest, 16);
+            // intentionally not using TypeUtil.parseBytes as using TypeUtil introduces
+            // a logging requirement that is not supported by Command Line Password util
+            byte[] digestBytes = new byte[digest.length() / 2];
+            for (int i = 0; i < digest.length(); i += 2)
+            {
+                digestBytes[i / 2] = (byte)TypeUtil.parseInt(digest, i, 2, 16);
+            }
+            _digest = digestBytes;
         }
 
         public byte[] getDigest()
@@ -274,7 +281,18 @@ public abstract class Credential implements Serializable
                     digest = __md.digest();
                 }
 
-                return __TYPE + TypeUtil.toString(digest, 16);
+                // Intentionally not using TypeUtil.toString() here as TypeUtil
+                // introduces a Logging requirement that is not compatible
+                // with Password command line.
+                final char[] HEX = "0123456789ABCDEF".toCharArray();
+                int len = digest.length;
+                char[] out = new char[len * 2];
+                for (int i = 0; i < len; i++)
+                {
+                    out[i * 2] = HEX[(digest[i] & 0xF0) >> 4];
+                    out[(i * 2) + 1] = HEX[(digest[i] & 0x0F)];
+                }
+                return __TYPE + String.valueOf(out);
             }
             catch (Exception e)
             {
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
@@ -23,8 +23,6 @@ import java.util.stream.Stream;
 
 import org.eclipse.jetty.util.TypeUtil;
 import org.eclipse.jetty.util.thread.AutoLock;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * Credentials. The Credential class represents an abstract mechanism for checking authentication credentials. A credential instance either represents a secret,
@@ -39,10 +37,12 @@ import org.slf4j.LoggerFactory;
  */
 public abstract class Credential implements Serializable
 {
+    // NOTE: DO NOT INTRODUCE LOGGING TO THIS CLASS
     private static final long serialVersionUID = -7760551052768181572L;
-    private static final Logger LOG = LoggerFactory.getLogger(Credential.class);
-    private static final List<CredentialProvider> CREDENTIAL_PROVIDERS = TypeUtil.serviceProviderStream(ServiceLoader.load(CredentialProvider.class))
-        .flatMap(p -> Stream.of(p.get()))
+    // Intentionally NOT using TypeUtil.serviceProviderStream
+    // as that introduces a Logger requirement that command line Password cannot use.
+    private static final List<CredentialProvider> CREDENTIAL_PROVIDERS = ServiceLoader.load(CredentialProvider.class).stream()
+        .flatMap((p) -> Stream.of(p.get()))
         .collect(Collectors.toList());
 
     /**
@@ -154,8 +154,7 @@ public abstract class Credential implements Serializable
         {
             if (credentials instanceof char[])
                 credentials = new String((char[])credentials);
-            if (!(credentials instanceof String) && !(credentials instanceof Password))
-                LOG.warn("Can't check {} against CRYPT", credentials.getClass());
+
             return stringEquals(_cooked, UnixCrypt.crypt(credentials.toString(), _cooked));
         }
 
@@ -229,13 +228,12 @@ public abstract class Credential implements Serializable
                 }
                 else
                 {
-                    LOG.warn("Can't check {} against MD5", credentials.getClass());
+                    // Not a MD5 or Credential class
                     return false;
                 }
             }
             catch (Exception e)
             {
-                LOG.warn("Failed message digest", e);
                 return false;
             }
         }
@@ -248,6 +246,9 @@ public abstract class Credential implements Serializable
             return false;
         }
 
+        /**
+         * Used only by Command Line Password utility
+         */
         public static String digest(String password)
         {
             try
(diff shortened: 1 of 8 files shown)


jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
```diff
@@ -39,10 +37,12 @@
...
 {
+    // NOTE: DO NOT INTRODUCE LOGGING TO THIS CLASS
     private static final long serialVersionUID = -7760551052768181572L;
-    private static final Logger LOG = LoggerFactory.getLogger(Credential.class);
-    private static final List<CredentialProvider> CREDENTIAL_PROVIDERS = TypeUtil.serviceProviderStream(ServiceLoader.load(CredentialProvider.class))
-        .flatMap(p -> Stream.of(p.get()))
+    // Intentionally NOT using TypeUtil.serviceProviderStream
+    // as that introduces a Logger requirement that command line Password cannot use.
+    private static final List<CredentialProvider> CREDENTIAL_PROVIDERS = ServiceLoader.load(CredentialProvider.class).stream()
+        .flatMap((p) -> Stream.of(p.get()))
```
sbordet:
```suggestion
        .map(ServiceLoader.Provider::get)
```

No need to wrap with `Stream.of()` and then `flatMap()`.

jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
```diff
@@ -263,7 +264,8 @@ public static String digest(String password)
                         }
                         catch (Exception e)
                         {
-                            LOG.warn("Unable to access MD5 message digest", e);
+                            System.err.println("Unable to access MD5 message digest");
+                            e.printStackTrace(System.err);
```
sbordet: No need to pass `System.err`, it's already the default.

jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
```diff
@@ -277,7 +279,8 @@ public static String digest(String password)
             }
             catch (Exception e)
             {
-                LOG.warn("Message Digest failure", e);
+                System.err.println("Message Digest Failure");
+                e.printStackTrace(System.err);
```
sbordet: Ditto.

jetty-util/src/main/java/org/eclipse/jetty/util/security/Password.java
```diff
@@ -224,7 +220,9 @@ public static Password getPassword(String realm, String dft, String promptDft)
             }
             catch (IOException e)
             {
-                LOG.warn("EXCEPTION", e);
+                // only seen with command line input style
+                System.err.println("ERROR: Bad/Invalid password.");
+                e.printStackTrace(System.err);
```
sbordet: Ditto.

sbordet (changes requested):

joakime:
@sbordet updates pushed.

olamy (dismissed):

sbordet:
@joakime is it possible to write a test case with a `ClassLoader` with only `jetty-util.jar` in its class-path and then invoke `Password.main()` to confirm that what we want actually works and we don't break it in future?

joakime:
> @joakime is it possible to write a test case with a `ClassLoader` with only `jetty-util.jar` in its class-path and then invoke `Password.main()` to confirm that what we want actually works and we don't break it in future?

The safest way is that we'll have to fork a new JVM.

sbordet:
> The safest way is that we'll have to fork a new JVM.

Yes, now with the `ProcessBuilder` APIs that should be quite simple.
I think such a test case should be added.

jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
```diff
@@ -187,7 +187,14 @@ public static String crypt(String user, String pw)
         MD5(String digest)
         {
             digest = digest.startsWith(__TYPE) ? digest.substring(__TYPE.length()) : digest;
-            _digest = TypeUtil.parseBytes(digest, 16);
+            // intentionally not using TypeUtil.parseBytes as using TypeUtil introduces
+            // a logging requirement that is not supported by Command Line Password util
+            byte[] digestBytes = new byte[digest.length() / 2];
+            for (int i = 0; i < digest.length(); i += 2)
+            {
+                digestBytes[i / 2] = (byte)TypeUtil.parseInt(digest, i, 2, 16);
```
sbordet: Seems strange to have a comment saying that we can't use `TypeUtil`, but then we use it.
`Integer.parseInt(digest.substring(i, i+2), 16)` does the same, so I'd get rid of `TypeUtil` usage here.
I would also drop the comment, as referencing a class that is not used makes the comment weird imho.

jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
```diff
@@ -274,7 +281,18 @@ public static String digest(String password)
                     digest = __md.digest();
                 }
 
-                return __TYPE + TypeUtil.toString(digest, 16);
+                // Intentionally not using TypeUtil.toString() here as TypeUtil
+                // introduces a Logging requirement that is not compatible
+                // with Password command line.
```
sbordet: I would drop the comment, no need to reference a class that is not used.

jetty-util/src/test/java/org/eclipse/jetty/util/security/PasswordTest.java
```diff
@@ -44,4 +56,35 @@ public void testObfuscateUnicode()
...
+            .directory(MavenTestingUtils.getTargetDir())
+            .command("java",
+                "-cp", MavenTestingUtils.getTargetPath("classes").toString(),
+                Password.class.getName(),
+                "user", "password")
+            .redirectErrorStream(true);
+
+        Process passwordProcess = passwordBuilder.start();
+        try (InputStreamReader inputStreamReader = new InputStreamReader(passwordProcess.getInputStream());
+             BufferedReader reader = new BufferedReader(inputStreamReader))
```
sbordet: Inline `inputStreamReader`, it does not need to be listed as a resource, since it's not used within the block.

jetty-util/src/test/java/org/eclipse/jetty/util/security/PasswordTest.java
```diff
@@ -44,4 +56,35 @@ public void testObfuscateUnicode()
...
+        {
+            String output = reader.lines().collect(Collectors.joining(System.lineSeparator()));
+            if (passwordProcess.waitFor(5, TimeUnit.SECONDS))
+            {
+                int exitCode = passwordProcess.exitValue();
+                assertThat("Non-error exit code: " + output, exitCode, is(0));
+                assertThat("Output", output, not(containsString("Exception")));
```
sbordet, line 80: I would add an assert for minimal expected output (e.g. the original password is printed back), something like `assertThat(output, containsString("password"));`.

jetty-util/src/test/java/org/eclipse/jetty/util/security/PasswordTest.java
```diff
@@ -44,4 +56,35 @@ public void testObfuscateUnicode()
...
+            }
+            else
+            {
+                System.out.println(output);
+                fail("Process didn't exit properly (was forcibly destroyed)");
+                passwordProcess.destroy();
```
sbordet, line 91: This line is never executed, as `fail()` above throws.

sbordet (dismissed):

jetty-util/src/test/java/org/eclipse/jetty/util/security/PasswordTest.java
joakime: good point

jetty-util/src/test/java/org/eclipse/jetty/util/security/PasswordTest.java
joakime: will do

jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
joakime: I want to break apart TypeUtil.

The String / Hex methods should exist in StringUtil (which has no logging).

The TypeUtil handling of types, and service loaders, and whatnot can stay.

jetty-util/src/test/java/org/eclipse/jetty/util/security/PasswordTest.java
joakime: Done.

jetty-util/src/test/java/org/eclipse/jetty/util/security/PasswordTest.java
joakime: Fixed

olamy (dismissed):

joakime:
@sbordet ping, mind a review of the changes?

jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
```diff
@@ -134,6 +133,58 @@ protected static boolean byteEquals(byte[] known, byte[] unknown)
         return result && l1 == l2;
     }
 
+    /**
+     * <p>Utility to convert a String of hex into an actual byte array</p>
+     * @param hstr the raw string
+     * @return the parsed byte array
+     */
+    protected static byte[] parseHexToByteArray(String hstr)
```
sbordet: Nope.
This can be implemented in a 2-liner in a `for` loop.
Please see `TypeUtil.fromHexString()`.
Also, make it `private`.

jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
```diff
@@ -134,6 +133,58 @@ protected static boolean byteEquals(byte[] known, byte[] unknown)
...
+        {
+            throw new IllegalArgumentException(String.format("Invalid string length of <%d>", hstr.length()));
+        }
+
+        int size = hstr.length() / 2;
+        byte[] buf = new byte[size];
+        byte hex;
+        int len = hstr.length();
+
+        int idx = (int)Math.floor(((size * 2) - (double)len) / 2);
```
sbordet: That is a really fancy way for `int idx = 0`. Why all the complication?

jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
```diff
@@ -134,6 +133,58 @@ protected static boolean byteEquals(byte[] known, byte[] unknown)
...
+
+        return buf;
+    }
+
+    /**
+     * Format the byte array as a hex string
+     * @param buf the byte array
+     * @return the hex string
+     */
+    protected static String formatAsHex(byte[] buf)
```
sbordet: Make it `private`.

sbordet (changes requested):

jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
joakime: Made it private, simplified it a great deal.
Didn't use Integer.parseInt() technique and int casting.
Kept it small still.

jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
joakime: Removed.

jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
joakime: Made private

jetty-util/src/main/java/org/eclipse/jetty/util/security/Credential.java
```diff
@@ -134,6 +133,48 @@ protected static boolean byteEquals(byte[] known, byte[] unknown)
...
+
+        byte[] buf = new byte[hstr.length() / 2];
+        int len = hstr.length();
+
+        for (int i = 0, idx = 0; i < len; i++)
+        {
+            byte hex = (byte)(Character.digit(hstr.charAt(i++), 16) << 4);
+            hex += (byte)(Character.digit(hstr.charAt(i), 16));
+            buf[idx++] = hex;
+        }
```
sbordet: Too complicated.
Use `TypeUtil.fromHexString()` implementation.

sbordet (changes requested):

sbordet (approved):

PR #7228, opened 2021-12-07
Regularize name of gcloud session module and package.

janbartel:
Change name of `jetty-gcloud-session-manager` to `jetty-gcloud` to be consistent with other session modules. Also change package from `org.eclipse.jetty.gcloud.session` to `org.eclipse.jetty.session.gcloud` to be consistent.


sbordet:
I don't like the name changes, see https://github.com/eclipse/jetty.project/pull/7232#issuecomment-988665822.

olamy (approved):

PR #7230, opened 2021-12-07
Regularize package of jetty-memcached to be consistent.

janbartel:
Change from `org.eclipse.jetty.memcached.sessions` to `org.eclipse.jetty.session.memcached` to be consistent with other sessions modules.

Also fixed the `.mod` file which was referring to the old name of the memcached artifact of `jetty-memcached-sessions` to `jetty-memcached`.


sbordet:
See https://github.com/eclipse/jetty.project/pull/7232#issuecomment-988665822.

olamy (approved):

PR #7231, opened 2021-12-08
Regularize jetty 12.0.x mongo session package

janbartel:
Change `org.eclipse.jetty.mongodb` to `org.eclipse.jetty.session.mongodb` to be consistent with other session modules.


sbordet:
Not in this PR but the restructuring to `jetty-mongodb` does not carry the information that this is about sessions.
I would prefer `jetty-session/session-mongodb`, see https://github.com/eclipse/jetty.project/pull/7232#issuecomment-988665822.

olamy (approved):

PR #7232, opened 2021-12-08
Split out jdbc and file sessions

janbartel:
Split file session and jdbc session implementation out of `jetty-server` and move it into `jetty-session/jetty-file` and `jetty-session/jetty-jdbc` respectively.


sbordet:
@janbartel I would not call it `jetty-file` or jetty-jdbc` because looking at these only it's not evident they refer to sessions.

Maven standards would require `jetty-session/jetty-session-file`, however I feel that's very repetitive to put the parent directory name as a prefix of the subdirectory.

In other places in the jetty project we have used the format `jetty-session/session-file`, for example for `http2` and `http3` where we have e.g. `jetty-http2/http2-server`.

olamy:
> @janbartel I would not call it `jetty-file` or jetty-jdbc` because looking at these only it's not evident they refer to sessions.
> 
> Maven standards would require `jetty-session/jetty-session-file`, however I feel that's very repetitive to put the parent directory name as a prefix of the subdirectory.
> 
> In other places in the jetty project we have used the format `jetty-session/session-file`, for example for `http2` and `http3` where we have e.g. `jetty-http2/http2-server`.

the problem with using only `jetty-http2/http2-server` users will have a jar named `http2-server.jar` so it's hard to know it's coming from the  jetty project.

why not following maven standard so users have jars named `jetty-....` 

another problem with `jetty-http2/http2-server` is the artifactId of the pom which is `http2-parent` maven standard usage would call it the same as the directory.

sbordet:
> the problem with using only jetty-http2/http2-server users will have a jar named http2-server.jar so it's hard to know it's coming from the jetty project.

Fair point, I'm fine with the standard Maven, then.

janbartel:
@sbordet I'm fine with changing all of all session artifacts to be prefixed with `jetty-`, however I would prefer to do that in a new, separate pr after these current crop of prs have been accepted.

sbordet:
@janbartel, @olamy suggested that your proposal would generate a jar called `jetty-file`, which does not hint it's about sessions, while my proposal would generate a jar called `session-file` which does not hint that it's from the Jetty project, so perhaps stick with Maven convention is better, so `jetty-session/jetty-session-file` is best.

Since none of the crop of PRs actually build fine, perhaps it's better to do the right renaming now?

janbartel:
@sbordet the jetty-12 branch doesn't pass distribution tests or the maven tests that use the distribution. It is nothing to do with these prs - it is the .mod files that don't take account of the different servlet spec versions of jetty modules like servlet and webapp. 

I would rather do the rename in another pr because there are other session modules that will need to be renamed for which there is no current pr - best to do the rename as a big bang all at once across all session modules for consistency.

gregw (approved):
I'm not that concerned that we get the naming exactly right.
It is more important to move stuff out of jetty-server, so that we can make progress there to the new architecture with files that no longer belong there nor losing the history of those files.

we can rename later - let's focus on grouping things correctly first

olamy (approved):

PR #7248, opened 2021-12-09
(10.0.x) ByteBufferPool and RetainableByteBufferPool now reset the buffer's endianness on release

### LINKED ISSUE #7243: Reset pooled ByteBuffer endianness
**Target Jetty version(s)**
9.4.x, 10.0.x, 11.0.x

**Enhancement Description**
User code may be passed `ByteBuffer`s that are pooled by Jetty.
They may use the buffer and change the endianness and then forget to reset it back, so the next usage of the buffer by Jetty may use the wrong endianness.

We should reset the endianness when we put back the buffer to the pool, like we already clear it, and like we already reset the interrupted status of pooled threads in the thread pool.

lorban:
See #7243.

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
@@ -21,6 +21,7 @@ import java.io.RandomAccessFile;
 import java.nio.Buffer;
 import java.nio.BufferOverflowException;
 import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 import java.nio.channels.FileChannel;
 import java.nio.channels.FileChannel.MapMode;
 import java.nio.charset.Charset;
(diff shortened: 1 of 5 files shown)


jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
```diff
@@ -166,6 +167,23 @@ public static ByteBuffer copy(ByteBuffer buffer)
         return clone;
     }
 
+    /**
+     * Reset the buffer's endianness to {@link ByteOrder#BIG_ENDIAN}
+     * and clear the buffer to be empty in flush mode.
```
sbordet: Use 3rd person as per javadoc standards ("Resets", "clears"). Wrap with `<p>` the paragraphs.

jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
```diff
@@ -166,6 +167,23 @@ public static ByteBuffer copy(ByteBuffer buffer)
...
+     * The position and limit are set to 0;
```
sbordet: Why semicolon at end of sentence? Something was forgot or typo?

jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
```diff
@@ -166,6 +167,23 @@ public static ByteBuffer copy(ByteBuffer buffer)
...
+     *
+     * @param buffer The buffer to clear.
```
sbordet:
```suggestion
     * @param buffer the buffer to reset
```

sbordet (changes requested):

jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
lorban: This is an exact copy of `clear()`'s javadoc, with the extra _Reset the buffer's endianness..._ bits added on top. I'll fix both.

jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
lorban: Ok.

jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
lorban: Ok.

sbordet (approved):

PR #7251, opened 2021-12-09
Issue #7250 - Update HostHeaderCustomizer for empty Host + no authority

### LINKED ISSUE #7250: HostHeaderCustomizer doesn't address empty Host header value with no authority set
**Jetty version(s)**
9.4.x and 10.x and 11.x

All

**OS type/version**
All

**Description**

If a request arrives without an authority and an empty `Host` header, the use of `HostHeaderCustomizer` doesn't correct this scenario.

Example:

```
GET /foo HTTP/1.1
Host:
Connection: close
```

This has no authority in the request uri line, nor in the Host header, nor any Forwarding headers.
The resulting APIs that rely on authority being valid return the connector local port instead. (eg: `HttpServletRequest.getServerName()` and `HttpServletRequest.getRequestURL()`)

The intent of `HostHeaderCustomizer` is to ensure that the Host header is set and the connector local IP is not accidentally exposed to the webapps.

joakime:

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HostHeaderCustomizer.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HostHeaderCustomizer.java
@@ -21,6 +21,8 @@ package org.eclipse.jetty.server;
 import java.util.Objects;
 import javax.servlet.http.HttpServletRequest;
 
+import org.eclipse.jetty.http.HttpHeader;
+
 /**
  * Customizes requests that lack the {@code Host} header (for example, HTTP 1.0 requests).
  * <p>
@@ -62,7 +64,20 @@ public class HostHeaderCustomizer implements HttpConfiguration.Customizer
     @Override
     public void customize(Connector connector, HttpConfiguration channelConfig, Request request)
     {
-        if (request.getHeader("Host") == null)
-            request.setAuthority(serverName, serverPort);  // TODO set the field as well?
+        String hostHeaderValue = request.getHeader("Host");
+        String authority = request.getHttpURI().getAuthority();
+
+        if (hostHeaderValue == null || // No Host Header
+            hostHeaderValue.trim().isEmpty()) // Empty Host Header
+        {
+            // only reset authority if it's not set yet
+            // other customizers could have set the authority (eg: ForwardedRequestCustomizer)
+            if (authority == null)
+            {
+                request.setAuthority(serverName, serverPort);
+                authority = request.getHttpURI().getAuthority();
+            }
+            request.getHttpFields().put(HttpHeader.HOST, authority);
+        }
     }
 }
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HostHeaderCustomizer.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HostHeaderCustomizer.java
@@ -22,6 +22,8 @@ import java.util.Objects;
 import javax.servlet.http.HttpServletRequest;
 
 import org.eclipse.jetty.http.HttpHeader;
+import org.eclipse.jetty.http.HttpURI;
+import org.eclipse.jetty.http.HttpVersion;
 
 /**
  * Customizes requests that lack the {@code Host} header (for example, HTTP 1.0 requests).
@@ -40,15 +42,14 @@ import org.eclipse.jetty.http.HttpHeader;
  */
 public class HostHeaderCustomizer implements HttpConfiguration.Customizer
 {
-    private final String serverName;
-    private final int serverPort;
+    private final String hostValue;
 
     /**
      * @param serverName the {@code serverName} to set on the request (the {@code serverPort} will not be set)
      */
     public HostHeaderCustomizer(String serverName)
     {
-        this(serverName, 0);
+        this(serverName, HttpURI.NO_PORT);
     }
 
     /**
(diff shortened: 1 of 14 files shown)


joakime:
There is an argument that we should not be setting or caring about the authority.

If we get a request like ...

```
GET foo:///12345 HTTP/1.1
Host:
Connection: close
```

That is a request without a declared authority, as specified by https://datatracker.ietf.org/doc/html/rfc7230#section-5.4
The `foo` scheme could have an implied authority as well.

_For example `file:///path/to/bar` is a `file` scheme which says the implied authority is "localhost" (See https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2)_

If we reach a webapp in the above request that does `HttpServletRequest.sendRedirect("/bar")`, what does that mean to the `Location: ` response header? 
What about the `HttpServletRequest.getServerName()` ?  (oddly, we get `HttpServletRequest.getRequestURL()` correct in the above example)

Question still remaining, do we force the `Host` header, even in the "empty host header" scenario?

joakime:
Needs more tests for bad `Host` values.

Example:

```
Host: ":"
Host: ":43"
Host: -
Host: *
Host: -:88
Host: *.company.com
```

Should these be fixed by `HostHeaderCustomizer` as well?  or treated as `400 Bad Request`? 
I'm kinda leaning towards `400 Bad Request`, and having the new `RejectNoAuthorityCustomizer` renamed to `RejectInvalidAuthorityCustomizer` which also handles oddball authorities like above?  WDYT?

We'll also need a new module for this new customizer.

jetty-server/src/main/java/org/eclipse/jetty/server/RejectInvalidAuthorityCustomizer.java
```diff
@@ -0,0 +1,77 @@
...
+
+        String host = httpURI.getHost();
+        if (StringUtil.isBlank(host))
+            return false;
+        else
+        {
+            // Host must start with alpha-numeric.
+            int c = host.codePointAt(0);
+            if (!Character.isLetterOrDigit(c))
+                return false;
```
gregw, line 64: What about IPv6, which can start with '['

jetty-server/src/main/java/org/eclipse/jetty/server/RejectInvalidAuthorityCustomizer.java
```diff
@@ -0,0 +1,77 @@
...
+ * <p>
+ * This will return a 400 Bad Request on failed authority checks.
+ * </p>
+ */
+public class RejectInvalidAuthorityCustomizer implements HttpConfiguration.Customizer
+{
+    @Override
+    public void customize(Connector connector, HttpConfiguration channelConfig, Request request)
+    {
+        if (!isValidAuthority(request.getHttpURI()))
```
gregw, line 44: Is this correct? Don't we just care that there is an Authority in eitehr the URI or Host header (and if both then they match)?
Or has this been handled for us already and any Host header already copied over to the URI authority?

gregw (changes requested):
I'd really REALLY prefer not to make any changes to `HttpURI`.

Instead focus on making a really good `RejectInvalidAuthorityCustomizer` that will solve the main concern of this PR.  

Then by all means tweak the other Customizers to make them do a better job in the edge cases, but they don't need to be perfect if we have a good RIAC

jetty-server/src/main/java/org/eclipse/jetty/server/RejectInvalidAuthorityCustomizer.java
joakime: handled in different path, and in uncommited code i mentioned.

jetty-server/src/main/java/org/eclipse/jetty/server/RejectInvalidAuthorityCustomizer.java
joakime: Handled differently now, in `HostPort`.

jetty-server/src/main/java/org/eclipse/jetty/server/HostHeaderCustomizer.java
```diff
@@ -55,14 +58,29 @@ public HostHeaderCustomizer(String serverName)
...
 
     @Override
     public void customize(Connector connector, HttpConfiguration channelConfig, Request request)
     {
-        if (request.getHeader("Host") == null)
-            request.setAuthority(serverName, serverPort);  // TODO set the field as well?
+        String hostHeaderValue = request.getHeader("Host");
+
+        // No Host Header
+        if (request.getHttpVersion().getVersion() != HttpVersion.HTTP_1_1.getVersion() &&
```
abahl, line 73: BTW, I see IP gets leaked even for HTTP/1.1 when the Host header has single and/or multiple whitespace values like the following (see cookie value
```
curl  http://localhost:6109  -vl --http1.1 --Header "Host:  "
* Rebuilt URL to: http://localhost:6109/
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 6109 (#0)
> GET / HTTP/1.1
> Host: 
> User-Agent: curl/7.58.0
> Accept: */*
> 
< HTTP/1.1 302 Found
< Date: Tue, 14 Dec 2021 01:09:48 GMT
< Set-Cookie: CookieConsentPolicy=0:0; domain=127.0.0.1; path=/; expires=Wed, 14-Dec-2022 01:09:49 GMT; Max-Age=31536000
< Set-Cookie: LSKey-c$CookieConsentPolicy=0:0; domain=127.0.0.1; path=/; expires=Wed, 14-Dec-2022 01:09:49 GMT; Max-Age=31536000
< X-Content-Type-Options: nosniff
< X-XSS-Protection: 1; mode=block

```
If we have a null host then we get BadMessage as expected for HTTP/1.1, but for an empty host, we see IP. Do we need else case here to check the host with white spaces?

abahl, line 73: This can cause NPE, isn't it? request.getHttpVersion() can return null.

joakime:
This PR will be split into separate PRs.

jetty-server/src/main/java/org/eclipse/jetty/server/HostHeaderCustomizer.java
joakime: The request shouldn't reach this customizer then, as I would expect the request to fail well before the customizers have been called.

joakime:
This PR has now been broken down into ...

* PR #7283 
* PR #7279 
* PR #7292

abahl:
Thanks, Joakim 
It seems a lot of changes and we need to test, ingest and roll it out cautiously as we have many partners also.
We will work on it on our mainline post Holidays hopefully. For now, we pushed a simple fix to check empty host along with null on Http/1.0 as a customer was asking ETA about the fix.

Issue #7252, opened 2021-12-09
HttpChannel.Listener has no post-customizer and pre-dispatch event to hook into


joakime:
**Jetty version(s)**
9.4.x, 10.0.x, 11.0.x

**Java version/vendor** `(use: java -version)`
All

**OS type/version**
All

**Description**
If you want a `HttpChannel.Listener` (for metrics) to correlate events that occur during dispatch to a specific `Host` (or authority), there's no way to do that with the currently order of events.

I looked back at the whole history of `HttpChannel.Listener` and the following behavior has been in place since that listener was introduced in https://github.com/eclipse/jetty.project/issues/1807 (first seen in Jetty 9.4.8.v20171121)

Order of events ...

1. `HttpChannel.dispatch()`
2. `HttpChannel.Listener.onBeforeDispatch()`
3. `Customizer.customize()`
4. `Server.getHandle().handle()`
5. `HttpChannel.Listener.onAfterDispatch()`

If a metrics library wants to note changes to a specific `Host`, that host might not be valid until features like `HostHeaderCustomizer` or `ForwardedRequestCustomizer` are executed first.
There's no way in the current order of events to write a listener that knows the correct post-customizer values, before the dispatch actually occurs.

PR #7255, opened 2021-12-09
Issue #7252 - Customizers before HttpChannel.Listener.onBeforeDispatch

joakime:
Moved `Customizer.customize()` from after `HttpChannel.Listener.onBeforeDispatch()` to before `onBeforeDispatch()`

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
@@ -389,17 +389,15 @@ public class HttpChannel implements Runnable, HttpOutput.Interceptor
                         if (!_request.hasMetaData())
                             throw new IllegalStateException("state=" + _state);
 
-                        dispatch(DispatcherType.REQUEST, () ->
+                        for (HttpConfiguration.Customizer customizer : _configuration.getCustomizers())
                         {
-                            for (HttpConfiguration.Customizer customizer : _configuration.getCustomizers())
-                            {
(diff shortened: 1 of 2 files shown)


gregw (commented):
I'm not convinced we should make such a big behaviour change in 9.4
In hindsight, perhaps it should have been this way.... except that  customizers can also be considered user code and can actually handle the request.... so perhaps they are a dispatch?

I think this is something for jetty-12 consideration

gregw (changes requested):
I think the new ordering is better..... but we should have an already deprecated option to revert to the old ordering in case we break somebody.

joakime:
While I like the new ordering as it makes sense.
The motivation around this change is lessened due to changes in other PRs.
Perhaps this should be for Jetty 12+? and not Jetty 9+?

joakime:
> I think this is something for jetty-12 consideration

@gregw Still good with this being redone for `jetty-12.0.x-hackathon` branch?

gregw:
Kind of.   Currently 12 has dropped the HttpChannel.Listener concept... but I expect that we will need either to restore it, or a similar mechanism.
So the ordering concern expressed in this PR should definitely be considered when we add listeners back to 12.

Maybe for now, it should just become a TODO in the 12 branch(es) to indicate that listeners should be called after customization.

joakime:
We don't have Customizers or HttpChannel.Listener anymore, so ... ??

joakime:
Closing.

Opened as Issue #8218 instead

Issue #7259, opened 2021-12-10
InputStreamResponseListener should have a read timeout


sbordet:
**Target Jetty version(s)**
9.4.x

**Enhancement Description**
Class `InputStreamResponseListener` is currently using an infinite wait in case a read blocks.

This is sub-optimal because there is a risk that the caller thread waits forever, even if the class is designed to wake up blocked reads in case of errors, timeouts, close, etc.

PR #7260, opened 2021-12-10
Fixes InputStreamResponseListener should have a read timeout #7259

sbordet:
Added constructor with read timeout parameter.
Updated implementation to honor the read timeout.
Added test case.

### DIFF
--- a/documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/client/http/HTTPClientDocs.java
+++ b/documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/client/http/HTTPClientDocs.java
@@ -424,7 +424,8 @@ public class HTTPClientDocs
         httpClient.start();
 
         // tag::inputStreamResponseListener[]
-        InputStreamResponseListener listener = new InputStreamResponseListener();
+        long readTimeout = 5000; // milliseconds
+        InputStreamResponseListener listener = new InputStreamResponseListener(readTimeout);
         httpClient.newRequest("http://domain.com/path")
             .send(listener);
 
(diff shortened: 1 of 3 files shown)


lorban:
This looks redundant with the idle timeout mechanism to me. What's this bringing that idleTimeout cannot do, or what am I missing?

sbordet:
@lorban the idle timeout of the `ClientConnector` cannot be accessed from this class.

I can use that of the request by default, but only if it has been set by the application.

joakime:
Do you want this in upcoming 9.4.45 release too?

sbordet:
@joakime yes please

lorban:
@sbordet my point isn't where to get the timeout value from, but rather than the idleTimeout mechanism should be the only one that unblocks the awaiting thread after a certain time elapsed, at least IMHO.

Adding an extra timeout looks to me like replicating an existing feature just in case that said feature doesn't work as expected.

sbordet:
@gregw thoughts?

joakime:
@lorban what project do you want this PR in? (10.0.10?)

gregw:
@sbordet I'm with @lorban on this? Why is there a second timer?   If I understand correctly there is a HttpClient request, that will have one idle timeout, then a listener that can have another.
So without this PR, a user can make a HttpClient request and start reading from the input stream listener. They will block waiting for content and if the HttpClient request timesout, then their input stream will fail accordingly.
With this PR, the input stream may timeout before the HttpClient has?  This makes no sense to me?

Without the second timeout, is there ever a case this listener will wait forever for the HttpClient to succeed or fail?  The issue only says:

 > his is sub-optimal because there is a risk that the caller thread waits forever

But surely that would be a bug in HttpClient?

joakime:
@sbordet *bump* what's the status on this?

gregw:
I think this should be closed. @sbordet?

sbordet:
Closing this but leaving the issue open.

PR #7267, opened 2021-12-13
Issue #7266 - fix ALPN module selection based on runtime.feature.alpn property

### LINKED ISSUE #7266: Wrong ALPN jars are selected for newer versions of JDK8
**Jetty version(s)**
9.4.x

jdk1.8.0_281

**Description**
The `runtime.feature.alpn` is set to true and because of this the module `alpn-impl/alpn-available-true` is enabled instead of `alpn-impl/alpn-available-false`. 

This results in `jetty-alpn-openjdk8-server.jar` not being on the server classpath and the server giving the following exception on start.

```
...
Caused by: java.lang.IllegalStateException: No Server ALPNProcessors!
	at org.eclipse.jetty.alpn.server.ALPNServerConnectionFactory.<init>(ALPNServerConnectionFactory.java:52)
	at org.eclipse.jetty.alpn.server.ALPNServerConnectionFactory.<init>(ALPNServerConnectionFactory.java:45)
	... 26 more
	Suppressed: java.lang.UnsupportedClassVersionError: org/eclipse/jetty/alpn/java/server/JDK9ServerALPNProcessor has been compiled by a more recent version of the Java Runtime (class file version 53.0), this version of the Java Runtime only recognizes class file versions up to 52.0
		at java.lang.ClassLoader.defineClass1(Native Method)
		at java.lang.ClassLoader.defineClass(ClassLoader.java:756)
		at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
		at java.net.URLClassLoader.defineClass(URLClassLoader.java:468)
		at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
		at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
		at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
		at java.security.AccessController.doPrivileged(Native Method)
		at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
		at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
		at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
		at java.lang.Class.forName0(Native Method)
		at java.lang.Class.forName(Class.java:348)
		at java.util.ServiceLoader$LazyIterator.nextService(ServiceLoader.java:370)
		at java.util.ServiceLoader$LazyIterator.next(ServiceLoader.java:404)
		at java.util.ServiceLoader$1.next(ServiceLoader.java:480)
		at org.eclipse.jetty.alpn.server.ALPNServerConnectionFactory.<init>(ALPNServerConnectionFactory.java:59)
		... 27 more
```

Originally reported at https://github.com/eclipse/jetty.docker/issues/49

**How to reproduce?**
```
java -jar $JETTY_HOME/start.jar --add-to-startd=http2
java -jar $JETTY_HOME/start.jar
```

lachlan-roberts:
closes #7266

### DIFF
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
@@ -1502,7 +1502,8 @@ public class StartArgs
                 properties.setProperty("java.version.micro", Integer.toString(ver.getMicro()), "Deprecated");
 
                 // ALPN feature exists
-                properties.setProperty("runtime.feature.alpn", Boolean.toString(ver.getPlatform() >= 9), source);
+                properties.setProperty("runtime.feature.alpn", Boolean.toString(isMethodAvailable(javax.net.ssl.SSLParameters.class, "getApplicationProtocols", null)), source);
+                properties.setProperty("at.least.java9", Boolean.toString(ver.getPlatform() >= 9), source);
             }
             catch (Throwable x)
             {
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
@@ -1502,7 +1502,7 @@ public class StartArgs
                 properties.setProperty("java.version.micro", Integer.toString(ver.getMicro()), "Deprecated");
 
                 // ALPN feature exists
-                properties.setProperty("runtime.feature.alpn", Boolean.toString(isMethodAvailable(javax.net.ssl.SSLParameters.class, "getApplicationProtocols", null)), source);
+                properties.setProperty("runtime.feature.alpn", Boolean.toString(ver.getPlatform() >= 9), source);
             }
             catch (Throwable x)
             {
--- a/jetty-alpn/jetty-alpn-server/src/main/config/modules/alpn-impl/alpn-16.mod
+++ b/jetty-alpn/jetty-alpn-server/src/main/config/modules/alpn-impl/alpn-16.mod
@@ -1,4 +1,0 @@
-# DO NOT EDIT - See: https://www.eclipse.org/jetty/documentation/current/startup-modules.html
-
-[depend]
-alpn-impl/alpn-11
--- a/jetty-alpn/jetty-alpn-server/src/main/config/modules/alpn.mod
+++ b/jetty-alpn/jetty-alpn-server/src/main/config/modules/alpn.mod
@@ -5,7 +5,7 @@ Enables the ALPN (Application Layer Protocol Negotiation) TLS extension.
 
 [depend]
 ssl
-alpn-impl/alpn-available-${runtime.feature.alpn}
+alpn-impl/alpn-available-${at.least.java9}
 
 [lib]
 lib/jetty-alpn-client-${jetty.version}.jar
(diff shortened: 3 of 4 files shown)


jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
```diff
@@ -1502,7 +1502,7 @@ public void setProperty(String key, String value, String source)
                 properties.setProperty("java.version.micro", Integer.toString(ver.getMicro()), "Deprecated");
 
                 // ALPN feature exists
-                properties.setProperty("runtime.feature.alpn", Boolean.toString(isMethodAvailable(javax.net.ssl.SSLParameters.class, "getApplicationProtocols", null)), source);
+                properties.setProperty("runtime.feature.alpn", Boolean.toString(ver.getPlatform() >= 9), source);
```
joakime: This is wrong.
Java 8 has support too.
The previous version checked for the existence of a specific class to know if the ALPN support was backported.

joakime (changes requested):
I don't like this.

What's the scenario to reproduce with the old code?

joakime:
We have 3 states to worry about in Jetty 8.

1. Old Java 8, needs `-XBootclasspath`
2. New Java 8, needs `jetty-alpn-openjdk8-server-<jetty.version>.jar`
3. Java 9+, needs `jetty-alpn-java-server-<jetty.version>.jar`

The `runtime.feature.alpn` property was created to address 2 and 3 above, but not 1.
Since `runtime.feature.alpn` was created, the `alpn-agent` was integrated.
So now 1 and 2 above are handled by the `alpn-agent`, and 3 is handled by direct use of `jetty-alpn-java-server-<jetty.version>.jar`

There is no longer a purpose for `runtime.feature.alpn` for the shipped `jetty-home` modules.

Don't change the existing `runtime.feature.alpn` however, as there are still cases with embedded systems and custom jetty.base distributions with different software where we need this as it stood before (eg: Product Foo is attempting to be installed on a Java 9+ but with limited featureset that lacks alpn featureset).

What this PR should do.

* Delete `alpn-16.mod` that's just noise and is detritus from an old merge, and is no longer referenced AFAICT.
* Introduce a new detection of java.version is 9 or above, perhaps call it `java.version.modern` with a value of `old` (java 8) or `modern` (java 9).

As for how to detect Java 9+, some ideas:
* Does `java.lang.Runtime.Version` exist? (You are on Java 9+)
* Does `java.lang.Module` exist? (You are on Java 9+)
* A simple Version check like you are currently doing in this this PR.  (We historically didn't do this as we came across some wacky JVMs that had version strings that broke the version check, but I think those were all limited to Java 8 or older, and the versioning rules in Java 9+ became far more strict)

So you will wind up with ...

JVM | ALPN Support | Module
---- | ----- | ----
8u201 (needs `xbootclasspath`) | `alpn-agent` | `alpn-java-old`
8u282 (needs `jetty-alpn-openjdk8-server.jar`) | `alpn-agent` | `alpn-java-old`
9+ | lib=`jetty-alpn-java-server-<jetty.version>.jar` | `alpn-java-modern`

What needs to also be tested: The existing `alpn-available-false.mod` automatically includes the `jetty-alpn-openjdk8-server` which needs to be tested on an old pre-backport-of-alpn version of Java 8 to know if that's still a sane setup.

lachlan-roberts:
Tested and working with `8u201`, `8u281`, and `11.0.10`.

joakime (approved):
:+1:

PR #7270, opened 2021-12-14
Replace `SharedBlockingCallback` with `Blocking` class

gregw:
This single class replaces `SharedBlockingCallback` with the ability to support both `Runnable` and `Callback` APIs, plus single use instances.

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/Blocking.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/Blocking.java
@@ -1,0 +1,305 @@
+//
+// ========================================================================
+// Copyright (c) 1995-2021 Mort Bay Consulting Pty Ltd and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.util;
+
+import java.io.IOException;
+import java.io.InterruptedIOException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
+
+import org.eclipse.jetty.util.thread.Invocable;
+
+/**
+ * Utility class that provides blocking {@link java.lang.Runnable} and {@link org.eclipse.jetty.util.Callback}
+ * instances.  These can either be shared (and mutually excluded from concurrent usage) or single usage.
+ * The instances are autocloseable, so blocking takes place as a try block is exited.  Example usages:
+ *
+ * <h2>Non shared Runnable</h2>
+ * <pre>
+ *     try(Blocking.Runnable onAction = Blocking.runnable())
+ *     {
+ *         someMethod(onAction);
+ *     }
+ * </pre>
+ *
+ * <h2>Shared Runnable</h2>
+ * <pre>
+ *     Blocking.SharedRunnable shared = new Blocking.Shared();
+ *     // ...
+ *     try(Blocking.Runnable onAction = shared.runnable())
+ *     {
+ *         someMethod(onAction);
+ *     }
+ * </pre>
+ *
+ * <h2>Non shared Callback</h2>
+ * <pre>
+ *     try(Blocking.Callback callback = Blocking.callback())
+ *     {
+ *         someMethod(callback);
+ *     }
+ * </pre>
+ *
+ * <h2>Shared Callback</h2>
+ * <pre>
+ *     Blocking.SharedCallback blocker = new Blocking.Shared();
+ *     // ...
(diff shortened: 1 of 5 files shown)


jetty-util/src/main/java/org/eclipse/jetty/util/Blocking.java
```diff
@@ -0,0 +1,305 @@
...
+            }
+
+            @Override
+            public void failed(Throwable x)
+            {
+                _future.complete(x == null ? new Throwable() : x);
+            }
+
+            @Override
+            public void close() throws IOException
```
lachlan-roberts, line 149: This would probably be more useful if it also had a close method with a timeout.

See the `FutureCallback` class which does a similar thing, there is `block()` method for infinite timeout but also a `block(long timeout, TimeUnit unit)` method. Because of this it can be used for the blocking WebSocket sends.

gregw: I'd prefer to leave that until we actually have a use for it.    Specially for the shared instances, as if you wait for a time and it times out, then what does that mean for the shared instance? Is it still valid ? usable? is it reset?
Tell you what, I'll add it for the single use ones....

gregw: Ah I can't as they have the same signature returned.   So let's not add them unless there is a real use-case.

gregw:
@lorban @sbordet can I get review on these. Need to make some progress.

gregw:
Hmmmm

I have removed the explict call to `block()` so you can just do:
```java
    try (Blocking.Callback callback = Blocking.callback())
    {
        someMethod(callback);
    }
```

But the problem here is that if `someMethod` throws, (which it shouldn't but might as we are not that rigorous) then the autoclose still ends up blocking, which may block forever.
Thus I think it is safer to revert to:

```java
    try (Blocking.Callback callback = Blocking.callback())
    {
        someMethod(callback);
        callback.block();
    }
```

So the block is skiped if the method throws.   We can still be auto closeable and emit a warning if we have not blocked.

@sbordet @lachlan-roberts @lorban   Your thoughts?   Please don't wait for this for any other review, as I intend to merge with my 12 Handler branch and then merge to 12.0.x later today regardless.

jetty-util/src/main/java/org/eclipse/jetty/util/Blocking.java
```diff
@@ -0,0 +1,305 @@
...
+ * <pre>
+ *     try(Blocking.Runnable onAction = Blocking.runnable())
+ *     {
+ *         someMethod(onAction);
+ *     }
+ * </pre>
+ *
+ * <h2>Shared Runnable</h2>
+ * <pre>
+ *     Blocking.SharedRunnable shared = new Blocking.Shared();
```
sbordet, line 40: `Blocking.SharedRunnable` is not a valid class.

jetty-util/src/main/java/org/eclipse/jetty/util/Blocking.java
```diff
@@ -0,0 +1,305 @@
...
+ * <pre>
+ *     try(Blocking.Callback callback = Blocking.callback())
+ *     {
+ *         someMethod(callback);
+ *     }
+ * </pre>
+ *
+ * <h2>Shared Callback</h2>
+ * <pre>
+ *     Blocking.SharedCallback blocker = new Blocking.Shared();
```
sbordet, line 58: `Blocking.SharedCallback` is not a valid class.

jetty-util/src/main/java/org/eclipse/jetty/util/Blocking.java
```diff
@@ -0,0 +1,305 @@
...
+ *         someMethod(onAction);
+ *     }
+ * </pre>
+ */
+public class Blocking
+{
+    private static final Throwable ACQUIRED = new Throwable()
+    {
+        @Override
+        public synchronized Throwable fillInStackTrace()
```
sbordet, line 71: Remove `synchronized`.

jetty-util/src/main/java/org/eclipse/jetty/util/Blocking.java
```diff
@@ -0,0 +1,305 @@
...
+        {
+            return this;
+        }
+    };
+    private static final Throwable SUCCEEDED = new Throwable()
+    {
+        @Override
+        public synchronized Throwable fillInStackTrace()
```
sbordet, line 79: Remove `synchronized`.

jetty-util/src/main/java/org/eclipse/jetty/util/Blocking.java
```diff
@@ -0,0 +1,305 @@
...
+            @Override
+            public InvocationType getInvocationType()
+            {
+                return InvocationType.NON_BLOCKING;
+            }
+
+            @Override
+            public void succeeded()
+            {
+                _future.complete(SUCCEEDED);
```
sbordet, line 139: Would not `null` instead of `SUCCEEDED` be enough?

sbordet (changes requested):
Overall, don't like it.
The fact that the blocking happens hidden in the closing brace is too much magic.

I like the old semantic of:

```
try(Blocker blocker = sharedBlockingCallback.acquire())
 {
     someAsyncCall(args, blocker);
     blocker.block(); // Explicit blocking -- reading the code makes it obvious.
 }
```

Issue #7271, opened 2021-12-14
It is necessary to set MAX_CAPACITY to ArrayTernaryTrie/ArrayTrie


tinswzy:
**Jetty version(s)**
9.3+, 9.4+

**Java version/vendor** `(use: java -version)`
All

**OS type/version**
All

**Description**
It is necessary to backport MAX_CAPACITY bugfix to ArrayTernaryTrie/ArrayTrie，as there are still many mainstream projects using jetty9.x. 
In 9.x, the ContextHandlerCollection uses the ArrayTernaryTrie data structure by default. The data types of  _rows&_tree in ArrayTernaryTrie are char/char[]. Their capacity is limited, otherwise routing bugs will occur if the value is exceeded.

**How to reproduce?**
@Test
public void testArrayTernaryTrieBug() throws Exception
{
    Trie<String> trie = new ArrayTernaryTrie<>(false,400000);

    // put root path
    boolean success = trie.put("","/");
    assertTrue(success);
    String bestUrl = trie.get("/",1,0);
    assertEquals("/",bestUrl);

    // put a large path or many paths
    char[] c1 = new char[65535];
    Arrays.fill(c1,'a');
    success = trie.put(new String(c1),"/"+new String(c1));
    // expect put failed, but now it success
    assertFalse(success);

    bestUrl = trie.get("/",1,0);
    assertEquals("/",bestUrl);
}

sbordet:
Fixed by #7405.

PR #7273, opened 2021-12-14
Issue #7271 - Addressing MAX_CAPACITY limit in ArrayTernaryTrie and ArrayTrie

tinswzy:
#7271 
Signed-off-by: tinswzy <tinswzy@gmail.com>

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/ArrayTernaryTrie.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/ArrayTernaryTrie.java
@@ -70,6 +70,13 @@ public class ArrayTernaryTrie<V> extends AbstractTrie<V>
     private static final int ROW_SIZE = 4;
 
     /**
+     * The maximum capacity of the implementation. Over that,
+     * the 16 bit indexes can overflow and the trie
+     * cannot find existing entries anymore.
+     */
+    private static final int MAX_CAPACITY = Character.MAX_VALUE;
+
+    /**
      * The Trie rows in a single array which allows a lookup of row,character
      * to the next row in the Trie.  This is actually a 2 dimensional
      * array that has been flattened to achieve locality of reference.
@@ -140,6 +147,8 @@ public class ArrayTernaryTrie<V> extends AbstractTrie<V>
     public ArrayTernaryTrie(boolean insensitive, int capacity)
     {
         super(insensitive);
+        if (capacity > MAX_CAPACITY)
+            throw new IllegalArgumentException("ArrayTernaryTrie maximum capacity overflow (" + capacity + " > " + MAX_CAPACITY + ")");
         _value = (V[])new Object[capacity];
         _tree = new char[capacity * ROW_SIZE];
         _key = new String[capacity];
@@ -155,6 +164,8 @@ public class ArrayTernaryTrie<V> extends AbstractTrie<V>
     {
         super(trie.isCaseInsensitive());
         int capacity = (int)(trie._value.length * factor);
+        if (capacity > MAX_CAPACITY)
+            throw new IllegalArgumentException("ArrayTernaryTrie maximum capacity overflow (" + capacity + " > " + MAX_CAPACITY + ")");
         _rows = trie._rows;
         _value = Arrays.copyOf(trie._value, capacity);
         _tree = Arrays.copyOf(trie._tree, capacity * ROW_SIZE);
@@ -190,10 +201,15 @@ public class ArrayTernaryTrie<V> extends AbstractTrie<V>
                 if (t == _rows)
                 {
                     _rows++;
-                    if (_rows >= _key.length)
+                    if (_rows + 1 >= _key.length)
                     {
-                        _rows--;
-                        return false;
+                        if (k + 1 < limit)
+                        {
+                            // More than 1 character remains unprocessed, it must be failed,
+                            // otherwise it should be success
+                            _rows--;
+                            return false;
+                        }
                     }
                     _tree[row] = c;
                 }
@@ -223,10 +239,11 @@ public class ArrayTernaryTrie<V> extends AbstractTrie<V>
         if (t == _rows)
         {
             _rows++;
-            if (_rows >= _key.length)
+            if (_rows + 1 >= _key.length)
             {
                 _rows--;
-                return false;
+                // should not return false, here the element already fill into the trie
+                // return false;
             }
         }
 
@@ -530,7 +547,7 @@ public class ArrayTernaryTrie<V> extends AbstractTrie<V>
     @Override
     public boolean isFull()
     {
-        return _rows + 1 == _key.length;
+        return _rows + 1 >= _key.length;
     }
 
     public static int hilo(int diff)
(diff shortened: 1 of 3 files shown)


sbordet:
@lorban can you review this? Is the backporting fine, as in 9.4.x the implementation is different?

ocean613:
@tinswzy  @lorban  When will this patch be available? I have a similar problem when the “_rows” variable overflows, and the ArrayTernaryTrie.getBest method returns an unexpected value.

tinswzy:
@lorban @sbordet Any suggestions? Refactor  ArrayTernaryTrie/ArrayTire to support Integer.MAX_VALUE capacity or set the maximum Character.MAX_VALUE as before?

jetty-util/src/test/java/org/eclipse/jetty/util/TrieTest.java
```diff
@@ -250,4 +254,60 @@ public void testFull(Trie<Integer> trie) throws Exception
...
+    {
+        ArrayTernaryTrie<String> trie = new ArrayTernaryTrie<>(Character.MAX_VALUE);
+
+        char[] c1 = new char[Character.MAX_VALUE - 1];
+        Arrays.fill(c1, 'a');
+        String huge = new String(c1);
+        assertTrue(trie.put(huge, "wow"));
+        assertThat(trie.get(huge), is("wow"));
+
+        assertThrows(IllegalArgumentException.class, () -> new ArrayTrie<String>(Character.MAX_VALUE + 1));
```
lorban: There is already a separate test for this assertion, it can be removed.

jetty-util/src/test/java/org/eclipse/jetty/util/TrieTest.java
```diff
@@ -250,4 +254,60 @@ public void testFull(Trie<Integer> trie) throws Exception
...
+    }
+
+    @Test
+    public void testArrayTernaryOverflowReject()
```
lorban: This test name looks odd. Should it read `testArrayTrieOverflowReject()`?

Also, this test should also work for `ArrayTernaryTrie` so you should parameterize this test too.

jetty-util/src/test/java/org/eclipse/jetty/util/TrieTest.java
```diff
@@ -250,4 +254,60 @@ public void testFull(Trie<Integer> trie) throws Exception
...
+    }
+
+    @Test
+    public void testArrayTernaryTrieCapacityOverFlow()
+    {
+        assertThrows(IllegalArgumentException.class, () -> new ArrayTernaryTrie<String>(Character.MAX_VALUE + 1));
+    }
+
+    @Test
+    public void testArrayTrieCapacity()
```
lorban: This test and the one below (`testArrayTernaryTrieCapacity`) only differ by their trie implementation. You should refactor them as a single parameterized test.

jetty-util/src/test/java/org/eclipse/jetty/util/TrieTest.java
```diff
@@ -250,4 +254,60 @@ public void testFull(Trie<Integer> trie) throws Exception
...
+        assertTrue(trie.put("", "/"));
+        assertThat(trie.getBest("/", 1, 0), is("/"));
+
+        char[] c1 = new char[Character.MAX_VALUE];
+        Arrays.fill(c1, 'a');
+        String huge = new String(c1);
+        assertFalse(trie.put(huge, "overflow"));
+        assertNull(trie.get(huge));
+
+        // The previous variable is not overwritten
```
lorban: I guess you mean `The previous entry was not overridden`?

jetty-util/src/main/java/org/eclipse/jetty/util/ArrayTrie.java
```diff
@@ -49,6 +49,7 @@
  */
 public class ArrayTrie<V> extends AbstractTrie<V>
 {
+    public static int MAX_CAPACITY = Character.MAX_VALUE;
```
lorban, line 52: This should be `private` and `final`.

jetty-util/src/main/java/org/eclipse/jetty/util/ArrayTernaryTrie.java
```diff
@@ -530,7 +541,7 @@ public boolean isEmpty()
     @Override
     public boolean isFull()
     {
-        return _rows + 1 == _key.length;
+        return _rows + 1 >= _key.length;
```
lorban, line 550: This will overflow if `_rows == Character.MAX_VALUE`

jetty-util/src/main/java/org/eclipse/jetty/util/ArrayTernaryTrie.java
```diff
@@ -190,7 +201,7 @@ public boolean put(String s, V v)
                 if (t == _rows)
                 {
                     _rows++;
-                    if (_rows >= _key.length)
+                    if (_rows > _key.length)
```
lorban: This change will make line 209 throw `ArrayIndexOutOfBoundsException` if `_rows == _key.length` which happens when `t == Character.MAX_VALUE`.

lorban (changes requested):
I'm not against the idea of backporting the `MAX_CAPACITY` check from the jetty-10.0.x branch, but the 10.0.x `ArrayTernaryTrie` seems to be buggy in that regard so I would not backport those changes.

The simplest thing to do would probably be to keep the 9.4.x code as-is, add the max capacity check with the `MAX_CAPACITY` at `Character.MAX_VALUE -1` but that value should be tested.

I'm also not certain `ArrayTrie` can support 65k entries so that would need to be tested too.

jetty-util/src/test/java/org/eclipse/jetty/util/TrieTest.java
tinswzy: Got it!

jetty-util/src/test/java/org/eclipse/jetty/util/TrieTest.java
tinswzy: Ok

jetty-util/src/main/java/org/eclipse/jetty/util/ArrayTernaryTrie.java
tinswzy: I will fix it~

tinswzy:
> I'm not against the idea of backporting the `MAX_CAPACITY` check from the jetty-10.0.x branch, but the 10.0.x `ArrayTernaryTrie` seems to be buggy in that regard so I would not backport those changes.
> 
> The simplest thing to do would probably be to keep the 9.4.x code as-is, add the max capacity check with the `MAX_CAPACITY` at `Character.MAX_VALUE -1` but that value should be tested.
> 
> I'm also not certain `ArrayTrie` can support 65k entries so that would need to be tested too.

I think you're right. Just "Add the Max capacity check" is a good choice.

lorban:
There still are problems in your PR; they were quicker to fix than to explain so I've created the following PR you're welcome to check and comment on: https://github.com/eclipse/jetty.project/pull/7405

In a nutshell, the fundamental capacity of the tries is 65534 entries as index values 0 and 65535 are special cases respectively in ArrayTrie and ArrayTernaryTrie. I've added the missing checks and made sure to not overflow `char`. I've also modified the test using the empty string as a key as this is a special case of the ArrayTrie implementation.

joakime:
@lorban now that your PR #7405 is merged with a more comprehensive fix, can we close this one?

lorban:
Yes, this issue can now be closed as it was superseded by https://github.com/eclipse/jetty.project/pull/7405.

PR #7282, opened 2021-12-14
Destroy interceptor when content producer gets recycled

### LINKED ISSUE #7280: Interceptors don't get destroyed in HttpInput
**Jetty version(s)**
10.x and 11.x

**Java version/vendor**
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.7+10, mixed mode)

**OS type/version**
Windows 10 20H2, 64 bit

**Description**
When migrating our product from Jetty 9.4 to 11.0, we noticed that our custom interceptors were not destroyed any more. Between Jetty 9.4 and 10.x/11.x, it is obvious that a lot of code from `HttpInput.recycle()` (that did the trick before) has vanished:

image

This happened during a large refactoring of `HttpInput` starting with PR #4556. Maybe @lorban , @sbordet or @gregw know more? Is this an oversight or has the lifecycle of interceptors somehow changed? Are we the only ones who have this problem?

In our case, this solved the problem (but we don't know enough about Jetty to be sure this doesn't break something else):

```
    public void recycle()
    {
        try (var ignored = _contentProducer.lock()) {
            if (LOG.isDebugEnabled())
                LOG.debug("recycle {}", this);
            Interceptor interceptor = _contentProducer.getInterceptor();
            if (interceptor instanceof Destroyable) {
                ((Destroyable) interceptor).destroy();
            }
        }
    }

```

**How to reproduce?**
For both Jetty 9.4 and 11.0, `HttpInput.recycle()` is on a call-chain triggered by `HttpConnection.onCompleted()`.

lorban:
This is a behavior that was present in 9.4 and was mistakenly omitted in 10.0 during the Grand HttpInput Rewrite.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
@@ -19,6 +19,7 @@ import java.util.concurrent.locks.Condition;
 
 import org.eclipse.jetty.http.BadMessageException;
 import org.eclipse.jetty.http.HttpStatus;
+import org.eclipse.jetty.util.component.Destroyable;
 import org.eclipse.jetty.util.thread.AutoLock;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


sbordet (dismissed):
Just remove space after cast in `AsyncContentProducer`.

sbordet (approved):

PR #7283, opened 2021-12-14
Issue #7277 - Allow `Request.getLocalName()` and `.getLocalPort()` to be overridden

### LINKED ISSUE #7277: Allow override of `ServletRequest.getLocalName()` and `.getLocalPort()` in post-intermediary scenarios
**Jetty version(s)**
9+

All

**OS type/version**
All

**Description**
Per the Servlet API javadoc for `HttpServletRequest.getServerName()`

``` java
    /**
     * Returns the host name of the server to which the request was sent.
     * It is the value of the part before ":" in the <code>Host</code>
     * header value, if any, or the resolved server name, or the server IP
     * address.
     *
     * @return a <code>String</code> containing the name of the server
     */
    public String getServerName();
```

When we have a server that's behind a intermediary and not exposed directly to the internet, and
requests that lack an authority (see HTTP/1 specs) we don't return a valid "resolved server name" 
or "server IP address" that is usable by the application.  In some circles exposing this information
is considered a security issue.

The "resolved server name" internally comes from calling the `ServletRequest.getLocalName()` method,
which Jetty overrides in `org.eclipse.jetty.server.Request.getLocalName()`.

It would be nice to have a configurable (on the Connector? the Channel?) to allow an arbitrary override
of the "resolved server name" step in the `getLocalName()` API contract.

joakime:
Uses a Connection.Listener to set the override on connections that are opened using an `HttpChannel`.


jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
```diff
@@ -298,6 +301,26 @@ public int getAcceptors()
         return _acceptors.length;
     }
 
+    /**
+     * @return Returns the optional local name override
+     */
+    @ManagedAttribute("local name override")
+    public String getLocalNameOverride()
```
sbordet: I think just `getLocalName()` is enough and consistent with all the other APIs of the same name in `HttpChannel` and `Request`.

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
```diff
@@ -318,6 +341,12 @@ protected void doStart() throws Exception
         }
 
         _lease = ThreadPoolBudget.leaseFrom(getExecutor(), this, _acceptors.length);
+
+        if (_localNameOverride != null)
+        {
+            addBean(new HttpChannelLocalNameOverrideListener(_localNameOverride));
+        }
```
sbordet: Not necessary, see below.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -298,6 +301,64 @@ public EndPoint getEndPoint()
...
+     * such as absolute-URI creation (eg: Location response header).
+     * </p>
+     *
+     * @return the local name if overridden, or the local address, or
+     * the local host name.
+     */
+    public String getLocalName()
+    {
+        if (overriddenLocalName != null)
+            return overriddenLocalName;
```
sbordet: From here you can access the `Connector`.

So this code becomes something like:

```java
Connector connector = getConnector();
String localName = null;
if (connector instanceof AbstractConnector)
  localName = ((AbstractConnector)connector).getLocalName();
if (localName != null)
  return localName;
```

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannelLocalNameOverrideListener.java
```diff
@@ -0,0 +1,50 @@
...
+//
+
+package org.eclipse.jetty.server;
+
+import org.eclipse.jetty.io.Connection;
+
+/**
+ * Sets the {@link HttpChannel#setLocalNameOverride(String)} on all newly opened connections.
+ */
+public class HttpChannelLocalNameOverrideListener implements Connection.Listener
```
sbordet: Not necessary.

sbordet (changes requested):

joakime:
@sbordet changes applied

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -298,6 +301,70 @@ public EndPoint getEndPoint()
...
+     * This allows the name returned by APIs like HttpServletRequest.getServerName() to be overridden with
+     * an arbitrary name when behind an intermediary and the incoming request has no valid uri authority.
+     * </p>
+     *
+     * @param overriddenLocalName the overridden local name, use null to set behavior back to default
+     */
+    public void setLocalNameOverride(String overriddenLocalName)
+    {
+        this.overriddenLocalName = overriddenLocalName;
+    }
```
sbordet: Not needed anymore.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -298,6 +301,70 @@ public EndPoint getEndPoint()
...
+        {
+            String name = InetAddress.getLocalHost().getHostName();
+            if (StringUtil.ALL_INTERFACES.equals(name))
+                return null;
+            return name;
+        }
+        catch (UnknownHostException e)
+        {
+            LOG.ignore(e);
+        }
```
sbordet: Don't think this code should be there.
It's typically wrong if `/etc/hosts` is not setup properly, it does not work for Unix-Domain, leaks the internal host name or IP, etc.

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -1025,9 +1025,9 @@ public String getLocalName()
...
         {
-            InetSocketAddress local = _channel.getLocalAddress();
-            if (local != null)
-                return formatAddrOrHost(local.getHostString());
+            String localName = _channel.getLocalName();
+            if (localName != null)
+                return formatAddrOrHost(localName);
         }
 
         try
```
sbordet: See above, remove this code?

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
```diff
@@ -298,6 +301,24 @@ public int getAcceptors()
...
+    public String getLocalName()
+    {
+        return _localName;
+    }
+
+    /**
+     * Optional override of connection local name used within application API layer
+     * when identifying the local host name of a connected endpoint.
+     */
+    public void setLocalName(String localName)
```
sbordet: Do we just need a `localName` or we actually need a `localAuthority` (i.e. host _and_ port)?
@gregw thoughts?

sbordet (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
gregw: authority is better

joakime: Still on AbstractConnector?
As in `AbstractConnector.setLocalAuthority(String authority)`?

joakime: Changed to use authority instead.
Host AND port must be specified though.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -97,6 +97,7 @@
     private MetaData.Response _committedMetaData;
     private RequestLog _requestLog;
     private long _oldIdleTimeout;
+    private String overriddenLocalName;
```
gregw: this is not used

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
```diff
@@ -298,6 +302,42 @@ public int getAcceptors()
...
     }
 
+    /**
+     * @return Returns the optional local authority override
+     */
+    @ManagedAttribute("local authority")
+    public HostPort getLocalAuthority()
+    {
+        return _localAuthority;
+    }
```
gregw, line 309: This method should always return a local authority, either one set explicitly or one that is determined dynamically in `doStart` (or `open`).
Then the layers about can just ask for the authority and not have to check if null and then look at the address etc.

gregw: Also make this a method on `NetworkConnector`

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -298,6 +299,73 @@ public EndPoint getEndPoint()
...
+        {
+            HostPort localAuthority = ((AbstractConnector)connector).getLocalAuthority();
+            if (localAuthority != null)
+                return localAuthority.getHost();
+        }
+
+        InetSocketAddress local = getLocalAddress();
+        if (local != null)
+            return local.getHostString();
+
```
gregw, line 328: We should never fall through to here.  If there `Connector` is a `NetworkConnector` it should always provide a local authority, otherwise return null.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -298,6 +299,73 @@ public EndPoint getEndPoint()
...
+     * </p>
+     *
+     * @return the local authority port if overridden, or the local address port, or
+     * 0 in the case of no local address (usually seen in connectors not based on IP networking).
+     */
+    public int getLocalPort()
+    {
+        Connector connector = getConnector();
+
+        if (connector instanceof AbstractConnector)
```
gregw: Ditto, test should be for `NetworkConnector` which should always provide the local port

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -298,6 +299,73 @@ public EndPoint getEndPoint()
...
+     * Value can be overridden by {@link AbstractConnector#setLocalAuthority(HostPort)} for
+     * scenarios where Jetty is being an intermediary and the local name
+     * needs to be changed to satisfy public (pre-intermediary) HTTP behaviors
+     * such as absolute-URI creation (eg: Location response header).
+     * </p>
+     *
+     * @return the local name if overridden, or the local address, or
+     * null in the case of no local address (usually seen in connectors not based on IP networking).
+     */
+    public String getLocalName()
```
gregw, line 319: Why not keep the `getLocalAuthority()` here and only convert  to host/port in `Request`

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
gregw: But also let's not convert to Host/Port here.  Might need to leave these as deprecated, but pass the local Authority all the way to the `Request`

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -1050,10 +1039,16 @@ public String getLocalName()
...
-        InetSocketAddress local = _channel.getLocalAddress();
-        return local == null ? 0 : local.getPort();
+        if (_channel != null)
+        {
+            int localPort = _channel.getLocalPort();
+            if (localPort > 0)
+                return localPort;
+            InetSocketAddress local = _channel.getLocalAddress();
+            if (local != null)
+                return local.getPort();
```
gregw: We should not have to do this fall back yet again here.   Either is a local authority (so we have a port) or there is not (and we don't)

gregw (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
joakime: This is not a NetworkConnector specific concept.
We have UnixDomainConnector and LocalConnector that can benefit from setLocalAuthority() and are not network connectors.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
joakime: Removed

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
joakime: Tried this.
Going this route breaks expectations in ForwardRequestCustomizer and DetectorConnection and ProxyEndPoint behaviors, so I reverted it.

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
joakime: We cannot "always return a local authority".
I tried it.
Going this route breaks expectations in ForwardRequestCustomizer and DetectorConnection and ProxyEndPoint behaviors, so I reverted it.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
joakime: See comment above about this feature not being NetworkConnector specific.

gregw:
I can make all but the EmptyHost test pass on an unmodified jetty-9 with the following ammendment to the unit test:
```java
    private CloseableServer startServer(HostPort localAuthority) throws Exception
    {
        Server server = new Server();

        ServerConnector connector;
        if (localAuthority == null)
        {
            connector = new ServerConnector(server);
        }
        else
        {
            InetSocketAddress authorityInetSocketAddress = InetSocketAddress.createUnresolved(localAuthority.getHost(), localAuthority.getPort());
            connector = new ServerConnector(server)
            {
                @Override
                protected ChannelEndPoint newEndPoint(SocketChannel channel, ManagedSelector selectSet, SelectionKey key) throws IOException
                {
                    SocketChannelEndPoint endpoint = new SocketChannelEndPoint(channel, selectSet, key, getScheduler())
                    {
                        @Override
                        public InetSocketAddress getLocalAddress()
                        {
                            return authorityInetSocketAddress;
                        }
                    };
                    endpoint.setIdleTimeout(getIdleTimeout());
                    return endpoint;
                }
            };
        }
        connector.setPort(0);
```

So I'm really reluctant to make big changes in jetty-9 when a simple extention of `ServerConnector` server connector can achieve pretty much the same result.

joakime:
This PR replaced with PR #7316

PR #7292, opened 2021-12-15
Issue #7250 - Correct `HostHeaderCustomizer` logic and add new `RejectMissingAuthorityCustomizer`

### LINKED ISSUE #7250: HostHeaderCustomizer doesn't address empty Host header value with no authority set

joakime:
This is a minimal version of the draft PR #7251

+ Fixes HostHeaderCustomizer to only do Host
  header on non-HTTP/1.1
+ Introduces RejectMissingAuthorityCustomizer
+ Adds modules + xml for both

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HostHeaderCustomizer.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HostHeaderCustomizer.java
@@ -21,6 +21,9 @@ package org.eclipse.jetty.server;
 import java.util.Objects;
 import javax.servlet.http.HttpServletRequest;
 
+import org.eclipse.jetty.http.HttpHeader;
+import org.eclipse.jetty.http.HttpVersion;
+
 /**
  * Customizes requests that lack the {@code Host} header (for example, HTTP 1.0 requests).
  * <p>
@@ -38,15 +41,14 @@ import javax.servlet.http.HttpServletRequest;
  */
 public class HostHeaderCustomizer implements HttpConfiguration.Customizer
 {
-    private final String serverName;
-    private final int serverPort;
+    private final String hostValue;
 
     /**
      * @param serverName the {@code serverName} to set on the request (the {@code serverPort} will not be set)
      */
     public HostHeaderCustomizer(String serverName)
     {
-        this(serverName, 0);
+        this(serverName, -1);
     }
 
     /**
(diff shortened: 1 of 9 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/HostHeaderCustomizer.java
```diff
@@ -55,14 +57,29 @@ public HostHeaderCustomizer(String serverName)
...
+        {
+            if (request.getHttpURI().isAbsolute())
+            {
+                request.getHttpFields().put(HttpHeader.HOST, request.getHttpURI().getAuthority());
+            }
+            else
+            {
+                request.getHttpFields().put(HttpHeader.HOST, hostValue);
+            }
+        }
```
gregw, line 83: Since getHeader is a linear lookup, it is best to avoid doing it if possible, and then to use efficient non string lookup:
```suggestion
        // If not 1.1 and no Host Header
        if (request.getHttpVersion().getVersion() != HttpVersion.HTTP_1_1.getVersion() &&
            !request.getHttpFields().contains(HttpHeader.HOST))
        {
            if (request.getHttpURI().isAbsolute())
            {
                request.getHttpFields().put(HttpHeader.HOST, request.getHttpURI().getAuthority());
            }
            else
            {
                request.getHttpFields().put(HttpHeader.HOST, hostValue);
            }
        }
```

jetty-server/src/main/java/org/eclipse/jetty/server/RejectMissingAuthorityCustomizer.java
```diff
@@ -0,0 +1,84 @@
...
+
+                if (host instanceof HostPortHttpField)
+                {
+                    HostPortHttpField authority = (HostPortHttpField)host;
+                    metadata.getURI().setAuthority(authority.getHost(), authority.getPort());
+                    if (StringUtil.isNotBlank(authority.getHost()))
+                        return true;
+                }
+            }
+        }
```
gregw, line 80: I think this way reduces the number of tests in the hot path:

```suggestion
            if (host != null)
            if (host instanceof HostPortHttpField)
            {
                HostPortHttpField authority = (HostPortHttpField)host;
                metadata.getURI().setAuthority(authority.getHost(), authority.getPort());
                if (StringUtil.isNotBlank(authority.getHost()))
                    return true;
            }
            else if (host != null && StringUtil.isNotBlank(host.getValue()))
            {
                return true;
            }
        }
```

gregw (changes requested):
just a couple of optimizations

joakime:
@gregw would this be good for `jetty-12.0.x-hackathon` branch instead?

joakime:

gregw:
Note keen on any changes to behaviour in jetty-9.  Does the sponsor still require this?

joakime:
Nope.  
The other recent changes for overriding the authority at the connector level were sufficient.

PR #7293, opened 2021-12-16
Avoid NPE when the websocket session is not created

azurvii:

PR #7294, opened 2021-12-16
Fix possible NPE from WebSocketAdapter

lachlan-roberts:
originally from PR #7293 by @azurvii 

Fixes potential NPE from `WebSocketAdapter` if used before `onWebSocketConnect()` is called.

### DIFF
--- a/jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/WebSocketAdapter.java
+++ b/jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/WebSocketAdapter.java
@@ -21,10 +21,11 @@ package org.eclipse.jetty.websocket.api;
 public class WebSocketAdapter implements WebSocketListener
 {
     private volatile Session session;
+    private RemoteEndpoint remote;
 
     public RemoteEndpoint getRemote()
     {
-        return session.getRemote();
+        return remote;
     }
 
     public Session getSession()
@@ -34,7 +35,8 @@ public class WebSocketAdapter implements WebSocketListener
 
     public boolean isConnected()
     {
-        return session.isOpen();
+        Session sess = this.session;
+        return (sess != null) && (sess.isOpen());
     }
 
     public boolean isNotConnected()
@@ -58,6 +60,7 @@ public class WebSocketAdapter implements WebSocketListener
     public void onWebSocketConnect(Session sess)
     {
         this.session = sess;
+        this.remote = sess.getRemote();
     }
 
     @Override


joakime (approved):

joakime:
Do you want this in the 10.0.8 release?

jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/WebSocketAdapter.java
```diff
@@ -21,10 +21,11 @@
 public class WebSocketAdapter implements WebSocketListener
 {
     private volatile Session session;
+    private RemoteEndpoint remote;
 
     public RemoteEndpoint getRemote()
     {
-        return session.getRemote();
+        return remote;
```
azurvii, line 28: IMO `return session == null ? null : session.getRemote();` might be better. This is minor, but could save the `remote` field. And I think the readability is not affected too much either way.

jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/WebSocketAdapter.java
```diff
@@ -34,7 +35,8 @@ public Session getSession()
 
     public boolean isConnected()
     {
-        return session.isOpen();
+        Session sess = this.session;
+        return (sess != null) && (sess.isOpen());
```
azurvii, line 39: I think just `return session != null && session.isOpen();` should be more readable. Creating `sess` here seems redundant.

jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/WebSocketAdapter.java
joakime: It would be more surprising for `getRemote()` to suddenly start returning `null` after it was non-null.

Let the usage of `Remote` trigger the exceptions indicating that you no longer have a valid connection/session/whatnot.

jetty-websocket/websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/WebSocketAdapter.java
joakime: This is standard multi threading coding practice.
The value of `this.session` can actually change between the 2 evaluations (non-null check and isopen).
It could go from non-null to null between the two checks.

azurvii: Ah, I see. Thanks.

PR #7296, opened 2021-12-16
Pass special content to interceptors

### LINKED ISSUE #7281: EOFs are not passed to interceptors any more - shouldn't they?
**Jetty version(s)**
10.x, 11.x

**Java version/vendor**
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.7+10, mixed mode)

**OS type/version**
Windows 10 20H2, 64 bit

**Description**
Starting with the overhaul of `HttpInput` in PR #4556, EOFs are not passed to interceptors any more. The author was obviously also unsure whether this is a good idea, see `AsyncContentProducer.nextTransformedContent()`:
```
// TODO does EOF need to be passed to the interceptors?
```
So custom interceptors cannot react to EOF any more - which we have seen in use by an OSS library based on Jetty 9.4.

If one *would* pass EOF to the interceptors, this would require a small change in `GzipHttpInputInterceptor.readFrom()`, because it tries `getByteBuffer()` without a guard --> `IllegalStateException` in case of `SpecialContent`. For Comparison: In Jetty 9.4, `EofContent` had a `BufferUtil.EMPTY_BUFFER`, so `getByteBuffer()` worked like a charm.

Shouldn't custom interceptors be able to react to EOF nowadays?
This should at least be clarified in a detailed comment instead of the TODO comment.

**How to reproduce?**
Set a breakpoint next to the TODO comment in `AsyncContentProducer.nextTransformedContent()`.

lorban:
Modify the `AsyncContentProducer` accordingly and update `GzipHttpInputInterceptor` to correctly handle the existing `HttpInput.Content` contract.

I have not modified the special contents (error and EOF) to make them return an empty buffer when `getByteBuffer()` is called as IMHO the `Content` contract already is clear on that: calling `getByteBuffer()` when `isSpecial()` returns `true` does throw ISE.

So IMHO if there is some code out there that calls `Content.getByteBuffer()` without first checking the special flag, it is in breach of the contract as it was documented since 10.0.0; the fact that we never passed special content to interceptors should be irrelevant.


lorban:
Note: I have not modified the special contents (error and EOF) to make them return an empty buffer when `getByteBuffer()` is called as IMHO the `Content` contract already is clear on that: calling `getByteBuffer()` when `isSpecial()` returns `true` does throw ISE.

So IMHO if there is some code out there that calls `Content.getByteBuffer()` without first checking the special flag, it is in breach of the contract as it was documented since 10.0.0; the fact that we never passed special content to interceptors should be irrelevant.

gregw (changes requested):
Can we add and example to the javadoc somewhere like:
```java
@Override
public HttpInput.Content readFrom(HttpInput.Content content) {
    if (content.hasContent())
        processContent(content.getByteBuffer());
    if (content.isEof())
        endProcessing();
    return processedContent();
}
```

lorban:
@gregw I've added some javadoc to the `Interceptor` interface, including some code examples and a reference to `GzipHttpInputInterceptor`.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
```diff
@@ -414,6 +414,40 @@ public String toString()
...
+     * which would not do anything with the content besides logging it. A more involved implementation could look like the
+     * following:
+     * <pre>
+     * public HttpInput.Content readFrom(HttpInput.Content content)
+     * {
+     *     if (content.hasContent())
+     *         processContent(content.getByteBuffer());
+     *     if (content.isEof())
+     *         endProcessing();
+     *     return processedContent();
```
sbordet: This example is missing is `isSpecial()` call, which is the one actually used in the implementations.

I would suggest:

```suggestion
     *     if (content.hasContent())
     *         this.processedContent = processContent(content.getByteBuffer());
     *     if (content.isEof())
     *         disposeResources();
     *     return content.isSpecial() ? content : this.processedContent;
```

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
```diff
@@ -414,6 +414,40 @@ public String toString()
...
+     * }
+     * </pre>
+     * Please note that the {@link Content} contract must be respected by implementations, like for instance
+     * that calling {@link Content#getByteBuffer()} when {@link Content#isSpecial()} returns <code>true</code>
+     * throws {@link IllegalStateException}.
+     * </p>
+     * <p>Implementations implementing both this interface and {@link Destroyable} will have their {@link Destroyable#destroy()}
+     * method called when {@link #recycle()} is called.</p>
```
sbordet: I would also add what implementation should do, since `HttpInput.Content` is a `Callback`, when they return a different object, with respect to the callback methods. Should an implementation call them or not?

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
```diff
@@ -334,34 +334,11 @@ public boolean isReady()
...
-                if (LOG.isDebugEnabled())
-                    LOG.debug("raw content is special (with error = {}), returning it {}", _error, this);
-                return _rawContent;
-            }
-
             if (_interceptor != null)
             {
                 if (LOG.isDebugEnabled())
                     LOG.debug("using interceptor to transform raw content {}", this);
                 _transformedContent = intercept();
```
sbordet: If the interceptor returns an error or EOF, is it handled correctly in the block below?
Test cases for these 2 cases?

sbordet (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
lorban: I've added some doc explicitly that it's up to the caller of the interceptor to manage the `Callback` lifecycle.

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
lorban: Fixed and added some more tests, thanks.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
lorban: Okay, updated javadoc.

joakime: Wait, does this mean `content.isSpecial()` is mutually exclusive to `content.hasContent()` always?

You'll never have a scenario where `(content.hasContent() && content.isSpecial())` is true?

winne42: @joakime at least, the JavaDoc on `HttpInput.hasContent()` says so:

> Always false on a special content.

jetty-server/src/test/java/org/eclipse/jetty/server/BlockingContentProducerTest.java
```diff
@@ -116,6 +119,160 @@ public void testBlockingContentProducerNoInterceptorWithError()
...
+            {
+                contentSucceededCount.incrementAndGet();
+            }
+        })));
+        try (AutoLock lock = contentProducer.lock())
+        {
+            contentProducer.setInterceptor(content -> new HttpInput.ErrorContent(new Throwable("testBlockingContentProducerInterceptorGeneratesError interceptor error")));
+
+            HttpInput.Content content1 = contentProducer.nextContent();
+            assertThat(content1.isSpecial(), Matchers.is(true));
```
winne42: `import static org.hamcrest.Matchers.is;` like in `AsyncContentProducerTest` looks nicer

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
```diff
@@ -414,6 +414,52 @@ public String toString()
...
+     *     instances, once {@link ByteBuffer#hasRemaining()} returns <code>false</code> {@link HttpInput} will make sure
+     *     {@link Callback#succeeded()} is called, or {@link Callback#failed(Throwable)} if an error occurs.</li>
+     *     <li>After {@link #readFrom(Content)} is called for the first time, subsequent {@link #readFrom(Content)} calls will
+     *     occur only after the contained byte buffer is empty (see above) or at any time if the returned content was special.</li>
+     *     <li>Once {@link #readFrom(Content)} returned a special content, subsequent calls to {@link #readFrom(Content)} must
+     *     always return the same special content.</li>
+     *     <li>Implementations implementing both this interface and {@link Destroyable} will have their
+     *     {@link Destroyable#destroy()} method called when {@link #recycle()} is called.</li>
+     * </ul>
+     * </p>
```
winne42: brilliant JavaDoc!

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
lorban: @joakime a `Content` is special when it doesn't hold bytes but something more akin to a special event: EOF or an error. So yes, as was pointed out, the `Content` contract explicitly states that `hasContent()` returns `false` when content is special. Other methods also document specific behavior when the content is special.

I've modified the `Content.isSpecial()` javadoc to better explain this.

jetty-server/src/test/java/org/eclipse/jetty/server/BlockingContentProducerTest.java
lorban: Done, thanks.

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
```diff
@@ -316,120 +316,144 @@ public boolean isReady()
...
+
+            // If the interceptor returned the raw content, recycle the raw content now if its empty.
+            if (_transformedContent == _rawContent && _rawContent.isEmpty() && !_rawContent.isSpecial())
             {
                 if (LOG.isDebugEnabled())
-                    LOG.debug("transformed content is not empty {}", this);
+                    LOG.debug("interceptor returned the raw content, recycle the empty raw content now {}", this);
+                _rawContent.succeeded();
+                _rawContent = _transformedContent = null;
+                return;
```
sbordet: Unnecessary `return` statement.

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
```diff
@@ -316,119 +316,144 @@ public boolean isReady()
...
-        while (_transformedContent == null)
-        {
-            if (_interceptor != null)
-            {
-                if (LOG.isDebugEnabled())
-                    LOG.debug("using interceptor to transform raw content {}", this);
-                _transformedContent = intercept();
-            }
-            else
+            if (_transformedContent != null && (_transformedContent.isSpecial() || !_transformedContent.isEmpty()))
```
sbordet: 2nd & 3rd expressions are the opposite of the expressions at 343.
I think this could be simplified via:

```java
if (_transformed != null) 
{
    if (special && !empty) 
    {
        ...
    }
    else 
    {
        // block at 344
    }
}
```

In this way it would be IMHO clearer and less expression repetitions.

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
```diff
@@ -316,119 +316,144 @@ public boolean isReady()
...
+            _transformedContent = intercept();
+
+            // If the interceptor generated a special content, _rawContent must become that special content.
+            if (_transformedContent != null && _transformedContent.isSpecial() && _transformedContent != _rawContent)
+            {
+                if (LOG.isDebugEnabled())
+                    LOG.debug("interceptor generated a special content, _rawContent must become that special content {}", this);
+                _rawContent.succeeded();
+                _rawContent = _transformedContent;
+                return;
```
sbordet, line 392: If the interceptor generated an error content, should you not set `_error` as well?

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
```diff
@@ -316,119 +316,144 @@ public boolean isReady()
...
                 if (LOG.isDebugEnabled())
-                    LOG.debug("transformed content is not empty {}", this);
+                    LOG.debug("interceptor generated a null content, recycling the empty raw content now {}", this);
+                _rawContent.succeeded();
+                _rawContent = null;
+                return;
+            }
+
+            // If the interceptor returned the raw content, recycle the raw content now if its empty.
+            if (_transformedContent == _rawContent && _rawContent.isEmpty() && !_rawContent.isSpecial())
```
sbordet, line 406: This `if` condition basically states that interceptors MUST completely consume the content.

It is what it is, but A) needs to be javadoc'ed and B) needs a safeguard (and a test case) because otherwise the implementation goes into a spin loop!

sbordet (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
sbordet: Remove this and following HTML `p` tags.

sbordet (changes requested):
Please fix redundant `return`.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
lorban: ok.

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
lorban: This is already done in the `intercept()` method.

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
lorban: Ok, done.

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
lorban: Removed it.

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
lorban: Good catch, let me add the necessary javadoc, safeguard check and tests.

sbordet (changes requested):
I see 3 variables here:
- what type of input content (some content, empty content, EOF content, error content)
- what the interceptor does with the input content (full consume, partial consume, no consume)
- what type of output content (null, empty, non-empty, EOF, error, throw)

I would like, if possible, to have tests that cover all combinations of the 3 variables above.
I'd be happy to see tests for the most common cases.
Otherwise I'm good.

sbordet (dismissed):

gregw:
Doh! forced push!!!!! I can't review just the changes since my last review!!!!

gregw (dismissed):

sbordet (approved):

PR #7305, opened 2021-12-20
Experiment with different demandContent contract

gregw:
Currently the `Content.Provider` contract in Jetty-12.0.x is:
```java    
    interface Provider
    {
        Content readContent();
        void demandContent(Runnable onContentAvailable);
    }
```
which can be used in a simple way like:
```java
    try (Blocking.Runnable blocker = _blocker.runnable())
    {
        request.demandContent(blocker);
        blocker.block();
    }
```

There has been some review of this that dislikes the passing on of the `onContentAvailable` with each call to `demandContent`, since this means that `demandContent` cannot be idempotent and that multiple calls to it are meaningful and may result in a pending exception.  However, unlike lower layers, where the callback can be to a well known instance (eg HttpStream calling HttpChannel.onContentAvailable), there is no well know instance to call as the layer above is the application.   Thus this PR experiments with the contract:

```java
    interface Provider
    {
        Content readContent();
        void setOnContentListener(Runnable onContentAvailable);
        void demandContent();
    }
```

The simple blocking usage of this now becomes:
```java
    try (Blocking.Runnable blocker = _blocker.runnable())
    {
        request.setOnContentListener(blocker);
        request.demandContent();
        blocker.block();
    }
    finally
    {
        request.setOnContentListener(null);
    }
```

**Pros:**
 + The call to `demandContent()` is now idempontent, however there currently are no examples of it needing to be idempotent

**Cons:**
 + The lock has to be grabbed more times. For the simple case shown above the lock is grabbed 3 times instead of once.  In general it is grabbed N+2 times, where N is the number of calls to `demandContent`
 + The simple case code is a lot more verbose.
 + If we are to keep the current strict policing of no pending reads when `request.succeeded()` is called, then users have to remember to set the listener back to null after it has been used.
 + If users don't set the listener to null, then we may hold onto lambdas/instances longer than necessary and if succeeded is called we will generate a warning that a read may have been pending.
 + It doesn't really help with the case of two different concerns trying to read content.  If we ever did have two concerns both trying to call `demandContent(Runnable)` with different Runnables, then the same situation can occur with them both calling `setOnContentListener`.... except the outcome can be a lot more confusing, as instead of getting a demand pending exception, the wrong callback may be called.

In short, I see no benefit to this style.   If we really want to have an idempotent `demandContent()` call, then another solution is needed.

Thoughts?

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Content.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Content.java
@@ -245,7 +245,9 @@ public interface Content
     {
         Content readContent();
 
-        void demandContent(Runnable onContentAvailable);
+        void setOnContentListener(Runnable onContentAvailable);
+
+        void demandContent();
     }
 
     // TODO should these static methods be instance methods?   They are not very buffer efficient
@@ -263,9 +265,10 @@ public interface Content
                     Content c = provider.readContent();
                     if (c == null)
                     {
-                        provider.demandContent(this);
+                        provider.demandContent();
                         return;
                     }
+
                     if (c.hasRemaining())
                     {
                         out.copyBuffer(c.getByteBuffer());
@@ -273,6 +276,7 @@ public interface Content
                     }
                     if (c.isLast())
                     {
+                        provider.setOnContentListener(null);
                         if (c instanceof Content.Error)
                             content.failed(((Content.Error)c).getCause());
                         else
@@ -282,6 +286,7 @@ public interface Content
                 }
             }
         };
+        provider.setOnContentListener(onDataAvailable);
         onDataAvailable.run();
     }
 
@@ -312,7 +317,7 @@ public interface Content
                     Content c = provider.readContent();
                     if (c == null)
                     {
-                        provider.demandContent(this);
+                        provider.demandContent();
                         return;
                     }
                     if (c.hasRemaining())
@@ -322,6 +327,7 @@ public interface Content
                     }
                     if (c.isLast())
                     {
+                        provider.setOnContentListener(null);
                         if (c instanceof Content.Error)
                             content.failed(((Content.Error)c).getCause());
                         else
@@ -331,6 +337,7 @@ public interface Content
                 }
             }
         };
+        provider.setOnContentListener(onDataAvailable);
         onDataAvailable.run();
     }
 
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
@@ -212,7 +212,6 @@ public class HttpChannel extends Attributes.Lazy
             if (_request == null)
                 return null;
             Runnable onContent = _request._onContentAvailable;
-            _request._onContentAvailable = null;
             return _serializedInvoker.offer(onContent);
         }
     }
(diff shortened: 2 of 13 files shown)


lorban:
I agree that making the demand listener configurable on the `Provider` interface is not satisfying, but not making the `Provider` contract idempotent is not a need, it's a matter of simplifying the mental model: the fact that you're guaranteed you can call `demandContent()` any time you want without checking anything nor without any ill effects is the true value of idempotency: simplifying the mental model of the API's user.

Ideally, the listener should be set on construction of the `Provider` instance to make it both mandatory and immutable. I'll have a look to see what that would entail.

gregw:
I'm 50:50 on the benefits of `demandContent` being idempotent.   The problem with encouraging an laissez-faire approach to calling demand, is that opens the possibility that an outstanding demand call exists when the request is ultimately succeeeded.   Currently I treat that as an error - ie you can't be reading and then jsut give up and say - the response is OK after all.  That just gives the container the impossible task of trying to consume all and then deciding it to abort when that fails etc.

I think since we are building from the bottom, it is possible for use to be strict and not allow hanging reads and writes to exist when succeeded is called.  At least at this core level.  We probably need to allow them at the servlet level, but we will need to translate them into a strict interpretation of the core API rather than just kicking the can along the road to the network layer to try to resolve.

lorban:
Here is a working prototype of what I have in mind: https://github.com/eclipse/jetty.project/blob/jetty-12.0.x-demand-reducer-api/jetty-server/src/test/java/org/eclipse/jetty/server/HttpConnectionTest.java#L1294-L1317

sbordet:
> The problem with encouraging an laissez-faire approach to calling demand, is that opens the possibility that an outstanding demand call exists when the request is ultimately succeeeded.

I don't see an idempotent `demandContent()` as a laissez-faire approach?
I agree we should be strict about the state at the time `succeeded()` is called, but if you demand and don't consume the demand it's an error no matter if `demandContent()` is idempotent or not.

Idempotency has other benefits (clearer API, similarity with many other APIs we use in Jetty 10 and 11, similarity with reactive streams / `j.u.c.Flow` APIs people is used to, etc.) and no cons while keeping the same semantic, so I would try hard to go there.

gregw:
@lorban are all the tests passing in your branch?  because I really don't understand how it works?   There must be some magic with the blocker class, as I don't see how it is being called in most examples (eg DumpHandler or EchoHandler)?     I think it is important at this stage to not just have one working example... that's the easy part.  You need to have all the examples working, including the failure/error scenarios

gregw:
@sbordet Having an idempotent `demandContent()` is not directly laissez-faire, but it facilitates a laissez-faire approach.  If we are being strict about  state at the time `succeeded()` then we must already be tracking the state and know if we have demanded or not.   Thus it doesn't need to be idempotent.  Conversely, if you are depending on `demandContent()` being idempotent, then that means you don't know the state and are being laissez-faire.

I would not call `Flow` idempotent, because they pass the demand integer which is additive and thus multiple calls to `demand(1)` are different to a single call.

As for clarity of API, I think `demandContent(Runnable onContentAvailable)` cannot be described as un-clear.   It is obvious how to use it.    The issues may be how it might be misused, but currently in the examples we have that is just not an issue.

Can you come up with a test scenario that really needs idempotent `demandContent` so we can use it to evaluate the various approaches?  I too am attracted to the notion of an idempotent demand, but I've just not found any actual examples yet of why it is needed.

gregw:
Latest ideas from a call was to not have the `Content.Provider` interface as part of request, and instead have it obtainable from a method on the request:
```java
interface Request
{
    //...
   Content.Provider getContentProvider(Runnable onContentAvailable);
}
```

This would allow a no args `demandContent()` method.
We need to work out if this really is a getter - ie can it be called multiple times and the Runnable changed?

gregw:
So I'm closing this branch as that latest idea will be played with in a different branch and PR

PR #7314, opened 2021-12-20
Issue #7297 - Removing log4j 1.x

### LINKED ISSUE #7297: Deprecate log4j 1.x support
**Jetty version(s)**
9+

All

**OS type/version**
Alll

**Description**
Log4j 1.x was EOL in August 2015.

While we have no direct dependencies on log4j 1.x within our project, we do have modules across our 3 active versions of Jetty that have log4j 1.x options.

I propose we drop log4j 1.x entirely, modules and all.
If a user wants to continue to use log4j 1.x they have to manually add it, no module convenience anymore.

joakime:

### DIFF
--- a/jetty-util/src/main/config/modules/log4j-impl.mod
+++ b/jetty-util/src/main/config/modules/log4j-impl.mod
@@ -28,5 +28,5 @@ Log4j is released under the Apache 2.0 license.
 http://www.apache.org/licenses/LICENSE-2.0.html
 
 [ini]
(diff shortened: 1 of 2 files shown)


joakime:
@olamy the technique used here is just to move the `log4j.version` and `<dependency>` out of the pom so that we don't constantly trip various vulnerability scans for this deprecated support.

olamy (approved):

PR #7316, opened 2021-12-20
Issue #7277 - Allow `Request.getLocalName()` and `.getLocalPort()` to be overridden (Split authority approach)

### LINKED ISSUE #7277: Allow override of `ServletRequest.getLocalName()` and `.getLocalPort()` in post-intermediary scenarios

joakime:
This is the split authority approach as proposed by @sbordet as an alternate approach for PR #7283


jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
```diff
@@ -161,7 +163,18 @@
     private long _idleTimeout = 30000;
     private String _defaultProtocol;
     private ConnectionFactory _defaultConnectionFactory;
+    /* The name used to link up virtual host configuration to named connectors */
     private String _name;
+    /**
+     * The authority used to define the server uri authority fallback for the connection
+     * (see ServletRequest.getServerName(), and ServletRequest.getServerPort()).
+     */
+    private HostPort _serverUriAuthority;
```
gregw: I think just `_serverAuthority` is a better name.

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
```diff
@@ -298,6 +311,43 @@ public int getAcceptors()
...
+            throw new IllegalStateException(getState());
+
+        if (!authority.hasHost() || !authority.hasPort())
+            throw new IllegalStateException("Local Authority must have both host and port declared");
+        else
+            _localAuthority = authority;
+    }
+
+    @ManagedAttribute("server uri authority")
+    public HostPort getServerUriAuthority()
```
gregw:
```suggestion
    public HostPort getServerAuthority()
```

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
```diff
@@ -161,7 +163,18 @@
...
+    /**
+     * The authority used to define the server uri authority fallback for the connection
+     * (see ServletRequest.getServerName(), and ServletRequest.getServerPort()).
+     */
+    private HostPort _serverUriAuthority;
+    /**
+     * The authority used to define the local authority for the connection
+     * (see ServletRequest.getLocalName(), ServletRequest.getLocalAddr(), and ServletRequest.getLocalPort()).
+     */
+    private HostPort _localAuthority;
```
gregw: is this really a local authority?   Feels more like a local address

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -298,6 +298,64 @@ public EndPoint getEndPoint()
...
+     * scenarios where Jetty is being an intermediary and the local name
+     * needs to be changed to satisfy public (pre-intermediary) HTTP behaviors
+     * such as absolute-URI creation (eg: Location response header).
+     * </p>
+     *
+     * @return the local name if overridden, or the local address, or
+     * null in the case of no local address (usually seen in connectors not based on IP networking).
+
+     */
+    public String getLocalName()
```
gregw, line 320: Why convert to separate name/port methods at the channel level?
I think it would be better if the channel provided `getServerAuthority()` and `getLocalAuthority()` (or `getLocalAddress()` if name change above is accepted).    We only need separate name/port methods in the request where they are defined by servlet API

gregw: Note that the only real reason I see to resolve to name/port methods here is that the existing `getLocalAddress()` returns an `InetSocketAddress`, whilst the new methods return `HostPort`.      I'm thinking it would be simpler to just use `InetSocketAddress` everywhere.

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -1428,20 +1409,24 @@ private String findServerName()
             }
```
gregw, line 1409: Do we have coverage for the code above  here?   I.e. does the `findServerName()` method ever find a `Host` header field? as I'm pretty sure that will have already been copied to the authority in the HttpURI by now.... perhaps if a customizer added a Host header but didn't update the HttpURI?

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -1428,20 +1409,24 @@ private String findServerName()
...
+        {
+            Connector connector = _channel.getConnector();
+            if (connector != null)
+            {
+                HostPort serverAuth = connector.getServerUriAuthority();
+                if (serverAuth != null)
+                {
+                    return formatAddrOrHost(serverAuth.getHost());
+                }
+            }
```
gregw: This feels wrong to be diving down to the connector  here.  Would be better just to ask for `_channel.getServerAuthority()`

gregw (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
joakime: Yep, this block is used if a `Customizer` sets a `Host` header without updating the authority.

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
joakime: Done.

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
joakime: Done.

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
joakime: Changed to SocketAddress

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
```diff
@@ -161,7 +164,18 @@
...
+    /**
+     * The authority used to define the server uri authority fallback for the connection
+     * (see ServletRequest.getServerName(), and ServletRequest.getServerPort()).
+     */
+    private HostPort _serverAuthority;
+    /**
+     * The address used to define the local address for the connection
+     * (see ServletRequest.getLocalName(), ServletRequest.getLocalAddr(), and ServletRequest.getLocalPort()).
+     */
+    private SocketAddress _localAddress;
```
gregw: I think it is sufficient to have javadoc on the associated getters/setters rather than on the private fields

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/MavenServerConnector.java
```diff
@@ -55,6 +58,8 @@
     private String name;
     private int port;
     private long idleTimeout;
+    private HostPort serverAuthority;
+    private SocketAddress localAddress;
```
gregw: Can't you just use these from the delegate connector?

jetty-server/src/main/java/org/eclipse/jetty/server/ServerConnector.java
```diff
@@ -283,6 +285,19 @@ public void setInheritChannel(boolean inheritChannel)
...
+    {
+        _acceptChannel = acceptChannel;
+        _localPort = _acceptChannel.socket().getLocalPort();
+        SocketAddress localAddress = _acceptChannel.socket().getLocalSocketAddress();
+        if (localAddress instanceof InetSocketAddress)
+            _localName = ((InetSocketAddress)localAddress).getHostString();
+        else
+            _localName = localAddress.toString();
+        if (_localPort <= 0)
+            throw new IOException("Server channel not bound");
```
gregw: It is kind of strange that the `localAddress` here is used to initialize the `_localName`, but not the `_localAddress` or `_serverAuthority`.

I know you said there is a reason??? was it the Proxy protocol?   So either we need to make another attempt to initialize these here or we have to document somewhere in the code why not.   If they are not going to be initialized, then we need better javadoc on getLocalAddress to say what null return means

gregw (changes requested):
Getting close.   Mostly just needs some javadoc cleanups.
However, I would like to have one more go at always initializing the _localAddress and if that can't be done, document why

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
joakime: I'll move these to the `Connector` apidoc, as it's kinda useful information not represented there currently.

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/MavenServerConnector.java
joakime: delegate connector?  Let me review that code again, I didn't notice that.

jetty-server/src/main/java/org/eclipse/jetty/server/ServerConnector.java
joakime: I reverted this.  It's left over from when we attempted to default the localAddress during doStart().
That (the defaulting of localAddress at doStart) couldn't be done for the proxy cases.

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/MavenServerConnector.java
joakime: Fixed

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
joakime: Moved

jetty-server/src/main/java/org/eclipse/jetty/server/Connector.java
```diff
@@ -102,4 +104,48 @@
...
+     * </p>
+     *
+     * @param localAddress the address to use for host/addr/port, or null to reset to default behavior
+     */
+    void setLocalAddress(SocketAddress localAddress);
+
+    /**
+     * Get the optional Server URI authority default
+     *
+     * @return Returns the connection server authority (name/port).
```
gregw:
```suggestion
     * @return Returns the connection server authority (name/port) or null.
```

I really don't like this "or null".

jetty-server/src/main/java/org/eclipse/jetty/server/Connector.java
```diff
@@ -102,4 +104,48 @@
      * @return The connector name or null.
      */
     String getName();
+
+    /**
+     * Get the Local Address of the connection
+     *
+     * @return Returns the connection local address.
```
gregw:
```suggestion
     * @return Returns the connection local address or null.
```
I really don't like the or null.

gregw (changes requested):
I really don't like the fact that in most situations `getLocalAddress` and `getServerAuthority` are going to return null.   If I see those methods on the connector, they should provide valid values if the connector is running.

Currently the semantic of the methods is `getLocalAddressDefault()` and `getServerAuthorityDefault()` or something like that... but "Default" is not the right word, nor is "Override", as any verb we put here is describing an algorithm that is implemented outside of this instance and is just assuming how these methods are used.     So it reeks of bad design the fact that we can't get the name right.   Currently these getters/setters are entirely unrelated to the behaviour of the connector and are essentially fields for an external algorithm.

Can you describe again the use-case that prevents us making these real methods that actually returns a dynamically determined local address and authority (if none have been set)?

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConfiguration.java
```diff
@@ -78,6 +81,9 @@
     private boolean _notifyRemoteAsyncErrors = true;
     private boolean _relativeRedirectAllowed;
 
+    private HostPort _serverAuthority;
+    private SocketAddress _localAddress;
```
gregw, line 83: We need to either add these fields to the copy constructor OR well document that they are not copied by the copy constructor.

I'm probably happy with either approach... whatever works best in the XML and makes more sense for configuring http and https connectors.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConfiguration.java
```diff
@@ -662,6 +668,71 @@ public boolean isRelativeRedirectAllowed()
         return _relativeRedirectAllowed;
     }
 
+    /**
+     * Get the Local Address of the connection
+     *
+     * @return Returns the connection local address.
+     */
+    @ManagedAttribute("Local address override")
+    public SocketAddress getLocalAddress()
```
gregw:
```suggestion
    /**
     * Get the SocketAddress override to be reported as the local address of all connection
     *
     * @return Returns the connection local address override or null.
     */
    @ManagedAttribute("Local address override")
    public SocketAddress getLocalAddress()
```

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConfiguration.java
```diff
@@ -662,6 +668,71 @@ public boolean isRelativeRedirectAllowed()
...
+        _localAddress = localAddress;
+    }
+
+    /**
+     * Get the optional Server URI authority default
+     *
+     * @return Returns the connection server authority (name/port).
+     */
+    @ManagedAttribute("Server authority override")
+    public HostPort getServerAuthority()
```
gregw:
```suggestion
     * Get the Server URI authority default to be used if no authority is provided by a request.
     *
     * @return Returns the connection server authority (name/port) or null
     */
    @ManagedAttribute("Server authority default")
    public HostPort getServerAuthority()
```

gregw (dismissed):
Much better.
Just a couple of javadoc tweaks and the config copy constructor to resolve (one way or the other). 

Actually we should probably also define some properties in the mod files and update the xml that sets up our HttpConfiguration to test that it is usable as we think it should be.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -298,8 +299,73 @@ public EndPoint getEndPoint()
...
     }
 
+    /**
+     * <p>Return the Local Name of the connected channel.</p>
+     *
+     * <p>
+     * This is the name after the connector is bound and the connection is accepted.
+     * The value is typically the local address and optionally the local host name
+     * if the local address is unavailable.
+     * </p>
```
sbordet, line 308: This javadoc paragraph is incorrect, the address is never returned, only the host name.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConfiguration.java
sbordet: It's not the _URI_ authority, just the `server authority`.
The description of `@ManagedAttribute` should be "The server authority if none provided by requests".

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConfiguration.java
```diff
@@ -662,6 +668,71 @@ public boolean isRelativeRedirectAllowed()
...
+    {
+        return _serverAuthority;
+    }
+
+    /**
+     * <p>
+     * Specify the connection server uri authority (name/port) used within application API layer
```
sbordet: Here again, remove "uri" and add that's used if no authority is provided by the request.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConfiguration.java
```diff
@@ -662,6 +668,71 @@ public boolean isRelativeRedirectAllowed()
...
+     * </p>
+     *
+     * @param authority the authority host (and optional port), or null to reset to default behavior
+     */
+    public void setServerAuthority(HostPort authority)
+    {
+        if (authority == null)
+            _serverAuthority = null;
+        else if (!authority.hasHost())
+            throw new IllegalStateException("Server URI Authority must have host declared");
```
sbordet: Remove "URI".

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConfiguration.java
```diff
@@ -662,6 +668,71 @@ public boolean isRelativeRedirectAllowed()
...
+     * and {@code ServletRequest.getLocalPort()}.
+     * </p>
+     *
+     * @param localAddress the address to use for host/addr/port, or null to reset to default behavior
+     */
+    public void setLocalAddress(SocketAddress localAddress)
+    {
+        Objects.requireNonNull(localAddress, "Local Address");
+
+        _localAddress = localAddress;
```
sbordet:
```suggestion
        _localAddress = Objects.requireNonNull(localAddress);
```

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConfiguration.java
sbordet: I would go for full copy of these fields, like all the others.

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -1428,21 +1409,25 @@ private String findServerName()
...
+            HttpConfiguration httpConfiguration = _channel.getHttpConfiguration();
+            if (httpConfiguration != null)
+            {
+                HostPort serverAuth = httpConfiguration.getServerAuthority();
+                if (serverAuth != null)
+                {
+                    return formatAddrOrHost(serverAuth.getHost());
+                }
+            }
+        }
```
sbordet, line 1417: This block should be moved to `HttpChannel.getServerAuthority()`, to make it similar to `HttpChannel.getLocalName()`.

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -1474,19 +1460,29 @@ private int findServerPort()
...
-            return getLocalPort();
+        {
+            HttpConfiguration httpConfiguration = _channel.getHttpConfiguration();
+            if (httpConfiguration != null)
+            {
+                HostPort serverAuth = httpConfiguration.getServerAuthority();
+                if (serverAuth != null)
+                    return serverAuth.getPort();
+            }
+        }
```
sbordet, line 1472: Move this block to `HttpChannel.getServerAuthority()`.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -298,8 +299,73 @@ public EndPoint getEndPoint()
...
+     * <p>
+     * Value can be overridden by {@link HttpConfiguration#setLocalAddress(SocketAddress)} for
+     * scenarios where Jetty is being an intermediary and the local name
+     * needs to be changed to satisfy public (pre-intermediary) HTTP behaviors
+     * such as absolute-URI creation (eg: Location response header).
+     * </p>
```
sbordet, line 316: This paragraph is incorrect. It's not used for URI creation (and should not be).

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -298,8 +299,73 @@ public EndPoint getEndPoint()
...
+     *
+     * <p>
+     * This is the port the connector is bound to and is accepting connection on.
+     * </p>
+     * <p>
+     * Value can be overridden by {@link HttpConfiguration#setLocalAddress(SocketAddress)} for
+     * scenarios where Jetty is being an intermediary and the local port
+     * needs to be changed to satisfy public (pre-intermediary) HTTP behaviors
+     * such as absolute-URI creation (eg: Location response header).
+     * </p>
```
sbordet, line 350: This paragraph is incorrect. It's not used for URI creation (and should not be).

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -298,8 +299,73 @@ public EndPoint getEndPoint()
...
+     *
+     * @return the local authority port if overridden, or the local address port, or
```
sbordet: Remove the part on "authority".

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -298,8 +299,73 @@ public EndPoint getEndPoint()
...
+     * </p>
+     *
+     * @return the local name if overridden, or the local address, or
+     * null in the case of no local address (usually seen in connectors not based on IP networking).
+
+     */
+    public String getLocalName()
+    {
+        SocketAddress localAddress = getHttpConfiguration().getLocalAddress();
+        if (localAddress != null)
```
sbordet: Null check can be removed, as the `instanceof` below would take care of it implicitly.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConfiguration.java
sbordet: Why can't it be set to `null`? I think it should, as the server authority can.

sbordet (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -298,8 +300,89 @@ public EndPoint getEndPoint()
...
+                return ((InetSocketAddress)localAddress).getPort();
+        }
+
+        InetSocketAddress local = getLocalAddress();
+        return local == null ? 0 : local.getPort();
+    }
+
     public InetSocketAddress getLocalAddress()
     {
+        SocketAddress localAddress = getHttpConfiguration().getLocalAddress();
```
sbordet: Null check for HttpConfiguration?

sbordet (approved):

PR #7335, opened 2021-12-23
Issue #7299 - Enabling the logging-logback module prevents eclipse re…

### LINKED ISSUE #7299: Enabling the logging-logback module prevents eclipse remote debugging
**Jetty version(s)**
9.4.43.v20210629

openjdk version "1.8.0_292"

**OS type/version**
Ubuntu 18.04 (x64)

**eclipse version**
4.6 neon

**Description**
Attempting to debug a web application using eclipse remote debugging will fail if the logging-logback module of the jetty server is enabled. The eclipse debugger will see and connect to the debug port, but no breakpoints will be hit and execution will never pause.

**How to reproduce?**

1. Enable the logback module via the command line (java -jar ../start.jar --add-to-start=logging-logback)
2. Add the dependencies to the project pom 
 ```
       <dependency>
	    <groupId>org.slf4j</groupId>
	    <artifactId>slf4j-api</artifactId>
	    <version>1.7.30</version>
	</dependency>

	<dependency>
	    <groupId>ch.qos.logback</groupId>
	    <artifactId>logback-classic</artifactId>
	    <version>1.2.6</version>
	</dependency>
	
	<dependency>
	    <groupId>ch.qos.logback</groupId>
	    <artifactId>logback-core</artifactId>
	    <version>1.2.6</version>
	</dependency>
```
3. Build and deploy the web app.
4. Start the server with a debug port using the args '-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=9999'
5. Create and launch a debug configuration for a Remote Java Application with properties **host:localhost**  and **port:9999**
6. Wait for the debugger to connect and the suspended server startup to complete.
7. Add and eclipse breakpoint in a web service method.
8. Navigate to the web service via a browser.

Result: The web method will execute and return results. None of the eclipse breakpoints will be hit. If steps 3-8 are repeated with logging-logback disabled in start.ini, the breakpoints will be hit and execution will pause as expected.

sbordet:
…mote debugging.

Added documentation as discussed in the issue.

### DIFF
--- a/documentation/jetty-documentation/src/main/asciidoc/operations-guide/logging/logging-server.adoc
+++ b/documentation/jetty-documentation/src/main/asciidoc/operations-guide/logging/logging-server.adoc
@@ -58,8 +58,6 @@ Specifies the timezone ID (such as `PST`, or `America/Los_Angeles` or `GMT-8:00`
 The empty string specifies the `UTC` timezone.
 Default value is the local timezone.
 
-// TODO: add a xref to the console-capture module.
-
 The `logging-jetty` Jetty module, enabled transitively, provides the configuration file `$JETTY_BASE/resources/jetty-logging.properties` to configure the logging levels, for example:
 
 ----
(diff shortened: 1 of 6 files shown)


pbtura:
All the changes look reasonable. I think it might be worthwhile to also make a mention of this issue in the documentation for adding a logging implementation. Since adding a non-default logger is what creates the need for special handling of debugging, it seems like it would make sense to warn users about potential issues there.

Also, it might be worth updating the documentation for 9.x since this problem also affects that version.

joakime (dismissed):

olamy (dismissed):

sbordet:
@pbtura In Jetty 10 we don't fork a new JVM so it's not necessary to add that to the documentation.
Docs for 9.4.x are left untouched as we're focusing on Jetty 10+.
Thanks for your feedback!

PR #7350, opened 2022-01-04
Issue #3551 - Document correct Conscrypt usage

### LINKED ISSUE #3551: Document correct Conscrypt usage
When using the `Conscrypt` provider, a few things must be done to make sure it's used properly.

#### On the server

- Don't use `Security.addProvider(new OpenSSLProvider())`.
Instead use `Security.insertProviderAt(new OpenSSLProvider(), 1)`.
This is because the Conscrypt provider, internally, reaches for other component factories (such as `TrustManagerFactory`, etc.) and we want to default to the Conscrypt ones.
If the Conscrypt provider is not in the first position, the default for those component factories will be the OpenJDK provider.

#### On the client

- Use `Security.insertProviderAt(new OpenSSLProvider(), 1)` (see above).
- Because of google/conscrypt#656, when using explicitly the `HTTPS` endpointIdentificationAlgorithm, you must call `HttpsURLConnection.setDefaultHostnameVerifier(...)` or Conscrypt's 2.1 `Conscrypt.setHostnameVerifier(...)`.

lachlan-roberts:
closes #3551

- add some documentation about Conscrypt in the programming guide
    - document that it is necessary to set the `HostnameVerifier` when using `HTTPS` endpointIdentificationAlgorithm on the client.
    - That you should use `Security.insertProviderAt(new OpenSSLProvider(), 1)` instead of `Security.addProvider(new OpenSSLProvider())` when adding conscrypt for the server.

- change the conscrypt.xml file to reflect the documented way of adding the `OpenSSLProvider`.


documentation/jetty-documentation/src/main/asciidoc/programming-guide/client/client.adoc
```diff
@@ -35,3 +35,4 @@ include::http/client-http.adoc[]
 include::http2/client-http2.adoc[]
 include::http3/client-http3.adoc[]
 include::websocket/client-websocket.adoc[]
+include::conscrypt/client-conscrypt.adoc[]
```
sbordet, line 38: I don't think this deserves a whole section.
There already is a section about configuring TLS on the client, can you move this additional documentation there, section `pg-client-http-configuration-tls`?

documentation/jetty-documentation/src/main/asciidoc/programming-guide/client/conscrypt/client-conscrypt.adoc
```diff
@@ -0,0 +1,22 @@
...
+
+[[pg-client-conscrypt]]
+=== Conscrypt Java Security Provider
+
+Conscrypt is a Java Secure Socket Extension (JSSE) can be used with Jetty instead of the default provider from the JVM to provide TLS and the ALPN extension.
+
+See relevant information xref:pg-server-conscrypt[Conscrypt Server Configuration].
+
+In order to use the `HTTPS` endpointIdentificationAlgorithm, you must also set a `HostnameVerifier`, either via `HttpsURLConnection.setDefaultHostnameVerifier()` or `Conscrypt.setHostnameVerifier` / `Conscrypt.setDefaultHostnameVerifier()`.
+This is because Conscrypt implementation defers directly to the hostname verifier unlike OpenJDK which first delegates to an internal verification and only if that fails tries the hostname verifier.
```
sbordet, line 22: Are you sure? I ask because I just ran `ConscryptHTTP2ClientTest` removing the line `Conscrypt.setDefaultHostnameVerifier(...)` and it works fine, so apparently it is not necessary?

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/server.adoc
```diff
@@ -36,3 +36,4 @@ include::http3/server-http3.adoc[]
 include::sessions/sessions.adoc[]
 include::websocket/server-websocket.adoc[]
 include::server-io-arch.adoc[]
+include::conscrypt/server-conscrypt.adoc[]
```
sbordet, line 39: Same as for the client, I think this should be moved to the `pg-server-http-connector-protocol` section, perhaps as a child, with all the subsections that configure TLS referring to this new child.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/conscrypt/server-conscrypt.adoc
```diff
@@ -0,0 +1,25 @@
...
+//
+
+[[pg-server-conscrypt]]
+=== Conscrypt Java Security Provider
+
+Conscrypt is a Java Secure Socket Extension (JSSE) can be used with Jetty instead of the default provider from the JVM to provide TLS and the ALPN extension.
+
+Google maintains BoringSSL as a fork of the OpenSSL project that is used in their Chrome and Android products, thus it is an excellent well maintained library with good security and performance. Google have also built on the work of the Netty project to develop Conscrypt as the Java library that maps the native BoringSSL API to be a compliant JSSE SecurityProvider.
+
+When configuring Conscrypt for the Server, instantiate an instance of Conscrypt’s `OpenSSLProvider` to be registered with the static method on the `java.security.Security` class.
```
sbordet, line 21: Make a code example, rather than describing the code in words.

sbordet (changes requested):

documentation/jetty-documentation/src/main/asciidoc/programming-guide/client/conscrypt/client-conscrypt.adoc
lachlan-roberts: That was the information provided in the issue you linked (https://github.com/google/conscrypt/issues/656). 

But I can't seem to replicate the problem from that issue at all, even using the exact same snippet of code didn't fail for me. So maybe something has changed since then. I couldn't see that the hostname verifier was ever called at all.

PR #7360, opened 2022-01-06
Issue #7351 - large WebSocket payloads with permessage-deflate hang

### LINKED ISSUE #7351: Large WebSocket payloads with permessage-deflate hang on 10.0.7
**Jetty version(s)**

10.0.7

openjdk version "11.0.13" 2021-10-19 LTS

**OS type/version**

MacOS/Linux

**Description**

Starting from 10.0.7, large WebSocket messages hang when the WebSocket connection is initiated with the `permessage-deflate` extension. The "hanging" behaviour is only observed starting from 10.0.7, and the code works correctly on other 10.0.x and 9.4.x versions. The [`jetty-10.0.7-websocket-permessage-deflate`](https://github.com/SerCeMan/jetty-10.0.7-websocket-permessage-deflate) can help to reproduce the issue, the test that sends a single payload never completes. 

```java
    @Test
    void testDeflate() throws Exception {
        var client = new WebSocketClient();
        client.setMaxBinaryMessageSize(5 * 1024 * 1024);
        client.start();
        var latch = new CountDownLatch(1);
        var req = new ClientUpgradeRequest();
        req.addExtensions("permessage-deflate");
        var session = client.connect(new Socket(latch), URI.create("ws://localhost:8083/ws"), req).get();
        latch.await(3, TimeUnit.SECONDS);
        session.close();
    }
```

Removing the extension from the test, or changing the jetty version from `10.0.7` to `10.0.6` in [build.gradle](https://github.com/SerCeMan/jetty-10.0.7-websocket-permessage-deflate/blob/master/build.gradle#L16-L17) makes the test pass. 

From what I understand, the test fails on `10.0.7` when the message size exceeds `PerMessageDeflateExtension#DEFAULT_BUF_SIZE`.

**How to reproduce?**

The reproducer project: https://github.com/SerCeMan/jetty-10.0.7-websocket-permessage-deflate.

lachlan-roberts:
## Issue #7351

PerMessageDeflate cannot use iterating callback for incoming frame fragmentation as we must wait for more demand before inflating the next frame.

This PR adds a mechanism to override the demand in the `WebSocketCoreSession` using a stack. Extensions can use this to be notified once another frame has been requested.


jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/WebSocketCoreSession.java
```diff
@@ -422,7 +428,31 @@ public void demand(long n)
...
+        LongConsumer consumer = null;
+        try (AutoLock l = _lock.lock())
+        {
+            if (!_demandHandlers.isEmpty())
+                consumer = _demandHandlers.pop();
+        }
+        return consumer;
+    }
+
+    public void pushDemandHandler(LongConsumer demand)
```
gregw: I'm not opposed to a demand interception architecture.... as ultimately that is what we are doing in the jetty-12 HttpStream contract.   However, at the very least it needs a lot of javadoc to explain how to use it and why it is needed.

But I'm also very cautious that it doesn't go hand in hand with the interception of the actual content... again as it is done in the jetty-12 HttpStream contract.

gregw (changes requested):
OK so if I'm reading this right, the basic idea is to intercept demand so that if you have a large frame that is being inflated then demand is satisfied by inflating more data rather than demanding another frame.

This basic idea is sound.... but is there a better way of intercepting demand rather than pushing down a handler onto the core Session?  That feels like it is making the session need to worry about concerns that are not it's core business?

It feels more like there is a missing layer of abstraction that would allow the demand to be filtered by the same entity doing the inflation.

This needs careful consideration before merging.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/WebSocketCoreSession.java
gregw: I think you need to look at adding demand interception to the Extension class, perhaps as an optional interface extension.   Then WebSocketCoreSession, which it builds its stack of extentions that intercept incoming frames, it can also build a stack of demand interceptors (handlers), that will always be there - not added and removed.    Thus the inflate extension will always intercept demand and only pass it on if it has not more bytes to inflate into a synthetic frame.

lachlan-roberts: > at the very least it needs a lot of javadoc to explain how to use it and why it is needed.

Sure I can add some javadoc. The `DemandHandler` is internal mechanism and is not exposed through the `CoreSession` interface. So the only way to have access to this is to downcast the `CoreSession` to an internal class.

> I think you need to look at adding demand interception to the Extension class, perhaps as an optional interface extension.

This is not possible, it needs to be on the WebSocketCoreSession or the WebSocketConnection, the application has no interaction with the Extensions it only calls demand on the session. I originally had a permanent `DemandListener` to intercept demand, so that is possible but it still needs to be registered through the `WebSocketCoreSession`.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/WebSocketCoreSession.java
```diff
@@ -422,7 +428,35 @@ public void demand(long n)
...
+
+    public LongConsumer popDemandHandler()
+    {
+        LongConsumer consumer = null;
+        try (AutoLock l = _lock.lock())
+        {
+            if (!_demandHandlers.isEmpty())
+                consumer = _demandHandlers.pop();
+        }
+        return consumer;
```
sbordet:
```suggestion
        try (AutoLock l = _lock.lock())
        {
            return _demandHandlers.poll();
        }
```

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/PerMessageDeflateExtension.java
```diff
@@ -438,22 +465,72 @@ private boolean inflate(Callback callback) throws DataFormatException
...
+            _frame = null;
+            _framePayload = null;
+            _frameCallback = null;
+            _tailBytes = false;
+        }
+
+        @Override
+        protected void onCompleteFailure(Throwable cause)
+        {
+            _failure.set(cause);
```
sbordet: This should be a `compareAndSet()` to keep only the first failure and to make sure to notify only once?

jetty-websocket/websocket-jetty-tests/src/test/java/org/eclipse/jetty/websocket/tests/LargeDeflateTest.java
```diff
@@ -0,0 +1,101 @@
...
+        assertThat(message, is(sentMessage));
+    }
+
+    private static ByteBuffer largePayloads()
+    {
+        var bytes = new byte[4 * 1024 * 1024];
+        new Random(42).nextBytes(bytes);
+        return BufferUtil.toBuffer(bytes);
+    }
+}
```
sbordet, line 101: Can you please write also a test with explicit demand _and_ the permessage deflate extension?

sbordet (changes requested):

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/WebSocketCoreSession.java
sbordet: > Then WebSocketCoreSession, which it builds its stack of extentions that intercept incoming frames, it can also build a stack of demand interceptors (handlers), that will always be there - not added and removed. Thus the inflate extension will always intercept demand and only pass it on if it has not more bytes to inflate into a synthetic frame.

What's currently in this PR is a minimal, implementation-only, fix for the issue.
While perhaps a more generic (and publicly available) mechanism would be desired, I think for this release we should stick with this minimal change, as what you propose is a larger change that still feels like an afterthought (introducing new methods to `Extension` that should really be a new parameter in `IncomingFrames.onFrame()` instead, etc.).
I like better a re-design for Jetty 12 and a minimal change now.
Then if we really need to add a method to `Extension` we can discuss it more carefully for a future Jetty 10 release.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/PerMessageDeflateExtension.java
lachlan-roberts: `onCompleteFailure` will only ever be called once, so we already won't notify more than once. The only time `_failure` will already be set is when `failFlusher(Throwable t)` has been called, and if the iterating callback was processing at the time of `failFlusher` then process could also fail for a different reason.

So if we want to cover the case where it has failed for two different reasons, then I could change this to:
```java
Throwable suppressed = _failure.getAndSet(cause);
if (suppressed != null && suppressed != cause)
    cause.addSuppressed(suppressed);
```

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/WebSocketCoreSession.java
gregw: @sbordet the current PR doesn't need to be a perfect fix and we can land on a final API in 12.  But it should not be a step in the wrong direction.   I think the approach here is just wrong: a dynamic stack of interceptors and explicit demand bypass API!        We should at least fix it in a style that is likely to be similar to what we ultimately implement - ie a fixed set of demand interceptors that is initialized once the Extentions are known, that will always be called for all demand calls.   I see no reason not to try to do it right now.

gregw:
I mocked up how I think we could do the demand interception in https://github.com/eclipse/jetty.project/tree/jetty-10.0.x-ws-demand-extension
My main criteria is that the demand interception is not dynamic for the connection, but setup once with the extension stack.   My mockup has a new public API, but that probably should be package protected to keep this non-public for now.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/DemandChain.java
```diff
@@ -0,0 +1,25 @@
...
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.websocket.core.internal;
+
+import java.util.function.LongConsumer;
+
+public interface DemandChain
```
sbordet, line 25: Javadocs.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/ExtensionStack.java
```diff
@@ -210,6 +213,14 @@ public void negotiate(List<ExtensionConfig> offeredConfigs, List<ExtensionConfig
                 Extension ext = exts.next();
                 ext.setNextOutgoingFrames(outgoing);
                 outgoing = ext;
+
+                if (ext instanceof DemandChain)
+                {
+                    DemandChain demandingExtension = (DemandChain)ext;
+                    if (demandChain != null)
```
sbordet: `demandChain` cannot be `null`, right?

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/PerMessageDeflateExtension.java
```diff
@@ -335,85 +354,131 @@ else if (_frame.isFin())
...
+        public void setNextDemand(LongConsumer nextDemand)
+        {
+            _nextDemand = nextDemand;
+        }
+
+        public void demand(long n)
+        {
+            if (n <= 0)
+                throw new IllegalArgumentException("Demand must be positive");
+            _demand.addAndGet(n);
```
sbordet: Cap the demand with `MathUtils`.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/PerMessageDeflateExtension.java
```diff
@@ -236,6 +240,21 @@ protected void nextOutgoingFrame(Frame frame, Callback callback, boolean batch)
...
+        incomingFlusher.setNextDemand(nextDemand);
+    }
+
+    @Override
+    public void demand(long n)
+    {
+        if (n <= 0)
+            throw new IllegalArgumentException("Demand must be positive");
+        incomingFlusher._demand.addAndGet(n);
+        incomingFlusher.iterate();
```
sbordet:
```suggestion
        incomingFlusher.demand(n);
```

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/PerMessageDeflateExtension.java
```diff
@@ -335,85 +354,131 @@ else if (_frame.isFin())
...
+            return Action.IDLE;
         }
 
-        private boolean inflate(Callback callback) throws DataFormatException
+        private void inflate() throws DataFormatException
         {
             // Get a buffer for the inflated payload.
             long maxFrameSize = getConfiguration().getMaxFrameSize();
             int bufferSize = (maxFrameSize <= 0) ? inflateBufferSize : (int)Math.min(maxFrameSize, inflateBufferSize);
             final ByteBuffer payload = getBufferPool().acquire(bufferSize, false);
```
sbordet: Remove `final`.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/PerMessageDeflateExtension.java
```diff
@@ -438,22 +503,65 @@ private boolean inflate(Callback callback) throws DataFormatException
...
+                    frameCallback.succeeded();
+            }, t ->
+            {
+                // The error needs to be forwarded to the CoreSession if callback is failed.
+                getBufferPool().release(payload);
+                failFlusher(t);
+                coreSession.processHandlerError(t, NOOP);
+            });
+            _demand.decrementAndGet();
+            nextIncomingFrame(chunk, payloadCallback);
```
sbordet: If this method throws, the `payload` is not released back into the buffer pool?
I think you should move the payload release in `onCompleteFailure()`.

jetty-websocket/websocket-jetty-tests/src/test/java/org/eclipse/jetty/websocket/tests/LargeDeflateTest.java
```diff
@@ -0,0 +1,101 @@
...
+import org.eclipse.jetty.websocket.server.config.JettyWebSocketServletContainerInitializer;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class LargeDeflateTest
```
sbordet, line 39: Can you move this test to `PermessageDeflateDemandTest`? There is no need to have 1 class per test.

jetty-websocket/websocket-jetty-tests/src/test/java/org/eclipse/jetty/websocket/tests/PermessageDeflateBufferTest.java
```diff
@@ -124,5 +135,26 @@ public void testPermessageDeflateAggregation() throws Exception
...
+
+        ByteBuffer message = randomBytes(1024);
+        session.setMaxFrameSize(64);
+        session.getRemote().sendBytes(message);
+        assertThat(socket.binaryMessages.poll(5, TimeUnit.SECONDS), equalTo(message));
+
+        session.close();
+        assertTrue(socket.closeLatch.await(5, TimeUnit.SECONDS));
+        assertThat(socket.closeCode, equalTo(StatusCode.NORMAL));
     }
```
sbordet, line 159: Please add a test that verifies that the payload is released if an extension throws or fails the callback.

sbordet (changes requested):

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/PerMessageDeflateExtension.java
lachlan-roberts: `nextIncomingFrame` is not supposed to throw that's why we give it a callback. When we eventually call application code we have a try/catch so if the application throws we fail the callback. So if `nextIncomingFrame` throws it is a bug.

If we release in `onCompleteFailure()` then we have to keep a full list of all the payload callbacks in the flusher and remove them when the callback is succeeded (the payloads might not be succeeded before next iteration of the flusher). I think this is not a good idea.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/PerMessageDeflateExtension.java
lachlan-roberts: Ok, but will use `Math.addExact` instead as its probably more what we want.

You don't really want to cap the demand because then its out of sync with what the application is expecting, although this is not likely to be a problem with `long`.

jetty-websocket/websocket-jetty-tests/src/test/java/org/eclipse/jetty/websocket/tests/LargeDeflateTest.java
lachlan-roberts: `PermessageDeflateDemandTest` is testing the core API and this is using the Jetty WS API so is also testing the message aggregation through the message sinks.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/DemandChain.java
lachlan-roberts: done

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/ExtensionStack.java
lachlan-roberts: fixed

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/DemandingFlusher.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.BiConsumer;
+import java.util.function.LongConsumer;
+
+import org.eclipse.jetty.util.Callback;
+import org.eclipse.jetty.util.CountingCallback;
+import org.eclipse.jetty.util.IteratingCallback;
+import org.eclipse.jetty.websocket.core.Frame;
+
+public abstract class DemandingFlusher extends IteratingCallback implements DemandChain
```
sbordet, line 39: Needs javadocs both for the class and the methods.

jetty-websocket/websocket-jetty-tests/src/test/java/org/eclipse/jetty/websocket/tests/PermessageDeflateBufferTest.java
sbordet: Please implement what above.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/DemandingFlusher.java
gregw, line 39: javadoc

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/DemandingFlusher.java
```diff
@@ -0,0 +1,128 @@
...
+    public void failFlusher(Throwable t)
+    {
+        if (_failure.compareAndSet(null, t))
+        {
+            failed(t);
+            iterate();
+        }
+    }
+
+    public void forwardFrame(Frame frame, Callback callback)
```
gregw: I'm not sure "forward" is the right verb, as it implies only frames received can be forwarded.   Perhaps `emitFrame` ?

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/DemandingFlusher.java
```diff
@@ -0,0 +1,128 @@
...
+    private Callback _callback;
+    private boolean _needContent = true;
+    private boolean _first = true;
+
+    public DemandingFlusher(BiConsumer<Frame, Callback> forwardFrame)
+    {
+        _forwardFrame = forwardFrame;
+    }
+
+    protected abstract boolean handle(Frame frame, Callback callback, boolean first);
```
gregw, line 73: javadoc

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/DemandingFlusher.java
```diff
@@ -0,0 +1,128 @@
...
+        if (_failure.compareAndSet(null, t))
+        {
+            failed(t);
+            iterate();
+        }
+    }
+
+    public void forwardFrame(Frame frame, Callback callback)
+    {
+        _demand.decrementAndGet();
```
gregw: This needs to check that demand does not go negative and if so throw an ISE.
Currently there is nothing stopping a usage of this class calling forwardFrame more than it should.    In my `IteratingDemand` class I prevent that by returning the frame to be forwarded, so only a single frame can be produced each iteration and then demand is checked again.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/FragmentExtension.java
```diff
@@ -69,4 +88,68 @@ public void init(ExtensionConfig config, WebSocketComponents components)
...
+            Callback payloadCallback = Callback.from(() ->
+            {
+                if (finished)
+                    callback.succeeded();
+            }, t ->
+            {
+                // This is wrapped with CountingCallback so will only be failed once.
+                callback.failed(t);
+                failFlusher(t);
+            });
```
gregw, line 149: This feels like an unnecessary capturing lamda.  The only thing it is capturing is the callback, which is actually a field on the `DemandingFlusher`, so really this should just be able to be a reused callback on the instance.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/DemandingFlusher.java
```diff
@@ -0,0 +1,128 @@
...
+    @Override
+    public void setNextDemand(LongConsumer nextDemand)
+    {
+        _nextDemand = nextDemand;
+    }
+
+    public void onFrame(Frame frame, Callback callback)
+    {
+        _frame = frame;
+        _callback = new CountingCallback(callback, 1);
```
gregw, line 100: This feels rather lame.   If we can't control how often we call the callback, then I think our state machine is incomplete.  More over, implementations of callback should be robust in the face of multiple calls anyway with subsequent calls being noops.

Do we really need this?

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/DemandingFlusher.java
```diff
@@ -0,0 +1,128 @@
...
+
+    @Override
+    public void setNextDemand(LongConsumer nextDemand)
+    {
+        _nextDemand = nextDemand;
+    }
+
+    public void onFrame(Frame frame, Callback callback)
+    {
+        _frame = frame;
```
gregw, line 99: probably should ISE if we are not expecting a frame?

gregw (changes requested):
We are spending a lot of time on this PR so I don't want to get too picky.    So my only "must fix" issue is that `DemandingFlusher` needs javadoc.

But there are lots of other aspects I really don't like... but it is the best of the variations we've looked at over the last week.  So probably OK to go ahead with some javadoc.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/FragmentExtension.java
lachlan-roberts: The application doesn't need to succeed this `payloadCallback` until it decides it doesn't need the content anymore. This could be many iterations into this iterating callback where we have a callback for a different frame entirely.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/DemandingFlusher.java
lachlan-roberts: The implementation could multiple of its `payloadCallback`s at once. In the failure of this we need to fail the original frame callback, but we don't want to fail it multiple times. We can't rely on the state of the IteratingCallback because it may have another frame/callback at the time the `payloadCallback` is failed.

In addition to this there could be an error generated from inside the `IteratingCallback` which causes `onCompleteFailure` to be called, and we then want to fail the current frameCallback to relay that failure up to the `WebSocketCoreSession`, but the application may have access to this callback as well and might succeed/fail it at some point after.

So we do need this.

sbordet (dismissed):

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/DemandingFlusher.java
```diff
@@ -0,0 +1,181 @@
...
+    private final AtomicReference<Throwable> _failure = new AtomicReference<>();
+    private LongConsumer _nextDemand;
+
+    private Frame _frame;
+    private Callback _callback;
+    private boolean _needContent = true;
+    private boolean _first = true;
+
+    /**
+     * @param emitFrame where frames generated by {@link #handle(Frame, Callback, boolean)} are forwarded.
```
gregw, line 52:
```suggestion
     * @param emitFrame The BiConsumer to which generated frames are forwarded, along with a callback to succeed when the frame can be released, by {@link #handle(Frame, Callback, boolean)}.
```

We also need to clarify somewhere that `callback.failed(t)` means.   Perhaps that can be javadoc'd in one place (probably `IncomingFrames`) and then whenever  a like callback is used we can refer to that.   Perhaps also we should make the association between the BiConsumer and `IncomingFrames.onFrame`.... actually why isn't this just `IncomingFrames` rather than `BiConsumer`?

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/DemandingFlusher.java
```diff
@@ -0,0 +1,181 @@
...
+     * <p>Called when there is demand for a single frame to be produced. During this method a single call can be made
+     * to {@link #emitFrame(Frame, Callback)} which will forward this frame towards the application. Returning true
+     * from this method signals that you are done processing the current Frame, and the next invocation of this
+     * method will have the next frame.</p>
+     *
+     * <p>Note that the callback supplied here is specially wrapped so that you can call
+     * it multiple times and it will not be completed more than once. This simplifies the
+     * handling of failure cases.</p>
+     * @param frame the original frame.
+     * @param callback to succeed to release the frame payload.
```
gregw, line 69: again failure case is undefined.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/DemandingFlusher.java
```diff
@@ -0,0 +1,181 @@
...
+     *
+     * <p>This method should only be called ONCE within each invocation of {@link #handle(Frame, Callback, boolean)}
+     * otherwise</p>
+     * @param frame the WebSocket frame.
+     * @param callback to release frame payload.
+     */
+    public void emitFrame(Frame frame, Callback callback)
+    {
+        if (_demand.decrementAndGet() < 0)
+            throw new IllegalStateException("Negative Demand");
```
gregw, line 129: Do we need to increment demand again?

gregw (changes requested):
I know I'm asking you to fix problems a bit unassociated with this PR, but I think it would be good to take the 30 minutes to clarify the javadoc about what the failure case of the callback passed to `IncomingFrames` means.

Also wondering why we can't use `IncomingFrames` directly rather than a `BiConsumer`?

But these are niggles and I mostly approve.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/DemandingFlusher.java
lachlan-roberts: The increment is done inside the call to demand that we intercept.

lachlan-roberts: Hmm I get what you mean, the application has already demanded, but we have thrown ISE but don't have demand for the close Frame.

But I don't think this is a big issue because we don't wait for close response if we send an error close code.

And this ISE can only be trigged by a bad implementation of the `DemandingFlusher`. In the javadoc it said to only call `emitFrame` once per call to `handle()`.

gregw: yeah but no!   Ideally an implementation would receive an ISE in response to a bad call, that would not actually change the state of the flusher.     But kind of difficult to achieve atomically without a CAS loop:
```suggestion
        while(true)
        {
            int demand = _demand.get();
            if (demand <= 0)
                throw new IllegalStateException("Negative Demand");
           if (_demand.compareAndSet(demand, demand - 1)
               break;
       }
```

gregw (approved):
Still some quibbles, but let's not hold this up any longer.

sbordet (approved):

PR #7361, opened 2022-01-07
Fix resource leaks in PropertyUserStore, XmlConfiguraiton, and start.jar

lujiefsi:
We have found that some file description leaked in memory.

please review


sbordet (approved):
Thanks!

PR #7374, opened 2022-01-13
Issue #7344 - wait for forked jetty process

### LINKED ISSUE #7344: Incompatible with jacoco due to shutdown race condition
**Jetty version(s)**
jetty-maven-plugin 11.0.7

openjdk version "17.0.1" 2021-10-19
OpenJDK Runtime Environment (build 17.0.1+12)
OpenJDK 64-Bit Server VM (build 17.0.1+12, mixed mode)

**OS type/version**
Arch Linux

**Description**
I'm using jetty and jacoco to generate integration tests coverage.
Jetty shutdowns in post-integration-test phase and jacoco's report-integration goal runs at verify phase.
When i run stop goal in post-integration-test phase, plugin quits immediately even if jetty's VM is not shut down yet.
Because VM is not shutdown yet, jacoco agent that is attached to jetty's VM has not finished writing coverage data file.
While jetty's VM is not shutdown yet, jacoco in project that started and stopped jetty starts collecting report.
It may lead to 3 results:
1. VM shutdown before jacoco executed: good coverage results
2. VM not shutdown before jacoco executed: may lead to EOFException in Jacoco or empty coverage report, as if no coverage was done

Jetty maven plugin should check if VM is shut down or not. `stopWait` tag does not help, feels like it doesn't check anything.

**How to reproduce?**
1. Project properties:
```xml
<jacoco.output.dir>${project.build.directory}/jacoco/output</jacoco.output.dir>
<jacoco.ut.execution.data.file>${jacoco.output.dir}/jacoco-unit-tests.exec</jacoco.ut.execution.data.file>
<jacoco.it.execution.data.file>${jacoco.output.dir}/jacoco-integration-tests.exec</jacoco.it.execution.data.file>
<jetty.maven.plugin.version>11.0.7</jetty.maven.plugin.version>
```
2. Jacoco plugin config:
```xml
<plugin>
    <groupId>org.jacoco</groupId>
    <artifactId>jacoco-maven-plugin</artifactId>
    <version>${jacoco.maven.plugin.version}</version>
    <executions>
        <execution>
            <id>pre-unit-test-execution</id>
            <goals>
                <goal>prepare-agent</goal>
            </goals>
            <configuration>
                <destFile>${jacoco.ut.execution.data.file}</destFile>
                <propertyName>surefire.jacoco.args</propertyName>
            </configuration>
        </execution>
        <execution>
            <id>post-unit-test-execution</id>
            <phase>test</phase>
            <goals>
                <goal>report</goal>
            </goals>
            <configuration>
                <dataFile>${jacoco.ut.execution.data.file}</dataFile>
                <outputDirectory>${project.reporting.outputDirectory}/jacoco-unit-test-coverage-report</outputDirectory>
            </configuration>
        </execution>
        <execution>
            <id>pre-integration-test-execution</id>
            <phase>pre-integration-test</phase>
            <goals>
                <goal>prepare-agent-integration</goal>
            </goals>
            <configuration>
                <destFile>${jacoco.it.execution.data.file}</destFile>
                <propertyName>failsafe.jacoco.args</propertyName>
            </configuration>
        </execution>
        <execution>
            <id>post-integration-test-execution</id>
            <phase>verify</phase>
            <goals>
                <goal>report-integration</goal>
            </goals>
            <configuration>
                <dataFile>${jacoco.it.execution.data.file}</dataFile>
                <outputDirectory>${project.reporting.outputDirectory}/jacoco-integration-test-coverage-report</outputDirectory>
            </configuration>
        </execution>
    </executions>
</plugin>
```
3. Jetty plugin config:
```xml
<plugin>
    <artifactId>jetty-maven-plugin</artifactId>
    <version>${jetty.maven.plugin.version}</version>
    <configuration>
        <httpConnector>
            <port>8080</port>
        </httpConnector>
        <stopKey>quit</stopKey>
        <stopPort>9000</stopPort>
        <stopWait>90</stopWait>
    </configuration>
    <executions>
        <execution>
            <id>start-jetty</id>
            <phase>pre-integration-test</phase>
            <goals>
                <goal>start</goal>
            </goals>
            <configuration>
                <deployMode>FORK</deployMode>
                <jvmArgs>${failsafe.jacoco.args}</jvmArgs>
            </configuration>
        </execution>
        <execution>
            <id>stop-jetty</id>
            <phase>post-integration-test</phase>
            <goals>
                <goal>stop</goal>
            </goals>
        </execution>
    </executions>
</plugin>
```

Workaround:
Put some plugin in-between that might halt execution for a split second, like this.
If just having this plugin in-between does not help, launch maven with `-Dwait=true`
It will ask you to press enter, wait a sec or two and press enter.
```xml
<plugin>
    <groupId>net.stickycode.plugins</groupId>
    <artifactId>wait-maven-plugin</artifactId>
    <version>1.5</version>
    <executions>
        <execution>
            <phase>post-integration-test</phase>
            <goals>
                <goal>wait</goal>
            </goals>
            <configuration>
                <promptMessage>Just chilling?</promptMessage>
            </configuration>
        </execution>
    </executions>
</plugin>
```

janbartel:
Closes #7344 

Change `org.eclipse.jetty.server.ShutdownMonitor` to add a new command: `pid`. This reports the process id of the running jetty process. Changed the `org.eclipse.jetty.maven.plugin.JettyStopMojo` to first send this command to the running jetty process to obtain its id before sending a `stop` command. If the `pid` is retrievable, the stop mojo will wait (a configurable amount of time) for the jetty process to exit before terminating. If the `pid` is bad or missing, then the stop mojo falls back to it's previous behaviour of waiting a configurable amount of time to receive the `Stopped` message from the jetty process before terminating.


jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyStopMojo.java
```diff
@@ -72,38 +74,132 @@ public void execute() throws MojoExecutionException, MojoFailureException
...
-                        stopped = true;
                         getLog().info("Server reports itself as stopped");
+                    else
+                        getLog().info("Couldn't verify server as stopped, received " + response);
+                }
+                else
+                {
+                    //wait for pid to stop
+                    getLog().info("Waiting " + stopWait + " seconds for jetty " + pid + " to stop");
+                    send(stopKey + "\r\n" + command + "\r\n", false, 0);
```
XakepSDK: It's better for `send(...)` to be after `ProcessHandle#of` to avoid potential race condition

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyStopMojo.java
```diff
@@ -72,38 +74,132 @@ public void execute() throws MojoExecutionException, MojoFailureException
...
+     */
+    private String send(String command, boolean expectResponse, int wait)
+        throws Exception
+    {
+        String response = null;
+        try (Socket s = new Socket(InetAddress.getByName("127.0.0.1"), stopPort); OutputStream out = s.getOutputStream();)
         {
-            getLog().error(e);
+            out.write(command.getBytes());
+            out.flush();
```
sbordet, line 198: No, TCP is not very friendly with open/write/close.
You should always read to have a confirmation that the other peer received the data (even if it's only reading -1).
The other machine could be overloaded, so SYN+PSH+FIN may be received by the recipient at OS level (but not yet at application level), but because the machine is overloaded timeouts may happen, so the OS decides to RST back to the sender, which has already closed. 
The net result is that the sender does not know that something went wrong, the recipient never received the data at the application level, and everything went into /dev/null.

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/MockShutdownMonitor.java
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * MockShutdownMonitor
+ * 
+ * A helper class that grabs a ServerSocket, spawns a thread and then
+ * passes the ServerSocket to the Runnable. This class has a main so
+ * that it can be used for forking, to mimic the actions of the
+ * org.eclipse.jetty.server.ShutdownMonitor.
+ *
+ */
+public class MockShutdownMonitor
```
sbordet, line 27: Why can't we use `ShutdownMonitor`? 
It should be possible because there are tests that use `ShutdownMonitor` directly and not a mock.
We really want to test the real `ShutdownMonitor`, not this mock class.

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyStopMojo.java
```diff
@@ -72,38 +74,132 @@ public void execute() throws MojoExecutionException, MojoFailureException
...
+                    else
+                        getLog().info("Couldn't verify server as stopped, received " + response);
+                }
+                else
+                {
+                    //wait for pid to stop
+                    getLog().info("Waiting " + stopWait + " seconds for jetty " + pid + " to stop");
+                    send(stopKey + "\r\n" + command + "\r\n", false, 0);
+                    Optional<ProcessHandle> optional = ProcessHandle.of(pid);
+                    ProcessHandle handle = optional.orElse(null);
```
sbordet: Use `Optional.ifPresent(handle -> ...)` instead.

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyStopMojo.java
```diff
@@ -72,38 +74,132 @@ public void execute() throws MojoExecutionException, MojoFailureException
...
+                    if (handle != null)
+                    {
+                        try
+                        {
+                            CompletableFuture<ProcessHandle> future = handle.onExit();
+                            if (handle.isAlive())
+                            {
+                                handle = future.get(stopWait, TimeUnit.SECONDS);
+                            }
```
sbordet: Don't think this block is correct.
`future.get(wait, unit)` throws `TimeoutException` if the time elapses, but the code seems to hint something like "if timeout and still alive, do something", so the "do something" will be skipped.

Have you considered using `ProcessHandle.destroy[Forcibly]()` if the timeout expires?
Perhaps along with `CompletableFuture.orTimeout()`?

sbordet (changes requested):

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyStopMojo.java
janbartel: @sbordet this would be a behaviour change of the JettyStopMojo:  it has always just made a best effort to stop the remote jetty, waiting up to the configured time to get an acknowledgement from jetty that it is shutting down before giving up. I suppose we could add a new configuration option to forcibly destroy the process after the timeout expires, but I think that might be in another issue.

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyStopMojo.java
janbartel: If the sender is interested in getting confirmation from the remote jetty, then they configure a non-zero `stopWait` timeout, in which case we set the socket soTimeout and read from it to get the "Shutting down" message.  Alternatively, if we have been able to obtain the pid, we send the message and then wait up to `stopWait` seconds on the process itself to stop. So in either case, there will be a wait for either a response message, or the process to die.  If no wait is configured, then we don't wait for any response, we just send the message and exit.

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/MockShutdownMonitor.java
janbartel: @sbordet I'm not interested in testing ShutdownMonitor - we have other tests for that ;). The unit test it relates to is of the JettyStopMojo - all I want to do is to test the behaviour of the JettyStopMojo, and for that I need to control the behaviour of the ShutdownMonitor and its ShutdownMonitorRunnable. I can't do that by subclassing because ShutdownMonitor is effectively  not extensible, hence the mocks.

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyStopMojo.java
janbartel: Done.

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyStopMojo.java
janbartel: If the process isn't running when we do the send, then we will just log "Jetty not running".

XakepSDK: 1. Run `send`
2. Execution blocks for unknown reasons
3. Another app starts and system assigns PID previously used by jetty
4. We get `ProcessHandle` for unrelated application
5. We get unnecessary stall

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyStopMojo.java
sbordet: You are assuming that the message you send reaches the other end. I'm saying that it may not.
Now, if you care whether the message reaches the other end, you should read, at least a -1.
If you don't care whether the message reaches the other end, then you may not read (like the code above), just don't assume that `send()` actually produces the results you want.
It's not about getting confirmation about the specific command.
It's about getting a confirmation that the message even arrived to the other process.

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/MockShutdownMonitor.java
sbordet: I made `ShutdownMonitor.start()` and `stop()` public and I can rewrite these tests using directly `ShutdownMonitor` without 200 lines of mock code that must to be bug-free and maintained.
I'd prefer to fix `ShutdownMonitor` than adding a mock for it.

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/TestJettyStopMojo.java
```diff
@@ -0,0 +1,247 @@
...
+        
+        //start the responder
+        MockShutdownMonitorRunnable runnable = new MockShutdownMonitorRunnable();
+        runnable.setPidResponse("abcd");
+        MockShutdownMonitor monitor = new MockShutdownMonitor(stopKey, runnable, false);
+        monitor.start();
+        
+        TestLog log = new TestLog();
+        JettyStopMojo mojo = new JettyStopMojo();
+        mojo.stopKey = "foo";
```
sbordet: Use the variable `stopKey`.

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/TestJettyStopMojo.java
```diff
@@ -0,0 +1,247 @@
...
+        log.assertContains("Server returned bad pid");
+        log.assertContains("Server reports itself as stopped");
+    }
+
+    @Test
+    public void testStopWait() throws Exception
+    {
+        String stopKey = "foo";
+
+        List<String> cmd = new ArrayList<String>();
```
sbordet:
```suggestion
        List<String> cmd = new ArrayList<>();
```

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/TestJettyStopMojo.java
```diff
@@ -0,0 +1,247 @@
...
+        command.directory(MavenTestingUtils.getTargetDir());
+        Process fork = command.start();
+        
+        Thread.sleep(500);
+        while (!file.exists() && file.length() == 0)
+        {
+            Thread.sleep(300);
+        }
+
+        String str = null;
```
sbordet:
```suggestion
        String str;
```

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/TestJettyStopMojo.java
```diff
@@ -0,0 +1,247 @@
...
+        {
+            str = reader.readLine();
+        }
+        assertNotNull(str);
+
+        TestLog log = new TestLog();
+        JettyStopMojo mojo = new JettyStopMojo();
+        mojo.stopWait = 5;
+        mojo.stopKey = stopKey;
+        mojo.stopPort = Integer.valueOf(str).intValue();
```
sbordet:
```suggestion
        mojo.stopPort = Integer.parseInt(str);
```

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/TestJettyStopMojo.java
```diff
@@ -0,0 +1,247 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+public class TestJettyStopMojo
+{   
+    public static class TestLog implements org.apache.maven.plugin.logging.Log
+    {
+        List<String> sink = new ArrayList<String>();
```
sbordet:
```suggestion
        List<String> sink = new ArrayList<>();
```

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/TestJettyStopMojo.java
```diff
@@ -0,0 +1,247 @@
...
+        Process fork = command.start();
+        
+        Thread.sleep(500);
+        while (!file.exists() && file.length() == 0)
+        {
+            Thread.sleep(300);
+        }
+
+        String str = null;
+        try (LineNumberReader reader = new LineNumberReader(new FileReader(file));)
```
sbordet: Unnecessary semicolon.

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/MockShutdownMonitor.java
```diff
@@ -0,0 +1,97 @@
...
+//
+
+package org.eclipse.jetty.maven.plugin;
+
+import java.net.ServerSocket;
+
+import org.eclipse.jetty.toolchain.test.IO;
+
+/**
+ * MockShutdownMonitor
```
sbordet, line 21: Unnecessary line, adds nothing.

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/MockShutdownMonitor.java
```diff
@@ -0,0 +1,97 @@
...
+ * 
+ * A helper class that grabs a ServerSocket, spawns a thread and then
+ * passes the ServerSocket to the Runnable. This class has a main so
+ * that it can be used for forking, to mimic the actions of the
+ * org.eclipse.jetty.server.ShutdownMonitor.
+ *
```
sbordet: Unnecessary newline.

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/MockShutdownMonitorRunnable.java
```diff
@@ -0,0 +1,122 @@
...
+import java.io.LineNumberReader;
+import java.io.OutputStream;
+import java.net.ServerSocket;
+import java.net.Socket;
+import java.nio.charset.StandardCharsets;
+
+import org.eclipse.jetty.toolchain.test.IO;
+
+/**
+ * MockShutdownMonitorRunnable
```
sbordet, line 26: Unnecessary line, adds nothing.

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/MockShutdownMonitorRunnable.java
```diff
@@ -0,0 +1,122 @@
...
+ * 
+ * Mimics the actions of the org.eclipse.jetty.server.ShutdownMonitor.ShutdownMonitorRunnable
+ * to aid testing.
+ *
```
sbordet: Unnecessary newline.

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/MockShutdownMonitor.java
```diff
@@ -0,0 +1,97 @@
...
+        testerRunnable.setKey(key);
+        Thread thread = new Thread(testerRunnable);
+        thread.setDaemon(true);
+        thread.setName("Tester Thread");
+        thread.start();
+        if (join)
+            thread.join();
+    }
+
+    public static final void main(String[] args)
```
sbordet: `static` methods should not be also `final`.

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/MockShutdownMonitorRunnable.java
```diff
@@ -0,0 +1,122 @@
...
+    
+    public void setDefaultResponse(String response)
+    {
+        this.defaultResponse = response;
+    }
+
+    public void setStatusResponse(String statusResponse)
+    {
+        this.statusResponse = statusResponse;
+    }
```
sbordet: Never used, remove them.

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyStopMojo.java
sbordet: @XakepSDK good point.

sbordet (changes requested):

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/TestJettyStopMojo.java
janbartel: Done.

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/TestJettyStopMojo.java
janbartel: Done.

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/MockShutdownMonitor.java
janbartel: Done

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/MockShutdownMonitorRunnable.java
janbartel: Done.

jetty-maven-plugin/src/test/java/org/eclipse/jetty/maven/plugin/MockShutdownMonitor.java
janbartel: @sbordet I have changed the final test `testStopWait` to talk to the real `ShutdownMonitor`. None of the other tests can use the real class because I need to test failure/fallback modes, and `ShutdownMonitor` is unsubclassable (private no-args constructor).

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyStopMojo.java
janbartel: Done.

jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyStopMojo.java
janbartel: So @sbordet how long does the sender need to wait to read from the socket to ensure that the message was sent? The behaviour of the `ShutdownMonitor` is to stop jetty _before_ replying with "Stopped" and exiting. In the case where the caller has _not_ configured a `stopWait` time they do _not_ want to wait until the socket is closed etc.

janbartel:
@sbordet latest release has gone out - can you now approve this pr?

sbordet (commented):
@janbartel a few suggested fixes have not been addressed (e.g. javadocs, semicolons, unnecessary types, etc.)?

sbordet (approved):

olamy (approved):

PR #7377, opened 2022-01-13
Proposed IteratingDemand class

gregw:
This is a proposed class that may better model the demand interception model needed by websocket exceptions and potentially jetty-12.

There are some complexities with the nuances of releasing the input buffer that require the `O produce(I input, Runnable release)` method to sometimes be called with null input to see if anything can be produced from nothing. An alternative would be to have a method like `O produce(I input, Runnable release, Runnable complete)` so the method could separately indicate releasing of the input vs completion of production of everything that came from that input (which might be able to be produced after the input is released).


gregw:
I attempted to replace @lachlan-roberts's `DemandingFlusher` with this.   It worked OK for the data and demand flow, but the release buffer flow was really a hack.  The issue being that with websocket extension a Callback is used to signal either Frame release or and error in handling.   I don't think that is the right API and I don't really want to make a util class setup to deal with it.    
So I'm closing this for now.  I may resurrect once we see real need for this in the jetty-12 model where error flows are kept out of both the demand and release mechanisms, so life is simple.

PR #7393, opened 2022-01-14
Updated Authority Override test

gregw:
An empty host header is now a Bad Request unless the URI is absolute, also with no host.
Bad Requests in this test are returned as 302s from the redirection in the error handler.


gregw:
@joakime can you give this some priority to review, as it is currently breaking jetty-server in 12.0.x branch.

gregw:
@joakime I'm going to merge this today, as the broken test in main jetty-12.0.x is disruptive.     However, let's find time later this week or next to discuss the changes.

PR #7407, opened 2022-01-18
Implementation of utility Content.Processor #7403

### LINKED ISSUE #7403: Jetty-12 `Content.Provider` Producer?
**Jetty version**
12

**Question**
After the recent conversations regarding websocket and IteratingCallback I wanted to make sure that Jetty-12 has a good processor abstraction that handles both demand and release correctly.
But I'm currently tying myself in knots trying to come up with right abstract for how a truly general processor should work (ie one that can both silently consume and spontaneously produce content(s)).

So currently the `Content` signature is:

```java
public interface Content
{
    ByteBuffer getByteBuffer();
    boolean isSpecial();
    boolean isLast()
    void release();
}
```

and we have a Provider abstraction like:
```java
    interface Provider
    {
        Content readContent();
        void demandContent(Runnable onContentAvailable);
    }
```

So the challenge is to make a utility processor class with a signature like:
```java

public abstract class ContentProcessor implements Content.Provider
{
    public ContentProcessor(Content.Provider provider) { /* ... */ }

    public Content readContent() { /* ... */ }
    public  void demandContent(Runnable onContentAvailable) { /* ... */ }

    protected abstract Content produce(Content content);
}
```

The idea being that the processor will call the provider for demand/content as needed and then call `produce(Content content)` to transform that content somehow:
 + content may be passed straight through, in which case it is the responsibility of the ultimate consumer to release it.
 + content may be translated on a 1:1 basis to new content, in which case the processor must release the passed content and the consumer must release the produced content.
 + content may be consumed without producing any content.  It may or may not be released when it is consumed and it's remaining count may or may not be none zero.
 + we need to allow consumed content to produce multiple contents.
 + perhaps we should allows a processor to spontaneously produce content without any input content????

So the big question I have, is what exactly is the contract with the `Content produce(Content content)` method. Specifically how does it indicate that it has finished with the passed content and the processor should try to read more content from the provider?

For the purposes of the following discussion, I am going to user the name "input" for content read from the provider and passed to the `produce` method.  I will use the name "output" for content returned from the `produce` method. Yeah I know, not strictly correct but will do for the purposes of discussion.

**Option 1**

The same input content is passed repeatedly to the `produce` method until such time as it returns null.  So consider a producer that:
 + splits string content on white space into new content
 + consumes any strings "NOOP"
 + Copies any other upper case content into a new lower case content
 + passes all other content straight through

So if the following queue was available:
> "one"
> "TWO"
> "three four"
> "NOOP"
> "five NOOP six"
there would be the following sequence of calls and returns (note that I'm not showing any consumptions of bytes in the content)
```
 provider.readContent() -> "one"
 produce("one") -> "one"
 produce("one") -> null
 provider.readContent() -> "TWO"
 produce("TWO") -> "two"
 produce("TWO") -> null
 provider.readContent() -> "three four"
 produce("three four") -> "three"
 produce("three four") -> "four"
 produce("three four") -> null
 provider.readContent() -> "NOOP"
 produce("NOOP") -> null
 provider.readContent() -> "five NOOP six"
 produce("five NOOP six") -> "five"
 produce("five NOOP six") -> "six"
 produce("five NOOP six") -> null
```

PROS:
 + processor doesn't need to remember the input as it keeps getting passed in until it is finished

CONS:
 + two calls to `produce` for every content in a 1:1 scenario
 + if the processor releases the input content but returns non null, it is *wrong* to pass a released content to the processor on subsequent calls as the bytebuffer may be reused and contain anything!  So processors would need to delay releasing until they return null.
 + If the processor uses release to signal input demand, then it may be forced to hold onto large content longer than is optimal to signal back pressure.

**Option 2**
The same as Option 1 but input is represented until such time as null is return OR it is released.   So for the same example producer and same input:

```
 provider.readContent() -> "one"
 produce("one") -> "one"
 produce("one") -> null
 provider.readContent() -> "TWO"
 produce("TWO") -> "two"  && "TWO".release()
 provider.readContent() -> "three four"
 produce("three four") -> "three"
 produce("three four") -> "four" && "three four".release()
 provider.readContent() -> "NOOP"
 produce("NOOP") -> null
 provider.readContent() -> "five NOOP six"
 produce("five NOOP six") -> "five"
 produce("five NOOP six") -> "six" && "five NOOP six".release()
```

PROS:
 + If the processor does the releasing, you can avoid lots of null returning calls and get close to one call to `produce` for every content in a 1:1 scenario

CONS:
 + It's a little subtle and complex.
 + input content needs to be wrapped to intercept release

**Option 3**
A content is passed to the processor once and only once. Instead, it is called with null to see if anything can be produced without more input.   So for the same example, the calling sequence would be:
```
 produce(null) -> null
 provider.readContent() -> "one"
 produce("one") -> "one"
 produce(null) -> null
 provider.readContent() -> "TWO"
 produce("TWO") -> "two"
 produce(null) -> null
 provider.readContent() -> "three four"
 produce("three four") -> "three"
 produce(null) -> "four"
 produce(null) -> null
 provider.readContent() -> "NOOP"
 produce("NOOP") -> null
 provider.readContent() -> "five NOOP six"
 produce("five NOOP six") -> "five"
 produce(null) -> "six"
 produce(null) -> null
```

PROS:
 + the contract is simple to describe.
 + no issues with releasing the input (can be done at anytime by either processor or consumer)
 + the processor can produce output even before any input has arrived.
 + It is the only option I've found that has a clear signals for "input release" and "input demand" that does not allow processor abuse.

CONS:
 + lots of `produce(null)` calls
 + the processor has to keep a reference to the input if it wants to use it multiple times 

**Option 4**
Have a produce that returns a boolean like `boolean produce(Content input, Consumer<Content> output)`, that returns true when the processor is finished with the input.
PROS:
 + The boolean makes it a lot clearer when the producer no longer wants to deal with the input.

CONS:
 + We still have the issue of a producer that returns false, but has released the input... although that could be fixed by only passing an input once and then passing null until a true return
 + the output consumer can be called multiple times for each produce call, which can violate demand.

**Option 5**
The producer could be given the option to control it's own input demand like `Content produce(Content input, Runnable demandInput)`
PROS:
 + probably will work

CONS:
 + complicates the producer so there is significant risk of it not working

**Option 6**
Somehow use bytes remaining????

CONS:
 + Doesn't well represent a fully consumed input  that can still produce multiple outputs.  We don't want to request more input until it is truly finished.
 + Doesn't allows content to be held by processor nor consumer.

**Discussion**
I'm not exactly sure which one I like (or which I dislike least)..... but probably going to try to implement **option 3** to test.       However, this now put's yet another requirement on our consideration for better APIs/contracts in jetty-12.   Not only does anything we come up with have to work with all the jetty-server unit-tests that are currently passing, but I think we need to show that it has a good processor behavior as well.      The output side is not so bad as that is just traditional write with `Callback` and neither demand nor release are relevant.

thoughts?

gregw:
This PR adds the `Content.Processor` abstration and a `ContentProcessor` utility implementation. See discussion in #7403

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/ContentProcessor.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/ContentProcessor.java
@@ -53,6 +53,7 @@ public abstract class ContentProcessor extends Content.Processor
             _reading = true;
         }
 
+        Content input = null;
         try
         {
             while (true)
@@ -61,13 +62,17 @@ public abstract class ContentProcessor extends Content.Processor
                 if (output != null)
                     return output;
                 available = false;
-                Content input = getProvider().readContent();
+                input = getProvider().readContent();
                 output = process(input);
                 if (output != null)
                     return output;
                 if (input == null)
                     return null;
             }
+        }
+        catch (Throwable t)
+        {
+            return new Content.Error(t, input != null && input.isLast());
         }
         finally
         {
@@ -125,26 +130,33 @@ public abstract class ContentProcessor extends Content.Processor
         {
             if (onContentAvailable != null)
             {
+                Throwable error = null;
                 try
                 {
                     onContentAvailable.run();
-                    iterating = false;
+                }
+                catch (Throwable t)
+                {
+                    error = t;
                 }
                 finally
                 {
+                    onContentAvailable = null;
                     try (AutoLock ignored = _lock.lock())
                     {
-                        if (iterating)
-                            // exception must have been thrown
-                            _iterating = iterating = false;
-                        else if (_onContentAvailable != null)
+                        if (error != null)
                         {
-                            iterating = true;
-                            onContentAvailable = null;
+                            if (_output != null)
+                                _output.release();
+                            _output = new Content.Error(error, false);
                         }
-                        else
+
+                        if (_onContentAvailable == null)
+                            _iterating = iterating = false;
+                        else if (_output != null)
                         {
-                            _iterating = iterating;
+                            onContentAvailable = _onContentAvailable;
+                            _onContentAvailable = null;
                         }
                     }
                 }
@@ -163,6 +175,13 @@ public abstract class ContentProcessor extends Content.Processor
                         if (output == null && input == null)
                             getProvider().demandContent(this::onContentAvailable);
                     }
+                }
+                catch (Throwable t)
+                {
+                    if (output != null)
+                        output.release();
+                    output = new Content.Error(t, input != null && input.isLast());
+                    available = false;
                 }
                 finally
                 {
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/ContentProcessor.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/ContentProcessor.java
@@ -13,6 +13,7 @@
 
 package org.eclipse.jetty.server;
 
+import java.nio.channels.ReadPendingException;
 import java.util.Objects;
 
 import org.eclipse.jetty.util.thread.AutoLock;
@@ -38,7 +39,7 @@ public abstract class ContentProcessor extends Content.Processor
         try (AutoLock ignored = _lock.lock())
         {
             if (_reading)
-                return null;
+                throw new ReadPendingException();
 
             if (_output != null)
             {
@@ -56,12 +57,12 @@ public abstract class ContentProcessor extends Content.Processor
         {
             while (true)
             {
-                Content output = available ? null : produce(null);
+                Content output = available ? null : process(null);
                 if (output != null)
                     return output;
                 available = false;
                 Content input = getProvider().readContent();
-                output = produce(input);
+                output = process(input);
                 if (output != null)
                     return output;
                 if (input == null)
@@ -153,11 +154,11 @@ public abstract class ContentProcessor extends Content.Processor
                 Content input = null;
                 try
                 {
-                    output = available ? null : produce(null);
+                    output = available ? null : process(null);
                     if (output == null)
                     {
                         input = getProvider().readContent();
-                        output = produce(input);
+                        output = process(input);
                         available = false;
                         if (output == null && input == null)
                             getProvider().demandContent(this::onContentAvailable);
@@ -201,5 +202,5 @@ public abstract class ContentProcessor extends Content.Processor
         iterate(true, null);
     }
 
-    protected abstract Content produce(Content content);
+    protected abstract Content process(Content content);
 }
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/ContentProcessor.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/ContentProcessor.java
@@ -1,0 +1,205 @@
+//
+// ========================================================================
+// Copyright (c) 1995-2021 Mort Bay Consulting Pty Ltd and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.server;
+
+import java.util.Objects;
+
+import org.eclipse.jetty.util.thread.AutoLock;
+
+public abstract class ContentProcessor extends Content.Processor
+{
+    private final AutoLock _lock = new AutoLock();
+    private boolean _reading;
+    private boolean _iterating;
+    private boolean _available;
+    private Content _output;
+    private Runnable _onContentAvailable;
+
+    public ContentProcessor(Content.Provider provider)
+    {
+        super(provider);
+    }
+
+    @Override
+    public Content readContent()
+    {
+        boolean available;
+        try (AutoLock ignored = _lock.lock())
+        {
+            if (_reading)
+                return null;
+
+            if (_output != null)
+            {
+                Content output = _output;
+                _output = null;
+                return output;
+            }
+
+            available = _available;
+            _available = false;
+            _reading = true;
+        }
+
+        try
+        {
+            while (true)
+            {
+                Content output = available ? null : produce(null);
+                if (output != null)
+                    return output;
+                available = false;
+                Content input = getProvider().readContent();
+                output = produce(input);
+                if (output != null)
+                    return output;
+                if (input == null)
+                    return null;
+            }
+        }
+        finally
+        {
+            try (AutoLock ignored = _lock.lock())
+            {
+                _reading = false;
+            }
+        }
+    }
+
+    @Override
+    public void demandContent(Runnable onContentAvailable)
+    {
+        Objects.requireNonNull(onContentAvailable);
+        iterate(false, onContentAvailable);
+    }
+
+    private void iterate(boolean available, Runnable onContentAvailable)
+    {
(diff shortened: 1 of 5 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/ContentProcessor.java
```diff
@@ -0,0 +1,205 @@
...
+            }
+        }
+    }
+
+    private void onContentAvailable()
+    {
+        iterate(true, null);
+    }
+
+    protected abstract Content produce(Content content);
```
sbordet: This should be called `process()` or `transform()`.

jetty-server/src/main/java/org/eclipse/jetty/server/ContentProcessor.java
```diff
@@ -0,0 +1,205 @@
...
+
+        Content output = null;
+        boolean iterating = true;
+        while (iterating)
+        {
+            if (onContentAvailable != null)
+            {
+                try
+                {
+                    onContentAvailable.run();
```
sbordet, line 136: I still think that this is not enough.
A downstream application could read&throw or just decide to fail. 
If this class has allocated resources that is needs to release, how does the application communicate the failure upstream?

I was thinking that `Provider` should have a `cancel(Throwable)` method for that.
Throwing `x` from `run()` implicitly calls `cancel(x)`.

In this way, all places in this class that throw e.g. `IllegalStateException` will arrive to the application which can decide whether to handle them, but if they are not handled, they result in `cancel(x)` and resources are released along the whole chain.

jetty-server/src/main/java/org/eclipse/jetty/server/ContentProcessor.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    @Override
+    public Content readContent()
+    {
+        boolean available;
+        try (AutoLock ignored = _lock.lock())
+        {
+            if (_reading)
+                return null;
```
sbordet: I would throw `ReadPendingException` here.

jetty-server/src/main/java/org/eclipse/jetty/server/ContentProcessor.java
```diff
@@ -0,0 +1,205 @@
...
+            available = _available;
+            _available = false;
+            _reading = true;
+        }
+
+        try
+        {
+            while (true)
+            {
+                Content output = available ? null : produce(null);
```
sbordet: `produce(null)` should be avoided (also, not sure the logic above is correct for the first pass -- available seems to either be `false` or skip `produce()`?).

If upstream always has content, `produce(null)` should never happen.
I would prefer the contract of `produce()` to be fully consuming the input content, and return the first transformed.
Method `readContent()` will take care of returning transformed content if present; if not present it should demand for it, similar to HTTP/1 `fillAndParse()` -- we only return `null` from `readContent()` if we tried and we are truly stalled.

sbordet (changes requested):
It's a lot of locking and state to cope with the most generic case (where demands and reads may be called at any time by any thread), but a useful utility class for transforming content.

Still think the main contract is lacking upstream failure notification though.

jetty-server/src/main/java/org/eclipse/jetty/server/ContentProcessor.java
gregw: > If upstream always has content, `produce(null)` should never happen.

That was the point of the discussion I tried to have in the issue.   It may be that upstream has content, but the previous content is still able to produce more content (eg being expanded), so we need to call the produce again when there is more demand to see if it can produce more.   There are designs that don't need to pass null, but they have other issues (like letting the produce go beyond demand).

> I would prefer the contract of produce() to be fully consuming the input content,
Because we are using `Content` with `release()` rather than just `ByteBuffers`, then we cannot use consumption as a signal for anything.   A Processor is free to hold onto a content over multiple cycles and release it when it wants.  It is also free to pass it onto the application, which may then hold onto it for a while until it calls release.  The whole point of `release()` is to avoid copying (ie forcing consumption before it is necessary).       If the ultimate target of this chain is a servlet application reading 1 byte at a time, we don't want to force anybody to fully consume content as that will imply a needless copy and buffering.

As far as I can tell, the state of consumption of a content can have no influence on how we process it.   Even a consumed content might be retained so that the processor or application can pick bytes out of it later.

> Method readContent() will take care of returning transformed content if present;

If we go with that approach, then it may be that a processor fully consumes some content that arrives.  In that case, we will have called the consumers/applications onContentAvailable method and it will do a `readContent` that returns null.     That may be OK, but we'd have to rename the concept from `onContentAvailable` to 'tryAgainNowButItMayStillBeNull` (or something better to that intent).       I think the impl I have here shows that it is at least possible to "filter" content availability and to only call the consumer/application once there really is content available.  Cost is we have to "buffer" one content between when notification and a subsequent call to `readContent()`.

jetty-server/src/main/java/org/eclipse/jetty/server/ContentProcessor.java
gregw: I'm not sure we can ever really support any `cancel` semantic as you can't take back demand.

But your point is valid - we need to better test error handling in the `Processor` and probably in `Provider` generally.    I've got some tests of error cases for the direct usage of `Provider` in the `Request` API, but probably not enough.... and none in this PR.

I'll add the simple cases of an exception thrown from an 'onContentAvailable` callback, and also one thrown from `produce`.   If you can think of any other specific error use-cases you want tested... and the outcome you expect, then do either write them or describe them here.

In my other handling, I have tried to avoid additional error channel and instead have kept errors "in-band", so inboud errors are reported via a `readContent` and outbound errors are signalled by a `Callback.failed(x)` on the original request invocation.    Any `onError` methods just invite all the complexities of concurrent invocation and errors thrown from `onError` etc.          But let's do the testing here and see how it looks.

gregw:
@sbordet 
> It's a lot of locking and state to cope with the most generic case

geeeeeeze !!!!  You should have seen the first lock happy attempt(s)!!!!      I thought I've done pretty good to only grab the lock once per iteration yet still never call out with the lock held!    I'm not sure we can do better than that.

But there are probably some special cases of `Processor`s that don't need the full machinery and could get by with less/no/atomic locking.    So yes, the generic case requires at least a lock per iteration, with perhaps a lock of an already held lock when demand/read is called from within an onContentAvailable callback.  Actually a read can sometimes grab the lock twice, but only in difficult cases.

jetty-server/src/main/java/org/eclipse/jetty/server/ContentProcessor.java
gregw: @sbordet @lorban I've pushed an update that now handles errors thrown from `process` or `available` calls.   It handles all errors inband, by providing them via `readContent()`.  If the consumer is not reading nor demanding, then the error is held until they are.     

To make this sensible, I've introduced the concept of a non-last error!  So you can read an `Content.Error` from `readContent()`, that if it did not originate from the original provider will not be last (unless the input that caused the error to be generated was already last).

There is some risk of applications infinitely looping if in their available method they: 1) readContent; 2) demandContent; 3) throw.   But this is really no different to the looping danger if an onError callback throws.  I could add loop protection?   Actaully I have and idea for that..... stand by....

gregw: hmmm no that idea didn't work. anyway have a look and feedback please.

gregw:
I'm, going to merge this branch, so this class is available for the experiments to be conducted with a no-args `demandContent` method.
The current jetty-12.0.x branch will ultimately become a jetty-12.0.x-experiment2 branch and a new jetty-12.0.x branch started with the leasons learnt

PR #7413, opened 2022-01-18
GunzipContentProvider implementation without Content.Processor

lorban:
**Jetty version**
12

The simplest possible `Content.Provider` implementation I could think of that gunzips content sourced from another `Content.Provider`.

See #7407.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/GunzipContentProvider.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/GunzipContentProvider.java
@@ -1,0 +1,108 @@
+package org.eclipse.jetty.server;
+
+import java.nio.ByteBuffer;
+
+import org.eclipse.jetty.http.GZIPContentDecoder;
+import org.eclipse.jetty.util.BufferUtil;
+
+public class GunzipContentProvider implements Content.Provider
+{
+    private final GZIPContentDecoder decoder = new GZIPContentDecoder();
+    private final Content.Provider deflatedProvider;
+
+    private Content currentDeflatedContent;
+    private ByteBuffer inflatedBuffer;
+
+    public GunzipContentProvider(Content.Provider deflatedProvider)
+    {
+        this.deflatedProvider = deflatedProvider;
+    }
+
+    @Override
+    public Content readContent()
+    {
+        while (true)
+        {
+            if (currentDeflatedContent == null)
+            {
+                currentDeflatedContent = deflatedProvider.readContent();
+                if (currentDeflatedContent == null)
+                    return null;
+            }
+            if (currentDeflatedContent.isSpecial())
+                return currentDeflatedContent;
+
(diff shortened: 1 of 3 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/GunzipContentProvider.java
```diff
@@ -0,0 +1,121 @@
...
+            inflatedBuffer = null;
+            currentDeflatedContent.release();
+            currentDeflatedContent = null;
+        }
+    }
+
+    @Override
+    public void demandContent(Runnable onContentAvailable)
+    {
+        deflatedProvider.demandContent(onContentAvailable);
```
sbordet, line 85: I think you're covering only the case where the user of this class uses the style:
```java
onContentAvailable() {
  while (true) {
    Content c = readContent();
    if (c == null)
      demandContent(this::onContentAvailable);
  } else {
    consumeContent(c);
}
```

But there is also the style:

```java
onContentAvailable() {
  Content c = readContent();
  consumeContent(c);
  demandContent(this::onContentAvailable);
}
```
I don't think the second style will work if you always forward the demand upstream, right?

sbordet (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/GunzipContentProvider.java
lorban: You're right, but that's a different problem that can be solved independently. Let me add an extra test for this use-case as well as the required implementation.

lorban:
Something important I noticed about the `Provider` interface's contract: the `void demandContent(Runnable onContentAvailable)` method must be thread-safe while the `Content readContent()` one must not.

This must be carefully documented (or discussed in case we disagree on this).

jetty-server/src/main/java/org/eclipse/jetty/server/GunzipContentProvider.java
lorban: @sbordet I've tackled this problem with the simplest possible implementation I could think of (again), without modifying anything related to `readContent()`.

The only thing worth mentioning I relied upon is that `Provider.demandContent()` is thread-safe while `Provider.readContent()` is not.

gregw:
There are 3 "events": a call to `demandContent(Runnable)`; a call to `readContent()`; and the arrival of data available notification that can happen any time during/after `demandContent()`.
It is unlikely, but not impossible for the app to call `demandContent` and/or `readContent` concurrently, so the main race is in the arrival of available notification in the implementation of `demandContent`.  All other races could be said to be bad applications, but we should strive to behaive nicely none the less.

gregw:
Oh and the secondary race of available notification or `demandContent` call   racing with the return of a call to the Runnable passed to the previous `demandContent` call.

PR #7417, opened 2022-01-18
Clarify that requestHeaderSize is a cumulative limit

mszabo-wikia:
HttpConfiguration documents the requestHeaderSize configuration option
as being a limit on the size of a single request header, but it is in
fact a limit on the cumulative size of all request headers as well as
the request URI. This patch updates the documentation accordingly, and
adds test cases for the HTTP/1.x and HTTP/2 parsers to verify the
behavior.

NB.: the HTTP/3 parser and configuration seem to correctly document this
option as being a global limit on header size.

Signed-off-by: Máté Szabó <mszabo@wikia-inc.com>

### DIFF
--- a/jetty-http/src/test/java/org/eclipse/jetty/http/HttpParserTest.java
+++ b/jetty-http/src/test/java/org/eclipse/jetty/http/HttpParserTest.java
@@ -2113,7 +2113,7 @@ public class HttpParserTest
                 "Connection: close\r\n" +
                 "\r\n");
 
-        int maxHeaderBytes = 20;
+        int maxHeaderBytes = 5;
         HttpParser.RequestHandler handler = new Handler();
         HttpParser parser = new HttpParser(handler, maxHeaderBytes);
 
--- a/jetty-http/src/test/java/org/eclipse/jetty/http/HttpParserTest.java
+++ b/jetty-http/src/test/java/org/eclipse/jetty/http/HttpParserTest.java
@@ -2105,6 +2105,41 @@ public class HttpParserTest
     }
 
     @Test
+    public void testRequestMaxHeaderBytesURITooLong()
+    {
+        ByteBuffer buffer = BufferUtil.toBuffer(
+                "GET /long/nested/path/uri HTTP/1.1\r\n" +
+                "Host: example.com\r\n" +
+                "Connection: close\r\n" +
+                "\r\n");
+
+        int maxHeaderBytes = 20;
+        HttpParser.RequestHandler handler = new Handler();
+        HttpParser parser = new HttpParser(handler, maxHeaderBytes);
+
+        parseAll(parser, buffer);
+        assertEquals("414", _bad);
+    }
+
+    @Test
+    public void testRequestMaxHeaderBytesCumulative()
+    {
+        ByteBuffer buffer = BufferUtil.toBuffer(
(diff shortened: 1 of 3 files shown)


mszabo-wikia:
CI failed in the Java 11 branch with an apparent timeout

jetty-http/src/test/java/org/eclipse/jetty/http/HttpParserTest.java
```diff
@@ -2104,6 +2104,41 @@ public void testEmptyHostPort()
...
+    @Test
+    public void testRequestMaxHeaderBytesURITooLong()
+    {
+        ByteBuffer buffer = BufferUtil.toBuffer(
+                "GET /long/nested/path/uri HTTP/1.1\r\n" +
+                "Host: example.com\r\n" +
+                "Connection: close\r\n" +
+                "\r\n");
+
+        int maxHeaderBytes = 20;
```
sbordet: Using 20 makes not really clear that it's the URI too long that fails the tests. I would use 5, which makes that instantly clear.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConfiguration.java
```diff
@@ -193,7 +193,7 @@ public int getOutputAggregationSize()
         return _outputAggregationSize;
     }
 
-    @ManagedAttribute("The maximum allowed size in bytes for an HTTP request header")
+    @ManagedAttribute("The maximum allowed size in bytes for the combined content of all HTTP request headers and the request URI")
```
sbordet:
```suggestion
    @ManagedAttribute("The maximum allowed size in bytes for the HTTP request line and HTTP request headers")
```

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConfiguration.java
```diff
@@ -406,11 +406,13 @@ public void setOutputAggregationSize(int outputAggregationSize)
     }
 
     /**
+     * Set the maximum allowable size for the cumulative content of all request headers and the URI.
```
sbordet: - Wrap with `<p>`.
- Use third person for javadocs, so "Sets the maximum...".
- Use the same description as in the getter's `@ManagedAttribute`.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpConfiguration.java
```diff
@@ -406,11 +406,13 @@ public void setOutputAggregationSize(int outputAggregationSize)
...
+     *
      * <p>Larger headers will allow for more and/or larger cookies plus larger form content encoded
      * in a URL. However, larger headers consume more memory and can make a server more vulnerable to denial of service
      * attacks.</p>
      *
-     * @param requestHeaderSize the maximum size in bytes of the request header
+     * @param requestHeaderSize the maximum size in bytes for the cumulative content of all request headers and the URI
```
sbordet: Use the same description as in the getter's `@ManagedAttribute`.

sbordet (changes requested):

mszabo-wikia:
Thanks for the quick response! I've updated the branch accordingly.

sbordet (approved):

PR #5142, opened 2020-08-12
Issue #5133 - Improve ResourceFactory and Resource list handling

joakime:
* Now parsed by `WebAppContext` into `List<Resource>`
* Reintroduced `Resource.fromList()` from old PR.
* Refactored `ResourceFactory` to never return null and always throw an exception if unable to get/create/resolve the `Resource`
* `ResourceHandler.getResource(String)` cleaned up in light of Exception handling requirement
* `Resource.addPath(String)` implementations can
  never return null now

PR #7419, opened 2022-01-19
Update resource base example so it works

Artur-:
The code change was done in #5142


sbordet:
The PR looks good, but you need to comply with the contribution guidelines:
https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/CONTRIBUTING.md

Artur-:
I dealt with the ECA but I am not sure this change makes sense because it seems to still fail, just with a different error message
```
Unable to parse configuration of mojo org.eclipse.jetty:jetty-maven-plugin:11.0.7:run for parameter #: Cannot create instance of class org.eclipse.jetty.util.resource.Resource
```

Maybe I am just holding it wrong

joakime (approved):
LGTM

PR #7437, opened 2022-01-24
Void handle method #7424

### LINKED ISSUE #7424: Jetty-12 Handler.handle boolean return?
**Jetty version**
12

**Question**

The current proposed signature for a `Handler` in jetty-12 is:
```java
    boolean handle(Request request, Response response) throws Exception;
```
Where the `Request` class extends `Callback` and is used to asynchronously signal success or failure of handling the request.   The boolean return is used only to indicate that this handler will be handling the request and has undertaken to call `succeeded()` or `fail(x)`.

This allows us to keep a `Handler` structure similar to Jetty<=11 like:
```
 server
   +- StatsHandler
   |     +- ContextHandlerCollection
   |           +- ContextA
   |           |     +- SomeHandler
   |           +- ContextB
   |                 +- SessionHandler
   |                       +- ServletHandler
   +- ErrorHandler
```
The boolean return allows:
 + The `Server` to know that no `Handler` has accepted the request, and thus it needs to call the `ErrorHandler` to generate a 404.
 + The  `ContextHandlerCollection` to know if a context has accepted the request (by contextPath, virtual host and/or connector name)

The problem with this approach is what if the Handler needs to wait for something (eg external DB) to decide if it will handle the request or not?
If the handler blocks waiting to know if it will return true or not, then it will block all the handling.

Now currently, this is not a huge problem as we invoke handle with a blocking invokcation anyway, as it might be a ServletHandler that calls a servlet that blocks in the normal IO.   So whilst it will often be invoked with a selector thread, it will never be invoked with the last selector thread. Thus blocking will not block the server.

So two questions:
 1) Do we need to fix this?
 2) How do we fix this?

gregw:
request no longer extends Callback #7424
```java
interface Request
{
    Callback setHandling();
    boolean isHandling();
}

interface Handler
{
    void handle(Request request, Response response) throws Exception;
}
```

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java
@@ -43,13 +43,13 @@ import org.slf4j.LoggerFactory;
  * <ul>
  * <li>Do nothing</li>
  * <li>Completely generate the HTTP Response and call {@link Callback#succeeded()} on the {@link Callback}
- * returned from {@link Request#setHandling()}.</li>
- * <li>Call {@link Request#setHandling()} and arrange for an async process to generate the HTTP Response and call
+ * returned from {@link Request#accept()}.</li>
+ * <li>Call {@link Request#accept()} and arrange for an async process to generate the HTTP Response and call
  * {@link Callback#succeeded()} or {@link Callback#failed(Throwable)} on the {@link Callback} returned.</li>
  * <li>Pass the request to one or more other Handlers.</li>
  * <li>Wrap the request and/or response and pass them to one or more other Handlers.</li>
  * <li>Fail the request by calling {@link Callback#failed(Throwable)} on the {@link Callback} returned from
- * {@link Request#setHandling()}.</li>
+ * {@link Request#accept()}.</li>
  * </ul>
  *
  */
@@ -60,7 +60,7 @@ public interface Handler extends LifeCycle, Destroyable
     /**
      * Handle an HTTP request and produce a response.
      * @param request The immutable request, which is also a {@link Callback} used to signal success or failure. The Handler
-     * or one of it's nested Handlers must call {@link Request#setHandling()} to indicate that it will ultimately succeed or
+     * or one of it's nested Handlers must call {@link Request#accept()} to indicate that it will ultimately succeed or
      * fail the {@link Callback} returned.
      *
      * @param response The muttable response
@@ -340,7 +340,7 @@ public interface Handler extends LifeCycle, Destroyable
         {
             for (Handler h : _handlers)
             {
-                if (!request.isHandling())
+                if (!request.isAccepted())
                     h.handle(request, response);
             }
         }
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java
@@ -41,11 +41,15 @@ import org.slf4j.LoggerFactory;
  * <p>
  * A call to {@link #handle(Request, Response)} may:
  * <ul>
- * <li>return false to indicate that it will not handle the request</li>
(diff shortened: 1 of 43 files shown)


gregw:
This works pretty well and the branch is now passing all tests in jetty-server.

There is probably a tiny cost in performance as we do need to grab the lock on the request at least two extra times for every request.   So frequently instead of doing:
```java
    // ...
   request.succeeded();
   return true;
}
```
we now do
```java
    // ...
   request.setHandling().succeeded();
}
```

I think the second is clearer code (although the name is not that good).  Specifically it makes it clear that 2 actions are taking place: 1) a committment to handle the request by calling `setHandling` ; and 2) completing handling of the request with a call to succeeded.   Before the lock was just grabbed in `succeeded()`, but now it is grabbed in `setHandling()`, `succeeded()` and then again by the caller to check `isHandling()`.        @lorban it would be good to benchmark both branches to see if those extra locks are at all significant.  I'm expecting them to be down in the noise????  Potentially we could put just the handling field into a volatile to avoid the lock?

I'm not enthralled with the method names `setHandling()` and `isHandling()`:
 + `setHandling()` is not really a setter as nothing is passed.
 + `setHandling()` is almost a getter as it returns the `Callback`, but it does have a state change associated with calling it, so it is not a pure getter either.
 + `isHandling()` is close to the previous `isHandled()`, but the past tense doesn't work in the async case.

It is pretty easy to use and there are lots of simple cases liked:

```java
    @Override
    public void handle(Request request, Response response) throws Exception
    {
        response.setStatus(200);
        response.setContentType(MimeTypes.Type.TEXT_PLAIN_UTF_8.asString());
        response.write(true, request.setHandling(), BufferUtil.toBuffer("Hello World"));
    }
```

Another common style is:

```java
    @Override
    public void handle(Request request, Response response) throws Exception
    {
        response.setStatus(HttpStatus.MOVED_PERMANENTLY_301);
        response.getHeaders().add(new HttpField(HttpHeader.LOCATION, "/somewhere"));
        request.setHandling().succeeded();
    }
```

But it is rather easy to make at least one common mistake.  Spot the error:
```java
    @Override
    public void handle(Request request, Response response) throws Exception
    {
        response.setStatus(200);
        response.setContentType(MimeTypes.Type.TEXT_PLAIN_UTF_8.asString());
        response.write(true, Callback.from(
                () -> {somethingElse(); request.setHandling().succeeded();}
                t -> {request.setHandling().failed(t);}
            ), 
            BufferUtil.toBuffer("Hello World"));
    }
```

Maybe better naming would help?

On balance, I like it more that `Request implements Callback`, but only if we can show no significant extra locking cost and come up with some better names.

However, I can't see how we can avoid the no-callback-unless-handling paradox unless we keep the extra state of handling status, which needs extra locks.

**Name ideas:**

Current in branch:
```java
   if (request.isHandling())
       request.setHandling().failed(new AlreadyHandled());
   else
       response.write(true, request.setHandling(), someContent);
```
or perhaps
```java
   if (request.isHandling())
       request.handling().failed(new AlreadyHandled());
   else
       response.write(true, request.handling(), someContent);
```
or perhaps
```java
   if (request.isCallbackAccessed())
       request.callback().failed(new AlreadyHandled());
   else
       response.write(true, request.callback(), someContent);
```
or perhaps
???

gregw:
or perhaps:
```java
if (request.isAccepted())
       request.accept().failed(new AlreadyHandled());
   else
       response.write(true, request.accept(), someContent);
```

gregw:
Informal feedback from @sbordet : 
> We don't need this as it is still not fully asynchronous acceptance of handling a request and it creates a race condition where an async process may call `accept()` after  `isAccepted()` is checked

So currently both options have some bad aspect:
 + boolean return allows for a bad handler to return false, but still call `succeeded()`/`failed(x)` on the request.  It also has `Request implements Callback`, which is irksome (but perhaps fixable other ways).
 + The `accept()` method could be used in an async race with a call to `isAccepted()`.

More thought needed...

PR #7441, opened 2022-01-24
Issue #6282 - Allow SecuredRedirectHandler status code to be configurable

### LINKED ISSUE #6282: SecuredRedirectHandler should probably redirect with 301
**Jetty version**
10.0

**Description**
The status code used by SecuredRedirectHandler isn't configurable and currently redirects with 302. If there is no redirect handler set, jetty seems to redirect with 303 by default if the transport-guarantee is set to confidential in web.xml. (some services like the http-01 domain validation challenge used by letsencrypt are not following 303 redirects anymore)

Plain text http is slowly going away in the modern web and the recommended way to redirect from http to https is via 301 - permanently moved.

Consider making 301 the default for SecuredRedirectHandler, or allow it to be configured.

dellgreen:


joakime (approved):
LGTM

PR #7479, opened 2022-01-28
Jetty 12.0.x 7424 void handle 2

### LINKED ISSUE #7424: Jetty-12 Handler.handle boolean return?

gregw:
This is another alternative fix for #7424 that extends #7437 so that if the request is not accepted then:
 + request content cannot be read
 + the response can't be seen (and even if it could be, then content could not be written)
 + the callback cannot be seen and thus not succeeded nor failed.

A request can only be accepted once, so that if two handlers try to accept only 1 will succeed.  Furthermore, as the response is not available until the request is accepted, this prevents concurrent threads writing to the response, as only one will succeed in accepting the request.   Whilst on its own, it does not fully support async acceptance, there is a path to fully async acceptance if that proves necessary. This gives very safe handler usage, as bad states are simply prevented, by hiding the response and callback until the request is accepted.

The main changes can be summarized as:
```java
public interface Request 
{
    /**
     * @return The response instance or null if the request has already been accepted.
     */
    Response accept();
    // ...
}

public interface Handler
{
    void handle(Request request) throws Exception;
    // ...
}
```

This allows for simple handlers like:
```java
public class HelloHandler extends Handler.Abstract
{
    @Override
    public void handle(Request request) throws Exception
    {
        Response response = request.accept();
        response.setStatus(200);
        response.setContentType(MimeTypes.Type.TEXT_PLAIN_UTF_8.asString());
        response.write(true, response.getCallback(), BufferUtil.toBuffer("Hello World!"));
    }
}
```
Handlers can wrap requests as they always have done.  If they wish to wrap the response, they need to wrap the request and intercept the accept method.  However, this gives the benefit that the response is only wrapped if some nested handler accepts:
```java

public class GzipHandler extends Handler.Wrapper
{
    @Override
    public void handle(Request request) throws Exception
    {
        if (!request.getHeaders().contains(ACCEPT_GZIP))
        {
            super.handle(request);
            return;
        }

        HttpFields updated = HttpFields.from(request.getHeaders(), f ->
        {
            if (f.getHeader() != null)
            {
                if (CONTENT_ENCODING_GZIP.equals(f) || f.getHeader().equals(HttpHeader.CONTENT_LENGTH))
                    return null;
            }
            return f;
        });

        // wrap and handle request
        super.handle(
            new Request.Wrapper(request)
            {
                @Override
                public HttpFields getHeaders()
                {
                    return updated;
                }

                @Override
                public long getContentLength()
                {
                    return -1;
                }

                @Override
                public Content readContent()
                {
                    // TODO inflate data
                    return super.readContent();
                }

                @Override
                public Response accept()
                {
                    Response response = super.accept();
                    if (response != null)
                    {
                        return new Response.Wrapper(request, request.accept())
                        {
                            @Override
                            public void write(boolean last, Callback callback, ByteBuffer... content)
                            {
                                // TODO deflate data
                                super.write(last, callback, content);
                            }
                        };
                    }
                    return null;
                }
            }
        );
    }
}
```

This also allows atomic handling of exceptions that switch on if the request has been accepted or not, so rather than the anti pattern:
```java
      if (!request.isAccepted())
      {
          Response response = request.accept();
          response.setStatus(404);
          // ...
      }
```
We can do the atomic:
```java
      Response response = request.accept();
      if (response != null)
      {
          response.setStatus(404);
          // ...
      }
```

Finally, this is not fully asynchronous, but even in this form, you could concurrently call multiple handlers and only the first to accept it would be able to generate a response and complete the callback.  However, there is currently no mechanism to be notified of rejection, so it is not possible to know how long to wait before issuing a 404.   However, if we ever needed to go fully async, we could add the following method:

```java
public interface Request 
{
    Function<Boolean, Response> getAcceptor();
    // ...
}
```
A handler that has called `getAcceptor()` has indicated that it will eventually call the returned function with true/false to accept/reject the request.  If it passes true and no other handler has already accepted, then it gets the response to proceed with.  This could work in parallel with other simple handlers just calling the `accept()` method, and only a few classes would need to be modified to know if there were any outstanding Acceptors and thus to wait for a accept/reject before proceeding.

### DIFF
--- a/jetty-connectors/jetty-http2/http2-server/src/main/java/org/eclipse/jetty/http2/server/HttpChannelOverHTTP2.java
+++ b/jetty-connectors/jetty-http2/http2-server/src/main/java/org/eclipse/jetty/http2/server/HttpChannelOverHTTP2.java
@@ -689,7 +689,7 @@ public class HttpChannelOverHTTP2 extends HttpChannel implements Closeable, Writ
     {
         ContextHandler context = getState().getContextHandler();
         if (context != null)
-            context.handle(getRequest(), this);
+            context.handle(getRequest());
         else
             handle();
     }
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java
@@ -33,13 +33,10 @@ import org.slf4j.LoggerFactory;
  * <p>
  * Incoming requests to the Server (itself a Handler) are passed to one or more Handlers
  * until the request is handled and a response is produced.  Handlers are asynchronous,
- * so handling may happen during or after a call to {@link #handle(Request, Response)}.
- * A handler indicates that returns true from the {@link #handle(Request, Response)} method
- * is indicated that it (or one of it's contained handlers) has undertaken to produce the
- * response and ultimately call {@link Request#succeeded()} or {@link Request#failed(Throwable)}
- * to indicate the end of the request handling.
+ * so handling may happen during or after a call to {@link #handle(Request)}.
+ *
  * <p>
- * A call to {@link #handle(Request, Response)} may:
+ * A call to {@link #handle(Request)} may:
  * <ul>
  * <li>Do nothing</li>
  * <li>Completely generate the HTTP Response and call {@link Callback#succeeded()} on the {@link Callback}
@@ -63,10 +60,9 @@ public interface Handler extends LifeCycle, Destroyable
      * or one of it's nested Handlers must call {@link Request#accept()} to indicate that it will ultimately succeed or
      * fail the {@link Callback} returned.
      *
-     * @param response The muttable response
      * @throws Exception Thrown if there is a problem handling.
      */
-    void handle(Request request, Response response) throws Exception;
+    void handle(Request request) throws Exception;
 
     @ManagedAttribute(value = "the jetty server for this handler", readonly = true)
     Server getServer();
@@ -286,11 +282,11 @@ public interface Handler extends LifeCycle, Destroyable
         }
 
         @Override
-        public void handle(Request request, Response response) throws Exception
+        public void handle(Request request) throws Exception
         {
             Handler next = getHandler();
             if (next != null)
-                next.handle(request, response);
+                next.handle(request);
         }
     }
 
@@ -318,7 +314,7 @@ public interface Handler extends LifeCycle, Destroyable
 
     /**
      * A Handler Container that wraps a list of other Handlers.
-     * By default, each handler is called in turn until one returns true from {@link Handler#handle(Request, Response)}.
+     * By default, each handler is called in turn until one returns true from {@link Handler#handle(Request)}.
      */
     class Collection extends AbstractContainer
     {
@@ -336,12 +332,12 @@ public interface Handler extends LifeCycle, Destroyable
         }
 
         @Override
-        public void handle(Request request, Response response) throws Exception
+        public void handle(Request request) throws Exception
         {
             for (Handler h : _handlers)
             {
                 if (!request.isAccepted())
-                    h.handle(request, response);
+                    h.handle(request);
             }
         }
 
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java
@@ -43,13 +43,13 @@ import org.slf4j.LoggerFactory;
  * <ul>
  * <li>Do nothing</li>
  * <li>Completely generate the HTTP Response and call {@link Callback#succeeded()} on the {@link Callback}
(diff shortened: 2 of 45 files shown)


lorban:
I find this API very elegant as it follows the path of the API restricting invalid usages. :+1: 

In the form you're proposing, it still allows reading the input before the request is accepted so we could take it a step even further by making both the response and the request only available after the acceptation:

```java
public class HelloHandler extends Handler.Abstract
{
    @Override
    public void handle(MetaData.Request metadata) throws Exception
    {
        Exchange exchange = metadata.accept();

        Request request = exchange.getRequest();
        IO.copy(request.getInputStream(), "/dev/null");

        Response response = exchange.getResponse();
        response.setStatus(200);
        response.setContentType(MimeTypes.Type.TEXT_PLAIN_UTF_8.asString());
        response.write(true, response.getCallback(), BufferUtil.toBuffer("Hello World!"));
    }
}
```

I didn't think through all the details so maybe my example is broken but WDYT of pushing your idea to its logical conclusion?

gregw:
@lorban firstly more that whilst the read API is available, it is not active until accept.
Also we often need more info than just the metada to accept, including connectionMetaData and request attributes.... Ie all of the request except to the Io.  So it's not so bad to just have them disabled until accept.

I had to change a few handlers that were reading first and then accepting, so it does work.

I have also tried using exchange objects a few times in various ideas and they do make a wrap heavy design more cumbersome

gregw:
So the only badness this API allows that the boolean return did not, is accepting after returning from the handle method.   But ultimately, even if there is a race to accept, at least it is a safe race and only the winner can get the response and the losers know they lost.

gregw:
@lachlan-roberts, @janbartel  note that an interesting aspect of this approach is how it will affect the session/security/servlet handlers:
+ Currently, at the context level, we do not assume a request will be accepted, but our ServletHandler by default will always handle a request, thus it will always accept.
+ Also, the security handlers might need to accept a request in order to send a challenge.  
+ The session handler may need to add cookies to the response, but can only get the response if it is accepted.

I think all this can be done OK, but we can also give some thought to accepting the request at the context level, and then it is always available to nested handlers via `request.getResponse()`.  However, such handlers would then always need to be deployed behind an accepting handler.

lorban:
@gregw my point is while this PR is about changing the API so that the `Response` cannot be accessed before accepting the `Request`, why not going one step further and also preventing the input from being consumed _at the API level_ before the acceptation so that there's no need to rely on runtime checks?

gregw:
@lorban yep. I'm all for that, but only if we don't make the API more complex than necessary to achieve it.   I think this PR works nicely as we didn't need to introduce any extra classes and the API flows naturally from Request to Response to Callback.
Now this almost works as a natural flow from MetaData to Request to Response to Callback, but it turns out we need more than MetaData we need ConnectionMetaData and Attributes.  Also accepting the request enough to read the content means that the response is also available at the same stage.   We thus get (MetaData.Request, ConnectionMetaData, Attributes) to (Request, Response) to Callback, so we end up needing holder classes like `Exchange`.    Such holder classes in a wrap-centric architecture become a PITA really quickly.

If we can find a way without holder classes, I'm more interested.

gregw:
@sbordet @lorban I like this approach enough that I want to merge it into my current 12.0.x branch, which will become 12.0.x-experiment branch once the NG branch is ready to go.
OK if I merge?   All things still up for debate.

lorban:
Sure, go ahead merging it. This proposal may not be final but my gut feeling is that it's a big step in a good direction.

PR #7503, opened 2022-01-31
Jetty 9.4.x : fix tries mistakenly throwing ArrayIndexOutOfBoundsException

### LINKED ISSUE #7496: Transient 400: Bad Request responses
jetty-9.4.45.v20220128 Compact 3 (Not yet released)

java -version
openjdk version "1.8.0_322"
OpenJDK Runtime Environment (Zulu 8.60.0.22-SA-qnx_aarch32hf) (build 1.8.0_322-b06, profile compact3)
OpenJDK Client VM (Zulu 8.60.0.22-SA-qnx_aarch32hf) (build 25.322-b06, mixed mode)

QNX 7.0 ARMLEv7

I understand that this version is not yet released, but I have done some early testing against the 9.4.45.v20220128 tag made yesterday in anticipation of adopting the release as soon as it is made available. I am seeing a substantial regression in our application behavior where a 400 Bad Request response is being returned intermittently. The only change in our runtime environment is that the 9.4.45 Jetty version (jetty-all-compact3-9.4.45.v20220128.jar) replaces the 9.4.44 Jetty version (jetty-all-compact3-9.4.44.v20210927.jar) in our classpath. Sometimes the request goes through normally (200 response), and other times the page fails to load, or partially loads (with 400 response). The page always loads normally with 9.4.44.

The request in question is HTTP so no HTTPS/TLS negotiation is a factor.

I apologize that this description is vague, and that this 9.4.45 version might not be released yet if still a work-in-progress, but I wanted to provide early feedback to see if this change in behavior rings any bells about what may have changed between 9.4.44 and 9.4.45.20220128 that might bring about this behavior (and prevent our uptake of 9.4.45). Is there any particular release note I should be paying attention to that might require change in our application configuration? We've been using the 9.4.X branch since 9.4.26 and have not seen this type of behavior until testing with recent 9.4.45.

Thanks, please let me know if you have any questions that might help me clarify this behavior.

lorban:
* Add missing `ByteBuffer` limit checks in `get()` and `getBest()` to make sure that `ArrayIndexOutOfBoundsException` is never going to be thrown.
* Never write an index that points out of the array in `put()` by always incrementing the row count before checking the limit but capping it at its max value.
* Make sure the row count cannot overflow `Character.MAX_VALUE`.


jetty-util/src/main/java/org/eclipse/jetty/util/ArrayTrie.java
```diff
@@ -190,9 +190,10 @@ else if (c > 127)
                 t = big[c];
                 if (t == 0)
                 {
+                    _rows = (char)Math.min(_rows + 1, _value.length);
                     if (_rows == _value.length)
                         return false;
-                    t = big[c] = ++_rows;
+                    t = big[c] = _rows;
```
gregw: Can we make the small index logic exactly the same as this.

gregw (changes requested):

jetty-util/src/main/java/org/eclipse/jetty/util/ArrayTrie.java
gregw: I pushed this change.

sbordet (approved):

gregw (commented):
LGTM (I know too late if it didn't :)

PR #7505, opened 2022-01-31
Removing jetty-maven-plugin IT test javax-annotation-api

joakime:
Fixing `jetty-9.4.x` branch build.

The javax.annotation-api 1.3 dependency conflicts with the maven built-in jsr250-api 1.0 dependency.
Removing the jsr dependency from `<extensions>` fixes this build issue.


olamy (dismissed):

olamy:
due to upgrade to maven 3.8.3 see https://issues.apache.org/jira/browse/MNG-7045

olamy:
this will still fail with maven 3.8.2-
We need to add a selector  `invoker.maven.version = 3.8.3+`
see a3d0c13802646d1d2201a8d0957520fb73a4a399

olamy (dismissed):

janbartel:
Do we even need this test any more if we have upgraded the jetty build to a suitably recent version that doesn't use the old jsr annotations jar?

janbartel (changes requested):
See my comment - why do we even need this test any more?

olamy:
Yup I agree we can get rid of it it was just an example of workaround for Java 1.8 and maven 3.8.2-

joakime:
@janbartel @olamy just removed that test entirely in this PR.

olamy (approved):

janbartel (approved):

PR #7506, opened 2022-02-01
Issue #7414 Deprecate QoSFilter setters

### LINKED ISSUE #7414: QoSFilter.setMaxRequests throws NullPointerException
**Jetty version(s)**
9.4.44
openjdk version "11.0.10" 2021-01-19 LTS

**OS type/version**
macOS 11.6.2

**Description**
I'm trying to use the QosFilter into an existing Spring application but if I try to tune the filter calling `setMaxRequests` at configuration time, it throws the following NullPointer exception:
```
Caused by: java.lang.NullPointerException: null
	at org.eclipse.jetty.servlets.QoSFilter.setMaxRequests(QoSFilter.java:351)
```
this is the code I'm using to create a Spring Bean with the QosFilter:
```
@Bean
public QoSFilter sfdcJettyQualityOfServiceFilter(MyQoSProperties properties) {
    QoSFilter qoSFilter = new QoSFilter();
    qoSFilter.setSuspendMs(properties.getSuspendMs());
    qoSFilter.setWaitMs(properties.getWaitMs());
    qoSFilter.setMaxRequests(properties.getMaxRequests());

    return qoSFilter;
}
```
I think it blows up because `QosFilter.init()` method has not been called yet at that stage and the private property `_passes` is null [here](https://github.com/eclipse/jetty.project/blob/25306366d4b5d2cdea3eea07861a5d3d145683d6/jetty-servlets/src/main/java/org/eclipse/jetty/servlets/QoSFilter.java#L345).

**How to reproduce?**
- creating a SpringBoot application and trying to load the QosFilter in a `@Configuration` annotated class

janbartel:
Closes #7414 

The OP reported that calling `QoSFilter.setMaxRequests(int)`  throws NPE. This is because it uses a field that is only initialized _after_ `init()` is called. The filter seems to expect that values are set via `init-params`, not via setters. Certainly there is no code that would handle one of the setters being called whilst the filter is in service. The best way forward seems to be to eventually remove those setters, after being `deprecated`.

### DIFF
--- a/jetty-servlets/src/main/java/org/eclipse/jetty/servlets/QoSFilter.java
+++ b/jetty-servlets/src/main/java/org/eclipse/jetty/servlets/QoSFilter.java
@@ -293,7 +293,9 @@ public class QoSFilter implements Filter
      * for the semaphore to become available before suspending a request.
      *
      * @param value wait time (in milliseconds)
+     * @deprecated do not use
      */
+    @Deprecated
     public void setWaitMs(long value)
     {
         _waitMs = value;
@@ -316,7 +318,9 @@ public class QoSFilter implements Filter
      * a request for while waiting for the semaphore to become available.
      *
      * @param value suspend time (in milliseconds)
+     * @deprecated do not use
      */
+    @Deprecated
     public void setSuspendMs(long value)
     {
         _suspendMs = value;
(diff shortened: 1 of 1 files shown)


jetty-servlets/src/main/java/org/eclipse/jetty/servlets/QoSFilter.java
```diff
@@ -293,7 +293,9 @@ public long getWaitMs()
      * for the semaphore to become available before suspending a request.
      *
      * @param value wait time (in milliseconds)
+     * @deprecated do not use
```
gregw: Should we say use init parameters instead?

gregw (changes requested):
Should we make the bodies of these methods noops to avoid the NPEs etc.?

joakime (approved):
LGTM

gregw (approved):

Issue #6017, opened 2021-02-28
Property overriding does not work


sbordet:
**Jetty version**
10.0.x

**Description**
These commands:

```
$ java -jar $JETTY_HOME/start.jar --add-modules=ssl,https,test-keystore
$ java -jar $JETTY_HOME/start.jar jetty.sslContext.keyStorePath=bogus
```

produce:

```
java.lang.IllegalStateException: /tmp/jetty.base/etc/test-keystore.p12 is not a valid keystore
```

The wrong property value is reported in the error message.

But furthermore:

```
$ java -jar $JETTY_HOME/start.jar jetty.sslContext.keyStorePassword=bogus
```

starts Jetty without errors.

There is something wrong in how properties are handled.

sbordet:
Works properly in Jetty 9.4.x.

sbordet:
@joakime any idea?

joakime:
Isn't this related to #5915 ?

sbordet:
@joakime maybe -- the `test-keystore` module does have an `[ini]` section that seems to affect this issue and also #5990.

joakime:
Could be, the `--list-config` shows ...

```
Enabled Modules:
----------------
    0) resources       transitive provider of resources for logging-jetty
    1) logging/slf4j   transitive provider of logging/slf4j for logging-jetty
                       dynamic dependency of logging-jetty
    2) logging-jetty   transitive provider of logging for threadpool
                       transitive provider of logging for bytebufferpool
                       transitive provider of logging for server
    3) bytebufferpool  transitive provider of bytebufferpool for server
                       init template available with --add-module=bytebufferpool
    4) threadpool      transitive provider of threadpool for server
                       init template available with --add-module=threadpool
    5) server          transitive provider of server for ssl
                       init template available with --add-module=server
    6) ssl             ${jetty.base}/start.d/ssl.ini
    7) https           ${jetty.base}/start.d/https.ini
    8) test-keystore   ${jetty.base}/start.d/test-keystore.ini

...(snip)...

Properties:
-----------
 bouncycastle.version = 1.62
 java.version = 11.0.10
 java.version.major = 11
 java.version.micro = 10
 java.version.minor = 0
 java.version.platform = 11
 jetty.base = /home/joakim/code/jetty/distros/bases/prop-base
 jetty.base.uri = file:///home/joakim/code/jetty/distros/bases/prop-base
[...]

PR #7511, opened 2022-02-01
Issue #6017 - backport module property override PR

joakime:
This is a backport of changes in `jetty-10.0.x` to `jetty-9.4.x`


joakime:
Closing.

This feature doesn't need to be backported.
Simply using `<k>?=<v>` instead in your module files will give the same end result.

PR #7522, opened 2022-02-02
Fixes #7514 - Adding InheritedListeners to already-started components…

### LINKED ISSUE #7514: Adding InheritedListeners to already-started components can cause IllegalStateException
**Jetty 10.0.x**

When a webapp is hot-deployed to an already-started jetty via the `deploy` module, the `o.e.j.deploy.bindings.StandardStarter` does:

```
        if (contexts.isStarted() && handler.isStopped())
        {
            // start the handler manually
            handler.start();

            // After starting let the context manage state
            contexts.manage(handler);
        }
```

So the webapp is first started, and then the `ContextHandlerCollection` is told to manage it.  During the call to `manage`, if an `o.e.j.u.component.Container.InheritedListener` is registered with jetty, then it will be added to every component of the newly started webapp.

If one of those components happens to be a `o.e.j.io.ClientConnector`, when the `InheritedListener` is registered with the connector's `o.e.j.io.ClientConnector.ClientSelectorManager`, then the call to `addEventListener(EventListener)` will fail because the `ClientConnector.ClientSelectorManager` is _already_ in the `started` state, see:
https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-io/src/main/java/org/eclipse/jetty/io/SelectorManager.java#L399.

This does _not_ happen if the webapp is deployed during startup of jetty, because none of the component hierarchy is already in the `started` state, it is just `starting`.

This can be reproduced in the standard jetty distro by:

1. enabling the `jmx` module (which has an `InheritedListener`)
2. enabling the `deploy` module, along with `http` and `websocket`
3. starting jetty the usual way
4. deploying the `cometd-6.0.5` demo webapp (which will contain a `ClientConnector` and `ClientSelectorManager`)

sbordet:
… can cause IllegalStateException

Removed the unnecessary check-and-throw statements from SelectorManager.

### DIFF
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/SelectorManager.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/SelectorManager.java
@@ -396,8 +396,6 @@ public abstract class SelectorManager extends ContainerLifeCycle implements Dump
     @Override
     public boolean addEventListener(EventListener listener)
     {
-        if (isRunning())
-            throw new IllegalStateException(this.toString());
         if (super.addEventListener(listener))
         {
             if (listener instanceof AcceptListener)
@@ -410,8 +408,6 @@ public abstract class SelectorManager extends ContainerLifeCycle implements Dump
     @Override
     public boolean removeEventListener(EventListener listener)
     {
-        if (isRunning())
-            throw new IllegalStateException(this.toString());
         if (super.removeEventListener(listener))
         {
             if (listener instanceof AcceptListener)
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/SelectorManager.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/SelectorManager.java
@@ -22,10 +22,10 @@ import java.nio.channels.SelectionKey;
 import java.nio.channels.Selector;
 import java.nio.channels.ServerSocketChannel;
 import java.nio.channels.SocketChannel;
-import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


jetty-io/src/main/java/org/eclipse/jetty/io/SelectorManager.java
```diff
@@ -396,8 +396,6 @@ protected void connectionFailed(SelectableChannel channel, Throwable ex, Object
     @Override
     public boolean addEventListener(EventListener listener)
     {
-        if (isRunning())
```
gregw, line 399: I think this theoretically exposes us to `ConcurrentModificationException`s

jetty-io/src/main/java/org/eclipse/jetty/io/SelectorManager.java
```diff
@@ -410,8 +408,6 @@ public boolean addEventListener(EventListener listener)
     @Override
     public boolean removeEventListener(EventListener listener)
     {
-        if (isRunning())
```
gregw, line 413: I think this definitely exposes us to `ConcurrentModificationException`s if the selector is iterating over events when this is called.

gregw (changes requested):
We need to protect against concurrent modifications whilst iterating over the listeners.  Maybe a CoW list?

sbordet:
Merged to `jetty-12.0.x` on 3 Feb 2022, commit 40378616681e8b8664b4fcdab7cc4efe1ce64244.

PR #7525, opened 2022-02-02
Fixes #7524 - Missing package in JmxConfiguration

### LINKED ISSUE #7524: Missing package in JmxConfiguration
**Jetty version(s)**
10.0.x

**Description**
`JmxConfiguration` protects and exposes the Jetty JMX package `org.eclipse.jetty.jmx` , but it should also protect and expose `org.eclipse.jetty.util.annotation` because that's where the JMX annotations such as `@ManagedAttribute` are.

Failing to expose `org.eclipse.jetty.util.annotation` causes web applications that have `jetty-util` in `WEB-INF/lib` to use the web application version, which means that the server mechanism to register MBeans won't be able to find attributes and operations (because it is searching for server-loaded `@ManagedAttribute`, while instead the web application classes are annotated with webapp-loaded `@ManagedAttribute`.

sbordet:
Also protectAndExpose "org.eclipse.jetty.util.annotation" to make JMX annotations loaded by the server available to web applications.
This allows the server JMX mechanism (MBeanContainer) to work on web application classes too.

### DIFF
--- a/jetty-webapp/src/main/java/org/eclipse/jetty/webapp/JmxConfiguration.java
+++ b/jetty-webapp/src/main/java/org/eclipse/jetty/webapp/JmxConfiguration.java
@@ -32,7 +32,7 @@ public class JmxConfiguration extends AbstractConfiguration
     public JmxConfiguration()
     {
         addDependents(WebXmlConfiguration.class, MetaInfConfiguration.class, WebInfConfiguration.class);
-        protectAndExpose("org.eclipse.jetty.jmx.");
+        protectAndExpose("org.eclipse.jetty.util.annotation", "org.eclipse.jetty.jmx.");
     }
 
     @Override


jetty-webapp/src/main/java/org/eclipse/jetty/webapp/JmxConfiguration.java
```diff
@@ -32,7 +32,7 @@
     public JmxConfiguration()
     {
         addDependents(WebXmlConfiguration.class, MetaInfConfiguration.class, WebInfConfiguration.class);
-        protectAndExpose("org.eclipse.jetty.jmx.");
+        protectAndExpose("org.eclipse.jetty.util.annotation", "org.eclipse.jetty.jmx.");
```
gregw, line 35: This is correctly restoring the jetty-9 behaviour, so good to proceed with.

BUT ultimately we should avoid exposing jetty classes like this.  So we should consider an option of how we can use Jmx in the server and not expose to the webapp: either by making this line conditional or moving into another alternate optional configuration (which would need it's own jar).

gregw (approved):

PR #7527, opened 2022-02-03
Fix #7518 Trie.getBest with empty Key

### LINKED ISSUE #7518: ArrayTrie getBest fails to match the empty string entry in certain cases
**Jetty version(s)**
9.4.x

**Description**

When an `ArrayTrie` contains an empty string, `getBest()` should always match it as a last resort. But when `getBest()` is given a key that starts with common letters but does not match an existing entry then `null` is returned instead of the empty string's value.

Consider the following:

```java
ArrayTrie<Integer> trie = new ArrayTrie<>();

assertTrue(trie.put("", 0));
assertTrue(trie.put("prefix", 1));

assertEquals(0, trie.getBest("unknown"));
assertEquals(0, trie.getBest("pre-unknown")); // fails
```

gregw:
Fixes #7518 
Only increment current row if not recursing.

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/ArrayTrie.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/ArrayTrie.java
@@ -307,7 +307,7 @@ public class ArrayTrie<V> extends AbstractTrie<V>
             {
                 char[] big = _bigIndex == null ? null : _bigIndex[t];
                 if (big == null)
-                    return null;
+                    break;
                 nt = big[c];
                 if (nt == 0)
                     break;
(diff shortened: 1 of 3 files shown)


lorban (changes requested):
As for #7528 `HttpParserTest` got broken.

gregw:
@lorban Added extra test. The problem was a bug with the return if the first non matching character was a big index one.

gregw:
@lorban Added extra test. The problem was a bug with the return if the first non matching character was a big index one.

lorban (approved):
It looks like #7528 duplicates these changes, otherwise LGTM.

PR #7532, opened 2022-02-04
Issue #7513 Move jmx annotation to PropertiesConfigurationManager.getFile

### LINKED ISSUE #7513: Getter/setter type mismatch for mbean attribute file in class org.eclipse.jetty.deploy.PropertiesConfigurationManager
**Jetty 10.0.x**

Enable `jmx` module on jetty-10, results in this log at startup:
2022-02-02 18:30:26.063:INFO :oejj.MetaData:main: Getter/setter type mismatch for mbean attribute file in class org.eclipse.jetty.deploy.PropertiesConfigurationManager

janbartel:
Closes #7513

The `PropertiesConfigurationManager` class had a jmx attribute annotation on `setFile` method instead of `getFile` method.

### DIFF
--- a/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/PropertiesConfigurationManager.java
+++ b/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/PropertiesConfigurationManager.java
@@ -57,7 +57,6 @@ public class PropertiesConfigurationManager implements ConfigurationManager, Dum
         this(null);
     }
 
-    @ManagedAttribute("A file or URL of properties")
     public void setFile(String resource) throws IOException
     {
         _properties = resource;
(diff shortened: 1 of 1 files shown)


janbartel:
@sbordet whilst I've moved the annotation to the getter instead of the setter, which fixes the immediate problem, do you think we need a check for this situation?  If the annotation is put on the _setter_ instead of the _getter_, then https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-jmx/src/main/java/org/eclipse/jetty/jmx/MetaData.java#L441  is passed the _setter_ instead of the _getter_, and so  line https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-jmx/src/main/java/org/eclipse/jetty/jmx/MetaData.java#L458 incorrectly compares the return type of the _setter_ (which is always `void`) against the type of the arg to the setter.

sbordet (approved):

sbordet:
@janbartel I opened #7561.

Issue #7529, opened 2022-02-03
Upgrade quiche to version 0.11.0


lorban:
**Target Jetty version(s)**
10.0.x / 11.0.x

**Enhancement Description**
Quiche version 0.11.0 got released, we need to build it for our supported platforms and upgrade to that version of the native lib.

PR #7534, opened 2022-02-04
 #7529 upgrade quiche native lib to version 0.11.0.1 to work on less recent linux libc

lorban:
The previous Linux native build was made against a too recent version of libc which fails to load on less recent systems.


sbordet (changes requested):
Let's use 0.11.0.1 as our own "fix" versions.

sbordet (approved):

PR #7556, opened 2022-02-09
Fixing Jetty 10 release - must use Java 17

joakime:
+ `/javadoc/` (aggregate) module now has missing `jdk.incubator.foreign` 
+ the `-Peclipse-release` profile now enforces Java 17 minimum version.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,12 +7,12 @@ pipeline {
   stages {
     stage("Parallel Stage") {
       parallel {
-        stage("Build / Test - JDK11") {
+        stage("Build / Test - JDK17") {
           agent { node { label 'linux' } }
           steps {
             container('jetty-build') {
               timeout( time: 120, unit: 'MINUTES' ) {
-                mavenBuild( "jdk11", "clean install -Perrorprone", "maven3")
+                mavenBuild( "jdk17", "clean install -Perrorprone", "maven3")
                 // Collect up the jacoco execution results (only on main build)
                 jacoco inclusionPattern: '**/org/eclipse/jetty/**/*.class',
                        exclusionPattern: '' +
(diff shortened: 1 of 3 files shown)


sbordet (changes requested):
Please verify that `Jenkinsfile` is correct and that the CI builds do the javadocs with 17 too (and perhaps drop those done with 11).
I think it's time to make the 17 build the main one (with JaCoCo, etc.) and run 11 only as a secondary.

joakime:
> Please verify that `Jenkinsfile` is correct and that the CI builds do the javadocs with 17 too

I did this already, downloaded the entire build logs and verified that the javadoc is executing.
But now that I think of it, the change to `maven-compiler-plugin` on `/javadoc/pom.xml` is likely going to break the java 11 build.
I need to fix that.

> (and perhaps drop those done with 11). I think it's time to make the 17 build the main one (with JaCoCo, etc.) and run 11 only as a secondary.

I'll check what it takes to get jacoco running on Java 17.
Not sure it's been updated yet.

joakime:
@sbordet jacoco is working on JDK 17 now.
The javadoc failures on jdk17 are easiest fixed by not including quic-quiche-foreign-incubator in the javadoc aggregate in the first place.

Jenkinsfile
```diff
@@ -12,7 +12,7 @@ pipeline {
           steps {
             container('jetty-build') {
               timeout( time: 120, unit: 'MINUTES' ) {
-                mavenBuild( "jdk17", "clean install -Perrorprone", "maven3")
+                mavenBuild( "jdk17", "clean install", "maven3")
```
olamy, line 15: it's documented as supported now with jdk17.
but yeah can be fix in a later PR

olamy (dismissed):

Jenkinsfile
```diff
@@ -31,19 +31,19 @@ pipeline {
...
-        stage("Build / Test - JDK17") {
+        stage("Build / Test - JDK11") {
           agent { node { label 'linux' } }
           steps {
             container( 'jetty-build' ) {
               timeout( time: 120, unit: 'MINUTES' ) {
-                mavenBuild( "jdk17", "clean install -Dspotbugs.skip=true -Djacoco.skip=true", "maven3")
-                recordIssues id: "jdk17", name: "Static Analysis jdk17", aggregatingResults: true, enabledForFailure: true, tools: [mavenConsole(), java(), checkStyle(), spotBugs(), pmdParser()]
+                mavenBuild( "jdk11", "clean install -Dspotbugs.skip=true -Djacoco.skip=true -Perrorprone", "maven3")
+                recordIssues id: "jdk11", name: "Static Analysis jdk11", aggregatingResults: true, enabledForFailure: true, tools: [mavenConsole(), java(), checkStyle(), errorProne()]
```
sbordet, line 46: `spotbugs()` and `pmdParser()` have been removed from both builds. On purpose?
If so, do we have sections/dependencies in the POM that we need to remove as well?

joakime: pmd is no longer used in build, it shouldn't have been looked fro in the recordIssues().

spotbugs is only done for the first build, which is now jdk17.
However, spotbugs, like errorprone, does not work for jdk17 (yet).

We will fix spotbugs and errorprone in later PRs.

sbordet (approved):

Issue #6184, opened 2021-04-16
Remove usages of classes associated with JEP-411 that deprecate/remove the SecurityManager from the JVM


joakime:
**Jetty version**
All

**Java version**
All

**OS type/version**
All

**Description**
JEP-411 - https://openjdk.java.net/jeps/411

Will deprecate the SecurityManager in a specific way with an eye on removing it entirely in the future.
We should start to either alter our usages and expectations of the SecurityManager, or remove our usages entirely.

One of the techniques JEP-411 is considering is to have `System.getSecurityManager()` always return null.

That could impact our code when we use it like this ...

https://github.com/eclipse/jetty.project/blob/d825299da47a1d923648c52c80df11ac76fed77c/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java#L2529

Another aspect to consider is that `AccessController.doPrivileged(...)` will have no meaning anymore, the action will always be run in the JEP-411 deprecated mode.
Once JEP-411 enters removal mode, our code using AccessController will start to fail.

sbordet:
@janbartel please use this issue as an umbrella for other issues if you do any work related to the `SecurityManager`.

janbartel:
Even if we removed all calls to SecurityManager in jetty itself, it may still be present in 3rd party jars. For example, the JASPI api still refers to `SecurityManager`: https://github.com/jakartaee/authentication/blob/master/api/src/main/java/jakarta/security/auth/message/config/AuthConfigFactory.java#L157

sbordet:
@janbartel sure, but that's on them to fix JASPI.

janbartel:
@sbordet what I'm saying is that the JASPI api refers to SecurityExceptions that should be thrown if the caller does not have the correct permission. The api impl calls SecurityManager and expects that subclasses will do the same.

joakime:
@janbartel I went ahead and opened https://github.com/jakartaee/authentication/issues/138 about that specific case.

joakime:
The Jakarta EE platform level is tracking this requirement as well.

https://github.com/eclipse-ee4j/jakartaee-platform/issues/406

janbartel:
I think one of the biggest stumbling blocks for JEP-411 will be the fixes for #5859. The fix for the classloader pinning by the constructor for Thread requires the use of some java.security classes. We will need to support both pre-JEP-411 runtimes and post JEP-411 runtimes (possibly with reflection based off the runtime version?).

[...]

PR #7562, opened 2022-02-10
Removing SecurityManager in prep of JEP411

joakime:
Removing the usages of deprecated (and slated for removal) methods and classes defined in JEP-411.


sbordet (commented):
There are `SecurityManager` direct usages that it's ok to remove; fine with that.

There are `doPrivileged()` calls that I think were only used to "hide" operations that required permissions (e.g. load the logging properties) and that is ok to remove; fine with that.

There are `doPrivileged()` calls that were done to "cut" the `AccessControlContext` that was captured, which otherwise was retaining `ClassLoader` instances; I don't think it's ok to remove those, but I may be wrong.
If it's not ok, then perhaps try reflection?

@janbartel?

janbartel:
@sbordet @joakime see my comment over here https://github.com/eclipse/jetty.project/issues/6184#issuecomment-1035620780.

The `java.security` classes were essential in fixing the implementation of the `Thread`' constructor pinning of the classloader.

sbordet:
@janbartel I agree, `Thread` pinning the `ClassLoader` is category 3 of my comment above and I think it should remain, so that people using Java 17 will get the right effect, and those using a Java version without `SecurityManager` will have a modified `Thread` class that does not pin the `ClassLoader`, hopefully.

sbordet:
@janbartel looking at this in more details, I am dubious about the fix for the `ClassLoader` pinning.

See how `Executors.privilegedThreadFactory()` is doing things differently from our `PrivilegedThreadFactory`, in particular it does not replace the `AccessControlContext`, which pins a `ProtectionDomain` which pins the `ClassLoader`.

janbartel:
@sbordet `Executors.privilegedThreadFactory` will pin the `AccessControlContext`, thus the `ProtectionDomain` and the `ClassLoader` to be the same as the calling `Thread`.  Thus, depending on when jetty called `Executors.privilegedThreadFactory`, we could be just pinning the server class loader, which would be fine. However, if it was called by code that was _already_ in a context, then it would pin the context's classloader. 

Our solution executes in a privileged block, which means that the `AccessControlContext` is effectively not inherited, thus the `ProtectionDomain` and attendant `ClassLoader` cannot be pinned.

joakime:
@janbartel the `AccessController` is one of the classes deprecated by [JEP-411](https://openjdk.java.net/jeps/411), it's role in JDK-17 is a mere shell now, it barely does anything anymore as you cannot have a SecurityManager (most of the remaining methods now do nothing because of this, compared to previous JDKs).
A future JDK will just delete/remove `AccessController` entirely.

joakime:
Way out of date.
Will redo if there is a desire for it.
Closing.

PR #7563, opened 2022-02-10
Fixes #7548 - Interrupt flag is not always cleared in between requests.

### LINKED ISSUE #7548: Interrupt flag is not always cleared in between requests
**Jetty version(s)**

`9.4.44.v20210927`

```
openjdk version "11.0.14" 2022-01-18 LTS
OpenJDK Runtime Environment Zulu11.54+23-CA (build 11.0.14+9-LTS)
OpenJDK 64-Bit Server VM Zulu11.54+23-CA (build 11.0.14+9-LTS, mixed mode)
```

**OS type/version**

```
Linux fedora 5.16.5-200.fc35.x86_64 #1 SMP PREEMPT Tue Feb 1 21:37:11 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

**Description**

Sometimes when a thread processing a request is interrupted the "interrupted" flag is carried over to the next request handled by the same thread.

**How to reproduce?**

1. Interrupt a thread while it is processing `Request 1`
2. Send `Request 2`. When `Request 2` is processed by the same thread check if the current thread is interrupted. 

Unfortunately the issue doesn't reproduce all the time.

sbordet:
Now clearing the interrupt flag.

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/thread/ReservedThreadExecutor.java
@@ -410,6 +410,11 @@ public class ReservedThreadExecutor extends AbstractLifeCycle implements TryExec
                     {
                         LOG.warn("Unable to run task", e);
                     }
+                    finally
+                    {
+                        // Clear any interrupted status.
+                        Thread.interrupted();
+                    }
                 }
             }
             finally


sbordet:
Note that it is very difficult to write a test case, because a new task submitted via `tryExecute()` is only accepted if there is a reserved thread waiting.

The reserved thread loops around after running the interrupting task, and offers itself to the `SynchronousQueue`, and exactly at that moment the queue should find a new task, because if it does not find it the fact that the thread is interrupted causes an `InterruptedException` to be thrown (which is then caught, which resets the interrupted flag).
If the queue finds a new task, it returns it even if the thread is interrupted.

So the race window to hit to write a test case is very tiny: the new task must be offered exactly when a reserved thread is offered to the queue (to be accepted), but before the reserved thread executes the path that would cause to wait.

joakime:
Would there be any value in clearing the interrupt status before the thread is used from the pool? (as an alternate way to accomplish this)

sbordet:
@joakime no, we use interruption also to stop, so the flag must be cleared in a finally block.
Furthermore, if the thread is interrupted, we don't clear the flag, and then there are no jobs to run and the thread will wait, then it will rethrow which will clear the flag and loop around to see if there are more jobs, but it's an unnecessary loop.

joakime (approved):

gregw (approved):

sbordet:
Merged to `jetty-12.0.x`, commit 4532161b13b38bdc94623b7f9db245b2185b6bc1.

Issue #7531, opened 2022-02-03
Fix ClassLoaderServlet test


janbartel:
`>= jetty-9`

Test produces an exception on jetty9, but is largely commented out on jetty-10,  with a note to uncomment after `jetty-9.4.19`.

joakime:
This was changed as part of PR #4332 originally.

We should either not use `jetty-util` as the test dependency for this ClassLoaderTest, or isolate this change to the dependency tree within tests only, and not let it spill out into the rest of the jetty reactor breaking the release process.

janbartel:
Had a go at reinstating this in PRs #7565 and #7564 but after a great deal of discussion it seems like the fix is too difficult/complicated to achieve in the short term, so this is going onto the backburner.

PR #7564, opened 2022-02-11
Issue #7531 Reinstate ClassLoaderServlet demo class

janbartel:
Re-instated the ClassLoaderServlet class in the demo-spec-webapp. The stated purpose of the test is to show that the `jetty-util` jar can be included in a webapp's `WEB-INF/lib` at a _different_ version level to that of the server and have the webapp use that.

### DIFF
--- a/demos/demo-spec/demo-spec-webapp/pom.xml
+++ b/demos/demo-spec/demo-spec-webapp/pom.xml
@@ -65,9 +65,9 @@
           </supportedProjectTypes>
           <instructions>
             <Bundle-Description>Test Webapp for Servlet 4.0 Features</Bundle-Description>
-            <!-- TODO Add 'org.eclipse.jetty.util;version="[9.4.19,9.4.20)",' below, once 9.4.19 is released with a fix for #3726 -->
+            <!-- Specifically import an old jar to test webapp isolated classloading -->
             <Import-Package>
-              javax.transaction*;version="[1.1,2.0)", javax.servlet*;version="[2.6,4.1)", org.eclipse.jetty*;version="[$(version;===;${parsedVersion.osgiVersion}),$(version;==+;${parsedVersion.osgiVersion}))", org.eclipse.jetty.webapp;version="[$(version;===;${parsedVersion.osgiVersion}),$(version;==+;${parsedVersion.osgiVersion}))";resolution:="optional", org.eclipse.jetty.plus.jndi;version="[$(version;===;${parsedVersion.osgiVersion}),$(version;==+;${parsedVersion.osgiVersion}))";resolution:="optional", com.acme;version="${parsedVersion.majorVersion}.${parsedVersion.minorVersion}.${parsedVersion.incrementalVersion}", *
+                javax.transaction*;version="[1.1,2.0)", javax.servlet*;version="[2.6,4.1)", org.eclipse.jetty.webapp;version="[$(version;===;${parsedVersion.osgiVersion}),$(version;==+;${parsedVersion.osgiVersion}))";resolution:="optional", org.eclipse.jetty.plus.jndi;version="[$(version;===;${parsedVersion.osgiVersion}),$(version;==+;${parsedVersion.osgiVersion}))";resolution:="optional", com.acme;version="${parsedVersion.majorVersion}.${parsedVersion.minorVersion}.${parsedVersion.incrementalVersion}", org.eclipse.jetty.monitor.thread;version="(9.3.29, 9.3.30]", *
             </Import-Package>
             <_nouses />
             <Export-Package>com.acme.test;version="${parsedVersion.majorVersion}.${parsedVersion.minorVersion}.${parsedVersion.incrementalVersion}";-noimport:=true</Export-Package>
--- a/demos/demo-spec/demo-spec-webapp/pom.xml
+++ b/demos/demo-spec/demo-spec-webapp/pom.xml
@@ -177,6 +177,12 @@
               <artifactId>demo-mock-resources</artifactId>
               <version>${project.version}</version>
             </dependency>
+            <!-- deliberately newer version to test with -->
+            <dependency>
+              <groupId>org.eclipse.jetty</groupId>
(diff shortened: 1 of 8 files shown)


joakime (changes requested):
Don't use jetty-util for this test, use something that Jetty itself doesn't rely on.
This change has broken the release process.

janbartel:
@joakime see my comment on #7565.

Also, the whole point of this test is to check that a jar that is used by the _server_ can be used _at a different version_ by the _webapp_.

janbartel:
@joakime I have reworked the test. It was very difficult to find a jar that had:

* multiple old versions
* not currently part of the jetty build
* contained `Implementation-Version` in the manifest

I settled on the old `jetty-monitor` jar, which ticks all of the above boxes, _plus_ as it is in the `org.eclipse.jetty` package it is already treated as server classes.

joakime:
@janbartel definitely a step in the right direction.

But can we choose something that has no jetty dependencies of their own that clash with the reactor?

How about `jetty-monitor` from the Jetty 7.x days?

https://repo1.maven.org/maven2/org/mortbay/jetty/jetty-monitor/

At least that doesn't use `org.eclipse.jetty:jetty-util` as a dependency.

Or what about a unique test class/jar?

https://github.com/joakime/eclipse-jetty-classloader-testing

joakime:
> Or what about a unique test class/jar?
> 
> https://github.com/joakime/eclipse-jetty-classloader-testing

The oss.sonatype.org staging repository has a 1.0 and 2.0 versions of the above JAR.

https://oss.sonatype.org/content/repositories/jetty-1749/

Using classname `org.eclipse.jetty.tests.ExampleClazz` (which should satisfy the `org.eclipse.jetty` classloader rules you pointed out)
The META-INF/MANIFEST.MF has a `Implementation-Version` for both versions of the jar, with either `"1.0"` or `"2.0"` returned depending on the version.
These jars have poms that depend on nothing in the Jetty reactor.

janbartel:
@joakime the jetty-7 `jetty-monitor` jars have dependencies on several jetty-jars, so there would be just as many `<exclusion>` as for the `jetty-monitor` jar I'm using.

As for the jar in the staging repo, I don't see how that helps us, because we need to be able to reference the jar in the `demo-spec.mod` file, which means it must be in maven central, and we don't upload any of the test jars.

What's the problem with the `jetty-monitor` jar I've used in this PR?

joakime:
> @joakime the jetty-7 `jetty-monitor` jars have dependencies on several jetty-jars, so there would be just as many `<exclusion>` as for the `jetty-monitor` jar I'm using.

This still mucks with the same coordinates space as what's present in the reactor.

> As for the jar in the staging repo, I don't see how that helps us, because we need to be able to reference the jar in the `demo-spec.mod` file, which means it must be in maven central, and we don't upload any of the test jars.

We most certainly upload test jars, we have many such examples present on maven central.
We just don't upload test jars with every jetty release.

> What's the problem with the `jetty-monitor` jar I've used in this PR?

It is still overlapping with the same coordinates spaces as what's on the reactor.

Can we use the old mortbay jars instead?

* https://repo1.maven.org/maven2/org/mortbay/jetty/jetty-util/6.1.9/jetty-util-6.1.9.jar
* https://repo1.maven.org/maven2/org/mortbay/jetty/jetty-util/6.0.1/jetty-util-6.0.1.jar

These have no dependencies on anything `org.eclipse.jetty` and also have the `implementation-version` you mentioned is needed.

janbartel:
@joakime I don't understand the problem "with the coordinate space in the reactor". The build works, tests pass, and jetty-home looks fine.
We don't want to use something in the `org.mortbay` package as they aren't protected server classes.

joakime:
> @joakime I don't understand the problem "with the coordinate space in the reactor". The build works, tests pass, and jetty-home looks fine.

The release will fail, just like when using jetty-util.

That's because this change brings in a different version of `org.eclipse.jetty:jetty-util` via dependencies on the `jetty-monitor` you have selected.
The use of `<exclusion>` only moves the goal posts, the dependency is still present and just selected away for the few test cases that you run.

> We don't want to use something in the `org.mortbay` package as they aren't protected server classes.

Its easy to setup more server classes, especially so as you have a module in this PR that can do that easily.  This is a non-issue in my opinion.

Also, the proposed test jars at https://github.com/joakime/eclipse-jetty-classloader-testing don't have this problem with the default server classes OR the unintentional dependency version changes that the original `jetty-util` or this new `jetty-monitor` brings in.

If you want to use `jetty-monitor` and test this classloader issue, then this should not be in a `/demo/` webapp.
Make it isolated in `/tests/webappclassloader-tests/`.
Don't use these new maven modules in `/tests/test-integration/` though, as that just messes everything up again due to transitive dependencies.

janbartel:
@joakime chatting to @gregw it seems the original intent of this servlet was specifically to test older versions of `jetty-util` were useable from within the webapp. So, as that does not seem possible then it may be better to just remove the servlet. We have other tests for server class hiding.

joakime:
If that's all that's being tested, then just create a class in the webapp itself with the same name as something in jetty-util, but does something different.

Instead of testing the `class.getPackage().getImplementationVersion();` to know which version you got, just do a `toString()` and verify if you got the `WEB-INF/classes` version or the real `jetty-util` version.

Something like this in the webapp ...

``` java
package org.eclipse.jetty.util.security;

public class Password {
  public String toString() {
    return "this is the webapp version";
  }
}
```

This is simpler, and has no dependencies that change any existing reactor project declaration.

janbartel:
@joakime sounds like a reasonable idea, but will it cause JPMS to wig out (having split packages)? It would also cause OSGi some problems (also doesn't like split packages), but we may be able to prevent that class being exposed via OSGi.

gregw:
or just delete the test..... this is eating a lot of time and we have lots of things to do.
Let's think of a new test for jetty-12!

janbartel:
Dropping PR, it seems its just too hard to solve this right now.

PR #4332, opened 2019-11-19
Issue #4328 - Including jetty-util-9.3.0.RC0 as WEB-INF/lib

joakime:
+ not including it as a formal `<dependency>`
+ jetty-util-9.3.0.RC0 is pulled in via plugin dependency
  so that it only lives for the duration of a single plugin
  execution and doesn't contaminate the dependency convergence
  on the rest of the project.
+ verified that META-INF/MANIFEST.MF is consistent to 9.4.22 and 9.4.21
  configuration of OSGI manifest, despite the jetty-util artifact
  not being a formal maven dependency.

Signed-off-by: Joakim Erdfelt <joakim.erdfelt@gmail.com>

Closes #4328

PR #7565, opened 2022-02-11
Issue #7531 Reinstate ClassLoaderServlet test class

janbartel:
Re-instates the ClassLoaderSerlvet test class from `test-spec-webapp`, which had been throwing exceptions. The intention is for the ClassLoaderServlet class to be able to load a class from `jetty-util` at a _different_ version level than the server.

### DIFF
--- a/tests/test-webapps/test-servlet-spec/test-spec-webapp/pom.xml
+++ b/tests/test-webapps/test-servlet-spec/test-spec-webapp/pom.xml
@@ -210,6 +210,11 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-util</artifactId>
+      <version>9.3.0.RC0</version>
+    </dependency>
+    <dependency>
       <groupId>org.eclipse.jetty.tests</groupId>
       <artifactId>test-web-fragment</artifactId>
       <version>${project.version}</version>


joakime (changes requested):
This simple change is not enough to isolate the issues with this.

The entire tree, and everything that uses it must be isolated as well.
It cannot spill out into demos, or other modules, or jetty-home.

My last proposal was that we create a /tests/test-classloader/ tree and keep this isolated within there, and add enforcer rules to reject the use of those test modules anywhere else in the project tree.

joakime:
This change has effectively broken the release, as now the jetty-util version is indeterminate from here, to demos, into jetty-home, and jetty-runner.

Another choice is to NOT use jetty-util as this (one of our own modules in this multi-module setup), but something else, that doesn't exist in our own tree.

janbartel:
@joakime can you please explain that? I can't see how a dependency from a single `war` module can leak into any other module? I've just built jetty-9 and 10 and I don't see the older `jetty-util` jar in `jetty-home`?

joakime:
That war is referenced in demos, and test-distribution, and jetty-home, it's rather interconnected.
This change breaks the release 100% of the time.
It also breaks IDE usage as you now have 2 versions of the jetty-util present in your IDE. (which breaks compile/testing in the IDE as well).

If you insist on this being jetty-util, and ONLY jetty-util, and only in this war, then the changes in #4332 is the proper way to test this (which specifically does not use a normal reactor dependency for that old jetty-util).
Frankly, this test should NOT exist in the demos, it should be isolated in its own war, a war that is not referenced by the rest of the reactor (this means no test-distribution or jetty-osgi references back to this war).

This is better accomplished by having a non-jetty artifact (that we don't use anywhere else in jetty) on the server classpath and war classpath and verifying that we handle this situation properly (this fixes the heart of the issue, using `jetty-util` this way, and still allows for for `jetty:run` and test-distribution, and even jetty-osgi usage.

joakime (changes requested):
Just drop PR, not going to happen this simply in 9.4.x

joakime:
Moving this from project 9.4.46 to 9.4.47

janbartel:
Dropping PR, it seems its too hard to solve this right now.

PR #7566, opened 2022-02-11
Work in progress for alternate AcceptorProcessorExchange

gregw:
This is an attempt at introducing the `Acceptor`, `Processor & `Exchange` concepts to the handling.  It is essentially @sbordet method from jetty-12-ng-sbordet, but with an Acceptor separated out that takes the request, so request wrappers are not lost (see discussion in #7424):

 + `Handler`s are offerred an immutable `Request` along with an `Acceptor`.  There is no content API on request.
 + A `Handler` can accept the offered request by calling `Acceptor.accept(Request, Processor)` passing in the request that may have been wrapped by this or a previous `Handler`. 
 + Once accepted, the `Processor.process(Exchange)` method will subsequently be called to actually process the request and generate the response. The `process` call may happen within the scope of `accept` or it may be deferred waiting for content, or QoS issues etc.
 + The `Exchange` passed to `process` contains the `Request` and `Response` instances, it also is-a `Content.Provider` (to get the request body) and a `Callback` to signal success or failure of the exchange.

It is mostly working (8 tests failing) and those should be able to be fixed easily enough......

However, I don't like it.  I almost REALLY don't like it.  I'm close to hating it... but there is enough goodness to make me keep trying a bit more.

It is very technically correct and the concepts of `Acceptor`, `Processor` and `Exchange` feel real and not contrived.     But geeze it is hard to work with and makes many simple cases a lot more complex.   Several times I found myself wrapping an acceptor to wrap the processor so that the exchange could be wrapped to return a wrapped Content.Producer!    That's 4 wrappers instead of something that could have been done with 1.

I think it is going to produce a lot more garbage, have a lot deeper stacks, be a whole bunch hard to debug and the basic handler code will be harder to write correctly.

BUT  I really do like how well it allows deferred call to a Processor (waiting for content, or perhaps a QoS filter etc.)   

BUT I'm wondering if we are messing up the 99% case to make the 1% case work well?

I'll make a few more inline comments to illustrate some of my issues.... and then I'll keep working on it at least until it is green build and we (@lorban) can benchmark it.  Perhaps if it doesn't suck noticeably badly in benchmarks, then I might like it more.

### DIFF
--- a/jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextHandler.java
+++ b/jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextHandler.java
@@ -446,7 +446,7 @@ public class ContextHandler extends Handler.Wrapper implements Attributes, Grace
     }
 
     @Override
-    public void handle(Request request) throws Exception
+    public void offer(Request request, Acceptor acceptor) throws Exception
     {
         if (getHandler() == null)
             return;
@@ -460,16 +460,17 @@ public class ContextHandler extends Handler.Wrapper implements Attributes, Grace
 
         if (pathInContext.isEmpty() && !getAllowNullPathInfo())
         {
-            String location = _contextPath + "/";
-            if (request.getHttpURI().getParam() != null)
-                location += ";" + request.getHttpURI().getParam();
-            if (request.getHttpURI().getQuery() != null)
-                location += ";" + request.getHttpURI().getQuery();
+            String location = _contextPath + "/" +
+                ((request.getHttpURI().getParam() != null) ? (";" + request.getHttpURI().getParam()) : "") +
+                ((request.getHttpURI().getQuery() != null) ? (";" + request.getHttpURI().getQuery()) : "");
 
-            Response response = request.accept();
-            response.setStatus(HttpStatus.MOVED_PERMANENTLY_301);
-            response.getHeaders().add(new HttpField(HttpHeader.LOCATION, location));
-            response.getCallback().succeeded();
+            acceptor.accept(request, e ->
+            {
+                Response response = e.getResponse();
+                response.setStatus(HttpStatus.MOVED_PERMANENTLY_301);
+                response.getHeaders().add(new HttpField(HttpHeader.LOCATION, location));
+                e.succeeded();
+            });
             return;
         }
 
@@ -479,7 +480,7 @@ public class ContextHandler extends Handler.Wrapper implements Attributes, Grace
         if (scoped == null)
             return; // TODO 404? 500? Error dispatch ???
 
-        _context.call(scoped);
+        _context.call(() -> super.offer(scoped, acceptor));
     }
 
     /**
--- a/jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextHandler.java
+++ b/jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextHandler.java
@@ -480,7 +480,22 @@ public class ContextHandler extends Handler.Wrapper implements Attributes, Grace
         if (scoped == null)
             return; // TODO 404? 500? Error dispatch ???
 
-        _context.call(() -> super.offer(scoped, acceptor));
+        _context.call(() -> super.offer(scoped, (r, processor) -> acceptor.accept(r, exchange -> _context.call(() -> processor.process(
+            new Exchange.Wrapper(exchange)
+            {
+                @Override
+                public void demandContent(Runnable onContentAvailable)
+                {
+                    super.demandContent(() -> _context.run(onContentAvailable));
+                }
+
+                @Override
+                public Response getResponse()
+                {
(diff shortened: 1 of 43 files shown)


jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/gzip/GzipHandler.java
```diff
@@ -72,32 +72,38 @@ public long getContentLength()
...
+                acceptor.accept(r, exchange ->
+                    processor.process(new Exchange.Wrapper(exchange)
                     {
-                        return new Response.Wrapper(request, request.accept())
+                        @Override
+                        public Response getResponse()
                         {
-                            @Override
-                            public void write(boolean last, Callback callback, ByteBuffer... content)
+                            return new Response.Wrapper(exchange.getRequest(), super.getResponse())
```
gregw, line 83: Wrapping the acceptor to wrap the processor to wrap the Exchange to wrap the response to intercept write calls!!!!

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/HandleOnContentHandler.java
```diff
@@ -13,78 +13,29 @@
...
-        {
-            _request = request;
-            _response = response;
-        }
-
-        @Override
-        public void run()
-        {
-            try
+            super.offer(request, (req, processor) -> acceptor.accept(req, exchange -> exchange.demandContent(() ->
```
gregw, line 29: Three `->` lambdas on a single line!!! Brain exploding!!!

jetty-core/jetty-core-server/src/test/java/org/eclipse/jetty/core/server/handler/HelloHandler.java
```diff
@@ -52,12 +52,15 @@ public String getMessage()
...
-        response.setContentLength(_byteBuffer.remaining());
-        response.write(true, response.getCallback(), _byteBuffer.slice());
+        acceptor.accept(originalRequest, exchange ->
+        {
+            Response response = exchange.getResponse();
+            response.setStatus(200);
+            response.setContentType(MimeTypes.Type.TEXT_PLAIN_UTF_8.asString());
+            response.setContentLength(_byteBuffer.remaining());
+            response.write(true, exchange, _byteBuffer.slice());
+        });
```
gregw, line 64: The simple case is not tooooo ugly.  But there is definitely a danger of a user referencing `originalRequest` in the lambda rather than `exchange.getRequest()` and thus accidentally making it a capturing lambda.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextHandler.java
```diff
@@ -479,7 +480,22 @@ public void handle(Request request) throws Exception
...
+                    super.demandContent(() -> _context.run(onContentAvailable));
+                }
+
+                @Override
+                public Response getResponse()
+                {
+                    // TODO cache this
+                    return new ContextResponse(scoped, super.getResponse());
+                }
+            })))));
```
gregw, line 498: Is this code maintainable?   Quadruple lambda just to intercept demand and to wrap the exchange just to wrap the response (so it can wrap the write callback).    That's 4 lambdas and 3 wrappers!!!!

janbartel: This code is both incomprehensible and unmaintainable.

gregw: ah yes... but it works and is oh so clever :)

gregw: Ah but it is not correct.... somewhere in there I need to catch any Throwables to ensure error pages are also generated by the context error handler.... actually I probably need to catch twice... or three times?

gregw:
See #7597 for an alternate way to support delayed handling with the existing API

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextHandler.java
sbordet: To be fair, the code is exploded in place. A custom `Exchange` subclass and the need to wrap into `Callable`s makes this hard to read but no more than the current wrapping of `Runnable`s that are not exploded yet equally difficult to follow.

sbordet (commented):
@gregw I agree that this model probably safer from the API point of view, but less palatable due to the excessive wrapping.

Unfortunately I don't see a way out.
We need to go back to APIs that throw `IllegalStateException` when called at the wrong time, but require less wrapping.

gregw:
we all agree this approach is horrid.   Alternates are in progress.

PR #7574, opened 2022-02-13
Turn off peer certificate verification for quic-server by default

sunng87:
Current version of quic-server is configured to verify peer certificate (aka mutual tls) by default. I think this is accidentally set to `true` and should be changed to `false` in order to get examples working.

Signed-off-by: Ning Sun <sunng@protonmail.com>

### DIFF
--- a/jetty-quic/quic-server/src/main/java/org/eclipse/jetty/quic/server/QuicServerConnector.java
+++ b/jetty-quic/quic-server/src/main/java/org/eclipse/jetty/quic/server/QuicServerConnector.java
@@ -89,7 +89,7 @@ public class QuicServerConnector extends AbstractNetworkConnector
         // One bidirectional stream to simulate the TCP stream, and no unidirectional streams.
         quicConfiguration.setMaxBidirectionalRemoteStreams(1);
         quicConfiguration.setMaxUnidirectionalRemoteStreams(0);
-        quicConfiguration.setVerifyPeerCertificates(true);
+        quicConfiguration.setVerifyPeerCertificates(false);
     }
 
     public QuicConfiguration getQuicConfiguration()


sbordet:
@sunng87 the default configuration should be secure, so I think it should be left to true.

What examples are not working for you?

sunng87:
@sbordet by turning on this option, it requires any client to have a client certificate in secure connection handshake. This mechanism is typically used in [mutual tls authentication](https://www.cloudflare.com/learning/access-management/what-is-mutual-tls/). In most cases, like public https service, clients do not have such certificate and we do not auth client like this. So with current defaults, if we call the server with `curl --http3 -v [url]`, it will be blocked by a TLS handshake error.

joakime:
@sunng87 can you show how you initialize/configure your server?

sunng87:
@joakime I'm using the code example from https://www.eclipse.org/jetty/documentation/jetty-10/programming-guide/index.html#pg-server-http-connector-protocol-http3

sbordet:
@sunng87 this is surprising. The server typically configures `[want|need]ClientAuthentication` in order to request the client certificates. 
Whether to verify the certificates is an orthogonal issue (if the certificates are not sent, then the verification is a no-op).
I really hope that Quiche is not requiring this.

sbordet:
Yuck. Quiche requires this 😒 
https://docs.rs/quiche/0.12.0/quiche/struct.Config.html#method.verify_peer

sbordet (approved):

sbordet:
Merged to `jetty-12.0.x` on 12 Feb 2022, commit 65a47a369d3f6e56ecb9d55fb052cfab7c0540a9.

PR #7576, opened 2022-02-14
Issue #7567 - don't compare params when checking MIME type for GzipHandler

### LINKED ISSUE #7567: Gzip compression not working for multipart/form-data when added to the allowed list using addIncludedMimeTypes.
**Jetty version(s)**
9.4.22.v20191022

openjdk 16.0.1 2021-04-20

**OS type/version**
Linux

**Description**
When multipart/form-data is added to the allowed list using addIncludedMimeTypes, compression is not being applied because of the **boundary parameter**. Getting the following error.
`
2022-02-10 21:57:55,024 DEBUG [qtp112619572-112] org.eclipse.jetty.server.handler.gzip.GzipHttpOutputInterceptor - org.eclipse.jetty.server.handler.gzip.GzipHttpOutputInterceptor@3e373008 exclude by mimeType multipart/form-data;boundary=Boundary_2_393478717_1644530275023
`

Included it like this:
```
<Call name="addIncludedMimeTypes">
    <Arg>
    <Array type="String">
       <Item>application/json</Item>
        <Item>multipart/form-data</Item>
    </Array>
    </Arg>
</Call>
```

The comment for addIncludedMimeTypes says ["@param types The mime types to include (without charset or other parameters)" ](https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipHandler.java#L398) but only charset is being removed but not the boundary. [https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipHandler.java#L697](url)

Usually, multipart/form-data is followed by other parameters and other parameters are not removed while checking the condition, hence compression is likely to always fail for multipart/form-data when added to the allowed list.

Is this a bug? Is there any work around?

lachlan-roberts:
## Issue #7567

- add testing for GzipHandler with multipart/form-data content types
- strip params before comparing MIME type with the included/excluded mime types in GzipHandler

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipHandler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipHandler.java
@@ -694,7 +694,7 @@ public class GzipHandler extends HandlerWrapper implements GzipFactory
         String mimeType = context == null ? MimeTypes.getDefaultMimeByExtension(path) : context.getMimeType(path);
         if (mimeType != null)
         {
-            mimeType = MimeTypes.getContentTypeWithoutCharset(mimeType);
+            mimeType = HttpField.valueParameters(mimeType, null);
             if (!isMimeTypeGzipable(mimeType))
             {
                 LOG.debug("{} excluded by path suffix mime type {}", this, request);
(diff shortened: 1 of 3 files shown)


gregw (approved):

joakime (approved):
This change looks odd (going from `MimeTypes.getContentTypeWithoutCharset(ct)` to `HttpField.valueParameters(ct, null)` seems like a step backwards.

But it works, so ship it!

PR #7577, opened 2022-02-14
Jetty 12.0.x Align artifactId naming

olamy:


sbordet:
@olamy I'm not sure about this... changing the Maven artifactIds will really break everybody.
It is true that if we break this is the right time, but should we really break?

joakime:
I think now is the time to fix this.

This change also changes the filename when we see it in classloaders and such.

eg: `http2-client-11.0.8.jar` now becomes `jetty-http2-client-12.0.0.jar` (which is more clear)

olamy:
same comment as @joakime time to do. as a lot will change with jetty 12 so this groupId change will look very minor change.

janbartel:
I'm ok with doing the artifact renaming to always be the same as the module, and include `jetty` prefix.

olamy:
@simone ? still not?

gregw (changes requested):
I think we need to hold of on changing artifact names etc. until we decide what is the final structure.    I am also concerned that we should not needlessly pollute the space with new names, even if they are "more correct".

Issue #7573, opened 2022-02-13
WebSockets - "Unsupported PathParam Type: java.lang.Integer"


andrus:
**Jetty version(s)**

10.0.7

**Java version/vendor** `(use: java -version)`

11.0.12 (Temurin)

**OS type/version**

MacOS

**Description**
WebSocket endpoint like this:

```java
@ServerEndpoint(value = "/my/{id}")
public class Endpoint {

    @OnOpen
    public void onOpen(@PathParam("id") Integer id, Session session) {
        
    }
}
```
on Jetty startup results in an [exception here](https://github.com/eclipse/jetty.project/blob/cb127793e5d8b5c5730b964392a9a905ba49191d/jetty-websocket/websocket-javax-server/src/main/java/org/eclipse/jetty/websocket/javax/server/internal/PathParamIdentifier.java#L63) - "Unsupported PathParam Type: java.lang.Integer". Changing `Integer` to `int` fixes the issue. The expected behavior would be for `PathParamIdentifier` to compare the type against both `X.TYPE` and `X.class`, otherwise boxed types will not work.

lachlan-roberts:
Opened PR #7579 to fix this.

lachlan-roberts:
fixed with PR #7579, will be available with the release of `10.0.9`/`11.0.9`.

PR #7579, opened 2022-02-14
Issue #7573 - fix support in websocket for class path param types

lachlan-roberts:
## Issue #7573
Add testing for path params with all primitive types and their class versions.
Fix `PathParamIdentifier` and `JavaxWebSocketFrameHandlerFactory` to make the path params work.


joakime (approved):
LGTM, glad for the new test cases.

PR #7580, opened 2022-02-14
simplify pom by using simple property rather than xml fragment

olamy:

### DIFF
--- a/jetty-http3/http3-tests/pom.xml
+++ b/jetty-http3/http3-tests/pom.xml
@@ -10,6 +10,11 @@
   <artifactId>http3-tests</artifactId>
   <name>Jetty :: HTTP3 :: Tests</name>
 
+  <properties>
+    <maven.deploy.skip>true</maven.deploy.skip>
+    <maven.javadoc.skip>true</maven.javadoc.skip>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>org.eclipse.jetty.http3</groupId>
--- a/jetty-osgi/test-jetty-osgi-webapp-resources/pom.xml
+++ b/jetty-osgi/test-jetty-osgi-webapp-resources/pom.xml
@@ -11,7 +11,7 @@
   <packaging>war</packaging>
   <properties>
     <bundle-symbolic-name>${project.groupId}.webapp.resources</bundle-symbolic-name>
-    <maven.deploy.skip>false</maven.deploy.skip>
+    <maven.deploy.skip>true</maven.deploy.skip>
     <maven.javadoc.skip>true</maven.javadoc.skip>
   </properties>
   <build>
--- a/jetty-osgi/test-jetty-osgi-webapp-resources/pom.xml
+++ b/jetty-osgi/test-jetty-osgi-webapp-resources/pom.xml
@@ -11,41 +11,31 @@
   <packaging>war</packaging>
   <properties>
     <bundle-symbolic-name>${project.groupId}.webapp.resources</bundle-symbolic-name>
+    <maven.deploy.skip>false</maven.deploy.skip>
+    <maven.javadoc.skip>true</maven.javadoc.skip>
   </properties>
   <build>
-    <pluginManagement>
-      <plugins>
-        <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-javadoc-plugin</artifactId>
-          <configuration>
-            <!-- No point building javadoc on testing projects -->
-            <skip>true</skip>
-          </configuration>
-        </plugin>
-      </plugins>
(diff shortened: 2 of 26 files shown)


jetty-http3/http3-tests/pom.xml
```diff
@@ -10,6 +10,11 @@
   <artifactId>http3-tests</artifactId>
   <name>Jetty :: HTTP3 :: Tests</name>
 
+  <properties>
```
olamy, line 13: @sbordet I assume we should not deploy those tests as part of the release and was a mistake? or I am wrong?

jetty-osgi/test-jetty-osgi-webapp-resources/pom.xml
```diff
@@ -70,14 +60,6 @@
           </archive>
         </configuration>
       </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-deploy-plugin</artifactId>
-        <configuration>
-          <!-- must not deploy -->
```
olamy, line 77: I have to say I was confused here and not sure to believe the comment or the code. :)

jetty-osgi/test-jetty-osgi-webapp-resources/pom.xml
```diff
@@ -70,14 +60,6 @@
...
-          <skip>false</skip>
-        </configuration>
```
olamy, line 79: I have to say I was confused here and not sure what to believe the comment or the code. :)

jetty-http3/http3-tests/pom.xml
sbordet: This module will only produce a test jar (no sources), so it won't be deployed anyway, right?
Ah, I see it deployed an empty jar 😒 
So yes let's skip deploy for this module.

jetty-websocket/websocket-jetty-tests/pom.xml
```diff
@@ -12,6 +12,7 @@
 
   <properties>
     <bundle-symbolic-name>${project.groupId}.tests</bundle-symbolic-name>
+    <maven.javadoc.skip>true</maven.javadoc.skip>
```
sbordet, line 16: Should not the deploy be skipped too?

sbordet (changes requested):

jetty-websocket/websocket-jetty-tests/pom.xml
olamy: correct I forgot that

olamy: fixed https://github.com/eclipse/jetty.project/pull/7580/commits/10a9706d6ddd9c24767759742fbf2249d8d2b028

jetty-http3/http3-tests/pom.xml
olamy: after digging a bit the jar is not empty as it included some content we put in `target/classes` as part of the build such: generated MANIFEST.MF, LICENSE and NOTICE.txt

sbordet (dismissed):

jetty-osgi/test-jetty-osgi-webapp-resources/pom.xml
janbartel: I think it should be the same as the other osgi test projects ;)

tests/jetty-jmh/pom.xml
```diff
@@ -57,18 +57,6 @@
...
       </plugin>
     </plugins>
-    <pluginManagement>
-      <plugins>
-        <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-deploy-plugin</artifactId>
-          <configuration>
-            <!-- No point deploying jmh project -->
-            <skip>true</skip>
```
janbartel, line 67: Needs the property defined? Or is that covered by the toplevel tests/pom.xml properties?

janbartel (changes requested):

tests/jetty-jmh/pom.xml
olamy: correct. properties are inherited.

sbordet (dismissed):

janbartel:
@olamy this one needs merge conflicts fixed

janbartel (changes requested):
Changes themselves look ok, but this branch needs re-merging from jetty-10.0.x.

janbartel:
@olamy nudge, needs merge.

olamy:
@janbartel oh yes I definitely forgot this old PR. thanks for the ping :)

janbartel (approved):

PR #7597, opened 2022-02-15
Added Jetty-12 DelayedHandler, variants and tests

gregw:
One of the good elements of #7566 for #7424 is that it well supports delayed handling of requests.
This PR is an alternate approach that provides a `DelayedHandler` with implementations for `OnContent` and `QualityOfService`.    This shows that the existing API is able to support this semantic, albeit with the limitation that the request must be accepted.


sbordet (commented):
I see how it is working, but I like better a clear separation between accept and process.
What about a QoSHandler in front of 2 ContextHandlers? If neither accepts the request, then the request is accepted but no response provided.
`Server` will see the request accepted (by the QoSHandler), but does not know it was not responded because neither ContextHandler matched, right?

gregw:
@sbordet if neither context handler accepts, then the QoSHandler will send a 404, which is pretty much what the server would do.    Alternately, it could just look for and call the default handler, so you get exactly the same 404 that the server would do.

I agree that the accept + process model is technically more correct.  But I just think that the additional complexity is not worth it to just cover this case.

gregw:
@sbordet I've pushed an update to show how if the nested handlers don't accept the request, then the DelayedHandler can pass it to the DefaultHandler to handle.   This could be made configurable/flexible, so any handler could be called if a nested handler did not accept.

gregw:
@sbordet when you say:
> `Server` will see the request accepted (by the QoSHandler), but does not know it was not responded because neither ContextHandler matched, right?

The important distinction is that `DelayedHandler` does know if the request was responded to because it wraps `accept()` and shades it's own accepted state.    Thus it can always mimic exactly what the server would have done if a handler had not accepted.

gregw:
@sbordet @lorban I think this PR is non controversial, at least against the current head.    Even if we introduce API changes to change the way that delayed processing is done, the `DelayedHandler` and tests here are useful and could be converted to the new API to test it.       Thus I'd like to get this reviewed and merged soon - but not as a replacement for any Acceptor/Processor API change, but rather as a platform on which to test proposals.

sbordet (approved):

PR #7598, opened 2022-02-15
Jetty 12.0.x server context

gregw:
Moved `Context` to be an interface in `Request`, so that all requests always have a `Context` (which may be replaced).
This simplifies things like error handling and allows for all the executor methods to be moved off the request and to just be on the context itself.


gregw:
close in favour of #7649

PR #7601, opened 2022-02-16
Use the InvocationType of Handler

gregw:
+ Added InvocationType to the handlers
 + ensure type flows all the way through to the ManagedSelector
 + ensure handlers combine InvocationType when aggregated
 + cache InvocationType at the server to avoid handler tree traversal on every task.   Handlers that can be dynamically updated check for a change in the cached invocation type and throw IAE if so.  Could be made to update cached value?


gregw:
With this change, we can have handlers that declare as non_blocking and all handling is done in a selector thread with `ProduceConsume`.     Anything that goes to a Servlet or similar will have to be BLOCKING, so will only be PC if a reserved thread is available, otherwise `ProduceExecuteConsume`.    Will be interesting to see if that makes a difference to things like the techempower benchmarks, which are just really simple requests that don't need to be servlets.

jetty-core/jetty-core-server/src/test/java/org/eclipse/jetty/core/server/ServerConnectorHttpServerTest.java
```diff
@@ -24,6 +34,88 @@
...
+
+            // Read the response.
+            String response = readResponseHeader(client);
+
+            assertThat(response, containsString("HTTP/1.1 200 OK"));
+        }
+
+        // TODO need a better way to test this
+        assertThat(_connector.getSelectorManager().getBean(ManagedSelector.class).dump(), not(containsString("[pc=0")));
+        assertThat(_connector.getSelectorManager().getBean(ManagedSelector.class).dump(), containsString("pic=0,pec=0,epc=0]"));
```
gregw: @sbordet Can you think of a better way to test this?

jetty-core/jetty-core-server/src/test/java/org/eclipse/jetty/core/server/ServerConnectorHttpServerTest.java
```diff
@@ -24,6 +34,88 @@
...
+
+            // Read the response.
+            String response = readResponseHeader(client);
+
+            assertThat(response, containsString("HTTP/1.1 200 OK"));
+        }
+
+        // TODO need a better way to test this
+        assertThat(_connector.getSelectorManager().getBean(ManagedSelector.class).dump(), containsString("[pc=0"));
+        assertThat(_connector.getSelectorManager().getBean(ManagedSelector.class).dump(), containsString(",pic=0,"));
```
gregw: @sbordet Also here.

jetty-core/jetty-core-server/src/test/java/org/eclipse/jetty/core/server/ServerConnectorHttpServerTest.java
sbordet: Ensure there is only 1 selector when configuring the connector, then:

```java
ManagedSelector selector = _connector.getSelectorManager().getBean(ManagedSelector.class);
AdaptiveExecutionStrategy strategy = selector.getBean(AdaptiveExecutionStrategy.class);
assertEquals(0, strategy.getPICTasksExecuted());
```

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Handler.java
```diff
@@ -243,6 +267,11 @@ public Handler getHandler()
 
         public void setHandler(Handler handler)
         {
+            Server server = getServer();
+            if (server != null && server.isStarted() && handler != null &&
+                server.getInvocationType() != Invocable.combine(server.getInvocationType(), handler.getInvocationType()))
+                throw new IllegalArgumentException("Cannot change invocation type of started server");
```
sbordet, line 249: Not sure about this? Deploying a new webapp along with a non-blocking handler would fail, no?
Can we recompute the `HttpChannel.RunHandle` we give to the strategy with the new `InvocationType`?

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/HttpConnection.java
```diff
@@ -129,6 +129,12 @@ public HttpConnection(HttpConfiguration configuration, Connector connector, EndP
             LOG.debug("New HTTP Connection {}", this);
     }
 
+    @Override
+    public InvocationType getInvocationType()
+    {
+        return getServer().getInvocationType();
+    }
+
```
sbordet, line 137: Is this necessary? If so, should it not be less coarse and return the `InvocationType` of the `Handler` that handles the request?

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Server.java
```diff
@@ -161,6 +162,13 @@ public void handle(Request request)
         }
     }
 
+    @Override
+    public InvocationType getInvocationType()
+    {
+        // Return cached type to avoid a full handler tree walk.
+        return _invocationType;
+    }
+
```
sbordet, line 160: Unless we can associate the `InvocationType` to the request after being accepted?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BlockUtil.java
```diff
@@ -31,7 +31,7 @@
  *
  * <h2>Non shared Runnable</h2>
  * <pre>
- *     try(Blocking.Runnable onAction = Blocking.runnable())
+ *     try(BlockUtil.Runnable onAction = BlockUtil.runnable())
```
sbordet: I liked `Blocking` much more than `BlockUtil`.

sbordet (changes requested):
Yes! I was waiting for this since a long time, and would be great if we could do this at a more refined level (ideally, per-request).

Some stuff to clean up and some suggestions/ideas, but it's a much wanted change.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Handler.java
gregw: @sbordet we don't really want to recompute the runhandle for every channel, as if they are not reused (or connection is not long lived) then that is a lot of extra traverses of the full handler tree.
The ISE will only be thrown if the computed type is changed to blocking when it is cached as non-blocking.  So it should be rare.   But we could equally just recompute the type instead of throwing IAE

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/HttpConnection.java
gregw: Server is-a Handler.Wrapper, so it is the root Handler of the handler tree

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Handler.java
joakime: Many folks have dynamic handler trees now, we should still allow for this kind of thing, even if it puts more burden on the developers using it.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BlockUtil.java
gregw: So then what should I call `Handler.Blocking`?   I renamed to avoid a name clash.
Happy to rename it back, but need a better name for the handler.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Handler.java
gregw: @joakime this doesn't prevent dynamic trees.  It only currently prevents dynamic trees that change the dispatch type... but as any app that has a servlet is already going to be BLOCKING, this will not affect them.     It would only affect users that created dynamic handler trees that are async and then added a blocking handler.  We have zero users like that now, because we don't have async handlers.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Server.java
gregw: We really want to avoid traversing the handler tree just to work out the type.    But if we do decide to use a two phase accept/process approach, then it is possible that the accept phase could return the type of the process phase.   But that does us no good if the accept phase blocks, so we'd still need handlers to declare if their accept phase blocks or not.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Handler.java
gregw: @sbordet @joakime I've had a look at making the type dyanmic as well, but I don't think it can be done without making a global lock on all handling. Anytime you ask the handler tree for a type, then an async modification can sneak in and change the tree before you actually call the first handler.  Thus we are always in a race.

Choices are:  
 1. always assume handlers are blocking (as we currently do)
 2. Allow async handlers, but once we have started with a fully async tree, then we can't change it to blocking dynamically.  We can always add a async handler to a blocking tree, but not vice versa

gregw:
@sbordet I think this PR would also be good to get merged before considering the Accept/Processor branch.  How the invocation type is handled in accept vs process is going to be interesting.... is the process another task that should be fed back to the ExecutionStrategy? or are they just a combined task?  in which case how can we know the type before accept?

gregw:
changes merged with #7624

PR #7604, opened 2022-02-16
Enable errorprone and spotbugs for java 17

olamy:
- add compiler flag to support errorprone for jdk17 and restore spotbugs parsing
- spotbugs is fine with jdk17


Jenkinsfile
```diff
@@ -12,7 +12,9 @@ pipeline {
           steps {
             container('jetty-build') {
               timeout( time: 120, unit: 'MINUTES' ) {
-                mavenBuild( "jdk17", "clean install", "maven3")
+                // configuration needed for errorprone but not for normal build
+                sh "mv .mvn/.jvm.config .mvn/jvm.config"
```
joakime: Is there any harm in leaving this `.mvn/jvm.config` configuration enabled for both builds?

olamy: nope no issue with jdk11.
maybe to ease this for every build even local? (i.e just having a file ` .mvn/jvm.config`)
look at the content this doesn;t affect anything else than com.sun.tools.javac modules
wdyt?

joakime: Yes, lets leave it enabled for any JVM. (even JDK 18 and JDK 19)

olamy: fixed

joakime: Going to wait for a green build now ... :-)

joakime (approved):
Build is green.
Lets merge this.

PR #7606, opened 2022-02-16
Issue #5965 - fixing invalid dot output from start.jar

### LINKED ISSUE #5965: Option --write-module-graph produces wrong .dot file
**Jetty version**
10.0.x

**Description**
```
java -jar $JETTY_HOME/start.jar --write-module-graph=/tmp/graph.dot
```

produces a file, but then:

```
$ dot -V
dot - graphviz version 2.43.0 (0)
$ dot -Tpng -o graph.png graph.dot 
Error: graph.dot: syntax error in line 889 near '>'
```

joakime:
Fixes #5965

### DIFF
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/ModuleGraphWriter.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/ModuleGraphWriter.java
@@ -34,6 +34,7 @@ public class ModuleGraphWriter
 {
     private String colorModuleBg;
     private String colorEnabledBg;
+    private String colorEdgeOptional;
     private String colorTransitiveBg;
     private String colorCellBg;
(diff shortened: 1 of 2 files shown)


sbordet (dismissed):
@joakime possible to write a test case, running `dot` as in the issue to see if it can process the file?

`dot` is required to build Jetty (because of the documentation) so it should not be a problem?

joakime:
@sbordet done, added testcase that uses `dot` if it exists

joakime:
> `dot` is required to build Jetty (because of the documentation) so it should not be a problem?

Actually, it's not required, the documentation will still build and produce output, but without diagrams if `dot` is missing.

olamy (approved):

sbordet (approved):

Issue #7615, opened 2022-02-17
HttpServletResponse.encodeURL not working for URLs starting with ../


anca-jalbum:
**Jetty version(s)**
9.4.37 - 9.4.45

**Java version/vendor**
Java 1.8.0

**OS type/version**
Mac

**Description**
When doing HttpServletResponse.encodeURL("../foo/bar.jsp") when client has no session cookie so the URL should be rewritten as "../foo/bar.jsp;jsessionid=[sessionID]" the HttpURI.parse method responds with "Bad URI".

That is not a "Bad URI". Relative URLs with any number of ../ must be allowed.

**How to reproduce?**
Create a foo.jsp file in a foo-folder that just contains:
`<%= response.encodeURL("../foo/bar.jsp") %>`

Access the page http://[server]/foo/foo.jsp with a client that doesn't send cookies, or have no JSESSIONID cookie, like
`curl http://[server]/foo/foo.jsp`

This worked upto and including Jetty 9.4.36, but don't work with Jetty 9.4.37 and newer. So some update in 9.4.37 must have made this break.

I have 3rd party libraries that does this within servlets that I can't change, so it's a showstopper for me to upgrade to the latest version of Jetty.

joakime:
I haven't replicated this yet, but this issue should also be impacting `HttpServletResponse.encodeRedirectURL(String)` as well. (similar code path).

joakime:
I have been able to replicate easily with this test case...

https://github.com/eclipse/jetty.project/commit/d9c6414a36432528cf6070e55b9ad871919e371d

The `HttpURI` is used within the `encodeURL(String)` method only to parse out various values.
But our `HttpURI` doesn't support relative path only Strings.

joakime:
Opened PR #7616 with a fix.

joakime:
Closing, Merged PR #7763

gregw:
Reopening as not yet merged to 10, 11 & 12!

PR #7616, opened 2022-02-17
Issue #7615 HttpServletResponse.encodeURL() should support relative paths

joakime:
Fixing `Response.encodeURL(String)` to work with relative paths like `encodeURL("../foo/bar.txt")`

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
@@ -20,6 +20,8 @@ package org.eclipse.jetty.server;
 
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.nio.channels.IllegalSelectorException;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


joakime:
@gregw bump, review please.

jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -327,10 +328,18 @@ public String encodeURL(String url)
...
             return url;
 
-        HttpURI uri = null;
+        URI uri = null;
         if (sessionManager.isCheckingRemoteSessionIdEncoding() && URIUtil.hasScheme(url))
         {
-            uri = new HttpURI(url);
+            try
+            {
+                uri = new URI(url);
```
gregw, line 336: I'm very much against introducing the jdk URI class here.   
I'm going to have a little bit of a play to see what our alternatives are. Standby....

gregw (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
gregw: Also it is not needed at this point as a URI starting with .. will never have a scheme

gregw:
see #7763 for a simpler alternative

joakime:
Closing in favor of #7763

PR #7623, opened 2022-02-18
Jetty 12.0.x sbordet's take and Handler accept/process

sbordet:


jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Handler.java
```diff
@@ -122,8 +121,20 @@
...
 
+        @Override
+        public void accept(Request request) throws Exception
+        {
+            request.accept(_processor);
+        }
+
+        protected void handle(Request request, Response response, Callback callback) throws Exception
+        {
+        }
```
gregw: It is a bit wierd/confusing that all Abstract handlers have `handle` method and a `_processor` field, even though many of them will not actually use it.  I think these should be taken out of `Abstract` and then have another base class that is used only for those handlers that use this method.

Also should be called `process`

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/QualityOfServiceHandler.java
```diff
@@ -0,0 +1,111 @@
...
+            super(wrapped);
+        }
+
+        @Override
+        public void accept(Processor processor) throws Exception
+        {
+            getWrapped().accept((rq, rs, cb) -> handle(processor, rs, cb));
+        }
+
+        private void handle(Processor processor, Response response, Callback callback)
```
gregw, line 64:
```suggestion
        private void process(Processor processor, Response response, Callback callback)
```

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/QualityOfServiceHandler.java
```diff
@@ -0,0 +1,111 @@
...
+            try
+            {
+                processor.process(request, response, callback);
+            }
+            catch (Exception x)
+            {
+                // TODO: better exception handling?
+                callback.failed(x);
+            }
+        }
```
gregw: inline this method. The last thing this branch needs is another method called `process`  :)

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextRequest.java
```diff
@@ -39,27 +42,36 @@ protected ContextRequest(ContextHandler contextHandler, Request wrapped, String
...
-        return _response;
+        try
+        {
+            processor.process(request, response, callback);
+        }
+        catch (Throwable x)
+        {
+            // TODO: see below about QuietException.
+            if (!response.isCommitted())
+                response.writeError(request, x, callback);
```
gregw: This comment is common to both branches.

Sometimes when we catch and exception we do `callback.failed(x)` other times we do `response.writeError(x,callback)`.  We need to pick one.   I actually think the later is better and it can do the quiet logging and check for committed.  In fact it already checks for committed and falls back to `callback.failed(x)` if it is.

Anyway - that applies equally to both branches.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextRequest.java
```diff
@@ -134,14 +142,11 @@ public String getPath()
...
-                return _contextHandler.getContext().getContextPath();
-            case "o.e.j.s.h.ScopedRequest.pathInContext":
-                return _pathInContext;
-            default:
-                return super.getAttribute(name);
-        }
+            case "o.e.j.s.h.ScopedRequest.contextPath" -> _contextHandler.getContext().getContextPath();
+            case "o.e.j.s.h.ScopedRequest.pathInContext" -> _pathInContext;
+            default -> super.getAttribute(name);
+        };
```
gregw, line 148: such sexy syntax!  But we should refrain from such changes until we pick a branch... then we can convert lots of things.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/DefaultHandler.java
```diff
@@ -84,16 +85,15 @@ public DefaultHandler()
     }
 
     @Override
-    public void handle(Request request) throws Exception
+    protected void handle(Request request, Response response, Callback callback) throws Exception
     {
-        Response response = request.accept();
-        if (response == null || response.isCommitted())
+        if (response.isCommitted())
```
gregw: Can this ever be true?  It can only be committed if accepted and if this handler didn't accept then this wont be called.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/HttpChannel.java
```diff
@@ -637,15 +625,16 @@ public void failed(Throwable x)
...
 
         @Override
-        public Response accept()
+        public void accept(Processor processor) throws Exception
         {
             try (AutoLock ignored = _lock.lock())
             {
                 if (_accepted)
-                    return null;
+                    throw new IllegalStateException("request already accepted " + this);
```
gregw, line 633: In the 12.0.x branch, during error/exception handling, I found it useful to avoid races by calling `accept()` and then only acting if a response was returned.
Throwing here makes that semantic more difficult/costly.  Hmmm let me see if I can find and example....

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Server.java
```diff
@@ -113,53 +113,49 @@ public Server(@Name("threadpool") ThreadPool pool)
...
-                Callback callback = response.getCallback();
-                if (response.isCommitted())
-                    callback.failed(new IllegalStateException("No Handler for committed request"));
-                else
-                    response.writeError(404, null, callback);
-            }
+            request = new ServerRequest(request);
+            accept(request);
+            if (!request.isAccepted())
+                request.accept((rq, rs, cb) -> rs.writeError(rq, 404, cb));
```
gregw, line 137: If we ever become more async in the way we accept requests, then this pattern is a test-then-act race.   The other branch allows an atomic test-and-accept-if-not-accepted semantic.      However, I'm not sure if this will ever be a real problem.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Server.java
```diff
@@ -113,53 +113,49 @@ public Server(@Name("threadpool") ThreadPool pool)
...
+        catch (Throwable x)
         {
-            // Let's be less verbose with BadMessageExceptions & QuietExceptions
-            if (!LOG.isDebugEnabled() && (t instanceof BadMessageException || t instanceof QuietException))
-                LOG.warn("bad message {}", t.getMessage());
+            if (!request.isAccepted())
+                acceptAndFail(request, x);
             else
-                LOG.warn("handle failed {}", this, t);
+                callback.failed(x);
```
gregw: again, this looks like a race, but would only be so if a handler was accepting in a different thread whilst the handling thread threw.  Possible but unlikely

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Server.java
```diff
@@ -684,5 +678,36 @@ public DateField(long seconds, HttpField dateField)
...
+        @Override
+        public void accept(Processor processor) throws Exception
+        {
+            getWrapped().accept((rq, rs, cb) ->
+            {
+                try
+                {
+                    processor.process(this, rs, cb);
+                }
+                catch (Throwable x)
```
gregw: This is a clever "trick" to ensure processors see the correct wrappers.   But it is a little strange that:
 + The ServerRequest is a wrapper applied after customizer wrappers
 + The passed request is ignored and this is passed on.

Maybe should have an assert?

```suggestion
        public void accept(Processor processor) throws Exception
        {
            getWrapped().accept((rq, rs, cb) ->
            {
                assert rq.getWrapper() == this;
                try
                {
                    processor.process(this, rs, cb);
                }
                catch (Throwable x)
```

We are going to have to document things very carefully to ensure that users do this trick in their own handlers.   This API does open up the possiblility of wrappers being lost, different wrappers used for accept vs process and all sorts of other strange edge cases.   I fear some users might find a "use" for such anti-tricks

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ErrorProcessor.java
```diff
@@ -96,36 +95,43 @@ public void handle(Request request) throws IOException
...
 
-            generateAcceptableResponse(request, response, code, message, cause);
+            try
+            {
+                generateAcceptableResponse(request, response, code, message, cause, callback);
+            }
+            catch (Throwable x)
+            {
+                // TODO: cannot write the error response, give up and close the stream.
+                x.printStackTrace();
```
gregw, line 119: why not `callback.fail(x)` ?

gregw: or why not just throw from process?

gregw (commented):
I don't hate it.... which is a start.  But remain concerned about the additional complexity and the potential for miss use of the API with bad wrappings of Processors.

But there are aspects of the API I do like... like the Processor API itself is good.

Ultimately for me it is going to come down to performance and garbage (and that goes for the 12.0.x branch vs jetty-11 as well).

gregw:
Some benchmark results:

12.0.x:
```
Benchmark                                                    Mode  Cnt        Score       Error   Units
HandlerBenchmark.testPost                                   thrpt   10  1047612.732 ± 38818.355   ops/s
HandlerBenchmark.testPost:·gc.alloc.rate                    thrpt   10     2057.230 ±   504.670  MB/sec
HandlerBenchmark.testPost:·gc.alloc.rate.norm               thrpt   10     4336.330 ±     0.072    B/op
HandlerBenchmark.testPost:·gc.churn.G1_Eden_Space           thrpt   10     2100.520 ±   537.299  MB/sec
HandlerBenchmark.testPost:·gc.churn.G1_Eden_Space.norm      thrpt   10     4424.094 ±   232.899    B/op
HandlerBenchmark.testPost:·gc.churn.G1_Survivor_Space       thrpt   10        0.015 ±     0.017  MB/sec
HandlerBenchmark.testPost:·gc.churn.G1_Survivor_Space.norm  thrpt   10        0.031 ±     0.034    B/op
HandlerBenchmark.testPost:·gc.count                         thrpt   10       69.000              counts
HandlerBenchmark.testPost:·gc.time                          thrpt   10       61.000                  ms
```

This branch:
```
Benchmark                                                    Mode  Cnt       Score       Error   Units
HandlerBenchmark.testPost                                   thrpt   10  960666.254 ± 79411.056   ops/s
HandlerBenchmark.testPost:·gc.alloc.rate                    thrpt   10    1914.510 ±   503.539  MB/sec
HandlerBenchmark.testPost:·gc.alloc.rate.norm               thrpt   10    4376.313 ±     0.100    B/op
HandlerBenchmark.testPost:·gc.churn.G1_Eden_Space           thrpt   10    1935.543 ±   515.293  MB/sec
HandlerBenchmark.testPost:·gc.churn.G1_Eden_Space.norm      thrpt   10    4445.168 ±   636.745    B/op
HandlerBenchmark.testPost:·gc.churn.G1_Survivor_Space       thrpt   10       0.012 ±     0.010  MB/sec
HandlerBenchmark.testPost:·gc.churn.G1_Survivor_Space.norm  thrpt   10       0.027 ±     0.019    B/op
HandlerBenchmark.testPost:·gc.count                         thrpt   10      58.000              counts
HandlerBenchmark.testPost:·gc.time                          thrpt   10      50.000                  ms
```

This branch has less through put, but about the same performance on garbage/op (which is a surprise).   Will need to do a similar JMH on jetty-11, but I think the differences in the mock connection will swamp anything real.... so perhaps need to do a real benchmark over real HTTP

lorban:
@gregw How about using the following JMH profiler?

```java
.addProfiler(AsyncProfiler.class, "output=flamegraph")
```

Note that you need a recent async profiler binary installed locally (get it from https://github.com/jvm-profiling-tools/async-profiler/releases) and need to set the following env variable to get it to work: `LD_LIBRARY_PATH=${ASYNC_PROFILER_HOME}/build`

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Handler.java
```diff
@@ -33,36 +33,35 @@
  * <p>
  * Incoming requests to the Server (itself a Handler) are passed to one or more Handlers
  * until the request is handled and a response is produced.  Handlers are asynchronous,
- * so handling may happen during or after a call to {@link #handle(Request)}.
+ * so handling may happen during or after a call to {@link #accept(Request)}.
  *
  * <p>
- * A call to {@link #handle(Request)} may:
+ * A call to {@link #accept(Request)} may:
```
gregw, line 39: This should be renamed to `offer`

gregw (changes requested):

gregw:
I've been looking at if it makes more sense to return a `Processor` rather than have an API on request for accept that can be called at the wrong time or multiple times etc.

Currently this branch has the style:
```java
    class TestHandler extends Wrapper
    {
        @Override
        public void accept(Request request) throws Exception
        {
            Request.Wrapper wrapped = new Request.Wrapper(request)
            {
                @Override
                public void accept(Processor processor) throws Exception
                {
                    super.accept((ignored, response, callback) -> processor.process(this, response, callback));
                }
                
                // Override methods of interest
            };
            super.accept(wrapped);
        }
    }
```

This has one request wrapper and then a capturing lamda (capturing processor) that wraps the `Processor`.   The accept method on the request is there forever (even during process) and we have to guard and throw ISE if it is called wrongly.... which was kind of the thing this branch was meant to stop!

So if instead `accept` (which should be called `offer`) returned the `Processor` we could have:
```java
    class AltTestHandler extends AltHandlerWrapper
    {
        @Override
        public Processor accept(Request request) throws Exception
        {
            Request.Wrapper wrapped = new Request.Wrapper(request)
            {
                // Override methods of interest
            };

            Processor processor = super.accept(wrapped);
            return processor == null ? null : (ignored, response, callback) -> processor.process(wrapped, response, callback);
        }
    }
```

This has one request wrapper and a capturing lambda (capturing processor & wrapped) that wraps the `Processor`.   So it's pretty much the same thing, but now we don't have the issue of and accept method left on the request that can be called at the wrong time.  The only way to accept is to return a `Processor` from `accept`/`offer`.  I think this is cleaner.    I think this might also get rid of the need for the `ServerRequest` wrapper, so one less wrapper!

Only problem now is customizers that want to process the request.  Maybe we just don't let them do that..... that's a job for a handler!


OK. Now I'm stopping working on the weekend!

gregw:
I used my jet lag as a force for good and converted this branch to one in which the Processor is returned from offer: #7624    I like it!  Now my preferred branch/option   Although some handlers could do with a bit of a cleanup to avoid some lambdas.

joakime:
@sbordet this PR is now way out of date.
IIRC, this is now a settled topic.
Can we close this?

PR #7624, opened 2022-02-20
Jetty 12.0.x offer returns processor

gregw:
Implemented this alternative to #7623
```java
public interface Handler extends LifeCycle, Destroyable
{
    Processor offer(Request request) throws Exception;
}
```

I like it!


gregw:
This branch is a 5% performance improvement to:
```
Benchmark                   Mode  Cnt        Score       Error  Units
HandlerBenchmark.testPost  thrpt   10  1939445.743 ± 44806.967  ops/s
```
over 12.0.x
```
Benchmark                   Mode  Cnt        Score       Error  Units
HandlerBenchmark.testPost  thrpt   10  1825620.422 ± 17636.821  ops/s
```

This makes sense, as I think returning is decision is always going to be better than grabbing a lock to set a state and then grabbing the lock to check the state.

gregw:
It also has a nice shallow stack with mostly meaningful names:
```
java.lang.Throwable
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.HandlerStackTest$1$1.succeeded(HandlerStackTest.java:57)
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.HttpConnection$SendCallback.onCompleteSuccess(HttpConnection.java:924)
	at org.eclipse.jetty.util/org.eclipse.jetty.util.IteratingCallback.processing(IteratingCallback.java:310)
	at org.eclipse.jetty.util/org.eclipse.jetty.util.IteratingCallback.iterate(IteratingCallback.java:214)
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.HttpConnection$Http1Stream.send(HttpConnection.java:1365)
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.HttpStream$Wrapper.send(HttpStream.java:106)
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.HttpChannel$ChannelRequest$RequestCallback.succeeded(HttpChannel.java:682)
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.handler.EchoHandler$Echo.onCompleteSuccess(EchoHandler.java:93)
	at org.eclipse.jetty.util/org.eclipse.jetty.util.IteratingCallback.processing(IteratingCallback.java:310)
	at org.eclipse.jetty.util/org.eclipse.jetty.util.IteratingCallback.iterate(IteratingCallback.java:214)
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.handler.EchoHandler.process(EchoHandler.java:44)
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.handler.ContextRequest.run(ContextRequest.java:75)
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.handler.ContextHandler$Context.run(ContextHandler.java:737)
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.handler.ContextRequest.process(ContextRequest.java:67)
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.handler.DelayUntilContentHandler$DelayedProcessor.run(DelayUntilContentHandler.java:63)
	at org.eclipse.jetty.util/org.eclipse.jetty.util.thread.SerializedInvoker$Link.run(SerializedInvoker.java:146)
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.HttpConnection$Http1Stream.demandContent(HttpConnection.java:1322)
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.HttpStream$Wrapper.demandContent(HttpStream.java:101)
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.HttpChannel$ChannelRequest.demandContent(HttpChannel.java:560)
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.handler.DelayUntilContentHandler$DelayedProcessor.process(DelayUntilContentHandler.java:55)
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.Server.process(Server.java:144)
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.HttpChannel$RunHandle.run(HttpChannel.java:381)
	at org.eclipse.jetty.core.server@12.0.0-SNAPSHOT/org.eclipse.jetty.core.server.HttpConnection.onFillable(HttpConnection.java:429)
	at org.eclipse.jetty.io/org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:317)
	at org.eclipse.jetty.io/org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:100)
	at org.eclipse.jetty.io/org.eclipse.jetty.io.SocketChannelEndPoint$1.run(SocketChannelEndPoint.java:101)
	at org.eclipse.jetty.util/org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:894)
	at org.eclipse.jetty.util/org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1038)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

gregw:
I really like this branch:
 - [x] No state APIs on Request (OK there is `Content.Provider`, but that could be an arg of `Processor` if we really want).
 - [X] Simple handlers have simple implementations (OK the method is now called `process` rather than `handle`, but it is pretty good)
 - [X] Shallow stack.  The offer/accept conversation is no longer in every stack frame
 - [X] Performance. Fastest branch we have tried so far.
 - [X] Complexity. A clear simple offer enquiry of the handler tree is followed by a clear simple process call.
 - [X] Async. Offer cannot be accepted async, but processing is fully async.

gregw:
@lorban @sbordet ready for serious review.   The `s/offer/handle/` is done.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextHandler.java
```diff
@@ -446,40 +439,42 @@ public void destroy()
...
+        // wrap might fail (eg ServletContextHandler could not match a servlet)
         if (scoped == null)
-            return; // TODO 404? 500? Error dispatch ???
+            return null;
 
-        _context.call(scoped);
+        return scoped.asProcessor(_context.get(scoped));
+    }
+
+    void processMove(Request request, Response response, Callback callback)
```
sbordet: Rename to be more explicit such as `processMovedPermanently` or `send301`

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/StatisticsHandler.java
```diff
@@ -94,32 +95,70 @@ public void failed(Throwable x)
...
+        {
+            getProcessor().process(this, response, this);
+        }
+
+        @Override
+        public void succeeded()
+        {
+            try
+            {
+                Callback.super.succeeded();
```
sbordet: No need to call `super`.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/gzip/GzipHandler.java
```diff
@@ -30,74 +29,63 @@
...
-                @Override
-                public long getContentLength()
-                {
-                    // TODO hide the content length
-                    return -1;
-                }
+        @Override
+        public void process(Request ignored, Response response, Callback callback) throws Exception
+        {
+            getProcessor().process(this, new GzipResponse(response), callback);
```
sbordet: No need to override, this should just call `super`.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Server.java
```diff
@@ -114,50 +112,42 @@ public Server(@Name("threadpool") ThreadPool pool)
...
-            // Try to accept to test if already accepted
-            Response response = request.accept();
-            if (response != null)
-            {
-                Callback callback = response.getCallback();
-                response.writeError(HttpStatus.NOT_FOUND_404, "Not accepted", callback);
-            }
+            Processor processor = handle(customized);
+            accepted = true;
+            request.setAccepted();
```
sbordet: Should not this be called `setHandled()` now?

sbordet (changes requested):
Small things to fix, otherwise good.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Server.java
gregw: All it is used for is to enable the Content.Provider, so I how about `request.enableContent()`

gregw: ahhh it enables the callback as well.  So `request.enableProcessing()` it is then!

gregw:
Ooops, I accidentally merged in the changes from #7601 here (InvocationType), but I think they are not controversial so I'm going to leave them here.
This is now with the new naming of `Handler.Processor extends Handler.Abstract implements Request.Processor`

lorban:
There is a compilation error caused by the `takeAsImmutable()` change.

lorban:
:+1: I like the `Handler` and `Processor` aspects of this PR.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/HttpConnection.java
```diff
@@ -129,6 +129,12 @@ public HttpConnection(HttpConfiguration configuration, Connector connector, EndP
             LOG.debug("New HTTP Connection {}", this);
     }
 
+    @Override
+    public InvocationType getInvocationType()
+    {
+        return getServer().getInvocationType();
```
sbordet, line 135: Don't think this is right?
The same connection could be processed by different handlers, which have different invocation types.
Why do we even ask to a `HttpConnection` its invocation type?

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Request.java
```diff
@@ -293,22 +299,22 @@ public Request getWrapped()
         }
     }
 
-    abstract class ProcessingWrapper extends Wrapper implements Handler.Processor
+    abstract class ProcessingWrapper extends Wrapper implements Processor
```
sbordet, line 302: By all the other naming, this is a `WrapperProcessor`.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/StatisticsHandler.java
```diff
@@ -103,6 +103,7 @@ public void failed(Throwable x)
     {
         private final LongAdder _bytesRead;
         private final LongAdder _bytesWritten;
+        private Callback _callback;
 
         public StatsRequest(Request request, LongAdder bytesRead, LongAdder bytesWritten)
```
sbordet: Class should be `private` and called `StatisticsRequest` similar to its enclosing `Handler`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Blocking.java
```diff
@@ -31,7 +31,7 @@
  *
  * <h2>Non shared Runnable</h2>
  * <pre>
- *     try(Blocking.Runnable onAction = Blocking.runnable())
+ *     try(BlockUtil.Runnable onAction = BlockUtil.runnable())
```
sbordet: No `BlockUtil` please, just `Blocking`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/thread/Invocable.java
```diff
@@ -29,7 +29,7 @@
  */
 public interface Invocable
 {
-    static ThreadLocal<Boolean> __nonBlocking = new ThreadLocal<>();
+    ThreadLocal<Boolean> __nonBlocking = new ThreadLocal<>();
```
sbordet, line 32: Why non-`static`? I don't think it's correct.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/AbstractConnection.java
```diff
@@ -32,7 +32,7 @@
  * will schedule a callback to {@link #onFillable()} or {@link #onFillInterestedFailed(Throwable)}
  * as appropriate.</p>
  */
-public abstract class AbstractConnection implements Connection
+public abstract class AbstractConnection implements Connection, Invocable
```
sbordet, line 35: Why does `AbstractConnection` needs to be `Invocable`? 
We should rely on `Callback`s to be `Invocable`, the connection itself seems too coarse of an object.

sbordet (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Blocking.java
gregw: Ooops that's just old javadoc!

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/HttpConnection.java
gregw: The InvocationType of the entire server is set when the server is started.   We can't have individual invocations for each individual request because:
1) we don't know who will process a request until we have called handle, by which time it is too late to go back to the execution strategy, unless we break up handle and process at the level of the managed selector.
2) handlers can be changed dynamically at run time, so if we do query the handlers for the invocation type, they will then race with any dynamic changes - unless we put in a global lock

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Request.java
gregw: This class is primarily a request Wrapper, that you can call `asProcessor` on to get some utility logic to provide a Processor wrapper that substitutes the request in the processing chain.  It is an implementation detail that this class is-a Processor and it could be implemented with an internal private (or dynamic) processor.... but that would avoid the benefit of creating only 1 object.   The class is only used as a Wrapper and never directly used as a Processor.

So I'm happy to consider better names, but it is-not-a Processor.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/thread/Invocable.java
gregw: It's in an interface. It is always static.  Just getting rid of an intellij warning

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/AbstractConnection.java
gregw: XyzConnections use the plain `fillInterested()` registration which uses the `ReadCallback` in AbstractConnection.  Previously that was always BLOCKING.  We now need it to be Invocable if we want to get non blocking dispatch for the very first part of the request.

Alternately, I guess we could make HttpConnection not use `fillInterested()` and instead use `tryFillInterested(Callback)`, but then h2 and h3 would have to do the same.

We should probably review the whole `tryFillInterest()` thing

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/HttpConnection.java
gregw: I have often wondered if ExecutionStrategy should work with `Supplier<Runnable>`s rather than just `Runnable`s.  That way, any task could produce another task that could be put back in the execution strategy, potentially with a different invocation type.
We'd then have the `HttpConnection` having a NON_BLOCKING type as a task that would call the `Handler.handle` methods to determine a Processor, which would then be returned from the first task as the next task and the invocationtype re-evaluated for the next dispatch.    But that is a big change, beyond the scope of the PR.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Request.java
gregw: If the Processor was a final field within the wrapper class, would the optimizer treat an allocation of the wrapper as 1 object or 2?

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/DelayedHandler.java
```diff
@@ -14,123 +14,150 @@
...
     @Override
-    public void handle(Request request) throws Exception
+    public Request.Processor handle(Request request) throws Exception
     {
-        // Defer to extended class to accept the request
-        Response response = accept(request);
+        Handler handler = getHandler();
+        if (handler == null)
+            return null;
+        Request.Processor processor = handler.handle(request);
```
sbordet:
```suggestion
        Request.Processor processor = super.handle(request);
```

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/DelayedHandler.java
```diff
@@ -14,123 +14,150 @@
...
+            if (request.getContentLength() <= 0 && !request.getHeaders().contains(HttpHeader.CONTENT_TYPE))
+                return processor;
+
+            return new OnContentProcessor(request, processor);
         }
+    }
 
-        @Override
-        public Response getResponse()
+    private static class OnContentProcessor implements Request.Processor, Runnable
```
sbordet: For consistency, rename to `UntilContentProcessor`?

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/DelayedHandler.java
```diff
@@ -139,70 +166,93 @@ public QualityOfService(int permits)
...
-            if (response == null)
-                return null;
-            // wrap the response so that completion can be intercepted
-            return new QoSResponse(request, response);
+            return new QosProcessor(request, processor);
         }
 
-        @Override
-        protected void schedule(Request request, Runnable handle)
+        private class QosProcessor implements Request.Processor, Callback
```
sbordet: Rename to `QualityOfServiceProcessor`.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/StatisticsHandler.java
```diff
@@ -94,32 +95,72 @@ public void failed(Throwable x)
             }
         });
 
-        try
+        StatsRequest statsRequest = new StatsRequest(request, bytesRead, bytesWritten);
+        return statsRequest.asProcessor(super.handle(statsRequest));
+    }
+
+    private class StatsRequest extends Request.ProcessingWrapper implements Callback
```
sbordet: Rename to `StatisticsRequest`.

jetty-core/jetty-core-server/src/test/java/org/eclipse/jetty/core/server/handler/EchoHandler.java
```diff
@@ -13,43 +13,49 @@
...
-    public void handle(Request request) throws Exception
+    public void process(Request request, Response response, Callback callback)
     {
-        Response response = request.accept();
         response.setStatus(200);
         String contentType = request.getHeaders().get(HttpHeader.CONTENT_TYPE);
         if (StringUtil.isNotBlank(contentType))
             response.setContentType(contentType);
+        if (request.getHeaders().contains(HttpHeader.TRAILER))
+            response.getTrailers();
```
sbordet: Get the response trailers, but then do nothing with them?

lorban (approved):

sbordet (approved):
LGTM, but please do the renamings.

jetty-core/jetty-core-server/src/test/java/org/eclipse/jetty/core/server/handler/EchoHandler.java
gregw: Well there is a side effect of getting the trailers that prepares to response to accept trailers (ie forces chunking when it commits).
So you need to get them early before writing them.   I then get them again later rather than pass them around.   But that's a little wrong, so I'll pass the trailers into the Echo class. Standby....

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/DelayedHandler.java
gregw: oh but our stack traces will look so much deeper :)

PR #7626, opened 2022-02-21
add used but non declared dependencies and fix some scope

olamy:

### DIFF
--- a/javadoc/pom.xml
+++ b/javadoc/pom.xml
@@ -395,6 +395,12 @@
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-jspc-maven-plugin</artifactId>
       <scope>provided</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>javax.annotation</groupId>
+          <artifactId>javax.annotation-api</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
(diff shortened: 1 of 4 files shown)


joakime (dismissed):

janbartel (approved):

javadoc/pom.xml
```diff
@@ -395,6 +395,12 @@
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-jspc-maven-plugin</artifactId>
       <scope>provided</scope>
+      <exclusions>
```
janbartel, line 398: This isn't going to help with #7677, as it's not the `javax.annotation` jar that is the problem, it is the `findbugs jsr305` jar that is the issue.

jetty-maven-plugin/pom.xml
```diff
@@ -128,20 +128,33 @@
...
+    </dependency>
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-artifact</artifactId>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-core</artifactId>
+      <scope>provided</scope>
```
janbartel, line 157: Will have to add an exclusion for `findbugs jsr305`.

janbartel (commented):
See comments in code

jetty-maven-plugin/pom.xml
joakime: I read this change as not exposing maven-core / findbugs jsr305 on the plugin classpath.
Perhaps we should add a new `it` test for this, something that can replicate the report in #7677 and then prove out that this PR will fix things?

olamy: this should not expose transitive dependencies of maven-core because `provided`.
Yup I will have a look to add an IT.

PR #7633, opened 2022-02-21
Issue #7632 Make osgi build work with snapshot jsp jars

### LINKED ISSUE #7632: Make jetty-10 amenable to testing with SNAPSHOT jsp jars
**jetty-10**

Use the same `pom.xml` code as in jetty-11 to convert a `version` that may include `SNAPSHOT` into an osgi-acceptable form:

```xml
      <plugin>
        <groupId>org.codehaus.mojo</groupId>
        <artifactId>build-helper-maven-plugin</artifactId>
        <executions>
          <execution>
              <id>set-jsp-api-version</id>
              <phase>validate</phase>
              <goals>
                  <goal>parse-version</goal>
              </goals>
              <configuration>
                  <versionString>${jsp.impl.version}</versionString>
                  <propertyPrefix>jspImpl</propertyPrefix>
              </configuration>
          </execution>
        </executions>
      </plugin>
```

janbartel:
Closes #7632

### DIFF
--- a/jetty-osgi/jetty-osgi-boot-jsp/pom.xml
+++ b/jetty-osgi/jetty-osgi-boot-jsp/pom.xml
@@ -44,6 +44,25 @@
 
   <build>
     <plugins>
+      <!-- necessary because x.y.x-SNAPSHOT is not a valid osgi version -->
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>build-helper-maven-plugin</artifactId>
+        <executions>
+          <execution>
+              <id>set-jsp-api-version</id>
+              <phase>validate</phase>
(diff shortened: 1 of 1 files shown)


joakime:
Interesting.

Does this just create a new maven property for use in the manifest?
or does it also impact the installed artifacts too? (eg: the installed pom)

joakime (approved):

olamy (approved):

janbartel:
@joakime it creates a property, which you can use just in the maven build, or you can use it to affect how you generate the manifest. In this case we're using it to set the minimum version of packages in generated manifest. It essentially just lops off any "-SNAPSHOT" from the maven version property, which isn't valid for osgi.

PR #7644, opened 2022-02-23
Jetty Core APIs review and javadocs.

sbordet:

### DIFF
--- a/jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Handler.java
+++ b/jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Handler.java
@@ -35,7 +35,7 @@ import org.slf4j.LoggerFactory;
  * which may recursively forward it to their children {@code Handler}s, until one of them
  * returns a non-null {@link Request.Processor}.</p>
  * <p>Returning a non-null {@code Request.Processor} indicates that the {@code Handler}
- * wants to process the HTTP request, and subsequent sibling or children {@code Handler}s
+ * will process the HTTP request, and subsequent sibling or children {@code Handler}s
  * are not invoked.</p>
  * <p>If none of the {@code Handler}s returns a {@code Request.Processor}, a default HTTP 404
  * response is generated.</p>
--- a/jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Handler.java
+++ b/jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Handler.java
@@ -14,14 +14,12 @@
 package org.eclipse.jetty.core.server;
 
 import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
 
+import org.eclipse.jetty.core.server.handler.ErrorProcessor;
 import org.eclipse.jetty.util.Callback;
 import org.eclipse.jetty.util.annotation.ManagedAttribute;
 import org.eclipse.jetty.util.annotation.ManagedObject;
-import org.eclipse.jetty.util.annotation.ManagedOperation;
 import org.eclipse.jetty.util.component.ContainerLifeCycle;
 import org.eclipse.jetty.util.component.Destroyable;
 import org.eclipse.jetty.util.component.LifeCycle;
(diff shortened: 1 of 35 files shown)


jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Handler.java
```diff
@@ -30,85 +28,147 @@
...
- * until the request is handled and a response is produced.  Handlers are asynchronous,
- * so handling may happen during or after a call to {@link #handle(Request)}.
+ * <p>A Jetty component that handles HTTP requests, of any version (HTTP/1.1, HTTP/2 or HTTP/3).</p>
+ * <p>{@code Handler}s are organized in a tree structure.</p>
+ * <p>An incoming HTTP request is first delivered to the {@link Server} instance
+ * (itself the root {@code Handler}), which forwards it to one or more children {@code Handler}s,
+ * which may recursively forward it to their children {@code Handler}s, until one of them
+ * returns a non-null {@link Request.Processor}.</p>
+ * <p>Returning a non-null {@code Request.Processor} indicates that the {@code Handler}
+ * wants to process the HTTP request, and subsequent sibling or children {@code Handler}s
```
gregw:
```suggestion
 * will process the HTTP request, and subsequent sibling or children {@code Handler}s
```

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Handler.java
```diff
@@ -30,85 +28,147 @@
...
 
         /**
-         * @param byclass the child handler class to get
-         * @param <T> the type of handler
-         * @return first handler of all handlers contained by this handler and it's children of the passed type.
+         * @param type the type of {@code Handler}
+         * @param <T> the type of {@code Handler}
+         * @return an immutable collection of {@code Handler}s of the given type, descendants of this {@code Handler}
+         */
+        <T extends Handler> List<T> getDescendantsByClass(Class<T> type);
```
gregw:
```suggestion
        <T extends Handler> List<T> getDescendants(Class<T> type);
```

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Handler.java
```diff
@@ -30,85 +28,147 @@
...
+
+        /**
+         * @param type the type of {@code Handler}
+         * @param <T> the type of {@code Handler}
+         * @return the first {@code Handler} of the given type, descendants of this {@code Handler},
+         * or null if no such {@code Handler} exist
+         */
+        <T extends Handler> T getDescendantByClass(Class<T> type);
```
gregw:
```suggestion
        <T extends Handler> T getDescendant(Class<T> type);
```

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Handler.java
```diff
@@ -30,85 +28,147 @@
...
+
+        /**
+         * @param handler the child {@code Handler}
+         * @param type the type of {@code Handler}
+         * @param <T> the type of {@code Handler}
+         * @return the {@code Handler.Container} of the given type, parent of the given {@code Handler}
          */
-        <T extends Handler> T getChildHandlerByClass(Class<T> byclass);
+        default <T extends Handler.Container> T getContainerByClass(Handler handler, Class<T> type)
```
gregw:
```suggestion
        default <T extends Handler.Container> T getContainer(Handler handler, Class<T> type)
```

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Handler.java
```diff
@@ -398,6 +452,11 @@ public void removeHandler(Handler handler)
         }
     }
 
+    /**
+     * <p>A {@link Handler.Abstract} that implements {@link Request.Processor},
+     * therefore handling any {@link Request} so that subclasses only need to
+     * implement {@link #process(Request, Response, Callback)}.</p>
+     */
```
gregw:
```suggestion
    /**
     * <p>A {@link Handler} that itself implements {@link Request.Processor}
     * and that returns itself from a call to {@link Handler#handle(Request)}.
     * Subclasses only need to implement 
     * {@link #process(Request, Response, Callback)}.</p>
     */
```

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Request.java
```diff
@@ -30,140 +31,261 @@
...
+ * the request or not.</p>
+ * <p>During the handling phase, the {@code Request} APIs can be used, but its content cannot be read.
+ * Attempting to read the {@code Request} content during the handling phase results in an
+ * {@link IllegalStateException} to be thrown.</p>
+ * <p>A {@code Handler} that handles the request returns a {@link Processor}, that is then invoked
+ * to process the request and the response (the processing phase).</p>
+ * <p>Only during the processing phase the {@code Request} content can be read.</p>
+ * <p>The typical idiom to read request content is the following:</p>
+ * <pre>{@code
+ * class YourHandler
```
gregw:
```suggestion
 * class YourHandler extends Handler.Processor
```

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Request.java
```diff
@@ -30,140 +31,261 @@
...
+ *
+ *             // Normal content, process it.
+ *             processContent(content);
+ *             // Release the content after processing.
+ *             content.release();
+ *
+ *             if (content.isLast())
+ *             {
+ *                 // Reached end-of-file.
+ *                 return;
```
gregw: You need to generate response and succeed callback here.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Request.java
gregw: I'm not sure `Request` is the place for the full handler example.
It should link to the `Handler` which can document how a request is handled.  And also link to `Content.Provider` which can have an example of processing content like you have below

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Request.java
```diff
@@ -299,30 +455,53 @@ public Request getWrapped()
...
+     *
+     *         // Wrap the processor.
+     *         wrapped.wrapProcessor(processor);
+     *
+     *         return wrapped;
+     *     }
+     * }
+     * </pre>
+     */
+    class WrapperProcessor extends Wrapper implements Processor
```
gregw, line 478: I don't hate this name, but I preferred ProcessingWrapper, because it is primarily is-a wrapper.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Request.java
```diff
@@ -299,30 +455,53 @@ public Request getWrapped()
...
+     *         // Wrap the request.
+     *         WrapperProcessor wrapped = new YourWrapperProcessor(request);
+     *
+     *         // Delegate handling using the wrapped request.
+     *         Processor processor = super.handle(wrapped);
+     *
+     *         // Wrap the processor.
+     *         wrapped.wrapProcessor(processor);
+     *
+     *         return wrapped;
```
gregw: This code is wrong if the processor is null.

I think I preferred `asProcessor` as it avoids this easily made mistake.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ProxiedRequestHandler.java
```diff
@@ -72,6 +72,10 @@ public ConnectionMetaData getConnectionMetaData()
                 return proxiedFor;
             }
         };
-        return wrapProcessor(super.handle(wrapper), wrapper);
+        Request.Processor processor = super.handle(wrapper);
+        if (processor == null)
+            return null;
+        wrapper.setProcessor(processor);
+        return wrapper;
```
gregw: So much nicer as:
```suggestion
        return wraper.asProcessor(super.handle(wrapper));
```

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/RewriteHandler.java
```diff
@@ -23,14 +23,18 @@
...
     public Request.Processor handle(Request request) throws Exception
     {
-        Request rewritten = rewrite(request);
-        return wrapProcessor(super.handle(rewritten), rewritten);
+        Request.WrapperProcessor rewritten = rewrite(request);
+        Request.Processor processor = super.handle(rewritten);
+        if (processor == null)
+            return null;
+        rewritten.setProcessor(processor);
+        return rewritten;
```
gregw: simpler as:
```suggestion
        Request.WrapperProcessor rewritten = rewrite(request);
        return rewritten.asProcessor(super.handle(rewritten));
```

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextHandler.java
```diff
@@ -461,7 +461,13 @@ public void destroy()
...
             return null;
 
-        return scoped.asProcessor(_context.get(scoped));
+        Request.Processor processor = _context.get(scoped);
+        if (processor == null)
+            return null;
+
+        scoped.setProcessor(processor);
+
+        return scoped;
```
gregw: I preferred:
```suggestion
        return scoped.asProcessor(_context.get(scoped));
```

gregw (changes requested):
Mostly OK.... a few quibbles..... but I really preferred the `asProcessor` style.   It will prevent bugs like the one you made in the javadoc!

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Request.java
sbordet: Agreed.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Request.java
sbordet: I see it's mostly used as a `Processor`. In fact, the only reason to use this class is to wrap another `Processor` via `asProcessor()`, so it's all about `Processor`s.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Request.java
sbordet: Done.

gregw (approved):
LGTM

PR #7649, opened 2022-02-24
Jetty-12 Requests.getContext()

gregw:
+ Added ServerContext concept
 + Added decoration to the context
 + readded the content scope listener


jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Context.java
```diff
@@ -0,0 +1,43 @@
+package org.eclipse.jetty.core.server;
```
sbordet, line 14: Missing copyright header.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/HttpChannel.java
```diff
@@ -404,9 +405,8 @@ public InvocationType getInvocationType()
             if (metaData == null)
                 new Throwable().printStackTrace();
             Objects.requireNonNull(metaData);
-            _requests++;
+            _requestId = ++_requests;
```
sbordet: Does not look thread safe, and 2B seems like a too small number... I would use a `long`.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextHandler.java
```diff
@@ -620,6 +702,13 @@ public Object getAttribute(String name)
             return super.getAttribute(name);
         }
 
+        @Override
+        public Request.Processor getErrorProcessor()
+        {
+            return ContextHandler.this.getErrorProcessor();
```
sbordet, line 707: Should not this check for `null` and if so, call `super`?

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextHandler.java
```diff
@@ -631,17 +720,19 @@ public String toString()
             return "Context@" + ContextHandler.this.toString();
```
sbordet: Let's use the usual `getSimpleName()` + `hashCode()` rather than `"Context@"` which is wrong because it's a `ServerContext`.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextHandler.java
```diff
@@ -655,17 +746,25 @@ public Path getResourceBase()
...
+                enterScope(request);
                 return supplier.get();
             }
             finally
             {
+                exitScope(request);
                 __context.set(lastContext);
                 if (loader != null)
                     Thread.currentThread().setContextClassLoader(lastLoader);
             }
```
sbordet, line 757: This whole `try/finally` block is repeated multiple times... can it be factored out once, with only the 1 line that changes be passed in as a function?

sbordet (changes requested):

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextHandler.java
gregw: Maybe, but I didn't want to add any more capturing lambdas.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/HttpChannel.java
gregw: It is thread safe because a HttpChannel is only used for one stream and thus cannot be called concurrently.  Specifically it is protected with:
```java

            if (_request != null)
                throw new IllegalStateException();
            _request = new ChannelRequest(request);
```

I've made it a long

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextHandler.java
gregw: Note that currently the `ContextHandler.ScopedContext` does not wrap the `Server.ServerContext` nor whatever context the request might have had when received.     The main reason for this is the complexity of attributes.  We have the contexts persistent and transient(lost at stop) attributes arranged as a layer.   We'd need another layer to delegate to the server attributes in the server context.

I guess that is doable and would make contexts wrappers just like other aspects of the API.  Probably best not to be different?

For now I've done an explicit null check and call to the server errorProcessor.     Your thoughts?

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextHandler.java
```diff
@@ -750,6 +865,54 @@ public void run(Runnable task)
...
+            if (factory != null)
+                return factory.decorate(o);
+            factory = getServer().getBean(DecoratedObjectFactory.class);
+            if (factory != null)
+                return factory.decorate(o);
+            return o;
+        }
+
+        @Override
+        public void destroy(Object o)
```
lorban, line 876: I do not get what this `destroy` contract is about. What does it mean to destroy a context? And what does it mean to have a destroyed context still registered as a bean?

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextHandler.java
gregw: Hey, I like the new method:
```java
            return "%s@%x".formatted(getClass().getSimpleName(), hashCode());
```

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Context.java
```diff
@@ -0,0 +1,56 @@
...
+ * A Context is also an {@link Executor}, which allows tasks to be run by a thread pool, but scoped
+ * to the classloader and any other aspects of the context.   Methods {@link #run(Runnable)},
+ * {@link #call(Invocable.Callable)} and {@link #accept(Consumer, Throwable)}
+ * are also provided to allow various functional interfaces to be called scoped to the context,
+ * without being executed in another thread.
+ * <p>
+ * A Context is also a {@link Decorator}, allowing objects to be decorated in a context scope.
+ *
+ */
+public interface Context extends Attributes, Decorator, Executor
```
lorban, line 34: It feels like implementing `Executor` is either too little or too much. Since there are three other helper methods, maybe we should consider introducing a `JettyExecutor` ?

And what's the difference between `Context.run(Runnable task)` and `Executor.execute(Runnable command)`?

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/HttpChannel.java
```diff
@@ -500,7 +496,13 @@ public HttpURI getHttpURI()
...
 
         @Override
-        public String getPath()
+        public Context getContext()
+        {
+            return getServer().getContext();
+        }
+
+        @Override
+        public String getPathInContext()
```
lorban, line 504: The relation between `getContext().getContextPath()` and this method is unclear, and the fact that they're on different classes add to the confusion.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Request.java
```diff
@@ -128,10 +126,9 @@
      */
     HttpURI getHttpURI();
 
-    // TODO: remove?
-    // TODO: with isSecure() there seems to be a reason to have getOriginalHttpURI() and getHttpURI()
-    //  because why only replacing scheme and path, and not authority, query, etc.?
-    String getPath();
+    Context getContext();
+
+    String getPathInContext();
```
lorban, line 134: Same as above: the relation between `getPathInContext()` and `getContext().getContextPath()` is unclear. Maybe it's the lack of javadoc that ticks me?

lorban (changes requested):

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextHandler.java
gregw: will try fixing this later

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/handler/ContextHandler.java
gregw: This is the destroy from the Decorator, for CDI integration etc.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Context.java
gregw: Removed some the the helper methods.  The execution one is needed as it is the only one that uses the server threadpool.  I've added comments to make that clear.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/HttpChannel.java
gregw: I've imprived the javadoc, but in short: `request.getHttpURI.getPath() == request.getContext().getContextPath() + request.getPathInContext()`

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Request.java
gregw: how is it now?

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/HttpChannel.java
lorban: Okay, that's much clearer now.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/Request.java
lorban: LGTM

lorban (approved):

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/HttpChannel.java
```diff
@@ -81,7 +81,7 @@ public long getLongValue()
     private final HttpConfiguration _configuration;
     private final SerializedInvoker _serializedInvoker;
     private final Attributes _requestAttributes = new Attributes.Lazy();
-    private int _requests;
+    private long _requests;
```
sbordet, line 84: Unused now, remove it.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/HttpChannel.java
```diff
@@ -406,7 +406,6 @@ public InvocationType getInvocationType()
             Objects.requireNonNull(metaData);
             _requests++;
```
sbordet, line 407: Unused, remove this line.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/HttpChannel.java
```diff
@@ -463,15 +462,11 @@ public void clearAttributes()
...
-        public void execute(Runnable task)
-        {
-            _server.getThreadPool().execute(task);
-        }
-
         @Override
         public String getId()
         {
+            if (_id == null)
+                _id = getConnectionMetaData().getId() + "#" + _stream.getId();
```
sbordet, line 469: Not sure about this now. `_stream` may be `null` and must be read with a lock held, so maybe not worth it to use the `_stream` to generate the id of a request.

sbordet (changes requested):

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/HttpChannel.java
gregw: We will still need to count requests cycle per HttpChannel.  So I'm incline to leave it as the right place to be counting (if you count elsewhere you can miss bad requests)

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/HttpChannel.java
gregw: The whole ID thing needs thought and review.  It was never the subject of this PR.  Can't we punt that until we actually pin down what the ID requirements are?

PR #7654, opened 2022-02-25
Jetty-12 onCommit semantic

gregw:
This is a draft attempt at adding onCommit semantics.
Good thing are:
 + predetermined response headers are now only added at commit time (so immune from resets).
 + headers are recorded in the response, so can be logged correctly (i.e. not added by generator later)

Bad things are:
 - method is called with lock held.  This is a show stopper I think!  Specially when sessions are added that will write to databases in the onCommit

So struggling on how to have an atomic commit operation, but still allow response headers to be customized... potentially asynchronously

This needs more thought.


gregw:
@sbordet @lorban This needs more thought!!!!

gregw:
Big refactor of HttpFields so now we can have a Builder interface which the ResponseHttpFields implements. 
The onCommit call is now called prepareResponse and it is called without the lock held.

lorban:
The core of the change seems to be the introduction of a mutable `HttpFields` interface such as to have the following:

`HttpFields` read-only interface
`HttpFields.Builder` read-write interface
`HttpFields.Immutable` read-only implementation
`HttpFields.Mutable` read-write implementation

I find the idea sound (even if I'm not sure about having both ro and rw interfaces) but the naming / packaging is IMHO terribly confusing.

`HttpFields.Builder` has nothing to do with what's commonly called a builder, and quite far from the builder we introduced with the `Index` interface. That, and having the mutable interface an inner interface of the immutable one, mixed with both implementations next to it makes for an odd-looking API:

```java
HttpFields.Builder requestFields = HttpFields.build();
requestFields.put("User-Agent", client.getClass().getName() + "/" + Jetty.VERSION);
```

vs

```java
Index<Boolean> ASSUMED_CONTENT_METHODS = new Index.Builder<Boolean>()
        .caseSensitive(false)
        .with(HttpMethod.POST.asString(), Boolean.TRUE)
        .with(HttpMethod.PUT.asString(), Boolean.TRUE)
        .build();
```

How about the following instead?

```java
public interface HttpFields
{
    interface Mutable extends HttpFields
    {

        class Builder
        {
        }
    }

    class Builder
    {
    }
}

/*package-private*/ class ImmutableHttpFields implements HttpFields
{
}

/*package-private*/ class MutableHttpFields implements HttpFields.Mutable
{
}



```

gregw:
@lorban that's better.  Only 17 files changed as well - bonus!

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/ResponseHttpFields.java
```diff
@@ -17,106 +17,139 @@
...
 import org.eclipse.jetty.http.HttpFields;
 import org.eclipse.jetty.http.HttpHeader;
 import org.eclipse.jetty.http.HttpHeaderValue;
 
-class ResponseHttpFields extends HttpFields.Mutable
+class ResponseHttpFields implements HttpFields.Mutable
 {
-    private volatile boolean _readOnly;
+    private final Mutable _fields;
+    private final AtomicBoolean _committed = new AtomicBoolean();
```
lorban, line 32: It seems a simple `volatile boolean` would be enough.

jetty-core/jetty-core-server/src/main/java/org/eclipse/jetty/core/server/SecureRequestCustomizer.java
```diff
@@ -191,7 +191,7 @@ private void formatSTS()
     }
 
     @Override
-    public Request customize(Request request)
+    public Request customize(Request request, HttpFields.Mutable response)
```
lorban: `HttpFields.Mutable responseHeaders`?

jetty-core/jetty-core-server/src/test/java/org/eclipse/jetty/core/server/HttpConnectionTest.java
```diff
@@ -530,17 +530,21 @@ public void testSimple() throws Exception
     @Test
     public void testEmptyNotPersistent() throws Exception
     {
-        String response = _connector.getResponse("GET /R1?empty=true HTTP/1.0\r\n" +
-            "Host: localhost\r\n" +
-            "\r\n");
+        String response = _connector.getResponse("""
```
lorban, line 533: Is that `\r\n` -> `\r` change expected?

jetty-core/jetty-core-server/src/test/java/org/eclipse/jetty/core/server/HttpConnectionTest.java
```diff
@@ -530,17 +530,21 @@ public void testSimple() throws Exception
...
+            """);
 
         int offset = 0;
         checkContains(response, offset, "HTTP/1.1 200");
 
-        response = _connector.getResponse("GET /R1?empty=true HTTP/1.1\r\n" +
-            "Host: localhost\r\n" +
-            "Connection: close\r\n" +
-            "\r\n");
+        response = _connector.getResponse("""
```
lorban, line 542: Is that `\r\n` -> `\r` change expected?

jetty-core/jetty-core-server/src/test/java/org/eclipse/jetty/core/server/HttpConnectionTest.java
```diff
@@ -549,19 +553,25 @@ public void testEmptyNotPersistent() throws Exception
     @Test
     public void testEmptyPersistent() throws Exception
     {
-        String response = _connector.getResponse("GET /R1?empty=true HTTP/1.0\r\n" +
-            "Host: localhost\r\n" +
-            "Connection: keep-alive\r\n" +
-            "\r\n");
+        String response = _connector.getResponse("""
```
lorban, line 556: Is that `\r\n` -> `\r` change expected?

jetty-core/jetty-core-server/src/test/java/org/eclipse/jetty/core/server/HttpConnectionTest.java
```diff
@@ -549,19 +553,25 @@ public void testEmptyNotPersistent() throws Exception
...
+            GET /R1?empty=true HTTP/1.0\r
+            Host: localhost\r
+            Connection: keep-alive\r
+            \r
+            """);
+
+        System.err.println(response);
```
lorban: Unneeded call to `System.err.println()`.

jetty-core/jetty-core-server/src/test/java/org/eclipse/jetty/core/server/HttpConnectionTest.java
```diff
@@ -549,19 +553,25 @@ public void testEmptyNotPersistent() throws Exception
...
 
         int offset = 0;
         offset = checkContains(response, offset, "HTTP/1.1 200");
         checkContains(response, offset, "Content-Length: 0");
         checkNotContained(response, offset, "Connection: close");
 
-        response = _connector.getResponse("GET /R1?empty=true HTTP/1.1\r\n" +
-            "Host: localhost\r\n" +
-            "\r\n");
+        response = _connector.getResponse("""
```
lorban, line 568: Is that `\r\n` -> `\r` change expected?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpFields.java
```diff
@@ -536,15 +790,15 @@ default boolean isEqualTo(HttpFields that)
      *
      * <p>The cookie handling provided by this class is guided by the Servlet specification and RFC6265.
      */
-    class Mutable implements Iterable<HttpField>, HttpFields
+    class MutableHttpFields implements Mutable
```
lorban, line 793: I wish this class could be hidden, and be out of the interface.

lorban (changes requested):
There's still a few minor things here and there but overall, I find this to be much better.

jetty-core/jetty-core-server/src/test/java/org/eclipse/jetty/core/server/HttpConnectionTest.java
gregw: yes, text blocks have an implicit \n in them

jetty-core/jetty-core-server/src/test/java/org/eclipse/jetty/core/server/HttpConnectionTest.java
gregw: ditto

jetty-core/jetty-core-server/src/test/java/org/eclipse/jetty/core/server/HttpConnectionTest.java
gregw: ditto

jetty-core/jetty-core-server/src/test/java/org/eclipse/jetty/core/server/HttpConnectionTest.java
gregw: ditto

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpFields.java
gregw: yeah - I tried a number of things.... but then I actually do need to extend it sometimes, so the constructors are all protected, so that gives some limits

lorban (approved):

PR #7660, opened 2022-02-28
Issue #7635 - implement maxBlockedStreams logic in QpackDecoder

### LINKED ISSUE #7635: QPACK decoder should fail connection if the encoder blocks more than SETTINGS_QPACK_BLOCKED_STREAMS
Our `QpackEncoder` will not block more than `SETTINGS_QPACK_BLOCKED_STREAMS` streams. 
But the `QpackDecoder` also needs to perform verification of this and fail the connection if this is violated.

https://datatracker.ietf.org/doc/html/draft-ietf-quic-qpack-21#section-2.1.2
```
If a decoder encounters more blocked streams than it promised
   to support, it MUST treat this as a connection error of type
   QPACK_DECOMPRESSION_FAILED.
```

lachlan-roberts:
## Closes #7635

Add verification that the max blocked streams is not exceeded inside the `QpackDecoder`.


sbordet (changes requested):
Please add test case.

lachlan-roberts:
@sbordet I have copied over some of the same changes from #7802 to help with testing.

sbordet (dismissed):

PR #7687, opened 2022-03-04
remove some site configuration, non used checkstyle profile and turn skip xml into one line property

olamy:

### DIFF
--- a/jetty-jspc-maven-plugin/pom.xml
+++ b/jetty-jspc-maven-plugin/pom.xml
@@ -12,7 +12,7 @@
   <properties>
     <bundle-symbolic-name>${project.groupId}.jspc.plugin</bundle-symbolic-name>
     <jacoco.skip>true</jacoco.skip>
-    <skipTests>true</skipTests>
+    <maven.test.skip>true</maven.test.skip>
   </properties>
   <build>
     <plugins>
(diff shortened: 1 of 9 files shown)


jetty-jspc-maven-plugin/pom.xml
```diff
@@ -11,6 +11,8 @@
   <name>Jetty :: Jetty JSPC Maven Plugin</name>
   <properties>
     <bundle-symbolic-name>${project.groupId}.jspc.plugin</bundle-symbolic-name>
+    <jacoco.skip>true</jacoco.skip>
+    <maven.test.skip.exec>true</maven.test.skip.exec>
```
joakime: Why `maven.test.skip.exec` here?

joakime (changes requested):
One odd new property that doesn't appear before this.

jetty-jspc-maven-plugin/pom.xml
olamy: ah yeah good catch. I mixed some stuff when doing the same thing for 2 branches which are different sorry for that

olamy: oh but wait the surefire plugin configuration.
see next commit :)

joakime (approved):
LGTM

Issue #7688, opened 2022-03-04
Read data to native memory from HttpInput 


jiang-cao:
**Target Jetty version(s)**
jetty-11.0.9
**Enhancement Description**
Our purpose is read data direct to native memory from jetty direct buffers, then we could do CPU heavy compute with native lib without copy data between java heap and native heap.

Jetty has provide flexibility let us inherit class and implement our special code. The problem is some fields/methods in HttpChannel and HttpInput are private and can't be accessed in sub class.

The change is explore necessary methods to sub class. I will prepare a PR to show the change.

jiang-cao:
PR: https://github.com/eclipse/jetty.project/pull/7689

With this change, we implement a new method in sub class of HttpInput.

    public int read(long address, int off, int len) throws IOException, CSException {
    ...
    }

The input data will write to native address.

sbordet:
I'm not sure I understand this change?

If you use a direct `ByteBuffer` to read data from `HttpInput` there is no copy, and you can use the direct `ByteBuffer` with native libraries.

I'd rather add a `read(ByteBuffer)` method than a raw `read(long, ...)` that cannot be used without resorting to `Unsafe`.

Can you detail your exact use case? Would a `read(ByteBuffer)` be enough for you?

jiang-cao:
Thanks for your review!  
In my case, I did want the unsafe method, read the data to native buffer address.  In our case, the native memory is not owned by java layer,  so we prefer to have more flexibility to implement read(long, ...) in sub classes.   

read(ByteBuffer) is a good extension, but not match our current requirement.

joakime:
You haven't said why you need this.
What is your use case?

Most of the existing usages in Jetty are ByteBuffer based, and we are consuming more and more JDK APIs that themselves are ByteBuffer based.
Wouldn't a direct ByteBuffer be sufficient to satisfy your native vs heap memory concerns?

jiang-cao:
Explain more detail about my case:
We receive data from HTTP,  the data will be used do some CPU heavy compute work like md5, CRC, Erasing Coding, then the data will be send to remote node by RDMA NIC.   Avoid memory copy is important in the data path.  The idea way is direct used RDMA registered buffer to receive HTTP data and keep all CPU heavy work under native layer.  Use native buffer address is more straightforward here.

jiang-cao:
@sbordet @joakime any suggestion on what I could do to get this extend ability?

joakime:
[...]

PR #7689, opened 2022-03-04
Issue #7688 Read to native memory from HttpInput

jiang-cao:
Signed-off-by: Jiang Cao <jiang.cao@outlook.com>


jiang-cao:
Try a different approach,  add readContent(ByteBuffer) instead of expose these private fields and extend in sub class.

PR #7690, opened 2022-03-04
jetty 9.4.x remove non used maven plugin

olamy:
- remove some site configuration, non used checkstyle profile and turn skip xml into one line property
- do not skip execution of ITs as well
- bad idea
- fix that

### DIFF
--- a/jetty-maven-plugin/pom.xml
+++ b/jetty-maven-plugin/pom.xml
@@ -14,7 +14,7 @@
     <jetty.stopKey>FREEBEER</jetty.stopKey>
     <jetty.jvmArgs></jetty.jvmArgs>
     <jacoco.skip>true</jacoco.skip>
-    <maven.test.skip>true</maven.test.skip>
+    <maven.test.skip.exec>true</maven.test.skip.exec>
   </properties>
   <build>
     <plugins>
--- a/jetty-maven-plugin/pom.xml
+++ b/jetty-maven-plugin/pom.xml
@@ -13,15 +13,11 @@
     <bundle-symbolic-name>${project.groupId}.maven.plugin</bundle-symbolic-name>
     <jetty.stopKey>FREEBEER</jetty.stopKey>
     <jetty.jvmArgs></jetty.jvmArgs>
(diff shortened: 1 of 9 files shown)


jetty-maven-plugin/pom.xml
```diff
@@ -13,15 +13,11 @@
     <bundle-symbolic-name>${project.groupId}.maven.plugin</bundle-symbolic-name>
     <jetty.stopKey>FREEBEER</jetty.stopKey>
     <jetty.jvmArgs></jetty.jvmArgs>
+    <jacoco.skip>true</jacoco.skip>
+    <maven.test.skip.exec>true</maven.test.skip.exec>
```
joakime, line 17: Why do we have this for `jetty-maven-plugin` at all?
And we had a similar maven-surefire-plugin/config/skip=true on `jetty-jspc-maven-plugin` and you didn't include this `maven.test.skip.exec` property there.

joakime (changes requested):
our 2 maven plugins have some oddities (inconsistencies?) between them with surefire/skip

olamy:
> our 2 maven plugins have some oddities (inconsistencies?) between them with surefire/skip

because there is a difference between the 2 plugins:
- `jetty-jspc-maven-plugin` doesn't have any src/test/java source code so no test are compiled and surefire doesn't need to be skipped as there is not classes to scan  (target/test-classes is just empty)
- `jetty-maven-plugin` have code in src/test/java which is technically a class (TestGetContent) only here to be included as a common test for most of the ITs test. But this test cannot be runned apart from the maven-invoker-plugin tests (if we do not skip it simply fail as it need the context of our IT tests). That's why `maven-invoker-plugin` have the following configuration `<addTestClassPath>true</addTestClassPath>`  and most of the IT test have this 
```          
          <dependenciesToScan>
            <dependency>org.eclipse.jetty:jetty-maven-plugin</dependency>
          </dependenciesToScan>
```

joakime (approved):
Sounds good to me. 👍

PR #7729, opened 2022-03-12
Jetty 12 hackathon results

gregw:
Merge the results of the hackathon back to the main 12.0.x branch.

This is too big to review in detail, but we should scan to see if anything stands out as needing extra review


gregw:
OK, this is definitely too large to review here.   Will close and we can review some other way.

PR #7751, opened 2022-03-18
HttpInput read to ByteBuffer

### LINKED ISSUE #7688: Read data to native memory from HttpInput

jiang-cao:
add int read(ByteBuffer buffer) in HttpInput.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
@@ -130,9 +130,30 @@ public class HttpInput extends ServletInputStream implements Runnable
         }
     }
 
-    private int get(Content content, byte[] bytes, int offset, int length)
+    private int get(Content content, ByteBuffer des)
     {
-        int consumed = content.get(bytes, offset, length);
+        int consumed;
+        if (des.hasArray())
+        {
+            consumed = content.get(des.array(), des.arrayOffset() + des.position(), des.remaining());
+            des.position(des.position() + consumed);
+        }
+        else
+        {
+            var capacity = des.remaining();
+            var src = content.getByteBuffer();
+            if (src.remaining() > capacity)
+            {
+                var copy = src.duplicate().limit(src.position() + capacity);
+                des.put(copy);
+                src.position(src.position() + (capacity - copy.remaining()));
+            }
+            else
+            {
+                des.put(src);
+            }
+            consumed = capacity - des.remaining();
+        }
         _contentConsumed.add(consumed);
         return consumed;
     }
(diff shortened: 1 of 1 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
```diff
@@ -248,6 +269,11 @@ public int read() throws IOException
 
     @Override
     public int read(byte[] b, int off, int len) throws IOException
+    {
+        return read(ByteBuffer.wrap(b, off, len));
```
jiang-cao, line 273: The main flow in read(byte[] b, int off, int len)  is same as read(ByteBuffer buffer).  Wrap the array can avoid these duplicate code.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
```diff
@@ -130,9 +130,30 @@ public void addInterceptor(Interceptor interceptor)
         }
     }
 
-    private int get(Content content, byte[] bytes, int offset, int length)
+    private int get(Content content, ByteBuffer des)
     {
-        int consumed = content.get(bytes, offset, length);
+        int consumed;
+        if (des.hasArray())
```
jiang-cao, line 136: May I delete the des.hasArray branch?  The else branch can support same functionality.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
```diff
@@ -239,7 +260,7 @@ public int read() throws IOException
     {
         try (AutoLock lock = _contentProducer.lock())
         {
-            int read = read(_oneByteBuffer, 0, 1);
+            int read = read(ByteBuffer.wrap(_oneByteBuffer, 0, 1));
```
jiang-cao, line 263: _oneByteBuffer can change to ByBuffer to avoid this wrap().

sbordet:
@jiang-cao please base your work off the `jetty-10.0.x` branch, we will take care of merging to 11.

Please do not wrap `byte[]` into `ByteBuffer` because these methods are performance sensitive and we do not want to allocate a `ByteBuffer` for every read.

Let's start with adding the method you need, and if there are duplication happy to refactor, but let's not start by refactoring :)

Thanks!

jiang-cao:
Thanks @sbordet 
I create https://github.com/eclipse/jetty.project/pull/7778 base on jetty-10.0.x.

PR #7753, opened 2022-03-19
simplify m-compiler-p configuration by removing redundant configuration

olamy:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -13,9 +13,7 @@
 
   <properties>
     <!-- build -->
-    <compiler.source>11</compiler.source>
-    <compiler.target>11</compiler.target>
-    <compiler.release>11</compiler.release>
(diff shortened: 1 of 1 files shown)


joakime (changes requested):
This change, while valid to maven, will cause problems with various IDEs out there.
As the IDE will set the compiler level of the project based on source, not target or release.

This change will cause problems with our community.
I say we should not do this.

olamy:
> This change, while valid to maven, will cause problems with various IDEs out there. As the IDE will set the compiler level of the project based on source, not target or release.
> 
> This change will cause problems with our community. I say we should not do this.

fair enough. but modern IDE should manage this.
as it's part of the JEP: https://openjdk.java.net/jeps/247

```
for any N, the -source and -target options, as those are automatically set to N.
```

PR #7763, opened 2022-03-21
Issue #7615 HttpServletResponse.encodeURL() should support relative paths

### LINKED ISSUE #7615: HttpServletResponse.encodeURL not working for URLs starting with ../

gregw:
+ use presence of scheme to gate parsing as HttpURI

This replaces #7616

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
@@ -321,6 +321,9 @@ public class Response implements HttpServletResponse
     @Override
     public String encodeURL(String url)
     {
+        if (url == null)
+            return null;
+
         final Request request = _channel.getRequest();
         SessionHandler sessionManager = request.getSessionHandler();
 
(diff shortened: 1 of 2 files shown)


joakime (approved):
Yup, this is cleaner.
I like the removal of HttpURI usage too.

PR #7764, opened 2022-03-21
Issue #7683 Use direct buffers for gzip input/output, if configured.

### LINKED ISSUE #7683: GZIPContentDecoder ignores setUseInputDirectByteBuffers setting and always uses non-direct buffers (causing GC locking)
**Jetty version(s)**

10.0.8

```
openjdk version "17.0.2" 2022-01-18
OpenJDK Runtime Environment (build 17.0.2+8-86)
OpenJDK 64-Bit Server VM (build 17.0.2+8-86, mixed mode, sharing)
```

**OS type/version**

```
Ubuntu 20.04.2 LTS
```

**Description**

GZIPContentDecoder ignores setUseInputDirectByteBuffers setting and always uses non-direct buffers:

https://github.com/eclipse/jetty.project/blob/jetty-10.0.8/jetty-http/src/main/java/org/eclipse/jetty/http/GZIPContentDecoder.java#L435

This came to my attention because Java Inflater class when used with non-direct buffers will use JNI which in turn can cause GC issues. This is well explained in the README of this project (pure java inflate implementation, which has nothing to do with the issue except for having a good explanation of the problem) https://github.com/google/pjflate:

> This is not an attempt to make it fast. In fact, it's much slower than the Java standard Inflater. This is more like to avoid the GCLocker impact.
>
> JNI code in general is not GC aware, and in order to avoid arrays to be moved by JVM GC, it needs to pin primitive arrays. While pinning, JVM GC is paused. After this pinning is done, JVM can trigger a GC (GCLocker initiated GC). Because no GC is done while this pinning, depending on the heuristics JVM can trigger a full GC. For certain applications, it's more favorable to keep concurrent GCs running to prevent a full GC.
>
> [...]
>
> Starting from Java SE 11, the standard Inflater can take a ByteBuffer. In Open JDK implementation, it checks if it's a direct buffer and if it is, it doesn't lock GC. If GCLocker is a concern and Open JDK 11 or later is used, it's better to use a direct buffer with the standard Inflater. It gives you a better speed than pure Java implementations and you can avoid the GCLocker issue.

**How to reproduce?**

https://github.com/eclipse/jetty.project/blob/jetty-10.0.8/jetty-http/src/main/java/org/eclipse/jetty/http/GZIPContentDecoder.java#L435

jaroslawr:
Fix for #7683.

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/GZIPContentDecoder.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/GZIPContentDecoder.java
@@ -46,6 +46,7 @@ public class GZIPContentDecoder implements Destroyable
     private long _value;
     private byte _flags;
     private ByteBuffer _inflated;
+    private boolean _useDirectBuffers;
 
     public GZIPContentDecoder()
     {
@@ -59,15 +60,26 @@ public class GZIPContentDecoder implements Destroyable
 
     public GZIPContentDecoder(ByteBufferPool pool, int bufferSize)
     {
-        this(new InflaterPool(0, true), pool, bufferSize);
+        this(new InflaterPool(0, true), pool, bufferSize, false);
+    }
+
+    public GZIPContentDecoder(ByteBufferPool pool, int bufferSize, boolean useDirectBuffers)
+    {
+        this(new InflaterPool(0, true), pool, bufferSize, useDirectBuffers);
     }
 
     public GZIPContentDecoder(InflaterPool inflaterPool, ByteBufferPool pool, int bufferSize)
+    {
+        this(inflaterPool, pool, bufferSize, false);
+    }
+
+    public GZIPContentDecoder(InflaterPool inflaterPool, ByteBufferPool pool, int bufferSize, boolean useDirectBuffers)
     {
         _inflaterEntry = inflaterPool.acquire();
         _inflater = _inflaterEntry.get();
         _bufferSize = bufferSize;
         _pool = pool;
+        _useDirectBuffers = useDirectBuffers;
         reset();
     }
 
@@ -209,6 +221,13 @@ public class GZIPContentDecoder implements Destroyable
                             if (buffer == null)
                                 buffer = acquire(_bufferSize);
 
+                            if (_inflater.needsInput())
+                            {
+                                if (!compressed.hasRemaining())
+                                    return;
+                                _inflater.setInput(compressed);
+                            }
+
                             try
                             {
                                 int pos = BufferUtil.flipToFill(buffer);
@@ -226,12 +245,6 @@ public class GZIPContentDecoder implements Destroyable
                                 buffer = null;
                                 if (decodedChunk(chunk))
                                     return;
-                            }
-                            else if (_inflater.needsInput())
-                            {
-                                if (!compressed.hasRemaining())
-                                    return;
-                                _inflater.setInput(compressed);
                             }
                             else if (_inflater.finished())
                             {
@@ -432,7 +445,7 @@ public class GZIPContentDecoder implements Destroyable
      */
     public ByteBuffer acquire(int capacity)
     {
-        return _pool == null ? BufferUtil.allocate(capacity) : _pool.acquire(capacity, false);
+        return _pool == null ? BufferUtil.allocate(capacity) : _pool.acquire(capacity, _useDirectBuffers);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


jetty-http/src/main/java/org/eclipse/jetty/http/GZIPContentDecoder.java
```diff
@@ -209,6 +221,13 @@ else if ((_flags & 0x2) == 0x2)
                             if (buffer == null)
                                 buffer = acquire(_bufferSize);
 
+                            if (_inflater.needsInput())
+                            {
+                                if (!compressed.hasRemaining())
+                                    return;
+                                _inflater.setInput(compressed);
+                            }
+
```
jaroslawr, line 230: (I moved this up because otherwise there is a call done to _inflater.inflate with no input set yet)

joakime:
@jaroslawr this PR has test failures in both Java 11 and Java 17.

```
java.util.zip.ZipException: Corrupt GZIP trailer
	at java.base/java.util.zip.GZIPInputStream.readTrailer(GZIPInputStream.java:226)
	at java.base/java.util.zip.GZIPInputStream.read(GZIPInputStream.java:120)
	at org.eclipse.jetty.util@10.0.9-SNAPSHOT/org.eclipse.jetty.util.IO.copy(IO.java:166)
	at org.eclipse.jetty.util@10.0.9-SNAPSHOT/org.eclipse.jetty.util.IO.copy(IO.java:116)
	at org.eclipse.jetty.servlet@10.0.9-SNAPSHOT/org.eclipse.jetty.servlet.GzipHandlerTest.testAsyncLargeResponse(GzipHandlerTest.java:446)
```

jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipHttpOutputInterceptor.java
```diff
@@ -318,54 +328,69 @@ protected Action process() throws Exception
...
+                        return Action.SUCCEEDED;
                     }
-                    else
+
+                    int pos = BufferUtil.flipToFill(_buffer);
+                    int produced = deflater.deflate(_buffer, _syncFlush ? Deflater.SYNC_FLUSH : Deflater.NO_FLUSH);
+                    BufferUtil.flipToFlush(_buffer, pos);
+                }
+                else
+                {
```
lachlan-roberts, line 324: I would like to see us remove this branch completely and always use the `ByteBuffer` APIs for `Deflater` and `CRC32`.

If the buffer is not direct and does not have an array then `CRC32.update` will not copy the entire buffer at once, but allocate a single byte array of max size 4k and use that to read from the buffer.

jaroslawr: Done!

jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipHttpOutputInterceptor.java
```diff
@@ -264,8 +264,16 @@ public boolean mightCompress()
         public GzipBufferCB(ByteBuffer content, boolean complete, Callback callback)
```
lachlan-roberts, line 263: `_copy` can be removed as it is not used anymore.

lachlan-roberts (changes requested):
I'm happy with this change.

But I think we should also hear from @sbordet before merging.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipHttpOutputInterceptor.java
jaroslawr: Done

joakime:
The builds are taking much much longer with this latest change.
Old builds: 1 hour 23 minutes (on average).
New builds on this PR: 2 hours (and build is canceled due to overall build timeout)

jaroslawr:
@joakime Builds on other PRs/branches are now also timing out:
https://jenkins.webtide.net/blue/organizations/jenkins/jetty.project/detail/PR-7682/8/pipeline
https://jenkins.webtide.net/blue/organizations/jenkins/jetty.project/detail/jetty-10.0.x/1404/pipeline
So I suppose it's some issue with the CI environment and not with the change itself?

joakime:
@jaroslawr I see that now too.  We are investigating.

joakime:
Merged (manually) into `jetty-10.0.x` (and them merged up to `jetty-11.0.x`)

See: commit a35719367bf67611863f601eb9c150a5881556cc

PR #7769, opened 2022-03-22
Issue #5681 - clearer warning on JVM Arg / System Property use

### LINKED ISSUE #5681: Unrecognized jetty-home/start.jar command line option not reported clearly
**jetty-10.0.x**

If I use an unknown command option to start the distro like so:
[941] java -jar ../jetty-home/start.jar --foople

Jetty simply starts. I would expect at least an "unknown option" type warning, maybe even not starting.

joakime:
Fixes #5681 Unknown distro command ignored

### DIFF
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
@@ -501,11 +501,11 @@ public class Main
 
         if (args.hasJvmArgs() || args.hasSystemProperties())
         {
-            StartLog.warn("System properties and/or JVM args set.  Consider using --dry-run or --exec");
+            StartLog.warn("Unknown Arguments detected.  Consider using --dry-run or --exec");
             if (args.hasSystemProperties())
-                args.getSystemProperties().forEach((k, v) -> StartLog.warn("  Detected JVM System Property: %s=%s", k, System.getProperty(k)));
+                args.getSystemProperties().forEach((k, v) -> StartLog.warn("  Argument: -D%s=%s (interpreted as a System property, from %s)", k, System.getProperty(k), v));
             if (args.hasJvmArgs())
-                args.getJvmArgs().forEach((jvmArg) -> StartLog.warn("  Detected JVM Arg: %s", jvmArg));
+                args.getJvmArgSources().forEach((jvmArg, source) -> StartLog.warn("  Argument: %s (interpreted as a JVM argument, from %s)", jvmArg, source));
         }
 
         ClassLoader cl = classpath.getClassLoader();
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
@@ -502,6 +502,10 @@ public class Main
         if (args.hasJvmArgs() || args.hasSystemProperties())
         {
             StartLog.warn("System properties and/or JVM args set.  Consider using --dry-run or --exec");
+            if (args.hasSystemProperties())
+                args.getSystemProperties().forEach((k, v) -> StartLog.warn("  Detected JVM System Property: %s=%s", k, System.getProperty(k)));
+            if (args.hasJvmArgs())
+                args.getJvmArgs().forEach((jvmArg) -> StartLog.warn("  Detected JVM Arg: %s", jvmArg));
         }
 
         ClassLoader cl = classpath.getClassLoader();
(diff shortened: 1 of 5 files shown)


jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
```diff
@@ -502,6 +502,10 @@ public void run()
         if (args.hasJvmArgs() || args.hasSystemProperties())
         {
             StartLog.warn("System properties and/or JVM args set.  Consider using --dry-run or --exec");
+            if (args.hasSystemProperties())
+                args.getSystemProperties().forEach((k, v) -> StartLog.warn("  Detected JVM System Property: %s=%s", k, System.getProperty(k)));
+            if (args.hasJvmArgs())
+                args.getJvmArgs().forEach((jvmArg) -> StartLog.warn("  Detected JVM Arg: %s", jvmArg));
```
joakime: Other language options.

```
  Detected as JVM System Property: %s=%s
  Detected as JVM Arg: %s
```

```
  Detected possible JVM System Property: %s=%s
  Detected possible JVM Arg: %s
```

joakime: From @sbordet 

```
WARN  : Unknown Arguments detected.  Consider using --dry-run or --exec
WARN  :   Argument: -Dzed.key=0.value (interpreted as a system property, from <command-line>)
WARN  :   Argument: --foople (interpreted as a JVM argument)
```

joakime:
@sbordet new output syntax committed.

Example output:

```
WARN  : Unknown Arguments detected.  Consider using --dry-run or --exec
WARN  :   Argument: -Dzed.key=0.value (interpreted as a System property, from <command-line>)
WARN  :   Argument: --zed-0-zed (interpreted as a JVM argument, from /home/joakim/code/jetty/jetty.project-10.0.x/jetty-start/target/tests/test-base-example-unknown/start.d/zed.ini)
WARN  :   Argument: --foople (interpreted as a JVM argument, from <command-line>)
```

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
```diff
@@ -1331,10 +1357,12 @@ public void parse(final String rawarg, String source)
...
         if (arg.startsWith("-"))
         {
+            StartLog.debug("Unrecognized Arg: %s (from %s)", arg, source);
+
             // Only add non-duplicates
-            if (!jvmArgs.contains(arg))
+            if (!jvmArgSources.containsKey(arg))
             {
-                jvmArgs.add(arg);
+                jvmArgSources.put(arg, source);
```
sbordet: Use `putIfAbsent()` here.

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
```diff
@@ -331,21 +331,23 @@ public void dumpEnvironment(PrintStream out)
...
+        jvmArgSources.forEach((key, sourceRef) ->
+            {
+                String value = System.getProperty(key);
+                String source = StartLog.isDebugEnabled() ? '(' + sourceRef + ')' : "";
+                if (value != null)
+                    out.printf(" %s = %s %s%n", key, value, source);
+                else
+                    out.printf(" %s %s%n", key, source);
+            }
+        );
```
sbordet: Weird braces and parens alignment, can you make it normal?

sbordet (changes requested):

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
joakime: Disagree.
This arg can be overridden by a the same arg later on, which could have a different source.
The last source is what's used, so we replace each time to know the eventually used source.

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
joakime: This passes checkstyle.
I'll give it whirl.

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
sbordet: I don't understand? There is a `!containsKey()` followed by `put()` so it's equivalent to `putIfAbsent()`.

sbordet (changes requested):

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
joakime: You are correct that `putIfAbsent` is an appropriate replacement for what you reviewed.
I disagree that it should be used at all, the check `!containsKey()` shouldn't even be there too.
I pushed an update.

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
joakime: Done.

sbordet (approved):

PR #7778, opened 2022-03-25
HttpInput read into ByteBuffer

### LINKED ISSUE #7688: Read data to native memory from HttpInput

jiang-cao:
Add int read(ByteBuffer buffer) in HttpInput.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
@@ -143,9 +143,10 @@ public class HttpInput extends ServletInputStream implements Runnable
         var src = content.getByteBuffer();
         if (src.remaining() > capacity)
         {
-            var copy = src.duplicate().limit(src.position() + capacity);
-            des.put(copy);
-            src.position(src.position() + (capacity - copy.remaining()));
+            int limit = src.limit();
+            src.limit(src.position() + capacity);
+            des.put(src);
+            src.limit(limit);
         }
         else
         {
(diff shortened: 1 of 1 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
```diff
@@ -292,6 +311,51 @@ public int read(byte[] b, int off, int len) throws IOException
...
+                // If EOF do we need to wake for allDataRead callback?
+                if (onContentProducible())
+                    scheduleReadListenerNotification();
+                return -1;
+            }
+
+            throw new AssertionError("no data, no error and not EOF");
+        }
+    }
+
```
jiang-cao: The code in  read(ByteBuffer buffer) is duplicate of read(byte[] b ...)  except get(content, buffer).  But I didn't find a good way can avoid this duplication with out wrap byte[] to ByteBuffer.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
```diff
@@ -137,6 +137,25 @@ private int get(Content content, byte[] bytes, int offset, int length)
...
 
+    private int get(Content content, ByteBuffer des)
+    {
+        var capacity = des.remaining();
+        var src = content.getByteBuffer();
+        if (src.remaining() > capacity)
+        {
+            var copy = src.duplicate().limit(src.position() + capacity);
+            des.put(copy);
+            src.position(src.position() + (capacity - copy.remaining()));
```
sbordet: There is no need to allocate via `duplicate()`, this should work:

```suggestion
            int limit = src.limit();
            src.limit(src.position() + capacity);
            des.put(src);
            src.limit(limit);
```

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
sbordet: There are 2 methods now:

A=`read(ByteBuffer buffer)` and B=`read(byte[] b, int offset, int len)`.

Create another method:

C=`read(ByteBuffer buffer, byte[] b, int offset, int len)`, so that A calls C via `read(buffer, null, -1, -1)` and B calls C via `read(null, b, offset, len)`.

Inside C, the only line that changes can now be written as:
`int read = buffer == null ? get(content, b, offset, len) : get(content, buffer);`

sbordet (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
jiang-cao: Thanks, it's good approach.

jetty-server/src/main/java/org/eclipse/jetty/server/HttpInput.java
jiang-cao: It's more efficient, thanks!

sbordet (approved):

sbordet:
Merged to `jetty-12.0.x` via #8701.

PR #7802, opened 2022-03-30
HTTP/3 QPACK - do not expect section ack for zero required insert count

lachlan-roberts:
https://datatracker.ietf.org/doc/html/draft-ietf-quic-qpack-21#section-4.4.1

> After processing an encoded field section whose declared Required
   Insert Count is not zero, the decoder emits a Section Acknowledgment
   instruction.

>  If an encoder receives a Section Acknowledgment instruction referring
   to a stream on which every encoded field section with a non-zero
   Required Insert Count has already been acknowledged, this MUST be
   treated as a connection error of type QPACK_DECODER_STREAM_ERROR.

### DIFF
--- a/jetty-http3/http3-qpack/src/test/java/org/eclipse/jetty/http3/qpack/SectionAcknowledgmentTest.java
+++ b/jetty-http3/http3-qpack/src/test/java/org/eclipse/jetty/http3/qpack/SectionAcknowledgmentTest.java
@@ -15,6 +15,7 @@ package org.eclipse.jetty.http3.qpack;
 
 import java.nio.ByteBuffer;
 
+import org.eclipse.jetty.http3.qpack.QpackException.SessionException;
 import org.eclipse.jetty.http3.qpack.internal.instruction.SectionAcknowledgmentInstruction;
 import org.eclipse.jetty.util.BufferUtil;
 import org.junit.jupiter.api.BeforeEach;
@@ -25,6 +26,8 @@ import static org.eclipse.jetty.http3.qpack.QpackTestUtil.toBuffer;
 import static org.eclipse.jetty.http3.qpack.QpackTestUtil.toMetaData;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.greaterThan;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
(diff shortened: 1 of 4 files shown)


sbordet (changes requested):
Test case?

jetty-http3/http3-qpack/src/test/java/org/eclipse/jetty/http3/qpack/SectionAcknowledgmentTest.java
```diff
@@ -0,0 +1,79 @@
...
+        instruction = _decoderHandler.getInstruction();
+        assertNull(instruction);
+    }
+
+    @Test
+    public void testSectionAcknowledgmentForZeroRequiredInsertCountOnEncoder() throws Exception
+    {
+        // Encode a header with only a value contained in the static table.
+        ByteBuffer buffer = encode(_encoder, 0, toMetaData("GET", "/", "http"));
+        System.err.println(BufferUtil.toDetailString(buffer));
```
sbordet: Remove the `System.err.`.

jetty-http3/http3-qpack/src/test/java/org/eclipse/jetty/http3/qpack/SectionAcknowledgmentTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+        // Parsing a section ack instruction on the encoder when we are not expecting it should result in QPACK_DECODER_STREAM_ERROR.
+        SectionAcknowledgmentInstruction instruction = new SectionAcknowledgmentInstruction(0);
+        ByteBuffer instructionBuffer = toBuffer(instruction);
+        QpackException error = assertThrows(QpackException.class, () -> _encoder.parseInstructions(instructionBuffer));
```
sbordet: You want to test for `QpackException.StreamException` here to be sure it's not a SESSION error.

sbordet (changes requested):

jetty-http3/http3-qpack/src/test/java/org/eclipse/jetty/http3/qpack/SectionAcknowledgmentTest.java
lachlan-roberts: done

jetty-http3/http3-qpack/src/test/java/org/eclipse/jetty/http3/qpack/SectionAcknowledgmentTest.java
lachlan-roberts: It is a `SessionException` and thats what I would expect it to be. The rfc says it must be treated "as a connection error".

I have added this check to the test case.

sbordet (approved):

PR #7803, opened 2022-03-30
unwrap exception until we get the first non ServletException, as this can be wrap of wrap of wrap when using ContextHandlerCollection

olamy:
TCK test fail when using some embedded Jetty 


olamy:
I wonder if we should make unwrap true per default as it is expected by servlet spec?

joakime (approved):

PR #7834, opened 2022-04-04
Issue #7748 - Backport of allow override of path mapping behavior in ServletContextHandler

### LINKED ISSUE #7748: Allow overriding of url-pattern mapping in ServletContextHandler to allow for regex or uri-template matching

gregw:
backport and fix to 9.4 for #7748

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
@@ -140,7 +140,7 @@ public class RegexPathSpec extends AbstractPathSpec
         Matcher matcher = getMatcher(path);
         if (matcher.matches())
         {
-            if (matcher.groupCount() >= 1)
+            if (_group == PathSpecGroup.PREFIX_GLOB && matcher.groupCount() >= 1)
             {
                 int idx = matcher.start(1);
                 if (idx > 0)
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/UriTemplatePathSpec.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/UriTemplatePathSpec.java
@@ -358,7 +358,7 @@ public class UriTemplatePathSpec extends AbstractPathSpec
         Matcher matcher = getMatcher(path);
         if (matcher.matches())
         {
-            if (matcher.groupCount() >= 1)
+            if (_group == PathSpecGroup.PREFIX_GLOB && matcher.groupCount() >= 1)
             {
                 int idx = matcher.start(1);
                 if (idx > 0)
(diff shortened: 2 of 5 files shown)


joakime:
```
[INFO] --- license-maven-plugin:4.1:check (check-java-headers) @ jetty-project ---
[INFO] Checking licenses...
[WARNING] Missing header in: /home/jenkins/agent/workspace/jetty.project_PR-7834/jetty-servlet/src/test/java/org/eclipse/jetty/servlet/RegressionServletTest.java
[WARNING] Missing header in: /home/jenkins/agent/workspace/jetty.project_PR-7834/jetty-servlet/src/test/java/org/eclipse/jetty/servlet/RegexServletTest.java
```

Keep in mind that Jetty 9.4.x has the old EPL 1.0 license headers.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
```diff
@@ -140,7 +140,7 @@ public String getPathMatch(String path)
         Matcher matcher = getMatcher(path);
         if (matcher.matches())
         {
-            if (matcher.groupCount() >= 1)
+            if (_group == PathSpecGroup.PREFIX_GLOB && matcher.groupCount() >= 1)
```
lachlan-roberts, line 143: What is the reason for this change? 
This change is not present in 10 & 11.

gregw: @lachlan-roberts my intention is to move this change forward into 10 & 11.
I believe it is needed to make the methods `getPathMatch` and `getPathInfo` symmetric.

Specifically I would expect either `pathMatch==/test && pathInfo==/info` or `pathMatch==/test/info && pathInfo==null`, but never `pathMatch==/test && pathInfo==null`, since that loses information.  Currently we get the later behavior!

So I have added the following test.

```java
    @Test
public void testPathInfo()
    {
        RegexPathSpec spec = new RegexPathSpec("^/test(/.*)?$");
        assertTrue(spec.matches("/test/info"));
        assertThat(spec.getPathMatch("/test/info"), equalTo("/test"));
        assertThat(spec.getPathInfo("/test/info"), equalTo("/info"));

        spec = new RegexPathSpec("^/[Tt]est(/.*)?$");
        assertTrue(spec.matches("/test/info"));
        assertThat(spec.getPathMatch("/test/info"), equalTo("/test/info"));
        assertThat(spec.getPathInfo("/test/info"), nullValue());
    }
```

Note that this issue was not apparent in 10 & 11, because they do not use these methods to set the servletPath and pathInfo, but rather work it out differently (I'll check exactly why when I port this change forward).

lachlan-roberts (dismissed):

joakime (changes requested):
The UriTemplatePathSpec has the same issue with the difference in behavior.

Example path spec ... `/{first}/static/{other}`

gregw:
I just did a force push to squash this to 1 commit so I can start work on porting forward to 10, 11 etc.
yeah I know I always say don't do force pushes.... do as I say and not as I do :)

gregw:
@joakime @lachlan-roberts review please

joakime (approved):

PR #7837, opened 2022-04-05
Fix StatisticsHandler in the case a Handler throws exception.

lachlan-roberts:
If a `Handler` throws, the `StatisticsHandler` thinks it has seen a 200 response because the status code has not yet been set. This PR adds a separate stat `_responsesThrown` which is now incremented instead of using the incorrect status code in the case the handler throws.


joakime:
Having a late point of status tracking, to supplement the tracking in the StatisticsHandler is a good idea.
Maybe having a stats specific HttpChannel.Listener or stats RequestLog that forwards the status codes to the tracking in StatisticsHandler is in our future.

gregw:
Note that we have a late stats point in jetty-12, but I don't think we should make significant changes in <= 11

gregw (approved):

joakime:
@lachlan-roberts is there an open issue for this?

lachlan-roberts:
@joakime yes there is an open private issue which has been linked.

PR #7850, opened 2022-04-06
Fixes #7818 - Regression: allow HttpChannel.Listener.onResponseBegin to modify response headers

### LINKED ISSUE #7818: Modifying of HTTP headers in HttpChannel.Listener#onResponseBegin is no longer possible with Jetty 10
**Jetty version(s)**
Jetty 10.0.8

**Description**
I asked this already on the mailing list: 

I have a Jetty server that was upgraded from 9.4 to 10.0. There is one requirement to do some "header maintenance": Before the response is sent to client, I want all headers created by several addHeaders in a bunch of servlets, filters and other handlers to be cleaned up. In my case, this is deduplication and merging of "vary" and "cache-control" headers to assist NGINX in front so it only gets one header and not multiples.

In Jetty 9.4, I had the following code, added as bean to the connector:

```java
  private static HttpField mergeHttpFields(String name, List<HttpField> fields) {
    return (fields == null) ? null : new HttpField(name, fields.stream()
        .map(HttpField::getValues)
        .flatMap(Stream::of)
        .distinct()
        .collect(Collectors.joining(", "));
  }
  
  private static final HttpChannel.Listener HEADER_MERGER = new HttpChannel.Listener() {
    @Override
    public void onResponseBegin(Request request) {
      final HttpFields headers = request.getResponse().getHttpFields();
      headers.computeField("Vary", MyClass::mergeHttpFields);
      headers.computeField("Cache-Control", MyClass::mergeHttpFields);
    }
  };

  // This is how its added to connector
  connector.addBean(HEADER_MERGER);
```

In 10.0 this no longer compiled, but was easy to fix by using the `HttpFields.Mutable` interface instead of just `HttpFields`(`Response#getHttpFields()` returns a mutable instance).

After trying this out, I noticed that it did not work like it did in 9.4 -- the headers were no longer merged! Debuggig through the code, I figured out that my header merger was actually called before the response was submitted and after it ran, the HttpFields were correctly merged. But on the wire they did not appear, the original headers were sent.

What changed? It looks like it is caused by the change to have mutable vs. immutable `HttpFields` since Jetty 10: Before the `onResponseBegin()` handler is called, the `HttpChannel` has already made an immutable snapshot of the header fields, so all changes to the mutable ones are lost. I think the problem is in `HttpChannel#sendResponse()`: This method calls `Listener#onResponseBegin()` too late, it should call this before `_response.newResponseMetaData()` which makes an *immutable* snapshot of the `HttpFields`: https://github.com/eclipse/jetty.project/blob/a35719367bf67611863f601eb9c150a5881556cc/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java#L1040-L1045

I think the problem could be fixed by shuffling the calls a bit. Maybe it would even be better to add another callback in the Listener like: `onBeforeHeadersSend(HttpFields.Mutable fields)`.

**How to reproduce?**
Use the above code in Jetty 9.4: Here the `Vary` and `Cache-Control` headers are merged using `computeField()`, on Jetty 10, the method is called (I added log statements), but due to the call to `Response#newResponseMetaData()` before the callback, the modification to headers never go to the wire (only the snapshotted ones).

I think fixing this might be a longer process, but it would be good to have a statement by a developer how to fix this with current Jetty version - or do I need to go back to 9.4 for a while?

joakime:
Fixes #7850 - A regression in `HttpChannel.Listener.onResponseBegin` with the ability to modify response headers immediately before commit.

See PR #7851 for Jetty 9.4.x test demonstrating behavior. (Jetty 9.4.x does not need a code change)

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
@@ -1037,11 +1037,12 @@ public abstract class HttpChannel implements Runnable, HttpOutput.Interceptor
 
         if (committing)
         {
+            // Let HttpChannel.Listeners modify the response before commit
+            _combinedListener.onResponseBegin(_request);
             // We need an info to commit
             if (response == null)
                 response = _response.newResponseMetaData();
             commit(response);
-            _combinedListener.onResponseBegin(_request);
             _request.onResponseCommit();
 
             // wrap callback to process 100 responses
--- a/jetty-server/src/test/java/org/eclipse/jetty/server/HttpChannelEventTest.java
+++ b/jetty-server/src/test/java/org/eclipse/jetty/server/HttpChannelEventTest.java
@@ -69,7 +69,7 @@ public class HttpChannelEventTest
         start(new TestHandler()
         {
             @Override
-            protected void handle(HttpServletRequest request, HttpServletResponse response) throws IOException, ServletException
+            protected void handle(HttpServletRequest request, HttpServletResponse response) throws IOException
             {
                 ServletInputStream input = request.getInputStream();
                 int content = input.read();
@@ -111,7 +111,7 @@ public class HttpChannelEventTest
         start(new TestHandler()
         {
             @Override
-            protected void handle(HttpServletRequest request, HttpServletResponse response) throws IOException, ServletException
+            protected void handle(HttpServletRequest request, HttpServletResponse response) throws IOException
             {
                 response.getOutputStream().write(data);
             }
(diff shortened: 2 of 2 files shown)


uschindler (dismissed):
Fix looks fine to me and test also replicates what I did in my code.

Did you test if the test succeeded in 9.4 to confirm it was a regression?

joakime:
> Fix looks fine to me and test also replicates what I did in my code.
> 
> Did you test if the test succeeded in 9.4 to confirm it was a regression?

The 9.4.x test case which shows that this works is in PR #7851

gregw (commented):
Not building?  But looks good other than that.

sbordet (approved):

PR #7864, opened 2022-04-08
#7863 Include value of first accept-encoding header when there is more then one

### LINKED ISSUE #7863: Default servlet drops first accept-encoding header if there is more than one.
**Jetty version(s)**
11.0.9

N/A

**OS type/version**
N/A

**Description**
When `DefaultServlet` handles a request with multiple `accept-encoding` headers, it attempts to combine their values, but it drops the value of the first header (and produces a value with a leading comma).

**How to reproduce?**
Test case to follow.

markslater:
Test and fix for first occurrence of `accept-encoding` header being dropped when more than one `accept-encoding` header is given.

Signed-off-by: markslater <mark.slater@mail.com>

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
@@ -336,6 +336,7 @@ public class ResourceService
         if (headers.hasMoreElements())
         {
             StringBuilder sb = new StringBuilder(key.length() * 2);
+            sb.append(key);
             do
             {
                 sb.append(',').append(headers.nextElement());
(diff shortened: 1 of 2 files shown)


joakime:
This might be a problem all the way back in `jetty-9.4.x` too.

markslater:
Yes, the issue has existed since at least 2016, according to git blame.

How does the project's branching strategy work?  Should I resubmit my pull request against the `jetty-9.4.x` branch, or does it get back-ported from the `jetty-10.0.x` branch?

joakime:
Our branches ...

* `jetty-9.4.x` - in maintenance mode, we'll either do the work there initially and merge up, or cherry pick back into it as needed.
* `jetty-10.0.x` - our current stable mainline development, where all new bugs are typically fixed.
* `jetty-11.0.x` - we merge from `jetty-10.0.x` into here, unless there is jetty 11 specific changes needed.
* `jetty-12.0.x` - our next development tree (work in progress atm)

Don't worry about 9.4.x (i'll take care of that), but if you can rebase this PR against `jetty-10.0.x` that would eliminate a few steps for us :-)

markslater:
Rebased - sorry about that, I must have forgotten to pull before I made the change.

joakime (approved):

joakime:
Merged into `jetty-10.0.x` and merged up to `jetty-11.0.x`.

Cherry-pick to `jetty-9.4.x` handled in PR #7871

PR #7871, opened 2022-04-11
#7863 Include value of first accept-encoding header where multiple accept-encoding headers are given

### LINKED ISSUE #7863: Default servlet drops first accept-encoding header if there is more than one.

joakime:
Cherry-pick of commit 8bcb842fbbd2e4c9f57bfa7718b038786edc0aa0 for `jetty-9.4.x`

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
@@ -327,6 +327,7 @@ public class ResourceService
         if (headers.hasMoreElements())
         {
             StringBuilder sb = new StringBuilder(key.length() * 2);
+            sb.append(key);
             do
             {
                 sb.append(',').append(headers.nextElement());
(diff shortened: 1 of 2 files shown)


gregw (commented):
This code looks rather strange... even before the fix. @lorban can you have a look with your recent resource service experience and check that the while loop with no body is correct.

lorban:
@gregw I agree that this loop is quite convoluted. It feels like it was written that way to optimize the cases where there is no accept-encoding header or when it contains a single value.

The existing code also is buggy in the sense that it forgets to include the first accept-encoding value in the key it builds in case there are multiple values for that header, which poisons the cache with incorrect values.

But it feels like this `_preferredEncodingOrderCache` isn't going to be that efficient as it allows logical duplicates and requires multiple allocation before looking up a multi-accept-encoding value which feels to me like that would defeat most savings bypassing the `QuotedQualityCSV` parser could bring.

I think this fix should go in, but we should revisit this `_preferredEncodingOrderCache` and consider dropping it.

joakime:
Opened issue #7928 about that review of `_preferredEncodingOrderCache` at issue #7928

joakime:
@lorban or @gregw want to approve this then? (so we can get it merged)

lorban (approved):

Issue #7858, opened 2022-04-08
GZipHandler does not play nice with other handlers in HandlerCollection


JensGabe:
**Jetty version(s)**
Jetty 11.0.2

**Java version/vendor** `(use: java -version)`
Temurin 17.02+8

**OS type/version**
Windows Server 2016

**Description**
Handler configuration:
```
HandlerCollection
 - ResourceHandler (root static files - eg contextpath = / )
 - GZipHandler
   - ContextHandlerCollection
     - ServletContextHandler
     - ContextHander (other static files - eg contextpath= /foo )
       - ResourceHandler
 - DefaultHandler
```
The GZIPHandler is configured with minSize=1024 and includes javascript and other text files.

When requesting a root static file with a size over 96K, the GZIPHandler will gzip the part of the content which exceeds 96K - the first 96K of the content is plain text, and the remainder looks GZipped.
(The repsonse content length matched the actual file length, and the browser (chrome) report error content-length mismatch)

If the root static file is a pre-zipped file, it works fine, also for size > 96K.
If the same static file is requested via the /foo resource handler, it works fine.
On my Developer setup (Windows 10, same Java) it works fine - I cannot get it to fail as on the Server installation.

Checking the source code of the various handlers, I noticed that all the handlers i used, except GZipHandler starts with:
```
if (request.isHandled()) return;
```

Reconfiguring handler configuration to use HandlerList instead or move the ResourceHandler into a wrapped ContextHandler in the ContextHandlerCollection will both fix the issue for me.

It seems like the GZipHandler or a buffer handling deeper in the output handling is not working as intended if it is called after another handler handles the request.

**How to reproduce?**
See description.

joakime:
What is the declared context path of your ServletContextHandler ?

JensGabe:
"/" - The request in question never touches the ServletContextHandler.

joakime:
If your `ServletContextHandler` is on context path of `/`, then simplify your entire structure by using Servlet url-patterns properly and eliminating your `ResourceHandler` entries.

```
HandlerCollection
  - GZipHandler
    - ServletContextHandler (contextpath = /foo)
      - DefaultServlet (the default one on url-pattern `/`, using resourceBase from ServletContextHandler, serving static content from root)
      - DefaultServlet (a new one, on url-pattern `/foo/*`, using it's own resourceBase pointing to a different directory on disk)
 - DefaultHandler
```

See example at 
[...]

PR #7872, opened 2022-04-11
Issue #7858 - GzipHandler request.isHandled support

joakime:

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipHandler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipHandler.java
@@ -616,9 +616,6 @@ public class GzipHandler extends HandlerWrapper implements GzipFactory
     @Override
     public void handle(String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response) throws IOException, ServletException
     {
-        if (response.isCommitted() || baseRequest.isHandled())
-            return;
-
         ServletContext context = baseRequest.getServletContext();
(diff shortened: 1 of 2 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipHandler.java
```diff
@@ -616,6 +616,9 @@ public void setInflateBufferSize(int size)
     @Override
     public void handle(String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response) throws IOException, ServletException
     {
+        if (response.isCommitted() || baseRequest.isHandled())
```
gregw: I know we do sometimes check isCommitted elsewhere, but I think isHandled is enough here.

gregw (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipHandler.java
joakime: Actually, this is a bad if statement overall.

Have to rethink this, as we don't un-gzip request content with this if statement here.

joakime: but if the request is handled, do we even want to?

joakime: I submitted an alternative approach.

gregw (dismissed):

joakime:
The tests on CI failed, as the expectations were incorrect.
I improved the expectations and documented the reasoning.
This will need to be re-reviewed

gregw (approved):

PR #7890, opened 2022-04-18
Make deprecation message more useful

pzygielo:


jetty-server/src/main/java/org/eclipse/jetty/server/handler/AllowSymLinkAliasChecker.java
```diff
@@ -37,7 +37,7 @@ public class AllowSymLinkAliasChecker implements AliasCheck
 
     public AllowSymLinkAliasChecker()
     {
-        LOG.warn("Deprecated, use SymlinkAllowedResourceAliasChecker instead.");
+        LOG.warn("AllowSymLinkAliasChecker is deprecated, use SymlinkAllowedResourceAliasChecker instead.");
```
joakime, line 40: Including the name of the class will cause the name to be printed twice on the logging output.

Once from the Logger name itself, and again from the message.

pzygielo: > Including the name of the class will cause the name to be printed twice on the logging output.
> 
> Once from the Logger name itself, and again from the message.

Depends on the configuration of the logger I presume.

In my case - where I used application with jetty embedded - it was not present, thus made the tracking of offending class very hard, based on log only. (Given dozens of other deprecated methods/types used, this log gave no clue what is being deprecated.)
(And it's impossible to solve for me as the client anyway.)

And for example
https://github.com/eclipse/jetty.project/blob/69c208b6e4f8e16429a3e09b4b9e874e3c76cafb/jetty-servlet/src/main/java/org/eclipse/jetty/servlet/ServletContextHandler.java#L705

That's why I propose to include the type in message.

pzygielo:
@joakime - thanks for checking.

I'm not asking for myself, as I've learnt that already. It's rather for unnamed friend from the public.
Feel free to close this PR if you find this change unnecessary.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/AllowSymLinkAliasChecker.java
joakime: > Depends on the configuration of the logger I presume.

If you don't output the timestamp, logger name, and thread name (not the java thread id number) then you have a nearly useless logging configuration that wouldn't help you diagnose anything meaningful.

PR #7896, opened 2022-04-20
Fixed flaky ssl fullMethod test

gregw:
The write of the test request was sometimes failing before the write was complete.  A good response was produced.
In the case of too large messages, we should fail before reading them all.

### DIFF
--- a/jetty-server/src/test/java/org/eclipse/jetty/server/HttpServerTestBase.java
+++ b/jetty-server/src/test/java/org/eclipse/jetty/server/HttpServerTestBase.java
@@ -217,7 +217,7 @@ public abstract class HttpServerTestBase extends HttpServerTestFixture
             catch (Exception e)
             {
                 if (LOG.isDebugEnabled())
-                    LOG.warn("write exception", e);
+                    LOG.debug("write exception", e);
             }
 
             // Read the response.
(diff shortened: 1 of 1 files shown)


jetty-server/src/test/java/org/eclipse/jetty/server/HttpServerTestBase.java
```diff
@@ -209,8 +209,16 @@ public void testFullMethod() throws Exception
...
-            os.flush();
+            try
+            {
+                os.write(buffer);
+                os.flush();
+            }
+            catch (Exception e)
+            {
+                if (LOG.isDebugEnabled())
+                    LOG.warn("write exception", e);
```
joakime: is debug, then warn?

joakime (dismissed):

lorban (dismissed):

jetty-server/src/test/java/org/eclipse/jetty/server/HttpServerTestBase.java
sbordet: Should be `LOG.debug()`, not `LOG.warn()`. It's just a test case, no need to be fancy and we will be consistent with the rest of the code.

sbordet (changes requested):

lorban (approved):

sbordet (approved):

PR #7919, opened 2022-04-26
Fix #7918 Root path spec

### LINKED ISSUE #7918: PathMappings.asPathSpec does not allow root ServletPathSpec
**Jetty version(s)**
all

**Description**

The ServletPathSpec of "" (empty string) is the root only match.  This is not allowed by `PathMappings.asPathSpec`, which can prevent code that does not explicitly use ServletPathSpec from matching root.

gregw:
Handle root pathspec in PathMappings.asPathSpec

Not used in jetty-9, but will be ported forward to where it is used in 10,11 & 12

### DIFF
--- a/jetty-http/src/test/java/org/eclipse/jetty/http/pathmap/PathMappingsTest.java
+++ b/jetty-http/src/test/java/org/eclipse/jetty/http/pathmap/PathMappingsTest.java
@@ -25,6 +25,7 @@ import org.junit.jupiter.params.provider.ValueSource;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.notNullValue;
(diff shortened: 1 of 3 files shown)


jetty-http/src/test/java/org/eclipse/jetty/http/pathmap/PathMappingsTest.java
```diff
@@ -461,4 +462,18 @@ public void testRemoveServletPathSpec()
         assertThat(p.remove(new ServletPathSpec("/a/b/c")), is(true));
         assertThat(p.remove(new ServletPathSpec("/a/b/c")), is(false));
     }
+
+    @Test
+    public void testAsPathSpec()
+    {
+        assertThat(PathMappings.asPathSpec(""), instanceOf(ServletPathSpec.class));
```
joakime, line 469: What about `asPathSpec(null)`, should that be interpreted as ROOT too?

gregw: No,  it is very precisely the empty string that matches "/" (and "/" matches everything - go figure!)

lorban (approved):

PR #7920, opened 2022-04-26
Jetty 10 7918 root pathspec

### LINKED ISSUE #7918: PathMappings.asPathSpec does not allow root ServletPathSpec

gregw:
In addition to cherry-picking asPathSpec fix from 9, this PR makes asPathSpec extensible in ConstraintSecurityHandler in the same way it is now extensible in ServletHandler

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
@@ -199,12 +199,12 @@ public class PathMappings<E> implements Iterable<MappedResource<E>>, Dumpable
 
     public static PathSpec asPathSpec(String pathSpecString)
     {
-        if ((pathSpecString == null) || (pathSpecString.length() < 1))
-        {
-            if (pathSpecString != null)
-                return new ServletPathSpec("");
+        if (pathSpecString == null)
             throw new RuntimeException("Path Spec String must start with '^', '/', or '*.': got [" + pathSpecString + "]");
-        }
+
+        if (pathSpecString.length() == 0)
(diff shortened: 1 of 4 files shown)


jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -201,6 +201,8 @@ public static PathSpec asPathSpec(String pathSpecString)
     {
         if ((pathSpecString == null) || (pathSpecString.length() < 1))
         {
+            if (pathSpecString != null)
```
lorban: The conditions are quite convoluted when written this way. Can't they be rewritten in a clearer way like the following?
```java
if (pathSpecString == null)
    throw new RuntimeException("Path Spec String must start with '^', '/', or '*.': got [" + pathSpecString + "]");
if ("".equals(pathSpecString))
    return new ServletPathSpec("");
return pathSpecString.charAt(0) == '^' ? new RegexPathSpec(pathSpecString) : new ServletPathSpec(pathSpecString);
```

lorban (changes requested):

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
gregw: evolved solutions are mostly ugly!  will do!

lorban (approved):

PR #7923, opened 2022-04-26
Jetty 12.0.x hackathon pathcollection

gregw:


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/ClassFilePredicate.java
```diff
@@ -0,0 +1,51 @@
...
+
+package org.eclipse.jetty.util.paths;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.List;
+import java.util.function.BiPredicate;
+
+public class ClassFilePredicate implements BiPredicate<Path, BasicFileAttributes>
```
gregw, line 22: javadoc needed to explain why this class is needed

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathCollection.java
```diff
@@ -0,0 +1,291 @@
...
+
+import org.eclipse.jetty.util.StringUtil;
+
+/**
+ * Generic Loader for Multiple Path references.
+ * <p>
+ * This fits in the same space as the ResourceCollection of past Jetty versions.
+ * </p>
+ */
+public class PathCollection extends ArrayList<Path> implements AutoCloseable
```
gregw: The `Path` class is immutable, so I think `PathCollection` should also be immutable as it will be often used as a `Path`.

So I think it should instead be an interface with static builder methods:
```java
interface PathCollection extends List<Path>,  AutoCloseable
{
    PathCollection from(Path... path);
    PathCollection from(List<Path> paths, Path... path);
    PathCollection from(List<Path> paths, Predicate<Path> filter);
    
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathCollection.java
```diff
@@ -0,0 +1,291 @@
...
+     * Glob references will only iterate through the level specified and will not traverse
+     * found directories within the glob reference.
+     * </p>
+     *
+     * @param resources the comma {@code ,} or semicolon {@code ;} delimited
+     * String of path references.
+     * @param globDirs true to return directories in addition to files at the level of the glob
+     * @return the PathCollection parsed from input string.
+     */
+    public static PathCollection fromList(String resources, boolean globDirs) throws IOException
```
gregw, line 98: What uses this?  I'm guessing our XML???
We should try doing it in XML without this helper method

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/FileSystemPool.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * A pool of {@code zipfs} based {@code java.nio.file.FileSystem} objects to allow reference
+ * counting of open FileSystem and proper closure when the last reference count is decremented.
+ */
+public class FileSystemPool
```
gregw, line 30: name should indicate it is for jar files:  `JarFileSystemPool` ?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/FileSystemPool.java
```diff
@@ -0,0 +1,114 @@
...
+            URI uri = URI.create("jar:" + jarFile.toUri().toASCIIString());
+            refCount = new FileSystemRefCount();
+            refCount.fileSystem = FileSystems.newFileSystem(uri, env);
+            refCount.rootPath = refCount.fileSystem.getPath("/");
+            cache.put(jarFile, refCount);
+        }
+
+        refCount.count.incrementAndGet();
+        return refCount.fileSystem;
+    }
```
gregw, line 75: This needs to be atomic.  I know it is very unlikely to be hit concurrently, but it is possible.  So a null check then add is not sufficient as two threads can add at the same time.
Probably just needs a bit lock as any of the IfAbsent  style method might still create a new FS that has to be discarded.
Hmmm maybe @lorban has a pool that can be used?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/FileSystemPool.java
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+
+        if (refCount == null)
+            return true;
+
+        int count = refCount.count.decrementAndGet();
+        if (count <= 0)
+        {
+            cache.remove(key);
```
gregw, line 106: this is again a race condition

gregw (commented):
@joakime review comments will come to me because I made the PR....

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathCollection.java
joakime: The old WebAppContext extraClasspath logic has this, it was moved into here, just to support the "*" logic that is present when dealing with classloader style of reference lists (vs resource style reference lists)

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/FileSystemPool.java
joakime: Sure, it could also be called `ZipFsFileSystemPool` (as that's what is being used)

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/FileSystemPool.java
joakime: If the pool class can support this use case, I'm all for moving to a standardized pooling layer.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathCollection.java
```diff
@@ -0,0 +1,291 @@
...
+            throw error;
+    }
+
+    /**
+     * Resolve the first path that exists against the collection of paths.
+     *
+     * @param other the other path to resolve against, assumes a String in unix format, that is already URI decoded from URI space.
+     * @see Path#resolve(String)
+     */
+    public Path resolveFirstExisting(String other)
```
gregw: I think this can just be called `resolve(String)` to make the usage similar to using `'Path`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathCollection.java
```diff
@@ -0,0 +1,291 @@
...
+        return resolveFirst(other, Files::exists);
+    }
+
+    /**
+     * Resolve the first path that matches the predicate.
+     *
+     * @param other the other path to resolve against, assumes a String in unix format, that is already URI decoded from URI space.
+     * @param pathPredicate the predicate of {@code Path} to evaluate for first valid hit.
+     */
+    public Path resolveFirst(String other, Predicate<Path> pathPredicate)
```
gregw, line 222: this again can just be `resolve(String, Predicate<Path>)`

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathCollection.java
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    /**
+     * Find all files (up to 300 levels deep) in the path collection that match the provided {@link BiPredicate}
+     * of &lt;{@link Path}, {@link BasicFileAttributes}&gt;
+     *
+     * @param pathPredicate the predicate to evaluate across all paths in this collection
+     * @return the stream of hits for the find operation.
+     */
+    public Stream<Path> find(BiPredicate<Path, BasicFileAttributes> pathPredicate)
```
gregw: you may as well pass in the depth rather than hard code to 300.  This makes it very clear the difference with resolveAll, which is essentially a find with depth 0

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathCollection.java
```diff
@@ -0,0 +1,291 @@
...
+                Stream<Path> stream = Files.find(path, 300, pathPredicate);
+                ret = Stream.concat(ret, stream);
+            }
+            catch (IOException e)
+            {
+                e.printStackTrace();
+            }
+        }
+        return ret;
+    }
```
gregw, line 276: We need to have a `resolve(URI other)` method, as that will be the primary path from a request to a resource.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/FileSystemPool.java
lorban: The easiest fix would be to use `ConcurrentMap.computeIfAbsent()` to create the entry. The `Pool` class is not a key-value store so it does fit the bill of this use-case.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/FileSystemPool.java
lorban: That can be solved by using `ConcurrentMap.compute()` to atomically decrement the counter and remove the entry from the map.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathCollection.java
lorban: I agree that `PathCollection` should be immutable, which prevents it from extending from `ArrayList` or implementing `java.util.Collection` in any form.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathCollection.java
lorban: Since this resolves from a collection of `Path`s to a single `Path` the `resolveFirst` prefix is quite accurate IMHO

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathCollection.java
lorban: Do we want to be able to resolve anything but `URI`s?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathComparator.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Comparator;
+
+/**
+ * Provider neutral {@link Path} comparator.
+ * <p>
+ * Sorting is done via the {@code Path.toUri()} objects.
+ * This is useful when sorting Path objects across FileSystem types.
+ * </p>
+ */
+public class PathComparator implements Comparator<Path>
```
lorban, line 26: Does that make sense to sort by `URI`? If so, I think this class should reflect in its name that it compares by `URI` and possibly it should be moved to the `PathCollators` class.

lorban (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/FileSystemPool.java
gregw: I'm not sure computeIfAbsent is good enough as it can end up creating an entry which is ultimately not used.  So in this case it would create a zip filesystem that loses the race to be added and then has to be disposed of.
I think a big fat lock is probably best here.  Not on critical path.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/FileSystemPool.java
gregw: In this case, I think compute could be used.... but only if the computeIfAbsent can be used above.  If we end up using a lock there, then this needs to be a lock here as well.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathCollection.java
joakime: What about read-only apis for Stream or Iterable ?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathCollection.java
joakime: I'm leery of using only URI as resolve against.  
As that's not the contract on `Path.resolve(*)` which is either String or Path, not URI.
Treating PathCollection.resolve() and Path.resolve() the same seems like it would just lower the number of bugs.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathComparator.java
joakime: This is a low odds scenario.

But you have a collection consisting of multiple Path roots that are all different java.nio.file.FileSystem types.   Doing the sort by URI keeps the rules the same across the variances of those FileSystems. 

Having this in the PathCollators makes sense too.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/FileSystemPool.java
lorban: I think you're misunderstanding `computeIfAbsent`: the lambda is only called when the entry is null _and_ you won the race. If the entry isn't null, it's simply returned as a `get` would. I think it's safe here, but I'm not against using a lock instead.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/FileSystemPool.java
lorban: Agreed: `compute` can be used if it's also used above, otherwise you must use the same lock.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathCollection.java
lorban: Do we need to be able to iterate the contained `Path`s, except for logging/monitoring purposes?

If you want to provide a powerful API to be able to select a specific `Path` using a complex query, you could eventually add a `stream()` method that would return a `Stream<Path>`.

gregw: That shouldn't be a problem.  If the interface is a `List<String>`, then an implementation can use an immutable list behind that.

gregw:
@joakime note that I've asked @lorban to work with this branch and you to get this merged and integrated into the ContextHandler/ResourceHandler etc.

This is now kind of on the critical path to a deployable jetty-12.   ie we need some kind of ResourceCollection replacement in ContextHandler before we can make ee10-webapp work and we need ee10-webapps before I can complete my work on the deployer to deploy different environments.

gregw:
... but I'll continue to comment ideas here....

@lorban Currently nothing is calling `Context.getResourceBase()`.   The `ResourceHandler` currently has it's own getter/setter for a resource base, but that was only really ever intended as an override and it should default to calling `Context.getResourceBase()` to initialize that.    Note that all requests now have Context and there is always a `Server.getContext()` available if you can't find a `ContextHandler` if you are looking for a context without a request.

Anyway,  it should be easy for you to change the signature so that `Context.getResourceBase()` returns either a `PathCollection` or a new class `ResourceBase extends PathCollection`.

Only slight complication you will get will be in ee9, where we still have the old API on the legacy `ContextHandler` that wraps a core `ContextHandler`.   Currently that has:
```java
    public void setBaseResource(Resource base)
    {
        try
        {
            _coreContextHandler.setResourceBase(base == null ? null : base.getFile().toPath());
            _baseResource = base;
        }
        catch (IOException e)
        {
            throw new IllegalArgumentException(e);
        }
    }
```

Which converts a `Resource` to a `Path`.  So that will need to be modified to be able to convert a `ResourceCollection` to a `PathCollection`.  Should be pretty trivial.

Hmmm I need to check what happens if ResourceCollection is mutated after it is set????  Need to check if we ever actually do that.... I hope not.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathCollection.java
joakime: I'm onboard with what @lorban is thinking.
I'm struggling to find a use case where iterating through the path entries has any purpose that isn't fulfilled by the API already.
Only one I can think of dump/debug.

gregw: The only use-case that I can think of is that if `PathCollection` is immutable, than treating it as a `List<Path>` then it can be passed to a builder method like `PathCollection.from(List<Path>)` or better yet streamed/filtered or otherwise manipulated to create the a new `PathCollection`.

But it doesn't need to be the full `List` interface... perhaps just `Iterable`? maybe a `Stream<Path> stream()` method?

Eitherway, my main concern is that it is easy to create a new PathCollection from an old one. Something like:
```java
    PathCollection pathsBC = PathCollection.of(pathB, pathC);
    PathCollection pathsABC = PathCollection.of(pathA, pathsBC);
    PathCollection pathsABCDE = PathCollection.of(pathsABC, pathD, pathE);
```

joakime: I can see that usecase easily enough.
The existing resource handling (`.close()`) would still be supported that way too.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathCollection.java
joakime: This is the replacement of `Resource.fromList()`

See: https://github.com/eclipse/jetty.project/blob/jetty-11.0.9/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java#L946-L1018

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathCollection.java
janbartel: If PathCollection is immutable, then I'll need a method like PathCollection.from(Path, PathCollection).  This would be used eg by the jetty maven plugin as it discovers each of the overlaid wars that need to form part of the resource base, or maybe by the MetaInfConfiguration that discovers fragments. Ideally the code would look like:

```
Path discovered;
PathCollection existing = webappContext.getResourceBase();
webAppContext.setResourceBase(PathCollection.from(discovered, existing));
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/paths/PathCollection.java
gregw: The whole point of the changes to URI handling in jetty-12 is that we now have an unambiguous URI encoded pathInContext.   We can't give that as a string to a Path.resolve or PathCollection.resolve, because it will have encodings in it like %20 and %2F.  We can't decode to a string, as that introduces the ambiguities again as we lose information (was the / a / or a %2F encoding?).

Hence we need to resolve via something equivalent to `baseResoure.toURI().resolve(new URI(pathInContext))`

lorban:
Superceded by https://github.com/eclipse/jetty.project/pull/7974

PR #7947, opened 2022-05-02
Improved PathSpec handling for servletName & pathInfo

### LINKED ISSUE #7891: Better Servlet PathMappings for Regex
**Jetty version(s)**
 >= 10

**Description**

The path mappings for RegexPathSpec does not use groups for prefix matches to set `servletPath` and `pathInfo`.

joakime:
Introduction of MatchedResource and MatchedPath to capture the results of a PathSpec match.

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/AbstractPathSpec.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/AbstractPathSpec.java
@@ -36,7 +36,12 @@ public abstract class AbstractPathSpec implements PathSpec
             return diff;
 
         // Path Spec Name (alphabetical)
-        return getDeclaration().compareTo(other.getDeclaration());
+        diff = getDeclaration().compareTo(other.getDeclaration());
+        if (diff != 0)
+            return diff;
+
+        // Path Implementation
+        return getClass().getName().compareTo(other.getClass().getName());
     }
 
     @Override
@@ -55,7 +60,7 @@ public abstract class AbstractPathSpec implements PathSpec
     @Override
     public final int hashCode()
     {
-        return Objects.hash(getDeclaration());
+        return Objects.hash(this.getClass().getSimpleName() + ":" + getDeclaration());
     }
 
     @Override
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/AbstractPathSpec.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/AbstractPathSpec.java
@@ -60,7 +60,7 @@ public abstract class AbstractPathSpec implements PathSpec
     @Override
     public final int hashCode()
     {
-        return Objects.hash(this.getClass().getSimpleName() + ":" + getDeclaration());
+        return Objects.hash(getGroup().ordinal(), getSpecLength(), getDeclaration(), getClass().getName());
     }
 
     @Override
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedPath.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedPath.java
@@ -37,7 +37,7 @@ public interface MatchedPath
         @Override
         public String toString()
         {
-            return "MatchedPath.EMPTY";
+            return MatchedPath.class.getSimpleName() + ".EMPTY";
         }
     };
 
@@ -60,7 +60,7 @@ public interface MatchedPath
             @Override
             public String toString()
             {
-                return "MatchedPath.from[pathMatch=" + pathMatch + ", pathInfo=" + pathInfo + "]";
+                return MatchedPath.class.getSimpleName() + "[pathMatch=" + pathMatch + ", pathInfo=" + pathInfo + "]";
             }
         };
     }
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedPath.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedPath.java
@@ -1,0 +1,41 @@
+//
+//  ========================================================================
+//  Copyright (c) 1995-2022 Mort Bay Consulting Pty Ltd and others.
+//  ------------------------------------------------------------------------
+//  All rights reserved. This program and the accompanying materials
+//  are made available under the terms of the Eclipse Public License v1.0
+//  and Apache License v2.0 which accompanies this distribution.
+//
+//      The Eclipse Public License is available at
+//      http://www.eclipse.org/legal/epl-v10.html
+//
+//      The Apache License v2.0 is available at
+//      http://www.opensource.org/licenses/apache2.0.php
+//
+//  You may elect to redistribute this code under either of these licenses.
+//  ========================================================================
+//
+
+package org.eclipse.jetty.http.pathmap;
+
+public interface MatchedPath
+{
+    /**
+     * @return the raw input path as provided.
+     */
+    String getPath();
+
+    /**
+     * Return the portion of the path that matches a path spec.
+     *
+     * @return the path name portion of the match.
+     */
+    String getPathMatch();
+
+    /**
+     * Return the portion of the path that is after the path spec.
+     *
+     * @return the path info portion of the match, or null if there is no portion after the {@link #getPathMatch()}
+     */
+    String getPathInfo();
+}
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedPath.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedPath.java
@@ -33,7 +33,37 @@ public interface MatchedPath
         {
             return null;
         }
+
+        @Override
+        public String toString()
+        {
+            return "MatchedPath.EMPTY";
+        }
     };
+
+    static MatchedPath from(String pathMatch, String pathInfo)
+    {
+        return new MatchedPath()
+        {
+            @Override
+            public String getPathMatch()
+            {
+                return pathMatch;
+            }
+
+            @Override
+            public String getPathInfo()
+            {
+                return pathInfo;
+            }
+
+            @Override
+            public String toString()
+            {
+                return "MatchedPath.from[pathMatch=" + pathMatch + ", pathInfo=" + pathInfo + "]";
+            }
+        };
+    }
 
     /**
      * Return the portion of the path that matches a path spec.
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedResource.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedResource.java
@@ -20,6 +20,12 @@ package org.eclipse.jetty.http.pathmap;
 
 import java.util.Map;
 
+/**
+ * The match details when using {@link PathMappings#getMatched(String)}, used to minimize return to the PathSpec or PathMappings for subsequent details
+ * that are now provided by the {@link MatchedPath} instance.
+ *
+ * @param <E> the type of resource (IncludeExclude uses boolean, WebSocket uses endpoint/endpoint config, servlet uses ServletHolder, etc)
+ */
 public class MatchedResource<E>
 {
     private final E resource;
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedResource.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedResource.java
@@ -18,34 +18,34 @@
 
 package org.eclipse.jetty.http.pathmap;
 
-public interface MatchedPath
+public class MatchedResource<E>
 {
-    MatchedPath EMPTY = new MatchedPath()
+    private final MappedResource<E> mappedResource;
+    private final MatchedPath matchedPath;
+
+    public MatchedResource(MappedResource<E> resource, MatchedPath matchedPath)
     {
-        @Override
-        public String getPathMatch()
-        {
-            return null;
-        }
+        this.mappedResource = resource;
+        this.matchedPath = matchedPath;
+    }
 
-        @Override
-        public String getPathInfo()
-        {
-            return null;
-        }
-    };
+    public MappedResource<E> getMappedResource()
+    {
+        return this.mappedResource;
+    }
 
-    /**
-     * Return the portion of the path that matches a path spec.
-     *
-     * @return the path name portion of the match.
-     */
-    String getPathMatch();
+    public PathSpec getPathSpec()
+    {
+        return mappedResource.getPathSpec();
+    }
 
-    /**
-     * Return the portion of the path that is after the path spec.
-     *
-     * @return the path info portion of the match, or null if there is no portion after the {@link #getPathMatch()}
-     */
-    String getPathInfo();
+    public E getResource()
+    {
+        return mappedResource.getResource();
+    }
+
+    public MatchedPath getMatchedPath()
+    {
+        return matchedPath;
+    }
 }
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedResource.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedResource.java
@@ -18,34 +18,53 @@
 
 package org.eclipse.jetty.http.pathmap;
 
+import java.util.Map;
+
 public class MatchedResource<E>
 {
-    private final MappedResource<E> mappedResource;
+    private final E resource;
+    private final PathSpec pathSpec;
     private final MatchedPath matchedPath;
 
-    public MatchedResource(MappedResource<E> resource, MatchedPath matchedPath)
+    public MatchedResource(E resource, PathSpec pathSpec, MatchedPath matchedPath)
     {
-        this.mappedResource = resource;
+        this.resource = resource;
+        this.pathSpec = pathSpec;
         this.matchedPath = matchedPath;
     }
 
-    public MappedResource<E> getMappedResource()
+    public static <E> MatchedResource<E> of(Map.Entry<PathSpec, E> mapping, MatchedPath matchedPath)
     {
-        return this.mappedResource;
+        return new MatchedResource<>(mapping.getValue(), mapping.getKey(), matchedPath);
     }
 
     public PathSpec getPathSpec()
     {
-        return mappedResource.getPathSpec();
+        return this.pathSpec;
     }
 
     public E getResource()
     {
-        return mappedResource.getResource();
+        return this.resource;
     }
 
-    public MatchedPath getMatchedPath()
+    /**
+     * Return the portion of the path that matches a path spec.
+     *
+     * @return the path name portion of the match.
+     */
+    public String getPathMatch()
     {
-        return matchedPath;
+        return matchedPath.getPathMatch();
+    }
+
+    /**
+     * Return the portion of the path that is after the path spec.
+     *
+     * @return the path info portion of the match, or null if there is no portion after the {@link #getPathMatch()}
+     */
+    public String getPathInfo()
+    {
+        return matchedPath.getPathInfo();
     }
 }
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
@@ -147,11 +147,17 @@ public class PathMappings<E> implements Iterable<MappedResource<E>>, Dumpable
         MatchedPath matchedPath;
         PathSpecGroup lastGroup = null;
 
+        boolean skipRestOfGroup = false;
         // Search all the mappings
         for (MappedResource<E> mr : _mappings)
         {
-            boolean skipMatch = false;
             PathSpecGroup group = mr.getPathSpec().getGroup();
+            if (group == lastGroup && skipRestOfGroup)
+            {
+                continue; // skip
+            }
+
+            // Run servlet spec optimizations on first hit of specific groups
             if (group != lastGroup)
             {
                 // New group in list, so let's look for an optimization
@@ -178,7 +184,7 @@ public class PathMappings<E> implements Iterable<MappedResource<E>>, Dumpable
                                 i--;
                             }
                             // If we reached here, there's NO optimized EXACT Match possible, skip simple match below
-                            skipMatch = true;
+                            skipRestOfGroup = true;
                         }
                         break;
                     }
@@ -201,7 +207,7 @@ public class PathMappings<E> implements Iterable<MappedResource<E>>, Dumpable
                                 i--;
                             }
                             // If we reached here, there's NO optimized PREFIX Match possible, skip simple match below
-                            skipMatch = true;
+                            skipRestOfGroup = true;
                         }
                         break;
                     }
@@ -223,7 +229,7 @@ public class PathMappings<E> implements Iterable<MappedResource<E>>, Dumpable
                                     return new MatchedResource<>(candidate.getResource(), candidate.getPathSpec(), matchedPath);
                             }
                             // If we reached here, there's NO optimized SUFFIX Match possible, skip simple match below
-                            skipMatch = true;
+                            skipRestOfGroup = true;
                         }
                         break;
                     }
@@ -232,13 +238,9 @@ public class PathMappings<E> implements Iterable<MappedResource<E>>, Dumpable
                 }
             }
 
-            // Only perform simple match if optimized match lets us
-            if (!skipMatch)
-            {
-                matchedPath = mr.getPathSpec().matched(path);
-                if (matchedPath != null)
-                    return new MatchedResource<>(mr.getResource(), mr.getPathSpec(), matchedPath);
-            }
+            matchedPath = mr.getPathSpec().matched(path);
+            if (matchedPath != null)
+                return new MatchedResource<>(mr.getResource(), mr.getPathSpec(), matchedPath);
 
             lastGroup = group;
         }
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
@@ -160,6 +160,9 @@ public class PathMappings<E> implements Iterable<MappedResource<E>>, Dumpable
             // Run servlet spec optimizations on first hit of specific groups
             if (group != lastGroup)
             {
+                // New group, reset skip logic
+                skipRestOfGroup = false;
+
                 // New group in list, so let's look for an optimization
                 switch (group)
                 {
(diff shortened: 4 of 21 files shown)


jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedPath.java
```diff
@@ -0,0 +1,51 @@
...
+//      The Apache License v2.0 is available at
+//      http://www.opensource.org/licenses/apache2.0.php
+//
+//  You may elect to redistribute this code under either of these licenses.
+//  ========================================================================
+//
+
+package org.eclipse.jetty.http.pathmap;
+
+public interface MatchedPath
```
gregw, line 21: I think it would be more intuitive to have a `getPathSpec()` method on this class, but not have it on `MatchedResource`

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/ServletPathSpec.java
```diff
@@ -270,6 +284,44 @@ private boolean isWildcardMatch(String path)
...
+                }
+                break;
+            case SUFFIX_GLOB:
+                if (path.regionMatches((path.length() - _specLength) + 1, _declaration, 1, _specLength - 1))
+                    return new ServletMatchedPath(path, null);
+                break;
+            case ROOT:
+                // Only "/" matches
+                if ("/".equals(path))
+                    return new ServletMatchedPath("", path); // TODO: review this
```
gregw: bizarrely I think that is correct!  "" matches "/" and returns "" as the match and "/" as the part after the match !?!?!?!?

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/ServletPathSpec.java
```diff
@@ -270,6 +284,44 @@ private boolean isWildcardMatch(String path)
...
+                break;
+            case DEFAULT:
+                // If we reached this point, then everything matches
+                return new ServletMatchedPath(path, null);
```
gregw, line 320: this should also be a return of a known `MatchedPath`:
```suggestion
                return _matchedPath;
```

gregw: So this will be the main cost of this change (in garbage).  Previously we could just return path here, but now will need to wrap it in a `ServletMatchedPath`.    Probably worth it if we can reduced string comparisons elsewhere.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/ServletPathSpec.java
```diff
@@ -291,4 +343,28 @@ public boolean matches(String path)
                 return false;
         }
     }
+
+    public static class ServletMatchedPath implements MatchedPath
```
gregw: There is nothing "Servlet" about this.   Just add a static method on `MatchedPath` so you can say `MatchedPath.from(pathMatch, pathInfo)`

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/ServletPathSpec.java
```diff
@@ -291,4 +343,28 @@ public boolean matches(String path)
...
+    {
+        private final String servletName;
+        private final String pathInfo;
+
+        public ServletMatchedPath(String servletName, String pathInfo)
+        {
+            this.servletName = servletName;
```
gregw: this isn't the servletName, it is the pathMatch!

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -182,15 +191,25 @@ public MappedResource<E> getMatch(String path)
                 }
             }
 
-            if (mr.getPathSpec().matches(path))
-                return mr;
+            matchedPath = mr.getPathSpec().matched(path);
```
gregw, line 244: I think we have a big inefficiency here.... even in the existing code.

Consider if we have 3 EXACT mappings, 3 PREFIX and the DEFAULT mappings, then we will:
1. iterate over this loop 7 times.
2. twice we will fire the `group != lastGroup` test and do an efficient Trie lookup
3. once we fire the `group != lastGroup` and do nothing
4. So all 7 iterations come to this line and call `matched(path)`, even though we know for 6 of them that they will not match because they failed in the Trie

We need to change this loop so that if an EXACT or PREFIX match fails in the Trie, then all iterations for the same group are skipped and we don't try the explicit match here for them all.   

This change will more than make up for the small amount of extra garbage

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/ServletPathSpec.java
joakime: Done.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/ServletPathSpec.java
joakime: Done, using MatchedPath.from() instead.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
joakime: I think the original loop without the Tries is more efficient (in terms of speed and memory).

I'll work with @lorban and mock up some jmh tests to see what is truely better.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/ServletPathSpec.java
joakime: Now using `MatchedPath.from("", path);`

joakime: The logic here is identical to how it was in 9.4.46 too.

See: 

* https://github.com/eclipse/jetty.project/blob/jetty-9.4.46.v20220331/jetty-http/src/test/java/org/eclipse/jetty/http/pathmap/ServletPathSpecTest.java#L122
* https://github.com/eclipse/jetty.project/blob/jetty-9.4.46.v20220331/jetty-http/src/test/java/org/eclipse/jetty/http/pathmap/ServletPathSpecTest.java#L152-L153

https://github.com/eclipse/jetty.project/blob/bc17a0369a11ecf40bb92c839b9ef0a8ac50ea18/jetty-http/src/test/java/org/eclipse/jetty/http/pathmap/ServletPathSpecTest.java#L122

https://github.com/eclipse/jetty.project/blob/bc17a0369a11ecf40bb92c839b9ef0a8ac50ea18/jetty-http/src/test/java/org/eclipse/jetty/http/pathmap/ServletPathSpecTest.java#L152-L153

It surprised me too, and we have many tests elsewhere in jetty that test for ROOT path hits.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedPath.java
joakime: That would mean MappedResource doesn't really need to exist too, as we can do a simple resource return `<E>` now.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
joakime: @gregw  I think we've addressed your concern on this block of code.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedPath.java
joakime: This change caused too many changes.
Be happy to make it in Jetty 12 though.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -231,24 +295,48 @@ public boolean put(PathSpec pathSpec, E resource)
...
+                    while (exact != null && !_exactMap.put(exact, entry))
+                    {
+                        _exactMap = new ArrayTernaryTrie<>((ArrayTernaryTrie<MappedResource<E>>)_exactMap, 1.5);
+                    }
+                }
+                else
                 {
-                    _exactMap = new ArrayTernaryTrie<>((ArrayTernaryTrie<MappedResource<E>>)_exactMap, 1.5);
+                    // This is not a Servlet mapping, turn off optimization on Exact
+                    _optimizedExact = false;
```
gregw: I feel that we probably can do optimized exact matches, even for regex etc.   The Trie maps to a MappedResource that includes the PathSpec, so if match produces a non ServletPathSpec match, then we know we have to run it again to ensure we match any groups etc.

But maybe this is OK for now.  Perhaps add a TODO to investigate if an exact map can be done.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -231,24 +295,48 @@ public boolean put(PathSpec pathSpec, E resource)
         switch (pathSpec.getGroup())
         {
             case EXACT:
-                String exact = pathSpec.getPrefix();
-                while (exact != null && !_exactMap.put(exact, entry))
+                if (pathSpec instanceof ServletPathSpec)
+                {
+                    String exact = pathSpec.getDeclaration();
```
gregw: I know it is not part of this PR, but can you put a comment in the while below that says what the loop is for (i.e. it is growing the capacity of the trie).  Took me a few minutes to work that out.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -231,24 +295,48 @@ public boolean put(PathSpec pathSpec, E resource)
...
+                    while (prefix != null && !_prefixMap.put(prefix, entry))
+                    {
+                        _prefixMap = new ArrayTernaryTrie<>((ArrayTernaryTrie<MappedResource<E>>)_prefixMap, 1.5);
+                    }
+                }
+                else
                 {
-                    _prefixMap = new ArrayTernaryTrie<>((ArrayTernaryTrie<MappedResource<E>>)_prefixMap, 1.5);
+                    // This is not a Servlet mapping, turn off optimization on Prefix
+                    _optimizedPrefix = false;
```
gregw: Again, we should be able to extract the prefix from any PREFIX_GLOB and put that in the Trie.
So again, for now add a TODO to optimize later.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -231,24 +295,48 @@ public boolean put(PathSpec pathSpec, E resource)
...
+                    String suffix = pathSpec.getSuffix();
+                    while (suffix != null && !_suffixMap.put(suffix, entry))
+                    {
+                        _suffixMap = new ArrayTernaryTrie<>((ArrayTernaryTrie<MappedResource<E>>)_prefixMap, 1.5);
+                    }
+                }
+                else
+                {
+                    // This is not a Servlet mapping, turn off optimization on Suffix
+                    _optimizedSuffix = false;
```
gregw: Hmmm I really don't like all these optimized booleans.  Might just be better to null out the map/trie and have a null mean don't look up.  That is a single test rather than always testing boolean then looking at the map/trie

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -263,21 +351,31 @@ public boolean put(PathSpec pathSpec, E resource)
...
         switch (pathSpec.getGroup())
         {
             case EXACT:
-                if (prefix != null)
-                    _exactMap.remove(prefix);
+                String exact = pathSpec.getDeclaration();
+                if (exact != null)
+                {
+                    _exactMap.remove(exact);
+                    // TODO: recalculate _optimizeExact
```
gregw: Do this before PR can be accepted

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -231,24 +295,48 @@ public boolean put(PathSpec pathSpec, E resource)
         switch (pathSpec.getGroup())
```
gregw: There is an existing bug with this switch.   It should do the `boolean added = _mappings.add(entry);` first and run the switch only if `added==true`

gregw (changes requested):
A few things from a quick look through.
I don't love that we are not optimizing if there are Regex patterns, but I think that can be a TODO for later.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/ServletPathSpec.java
```diff
@@ -270,6 +284,44 @@ private boolean isWildcardMatch(String path)
...
     }
 
+    @Override
+    public MatchedPath matched(String path)
+    {
+        switch (_group)
+        {
+            case EXACT:
+                if (_declaration.equals(path))
+                    return MatchedPath.from(path, null);
```
gregw: This should be a final field as it will always be the immutable `MatchedPath.from(_declaration, null)`

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/ServletPathSpec.java
```diff
@@ -270,6 +284,44 @@ private boolean isWildcardMatch(String path)
...
+                if (isWildcardMatch(path))
+                {
+                    String pathMatch = path;
+                    String pathInfo = null;
+                    if (path.length() != (_specLength - 2))
+                    {
+                        pathMatch = path.substring(0, _specLength - 2);
+                        pathInfo = path.substring(_specLength - 2);
+                    }
+                    return MatchedPath.from(pathMatch, pathInfo);
```
gregw: This also needs to avoid allocation in the common case of "/foo" matching "/foo/*"

```suggestion
                if (isWildcardMatch(path))
                {
                    if (path.length() == (_specLength - 2))
                        return finalFieldAllocatedInConstructor;
                    String pathMatch = anotherFinalFiledAllocatedInConstructor ;
                    String pathInfo = path.substring(_specLength - 2);
                    return MatchedPath.from(pathMatch, pathInfo);
```

gregw (dismissed):

gregw:
@joakime do you want me to have a go at fixing some of these issues?

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
joakime: Lets punt on optimized regex for now.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/ServletPathSpec.java
joakime: final where?  if you have multiple PathSpecs, you have multiple final fields, which means a lookup/map again.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/ServletPathSpec.java
joakime: You'll have a map of values, mapping each PathSpec and potential path.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
joakime: Added some comments about the concerns on exact/prefix/suffix and non-servlet path specs

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
joakime: Done.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
joakime: It's not quite as easy as you think.
The comments added show what needs to be resolved.
TODO added as well.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
joakime: That means creating the Trie anew with every .add() or .put()

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
joakime: Done.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
joakime: Fixed in .put() and .remove()

joakime:
I addressed most of the review comments.

The comments about final fields for various things don't make much sense to me, as we are dealing with multiple PathSpec objects in the collection, so having a final field make sense will result in another Map lookup, which kinda defeats the whole final field benefit.

If you can make it work without a new collection/trie/etc, give it a whirl.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/ServletPathSpec.java
gregw: The pathSpec itself should know it's prefix, so that is what is returned.  No map needed.

gregw:
I'll do a PR against this PR addressing some of my performance concerns.... good task for on the train/plane tomorrow.

joakime:
@gregw :triangular_flag_on_post: bump :triangular_flag_on_post: : this need attention for the next release.

gregw:
Ack

gregw:
@joakime I pushed a change to this branch that should improve performance.
You'll need to get somebody else to review now.

joakime:
+1 from me.
Now lets get @janbartel @lachlan-roberts @sbordet or @lorban to review it (and hopefully approve it)

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/AbstractPathSpec.java
```diff
@@ -55,7 +60,7 @@ public final boolean equals(Object obj)
     @Override
     public final int hashCode()
     {
-        return Objects.hash(getDeclaration());
+        return Objects.hash(this.getClass().getSimpleName() + ":" + getDeclaration());
```
sbordet: Implementation of `hashCode()` should be based on the same items used for `equals()`.
Consider:

```suggestion
        return Objects.hash(getGroup().ordinal(), getSpecLength(), getDeclaration(), getClass().getName());
```

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedPath.java
```diff
@@ -0,0 +1,81 @@
...
+        @Override
+        public String getPathInfo()
+        {
+            return null;
+        }
+
+        @Override
+        public String toString()
+        {
+            return "MatchedPath.EMPTY";
```
sbordet: Better to use `getClass().getSimpleName()`.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedPath.java
```diff
@@ -0,0 +1,81 @@
...
+            @Override
+            public String getPathInfo()
+            {
+                return pathInfo;
+            }
+
+            @Override
+            public String toString()
+            {
+                return "MatchedPath.from[pathMatch=" + pathMatch + ", pathInfo=" + pathInfo + "]";
```
sbordet: Better to use `getClass().getSimpleName()`.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedResource.java
```diff
@@ -0,0 +1,70 @@
...
+//
+//  You may elect to redistribute this code under either of these licenses.
+//  ========================================================================
+//
+
+package org.eclipse.jetty.http.pathmap;
+
+import java.util.Map;
+
+public class MatchedResource<E>
```
sbordet, line 29: Javadocs? Why it is generic on `E` and what types `E` can be?

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/MatchedPath.java
```diff
@@ -0,0 +1,81 @@
...
+//      The Apache License v2.0 is available at
+//      http://www.opensource.org/licenses/apache2.0.php
+//
+//  You may elect to redistribute this code under either of these licenses.
+//  ========================================================================
+//
+
+package org.eclipse.jetty.http.pathmap;
+
+public interface MatchedPath
```
sbordet, line 21: Javadocs?

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -292,8 +378,50 @@ public boolean remove(PathSpec pathSpec)
...
+            {
+                case EXACT:
+                    String exact = pathSpec.getDeclaration();
+                    if (exact != null)
+                    {
+                        _exactMap.remove(exact);
+                        // Recalculate _optimizeExact
+                        _optimizedExact = _mappings.stream()
+                            .filter((mapping) -> mapping.getPathSpec().getGroup() == PathSpecGroup.EXACT)
+                            .allMatch((mapping) -> mapping.getPathSpec() instanceof ServletPathSpec);
```
sbordet: Consider moving this logic into a method and reused it in the other `case` blocks.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathSpec.java
```diff
@@ -25,6 +25,17 @@
  */
 public interface PathSpec extends Comparable<PathSpec>
 {
+    static PathSpec from(String pathSpecString)
+    {
+        if (pathSpecString == null)
+            throw new RuntimeException("Path Spec String must start with '^', '/', or '*.': got [" + pathSpecString + "]");
```
sbordet: Should not this be a NPE? 
Furthermore, if it's non-null, there is no check that `pathSpecString` starts with ^, /, etc. The only check below is for ^ to distinguish between regex or not.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
```diff
@@ -181,16 +264,88 @@ public Pattern getPattern()
...
         {
-            // match only non-query part
-            return getMatcher(path.substring(0, idx)).matches();
+            return new RegexMatchedPath(this, path, matcher);
         }
-        else
+        return null;
+    }
+
+    public class RegexMatchedPath implements MatchedPath
```
sbordet: Can be private?

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
```diff
@@ -181,16 +264,88 @@ public Pattern getPattern()
...
+            }
+
+            // default is null
+            return null;
+        }
+
+        @Override
+        public String toString()
+        {
+            return "RegexMatchedPath[" +
```
sbordet: Use class.simpleName.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/ServletPathSpec.java
```diff
@@ -90,13 +96,15 @@ public ServletPathSpec(String servletPathSpec)
             group = PathSpecGroup.PREFIX_GLOB;
             prefix = servletPathSpec.substring(0, specLength - 2);
             suffix = null;
+            _preMatchedPath = MatchedPath.from(prefix, null);
```
sbordet: Store in a local variable and assign below.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/UriTemplatePathSpec.java
```diff
@@ -421,4 +440,62 @@ public String[] getVariables()
     {
         return _variables;
     }
+
+    public static class UriTemplateMatchedPath implements MatchedPath
```
sbordet: Make private.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/UriTemplatePathSpec.java
```diff
@@ -421,4 +440,62 @@ public String[] getVariables()
...
+                else
+                    return pathInfo;
+            }
+            return null;
+        }
+
+        @Override
+        public String toString()
+        {
+            return "UriTemplateMatchedPath[" +
```
sbordet: Use class.simpleName.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
```diff
@@ -38,34 +59,80 @@ public RegexPathSpec(String regex)
...
+                    break;
+                case ']':
+                    inTextList = false;
                     signature.append('g'); // glob
                     break;
                 case '/':
-                    if (!inGrouping)
+                    if (!inTextList && !inQuantifier)
                         pathDepth++;
                     break;
```
sbordet, line 102: Should be there a `case` for `<` and `>` (for named groups)?

jetty-http/src/test/java/org/eclipse/jetty/http/pathmap/RegexPathSpecTest.java
```diff
@@ -151,6 +152,42 @@ public void testSuffixSpec()
...
+        assertEquals(PathSpecGroup.SUFFIX_GLOB, spec.getGroup(), "Spec.group");
+
+        assertMatches(spec, "/a/middle/c.do");
+        assertMatches(spec, "/a/b/c/d/middle/e/f");
+        assertMatches(spec, "/middle/");
+
+        assertNotMatches(spec, "/a.do");
+        assertNotMatches(spec, "/a/middle");
+        assertNotMatches(spec, "/middle");
+    }
```
sbordet, line 190: Is there a test for named groups?

sbordet (changes requested):

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
joakime: Not needed, the group detects a signature `g` properly and that sets up the signature check later approriately.
Other uses of `<` and `>` are allowed.
New test case added.

jetty-http/src/test/java/org/eclipse/jetty/http/pathmap/RegexPathSpecTest.java
joakime: Added testcase

sbordet (approved):

Issue #1446, opened 2017-04-04
100% CPU in selector on Windows


amshf:
There is a Java web-app that is running by Jetty 9.4.3.v20170317
After some minutes/hours of working, the web-app starts to use 100% CPU without any specific reason. **Even when it is idle**.

After profiling and getting thread-dumps, it seems that Jetty is the source of high CPU consumption.

So what is wrong with Jetty? is this a bug? **_(Please pay attention to the second image)_**

Below I list some evidences after profiling (using VisualVM) when CPU consumption reaches to 100%:
--------

- Running threads (**CLICK to enlarge images**): 
![threads](https://cloud.githubusercontent.com/assets/11075737/24646571/bab0dec8-1931-11e7-95dd-980ee504398e.PNG)

- CPU Sampler: 
![cpu-samples-cpusamples-after100cpu](https://cloud.githubusercontent.com/assets/11075737/24646603/d4d40fd2-1931-11e7-9427-03bb76c4ed75.PNG)

- CPU Sampler-per thread: 
![cpu-samples-perthread-after100cpu](https://cloud.githubusercontent.com/assets/11075737/24646636/eec6da8c-1931-11e7-8d49-02746cc5b477.PNG)

- Memory Sampler: 
![mem-samples-perthread-after100cpu](https://cloud.githubusercontent.com/assets/11075737/24646668/f9b0f3ec-1931-11e7-9478-435c62f8cc6d.PNG)

- Process status (using processExplorer): 
![process](https://cloud.githubusercontent.com/assets/11075737/24646689/0675d962-1932-11e7-8dc1-f9b9cbdaa2d8.PNG)

- Process performance (using processExplorer): 
![process-performance2](https://cloud.githubusercontent.com/assets/11075737/24646712/15d2adcc-1932-11e7-8b5f-57c0e478f7cd.PNG)

joakime:
What does the thread dump for that thread look like over time? (Does its stacktrace change?)
What version of Java? (Are you running the latest JVM?)
What OS? (there's been sporadic reports of NIO issues on Windows, esp with various home firewall software or network optimization software installed on dev/test machines)

joakime:
More questions...
What are the clients connecting to your server? (be specific)
What is the timeout configuration on your server?
What is your server attempting to do?

amshf:
Let me describe the test scenarios first:
I've tested the app on:
-two OS,
-on Eclipse (by Jetty) OR run it by Jetty itself,
[...]

PR #7950, opened 2022-05-03
#4717 Solve high CPU spikes happening on Mac with JDK 1.8

### LINKED ISSUE #4717: High CPU spikes with jetty winstone threads
OS: | Mac OS X, 10.14.6 , x86_64/64 (4 cores)
Java: | Java(TM) SE Runtime Environment, 1.8.0_241-b07
JVM: | Java HotSpot(TM) 64-Bit Server VM, 25.241-b07

**Description**
We are upgraded Jenkins v2.187 to 2.219 and this issue started happening, we tried with Jenkins 2.225 but issue remains.

our servers in production reported 100% CPU usage. The instance had been running for several days. The CPU spike occurred instantly and remained until the instance was rebooted.

**Hundred's of Jetty (winstone) threads are filing up not getting closed with the method sun.nio.ch.FileDispatcherImpl.preClose0(Native Method)**

**Jetty (winstone)-66537**
sun.nio.ch.FileDispatcherImpl.preClose0(Native Method)
sun.nio.ch.SocketDispatcher.preClose(SocketDispatcher.java:59)
sun.nio.ch.SocketChannelImpl.implCloseSelectableChannel(SocketChannelImpl.java:838)
java.nio.channels.spi.AbstractSelectableChannel.implCloseChannel(AbstractSelectableChannel.java:234)
java.nio.channels.spi.AbstractInterruptibleChannel.close(AbstractInterruptibleChannel.java:115)
org.eclipse.jetty.io.ChannelEndPoint.doClose(ChannelEndPoint.java:197)
org.eclipse.jetty.io.AbstractEndPoint.doOnClose(AbstractEndPoint.java:220)
org.eclipse.jetty.io.AbstractEndPoint.close(AbstractEndPoint.java:192)
org.eclipse.jetty.io.AbstractEndPoint.close(AbstractEndPoint.java:175)
org.eclipse.jetty.io.ssl.SslConnection$DecryptedEndPoint.doClose(SslConnection.java:1366)
org.eclipse.jetty.io.AbstractEndPoint.doOnClose(AbstractEndPoint.java:220)
org.eclipse.jetty.io.AbstractEndPoint.close(AbstractEndPoint.java:192)
org.eclipse.jetty.io.AbstractEndPoint.close(AbstractEndPoint.java:175)
org.eclipse.jetty.io.AbstractConnection.close(AbstractConnection.java:254)
org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:258)
org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:311)
org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:103)
org.eclipse.jetty.io.ssl.SslConnection$DecryptedEndPoint.onFillable(SslConnection.java:543)
org.eclipse.jetty.io.ssl.SslConnection.onFillable(SslConnection.java:398)
org.eclipse.jetty.io.ssl.SslConnection$2.succeeded(SslConnection.java:161)
org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:103)
org.eclipse.jetty.io.ChannelEndPoint$2.run(ChannelEndPoint.java:117)
org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.runTask(EatWhatYouKill.java:336)
org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.doProduce(EatWhatYouKill.java:313)
org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.tryProduce(EatWhatYouKill.java:171)
org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.run(EatWhatYouKill.java:129)
org.eclipse.jetty.util.thread.ReservedThreadExecutor$ReservedThread.run(ReservedThreadExecutor.java:375)
org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:806)
org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:938)

**When normal state we are seeing below for jetty winstone thread looks like**

Jetty (winstone)-14
sun.misc.Unsafe.park(Native Method)
java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:215)
java.util.concurrent.SynchronousQueue$TransferStack.awaitFulfill(SynchronousQueue.java:460)
java.util.concurrent.SynchronousQueue$TransferStack.transfer(SynchronousQueue.java:362)
java.util.concurrent.SynchronousQueue.poll(SynchronousQueue.java:941)
org.eclipse.jetty.util.thread.ReservedThreadExecutor$ReservedThread.reservedWait(ReservedThreadExecutor.java:304)
org.eclipse.jetty.util.thread.ReservedThreadExecutor$ReservedThread.run(ReservedThreadExecutor.java:366)
org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:806)
org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:938)
java.lang.Thread.run(Thread.java:748)

lorban:
Explicitly cancel the selection key before closing the channel to work around what seems to be a bug in MacOS / JDK 1.8.

### DIFF
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/ChannelEndPoint.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/ChannelEndPoint.java
@@ -207,8 +207,7 @@ public abstract class ChannelEndPoint extends AbstractEndPoint implements Manage
             LOG.debug("doClose {}", this);
         try
         {
-            if (_key != null)
-                _key.cancel();
+            _key.cancel();
             _channel.close();
         }
         catch (IOException e)
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/ChannelEndPoint.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/ChannelEndPoint.java
@@ -207,6 +207,8 @@ public abstract class ChannelEndPoint extends AbstractEndPoint implements Manage
             LOG.debug("doClose {}", this);
         try
         {
+            if (_key != null)
+                _key.cancel();
             _channel.close();
         }
         catch (IOException e)
--- a/jetty-io/src/test/java/org/eclipse/jetty/io/SocketChannelEndPointOpenCloseTest.java
+++ b/jetty-io/src/test/java/org/eclipse/jetty/io/SocketChannelEndPointOpenCloseTest.java
@@ -19,6 +19,9 @@
 package org.eclipse.jetty.io;
 
 import java.nio.ByteBuffer;
+import java.nio.channels.SelectableChannel;
+import java.nio.channels.SelectionKey;
+import java.nio.channels.Selector;
 import java.nio.channels.ServerSocketChannel;
 import java.nio.channels.SocketChannel;
 
@@ -58,9 +61,51 @@ public class SocketChannelEndPointOpenCloseTest
     private EndPointPair newConnection() throws Exception
     {
         EndPointPair c = new EndPointPair();
+        SelectionKey k = new SelectionKey()
+        {
+            @Override
+            public SelectableChannel channel()
+            {
+                throw new UnsupportedOperationException();
+            }
 
-        c.client = new SocketChannelEndPoint(SocketChannel.open(connector.socket().getLocalSocketAddress()), null, null, null);
-        c.server = new SocketChannelEndPoint(connector.accept(), null, null, null);
+            @Override
+            public Selector selector()
+            {
+                throw new UnsupportedOperationException();
+            }
+
+            @Override
+            public boolean isValid()
+            {
+                throw new UnsupportedOperationException();
+            }
+
+            @Override
+            public void cancel()
+            {
+            }
+
+            @Override
+            public int interestOps()
+            {
+                throw new UnsupportedOperationException();
+            }
+
+            @Override
+            public SelectionKey interestOps(int ops)
+            {
+                throw new UnsupportedOperationException();
+            }
+
+            @Override
+            public int readyOps()
+            {
+                throw new UnsupportedOperationException();
+            }
+        };
+        c.client = new SocketChannelEndPoint(SocketChannel.open(connector.socket().getLocalSocketAddress()), null, k, null);
+        c.server = new SocketChannelEndPoint(connector.accept(), null, k, null);
         return c;
     }
 


joakime:
Good stuff.

Do we need this same logic anywhere else in our code?

jetty-io/src/main/java/org/eclipse/jetty/io/ChannelEndPoint.java
```diff
@@ -207,6 +207,8 @@ public void doClose()
             LOG.debug("doClose {}", this);
         try
         {
+            if (_key != null)
```
sbordet: `_key` is never `null`, so the check seems redundant, and is now also racy with `replaceKey()`.
I would remove the null check and perhaps `catch (Throwable)` below.

sbordet (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/ChannelEndPoint.java
lorban: `SocketChannelEndPointOpenCloseTest` fails without this check, but it passes a null `SelectionKey` to the `ChannelEndPoint` constructor.

I thought adding that null check was the safest thing to do, but you're right, I overlooked the `replaceKey()` mechanism which makes this null check odd.

I've changed the test so that it passes a non-null `SelectionKey` but I wonder if we shouldn't be stricter in the constructor about accepting null arguments when that should not be allowed.

lorban:
@joakime I checked that together with @sbordet and no, there aren't any other places where that should be done as `ChannelEndPoint` is the central place where that is abstracted away.

I also do not think we should merge this change to other branches as the problem only occurs on MacOS with JDK 1.8.

jetty-io/src/test/java/org/eclipse/jetty/io/SocketChannelEndPointOpenCloseTest.java
```diff
@@ -58,9 +61,51 @@ public static void close() throws Exception
...
+            }
+
+            @Override
+            public int readyOps()
+            {
+                throw new UnsupportedOperationException();
+            }
+        };
+        c.client = new SocketChannelEndPoint(SocketChannel.open(connector.socket().getLocalSocketAddress()), null, k, null);
+        c.server = new SocketChannelEndPoint(connector.accept(), null, k, null);
```
sbordet, line 108: Oh, no! 😞 
Aren't unit/mock/whatever tests great? Not!

Would it be possible to rewrite this test to something less hacky that actually tests real code?

sbordet (changes requested):

jetty-io/src/test/java/org/eclipse/jetty/io/SocketChannelEndPointOpenCloseTest.java
lorban: I'm not sure of what you mean. This test is checking the open/shutdown state of `SocketChannelEndPoint`, which happened to not need the selection key but now does. Why would we want to change the scope of this test because we changed an implementation detail?

joakime:
> I also do not think we should merge this change to other branches as the problem only occurs on MacOS with JDK 1.8.

I wonder if the Windows users at #1446 and #2205 could benefit from this as well?

jetty-io/src/test/java/org/eclipse/jetty/io/SocketChannelEndPointOpenCloseTest.java
sbordet: I mean that creating "fake" SocketChannelEndPoints that lack the otherwise required ManagedSelector, the SelectionKey, etc. is basically just testing test code, not the real code path that will happen with real code.

It would be better to actually test with a normally connected pair of EndPoints, using ServerConnector and ClientConnector, so we would not need to "fake" the SelectionKey and half-create SocketChannelEndPoints without required components.

lorban:
#2205 looks strangely similar, so it might benefit from this change too. 

I'm still a bit on the fence about merging this fix or not. In my understanding this is clearly a JVM bug, but the fact that it's been observed on two different OS is making me question this conclusion a bit. But the problem has only ever been reported on JDK 1.8 and one person explicitly stated that it cannot be reproduced on newer JDKs so I'm still slightly in favor of not merging the fix forward.

@joakime WDYT?

joakime (approved):
I'm happy that this fix even exists.

And my reading of it is that even this extra key.cancel() is ultimately harmless on newer JVMs.
So why not keep it in for all JVMs in Jetty 9.4.x?

I also see the argument for not bothering merging this forward to Jetty 10+ as none of those have the ability to even run on a buggy JVM.

sbordet (approved):

Issue #7958, opened 2022-05-05
EOL (End of Life) for Jetty 9 - January 2025


joakime:
It’s been 9 years since Jetty 9 was released and it is time to announce some changes regarding the end of community support for Eclipse Jetty 9.x.

June 1st, 2022 will mark the official _End of Community Support_ for Eclipse Jetty 9.x.

While users of Jetty 9.x will continue to see releases for security and critical issues until 9.x goes _End of Life_ (EOL), the only ongoing support provided will be for Webtide customers.

Official EOL is January 2025.

Action | Before June 1st, 2022 | End of Community Support (June 1st, 2022) | End of Life
--- | --- | --- | ---
Community PRs reviewed and integrated | :heavy_check_mark: | :x: | :x: 
Webtide Customer PRs reviewed and integrated | :heavy_check_mark: | :heavy_check_mark: | :x:
Community triggered releases | :heavy_check_mark: | :x: | :x:
Webtide Customer triggered releases | :heavy_check_mark: | :heavy_check_mark:  | :x:
Security / Vulnerability triggered releases | :heavy_check_mark: | :heavy_check_mark: | :x:

Thank you for your continued usage and support of Jetty, please take a moment to look at Jetty release 12.0.x.

Note: As of January 2024, both Jetty 10 and Jetty 11 are also at End of Service/Support Life

* #10485

Jetty Version | Minimum JDK Version | Jakarta EE | Servlet Version | Namespace
---- | ---- | ---- | ---- | ----
12.0.x - env:`ee8` | 17 | EE8 | 4 | `javax.servlet.*`
12.0.x - env:`ee9` | 17 | EE9 | 5 | `jakarta.servlet.*`
12.0.x - env:`ee10` | 17 | EE10 | 6 | `jakarta.servlet.*`

jmcc0nn3ll:
Let's push it to June 1st, first of the month is always good for this sort of announcement and 10 days might panic someone.  No need to cause a heart attack.

joakime:
I agree. Updating ...

joakime:
We should send an email to the jetty mailing lists - https://www.eclipse.org/jetty/mailing_lists.php

* https://accounts.eclipse.org/mailing-list/jetty-users
* https://accounts.eclipse.org/mailing-list/jetty-dev
* https://accounts.eclipse.org/mailing-list/jetty-announce

And we should update the version table at https://www.eclipse.org/jetty/download.php

[...]

PR #7961, opened 2022-05-06
Add codeql-analysis.yml GitHub Actions Workflow

JLLeitschuh:
Hello! Your friendly neighborhood security researcher here!

GitHub is slowly (not entirely there yet) migrating away from LGTM.com
to establishing a way for security researchers to work with CodeQL databases
fully through GitHub.com. This 'new' way requires that databases get built
via GitHub actions and uploaded to GitHub.

The old way, supported by the `.lgtm.yaml` file will continue to exist for a while
(and I'll continue to use it), but as a researcher, if the Jetty project also built a
CodeQL database and uploaded it, I'd find it incredibly useful.

In summary, this PR adds a GitHub action to build a CodeQL database and runs
GitHub's CodeQL analysis tool which audits the code for security vulnerabilities.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,87 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ 'jetty-10.[1-9]?[0-9].x', 'jetty-11.[1-9]?[0-9].x', 'jetty-12.[1-9]?[0-9].x' ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ 'jetty-10.[1-9]?[0-9].x', 'jetty-11.[1-9]?[0-9].x', 'jetty-12.[1-9]?[0-9].x' ]
+  schedule:
+    - cron: '22 1 * * 2'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'java', 'javascript' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
+        # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      # Install and setup JDK 11
+      - name: Setup JDK 11
+        uses: actions/setup-java@v3
+        if: ${{
+          startsWith(github.ref, 'refs/heads/jetty-10.') ||
+          startsWith(github.ref, 'refs/heads/jetty-11.') ||
+          startsWith(github.base_ref, 'jetty-10.') ||
+          startsWith(github.base_ref, 'jetty-11.')
+          }}
+        with:
+          distribution: temurin
+          java-version: 11
+          cache: maven
+
+      # Install and setup JDK 17
+      - name: Setup JDK 17
+        uses: actions/setup-java@v3
+        if: ${{
+          startsWith(github.ref, 'refs/heads/jetty-12.') ||
+          startsWith(github.base_ref, 'jetty-12.')
+          }}
+        with:
+          distribution: temurin
+          java-version: 17
+          cache: maven
+
+      # Initializes the CodeQL tools for scanning.
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v2
+        with:
+          languages: ${{ matrix.language }}
+          # If you wish to specify custom queries, you can do so here or in a config file.
+          # By default, queries listed here will override any specified in a config file.
+          # Prefix the list here with "+" to use these queries and those in the config file.
+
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,69 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ 'jetty-[1-9]?[0-9].[0-9].x', 'feat/JLL/codeql-support' ]
```
joakime: Just hardcode `jetty-10.0.x`, `jetty-11.0.x`, and `jetty-12.0.x`
Note: Jetty 9.x is being deprecated soon. See: #7958

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,69 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      # Install and setup JDK 11
+      - name: Setup JDK 11
+        uses: actions/setup-java@v3
+        with:
+          distribution: temurin
+          java-version: 11
```
joakime, line 43: Jetty 12 will requires Java 17.

.github/workflows/codeql-analysis.yml
JLLeitschuh: If you support branching names like `9.4.x`, and will use that pattern for future development, wouldn't it be this?
`[ 'jetty-10.[1-9]?[0-9].x', 'jetty-11.[1-9]?[0-9].x', 'jetty-12.[1-9]?[0-9].x' ]`

.github/workflows/codeql-analysis.yml
JLLeitschuh: Thanks fixed! 🙂

JLLeitschuh:
Once this is merged, the maintainers of the eclipse jetty project will be able to see the alerts here.
https://github.com/eclipse/jetty.project/security/code-scanning

Also, merging this will establish a new baseline, and any PRs that unintentionally introduce a new security vulnerability will be flagged 😄

JLLeitschuh:
I think this PR is in a complete state. The one check 'Code scanning results / CodeQL' will not pass this time, but for other PRs it will pass because it will compare against what's on the default branch as a baseline.

joakime:
This is interesting, but we'll need some time to absorb what this means to our processes.

JLLeitschuh:
If you don't want to run this job on PRs in general, we can always remove these lines:

https://github.com/eclipse/jetty.project/pull/7961/files#diff-63bd641104d10e25f141d518a16b22a151d125e12701df2f9e79734b23b90188R6-R8

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,87 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'java', 'javascript' ]
```
olamy, line 24: we do not have any javascript here. So there is no real need of a matrix

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,87 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ 'jetty-10.[1-9]?[0-9].x', 'jetty-11.[1-9]?[0-9].x', 'jetty-12.[1-9]?[0-9].x' ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ 'jetty-10.[1-9]?[0-9].x', 'jetty-11.[1-9]?[0-9].x', 'jetty-12.[1-9]?[0-9].x' ]
```
olamy, line 8: not sure we want to run this on PRs.
just a weekly on hardcoded branches might be enough.

.github/workflows/codeql-analysis.yml
joakime: We do have javascript, in the demo apps. Check the bottom of the "files changed" tab in this PR to see where it's finding bugs.

.github/workflows/codeql-analysis.yml
joakime: I'm ok with PRs, it will catch accidentally introduced bugs.

.github/workflows/codeql-analysis.yml
olamy: ah yup demos 🤦

olamy:
> This is interesting, but we'll need some time to absorb what this means to our processes.

correct because we already some reports we do nothing about.
so here we will have to mark the reported errors or they will be sitting here and we will do nothing and this will be YASRWDC (Yet Another Static Report We Don't Care :) )

JLLeitschuh:
> so here we will have to mark the reported errors or they will be sitting here and we will do nothing and this will be YASRWDC (Yet Another Static Report We Don't Care :) )

Not a bad thing, but yes. I'm currently discussing with the GitHub team if there's a way to build the database without reporting query results. They say that there is, but it's a bit of a hack.

joakime (approved):

olamy:
is it possible to exclude `src/test` from the scanning?

JLLeitschuh:
It is, but I don't remember how to off hand. I suggest consulting the documentation

PR #7962, opened 2022-05-06
Add jetty maven plugin ITs as part of code coverage as we do it :)

olamy:
- enable of jacoco coverage for maven its plugin
- do not skip jacoco for maven plugins we want to record we test them
- fix excludes
- fix exclusion
- no need to exclude this

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,20 +14,24 @@ pipeline {
               timeout( time: 180, unit: 'MINUTES' ) {
                 mavenBuild( "jdk17", "clean install -Perrorprone", "maven3")
                 // Collect up the jacoco execution results (only on main build)
-                jacoco inclusionPattern: '**/org/eclipse/jetty/**/*.class',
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -17,8 +17,8 @@ pipeline {
                 jacoco inclusionPattern: '**/org/eclipse/jetty/**/*.class',
                        exclusionPattern: '' +
                                // build tools
-                               '**/org/eclipse/jetty/ant/**' + ',**/org/eclipse/jetty/maven/**' +
-                               ',**/org/eclipse/jetty/jspc/**' +
+                               '**/org/eclipse/jetty/ant/**,*/org/eclipse/jetty/maven/its/**' +
+                               ',**/org/eclipse/jetty/its/**' +
```
joakime, line 22: Please fix formatting here.
Each pattern should be its own line please.

olamy: fixed

janbartel (approved):

PR #7963, opened 2022-05-08
add a configuration field to MavenServerConnector to be able to force reUsePort

olamy:

### DIFF
--- a/jetty-maven-plugin/src/it/it-parent-pom/pom.xml
+++ b/jetty-maven-plugin/src/it/it-parent-pom/pom.xml
@@ -129,6 +129,11 @@
           <groupId>org.eclipse.jetty</groupId>
           <artifactId>jetty-maven-plugin</artifactId>
           <version>${jetty.version}</version>
+          <configuration>
+            <mavenServerConnector>
+              <reusePort>true</reusePort>
+            </mavenServerConnector>
+          </configuration>
         </plugin>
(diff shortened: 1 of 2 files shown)


janbartel:
@olamy what's the motivation for the change? I guess I'm ok with it, but I just want to know why you've added it?

joakime:
> @olamy what's the motivation for the change? I guess I'm ok with it, but I just want to know why you've added it?

I think his work on the jakarta tck with arquillian is the motivation.

olamy:
> > @olamy what's the motivation for the change? I guess I'm ok with it, but I just want to know why you've added it?
> 
> I think his work on the jakarta tck with arquillian is the motivation.

 I remember some issues with dependabot build few weeks ago (sorry but I cannot find it) and a build failing because of the famous  already in use message

janbartel:
> 

OK, but it's not the jetty builds that are failing right? We should always be choosing random ports.

janbartel:
@olamy nudge? This was needed for the tck, not for the jetty builds, correct?

olamy:
nope. TCK is not using maven plugin but Arquillian 
the problems are due to some IT tests (distro forked if I remember correctly) are using hardcoded port such 8080.
and sometimes those tests fail to start as a previous one was using the port already.

janbartel (approved):

PR #7965, opened 2022-05-09
revert back to bundle plugin 5.1.4 the bug makes the need of using clean for every build

olamy:
https://github.com/apache/felix-dev/pull/150


joakime (approved):
This needs to happen in 9.4.x too then?

olamy:
@joakime 9.4.x is using an old version 3.5.1 https://github.com/eclipse/jetty.project/blob/08d9ca0e32dfc840296def0b413b9446d044abbf/pom.xml#L104

PR #7969, opened 2022-05-09
restore jetty-home

olamy:
- get jetty-home back
- fix modules declaration

### DIFF
--- a/jetty-home/pom.xml
+++ b/jetty-home/pom.xml
@@ -218,7 +218,7 @@
             </goals>
             <configuration>
               <includeGroupIds>org.eclipse.jetty.toolchain,org.eclipse.jetty.ee9.websocket,org.eclipse.jetty.ee10.websocket</includeGroupIds>
-              <excludeArtifactIds>jetty-jakarta-servet-api,jetty-schemas,jetty-test-helper</excludeArtifactIds>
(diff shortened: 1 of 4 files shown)


gregw (commented):
We don't need to do PRs on jetty-12 just yet... at least not for simple things like this.

HOWEVER, please don't re-enable jetty-home until it can actually build something (and not break the build).

olamy:
> We don't need to do PRs on jetty-12 just yet... at least not for simple things like this.
> 
> HOWEVER, please don't re-enable jetty-home until it can actually build something (and not break the build).

@gregw it's green and assemble the jars the PR is to be sure it's green
see original branch (except flaky test_ https://jenkins.webtide.net/job/jetty.project/job/jetty-12.0.x-jetty-home/

jetty-home/pom.xml
```diff
@@ -236,6 +236,56 @@
...
+                  <classifier>sources</classifier>
+                </artifactItem>
+                <artifactItem>
+                  <groupId>org.eclipse.jetty.toolchain</groupId>
+                  <artifactId>jetty-jakarta-servlet-api</artifactId>
+                  <!-- FIXME MUST BE A PROPERTY -->
+                  <version>5.0.2</version>
+                  <classifier>sources</classifier>
+                </artifactItem>
+              </artifactItems>
```
gregw, line 285: Oh cool, you can bring in two different versions.

gregw (approved):
Let's give this a go

PR #7974, opened 2022-05-10
Jetty 12.0.x pathcollection

lorban:
There is still a million things to do, for instance:
 - making the DefaultServlet capable of serving something other than just files on disk
 - cleaning up the Resource API, removing almost all methods and only leaving a getPath() that can be used with the Files API
 - adding the multiple Path implementations to get rid of most Resource impls
 - figuring out if both PathResource and PathCollectionResource should be kept or only the latter.
 - make caching configurable
 - fix CachingPrecompressed* tests
 - ...

but the current state is to reach a point where the basics are set: keeping the Resource API, but much much simplified and mostly (only?) backed by Path/PathCollection.


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -36,7 +36,9 @@
  * The first resource in the collection is the main resource.
  * If a resource is not found in the main resource, it looks it up in
  * the order the resources were constructed.
+ * @deprecated
```
gregw: If we are a removing this in favour of `PathCollectionResource`, then we should say so here.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathCollectionResource.java
```diff
@@ -0,0 +1,374 @@
...
+
+import org.eclipse.jetty.util.StringUtil;
+import org.eclipse.jetty.util.URIUtil;
+import org.eclipse.jetty.util.paths.PathCollection;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * PathCollection Resource.
+ */
```
gregw: Can you explain somewhere (maybe here in this javadoc) why a PathCollectionResource is better than a ResourceCollection?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathCollectionResource.java
```diff
@@ -0,0 +1,374 @@
...
+        // resource, to preserve aliasing.
+        if (URIUtil.canonicalPath(subPath) == null)
+            throw new MalformedURLException(subPath);
+
+        if ("/".equals(subPath))
+            return this;
+
+        String toResolve = subPath.startsWith("/") ? subPath.substring(1) : subPath;
+        Path resolvedPath = pathCollection.resolveFirst(toResolve, (x) -> true);
+        return resolvedPath == null ? null : new PathResource(resolvedPath);
```
gregw: This still doesn't appear to handle the case that the resource returned is a collection (of some kind).  By using `PathCollection#resolveFirst` it is assuming a Path result.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathCollectionResource.java
```diff
@@ -0,0 +1,374 @@
...
+        catch (IOException e)
+        {
+            if (LOG.isDebugEnabled())
+                LOG.debug("ignored", e);
+        }
+        return false;
+    }
+
+    @Override
+    public Resource addPath(String subPath) throws IOException
```
gregw: I think the major work we need to do is to resolve what subPath is?   Is it a path as in a `Path` or is it a path from a `URI` or `HttpURI`?

This implementation is assuming the former, but what I really think we need is something that only supports the later.   Our resources must operate in the URI space and adding a path must be a URI operation not a Path operation.

gregw (changes requested):
@lorban I'm not convinced we need `PathCollectionResource`, specially as it currently doesn't solved the resolution to another collection.

More importantly, I think we have to do our operations in `URI` space, not in `Path` space.  Ie doing a `resource.addPath("foo\\bar")` should not be file system dependent.   That should give the same result as `Resource.newResource(resource.getURI() + '/' +"foo\\bar")`

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathCollectionResource.java
lorban: Agreed, the `String subPath` argument is a `URI` segment. `addPath` and `getResource` need to be merged, and the implementation cleaned up and fixed to properly take that into account.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathCollectionResource.java
lorban: Yes, this code should call `pathCollection.resolveAll()`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathCollectionResource.java
lorban: `PathCollection` contains the logic to lifecycle the `Path`'s `FileSystem`s. This can possibly be moved to `ResourceCollection` but downcasting to `PathResource` would then be needed.

lorban:
postponing PathCollection work, so this PR has been superseded by this other one: https://github.com/eclipse/jetty.project/pull/8090.

PR #7976, opened 2022-05-11
Add TRANSFER_ENCODING violation for MultiPart RFC7578 parser.

lachlan-roberts:
see https://datatracker.ietf.org/doc/html/rfc7578#section-4.7

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormInputStream.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormInputStream.java
@@ -638,7 +638,10 @@ public class MultiPartFormInputStream
             // Transfer encoding is not longer considers as it is deprecated as per
             // https://tools.ietf.org/html/rfc7578#section-4.7
             if (key.equalsIgnoreCase("content-transfer-encoding"))
-                _nonComplianceWarnings.add(NonCompliance.TRANSFER_ENCODING);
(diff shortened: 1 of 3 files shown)


jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormInputStream.java
```diff
@@ -610,7 +637,8 @@ else if (key.equalsIgnoreCase("content-type"))
 
             // Transfer encoding is not longer considers as it is deprecated as per
             // https://tools.ietf.org/html/rfc7578#section-4.7
-
+            if (key.equalsIgnoreCase("content-transfer-encoding"))
```
joakime, line 640: Do we want to be more subtle about this?

There are 3 `Content-Transfer-Encoding` scenarios that are not a violation (as they basically mean nothing more than "identity" or "no encoding")

* `Content-Transfer-Encoding: 8bit` (common enough in older httpclients)
* `Content-Transfer-Encoding: binary` (also common enough, Apache HttpClient will use this for example)
* `Content-Transfer-Encoding: 7bit`  (this one is iffy)

So that any other value is a violation

joakime (commented):
Question about the scope of the violation, to eliminate harmless conditions leaving harmful ones to only trigger.

joakime (approved):

PR #7983, opened 2022-05-12
Issue #1771 - add module for secure redirect

### LINKED ISSUE #1771: Add module for SecuredRedirect support
There doesn't seem to be a module that will inject/add the org.eclipse.jetty.server.handler.SecuredRedirectHandler
at the beginning of the server handler collection.

Does it make sense to have a module for this?

lachlan-roberts:
closes #1771

### DIFF
--- a/jetty-server/src/main/config/etc/jetty-secure-redirect.xml
+++ b/jetty-server/src/main/config/etc/jetty-secure-redirect.xml
@@ -2,20 +2,10 @@
 <!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "https://www.eclipse.org/jetty/configure_10_0.dtd">
 
 <Configure id="Server" class="org.eclipse.jetty.server.Server">
-  <Get id="Handler" name="handler"/>
-  <Call name="setHandler">
+  <Call name="insertHandler">
     <Arg>
-      <New class="org.eclipse.jetty.server.handler.HandlerList">
-        <Call name="addHandler">
-          <Arg>
-            <New id="SecuredRedirectHandler" class="org.eclipse.jetty.server.handler.SecuredRedirectHandler">
-              <Arg type="int"><Property name="jetty.redirect.code" default="302"/></Arg>
-            </New>
-          </Arg>
-        </Call>
-        <Call name="addHandler">
-          <Arg><Ref refid="Handler"/></Arg>
-        </Call>
+      <New id="SecuredRedirectHandler" class="org.eclipse.jetty.server.handler.SecuredRedirectHandler">
+        <Arg type="int"><Property name="jetty.secureredirect.code" default="302"/></Arg>
       </New>
     </Arg>
   </Call>
--- a/jetty-server/src/main/config/etc/jetty-secure-redirect.xml
+++ b/jetty-server/src/main/config/etc/jetty-secure-redirect.xml
@@ -1,0 +1,22 @@
+<?xml version="1.0"?>
+<!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "https://www.eclipse.org/jetty/configure_10_0.dtd">
+
+<Configure id="Server" class="org.eclipse.jetty.server.Server">
+  <Get id="Handler" name="handler"/>
+  <Call name="setHandler">
+    <Arg>
+      <New class="org.eclipse.jetty.server.handler.HandlerList">
+        <Call name="addHandler">
+          <Arg>
+            <New id="SecuredRedirectHandler" class="org.eclipse.jetty.server.handler.SecuredRedirectHandler">
+              <Arg type="int"><Property name="jetty.redirect.code" default="302"/></Arg>
+            </New>
+          </Arg>
+        </Call>
+        <Call name="addHandler">
+          <Arg><Ref refid="Handler"/></Arg>
+        </Call>
+      </New>
+    </Arg>
+  </Call>
+</Configure>
--- a/jetty-server/src/main/config/etc/jetty.xml
+++ b/jetty-server/src/main/config/etc/jetty.xml
@@ -60,7 +60,7 @@
     <!-- =========================================================== -->
     <New id="httpConfig" class="org.eclipse.jetty.server.HttpConfiguration">
       <Set name="secureScheme" property="jetty.httpConfig.secureScheme"/>
-      <Set name="securePort" property="jetty.httpConfig.securePort"/>
+      <Set name="securePort"><Property name="jetty.httpConfig.securePort" default="8443"/></Set>
       <Set name="outputBufferSize" property="jetty.httpConfig.outputBufferSize"/>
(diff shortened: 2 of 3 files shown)


jetty-server/src/main/config/etc/jetty-secure-redirect.xml
```diff
@@ -0,0 +1,22 @@
...
+
+<Configure id="Server" class="org.eclipse.jetty.server.Server">
+  <Get id="Handler" name="handler"/>
+  <Call name="setHandler">
+    <Arg>
+      <New class="org.eclipse.jetty.server.handler.HandlerList">
+        <Call name="addHandler">
+          <Arg>
+            <New id="SecuredRedirectHandler" class="org.eclipse.jetty.server.handler.SecuredRedirectHandler">
+              <Arg type="int"><Property name="jetty.redirect.code" default="302"/></Arg>
```
sbordet: `301` (a permanent redirect) is probably the best default here.

I also wonder if the property should have a more specific name because if we are going to write another, different, redirect module, `jetty.redirect.code` could clash.
Perhaps `jetty.redirect.secure.status.code`?

sbordet (dismissed):

jetty-server/src/main/config/etc/jetty-secure-redirect.xml
joakime: how about `jetty.secure.redirect.*` ?

jetty-server/src/main/config/etc/jetty.xml
```diff
@@ -60,7 +60,7 @@
     <!-- =========================================================== -->
     <New id="httpConfig" class="org.eclipse.jetty.server.HttpConfiguration">
       <Set name="secureScheme" property="jetty.httpConfig.secureScheme"/>
-      <Set name="securePort" property="jetty.httpConfig.securePort"/>
+      <Set name="securePort"><Property name="jetty.httpConfig.securePort" default="8443"/></Set>
```
joakime: I'm leery of this change.  
Isn't the in-code default `443`?

jetty-server/src/main/config/etc/jetty-secure-redirect.xml
joakime: The default is 302 (Moved Temporarily) in code.
If we change it here in the XML it should be changed in `SecureRedirectHandler` too.

joakime: From a Google perspective the response codes work like this ...

Permanent Redirect - show the new redirect target in search results.

* 301 (moved permanently)
* 308 (moved permanently)

Temporary Redirect - show the source page in search results

* 302 (found)
* 303 (see other) 
* 307 (temporary redirect)

sbordet: > how about jetty.secure.redirect.* ?

No, I prefer to be `jetty.<feature/module>.<other>`. `secure` is related to redirect (like status code), so it should be after `redirect`.

joakime: The class is called `SecureRedirect` it is specific for going `http` to `https` only.
It could have been called "ensure secure" or `ConfidentialTransportGuaranteeHandler` as well.
The "redirect" in that is just a hint on _how_ it does it.  That word "redirect" should be nowhere near the properties or module name.

No different than a servlet constraint enforcing a confidential transport-guarantee. (this is the exact words in the servlet spec, it also has no mention of "redirect")

jetty-server/src/main/config/etc/jetty.xml
lachlan-roberts: Well the default is already 8443 in `server.ini`. The problem is that `<Set name="securePort" property="jetty.httpConfig.securePort"/>` will not actually set the property if it is not explicitly defined in the ini file.

In the `HttpConfiguration` class if securePort is not set is is by default a value of 0, and if the securePort value is 0 in `SecureRedirectHandler` it gives a 403 response saying `HttpConfiguration.securePort not configured`.

So to get the secure-redirect module working you would have to uncomment this manually in `server.ini`:
```
## Port to use to build URIs for secure redirects
# jetty.httpConfig.securePort=8443
```

jetty-server/src/main/config/etc/jetty-secure-redirect.xml
lachlan-roberts: So what should I change the property name to, or is `jetty.redirect.code` ok?

I could do something like `jetty.secureredirect.code`.

jetty-server/src/main/config/etc/jetty-secure-redirect.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0"?>
+<!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "https://www.eclipse.org/jetty/configure_10_0.dtd">
+
+<Configure id="Server" class="org.eclipse.jetty.server.Server">
+  <Get id="Handler" name="handler"/>
+  <Call name="setHandler">
+    <Arg>
+      <New class="org.eclipse.jetty.server.handler.HandlerList">
```
gregw: `SecureRedirectHandler` is a handler wrapper, so you should not need this new HandlerList.
You can just do a `Server.insertHandler` to inject a secure redirect handler between the server and whatever its original handler was.

jetty-server/src/main/config/etc/jetty.xml
gregw: I agree that we should not put the default here, as that will change how we do defaults.

Instead we should modify the handler to be able to handle a 0 correctly, as I'm sure we already do in the SecurityHandler somewhere.  Hmmm no it also does a 403 if it sees 0 - so that definitely means we should not change the default.

Either they need to set the default port OR we can introduce some logic that 0 means redirect 80 to 443, but 403 response otherwise.

gregw (changes requested):

jetty-server/src/main/config/etc/jetty-secure-redirect.xml
sbordet: Don't like "secured" with the "d" at the end. Can we drop it from the class name and the property name?
I like `jetty.secureredirect.*`.

jetty-server/src/main/config/modules/secure-redirect.mod
```diff
@@ -0,0 +1,17 @@
+DO NOT EDIT - See: https://www.eclipse.org/jetty/documentation/current/startup-modules.html
+
+[description]
+Enable SecuredRedirectHandler to redirect all http requests to https.
```
gregw: Maybe add some note here that it redirects to the secure port set in the ???? module.

gregw (dismissed):
OK, but see note about doco

gregw (approved):

PR #7985, opened 2022-05-12
Rewrite of the content handling, both on client and on server.

sbordet:
Introduced a new io.Content class, with Content.Source, Content.Sink and Content.Chunk.

Updated the server-side to use io.Content instead of the previous server-side Content class.

Updated the client-side to use io.Content with a little specialization necessary for clients only.


gregw:
@sbordet I think some of those test failures are me.  I must have broken the build and then gone on vacation 2 days.... how much into the jar for that?
Fixing now and will tell you when you can merge

sbordet:
@gregw we found the issue and pushed a fix to `jetty-12.0.x`.

joakime:
> @gregw we found the issue and pushed a fix to `jetty-12.0.x`.

@sbordet looks like we broke the ee10 side with that fix (commit bfcd7845eb43f0c3d73b2e0f2369eff306cdb58b)

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -0,0 +1,403 @@
...
+import org.eclipse.jetty.util.Callback;
+import org.eclipse.jetty.util.FutureCallback;
+import org.eclipse.jetty.util.FuturePromise;
+import org.eclipse.jetty.util.IO;
+import org.eclipse.jetty.util.Promise;
+
+/**
+ * TODO: expand javadocs.
+ */
+public interface Content
```
gregw: Since Content is only a name scope, should it be a class with a private constructor_

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -0,0 +1,403 @@
...
+        //  read()!=null + demand() is legal usage
+        //  read()==null + demand() + demand callback + read()==null (AKA spurious wakeup) is possible
+        //  demand() + demand callback + read() is legal
+        //  demand() + demand() is illegal (no demand callback between demands AKA demand pending)
+        //  read()==isLast + demand() should invoke demand callback
+        //  read() with demand pending => not recommended (because we cannot tell due to races)
+        //  concurrent read() calls => not recommended (undefined, racy by definition)
+        //  read()==isLast, another read() will return isLast.
+        //  read()==Error, another read() will return the same error.
+        //  demandCallback.run() throws => fail() is called.
```
gregw: Javadoc doesn't need to be complete, but things likethis should be fixed

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -0,0 +1,403 @@
...
+        {
+            ByteBuffer byteBuffer = getByteBuffer();
+            length = Math.min(byteBuffer.remaining(), length);
+            byteBuffer.position(byteBuffer.position() + length);
+            return length;
+        }
+
+        // TODO: better name, because a last chunk not read yields terminal==false, but after it is read yields terminal==true
+        //  So maybe something like isLastAndConsumed()?
+        public default boolean isTerminal()
```
gregw, line 585: How about `isComplete`? That has a more transactional feel to it. 
Or `isContentComplete`?        Eitherway, pick a better name and remove the TODO.  If we kept TODOs for all the sub-optimal names, then the code would just be TODOs

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -0,0 +1,403 @@
...
+        {
+            return this instanceof Error || isLast() && !hasRemaining();
+        }
+
+        public static class Error implements Chunk
```
gregw: Strange that this is a nested class but `ByteBufferChunk` is not.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -0,0 +1,403 @@
...
+        public static Chunk from(ByteBuffer byteBuffer, boolean last)
+        {
+            return from(byteBuffer, last, null);
+        }
+
+        public static Chunk from(ByteBuffer byteBuffer, boolean last, Runnable releaser)
+        {
+            return new ByteBufferChunk(byteBuffer, last, releaser);
+        }
+
```
gregw, line 464:
```suggestion

￼
￼        public static Chunk from(Throwable error)
￼        {
￼            return new Error(error);
￼        }
```

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/internal/ByteBufferChunk.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public void release()
+    {
+        if (releaser != null)
+        {
+            Runnable runnable = releaser.getAndSet(null);
+            if (runnable != null)
+                runnable.run();
+        }
```
gregw, line 70: would it not be better for the releaser to be a `Consumer<ByteBuffer>`?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ContentOLD.java
```diff
@@ -45,7 +46,7 @@
  * the semantics of multiple methods like flush, close, onError, etc. to be
  * included in the read/write APIs.
  */
-public interface Content
+public interface ContentOLD
```
gregw: delete this

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -0,0 +1,403 @@
...
+     *         ByteBuffer buffer = chunk.getByteBuffer();
+     *         // ...
+     *
+     *         // Release the chunk when it has been consumed.
+     *         chunk.release();
+     *     }
+     * }
+     * }</pre>
+     */
+    public interface Source
```
gregw, line 122: Whilst I do like the symmetry of `Content.Source` and `Content.Sink` it does introduce a name only interface.
Whilst I like the name `Content` it is ultimately doing nothing for us here and violates least surprise when you can't do `content.getLength()` and instead have to ask the "source" of your content for the content length.  Then the "source" of the "content" doesn't provide content, it provides chunks.

Alternates are:

 + leave as is
 + move chunk API to top, so we have `Chunk.Source` and `Chunk.Sink`
 + make `Content` the source API so we just have `Content` and `Content.Sink`

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -0,0 +1,403 @@
...
+        }
+    }
+
+    public interface Sink
+    {
+        public static OutputStream asOutputStream(Sink sink)
+        {
+            return new ContentSinkOutputStream(sink);
+        }
+
```
gregw, line 369: Do we need an `Flow.Subscriber<Chunk> asSubscriber(Sink sink)` ?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -165,37 +167,38 @@ public interface Request extends Attributes, Content.Reader
...
+     * <li>a {@link Content.Chunk} instance, in case of normal request content</li>
      * </ul>
      * <p>When the returned {@code Content} is {@code null}, a call to
-     * {@link #demandContent(Runnable)} should be made, to be notified when more
+     * {@link #demand(Runnable)} should be made, to be notified when more
      * request content is available.</p>
      *
-     * @see #demandContent(Runnable)
+     * @see #demand(Runnable)
      */
```
gregw: should remove javadoc here and let interface fully define semantics

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -165,37 +167,38 @@ public interface Request extends Attributes, Content.Reader
...
      * <p>Demands to notify the given {@code Runnable} when request content is available to be read.</p>
-     * <p>It is not mandatory to call this method before a call to {@link #readContent()}.</p>
+     * <p>It is not mandatory to call this method before a call to {@link #read()}.</p>
      * <p>The given {@code Runnable} is notified only once for each invocation of this method,
      * and different invocations of this method may provide the same {@code Runnable}s or
      * different {@code Runnable}s.</p>
      *
-     * @see #readContent()
+     * @see #read()
      */
```
gregw: remove javadoc here

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -469,7 +461,7 @@ public void reset()
     // TODO test and document
     static OutputStream asOutputStream(Response response)
     {
-        return Content.asOutputStream(response);
+        return Content.Sink.asOutputStream(response);
     }
```
gregw: Do we need this convenience method?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -57,18 +57,10 @@ public interface Response extends Content.Writer
...
+    HttpFields.Mutable getOrCreateTrailers();
 
     @Override
-    void write(boolean last, Callback callback, ByteBuffer... content);
-
-    default void write(boolean last, Callback callback, String utf8Content)
-    {
-        write(last, callback, StandardCharsets.UTF_8.encode(utf8Content));
-    }
+    public void write(boolean last, Callback callback, ByteBuffer... buffers);
```
gregw: Do we need this empty override?  At very least do:
```suggestion
    void write(boolean last, Callback callback, ByteBuffer... buffers);
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -1073,6 +1079,22 @@ private HttpFields takeTrailers()
             return trailers;
         }
 
+        @Override
+        public void write(Content.Chunk chunk, Callback callback)
```
gregw: might need to define this method in the interface so we can add javadoc saying that Trailers are handled (and what steps have to be done to get them)

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -1073,6 +1079,22 @@ private HttpFields takeTrailers()
...
+        {
+            if (chunk instanceof Trailers trailersChunk)
+            {
+                ResponseHttpFields trailers = _request.getHttpChannel()._responseTrailers;
+                if (trailers != null)
+                    trailers.add(trailersChunk.getTrailers());
+                callback.succeeded();
```
gregw: Don't we still need to write here?
```suggestion
                Response.super.write(true, callback);
```

gregw (changes requested):
just some initial feedback

jetty-core/jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/internal/HttpStreamOverFCGI.java
```diff
@@ -187,14 +187,18 @@ private void notifyContentAvailable()
             onContentAvailable.run();
     }
 
-    public void onContent(Content content)
+    public void onContent(Content.Chunk chunk)
     {
-        _content = content;
```
lorban, line 192: Maybe you could add a paranoid check or an assertion here?
```java
assert _chunk == null;
```

jetty-core/jetty-fcgi/fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/internal/HttpStreamOverFCGI.java
```diff
@@ -187,14 +187,18 @@ private void notifyContentAvailable()
...
+        _chunk = chunk;
     }
 
     public void onComplete()
     {
-        _content = Content.last(_content);
+        if (_chunk == null)
+            _chunk = Content.Chunk.EOF;
+        // TODO: paranoid check, maybe remove it.
```
lorban: I quite like those paranoid checks, they catch bugs earlier and are cheap to compute. I'd keep it or at least turn it into an assertion.

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/util/ByteBufferRequestContent.java
```diff
@@ -13,77 +13,47 @@
...
     }
 
     @Override
-    protected Subscription newSubscription(Consumer consumer, boolean emitInitialContent)
-    {
-        return new SubscriptionImpl(consumer, emitInitialContent);
-    }
-
-    private class SubscriptionImpl extends AbstractSubscription
+    public boolean rewind()
```
lorban, line 55: This override is useless as it only calls super.

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/util/MultiPartRequestContent.java
```diff
@@ -54,7 +53,7 @@
  * &lt;/form&gt;
  * </pre>
  */
-public class MultiPartRequestContent extends AbstractRequestContent implements Closeable
+public class MultiPartRequestContent /*extends AbstractRequestContent*/ implements Request.Content, Closeable
```
lorban, line 56: This class looks unfinished. Did you leave it as-is on purpose?

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/util/PathRequestContent.java
```diff
@@ -66,107 +49,20 @@ public PathRequestContent(String contentType, Path filePath) throws IOException
...
-        return true;
-    }
-
-    public ByteBufferPool getByteBufferPool()
-    {
-        return bufferPool;
-    }
-
-    public void setByteBufferPool(ByteBufferPool byteBufferPool)
+    public boolean rewind()
```
lorban, line 64: This override is unneeded as it only calls super.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ContentProducer.java
```diff
@@ -104,16 +104,17 @@
      * The returned content is decoded by the interceptor set with {@link #setInterceptor(HttpInput.Interceptor)}
      * or left as-is if no intercept is set.
      * After this call, state can be either of UNREADY or IDLE.
+     *
      * @return the next content that can be read from or null if the implementation does not block
      * and has no available content.
      */
-    Content nextContent();
+    Content.Chunk nextContent();
```
lorban: This should be renamed `nextChunk()` or maybe just `next()`.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ContentProducer.java
```diff
@@ -89,11 +89,11 @@
     boolean hasContent();
```
lorban: Now that this class isn't dealing with `Content`s but `Chunk`s this method should be renamed. Maybe to `hasChunk()`?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: It's not only a name scope. It has `copy()` and possibly in future some other utility methods that take both `Source` and `Sink`.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ContentProducer.java
```diff
@@ -59,17 +59,17 @@
...
     /**
-     * Get the byte count produced by the underlying {@link org.eclipse.jetty.server.Content.Reader}.
+     * Get the byte count produced by the underlying {@link Content.Source}.
      *
      * This call is always non-blocking.
      * Doesn't change state.
-     * @return the byte count produced by the underlying {@link org.eclipse.jetty.server.Content.Reader}.
+     * @return the byte count produced by the underlying {@link Content.Source}.
      */
     long getRawContentArrived();
```
lorban: Now that this class isn't dealing with `Content`s but `Chunk`s this method should be renamed.  Maybe to `getRawBytesArrived()`?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ContentProducer.java
```diff
@@ -13,12 +13,12 @@
...
-import org.eclipse.jetty.server.Content;
+import org.eclipse.jetty.io.Content;
 import org.eclipse.jetty.util.component.Destroyable;
 import org.eclipse.jetty.util.thread.AutoLock;
 
 /**
- * ContentProducer is the bridge between {@link HttpInput} and {@link org.eclipse.jetty.server.Content.Reader}.
+ * ContentProducer is the bridge between {@link HttpInput} and {@link Content.Source}.
  */
 public interface ContentProducer
```
lorban, line 23: Now that this class isn't dealing with `Content`s but `Chunk`s this class should be renamed.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: Because `ByteBufferChunk` does not have extra APIs other than those of `Chunk`. `Chunk.Error` has `getCause()`.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: `isTerminal()` is as transactional as `isComplete()` to me, so I'd just remove the TODO?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ContentProducer.java
```diff
@@ -104,16 +104,17 @@
...
-    Content nextContent();
+    Content.Chunk nextContent();
 
     /**
-     * Free up the content by calling {@link Content#release()} on it
+     * Free up the content by calling {@link Content.Chunk#release()} on it
      * and updating this instance' internal state.
      */
-    void reclaim(Content content);
+    void reclaim(Content.Chunk content);
```
lorban: Now that this class isn't dealing with `Content`s but `Chunk`s the following method `onContentProducible` should be renamed.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HttpStream.java
```diff
@@ -19,20 +19,20 @@
...
      * @return an ID unique within the lifetime scope of the associated protocol connection.
      * This may be a protocol ID (eg HTTP/2 stream ID) or it may be unrelated to the protocol.
-     * @see HttpStream#getId();
      */
     String getId();
 
     long getNanoTimeStamp();
 
-    Content readContent();
+    Content.Chunk readContent();
```
lorban: Now that this class is dealing with `Chunk`s instead of `Content`s this method should be renamed.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/internal/ByteBufferChunk.java
sbordet: No, for tests we typically use `CountDownLatch::countDown`, and for other cases we don't use the chunk's ByteBuffer because we delegate the call to something else, e.g. a RetainableByteBuffer, etc.
I think Runnable is better.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HttpStream.java
```diff
@@ -19,20 +19,20 @@
...
 
     void demandContent(); // Calls back on Channel#onDataAvailable
```
lorban: Now that this class is dealing with `Chunk`s instead of `Content`s this method should be renamed. And the comment points to an invalid method name that should be fixed too. :-)

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ContentOLD.java
sbordet: Ah yes.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: See above, `Content` is not a namespace only interface.

lorban (changes requested):
First set of review comments, more will follow.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: done.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
sbordet: Done.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
sbordet: Done.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
```diff
@@ -177,16 +177,16 @@ public void checkMinDataRate()
...
 
     @Override
-    public long getRawContentArrived()
+    public long getRawBytesArrived()
     {
         assertLocked();
         if (LOG.isDebugEnabled())
-            LOG.debug("getRawContentArrived = {} {}", _rawContentArrived, this);
-        return _rawContentArrived;
+            LOG.debug("getRawContentArrived = {} {}", _rawBytesArrived, this);
```
lorban: The debug message's method name should be changed.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
```diff
@@ -207,38 +207,38 @@ public boolean consumeAll()
...
         // for instance the H2 implementation calls Stream.demand() that may
         // deliver the content asynchronously. Tests in StreamResetTest cover this.
-        boolean atEof = _servletChannel.failAllContent(x);
+        boolean atEof = _servletChannel.failAvailableContent(x);
         if (LOG.isDebugEnabled())
             LOG.debug("failed all content of http channel EOF={} {}", atEof, this);
         return atEof;
     }
 
     private void failCurrentContent(Throwable x)
```
lorban: This method is actually failing the current chunk.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
```diff
@@ -251,10 +251,10 @@ public boolean onContentProducible()
...
 
     @Override
-    public Content nextContent()
+    public Content.Chunk nextChunk()
     {
         assertLocked();
-        Content content = nextTransformedContent();
+        Content.Chunk content = nextTransformedContent();
         if (LOG.isDebugEnabled())
             LOG.debug("nextContent = {} {}", content, this);
```
lorban: The debug message's method name should be changed.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
```diff
@@ -306,52 +306,52 @@ boolean isUnready()
         return _servletChannel.getState().isInputUnready();
     }
 
-    private Content nextTransformedContent()
+    private Content.Chunk nextTransformedContent()
```
lorban: This method is actually returning the next transformed chunk.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
```diff
@@ -306,52 +306,52 @@ boolean isUnready()
...
     {
         if (LOG.isDebugEnabled())
             LOG.debug("nextTransformedContent {}", this);
```
lorban: The debug message's method name should be changed if the method gets renamed.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
```diff
@@ -476,20 +476,20 @@ else if (content != _rawContent && !_rawContent.isSpecial() && !_rawContent.isEm
                 _servletChannel.abort(failure);
             if (LOG.isDebugEnabled())
                 LOG.debug("interceptor threw exception {}", this, x);
-            return _transformedContent;
+            return _transformedChunk;
         }
     }
 
-    private Content produceRawContent()
+    private Content.Chunk produceRawContent()
```
lorban: The method actually produces a raw chunk.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
```diff
@@ -476,20 +476,20 @@ else if (content != _rawContent && !_rawContent.isSpecial() && !_rawContent.isEm
...
+        Content.Chunk content = _servletChannel.getRequest().read();
         if (content != null)
         {
-            _rawContentArrived += content.remaining();
+            _rawBytesArrived += content.remaining();
             if (_firstByteTimeStamp == Long.MIN_VALUE)
                 _firstByteTimeStamp = System.nanoTime();
             if (LOG.isDebugEnabled())
-                LOG.debug("produceRawContent updated rawContentArrived to {} and firstByteTimeStamp to {} {}", _rawContentArrived, _firstByteTimeStamp, this);
+                LOG.debug("produceRawContent updated rawContentArrived to {} and firstByteTimeStamp to {} {}", _rawBytesArrived, _firstByteTimeStamp, this);
```
lorban: The debug message's `rawContentArrived` should be renamed, as well as the method name should be changed if the method gets renamed.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
```diff
@@ -476,20 +476,20 @@ else if (content != _rawContent && !_rawContent.isSpecial() && !_rawContent.isEm
...
         }
         if (LOG.isDebugEnabled())
             LOG.debug("produceRawContent produced {} {}", content, this);
```
lorban: The debug message's method name should be changed if the method gets renamed.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/HttpInput.java
```diff
@@ -155,17 +155,17 @@ public long getContentReceived()
...
         }
     }
 
-    public boolean consumeAll()
+    public boolean consumeAvailable()
     {
         try (AutoLock lock = _contentProducer.lock())
         {
             if (LOG.isDebugEnabled())
                 LOG.debug("consumeAll {}", this);
```
lorban, line 167: The method name should be changed in the debug message.

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/AsyncContentProducer.java
```diff
@@ -33,13 +34,13 @@
 class AsyncContentProducer implements ContentProducer
```
lorban, line 34: The renames made in the ee10 module are also valid here. I suppose they weren't made on purpose?

jetty-core/jetty-http2/http2-server/src/main/java/org/eclipse/jetty/http2/server/internal/HttpStreamOverHTTP2.java
```diff
@@ -163,22 +164,9 @@ public Runnable onData(DataFrame frame, Callback callback)
         return _httpChannel.onContentAvailable();
     }
 
-    private Content.Abstract newContent(DataFrame frame, Runnable complete)
+    private Content.Chunk newChunk(DataFrame frame, Runnable complete)
```
lorban: This method can probably be inlined now.

jetty-core/jetty-http3/http3-server/src/main/java/org/eclipse/jetty/http3/server/internal/HttpStreamOverHTTP3.java
```diff
@@ -158,26 +159,13 @@ public Runnable onDataAvailable()
             return null;
         }
 
-        content = newContent(data);
+        chunk = newChunk(data);
         return httpChannel.onContentAvailable();
     }
 
-    private Content.Abstract newContent(Stream.Data data)
+    private Content.Chunk newChunk(Stream.Data data)
```
lorban: This method can probably be inlined now.

jetty-ee10/jetty-ee10-proxy/src/test/java/org/eclipse/jetty/ee10/proxy/AsyncMiddleManServletTest.java
```diff
@@ -66,10 +66,10 @@
 import org.eclipse.jetty.server.HttpConnectionFactory;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
-//import org.eclipse.jetty.server.internal.HttpChannelState;
```
lorban, line 69: Can be removed.

jetty-ee10/jetty-ee10-proxy/src/test/java/org/eclipse/jetty/ee10/proxy/ProxyServletFailureTest.java
```diff
@@ -40,11 +40,9 @@
 import org.eclipse.jetty.ee10.servlet.ServletHolder;
 import org.eclipse.jetty.http.HttpHeader;
 import org.eclipse.jetty.http.HttpTester;
-import org.eclipse.jetty.logging.StacklessLogging;
 import org.eclipse.jetty.server.HttpConnectionFactory;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
-//import org.eclipse.jetty.server.internal.HttpChannelState;
```
lorban, line 47: Can be removed.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HttpStream.java
```diff
@@ -19,22 +19,36 @@
...
+     * @return a chunk of content, possibly an {@link Chunk.Error error} or {@code null}.
+     */
+    Content.Chunk read();
 
-    void demandContent(); // Calls back on Channel#onDataAvailable
+    /**
+     * <p>Demands more content chunks to the underlying implementation.</p>
+     * <p>This method is called from the implementation of {@link Request#demand(Runnable)}.</p>
+     */
+    void demand();
```
lorban, line 51: Don't we want to document what gets called back?

lorban (changes requested):
A few minor things left, but LGTM.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/internal/ContentCopier.java
```diff
@@ -51,9 +51,6 @@ protected Action process() throws Throwable
             return Action.IDLE;
         }
 
-        if (current instanceof Content.Chunk.Error error)
-            throw error.getCause();
```
joakime: Is Content.Chunk.Error handled somewhere else now?

sbordet: Yes. Implementations of `Sink.write(Chunk, Callback)` will fail the callback if they receive a `Chunk.Error`.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSinkSubscriber.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public void onNext(Content.Chunk item)
+    {
+        sink.write(item, Callback.from(Callback.NOOP::succeeded, x -> subscription.cancel()));
+    }
+
+    @Override
+    public void onError(Throwable throwable)
+    {
+        sink.write(Content.Chunk.from(throwable), Callback.NOOP);
```
lorban: This implementation should actually result in a no-op since the `write(Chunk)` implementation is supposed to call the callback's `fail` method when the chunk is an error.

I'm not sure what to do in this case.

lorban (approved):
LGTM despite the suspicious `Flow` impl.

gregw:
@lorban Can you expand a bit here on the discussion you referenced in the email regarding `write(Chunk)`?
Specifically exactly why it is bad to have both.

I guess it makes the job of somebody that wants to wrap the response harder as they wrap write(boolean,ByteBuffer) and then themselves get wrapped, then either:
 1) the wrapper doesn't implement `write(Chunk)` and thus any ability to write trailers is lost, but at least the wrapped write is called; OR
 2) the wrapper does implement `write(Chunk)`, in which case the wrapped write is bypassed.

Ultimately I see `write(Chunk)` as 90% convenience method equivalent to `write(chunk.isLast(), chunk.getByteBuffer())` and 10% method for sending trailers....
So perhaps we just have a static `write(Chunk)` that does both the convenience thing, but also knows how to handle trailers.   We then always know that `write(Chunk)` will be on the outermost wrapper and both the trailers API and write API may be individually intercepted with specific APIs

gregw:
@lorban The other question about the write side is should it be `write(boolean, Callback, ByteBuffer...)` or `write(boolean, Callback, ByteBuffer)`?
The former is very rarely used, if ever.  But it costs a bit of allocation and complexity.      However it does have nice flush semantics of passing no buffers at all.
The later will need an empty buffer passed for flushes, but is simpler a implementation and covers 99.9999999% of the use-cases.

It is different at the lower layers, were we are often dealing with headers and body, thus gather writes make sense and can go all the way down to the network interface.... but for an application on top of http/2 or http/3, there is no way the gather write is going all the way down.  I guess it might influence the framing of the http>=2, but  is the cost of the gather worth it?     It may well be better to aggregate rather than gather.

lorban:
@gregw you mostly figured the problem: having more than one `write()` method forces you to be careful about how and when to call the wrapped Response. But it's more subtle than that: `write(Chunk, cb)` has to handle error content (which can be done at the io layer) and trailers content (which cannot be done at the io layer) so it's tempting to give it a default implementation that almost always does the wrong thing when Response is implemented.

This is why in the current PR's state `Content.write(Chunk, cb)` is overridden in Response to remove the default impl. And you also have the oddity of AsyncContent implementing `write(last, cb, ByteBuffer...)` by calling `write(Chunk, cb)`. During the multiple reviews of the code I discussed with @sbordet we figured that more often than not the principle of least surprise was shattered and there too often were non-obvious things to keep in mind w.r.t the implementation of those two `write()`.

Personally, I'm not fond of the `write(boolean, Callback, ByteBuffer...)` signature as being in a situation where we can perform gather writes IMHO is more of an exceptional than a common use-case. But I've also expressed my preference for `write(Chunk, cb)` which has the downside of forcing an allocation in some cases, but not in the most common one. `write(boolean, Callback, ByteBuffer)` would still be an improvement and I don't think having to pass an empty buffer instead of no buffer to signal a flush is a big deal.

lorban:
@gregw maybe we should get this merged before discussing the `write(xxx)` signatures? I think the current state of this PR is reasonable enough to get merged as-is and eventually reworking `write(xxx)` later doesn't sound too much of a big deal.

gregw:
@lorban @sbordet I tend to agree that we should merge this prior to resolving the write signature issue.    There is enough goodness here that we don't want it to go stale and maintaining 2 dev branches is needless double work.

Note also that the decision between `write(boolean, Callback, ByteBuffer)` and `write(Chunk, cb)` needs to be informed by benchmarks.  I too like the symmetry of the later and with `Chunk.of(buffer)` style coding it is simple enough to use.    It might also not be that much extra allocation for some common cases, as servlet HttpOutput will have an aggregating buffer, which could have a preallocated Chunk (or HttpOutput could be the Chunk itself).

It may even be that if we keep `write(boolean, Callback, ByteBuffer)` as a default wrapper of `write(Chunk.from(last,buffer), cb)` then the two APIs do work OK side by side?   Hmm maybe now I'm just talking myself into keeping both.... let's merge and think about it.

Build is green so let's go!

gregw (approved):

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -0,0 +1,621 @@
...
+         * @param chunk the chunk to write
+         * @param callback the callback to notify when the write operation is complete
+         * @see #write(boolean, Callback, ByteBuffer...)
+         */
+        public default void write(Chunk chunk, Callback callback)
+        {
+            if (chunk instanceof Content.Chunk.Error error)
+                callback.failed(error.getCause());
+            else
+                write(chunk.isLast(), callback, chunk.getByteBuffer());
```
gregw: This is actually wrong because it ignores the release conversation. Needs to be:
```suggestion
            if (chunk instanceof Content.Chunk.Error error)
            {
                chunk.release();
                callback.failed(error.getCause());
            }
            else
                write(chunk.isLast(), Callback.from(chunk::release, callback), chunk.getByteBuffer());
```

gregw (changes requested):

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
lorban: For the happy path, that's _another_ allocation on the fast path.
For the error path, since that `release()` call is a no-op I wonder if we should not document it as such on `Chunk.Error`.

lorban (approved):

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -0,0 +1,628 @@
...
+     * <p>If the predicate returns {@code false}, it means that the chunk is not
+     * handled, its callback will not be completed, and the implementation will
+     * handle the chunk and its callback.</p>
+     *
+     * @param source the source to copy from
+     * @param sink the sink to copy to
+     * @param callback the callback to notify when the copy is complete
+     * @param chunkHandler an optional predicate to handle the current chunk and its callback
+     */
+    public static void copy(Source source, Sink sink, Callback callback, BiPredicate<Chunk, Callback> chunkHandler)
```
gregw: maybe make the callback the last arg to match all other APIs with the callback last.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -0,0 +1,628 @@
...
+     * <p>If the predicate returns {@code true}, it means that the chunk is handled
+     * externally and its callback completed, or eventually completed.</p>
+     * <p>If the predicate returns {@code false}, it means that the chunk is not
+     * handled, its callback will not be completed, and the implementation will
+     * handle the chunk and its callback.</p>
+     *
+     * @param source the source to copy from
+     * @param sink the sink to copy to
+     * @param callback the callback to notify when the copy is complete
+     * @param chunkHandler an optional predicate to handle the current chunk and its callback
```
gregw: or null

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: Maybe have a version like:
```suggestion
    public static void copy(Source source, Sink sink, Callback callback)
```

gregw (approved):
Other than some niggles +1

PR #8008, opened 2022-05-17
Add compliance mode for LEGACY multipart parser in Jetty 10+

lachlan-roberts:
Introduce multipart `RFC7578`/`LEGACY` compliance modes to Jetty 10 with a default of `RFC7578`.
This allows an option to revert to the less efficient decoding of the LEGACY decoder which allows deprecated line terminations and decodes parts with the deprecated `Content-Transfer-Encoding`.

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/ReadLineInputStream.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/ReadLineInputStream.java
@@ -24,7 +24,7 @@ import java.util.EnumSet;
  *
  * Read from an input stream, accepting CR/LF, LF or just CR.
  */
-@Deprecated
+@Deprecated(forRemoval = true)
 public class ReadLineInputStream extends BufferedInputStream
 {
     boolean _seenCRLF;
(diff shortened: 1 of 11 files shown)


gregw (changes requested):
If we need this legacy mode going forward, then I don't think it is viable to continue with 2 different parsers.   We will need to work out how to make the one parser handle the different line terminations

jetty-util/src/main/java/org/eclipse/jetty/util/ReadLineInputStream.java
```diff
@@ -0,0 +1,161 @@
...
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.EnumSet;
+
+/**
+ * ReadLineInputStream
+ *
+ * Read from an input stream, accepting CR/LF, LF or just CR.
+ */
+@Deprecated
```
gregw: add marked for deletion

gregw (dismissed):
I am provisionally approving this just to solve the short term problem......

BUT this should not be merged forward to jetty-12 (well do the merge, but then revert the code).  In 12, we need to go to fully asynchronous multipart parsing and it will be unsustainable to have 2 parsers to be so maintained.   

So keep the issue open and let's solve it properly in 12.

gregw (approved):

PR #8015, opened 2022-05-17
Fixes #8014 - Review HttpRequest URI construction.

### LINKED ISSUE #8014: Review HttpRequest URI construction (CVE-2022-2047)
**Jetty version(s)**
10.0.x

**Description**
Certain URIs may be misconstructed by `HttpRequest`, possibly resulting in a wrong `Host` header.

**Fixes Security Advisory**
https://github.com/eclipse/jetty.project/security/advisories/GHSA-cj7v-27pg-wf7q
CVE-2022-2047

sbordet:
Now always adding a "/" before the path, if not already present.
Disabled flakey HTTP/3 test.

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/HttpRequest.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/HttpRequest.java
@@ -195,6 +195,8 @@ public class HttpRequest implements Request
             String rawPath = uri.getRawPath();
             if (rawPath == null)
                 rawPath = "";
+            if (!rawPath.startsWith("/"))
+                rawPath = "/" + rawPath;
             this.path = rawPath;
             String query = uri.getRawQuery();
             if (query != null)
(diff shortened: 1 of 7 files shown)


joakime (dismissed):
LGTM

gregw:
@Simone, i think we should also fix the asString method in HttpURI. I know it's not used, but it makes the same mistake with non root paths.

joakime (dismissed):
I like the IllegalArgumentExceptions on this

joakime (dismissed):
The OPTIONS path of `*` makes this interesting for sure.

`isPathValidForAuthority(String)` method makes this much easier to understand now too.

gregw:
Note, I'm working on fixing the test failure on this one.

gregw:
@sbordet @joakime I've made a change to treat the URI of a CONNECT request as the Authority rather than as a path (which I believe is more correct as per the RFC).
But it is kind of a significant change, so careful review needed.

joakime (approved):
Seems like a reasonable change.
Lets see what the CI build/test shows.

gregw:
The CI failure was an unrelated problem with a repository.  Retrying...

gregw:
Still having CI issues... which I think are unrelated.   If I get a clean build (again) locally, I will merge.

Issue #8021, opened 2022-05-19
Can we test the deployment of P2 with the new tycho version + signature?


laeubi:
**Jetty version**
10.x , 11.x

**Java version**

don't matter :-)

**Question**

I'd like to ask if it would be possible to once again spin a special "snapshot deploy" to test the new tycho-version before the next release (is there a schedule?) to make sure the following work:

- no additional items included
- PGP signatures are setup

Would be great if that's possible to catch any issues early :+1:

olamy:
they usually run daily if source code change
https://jenkins.webtide.net/job/nightlies/job/jetty-10.0.x-snapshots-deploy/
https://jenkins.webtide.net/job/nightlies/job/jetty-11.0.x-snapshots-deploy/

so you can see date/time of last run

laeubi:
@olamy the "problem" is that the daily run do not PGP sign the artifact, at least last time there must be some "magic" enabled for that (I can't find the issue right now).

olamy:
> @olamy the "problem" is that the daily run do not PGP sign the artifact, at least last time there must be some "magic" enabled for that (I can't find the issue right now).

oh ok. I guess the magic last time was my fingers 🤣
I will do it

olamy:
@laeubi both branches deployed with my sig 
enjoy!

laeubi:
Great, the signatures are there, but I notice that when looking at the `artifacts.xml` here

https://oss.sonatype.org/content/repositories/jetty-snapshots/org/eclipse/jetty/jetty-p2/10.0.10-SNAPSHOT/jetty-p2-10.0.10-20220519.075904-31-p2site.zip

it seems missing a lot of the jetty bundles missing, also for older builds this is the case, currently the repo references the

https://github.com/eclipse/jetty.project/blob/546c382255f69cbc3b467e6ae87a8d9ccb8bd036/jetty-p2/pom.xml#L44-L49

and includes all reactor projects but it seems not to be enough, I see two possible causes:

1. the bom is not enough to make the p2- to be build as a last step
2. not the full set of projects are build

any idea?

olamy:
Oh I think I know. I used -T3 so because of this it looks the p2 is build before some other modules and not at the end because they are not declared in p2 pom.
I will redeploy tomorrow and try to fix that

laeubi:
Good to find that early now :-) Maybe you have a better idea to let the p2 depend on "everything else" instead of referencing the bom or literally every single module?

[...]

PR #8026, opened 2022-05-20
make jetty-p2 depending jetty-home so we sure it's build last even when using -Tx option see https://github.com/eclipse/jetty.project/issues/8021

olamy:


joakime (approved):
makes sense.

PR #8032, opened 2022-05-20
do not use -Dmaven.test.failure.ignore=true

olamy:
Signed-off-by: Olivier Lamy <oliver.lamy@gmail.com>

PR #8036, opened 2022-05-23
use surefire 3.0.0 M7 SNAPSHOT so we can restore running all tests with -Dmaven.test.failure.ignore=true

olamy:
- test surefire 3.0.0-M7-SNAPSHOT and restore -Dmaven.test.failure.ignore=true
- add apache snapshots repository

@joakime why we need need this :)
because we need https://issues.apache.org/jira/browse/SUREFIRE-1426 which is fixed in M6 (if not we do not see jpms issues) but M6 has this https://issues.apache.org/jira/browse/SUREFIRE-2057


joakime (approved):

joakime:
Lots of failures on this PR build.
Make sure you address those before merging

olamy:
I just wanted to restore this so we have a full build of jetty 12 and report of all issues
not sure I can fix all the issues of jetty 12 branch in this single PR  😀

olamy:
to return to situation before https://github.com/eclipse/jetty.project/pull/8032

PR #8037, opened 2022-05-23
Jetty 12 gzip handler

gregw:
Whilst fixing some ee9 tests, I found a few issues in the core GzipHandler
@lachlan-roberts and I believed we fixed it, but review of the fix would be good.

The problem was mostly due to complications from support gather writes and/or writes of empty buffers.


gregw:
@lorban @sbordet I think the build failures here are flakes.  Can you review without waiting for a clean build.

janbartel (approved):

PR #8047, opened 2022-05-23
Jetty 12.0.x cookie cache rewrite

gregw:
CookieCache was leaking cookies from one request to the next.  It was too complex so I just rewrote it and added a unit test


janbartel (approved):
A couple more comments would not go astray in the getCookies() method.

PR #8049, opened 2022-05-24
jetty 10.0.x no port 8080

olamy:
- do not start on port 8080
- ensure it test is not started on port 8080 and use non random groupIds

### DIFF
--- a/jetty-maven-plugin/src/it/jetty-run-mojo-jar-scan-it/MyLibrary/pom.xml
+++ b/jetty-maven-plugin/src/it/jetty-run-mojo-jar-scan-it/MyLibrary/pom.xml
@@ -3,13 +3,13 @@
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <parent>
-    <artifactId>jetty-issue</artifactId>
-    <groupId>org.mehdi</groupId>
+    <groupId>org.eclipse.jetty.its.jetty-run-mojo-jar-scan-it</groupId>
+    <artifactId>jetty-jar-scan</artifactId>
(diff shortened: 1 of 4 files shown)


jetty-maven-plugin/src/it/jetty-run-mojo-jar-scan-it/MyLibrary/pom.xml
```diff
@@ -3,13 +3,13 @@
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <parent>
-    <artifactId>jetty-issue</artifactId>
-    <groupId>org.mehdi</groupId>
+    <groupId>org.eclipse.jetty.its.jetty-run-mojo-jar-scan-it</groupId>
+    <artifactId>jetty-jar-scan</artifactId>
     <version>1.0-SNAPSHOT</version>
```
lorban, line 8: Depending on SNAPSHOT versions is troublesome, don't you think?

gregw (changes requested):
yeah can we avoid a SNAPSHOT dependency?

jetty-maven-plugin/src/it/jetty-run-mojo-jar-scan-it/MyLibrary/pom.xml
olamy: no the **main** Jetty build it's not depending on snapshot with this :) 
src/it contains projects which are invoked by maven-invoker-plugin.
The **main** Jetty build do not try to download those projects.

olamy:
> yeah can we avoid a SNAPSHOT dependency?

explained here https://github.com/eclipse/jetty.project/pull/8049#discussion_r881065632

there is no SNAPSHOT dependency added here it's just test projects

gregw (approved):

PR #8058, opened 2022-05-26
Implement #8057 103 Early Hint

### LINKED ISSUE #8057: Support Http Response 103 (Early Hints)
**Target Jetty version(s)**
10+

**Enhancement Description**

What would it take to have Jetty support response code 103 (Early Hints)?

* https://datatracker.ietf.org/doc/html/rfc8297
* https://www.fastly.com/blog/beyond-server-push-experimenting-with-the-103-early-hints-status-code
* https://developers.cloudflare.com/cache/about/early-hints/

gregw:
Implementing #8057 Early hints

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/http/HttpReceiverOverHTTP.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/http/HttpReceiverOverHTTP.java
@@ -187,9 +187,12 @@ public class HttpReceiverOverHTTP extends HttpReceiver implements HttpParser.Res
                 }
                 else if (read == 0)
                 {
-                    releaseNetworkBuffer();
-                    fillInterested();
-                    return;
+                    if (networkBuffer.isEmpty())
+                    {
+                        releaseNetworkBuffer();
+                        fillInterested();
+                        return;
+                    }
                 }
                 else
                 {
@@ -245,19 +248,22 @@ public class HttpReceiverOverHTTP extends HttpReceiver implements HttpParser.Res
                 this.method = null;
                 if (getHttpChannel().isTunnel(method, status))
                     return true;
+
+                if (networkBuffer.isEmpty())
+                    return false;
+
+                if (!HttpStatus.isInformational(status))
+                {
+                    if (LOG.isDebugEnabled())
+                        LOG.debug("Discarding unexpected content after response {}: {}", status, networkBuffer);
+                    LOG.info("Discarding unexpected content after response {}: {}", status, networkBuffer);
+                    networkBuffer.clear();
+                }
+                return false;
             }
 
             if (networkBuffer.isEmpty())
                 return false;
-
-            if (complete)
-            {
-                // TODO this can clean a pipelined response after a 1xx
-                if (LOG.isDebugEnabled())
-                    LOG.debug("Discarding unexpected content after response: {}", networkBuffer);
-                networkBuffer.clear();
-                return false;
-            }
         }
     }
 
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/http/HttpReceiverOverHTTP.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/http/HttpReceiverOverHTTP.java
@@ -252,6 +252,7 @@ public class HttpReceiverOverHTTP extends HttpReceiver implements HttpParser.Res
 
             if (complete)
             {
+                // TODO this can clean a pipelined response after a 1xx
                 if (LOG.isDebugEnabled())
                     LOG.debug("Discarding unexpected content after response: {}", networkBuffer);
                 networkBuffer.clear();
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/http/HttpReceiverOverHTTP.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/http/HttpReceiverOverHTTP.java
@@ -187,12 +187,10 @@ public class HttpReceiverOverHTTP extends HttpReceiver implements HttpParser.Res
                 }
                 else if (read == 0)
                 {
-                    if (networkBuffer.isEmpty())
-                    {
-                        releaseNetworkBuffer();
-                        fillInterested();
-                        return;
-                    }
+                    assert networkBuffer.isEmpty();
+                    releaseNetworkBuffer();
+                    fillInterested();
+                    return;
                 }
                 else
                 {
@@ -256,7 +254,6 @@ public class HttpReceiverOverHTTP extends HttpReceiver implements HttpParser.Res
                 {
                     if (LOG.isDebugEnabled())
                         LOG.debug("Discarding unexpected content after response {}: {}", status, networkBuffer);
-                    LOG.info("Discarding unexpected content after response {}: {}", status, networkBuffer);
                     networkBuffer.clear();
                 }
                 return false;
@@ -379,7 +376,7 @@ public class HttpReceiverOverHTTP extends HttpReceiver implements HttpParser.Res
         }
 
         int status = exchange.getResponse().getStatus();
-        if (status != HttpStatus.CONTINUE_100)
+        if (!HttpStatus.isInterim(status))
         {
             inMessages.increment();
             complete = true;
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
@@ -385,7 +385,7 @@ public class HttpGenerator
 
                     // Handle 1xx and no content responses
                     int status = info.getStatus();
-                    if (status >= 100 && status < 200)
+                    if (HttpStatus.isInformational(status))
                     {
                         _noContentResponse = true;
                         switch (status)
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
@@ -388,12 +388,18 @@ public class HttpGenerator
                     if (status >= 100 && status < 200)
                     {
                         _noContentResponse = true;
-
-                        if (status != HttpStatus.SWITCHING_PROTOCOLS_101)
+                        switch(status)
                         {
-                            header.put(HttpTokens.CRLF);
-                            _state = State.COMPLETING_1XX;
-                            return Result.FLUSH;
+                            case HttpStatus.SWITCHING_PROTOCOLS_101:
+                                break;
+                            case HttpStatus.EARLY_HINT_103:
+                                generateHeaders(header, content, last);
+                                _state = State.COMPLETING_1XX;
+                                return Result.FLUSH;
+                            default:
+                                header.put(HttpTokens.CRLF);
+                                _state = State.COMPLETING_1XX;
+                                return Result.FLUSH;
                         }
                     }
                     else if (status == HttpStatus.NO_CONTENT_204 || status == HttpStatus.NOT_MODIFIED_304)
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
@@ -388,7 +388,7 @@ public class HttpGenerator
                     if (status >= 100 && status < 200)
                     {
                         _noContentResponse = true;
-                        switch(status)
+                        switch (status)
                         {
                             case HttpStatus.SWITCHING_PROTOCOLS_101:
                                 break;
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpStatus.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpStatus.java
@@ -342,6 +342,17 @@ public class HttpStatus
     }
 
     /**
+     * Tests whether the status code is informational but not {@code 101 Switching Protocols}.
+     *
+     * @param code the code to test
+     * @return whether the status code is informational but not {@code 101 Switching Protocols}
+     */
+    public static boolean isInterim(int code)
+    {
+        return isInformational(code) && code != HttpStatus.SWITCHING_PROTOCOLS_101;
+    }
+
+    /**
      * Simple test against an code to determine if it falls into the
      * <code>Success</code> message category as defined in the <a
      * href="http://tools.ietf.org/html/rfc1945">RFC 1945 - HTTP/1.0</a>, and <a
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/HttpStatus.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/HttpStatus.java
@@ -25,6 +25,7 @@ public class HttpStatus
     public static final int CONTINUE_100 = 100;
     public static final int SWITCHING_PROTOCOLS_101 = 101;
     public static final int PROCESSING_102 = 102;
+    public static final int EARLY_HINT_103 = 103;
 
     public static final int OK_200 = 200;
     public static final int CREATED_201 = 201;
@@ -103,6 +104,7 @@ public class HttpStatus
         CONTINUE(CONTINUE_100, "Continue"),
         SWITCHING_PROTOCOLS(SWITCHING_PROTOCOLS_101, "Switching Protocols"),
         PROCESSING(PROCESSING_102, "Processing"),
+        EARLY_HINT(EARLY_HINT_103, "Early Hint"),
 
         OK(OK_200, "OK"),
         CREATED(CREATED_201, "Created"),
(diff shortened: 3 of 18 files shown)


gregw:
@lorban can you look at why h3 is failing.

lorban:
@gregw I see no H3 test failure in the PR's build, only a checkstyle error. What am I overlooking?

gregw:
@lorban have you run org.eclipse.jetty.http.client.IntermediateResponseTest#test103EarlyHint ?

lorban:
@gregw I've added the missing support in H3 for 103 status code and pushed the changes to this branch.

While your test now passes and I'm reasonably confident my changes are correct, it would still be wise to have @sbordet review this bit eventually.

jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
```diff
@@ -388,12 +388,18 @@ public Result generateResponse(MetaData.Response info, boolean head, ByteBuffer
                     if (status >= 100 && status < 200)
```
sbordet: There is `HttpStatus.isInformational()` that we should use here.

jetty-http2/http2-server/src/main/java/org/eclipse/jetty/http2/server/HttpTransportOverHTTP2.java
```diff
@@ -98,7 +98,7 @@ public void sendHeaders(MetaData.Request request, MetaData.Response response, By
         boolean isHeadRequest = HttpMethod.HEAD.is(request.getMethod());
         boolean hasContent = BufferUtil.hasContent(content) && !isHeadRequest;
         int status = response.getStatus();
-        boolean interimResponse = status == HttpStatus.CONTINUE_100 || status == HttpStatus.PROCESSING_102;
+        boolean interimResponse = status >= HttpStatus.CONTINUE_100 && status < HttpStatus.OK_200 && status != HttpStatus.SWITCHING_PROTOCOLS_101;
```
sbordet: Replace with `HttpStatus.isInformational() && !101`.

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/IntermediateResponseTest.java
```diff
@@ -0,0 +1,298 @@
...
+    public void init(Transport transport) throws IOException
+    {
+        // Skip FCGI for now, not much interested in its server-side behavior.
+        Assumptions.assumeTrue(transport != FCGI);
+        setScenario(new TransportScenario(transport));
+    }
+
+    @ParameterizedTest
+    @ArgumentsSource(TransportProvider.class)
+    public void test100Continue(Transport transport) throws Exception
```
sbordet: There are a lot of 100-Continue tests elsewhere, so I don't think we need to duplicate this here.
I'd remove this test.

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/IntermediateResponseTest.java
```diff
@@ -0,0 +1,298 @@
...
+                        org.eclipse.jetty.client.api.Request request = response.getRequest();
+                        HttpConversation conversation = ((HttpRequest)request).getConversation();
+                        // Reset the conversation listeners, since we are going to receive another response code
+                        conversation.updateResponseListeners(null);
+
+                        HttpExchange exchange = conversation.getExchanges().peekLast();
+                        if (exchange != null && response.getStatus() == HttpStatus.EARLY_HINT_103)
+                        {
+                            // All good, continue.
+                            System.err.println("onSuccess\n" + response.getHeaders());
```
sbordet: Remove `System.err`

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/IntermediateResponseTest.java
```diff
@@ -0,0 +1,298 @@
...
+                            hints.add(response.getHeaders().get("Hint"));
+                            exchange.resetResponse();
+                            exchange.proceed(null);
+                        }
+                        else
+                        {
+                            throw new IllegalStateException("should not have accepted");
```
sbordet: This exception will only be logged, better call `response.abort(new ISE(...))`, so the test will fail.

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/IntermediateResponseTest.java
```diff
@@ -0,0 +1,298 @@
...
+                        HttpExchange exchange = conversation.getExchanges().peekLast();
+                        if (exchange != null && response.getStatus() == HttpStatus.PROCESSING_102)
+                        {
+                            // All good, continue.
+                            exchange.resetResponse();
+                            exchange.proceed(null);
+                        }
+                        else
+                        {
+                            throw new IllegalStateException("should not have accepted");
```
sbordet: Use `response.abort(new ISE(...))`.

jetty-http3/http3-common/src/main/java/org/eclipse/jetty/http3/internal/HTTP3StreamConnection.java
```diff
@@ -486,7 +486,7 @@ public void onHeaders(long streamId, HeadersFrame frame)
             else if (metaData.isResponse())
             {
                 MetaData.Response response = (MetaData.Response)metaData;
-                if (response.getStatus() != HttpStatus.CONTINUE_100)
+                if (response.getStatus() != HttpStatus.CONTINUE_100 && response.getStatus() != HttpStatus.EARLY_HINT_103)
```
sbordet: This should be `HttpStatus.isInformational()`.

jetty-http/src/main/java/org/eclipse/jetty/http/HttpStatus.java
```diff
@@ -103,6 +104,7 @@
         CONTINUE(CONTINUE_100, "Continue"),
         SWITCHING_PROTOCOLS(SWITCHING_PROTOCOLS_101, "Switching Protocols"),
         PROCESSING(PROCESSING_102, "Processing"),
+        EARLY_HINT(EARLY_HINT_103, "Early Hint"),
```
sbordet, line 107: Let's add also `HttpHeader.Link` from RFC 8288.

jetty-http3/http3-client/src/main/java/org/eclipse/jetty/http3/client/internal/HTTP3StreamClient.java
```diff
@@ -53,6 +53,8 @@ public void onResponse(HeadersFrame frame)
         boolean valid;
         if (response.getStatus() == HttpStatus.CONTINUE_100)
             valid = validateAndUpdate(EnumSet.of(FrameState.INITIAL), FrameState.CONTINUE);
+        else if (response.getStatus() == HttpStatus.EARLY_HINT_103)
+            valid = validateAndUpdate(EnumSet.of(FrameState.INITIAL, FrameState.HEADER, FrameState.CONTINUE), FrameState.CONTINUE);
```
sbordet: I would rename `FrameState.CONTINUE` to `INFORMATIONAL`.

State HEADER is the state used to validate data frames (i.e. we can only receive a DATA frame if we are in HEADER state).
So I think the "from" states can only be INITIAL and CONTINUE(INFORMATIONAL).

jetty-http3/http3-server/src/main/java/org/eclipse/jetty/http3/server/internal/HttpTransportOverHTTP3.java
```diff
@@ -71,7 +71,7 @@ private void sendHeaders(MetaData.Request request, MetaData.Response response, B
         boolean isHeadRequest = HttpMethod.HEAD.is(request.getMethod());
         boolean hasContent = BufferUtil.hasContent(content) && !isHeadRequest;
         int status = response.getStatus();
-        boolean interimResponse = status == HttpStatus.CONTINUE_100 || status == HttpStatus.PROCESSING_102;
+        boolean interimResponse = status >= HttpStatus.CONTINUE_100 && status < HttpStatus.OK_200 && status != HttpStatus.SWITCHING_PROTOCOLS_101;
```
sbordet: Replace with `HttpStatus.isInformational() && !101`.

jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -514,6 +517,23 @@ public void sendProcessing() throws IOException
...
+    /**
+     * Sends a 102-Processing response.
+     * If the connection is an HTTP connection, the version is 1.1 and the
+     * request has a Expect header starting with 102, then a 102 response is
+     * sent. This indicates that the request still be processed and real response
+     * can still be sent.   This method is called by sendError if it is passed 102.
+     *
+     * @throws IOException if unable to send the 102 response
+     * @see javax.servlet.http.HttpServletResponse#sendError(int)
+     */
```
sbordet, line 528: Javadoc is wrong.

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/IntermediateResponseTest.java
```diff
@@ -0,0 +1,298 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ArgumentsSource;
+
+import static org.eclipse.jetty.http.client.Transport.FCGI;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class IntermediateResponseTest extends AbstractTest<TransportScenario>
```
sbordet: Rename to `InformationalResponseTest`.

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/IntermediateResponseTest.java
```diff
@@ -0,0 +1,298 @@
...
+
+            @Override
+            public Response.Listener getResponseListener()
+            {
+                return new Response.Listener()
+                {
+                    @Override
+                    public void onSuccess(Response response)
+                    {
+                        org.eclipse.jetty.client.api.Request request = response.getRequest();
```
sbordet: Use `var request = ...`.

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/IntermediateResponseTest.java
```diff
@@ -0,0 +1,298 @@
...
```
sbordet: Use `var request = ...`.

sbordet (changes requested):

tests/test-http-client-transport/src/test/java/org/eclipse/jetty/http/client/IntermediateResponseTest.java
gregw: I'm not sure the other tests made sending the content actually conditional on receiving the 100, as I certainly could not find the magic needed in the HttpClient elsewhere.    So I'd be more inclined to delete the other tests and keep this one.  I'll look for any other differences

gregw: OK now I find the good 100 test!

gregw:
@lorban @sbordet I've updated for review feedback.
However, there is still something racy either in the tests or the impl.   UNIX_DOMAIN frequently fails for me if run with other tests (the second 103 is skipped).  I have once seen H2 fail by the 103 being reported as the final response instead of the 200, but I can't reproduce.     Something is not quiet right :(

gregw:
Hmmm and I've just had a flake of the 102 test for UNIX_DOMAIN as well... also not reproducible :(

gregw:
@sbordet what does the client do if it is still processing the 102/103 response and the final (or next) response turns up?    On the server, we don't wait for anything from the client before sending the next response, but if I put in a 10ms sleep between the responses then I don't get any flakes.      Either we are writing over the responses or the client is processing the next one before the last is finished?

gregw:
@sbordet I think the race is in the client.  If I put the 10ms sleeps in the server I pass every time. However, if I put a 100ms sleep in the onSuccess handling for the 1xx responses, then I fail more often than not, even for HTTP.

jetty-http3/http3-client/src/main/java/org/eclipse/jetty/http3/client/internal/HTTP3StreamClient.java
lorban: I agree that `FrameState.CONTINUE` should be renamed to `FrameState.INFORMATIONAL`.

I'm wondering if this code could be generalized or if we have no choice but to stick to a `if (status == 103)` special case.

jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -514,6 +517,22 @@ public void sendProcessing() throws IOException
         }
     }
 
+    /**
+     * Sends a 103 Early Hint response.
+     *
+     * Send a 103 response as per <a href="https://datatracker.ietf.org/doc/html/rfc8297">RFC8297</a>
+     * This method is called by sendError if it is passed 103.
+     *
+     * @throws IOException if unable to send the 102 response
```
lorban:
```diff
- if unable to send the 102 response
+ if unable to send the 103 response
```

lorban:
All tests in `IntermediateResponseTest` eventually fail when ran enough times, even if the H3 and UNIX_DOMAIN ones are more sensible.

jetty-http3/http3-common/src/main/java/org/eclipse/jetty/http3/internal/HTTP3StreamConnection.java
gregw: Is that OK for a 101 switching protocols?

jetty-http/src/main/java/org/eclipse/jetty/http/HttpStatus.java
gregw: Added in javadoc

jetty-http3/http3-client/src/main/java/org/eclipse/jetty/http3/client/internal/HTTP3StreamClient.java
gregw: There is no special case for 102???  Let's hangout and walk through this logic.

gregw:
@sbordet @lorban The flaky issue is somewhere below the generation of the response in the server and the parsing of it in the client.   I added some println's and for a good run I see:
```
generateResponseLine HTTP/1.1{s=103,h=2,cl=-1}
startResponse 103 Early Hint
generateResponseLine HTTP/1.1{s=103,h=2,cl=-1}
startResponse 103 Early Hint
generateResponseLine HTTP/1.1{s=200,h=2,cl=-1}
startResponse 200 OK
```
So each generated response is parsed (startResponse) in the client.

But for a bad run I see:
```
generateResponseLine HTTP/1.1{s=103,h=2,cl=-1}
generateResponseLine HTTP/1.1{s=103,h=2,cl=-1}
startResponse 103 Early Hint
generateResponseLine HTTP/1.1{s=200,h=2,cl=-1}
startResponse 200 OK
```
So all the responses are generated, but one of them is not received.     It feels like something might be clearing a buffer when it should not?     

@lorban can you focus on the H2/H3 flakes and I'll keep looking at h1

gregw:
oops sorry forced pushed after a rebase

jetty-client/src/main/java/org/eclipse/jetty/client/http/HttpReceiverOverHTTP.java
```diff
@@ -252,6 +252,7 @@ private boolean parse()
 
             if (complete)
             {
+                // TODO this can clean a pipelined response after a 1xx
                 if (LOG.isDebugEnabled())
                     LOG.debug("Discarding unexpected content after response: {}", networkBuffer);
                 networkBuffer.clear();
```
gregw: @sbordet @lorban this is the cause of the UNIX_DOMAIN flake.  A pipelined response can be discarded after a 1xx response.    I'm not sure of the fix.

gregw:
@sbordet I found and maybe fixed the flake in the HttpClient (discarding content after a 1xx), but would appreciate very careful review.
@lorban I can't get any flaky fails in h2 or h3? How about you?

lorban:
@gregw I can make `HttpClientStreamTest.testFileUpload` on H3 fail about once every couple dozen runs (I'm aware of that failure and I already started investigating it) but I cannot manage to get any of the tests in `InformationalResponseTest` to fail for H2 or H3 anymore.

That's strange...

sbordet:
@gregw looking

lorban (approved):

sbordet (approved):

PR #8082, opened 2022-05-31
Fix #8067 Use nanotime for DosFilter rate tracker

### LINKED ISSUE #8067: Wall time usage in DoSFilter RateTracker results in false positive alert
**Jetty version(s)**
9.4.46

openjdk version "1.8.0_312"
OpenJDK Runtime Environment (build 1.8.0_312-8u312-b07-0ubuntu1~20.04-b07)
OpenJDK 64-Bit Server VM (build 25.312-b07, mixed mode)

**OS type/version**
Ubuntu 20.04 LTS

**Description**
We have observed that when the system time is set backwards in our environment, either through manual intervention or
by NTP/chrony/timesync, any active clients captured in the RateTracker in the DoSFilter will instantly demonstrate an ALERT on the next request.

I see that the following issue:

https://github.com/eclipse/jetty.project/issues/121

highlights that Wall time is used in the filter RateTracker to measure time elapsed:

```
jetty-servlets/src/main/java/org/eclipse/jetty/servlets/DoSFilter.java:318:            final boolean overRateLimit = tracker.isRateExceeded(System.currentTimeMillis());
jetty-servlets/src/main/java/org/eclipse/jetty/servlets/DoSFilter.java:1280:            boolean hasRecentRequest = last != 0 && (System.currentTimeMillis() - last) < 1000L;
```

Current DoSFilter class definition demonstrates this is still in place at:

```
jetty-servlets\src\main\java\org\eclipse\jetty\servlets\DoSFilter.java(329): final OverLimit overLimit = tracker.isRateExceeded(System.currentTimeMillis());
jetty-servlets\src\main\java\org\eclipse\jetty\servlets\DoSFilter.java(1330): boolean hasRecentRequest = last != 0 && (System.currentTimeMillis() - last) < 1000L;
```

If this time is set backwards (date used is Sat 26 Oct 1985 01:21:00 AM) , the alert will appear as:

```
WARNING [01:21:05 26-Oct-85 UTC] DOS ALERT: Request rejected ip=192.168.56.11, overlimit=OverLimit@36fec835[type=IP, id=192.168.56.11, duration=PT-320728H-7M-49.388S, count=25], session=null, user=null
```

and the alert condition will not clear until time has moved forward once more past the original millis.

The condition may also manifest if time is moved forward, but in a less problematic way as it could mean that a potential DOS condition is not recognized; however, the time moving backwards scenario is the one more likely to cause frustration.

I've worked around this by using the removeFromRateTracker(String id) function to fix specific client's being actively tracked, but I think it would be very useful if the DoSFilter class exposed the _rateTracker().clear() functionality so that I could more easily inform the filter that time has meaningfully changed in the host and *all* clients should be reset. Our workaround is inadequate for some of the boundary conditions and a solution in the RateTracker itself is preferred.

The ideal solution may be to use nanoTime() or some other monotonic clock to track duration in the RateTracker. If switching to the monotonic clock is too disruptive, perhaps simply allow a negative duration value (in above duration=PT-320728H-7M-49.388S) to be thrown out as a recognized condition that time has gone backwards and my alert is false?

**How to reproduce?**
Enable the DoSFilter in a Jetty server instance.
Make a request to the server that will be captured by RateTracker, IP type used
Set system time back a significant amount
Make another request to the server from same IP address
Observe the DOS Alert is triggered

gregw:
Use nano time to avoid false positives when wall clock changes.

### DIFF
--- a/jetty-servlets/src/main/java/org/eclipse/jetty/servlets/DoSFilter.java
+++ b/jetty-servlets/src/main/java/org/eclipse/jetty/servlets/DoSFilter.java
@@ -1216,8 +1216,8 @@ public class DoSFilter implements Filter
         }
 
         /**
-         * @param now the time now (in nanoseconds)
-         * @return the current calculated request rate over the last second
+         * @param now the time now (in nanoseconds) used to calculate elapsed time since previous requests.
+         * @return the current calculated request rate over the last second if rate exceeded, else null.
          */
         public OverLimit isRateExceeded(long now)
         {
--- a/jetty-servlets/src/main/java/org/eclipse/jetty/servlets/DoSFilter.java
+++ b/jetty-servlets/src/main/java/org/eclipse/jetty/servlets/DoSFilter.java
@@ -1216,7 +1216,7 @@ public class DoSFilter implements Filter
         }
 
         /**
-         * @param now the time now (in milliseconds)
+         * @param now the time now (in nanoseconds)
          * @return the current calculated request rate over the last second
          */
         public OverLimit isRateExceeded(long now)
@@ -1237,7 +1237,7 @@ public class DoSFilter implements Filter
             long rate = (now - last);
             if (TimeUnit.NANOSECONDS.toSeconds(rate) < 1L)
             {
-                return new Overage(Duration.ofMillis(rate), _maxRequestsPerSecond);
+                return new Overage(Duration.ofNanos(rate), _maxRequestsPerSecond);
             }
             return null;
         }
--- a/jetty-servlets/src/main/java/org/eclipse/jetty/servlets/DoSFilter.java
+++ b/jetty-servlets/src/main/java/org/eclipse/jetty/servlets/DoSFilter.java
@@ -326,7 +326,7 @@ public class DoSFilter implements Filter
         tracker = getRateTracker(request);
 
         // Calculate the rate and check if it is over the allowed limit
-        final OverLimit overLimit = tracker.isRateExceeded(System.currentTimeMillis());
+        final OverLimit overLimit = tracker.isRateExceeded(System.nanoTime());
 
         // Pass it through if we are not currently over the rate limit.
         if (overLimit == null)
@@ -1235,7 +1235,7 @@ public class DoSFilter implements Filter
             }
 
             long rate = (now - last);
-            if (rate < 1000L)
+            if (TimeUnit.NANOSECONDS.toSeconds(rate) < 1L)
             {
                 return new Overage(Duration.ofMillis(rate), _maxRequestsPerSecond);
             }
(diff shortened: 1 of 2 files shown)


jetty-servlets/src/main/java/org/eclipse/jetty/servlets/DoSFilter.java
```diff
@@ -1235,7 +1235,7 @@ public OverLimit isRateExceeded(long now)
             }
 
             long rate = (now - last);
-            if (rate < 1000L)
+            if (TimeUnit.NANOSECONDS.toSeconds(rate) < 1L)
             {
                 return new Overage(Duration.ofMillis(rate), _maxRequestsPerSecond);
```
lachlan-roberts: Why is this still using `Duration.ofMillis` ?

jetty-servlets/src/main/java/org/eclipse/jetty/servlets/DoSFilter.java
```diff
@@ -326,7 +326,7 @@ protected void doFilter(HttpServletRequest request, HttpServletResponse response
         tracker = getRateTracker(request);
 
         // Calculate the rate and check if it is over the allowed limit
-        final OverLimit overLimit = tracker.isRateExceeded(System.currentTimeMillis());
+        final OverLimit overLimit = tracker.isRateExceeded(System.nanoTime());
```
lachlan-roberts, line 329: Can you update the javadoc of `isRateExceeded`. It specifies that the value is in ms.
>  now the time now (in milliseconds)

lachlan-roberts (changes requested):

jetty-servlets/src/main/java/org/eclipse/jetty/servlets/DoSFilter.java
gregw: oops

jetty-servlets/src/main/java/org/eclipse/jetty/servlets/DoSFilter.java
```diff
@@ -1216,7 +1216,7 @@ public RateTracker(ServletContext context, String filterName, String id, RateTyp
         }
 
         /**
-         * @param now the time now (in milliseconds)
+         * @param now the time now (in nanoseconds)
```
lorban: Could you please mention elapsed time or something in the javadoc as a hint that `nanoTime()` should be used instead of `currentTimeMillis()`?

lorban (dismissed):

olamy (dismissed):

sbordet (dismissed):

gregw:
@lachlan-roberts can you rereview to remove your objection.

lachlan-roberts (approved):

gregw:
I think the CI failure is some kind of CI problem. Distribution tests work locally

PR #8087, opened 2022-06-01
Issue #6328 - avoid binding WebSocket MethodHandles

### LINKED ISSUE #6328: High CPU usage of method handle invocations in Jetty 10
**Jetty version**

`10.0.3`

```
openjdk version "13.0.4" 2020-07-14
OpenJDK Runtime Environment Zulu13.33+25-CA (build 13.0.4+8-MTS)
OpenJDK 64-Bit Server VM Zulu13.33+25-CA (build 13.0.4+8-MTS, mixed mode, sharing)
```

**OS type/version**

`Ubuntu 18.04.5 LTS`

**Description**

Hi, Jetty maintainers! 

We've recently attempted a migration from Jetty 9 to Jetty 10 and we've noticed a regression related to WebSockets. According to our metrics, there seems to be a memory leak which I'm still currently investigating and I hope to provide more information soon. However, it also seems that Jetty 10 spends a large amount of CPU resolving lambda forms to perform method handle invocations inside. On the flame graphs, we saw a large number of lambdaforms - every tiny green tower on the flame graph is a separate instance of LabmdaForm. My assumption which I'm currently investigating and hope to provide more data soon is that the large number of labmdaforms filled up the java heap. 

**CPU flame graph:** 
Screen Shot 2021-05-27 at 2 23 30 pm

**Allocation flame graph:**
Screen Shot 2021-05-27 at 2 24 17 pm

I was wondering if you've observed this behaviour before or might know what it could be caused by. I'm also currently investigating the issue and will provide more info once I have it. Thanks!

lachlan-roberts:
Fixes #6328

Use a new `JettyMethodHandle` interface so that implementations can be used which avoid binding the MethodHandles for each WebSocket connection. This will avoid the multiple spikes seen on the flamegraph from #6328.

For Javax endpoints we are creating new method handles each time because we cannot have the same MetaData cache as used in the `JettyWebSocketFrameHandlerFactory` because it depends on both the endpointClass and the endpointConfig. However `javax.websocket.MessageHandler` does not actually use method handles at all and has a special implementation of  `JettyMethodHandle`.

### DIFF
--- a/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/AbstractMessageSink.java
+++ b/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/AbstractMessageSink.java
@@ -13,20 +13,20 @@
 
 package org.eclipse.jetty.websocket.core.internal.messages;
 
-import java.lang.invoke.MethodHandle;
 import java.util.Objects;
 
 import org.eclipse.jetty.websocket.core.CoreSession;
+import org.eclipse.jetty.websocket.core.internal.util.MethodHolder;
 
 public abstract class AbstractMessageSink implements MessageSink
 {
     protected final CoreSession session;
-    protected final MethodHandle methodHandle;
+    protected final MethodHolder methodHolder;
 
-    public AbstractMessageSink(CoreSession session, MethodHandle methodHandle)
+    public AbstractMessageSink(CoreSession session, MethodHolder methodHolder)
     {
         this.session = Objects.requireNonNull(session, "CoreSession");
-        this.methodHandle = Objects.requireNonNull(methodHandle, "MethodHandle");
+        this.methodHolder = Objects.requireNonNull(methodHolder, "MethodHolder");
     }
 
     @Override
--- a/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ByteBufferMessageSink.java
+++ b/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ByteBufferMessageSink.java
@@ -13,10 +13,7 @@
 
 package org.eclipse.jetty.websocket.core.internal.messages;
 
-import java.lang.invoke.MethodHandle;
-import java.lang.invoke.MethodType;
 import java.nio.ByteBuffer;
-import java.util.Objects;
 
 import org.eclipse.jetty.io.ByteBufferCallbackAccumulator;
 import org.eclipse.jetty.io.ByteBufferPool;
@@ -24,24 +21,16 @@ import org.eclipse.jetty.util.BufferUtil;
 import org.eclipse.jetty.util.Callback;
 import org.eclipse.jetty.websocket.core.CoreSession;
 import org.eclipse.jetty.websocket.core.Frame;
-import org.eclipse.jetty.websocket.core.exception.InvalidSignatureException;
 import org.eclipse.jetty.websocket.core.exception.MessageTooLargeException;
+import org.eclipse.jetty.websocket.core.internal.util.MethodHolder;
 
 public class ByteBufferMessageSink extends AbstractMessageSink
 {
     private ByteBufferCallbackAccumulator out;
 
-    public ByteBufferMessageSink(CoreSession session, MethodHandle methodHandle)
+    public ByteBufferMessageSink(CoreSession session, MethodHolder methodHolder)
     {
-        super(session, methodHandle);
-
-        // Validate onMessageMethod
-        Objects.requireNonNull(methodHandle, "MethodHandle");
-        MethodType onMessageType = MethodType.methodType(Void.TYPE, ByteBuffer.class);
-        if (methodHandle.type() != onMessageType)
-        {
-            throw InvalidSignatureException.build(onMessageType, methodHandle.type());
-        }
+        super(session, methodHolder);
     }
 
     @Override
@@ -61,9 +50,9 @@ public class ByteBufferMessageSink extends AbstractMessageSink
             if (frame.isFin() && (out == null))
             {
                 if (frame.hasPayload())
-                    methodHandle.invoke(frame.getPayload());
+                    methodHolder.invoke(frame.getPayload());
                 else
-                    methodHandle.invoke(BufferUtil.EMPTY_BUFFER);
+                    methodHolder.invoke(BufferUtil.EMPTY_BUFFER);
 
                 callback.succeeded();
                 session.demand(1);
@@ -89,7 +78,7 @@ public class ByteBufferMessageSink extends AbstractMessageSink
 
                 try
                 {
-                    methodHandle.invoke(buffer);
+                    methodHolder.invoke(buffer);
                 }
                 finally
                 {
--- a/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/InputStreamMessageSink.java
+++ b/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/InputStreamMessageSink.java
@@ -13,16 +13,15 @@
 
 package org.eclipse.jetty.websocket.core.internal.messages;
 
-import java.lang.invoke.MethodHandle;
-
 import org.eclipse.jetty.websocket.core.CoreSession;
 import org.eclipse.jetty.websocket.core.Frame;
+import org.eclipse.jetty.websocket.core.internal.util.MethodHolder;
 
 public class InputStreamMessageSink extends DispatchedMessageSink
 {
-    public InputStreamMessageSink(CoreSession session, MethodHandle methodHandle)
+    public InputStreamMessageSink(CoreSession session, MethodHolder methodHolder)
     {
-        super(session, methodHandle);
+        super(session, methodHolder);
     }
 
     @Override
--- a/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ReaderMessageSink.java
+++ b/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ReaderMessageSink.java
@@ -13,16 +13,15 @@
 
 package org.eclipse.jetty.websocket.core.internal.messages;
 
-import java.lang.invoke.MethodHandle;
-
 import org.eclipse.jetty.websocket.core.CoreSession;
 import org.eclipse.jetty.websocket.core.Frame;
+import org.eclipse.jetty.websocket.core.internal.util.MethodHolder;
 
 public class ReaderMessageSink extends DispatchedMessageSink
 {
-    public ReaderMessageSink(CoreSession session, MethodHandle methodHandle)
+    public ReaderMessageSink(CoreSession session, MethodHolder methodHolder)
     {
-        super(session, methodHandle);
+        super(session, methodHolder);
     }
 
     @Override
--- a/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/StringMessageSink.java
+++ b/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/StringMessageSink.java
@@ -13,22 +13,21 @@
 
 package org.eclipse.jetty.websocket.core.internal.messages;
 
-import java.lang.invoke.MethodHandle;
-
 import org.eclipse.jetty.util.Callback;
 import org.eclipse.jetty.util.Utf8StringBuilder;
 import org.eclipse.jetty.websocket.core.CoreSession;
 import org.eclipse.jetty.websocket.core.Frame;
 import org.eclipse.jetty.websocket.core.exception.MessageTooLargeException;
+import org.eclipse.jetty.websocket.core.internal.util.MethodHolder;
 
 public class StringMessageSink extends AbstractMessageSink
 {
     private Utf8StringBuilder out;
     private int size;
 
-    public StringMessageSink(CoreSession session, MethodHandle methodHandle)
+    public StringMessageSink(CoreSession session, MethodHolder methodHolder)
     {
-        super(session, methodHandle);
+        super(session, methodHolder);
         this.size = 0;
     }
 
@@ -50,7 +49,7 @@ public class StringMessageSink extends AbstractMessageSink
 
             out.append(frame.getPayload());
             if (frame.isFin())
-                methodHandle.invoke(out.toString());
+                methodHolder.invoke(out.toString());
 
             callback.succeeded();
             session.demand(1);
--- a/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/MethodHolder.java
+++ b/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/MethodHolder.java
@@ -1,0 +1,107 @@
+//
+// ========================================================================
+// Copyright (c) 1995 Mort Bay Consulting Pty Ltd and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.websocket.core.internal.util;
+
+import java.lang.invoke.MethodHandle;
+
+/**
+ * An interface for managing invocations of methods whose arguments may need to be augmented, by
+ * binding in certain parameters ahead of time.
+ *
+ * Implementations may use various invocation mechanisms, including:
+ * <ul>
+ *  <li>direct method invocation on an held object</li>
+ *  <li>calling a method pointer</li>
+ *  <li>calling a MethodHandle bound to the known arguments</li>
+ *  <li>calling a MethodHandle without binding to the known arguments</li>
+ * </ul>
+ *
+ * Implementations of this may not be thread safe, so the caller must use some external mutual exclusion
+ * unless they are using a specific implementation known to be thread-safe.
+ */
+public interface MethodHolder
+{
+    String METHOD_HOLDER_BINDING_PROPERTY = "jetty.websocket.methodholder.binding";
+
+    static MethodHolder from(MethodHandle methodHandle)
+    {
+        String property = System.getProperty(METHOD_HOLDER_BINDING_PROPERTY);
+        return from(methodHandle, Boolean.parseBoolean(property));
+    }
+
+    static MethodHolder from(MethodHandle methodHandle, boolean binding)
+    {
+        if (methodHandle == null)
+            return null;
+        return binding ? new BindingMethodHolder(methodHandle) : new NonBindingMethodHolder(methodHandle);
+    }
+
+    Object invoke(Object... args) throws Throwable;
+
+    default MethodHolder bindTo(Object arg)
+    {
+        throw new UnsupportedOperationException();
+    }
+
+    default MethodHolder bindTo(Object arg, int idx)
+    {
+        throw new UnsupportedOperationException();
+    }
+
+    default Class<?> parameterType(int idx)
+    {
+        throw new UnsupportedOperationException();
+    }
+
+    default Class<?> returnType()
+    {
+        throw new UnsupportedOperationException();
+    }
+
+    static Object doInvoke(MethodHandle methodHandle, Object arg1, Object arg2) throws Throwable
+    {
+        return methodHandle.invoke(arg1, arg2);
+    }
+
+    static Object doInvoke(MethodHandle methodHandle, Object... args) throws Throwable
+    {
+        return methodHandle.invokeExact(args[0], args[1]);
+
+//        switch (args.length)
+//        {
+//            case 0:
+//                return methodHandle.invoke();
+//            case 1:
+//                return methodHandle.invoke(args[0]);
+//            case 2:
+//                return methodHandle.invoke(args[0], args[1]);
+//            case 3:
+//                return methodHandle.invoke(args[0], args[1], args[2]);
+//            case 4:
+//                return methodHandle.invoke(args[0], args[1], args[2], args[3]);
+//            case 5:
+//                return methodHandle.invoke(args[0], args[1], args[2], args[3], args[4]);
+//            case 6:
+//                return methodHandle.invoke(args[0], args[1], args[2], args[3], args[4], args[5]);
+//            case 7:
+//                return methodHandle.invoke(args[0], args[1], args[2], args[3], args[4], args[5], args[6]);
+//            case 8:
+//                return methodHandle.invoke(args[0], args[1], args[2], args[3], args[4], args[5], args[6], args[7]);
+//            case 9:
+//                return methodHandle.invoke(args[0], args[1], args[2], args[3], args[4], args[5], args[6], args[7], args[8]);
+//            default:
+//                return methodHandle.invokeWithArguments(args);
+//        }
+    }
+}
--- a/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingMethodHolder.java
+++ b/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingMethodHolder.java
@@ -1,0 +1,103 @@
+//
+// ========================================================================
+// Copyright (c) 1995 Mort Bay Consulting Pty Ltd and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.websocket.core.internal.util;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.WrongMethodTypeException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * This implementation of {@link MethodHolder} is not thread safe.
+ * Mutual exclusion should be used when calling {@link #invoke(Object...)}, or this should only
+ * be invoked from a single thread.
+ */
+class NonBindingMethodHolder implements MethodHolder
+{
+    private final MethodHandle _methodHandle;
+    private final Object[] _parameters;
+    private final List<Integer> _unboundParamIndexes = new ArrayList<>();
+
+    public NonBindingMethodHolder(MethodHandle methodHandle)
+    {
+        _methodHandle = Objects.requireNonNull(methodHandle);
+        int numParams = methodHandle.type().parameterCount();
+        _parameters = new Object[numParams];
+        for (int i = 0; i < numParams; i++)
+        {
+            _unboundParamIndexes.add(i);
+        }
+    }
+
+    @Override
+    public Object invoke(Object... args) throws Throwable
+    {
+        try
+        {
+            insertArguments(args);
+            return MethodHolder.doInvoke(_methodHandle, _parameters);
+        }
+        finally
+        {
+            clearArguments();
+        }
+    }
+
+    @Override
+    public MethodHolder bindTo(Object arg, int idx)
+    {
+        _parameters[_unboundParamIndexes.get(idx)] = arg;
+        _unboundParamIndexes.remove(idx);
+        return this;
+    }
+
+    @Override
+    public MethodHolder bindTo(Object arg)
+    {
+        return bindTo(arg, 0);
+    }
+
+    private void insertArguments(Object... args)
+    {
+        if (_unboundParamIndexes.size() != args.length)
+            throw new WrongMethodTypeException(String.format("Expected %s params but had %s", _unboundParamIndexes.size(), args.length));
+
+        int argsIndex = 0;
+        for (int index : _unboundParamIndexes)
+        {
+            _parameters[index] = args[argsIndex++];
+        }
+    }
+
+    private void clearArguments()
+    {
+        for (int i : _unboundParamIndexes)
+        {
+            _parameters[i] = null;
+        }
+    }
+
+    @Override
+    public Class<?> parameterType(int idx)
+    {
+        return _methodHandle.type().parameterType(_unboundParamIndexes.get(idx));
+    }
+
+    @Override
+    public Class<?> returnType()
+    {
+        return _methodHandle.type().returnType();
+    }
+}
--- a/jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/JavaxMessagePartialMethodHolder.java
+++ b/jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/JavaxMessagePartialMethodHolder.java
@@ -1,0 +1,59 @@
+//
+// ========================================================================
+// Copyright (c) 1995 Mort Bay Consulting Pty Ltd and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.websocket.javax.common;
+
+import java.lang.invoke.WrongMethodTypeException;
+import javax.websocket.MessageHandler;
+
+import org.eclipse.jetty.websocket.core.internal.util.MethodHolder;
+
+class JavaxMessagePartialMethodHolder<T> implements MethodHolder
+{
+    private final MessageHandler.Partial<T> _messageHandler;
+
+    public JavaxMessagePartialMethodHolder(MessageHandler.Partial<T> messageHandler)
+    {
+        _messageHandler = messageHandler;
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public Object invoke(Object... args) throws Throwable
+    {
+        if (args.length != 2)
+            throw new WrongMethodTypeException(String.format("Expected %s params but had %s", 2, args.length));
+        _messageHandler.onMessage((T)args[0], (boolean)args[1]);
+        return null;
+    }
+
+    @Override
+    public Class<?> parameterType(int idx)
+    {
+        switch (idx)
+        {
+            case 0:
+                return Object.class;
+            case 1:
+                return boolean.class;
+            default:
+                throw new IndexOutOfBoundsException(idx);
+        }
+    }
+
+    @Override
+    public Class<?> returnType()
+    {
+        return void.class;
+    }
+}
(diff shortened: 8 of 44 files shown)


jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/JettyMethodHandle.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.lang.invoke.MethodHandle;
+
+/**
+ * Jetty interface for managing {@link MethodHandle}s.
+ *
+ * This differs from {@link MethodHandle}s that any calls the methods such as {@link #bindTo(Object)}
+ * will change the instance of {@link JettyMethodHandle} it is called on.
+ */
+public interface JettyMethodHandle
```
gregw: I don't like the name `JettyMethodHandle` as it is just using `Jetty` as a meaningless prefix.

Something like `MethodHolder` would be better.  We have lots of `Holder` classes in jetty, so the concept is well known... it's a class that holds a reference to another class and provides some services for that.  Specially as it appears that some impls don't even use `MethodHandle`, so that should not be in the name.

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/JavaxMessagePartialMethodHandle.java
```diff
@@ -0,0 +1,84 @@
...
+
+package org.eclipse.jetty.websocket.javax.common;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.WrongMethodTypeException;
+import javax.websocket.MessageHandler;
+
+import org.eclipse.jetty.websocket.core.internal.util.JettyMethodHandle;
+
+class JavaxMessagePartialMethodHandle<T> implements JettyMethodHandle
```
gregw: Again the naming would be better if it was:
```suggestion
class JavaxMessagePartialMethodHolder<T> implements MethodHolder
```

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingJettyMethodHandle.java
```diff
@@ -0,0 +1,133 @@
...
+// ========================================================================
+//
+
+package org.eclipse.jetty.websocket.core.internal.util;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.WrongMethodTypeException;
+import java.util.Objects;
+
+class NonBindingJettyMethodHandle implements JettyMethodHandle
```
gregw:
```suggestion
class NonBindingMethodHandleHolder implements MethodHolder
```

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingJettyMethodHandle.java
```diff
@@ -0,0 +1,133 @@
...
+        int index = 0;
+        int argsIndex = 0;
+        for (int i = 0; i < _numParams; i++)
+        {
+            if (_boundParams[i] != Boolean.TRUE)
+            {
+                if (index >= idx && argsIndex < args.length)
+                {
+                    Object val = args[argsIndex++];
+                    _parameters[i] = val;
```
gregw: So you need javadoc in this impl saying that this is not thread safe.  
You also need javadoc in the interface saying that some implementations may not be thread safe and that unless the caller knows the impls are, then external mutual exclusion is needed.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingJettyMethodHandle.java
```diff
@@ -0,0 +1,133 @@
...
+    {
+        insertArguments(false, 0, args);
+        Object o = _methodHandle.invokeWithArguments(_parameters);
+        if (_returnFilter != null)
+            o = _returnFilter.invoke(o);
+        return (_returnType == null) ? o : _returnType.cast(o);
+    }
+
+    @Override
+    public JettyMethodHandle bindTo(Object arg, int idx)
```
gregw: Is this style of binding to a specific index really necessary?   MethodHandle doesn't support that and you can only bind from the first args.

If we supported the same semantic, then you'd only need a single index to track the binding (rather than a boolean array) and you could use efficient arraycopy operations to move args into the array on every invoke rather than a loop with conditional tests over two arrays!

gregw (changes requested):
Pretty good.... but naming needs improvement.
Also I question the need for sparse binding of args.  It would be more efficient to only support the same binding style as MethodHandle

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/messages/DecodedTextStreamMessageSink.java
```diff
@@ -26,22 +25,32 @@
...
 import org.eclipse.jetty.websocket.core.internal.messages.ReaderMessageSink;
-import org.eclipse.jetty.websocket.javax.common.JavaxWebSocketFrameHandlerFactory;
+import org.eclipse.jetty.websocket.core.internal.util.AbstractJettyMethodHandle;
+import org.eclipse.jetty.websocket.core.internal.util.JettyMethodHandle;
 import org.eclipse.jetty.websocket.javax.common.decoders.RegisteredDecoder;
 
 public class DecodedTextStreamMessageSink<T> extends AbstractDecodedMessageSink.Stream<Decoder.TextStream<T>>
 {
-    public DecodedTextStreamMessageSink(CoreSession session, MethodHandle methodHandle, List<RegisteredDecoder> decoders)
+    public DecodedTextStreamMessageSink(CoreSession session, JettyMethodHandle methodHandle, List<RegisteredDecoder> decoders)
```
gregw: for the sake of clarity, I think a big renaming of args and fields from `methodHandle` to `methodHolder` would be useful

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/AbstractJettyMethodHandle.java
```diff
@@ -0,0 +1,55 @@
...
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.websocket.core.internal.util;
+
+import java.lang.invoke.MethodHandle;
+
+public abstract class AbstractJettyMethodHandle implements JettyMethodHandle
```
gregw: Does this class provide any value that you couldn't get with default impls in the interace?

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/BindingJettyMethodHandle.java
```diff
@@ -0,0 +1,73 @@
...
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.websocket.core.internal.util;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+
+class BindingJettyMethodHandle implements JettyMethodHandle
```
gregw:
```suggestion
class MethodHandleHolder implements MethodHolder
```

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/JettyMethodHandle.java
lachlan-roberts: Renamed to `MethodHolder`.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingJettyMethodHandle.java
lachlan-roberts: I have added some Javadoc to both this class and the interface.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingJettyMethodHandle.java
lachlan-roberts: `MethodHandles` allow you to do this with `insertArguments` and we are using this for binding the javax websocket path param values.

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/messages/DecodedTextStreamMessageSink.java
lachlan-roberts: done

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/AbstractJettyMethodHandle.java
lachlan-roberts: I have removed this abstract class and put default methods in the interface.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/AbstractMessageSink.java
```diff
@@ -13,19 +13,19 @@
...
     protected final CoreSession session;
-    protected final MethodHandle methodHandle;
+    protected final MethodHolder methodHolder;
 
-    public AbstractMessageSink(CoreSession session, MethodHandle methodHandle)
+    public AbstractMessageSink(CoreSession session, MethodHolder methodHolder)
     {
         this.session = Objects.requireNonNull(session, "CoreSession");
-        this.methodHandle = Objects.requireNonNull(methodHandle, "MethodHandle");
+        this.methodHolder = Objects.requireNonNull(methodHolder, "MethodHandle");
```
sbordet: The string should be `"MethodHolder"`.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/MethodHolder.java
```diff
@@ -0,0 +1,73 @@
...
+ * will change the instance of {@link MethodHolder} it is called on.
+ *
+ * Implementations of this may not be thread safe, so the caller must use some external mutual exclusion
+ * unless they are using a specific implementation known to be thread-safe.
+ */
+public interface MethodHolder
+{
+    static MethodHolder from(MethodHandle methodHandle)
+    {
+        return from(methodHandle, false);
```
gregw: Whilst I really don't like System properties, I think in this case it is worthwhile having one that will allow us to switch this from false to true.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/InputStreamMessageSink.java
```diff
@@ -17,12 +17,19 @@
...
 
 public class InputStreamMessageSink extends DispatchedMessageSink
 {
+    public InputStreamMessageSink(CoreSession session, MethodHolder methodHolder)
+    {
+        super(session, methodHolder);
+    }
+
+    @Deprecated
     public InputStreamMessageSink(CoreSession session, MethodHandle methodHandle)
```
sbordet: Why did you keep this constructor, while in all other `Sink`s it was removed (to only have the `MethodHolder` constructor)?
It's an internal class so we can break compatibility.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ReaderMessageSink.java
```diff
@@ -17,12 +17,19 @@
...
 
 public class ReaderMessageSink extends DispatchedMessageSink
 {
+    public ReaderMessageSink(CoreSession session, MethodHolder methodHolder)
+    {
+        super(session, methodHolder);
+    }
+
+    @Deprecated
     public ReaderMessageSink(CoreSession session, MethodHandle methodHandle)
```
sbordet: Ditto.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/StringMessageSink.java
```diff
@@ -20,18 +20,25 @@
...
-    public StringMessageSink(CoreSession session, MethodHandle methodHandle)
+    public StringMessageSink(CoreSession session, MethodHolder methodHolder)
     {
-        super(session, methodHandle);
+        super(session, methodHolder);
         this.size = 0;
     }
 
+    @Deprecated
+    public StringMessageSink(CoreSession session, MethodHandle methodHandle)
```
sbordet: Ditto.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/MethodHolder.java
```diff
@@ -0,0 +1,73 @@
...
+package org.eclipse.jetty.websocket.core.internal.util;
+
+import java.lang.invoke.MethodHandle;
+
+/**
+ * An interface for managing invocations of methods whose arguments may need to be augmented, by
+ * binding in certain parameters ahead of time.
+ *
+ * This differs from {@link MethodHandle}s that any calls the methods such as {@link #bindTo(Object)}
+ * will change the instance of {@link MethodHolder} it is called on.
```
gregw:
```suggestion
 * Implementations may use various invocation mechanisms, including:
  * <ul>
  *  <li>direct method invocation on an held object</li>
  *  <li>calling a method pointer</li>
  *  <li>calling a MethodHandle bound to the known arguments</li>
  *  <li>calling a MethodHandle without binding to the known arguments</li>
  * </ul>
```

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/JavaxMessagePartialMethodHolder.java
```diff
@@ -0,0 +1,84 @@
...
+    public Object invoke(Object... args) throws Throwable
+    {
+        if (args.length != 2)
+            throw new WrongMethodTypeException(String.format("Expected %s params but had %s", 2, args.length));
+        _messageHandler.onMessage((T)args[0], (boolean)args[1]);
+        return null;
+    }
+
+    @Override
+    public MethodHolder bindTo(Object arg)
```
sbordet: Aren't all these methods that throw UOE `default` in the interface?

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/JavaxMessageWholeMethodHolder.java
```diff
@@ -0,0 +1,82 @@
...
+    public Object invoke(Object... args) throws Throwable
+    {
+        if (args.length != 1)
+            throw new WrongMethodTypeException(String.format("Expected %s params but had %s", 1, args.length));
+        _messageHandler.onMessage((T)args[0]);
+        return null;
+    }
+
+    @Override
+    public MethodHolder bindTo(Object arg)
```
sbordet: Ditto.

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/JavaxWebSocketFrameHandler.java
```diff
@@ -81,10 +80,10 @@ public class JavaxWebSocketFrameHandler implements FrameHandler
     public JavaxWebSocketFrameHandler(JavaxWebSocketContainer container,
                                       UpgradeRequest upgradeRequest,
                                       Object endpointInstance,
-                                      MethodHandle openHandle, MethodHandle closeHandle, MethodHandle errorHandle,
+                                      MethodHolder openHandle, MethodHolder closeHandle, MethodHolder errorHandle,
                                       JavaxWebSocketMessageMetadata textMetadata,
                                       JavaxWebSocketMessageMetadata binaryMetadata,
-                                      MethodHandle pongHandle,
+                                      MethodHolder pongHandle,
                                       EndpointConfig endpointConfig)
```
sbordet, line 88: This class is not internal and has a signature change. Is that a problem? If not, can be made internal?

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingMethodHolder.java
```diff
@@ -0,0 +1,137 @@
...
+            throw new WrongMethodTypeException(String.format("Expected %s params but had %s", args.length, argsIndex + 1));
+        if (index != argsIndex)
+            throw new WrongMethodTypeException(String.format("Expected %s params but had %s", index + 1, argsIndex + 1));
+        return this;
+    }
+
+    @Override
+    public MethodHolder filterReturnValue(MethodHandle filter)
+    {
+        _returnFilter = filter;
```
gregw: This is not correct.  With `BindingMethodHolder` and `MethodHandler`, this method can be called multiple times to do multiple mutations of the return value.   So you need to check the _returnFilter and if non null set it to a lambda that invokes first the existing filter then the passed filter.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingMethodHolder.java
```diff
@@ -0,0 +1,137 @@
...
+        return this;
+    }
+
+    @Override
+    public MethodHolder changeReturnType(Class<Object> objectClass)
+    {
+        _returnType = objectClass;
```
gregw: This is also not correct.   With `MethodHandler` you can interleave multiple calls to filterReturnValue and changeReturnType and they will all apply.    You need to implement this as a filter, so it can be done in order with the other filters.

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/JavaxWebSocketFrameHandlerFactory.java
```diff
@@ -219,21 +220,21 @@ public static MessageSink createMessageSink(JavaxWebSocketSession session, Javax
         }
     }
 
-    public static MethodHandle wrapNonVoidReturnType(MethodHandle handle, JavaxWebSocketSession session)
+    public static MethodHolder wrapNonVoidReturnType(MethodHolder handle, JavaxWebSocketSession session)
```
sbordet: Make it non-`public`, as it has a signature change.

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/JavaxWebSocketMessageMetadata.java
```diff
@@ -58,14 +58,14 @@ public void setMaxMessageSize(int maxMessageSize)
...
 
-    public MethodHandle getMethodHandle()
+    public MethodHolder getMethodHolder()
     {
-        return methodHandle;
+        return methodHolder;
     }
 
-    public void setMethodHandle(MethodHandle methodHandle)
+    public void setMethodHolder(MethodHolder methodHolder)
```
sbordet, line 66: This setter and the getter have signature changes. Can this class be made internal?

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingMethodHolder.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        throw new IndexOutOfBoundsException(idx);
+    }
+
+    @Override
+    public Object invoke(Object... args) throws Throwable
+    {
+        insertArguments(false, 0, args);
+        Object o = _methodHandle.invokeWithArguments(_parameters);
```
gregw: You need to null out the non bound arguments after the invoke so they are not held in memory

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/messages/AbstractDecodedMessageSink.java
```diff
@@ -32,12 +32,12 @@ public abstract class AbstractDecodedMessageSink implements MessageSink
 {
     private static final Logger LOG = LoggerFactory.getLogger(AbstractDecodedMessageSink.class);
 
-    private final MethodHandle _methodHandle;
+    private final MethodHolder _methodHolder;
     private final MessageSink _messageSink;
 
-    public AbstractDecodedMessageSink(CoreSession coreSession, MethodHandle methodHandle)
+    public AbstractDecodedMessageSink(CoreSession coreSession, MethodHolder methodHolder)
```
sbordet, line 38: Signature change of public class. Internal or keep previous signature deprecated?

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/messages/DecodedBinaryMessageSink.java
```diff
@@ -34,18 +33,23 @@
 {
     private static final Logger LOG = LoggerFactory.getLogger(DecodedBinaryMessageSink.class);
 
-    public DecodedBinaryMessageSink(CoreSession session, MethodHandle methodHandle, List<RegisteredDecoder> decoders)
+    public DecodedBinaryMessageSink(CoreSession session, MethodHolder methodHolder, List<RegisteredDecoder> decoders)
```
sbordet, line 36: Ditto.

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/messages/DecodedBinaryStreamMessageSink.java
```diff
@@ -26,23 +25,28 @@
...
 import org.eclipse.jetty.websocket.core.internal.messages.InputStreamMessageSink;
 import org.eclipse.jetty.websocket.core.internal.messages.MessageSink;
-import org.eclipse.jetty.websocket.javax.common.JavaxWebSocketFrameHandlerFactory;
+import org.eclipse.jetty.websocket.core.internal.util.MethodHolder;
 import org.eclipse.jetty.websocket.javax.common.decoders.RegisteredDecoder;
 
 public class DecodedBinaryStreamMessageSink<T> extends AbstractDecodedMessageSink.Stream<Decoder.BinaryStream<T>>
 {
-    public DecodedBinaryStreamMessageSink(CoreSession session, MethodHandle methodHandle, List<RegisteredDecoder> decoders)
+    public DecodedBinaryStreamMessageSink(CoreSession session, MethodHolder methodHolder, List<RegisteredDecoder> decoders)
```
sbordet, line 33: Ditto.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingMethodHolder.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    @Override
+    public MethodHolder bindTo(Object arg)
+    {
+        return bindTo(arg, 0);
+    }
+
+    @SuppressWarnings({"UnusedReturnValue", "SameParameterValue"})
+    private MethodHolder insertArguments(boolean bind, int idx, Object... args)
```
gregw: Is this frequently called without any binding at all?  If so, it would be worthwhile having a short cut where the args themselves are used.

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/messages/DecodedTextMessageSink.java
```diff
@@ -33,18 +32,23 @@
 {
     private static final Logger LOG = LoggerFactory.getLogger(DecodedTextMessageSink.class);
 
-    public DecodedTextMessageSink(CoreSession session, MethodHandle methodHandle, List<RegisteredDecoder> decoders)
+    public DecodedTextMessageSink(CoreSession session, MethodHolder methodHolder, List<RegisteredDecoder> decoders)
```
sbordet, line 35: Ditto.

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/messages/DecodedTextStreamMessageSink.java
```diff
@@ -26,23 +25,28 @@
...
 import org.eclipse.jetty.websocket.core.internal.messages.MessageSink;
 import org.eclipse.jetty.websocket.core.internal.messages.ReaderMessageSink;
-import org.eclipse.jetty.websocket.javax.common.JavaxWebSocketFrameHandlerFactory;
+import org.eclipse.jetty.websocket.core.internal.util.MethodHolder;
 import org.eclipse.jetty.websocket.javax.common.decoders.RegisteredDecoder;
 
 public class DecodedTextStreamMessageSink<T> extends AbstractDecodedMessageSink.Stream<Decoder.TextStream<T>>
 {
-    public DecodedTextStreamMessageSink(CoreSession session, MethodHandle methodHandle, List<RegisteredDecoder> decoders)
+    public DecodedTextStreamMessageSink(CoreSession session, MethodHolder methodHolder, List<RegisteredDecoder> decoders)
```
sbordet, line 33: Ditto.

sbordet (changes requested):

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingMethodHolder.java
```diff
@@ -0,0 +1,137 @@
...
+                    _parameters[i] = val;
+                    _boundParams[i] = bind;
+                }
+                else
+                {
+                    _parameters[i] = null;
+                }
+                index++;
+            }
+        }
```
gregw: I'm not a hug fan of this algorithm.   I think I would prefer a lookup table that mapped the index of the passed args to the parameters index:
```suggestion
        for (int i = 0; i < args.length; i++)
          _parameter[_lookup[i]] = args[i];
```

It's a little bit more work to create the lookup table, but much simpler to invoke.  Also really simple to null out afterwards:
```java
  for (int i : _lookup)
    _parameter[i] = null;
```

gregw (changes requested):
Getting there...

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/JavaxWebSocketFrameHandler.java
lachlan-roberts: Anything in the websocket common modules are not part of the API and considered internal implementation.

except for websocket-core-common

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingMethodHolder.java
```diff
@@ -0,0 +1,136 @@
...
+    {
+        if (index < 0 || index >= _lookupTable.size())
+            throw new IndexOutOfBoundsException();
+        _lookupTable.remove(index);
+    }
+
+    private int getInternalIndex(int index)
+    {
+        if (index < 0 || index >= _lookupTable.size())
+            throw new IndexOutOfBoundsException();
```
gregw: Doesn't get below do this range check?

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingMethodHolder.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    @Override
+    public Object invoke(Object... args) throws Throwable
+    {
+        try
+        {
+            insertArguments(args);
+            Object o = _methodHandle.invokeWithArguments(_parameters);
+            if (_returnFilters != null)
```
gregw: Hmmm maybe better to initialize to an empty list to avoid this if.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingMethodHolder.java
```diff
@@ -0,0 +1,136 @@
...
+    @SuppressWarnings({"UnusedReturnValue", "SameParameterValue"})
+    private MethodHolder insertArguments(Object... args)
+    {
+        if (_lookupTable.size() != args.length)
+            throw new WrongMethodTypeException(String.format("Expected %s params but had %s", _lookupTable.size(), args.length));
+
+        for (int i = 0; i < args.length; i++)
+        {
+            _parameters[_lookupTable.get(i)] = args[i];
+        }
```
gregw:
```suggestion
        int from = 0;
        for (int to : _lookupTable)
            _parameters[to] = args[from++];
```

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingMethodHolder.java
```diff
@@ -0,0 +1,136 @@
...
+        for (int i = 0; i < numParams; i++)
+        {
+            _lookupTable.add(i);
+        }
+    }
+
+    private void dropFromLookupTable(int index)
+    {
+        if (index < 0 || index >= _lookupTable.size())
+            throw new IndexOutOfBoundsException();
```
gregw: doesn't remove do this bounds check?  Even if it doesn't, is it really needed?

gregw (changes requested):
I'm down to tiny quibbles now.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/MethodHolder.java
```diff
@@ -19,17 +19,25 @@
...
+ *  <li>calling a MethodHandle bound to the known arguments</li>
+ *  <li>calling a MethodHandle without binding to the known arguments</li>
+ * </ul>
  *
  * Implementations of this may not be thread safe, so the caller must use some external mutual exclusion
  * unless they are using a specific implementation known to be thread-safe.
  */
 public interface MethodHolder
 {
+    String METHOD_HOLDER_BINDING_PROPERTY = "jetty.methodholder.binding";
```
sbordet: The property name should have "websocket" in it as this only applied to websocket.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ByteBufferMessageSink.java
```diff
@@ -14,34 +14,30 @@
...
 
-        // Validate onMessageMethod
-        Objects.requireNonNull(methodHandle, "MethodHandle");
-        MethodType onMessageType = MethodType.methodType(Void.TYPE, ByteBuffer.class);
-        if (methodHandle.type() != onMessageType)
-        {
-            throw InvalidSignatureException.build(onMessageType, methodHandle.type());
-        }
+    @Deprecated
+    public ByteBufferMessageSink(CoreSession session, MethodHandle methodHandle)
```
sbordet: Remove this constructor (as you have removed all similar from all the other Sinks).

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingMethodHolder.java
```diff
@@ -25,51 +28,61 @@ class NonBindingMethodHolder implements MethodHolder
 {
```
sbordet, line 28: The javadocs of this class is incorrect. No mutual exclusion is ever used to call `invoke()`.
It relies on the fact that reads (and the processing of those reads) are naturally single threaded in the current WebSocket implementation.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingMethodHolder.java
```diff
@@ -25,51 +28,61 @@ class NonBindingMethodHolder implements MethodHolder
...
     private final MethodHandle _methodHandle;
     private final Object[] _parameters;
-    private final Boolean[] _boundParams;
-    private final int _numParams;
-    private Class<Object> _returnType;
-    private MethodHandle _returnFilter;
+    private final List<Integer> _lookupTable = new LinkedList<>();
```
sbordet: Can this be called `_unboundParameterIndexes` instead?

sbordet (changes requested):

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/util/NonBindingMethodHolder.java
gregw: @sbordet "naturally single threaded" is mutual exclusion

jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/JavaxWebSocketMessageMetadata.java
lachlan-roberts: It is internal, anything in javax-websocket-common is internal.

sbordet (dismissed):

joakime:
Do we want this in Jetty 10.0.10 release?

lachlan-roberts:
@joakime no this does not need to be in the `10.0.10` release.

I think it also needs some performance testing to assess the impact of this.

gregw (dismissed):

joakime:
Skipping this release, moving to 10.0.12 / 11.0.12 release.

joakime:
@lachlan-roberts you can merge this one now.

joakime:
@lachlan-roberts can this be merged?

lachlan-roberts:
@joakime no, the benchmarks indicated that using the wrapper around methodhandle is significantly slower. So we need to further assess the performance impact this change would have.

I don't think this should be classified as a bug, the current usage of MethodHandles is not wrong.

joakime:
Moving this to 10.0.14 then.

janbartel:
@lachlan-roberts  what's the status of this one? Should it be moved  back to Draft status?

lachlan-roberts:
@janbartel yes I converted it back to a draft.

Maybe this PR is something we want to consider for Jetty 12.

gregw:
@lachlan-roberts  I've asked the OP what their status is.
Can you update the branch against latest 10.0.x and repeat performance tests to check impact.

I would think this is something we want to do for 10/11/12 or not at all.   If there is a net benefit, then best to keep the code similar in all branches.

lachlan-roberts:
closing in favor of #10750 for Jetty 12

PR #8089, opened 2022-06-01
Issue #8088 Add STOP.EXIT System property to configure ShutdownMonitor.exitVm

### LINKED ISSUE #8088: Add option to configure exitVm on ShutdownMonitor from System properties
**9.4+**

Add a System property like `STOP.EXIT=[true|false]` to enable configuration of ShutdownMonitor `exitVm` from the command line.

janbartel:
Closes #8088

### DIFF
--- a/jetty-server/src/test/java/org/eclipse/jetty/server/ShutdownMonitorTest.java
+++ b/jetty-server/src/test/java/org/eclipse/jetty/server/ShutdownMonitorTest.java
@@ -152,15 +152,6 @@ public class ShutdownMonitorTest
             assertTrue(!ShutdownMonitor.isRegistered(server));
         }
     }
-    
-    @Test
-    public void testExitVmDefault() throws Exception
-    {
-        //Test that the default is to exit
-        ShutdownMonitor monitor = ShutdownMonitor.getInstance();
-        monitor.setPort(0);
-        assertTrue(monitor.isExitVm());
-    }
 
     /*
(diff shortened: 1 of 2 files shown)


janbartel:
@gregw note that even with the new `STOP.EXIT=false` System property - which prevents calling `System.exit()` -  the vm will still exit after the server is stopped. Probably  this is because there are only daemon threads left. Pretty sure we used to call `server.join()` somewhere to prevent that happening, but it doesn't seem to be the case any more.

joakime (dismissed):

olamy (dismissed):

olamy (dismissed):

jetty-server/src/test/java/org/eclipse/jetty/server/ShutdownMonitorTest.java
```diff
@@ -153,6 +153,48 @@ public void testNoExitSystemProperty() throws Exception
...
+    /*
+     * Disable these config tests because ShutdownMonitor is a 
+     * static singleton that cannot be unset, and thus would
+     * need each of these methods executed it its own jvm -
+     * current surefire settings only fork for a single test 
+     * class.
+     * 
+     * Undisable to test individually as needed.
+     */
+    @Disabled
```
joakime, line 165: Hmm.  Perhaps these exit VM tests should be in test-distribution ?

gregw (approved):

PR #8090, opened 2022-06-01
Jetty 12.0.x resource handler and default servlet

lorban:
WIP for resource handler and default servlet using the existing Resource API.


jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextHandler.java
```diff
@@ -873,7 +873,7 @@ public Resource getResource(String pathInContext) throws MalformedURLException
             // addPath with accept non-canonical paths that don't go above the root,
             // but will treat them as aliases. So unless allowed by an AliasChecker
             // they will be rejected below.
-            Resource resource = baseResource.addPath(pathInContext);
+            Resource resource = baseResource.getResource(pathInContext);
```
gregw, line 883: Isn't this now wrong?   The javadoc for getResource says that pathInContext is decoded, but we want it to be the canonically encoded path now.   Ie it must use the URI API to resolve, not string nor Path addition.

gregw (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -869,4 +832,44 @@ protected void putHeaders(HttpServletResponse response, HttpContent content, lon
          */
         String getWelcomeFile(String pathInContext) throws IOException;
     }
+
+    public interface GenericRequest
```
gregw, line 836: The core Request/Response APIs are not that fat and they are already interfaces.  So why not write the service in terms of core Request/Response and then only wrap the servlet Request/Response to be a core Request/Response.  We will almost certainly need such wrappers in future for cross context dispatch.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -13,8 +13,357 @@
...
+        @Override
+        public HttpURI getHttpURI()
+        {
+            return HttpURI.from(request.getRequestURI());
+        }
+
+        @Override
+        public String getPathInContext()
+        {
+            return request.getRequestURI();
```
gregw, line 163: I don't think this is correct.   It includes the context name  and it is fully encoded rather than canonically encoded.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -13,8 +13,357 @@
...
+        @Override
+        public long getHeaderDate(String name)
+        {
+            return request.getDateHeader(name);
+        }
+
+        @Override
+        public HttpURI getHttpURI()
+        {
+            return HttpURI.from(request.getRequestURI());
```
gregw, line 157: It would be good to have the optimization that if the request is not wrapped then we should be able to directly access the core values.  We should need to recreate them only on the minor case of a wrapped request.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
gregw: Also, should the include path attributes be resolved here?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -14,241 +14,168 @@
...
+    public static Resource getDefaultStylesheet()
+    {
+        // TODO the returned path should point to the classpath.
+        // This points to a non-existent file '/jetty-dir.css'.
+        return new PathResource(Path.of("/jetty-dir.css"));
+    }
 
-        boolean endsWithSlash = (pathInfo == null ? (_pathInfoOnly ? "" : servletPath) : pathInfo).endsWith(URIUtil.SLASH);
-        boolean checkPrecompressedVariants = _precompressedFormats.length > 0 && !endsWithSlash && !included && reqRanges == null;
+    public void doGet(GenericRequest request, GenericResponse response, Callback callback, HttpContent content) throws Exception
```
gregw, line 136: It feels wrong that this API passes in the content.  Surely the service has to be able to make content from a pathInContext itself (eg for welcome files)?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextHandler.java
gregw: I checked and this is using Path APIs for the add rather than URI apis.

gregw (changes requested):

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextHandler.java
gregw: This needs to be an efficient version of:
```suggestion
            Resource resource = Resource.newResource(_factory.getResource("/").getURI().resolve(pathInContext));
```

Actually, this is still wrong, because if the base resource is a ResourceCollection then the `getURI()` will collapse the "wave function" to a single URI.

I think we need a new API on `Resource` to do a URI resolve.... in fact why not call it `Resource.resolve(String)` and have it do the URI magic internally.
For a ResourceCollection, it will need to have an internal private list of URIs for each base, which it can resovlve against, convert to a Path and check for existence.  If it is a directory, return a new ResourceCollection, else return a PathResource

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/URLResource.java
```diff
@@ -252,16 +252,6 @@ public boolean delete()
         throw new SecurityException("Delete not supported");
```
janbartel, line 252: Do we still need URLResource? Aren't we doing everything with URIs and Paths?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -55,12 +55,11 @@
  * a file, a URL or an entry in a jar file.
  * </p>
  */
-//TODO remove
+// TODO Resource API should be cleaned up and simplified
```
janbartel, line 58: We might not need the `defaultUseCaches` settings any more if we're not using URLs?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactory.java
```diff
@@ -18,7 +18,6 @@
 /**
  * ResourceFactory.
  */
-// TODO remove
 public interface ResourceFactory
```
janbartel, line 21: Is this still needed?

janbartel:
@lorban this one needs a merge from 12.0.x.

Also, if you have addressed Greg's comments, could you mark them as resolved?

lorban:
Superseded by https://github.com/eclipse/jetty.project/pull/8276/

PR #8097, opened 2022-06-02
Rename demo-eeX to eeX-demo

gregw:


sbordet:
@gregw I thought we decided to stick with Maven conventions, so:

```
jetty-ee10/
  jetty-ee10-demos/
    jetty-ee10-demos-async-rest
```

That is, we always prefix the innermost module/directory with the outermost module/directory.

Another example was http2, which we wanted to rename from: `jetty-http2/http2-server` to `jetty-http2/jetty-http2-server`.

gregw:
@janbartel nudge

gregw:
@sbordet Sorry I missed you comment/question.

This PR was not really about the prefix, rather the order of the elements within that prefix.     I agree that there is no good pattern to why we sometimes drop the "jetty-" in a subdirectory, but I think that predated jetty-12.
Directories that used to be "demo/demo-async-rest" were moved to "jetty-eeX/jetty-ee10-demos/demo-ee10-async-rest".    This PR fixed the order to "jetty-eeX/jetty-ee10-demos/ee10-demo-async-rest", but didn't re-add the prefix "jetty-"... which maybe should also be done?

sbordet:
> but didn't re-add the prefix "jetty-"... which maybe should also be done?

I think jetty-12 is the time/place to do it. 

@olamy do you mind to do that in another PR?

olamy:
@sbordet  yup

olamy:
done via https://github.com/eclipse/jetty.project/pull/8172

PR #8103, opened 2022-06-02
Fixes #7994 - Ability to construct a detached client Request

### LINKED ISSUE #7994: Ability to construct a detached client Request
**Target Jetty version(s)**
10

**Enhancement Description**
It would be nice to be able to construct a client-side `Request` independently of an `HttpClient` instance.

Use-case:
* I want to distribute requests among proxies, and explicitly specify which proxy each request is sent to.
* `HttpClient` does provide the ability to specify proxies at request-construction time, so I am forced to construct multiple `HttpClient`s (one per proxy).
* Before choosing which proxy I want to send a request to, I want to look up whether the server response is already cached offline (i.e. in a local database)
* This means I need to be able to construct a `Request` independent of a specific `HttpClient`, look it up in the cache, and if no match is found bind it to a specific `HttpClient` and send out the request.

As an aside, Apache's HttpClient allows construction and configuration of unbound Request objects. Requests are bound to clients at the very last minute in order to send them out. `OkHttp` does the same.

It would also be great to be able to specify proxies on a per-`Request` basis. This would allow me to use a single `HttpClient`.

sbordet:
Implemented copy of the request tag that was mistakenly missing.


joakime (approved):
LGTM

PR #8134, opened 2022-06-07
Improve cleanup of deflater/inflater pools for PerMessageDeflateExtension

lachlan-roberts:
- Make Pool class available from `CompressionPool` via JMX.
- Add close method to clean up resources in Extensions.
- Close on `PermessageDeflateExtension` should release deflaters.
- Avoid using static exceptions from `FrameFlusher`.

### DIFF
--- a/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/Extension.java
+++ b/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/Extension.java
@@ -14,7 +14,6 @@
 package org.eclipse.jetty.websocket.core;
 
 import java.io.Closeable;
-import java.io.IOException;
 
 /**
  * Interface for WebSocket Extensions.
@@ -29,7 +28,7 @@ public interface Extension extends IncomingFrames, OutgoingFrames, Closeable
     /**
      * Used to clean up any resources after connection close.
      */
-    default void close() throws IOException
+    default void close()
     {
     }
 
--- a/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/Extension.java
+++ b/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/Extension.java
@@ -13,12 +13,15 @@
 
 package org.eclipse.jetty.websocket.core;
 
+import java.io.Closeable;
+import java.io.IOException;
+
 /**
  * Interface for WebSocket Extensions.
  * <p>
  * That {@link Frame}s are passed through the Extension via the {@link IncomingFrames} and {@link OutgoingFrames} interfaces
  */
-public interface Extension extends IncomingFrames, OutgoingFrames
+public interface Extension extends IncomingFrames, OutgoingFrames, Closeable
 {
 
     void init(ExtensionConfig config, WebSocketComponents components);
@@ -26,7 +29,7 @@ public interface Extension extends IncomingFrames, OutgoingFrames
     /**
      * Used to clean up any resources after connection close.
      */
-    default void close()
+    default void close() throws IOException
     {
     }
 
--- a/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/Extension.java
+++ b/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/Extension.java
@@ -24,6 +24,13 @@ public interface Extension extends IncomingFrames, OutgoingFrames
     void init(ExtensionConfig config, WebSocketComponents components);
 
     /**
+     * Used to clean up any resources after connection close.
+     */
+    default void close()
+    {
+    }
+
+    /**
      * The active configuration for this extension.
      *
      * @return the configuration for this extension. never null.
--- a/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/ExtensionStack.java
+++ b/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/ExtensionStack.java
@@ -60,6 +60,14 @@ public class ExtensionStack implements IncomingFrames, OutgoingFrames, Dumpable
         this.behavior = behavior;
     }
 
+    public void close()
+    {
+        for (Extension e : extensions)
+        {
+            e.close();
+        }
+    }
+
     @ManagedAttribute(name = "Extension List", readonly = true)
     public List<Extension> getExtensions()
     {
--- a/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/ExtensionStack.java
+++ b/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/ExtensionStack.java
@@ -62,9 +62,17 @@ public class ExtensionStack implements IncomingFrames, OutgoingFrames, Dumpable
 
     public void close()
     {
-        for (Extension e : extensions)
+        for (Extension ext : extensions)
         {
-            e.close();
+            try
+            {
+                ext.close();
+            }
+            catch (Throwable t)
+            {
+                if (LOG.isDebugEnabled())
+                    LOG.debug("Extension Error During Close", t);
+            }
         }
     }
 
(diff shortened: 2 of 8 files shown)


jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/PerMessageDeflateExtension.java
```diff
@@ -146,6 +147,17 @@ public void init(final ExtensionConfig config, WebSocketComponents components)
         super.init(configNegotiated, components);
     }
 
+    @Override
+    public void close()
+    {
+        // TODO: use IteratingCallback.close() instead of creating exception with failFlusher methods.
+        ClosedChannelException exception = new ClosedChannelException();
```
lorban: You should probably use the same `fillInStackTrace` trick as above, otherwise this `close()` implementation is going to be very expensive.

lorban (dismissed):
I think it would be wise to use a non-stack-filling exception in `close()`. Otherwise LGTM.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/Extension.java
```diff
@@ -23,6 +23,13 @@ public interface Extension extends IncomingFrames, OutgoingFrames
 
     void init(ExtensionConfig config, WebSocketComponents components);
 
+    /**
+     * Used to clean up any resources after connection close.
+     */
+    default void close()
```
sbordet, line 31: Should `Extension extend Closeable`? 
So that `ExtensionStack` can call `IO.close(extension)` and not worry about exceptions thrown by `close()`.

sbordet (changes requested):

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/FrameFlusher.java
```diff
@@ -185,7 +184,14 @@ public void onClose(Throwable cause)
     {
         try (AutoLock l = lock.lock())
         {
-            closedCause = cause == null ? CLOSED_CHANNEL : cause;
+            closedCause = cause == null ? new ClosedChannelException()
```
gregw, line 188: Can you include a TODO here indicating that this is a temporary solution until we can avoid using a throwable as an expensive boolean.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/PerMessageDeflateExtension.java
gregw: Is there a reason close doesn't work now in this PR?

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/TransformingFlusher.java
```diff
@@ -77,6 +77,29 @@ public final void sendFrame(Frame frame, Callback callback, boolean batch)
...
+    public void failFlusher(Throwable t)
+    {
+        boolean failed = false;
+        try (AutoLock l = lock.lock())
+        {
+            if (failure == null)
+            {
+                failure = t;
+                failed = true;
+            }
```
gregw, line 94: else add suppressed?

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/TransformingFlusher.java
```diff
@@ -103,8 +126,14 @@ private class Flusher extends IteratingCallback implements Callback
         private FrameEntry current;
 
         @Override
-        protected Action process()
+        protected Action process() throws Throwable
         {
+            try (AutoLock l = lock.lock())
+            {
+                if (failure != null)
+                    throw failure;
```
gregw, line 139: Need a TODO here that his is a temporary fix.     Essentially we need to check a closed boolean here and create an exception if we are processing whilst closed.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/PerMessageDeflateExtension.java
lachlan-roberts: Close doesn't work now because `DemandingFlusher` and `TransformingFlusher` will still allow new entries to be added and may not fail existing entries if close is called. I think we should carefully assess this in a separate PR.

lachlan-roberts: Opened Issue #8137

gregw: Ouch! That sounds like a bad bug in close... so please make 8137 a moderate high priority.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/ExtensionStack.java
```diff
@@ -62,9 +62,17 @@ public ExtensionStack(WebSocketComponents components, Behavior behavior)
...
-            e.close();
+            try
+            {
+                ext.close();
+            }
+            catch (Throwable t)
+            {
+                if (LOG.isDebugEnabled())
+                    LOG.debug("Extension Error During Close", t);
+            }
```
sbordet, line 75:
```suggestion
            IO.close(ext);
```

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/Extension.java
```diff
@@ -13,20 +13,23 @@
...
+public interface Extension extends IncomingFrames, OutgoingFrames, Closeable
 {
 
     void init(ExtensionConfig config, WebSocketComponents components);
 
     /**
      * Used to clean up any resources after connection close.
      */
-    default void close()
+    default void close() throws IOException
```
sbordet: Drop the `throws IOException` clause, you want those methods to never throw (if they throw, the only choice is to log and ignore anyway).

sbordet (changes requested):

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/Extension.java
lachlan-roberts: If I drop `IOException` here then none of the `Extension` implementations can declare that it throws `IOException`. And then why do I need to call `IO.close()` since an extension cannot declare that it throws `IOException`, the try/catch in `IO.close` is useless and I could just call `ext.close()` directly.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/ExtensionStack.java
lachlan-roberts: This is fine as long as no runtime exceptions are thrown as `IO.close()` will not catch them.

joakime:
Do we want this in Jetty 10.0.10 release?

lachlan-roberts:
@joakime yes this should be in the `10.0.10` release.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/Extension.java
sbordet: I thought IO.close() would catch Throwable, but I was wrong.
Remove the throws, and keep the try/catches in ExtensionStack.

sbordet (approved):

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/FrameFlusher.java
```diff
@@ -44,7 +44,6 @@ public class FrameFlusher extends IteratingCallback
 {
     public static final Frame FLUSH_FRAME = new Frame(OpCode.BINARY);
     private static final Logger LOG = LoggerFactory.getLogger(FrameFlusher.class);
-    private static final Throwable CLOSED_CHANNEL = new ClosedChannelException();
```
gregw, line 47: @lachlan-roberts as discussed on chat.... I think we can keep this static sentinael and just use the constructor that avoids suppressed exceptions and a stacktrace.

PR #8135, opened 2022-06-07
Rework static exceptions in ContentProducer

lorban:
As this is a potential memory leak if `Throwable.addSuppressed()` is ever called on the static exception, or on the exception contained in the static `ErrorContent`.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
@@ -19,6 +19,7 @@ import java.util.concurrent.locks.Condition;
 
 import org.eclipse.jetty.http.BadMessageException;
 import org.eclipse.jetty.http.HttpStatus;
+import org.eclipse.jetty.util.StaticException;
 import org.eclipse.jetty.util.component.Destroyable;
 import org.eclipse.jetty.util.thread.AutoLock;
 import org.slf4j.Logger;
@@ -31,8 +32,8 @@ import org.slf4j.LoggerFactory;
 class AsyncContentProducer implements ContentProducer
 {
     private static final Logger LOG = LoggerFactory.getLogger(AsyncContentProducer.class);
-    private static final HttpInput.ErrorContent RECYCLED_ERROR_CONTENT = new HttpInput.ErrorContent(new IllegalStateException("ContentProducer has been recycled"));
-    private static final Throwable UNCONSUMED_CONTENT_EXCEPTION = new IOException("Unconsumed content")
+    private static final HttpInput.ErrorContent RECYCLED_ERROR_CONTENT = new HttpInput.ErrorContent(new StaticException("ContentProducer has been recycled"));
+    private static final Throwable UNCONSUMED_CONTENT_EXCEPTION = new StaticException("Unconsumed content")
     {
         @Override
         public Throwable fillInStackTrace()
@@ -190,7 +191,7 @@ class AsyncContentProducer implements ContentProducer
         Throwable x = UNCONSUMED_CONTENT_EXCEPTION;
         if (LOG.isDebugEnabled())
         {
-            x = new IOException("Unconsumed content");
+            x = new StaticException("Unconsumed content");
             LOG.debug("consumeAll {}", this, x);
         }
         failCurrentContent(x);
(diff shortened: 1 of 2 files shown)


sbordet (dismissed):

gregw (dismissed):
I'm trying hard to think of why this is going to bite us.... but this all looks very straight forward and how it probably should have been done in the first place.

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
```diff
@@ -31,7 +31,7 @@
 class AsyncContentProducer implements ContentProducer
 {
     private static final Logger LOG = LoggerFactory.getLogger(AsyncContentProducer.class);
-    private static final HttpInput.ErrorContent RECYCLED_ERROR_CONTENT = new HttpInput.ErrorContent(new IllegalStateException("ContentProducer has been recycled"));
+    private static final HttpInput.ErrorContent SENTINEL_ERROR_CONTENT = new HttpInput.ErrorContent(null);
```
sbordet: How about getting rid of this constant instead?

Because it breaks the `ErrorContent` contract (it's an error content so it must have an error, but this one does not), and forces a number of `instanceof ErrorContent` now, and how sure we are we caught all the places where this should be done?

Can `SENTINEL_ERROR_CONTENT` _not_ be an `ErrorContent`?

sbordet (changes requested):

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
lorban: Actually after looking more carefully, `failCurrentContent` can safely set the final content to null if there is no error. And `recycle` can be more lenient by simply using EOF instead of an error content.

sbordet (dismissed):

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
lorban: Scratch that, there are scenarios that absolutely need the content to be an error after `consumeAll()` has been called.

jetty-server/src/main/java/org/eclipse/jetty/server/ContentProducer.java
```diff
@@ -148,5 +150,18 @@
...
+     * This exception used to report when there is some unconsumed content left at the end of a request's processing.
+     * Suppressed exceptions are disabled, meaning calling {@link #addSuppressed(Throwable)} has no effect.
+     */
+    class UnconsumedContentException extends Exception
+    {
+        private static final Logger LOG = LoggerFactory.getLogger(UnconsumedContentException.class);
+
+        public UnconsumedContentException()
+        {
+            super("Unconsumed content", null, false, LOG.isDebugEnabled());
```
gregw: I think this constructor is key!
Seeing that this constructor allows us to disable suppressed exceptions, there is no real reason for us to avoid static Throwables used as sentinels.

I'm not keen on using the LOG to hide stack traces.  I think if it is worth creating an exception of the fly, it is worth generating a stack trace.  I think we should avoid stack traces only for static sentinel Throwables.

@lachlan-roberts can you review your recent change in light of this constructor.

gregw (changes requested):
Do we need to make so many changes now that you've discovered the constructor that allows us to avoid suppressed exceptions for static sentinel Throwables.

Perhaps it is better for use to create a utility StaticThrowable class that has no stack trace and doesn't keep suppressed exceptions.  We can then use that as the base for our existing static exceptions rather than replacing them.

jetty-server/src/main/java/org/eclipse/jetty/server/ContentProducer.java
lorban: I agree. While this code might be worth some refactorings, that's clearly outside the scope of the original issue which is just to avoid leaking stack traces in static exceptions.

I'm going to revert all these changes, introduce a new `StaticException` and keep the changes to a minimum.

gregw (approved):

lorban:
Changes combined with https://github.com/eclipse/jetty.project/pull/8155/

PR #8140, opened 2022-06-08
change-dependabot-interval-to-monthly

olamy:

PR #8141, opened 2022-06-08
change-dependabot-interval-to-monthly

olamy:


gregw (approved):
Thanks

joakime (changes requested):
This doesn't solve anything, in fact it just makes for more work, more time spent, and harsher demands on the CI system.

joakime:
We don't need this one for `jetty-9.4.x`, the other one on PR #8140 is the "live" one.

PR #8155, opened 2022-06-10
Use static exceptions for closing websocket flushers and in ContentProducer

lachlan-roberts:


jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/exception/SentinelWebSocketCloseException.java
```diff
@@ -0,0 +1,32 @@
...
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.websocket.core.exception;
```
lachlan-roberts: Not sure if this should be here or in an internal implementation package.

lorban (changes requested):
LGTM, but I think we should align this PR with this other one: https://github.com/eclipse/jetty.project/pull/8135 as we should use have a common suppressed-exceptions-safe exception class.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/exception/SentinelWebSocketCloseException.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Suppressed exceptions are disabled for this implementation,
+ * meaning calling {@link #addSuppressed(Throwable)} has no effect.
+ * This means instances of {@link SentinelWebSocketCloseException} are suitable to be kept as static fields.
+ */
+public class SentinelWebSocketCloseException extends Exception
```
gregw: As @lorban suggested, perhaps create a util class called StaticException that forces the false,true args of the super constructor.

gregw (changes requested):

lorban (dismissed):

lorban:
Combined here the changes of https://github.com/eclipse/jetty.project/pull/8135 as they're pretty trivial.

jetty-util/src/main/java/org/eclipse/jetty/util/StaticException.java
```diff
@@ -0,0 +1,34 @@
...
+    {
+        this(null);
+    }
+
+    public StaticException(String message)
+    {
+        // Make sure to call the super constructor that disables suppressed exception.
+        super(message, null, false, true);
+    }
+}
```
gregw: no line ending

gregw (dismissed):
I think this is good, other than a tiny quibble.

Going foward, I still wonder if we should every throw/pass such a sentinel or if we should make a new exception when the sentinel is detected.... but for now, this is a good minimal change.

jetty-util/src/main/java/org/eclipse/jetty/util/StaticException.java
```diff
@@ -0,0 +1,34 @@
...
+{
+    public StaticException()
+    {
+        this(null);
+    }
+
+    public StaticException(String message)
+    {
+        // Make sure to call the super constructor that disables suppressed exception.
+        super(message, null, false, true);
```
sbordet: The last boolean parameter is `writableStackTrace`, which I think should be false as well, no?

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
```diff
@@ -190,7 +191,7 @@ public boolean consumeAll()
         Throwable x = UNCONSUMED_CONTENT_EXCEPTION;
         if (LOG.isDebugEnabled())
         {
-            x = new IOException("Unconsumed content");
+            x = new StaticException("Unconsumed content");
```
sbordet: If DEBUG is enabled, then here we want a normal exception like it was before, with stack trace and all, no?
Also, I think this should be moved to TRACE level, as with DEBUG it is just a lot of noise in the logs in case of the happy path.

sbordet (changes requested):

jetty-util/src/main/java/org/eclipse/jetty/util/StaticException.java
lorban: Not if we want to stick to the previous behavior as closely as we can, which is that if we want to disable the collection of stack traces, we subclass and override `fillInStackTrace()` otherwise static exceptions do have a stack trace.

jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContentProducer.java
lorban: This is what you get: `StaticException` collects stack traces by default. I've changed the exception type to make sure we don't get bad surprises by throwing a different type in debug mode and potentially catching the exception somewhere else or ending up in a different catch block. My goal is to make this change as minimal as possible, and eventually refactor later.

I can change the logging to TRACE mode, though.

jetty-util/src/main/java/org/eclipse/jetty/util/StaticException.java
lorban: Fixed.

jetty-util/src/main/java/org/eclipse/jetty/util/StaticException.java
sbordet: I'm not sure I understand. The stack trace of a static exception like this is irrelevant.
Why would we want to keep it?

sbordet (changes requested):

sbordet (approved):

jetty-util/src/main/java/org/eclipse/jetty/util/StaticException.java
lorban: Okay, discussing this offline has shown that we should keep the ability to throw a `StaticException` that has a stack trace, but that shouldn't be its default behavior. Changing accordingly.

PR #8162, opened 2022-06-13
Migrate code from jetty-util Logger to slf4j Logger

joakime:


gregw (approved):
It might be good to automatically detect this in future.

olamy (approved):

PR #8163, opened 2022-06-13
RegexPathSpec documentation and MatchedPath improvements

joakime:
Added javadoc, and improved RegexMatchedPath behavior with more unit tests.

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
@@ -33,15 +33,19 @@ import org.eclipse.jetty.util.log.Logger;
  * <p>
  * Supports {@link PathSpecGroup} for {@link PathSpecGroup#EXACT}, {@link PathSpecGroup#PREFIX_GLOB}, {@link PathSpecGroup#MIDDLE_GLOB}, and {@link PathSpecGroup#SUFFIX_GLOB}.
  * This is done by evaluating the signature or the provided regex pattern for what is a literal vs a glob (of any kind).
- *     <ul>
- *         <li>Only literals, it's a {@link PathSpecGroup#EXACT}.</li>
- *         <li>Starts with literals, ends with globs, it's a {@link PathSpecGroup#PREFIX_GLOB}</li>
- *         <li>Starts with glob, has at least 1 literal, then any thing else, it's a {@link PathSpecGroup#SUFFIX_GLOB}</li>
- *         <li>All other signatures are a {@link PathSpecGroup#MIDDLE_GLOB}</li>
- *     </ul>
- *     The use of regex capture groups, regex character classes, regex quantifiers, and regex special contructs
- *     will be identified as a glob (for signature determination), all other characters are identified
- *     as literal.  The regex {@code ^} beginning of line, and regex {@code $} end of line are ignored.
+ * </p>
+ *
+ * <ul>
+ *   <li>Only literals, it's a {@link PathSpecGroup#EXACT}.</li>
+ *   <li>Starts with literals, ends with globs, it's a {@link PathSpecGroup#PREFIX_GLOB}</li>
+ *   <li>Starts with glob, has at least 1 literal, then any thing else, it's a {@link PathSpecGroup#SUFFIX_GLOB}</li>
+ *   <li>All other signatures are a {@link PathSpecGroup#MIDDLE_GLOB}</li>
+ * </ul>
+ *
+ * <p>
+ * The use of regex capture groups, regex character classes, regex quantifiers, and regex special contructs
+ * will be identified as a glob (for signature determination), all other characters are identified
+ * as literal.  The regex {@code ^} beginning of line, and regex {@code $} end of line are ignored.
  * </p>
  *
  * <p>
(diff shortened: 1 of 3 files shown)


jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
```diff
@@ -21,6 +21,108 @@
...
+ *           and a null returned for {@link MatchedPath#getPathInfo()}</li>
+ *         <li>If the named regex capturing group {@code name} is present, that group
+ *           is returned in {@link MatchedPath#getPathMatch()}</li>
+ *         <li>If the named regex capturing group {@code info} is present, that group
+ *           is returned in {@link MatchedPath#getPathInfo()}</li>
+ *         <li>If the regex is of group type {@link PathSpecGroup#PREFIX_GLOB},
+ *           the beginning of the regex is a literal, so it is split at the start of
+ *           {@code java.util.regex.Matcher.group(1)}),
+ *           taking care to handle trailing slash properly so that {@link MatchedPath#getPathMatch()}
+ *           does not end in it, and {@link MatchedPath#getPathInfo()} starts with it.</li>
```
gregw: I think this correctly describes the current behavior (although I'm still concerned about having more than 1 group).  However, I think we could perhaps consider changing behavior to:

```suggestion
 *         <li>If there is only one capturing group, it is returned as the  {@link MatchedPath#getPathInfo()} and any characters prior to the start of the capture group are returned as the {@link MatchedPath#getPathMatch()}.</li>
```

Or perhaps to better align with the servlet spec

```suggestion
 *         <li>If there is only one anonymous capturing group, the path is split at the start of the capture and the first part of the split is returned as  the {@link MatchedPath#getPathMatch()} and the second part is returned as the  {@link MatchedPath#getPathInfo()}. The first part may be the empty string and the second part may be null</li>
```

Note that this doesn't depend on the type of the match.  I.e we could split an exact match this way.  Why not?

We could define what to do with 2 capture groups, but I don't really see the need.

gregw (changes requested):
I think you have well captured the current behaviour, but I think we should consider changing it a little as per my comment.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
```diff
@@ -279,38 +390,100 @@ private class RegexMatchedPath implements MatchedPath
...
+                        splitIdx = start;
+                        return splitIdx;
+                    }
+                }
+                catch (IllegalArgumentException ignore)
+                {
+                    // ignore if group info not found.
+                }
+
+                if (pathSpec.getGroup() == PathSpecGroup.PREFIX_GLOB)
```
gregw: I don't think this test is needed.    Should be able to split a pattern like `/foo(/*.xml)` .  Ie. if there is only 1 group, then it defines the start of the pathInfo, regardless of match type.

Instead we should test for only 1 capture group

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
```diff
@@ -279,38 +390,100 @@ private class RegexMatchedPath implements MatchedPath
...
+                // Try named group 'info'
+                try
+                {
+                    int start = matcher.start("info");
+                    if (start >= (-1))
+                    {
+                        splitIdx = start;
+                        return splitIdx;
+                    }
+                }
```
gregw: I think a named groups should only be used for splitting if there are not both name and info groups.  If there are both, then we just use them and don't split.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
```diff
@@ -21,6 +21,111 @@
...
+ *           and a null returned for {@link MatchedPath#getPathInfo()}</li>
+ *         <li>If the named regex capturing group {@code name} is present, that group
+ *           is returned in {@link MatchedPath#getPathMatch()}</li>
+ *         <li>If the named regex capturing group {@code info} is present, that group
+ *           is returned in {@link MatchedPath#getPathInfo()}</li>
+ *         <li>If the regex is of group type {@link PathSpecGroup#PREFIX_GLOB},
+ *           the beginning of the regex is a literal, so it is split at the start of
+ *           {@code java.util.regex.Matcher.group(1)}),
+ *           taking care to handle trailing slash properly so that {@link MatchedPath#getPathMatch()}
+ *           does not end in it, and {@link MatchedPath#getPathInfo()} starts with it.</li>
```
gregw: I think this should be as below, to make it clear that splitting only occurs if we don't have both name and info groups:

```suggestion
 *         <li>If there are no regex capturing groups,
 *           the entire path is returned in {@link MatchedPath#getPathMatch()},
 *           and a null returned for {@link MatchedPath#getPathInfo()}</li>
 *         <li>If both the named regex capturing groups {@code name} and {@code info} are present, then 
 *           the {@code name} group is returned in {@link MatchedPath#getPathMatch()} and the 
 *           {@code info} group is returned in {@link MatchedPath#getPathInfo()}</li>
 *         <li>If the named regex capturing group {@code name} is present, that group
 *           is returned in {@link MatchedPath#getPathMatch()} and any following characters (or null) 
 *           are returned in {@link MatchedPath#getPathInfo()} </li>
 *         <li>If the named regex capturing group {@code info} is present, that group
 *           is returned in {@link MatchedPath#getPathInfo()} and any preceediing characters (or empty string) 
 *           is returned in  {@link MatchedPath#getPathMatch()}</li>
 *         <li>If only a single regex capturing group is present, that group
 *           is returned in {@link MatchedPath#getPathInfo()} and any preceediing characters (or empty string) 
 *           is returned in  {@link MatchedPath#getPathMatch()}
```

gregw (changes requested):
I think the logic is getting a bit complex and a bit burried.  I've made a suggestion for how the description should be, but I think the following algorithm should be used as it is a bit more efficient with less name lookups:
```java
switch(matcher.groupCount())
{
    case 0: match=path; info=null; 
    break;

    case 1:
      if ("name".equals(group(1).name)
          match=group; info=path.substring(group.end);
      else
          match=path.substring(0,group.start); info=group; 
      break;

    default:
      if (group("name").exists)
      {   
          match=group("name");
              info=group("info");
          else
              info=path.substring(group("name").end);
      }
      else if (group("info").exists)
          match=path.substring(0,group.start); info=group; break;
      else
          match=path; info=null; 
    break;
}
```

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
joakime: IMO, the only valid split is if the signature starts with literals.
If the signature starts with glob then where is the split?

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
joakime: Fair enough

joakime: Are ok with having the split logic in a separate (and overridable) method?

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
gregw: We only split if we have a single capture group and that doesn't care which part is literal and which part is glob.    The logic described above in both your javadoc and my "optimized" algorithm works regardless of the PathSpecGroup.  Considering the scope for complex regex patterns to confuse the group logic, I don't think it is a big problem to just ignore the group when doing this split and just let the algorithm we've developed do it's stuff.

joakime:
@gregw I implemented your algorithm, but it makes quite the mess. (lots of test failures)
Take a look at the test case examples.

https://github.com/eclipse/jetty.project/blob/52de789437b5415ad7a1c594207e2d6d4c60fb05/jetty-http/src/test/java/org/eclipse/jetty/http/pathmap/RegexPathSpecTest.java#L90-L126

The test argument order is ...
`String regex, String input, String expectedPathMatch, String expectedPathInfo`

Do you think these are sane expectations?
If so, then we need to fix the algorithm, or reintroduce the check on PREFIX_GLOB.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
```diff
@@ -357,78 +371,92 @@ private class RegexMatchedPath implements MatchedPath
...
                 pathInfo = null;
+                return;
             }
-            else if (groupCount == 1)
+
+            if (groupCount == 1)
             {
-                if (idxNameStart >= 0)
+                // we know we are splitting
+                int idxNameEnd = endOf(matcher, "name");
```
gregw, line 358: Ultimately this lookup makes me itch!   If they are not using the named group, then an exception is created and ignored for every lookup attempt.    

For now, can we add a TODO that when we scan the pattern, we should look for the presence of "<name>" and only do this lookup if we have seen it in the pattern.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/RegexPathSpec.java
```diff
@@ -357,78 +371,92 @@ private class RegexMatchedPath implements MatchedPath
...
+                        pathInfo = '/' + pathInfo;
+                    }
+                    return;
                 }
             }
+
+            // Reach here we have 2+ groups
+
+            String gName = valueOf(matcher, "name");
+            String gInfo = valueOf(matcher, "info");
```
gregw, line 395: Two exceptions may be created and ignored here.  So again, I think we need a TODO to optimise that away in future.

gregw (dismissed):
I think this is good enough for now, but with TODOs for future optimizations.

joakime:
Once the javadoc issues on PR #8173 are green and merged, this should build green, then I can merge this one.

jmcc0nn3ll (approved):

Issue #8151, opened 2022-06-09
`JakartaWebSocketSession.close()` blocks long time when called from `SendHandlerCallback`


yamass:
**Jetty version(s)**
11.0.9
(worked with 9.4.44.v20210927) 

**Java version/vendor** `(use: java -version)`
temurin-17.0.2

**OS type/version**
Ubuntu Linux

**Description**

JakartaWebSocketSession.close() blocks when it is invoked from a SendHandlerCallback. In fact, it only unblocks after the idle timeout. Example (see also attached project):

```java
@ClientEndpoint
@ServerEndpoint(value = "/events/")
public class TestSocketJetty11 {

	@OnOpen
	public void onWebSocketConnect(Session session) {
		System.out.println("Socket Connected: " + session);
		Executors.newSingleThreadScheduledExecutor().schedule(() -> {
			System.out.println("outer...");
			session.getAsyncRemote().sendText("GOODBYE", sendResult -> { // call close inside the callback!
				try {
					System.out.println("inner...");
					long startTime = System.currentTimeMillis();
					session.close();
					System.out.println("inner done. Took " + (System.currentTimeMillis() - startTime) + "ms.");
				} catch (IOException e) {
					throw new RuntimeException(e);
				}
			});
			System.out.println("outer done");
		}, 2, TimeUnit.SECONDS);
	}

	@OnClose
	public void onWebSocketClose(CloseReason reason) {
		System.out.println("Socket Closed: " + reason);
	}

	@OnError
	public void onWebSocketError(Throwable cause) {
		cause.printStackTrace(System.err);
	}

}
```
Result:
```

Socket Connected: JakartaWebSocketSession@7fb54e1e[SERVER,JakartaWebSocketFrameHandler@5611479e[endpoint=org.eclipse.jetty.demo.TestSocketJetty11]]
outer...
inner...
org.eclipse.jetty.websocket.core.exception.WebSocketTimeoutException: Connection Idle Timeout
	at org.eclipse.jetty.websocket.core.internal.WebSocketConnection.onIdleExpired(WebSocketConnection.java:233)
	at org.eclipse.jetty.io.AbstractEndPoint.onIdleExpired(AbstractEndPoint.java:407)
	at org.eclipse.jetty.io.IdleTimeout.checkIdleTimeout(IdleTimeout.java:166)
	at org.eclipse.jetty.io.IdleTimeout.idleCheck(IdleTimeout.java:108)
[...]

PR #8164, opened 2022-06-14
Issue #8151 - make websocket close non-blocking

lachlan-roberts:
Proposed fix to #8151 by making the close non-blocking so it can be called from within the send callback.

### DIFF
--- a/jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/JavaxWebSocketSession.java
+++ b/jetty-websocket/websocket-javax-common/src/main/java/org/eclipse/jetty/websocket/javax/common/JavaxWebSocketSession.java
@@ -13,7 +13,6 @@
 
 package org.eclipse.jetty.websocket.javax.common;
 
-import java.io.IOException;
 import java.net.URI;
 import java.security.Principal;
 import java.time.Duration;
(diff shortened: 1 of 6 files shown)


sbordet:
I don't think changing the semantic of `close()` is a good idea. It has, and always had, a blocking semantic.

sbordet:
@lachlan-roberts I don't think making `close()` async is right, but I'm worried that the problem is a different one, i.e. sends are not reentrant.

What if you do another send() from a callback? Does it block?

lachlan-roberts:
@sbordet yes the problem with sends still exists if you do a blockingSend inside the callback for an asyncSend then the blockingSend will be stuck until connection times out. This is because we're using an iterating callback for the writes.

sbordet:
@lachlan-roberts then perhaps it's only a problem of ordering of when the callback is called... first we must complete the `IteratingCallback` and then succeeding the external callback.

sbordet:
After discussion, I think there is no other way other than making `close()` async. The use case in #8151 is legit and unfortunately the API does not give applications a choice of using blocking vs non-blocking `close()` (like it does with `send()`, so the only option is to make `close()` async.

sbordet (changes requested):
Can you please add if possible a `close(int, String, Callback)` to the Jetty WS APIs?

sbordet (approved):

Issue #8161, opened 2022-06-13
Improve SSLConnection buffers handling (CVE-2022-2191)


lorban:
**Jetty version(s)**
10+

**Description**
`SSLConnection`'s buffers utilization and their pooling should be reviewed.

**Fixes Security Advisory**
https://github.com/eclipse/jetty.project/security/advisories/GHSA-8mpp-f3f7-xc28
CVE-2022-2191

joakime:
In light of discoveries during this review, a better combined ByteBufferPool with easier to configure setup was just merged in PR #8171, due for the next Jetty 10.0.x release.

AB-xdev:
Does this also affect Jetty 9.x?

According to the issue description this is only relevant for Jetty 10+ but the advisory says <= 10.0.9 which also includes version 9.x.

lorban:
No, it does not affect any 9.4.x version.

joakime:
> According to the issue description this is only relevant for Jetty 10+ but the advisory says <= 10.0.9 which also includes version 9.x.

Good catch!

We'll update the advisory version range.

kadinwu:
affected versions in github advisories [(https://github.com/advisories/GHSA-8mpp-f3f7-xc28)] has < 10.0.10.
does it really affect 9.x versions?

joakime:
> affected versions in github advisories https://github.com/advisories/GHSA-8mpp-f3f7-xc28 has < 10.0.10.
> does it really affect 9.x versions?

See prior comments, and our advisory (the master database at github has not been updated yet):

* https://github.com/eclipse/jetty.project/issues/8161#issuecomment-1178788271
* https://github.com/eclipse/jetty.project/issues/8161#issuecomment-1178728623
* https://github.com/eclipse/jetty.project/security/advisories/GHSA-8mpp-f3f7-xc28

Also, Jetty 9.4.x is now at End of Community Support, you are strongly encouraged to upgrade to Jetty 10+ as soon as possible.

See:

* https://github.com/eclipse/jetty.project/issues/7958

joakime:
The github advisory database version of CVE-2022-2191 has its version range updated.
https://github.com/advisories/GHSA-8mpp-f3f7-xc28

PR #8165, opened 2022-06-14
Improve ssl buffers handling

lorban:
Solves https://github.com/eclipse/jetty.project/issues/8161


jetty-io/src/main/java/org/eclipse/jetty/io/ArrayRetainableByteBufferPool.java
```diff
@@ -91,8 +91,8 @@ protected ArrayRetainableByteBufferPool(int minCapacity, int factor, int maxCapa
         _maxCapacity = maxCapacity;
         _direct = directArray;
         _indirect = indirectArray;
-        _maxHeapMemory = maxHeapMemory;
-        _maxDirectMemory = maxDirectMemory;
+        _maxHeapMemory = (maxHeapMemory != 0L) ? maxHeapMemory : Runtime.getRuntime().maxMemory() / 4;
+        _maxDirectMemory = (maxDirectMemory != 0L) ? maxDirectMemory : Runtime.getRuntime().maxMemory() / 4;
```
gregw, line 139: javadoc this heuristic. Also include it in the PR description.

jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
```diff
@@ -790,6 +809,9 @@ public int fill(ByteBuffer buffer) throws IOException
                     }
                     catch (Throwable x)
                     {
+                        // Clear the input buffers so that releaseInputBuffers()
+                        // in the finally block will not leak them.
+                        clearInputBuffers();
```
gregw: I think a better solution would be to javadoc that "releaseInputBuffers" only releases buffers if they are not empty. Then here use a method "discardInputBuffers()" which is javadoc'd to say that it will release the buffer EVEN if not empty.

Or perhaps rename `releaseInputBuffers` to `releaseEmptyInputBuffers` and then have a `releaseInputBuffers` that always releases.    Eitherway, I don't like clearing to influence a later call - just release them here.

jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
```diff
@@ -1159,6 +1180,10 @@ public boolean flush(ByteBuffer... appOuts) throws IOException
                     }
                     catch (Throwable x)
                     {
+                        // Clear the encrypted output buffer so that
+                        // releaseEncryptedOutputBuffer() in the finally block
+                        // will not leak it.
+                        BufferUtil.clear(_encryptedOutput);
```
gregw: Ditto.  Use a method that clearly always releases the buffers and then you don't need the comment explanation.

jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
```diff
@@ -1323,21 +1353,32 @@ public void doShutdownOutput()
...
+                    LOG.trace("IGNORED", x);
+                disconnect();
+            }
+        }
+
+        private void disconnect()
+        {
+            try (AutoLock l = _lock.lock())
+            {
+                BufferUtil.clear(_encryptedOutput);
```
gregw: Is this a clear to force a release? if so, then just release here

gregw (changes requested):
I don't like the "clear to force a release later" pattern.    Instead just have clearly named methods that release if empty vs always releases:
 +  `releaseEmptyInputBuffers`; `releaseInputBuffers`
 +  `releaseInputBuffersIfEmpty`; `clearAndReleaseInputBuffers`

But ultimately, we should not need comments like "clearing so a later release will take effect". Instead we should just release immediately in those locations - the subsequent final releases are already null check protected.

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayRetainableByteBufferPool.java
sbordet: If not already, document the same heuristics in `ArrayByteBufferPool` as well.

jetty-server/src/main/config/modules/retainablebytebufferpool.mod
```diff
@@ -0,0 +1,31 @@
...
+## Requests for ByteBuffers larger than this value results
+## in the ByteBuffer being allocated but not pooled.
+#jetty.retainableByteBufferPool.maxCapacity=65536
+
+## Bucket capacity factor.
+## ByteBuffers are allocated out of buckets that have
+## a capacity that is multiple of this factor.
+#jetty.retainableByteBufferPool.factor=1024
+
+## Maximum size for each bucket.
```
sbordet:
```suggestion
## Maximum number of ByteBuffers for each bucket.
```

jetty-server/src/main/config/modules/retainablebytebufferpool.mod
```diff
@@ -0,0 +1,31 @@
...
+#jetty.retainableByteBufferPool.maxBucketSize=2147483647
+
+## Maximum heap memory retainable by the pool (0 for heuristic, -1 for unlimited).
```
sbordet:
```suggestion
## Maximum heap memory bytes retainable by the pool (0 for heuristic, -1 for unlimited).
```

jetty-server/src/main/config/modules/retainablebytebufferpool.mod
```diff
@@ -0,0 +1,31 @@
...
+#jetty.retainableByteBufferPool.maxHeapMemory=0
+
+## Maximum direct memory retainable by the pool (0 for heuristic, -1 for unlimited).
```
sbordet:
```suggestion
## Maximum direct memory bytes retainable by the pool (0 for heuristic, -1 for unlimited).
```

sbordet (changes requested):
I agree with @gregw comments about having a `discard*Buffer` that does both clearing and releasing.

jetty-server/src/main/config/modules/retainablebytebufferpool.mod
lorban: done.

jetty-server/src/main/config/modules/retainablebytebufferpool.mod
lorban: done.

jetty-server/src/main/config/modules/retainablebytebufferpool.mod
lorban: done.

jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
lorban: done.

jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
lorban: done.

jetty-io/src/main/java/org/eclipse/jetty/io/ssl/SslConnection.java
lorban: done.

gregw (dismissed):
LGTM

sbordet (approved):

PR #8168, opened 2022-06-15
Jetty 12.0.x fix websocket modules

janbartel:
Fix up websocket modules and jetty-home support for ee10 websocket. ee9 TBD.


janbartel:
I'm going to merge this into the 12.0.x-update-jakarta-apis branch instead of 12.0.x.

PR #8169, opened 2022-06-15
Make RetainableByteBufferPool a 1st class citizen in the API

lorban:


gregw (commented):
I don't love this.... could we consider doing this more like ThreadPool and ReservedThreadPool, where we just have a single pool that can be used in either way?

gregw:
Whilst there might be more work to do on #8171, I think it is a much better approach than this..... so closing this one.

PR #8171, opened 2022-06-16
Combined ByteBufferPool

gregw:
WIP on the concept of a single ByteBufferPool that can act either as a normal pool or as a retainable pool


gregw:
This is an alternative to #8169

jetty-io/src/main/java/org/eclipse/jetty/io/MappedByteBufferPool.java
```diff
@@ -99,13 +99,13 @@ public MappedByteBufferPool(int factor, int maxQueueLength, Function<Integer, Bu
      */
     public MappedByteBufferPool(int factor, int maxQueueLength, Function<Integer, Bucket> newBucket, long maxHeapMemory, long maxDirectMemory)
     {
-        super(factor, maxQueueLength, maxHeapMemory, maxDirectMemory);
+        super(factor, 0, maxQueueLength, maxHeapMemory, maxDirectMemory, -1, -1);
```
lorban: The default heap and direct memory should be 0 to use the heuristic, not -1 for unlimited.

lorban (commented):
AFAICT all the classes using the retainable pool still access it via `RetainableByteBufferPool.findOrAdapt()` except that we're not registering a retainable pool in the connector anymore.

I'm not sure what this brings compared to what we have now.

gregw:
@lorban said:
> I'm not sure what this brings compared to what we have now.

All abstract byte buffer pools now contain a retainable pool (either a null one or real one). This means we need to pass and configure only a single pool.  

I've kept findOrAdapt for now for backwards comparability for those who currently set a retainable pool on the server, but it could be simplified and then inlined/removed in 12.

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
```diff
@@ -30,36 +30,64 @@
...
         _maxHeapMemory = (maxHeapMemory != 0) ? maxHeapMemory : Runtime.getRuntime().maxMemory() / 4;
         _maxDirectMemory = (maxDirectMemory != 0) ? maxDirectMemory : Runtime.getRuntime().maxMemory() / 4;
+
+        if (retainedHeapMemory < 0 && retainedDirectMemory < 0)
+            _retainableByteBufferPool = RetainableByteBufferPool.from(this);
+        else
+            _retainableByteBufferPool = newRetainableByteBufferPool(maxCapacity, maxBucketSize,
+                (retainedHeapMemory != 0) ? retainedHeapMemory : Runtime.getRuntime().maxMemory() / 4,
+                (retainedDirectMemory != 0) ? retainedDirectMemory : Runtime.getRuntime().maxMemory() / 4);
+    }
```
gregw: @lorban this is the substantive change.   Every `AbstractByteBufferPool` contains a `RetainableByteBufferPool` that can be configured by only two properties: `retainedHeapMemory` and `retainedHeapMemory`.     This allows minimal changes to XML (see below) to support simple configuration of both types of pool.

jetty-server/src/main/config/etc/jetty-bytebufferpool.xml
```diff
@@ -8,5 +8,7 @@
     <Arg type="int"><Property name="jetty.byteBufferPool.maxQueueLength" default="-1"/></Arg>
     <Arg type="long"><Property name="jetty.byteBufferPool.maxHeapMemory" default="0"/></Arg>
     <Arg type="long"><Property name="jetty.byteBufferPool.maxDirectMemory" default="0"/></Arg>
+    <Arg type="long"><Property name="jetty.byteBufferPool.retainedHeapMemory" default="0"/></Arg>
+    <Arg type="long"><Property name="jetty.byteBufferPool.retainedDirectMemory" default="0"/></Arg>
```
gregw, line 12: @lorban With this simple/minimal change to the exist module/xml we get to configure both types of buffer pool in one.

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayRetainableByteBufferPool.java
```diff
@@ -191,9 +191,19 @@ public RetainableByteBuffer acquire(int size, boolean direct)
...
 
+    protected ByteBuffer allocate(int capacity)
+    {
+        return ByteBuffer.allocate(capacity);
+    }
+
+    protected ByteBuffer allocateDirect(int capacity)
+    {
+        return ByteBuffer.allocateDirect(capacity);
+    }
```
gregw, line 203: Note that retainable pool takes buffers from the normal pool, so that can reduce wastage as the retainable pool grows.   I have yet to implement release, but will do so.

gregw:
@sbordet your thoughts?

jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferPool.java
```diff
@@ -78,6 +78,11 @@ default ByteBuffer newByteBuffer(int capacity, boolean direct)
         return direct ? BufferUtil.allocateDirect(capacity) : BufferUtil.allocate(capacity);
     }
 
+    default RetainableByteBufferPool asRetainableByteBufferPool()
+    {
+        return RetainableByteBufferPool.from(this);
+    }
+
     public static class Lease
```
gregw: @sbordet is `Lease` really the right name for this?  It's kind of a `ByteBufferCollection`?    Eitherway, can you add some javadoc for the class

jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferPool.java
```diff
@@ -78,7 +72,17 @@ default ByteBuffer newByteBuffer(int capacity, boolean direct)
         return direct ? BufferUtil.allocateDirect(capacity) : BufferUtil.allocate(capacity);
     }
 
-    public static class Lease
+    /**
+     * Get this pool as a {@link RetainableByteBufferPool}, which supports reference counting of the
+     * buffers and possibly a more efficient lookup mechanism based on the {@link org.eclipse.jetty.util.Pool} class.
+     * @return This pool wrapped as a RetainableByteBufferPool.
+     */
+    default RetainableByteBufferPool asRetainableByteBufferPool()
```
lorban: Having the default implementation return a new instance each time while the `AbstractByteBufferPool` implementation always returns the same instance sounds like the lifecycle of the objects returned by this method is underdefined.

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
```diff
@@ -30,36 +35,64 @@
...
+    protected AbstractByteBufferPool(int factor, int maxCapacity, int maxBucketSize, long maxHeapMemory, long maxDirectMemory, long retainedHeapMemory, long retainedDirectMemory)
     {
         _factor = factor <= 0 ? 1024 : factor;
-        _maxQueueLength = maxQueueLength;
+        _maxCapacity = maxCapacity > 0 ? maxCapacity : 64 * _factor;
+        _maxBucketSize = maxBucketSize;
         _maxHeapMemory = (maxHeapMemory != 0) ? maxHeapMemory : Runtime.getRuntime().maxMemory() / 4;
         _maxDirectMemory = (maxDirectMemory != 0) ? maxDirectMemory : Runtime.getRuntime().maxMemory() / 4;
+
+        if (retainedHeapMemory < 0 && retainedDirectMemory < 0)
```
lorban: Since the allocation of the `_retainableByteBufferPool` instance is done in the constructor, its release should be done in the `clear` method, shouldn't it?

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
```diff
@@ -30,36 +35,64 @@
...
+            _retainableByteBufferPool = RetainableByteBufferPool.from(this);
+        else
+            _retainableByteBufferPool = newRetainableByteBufferPool(factor, maxCapacity, maxBucketSize,
+                (retainedHeapMemory != 0) ? retainedHeapMemory : Runtime.getRuntime().maxMemory() / 4,
+                (retainedDirectMemory != 0) ? retainedDirectMemory : Runtime.getRuntime().maxMemory() / 4);
+    }
+
+    protected RetainableByteBufferPool newRetainableByteBufferPool(int factor, int maxCapacity, int maxBucketSize, long retainedHeapMemory, long retainedDirectMemory)
+    {
+        return RetainableByteBufferPool.from(this);
```
lorban, line 88: This method has the exact same implementation as the super `asRetainableByteBufferPool` impl which looks odd to me. There's something not well defined about those two contracts.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
```diff
@@ -36,23 +36,38 @@
...
-     * @param byteBufferPool the {@link ByteBufferPool} to wrap if no memory pool was found in the container.
-     * @return the {@link RetainableByteBufferPool} found or the wrapped one.
+     * @param byteBufferPool Use  {@link ByteBufferPool#asRetainableByteBufferPool()} to convert if no memory pool was found in the container.
+     * @return the {@link RetainableByteBufferPool} found or the converted one.
      */
     static RetainableByteBufferPool findOrAdapt(Container container, ByteBufferPool byteBufferPool)
     {
         RetainableByteBufferPool retainableByteBufferPool = container == null ? null : container.getBean(RetainableByteBufferPool.class);
-        if (retainableByteBufferPool == null)
+        return retainableByteBufferPool == null ? byteBufferPool.asRetainableByteBufferPool() : retainableByteBufferPool;
```
lorban: If a `RetainableByteBufferPool` bean is added, this means there's going to be a useless `ArrayRetainableByteBufferPool` instance kept in the `ByteBufferPool`. This is going to be confusing in server dumps.

Plus, in the current implementation you return an object for whom the caller is sometimes responsible for its releasee and sometimes not.

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
```diff
@@ -190,9 +188,7 @@ public AbstractConnector(
         if (pool == null)
             pool = _server.getBean(ByteBufferPool.class);
         _byteBufferPool = pool != null ? pool : new ArrayByteBufferPool();
-        addBean(_byteBufferPool);
-        RetainableByteBufferPool retainableByteBufferPool = _server.getBean(RetainableByteBufferPool.class);
-        addBean(retainableByteBufferPool == null ? new ArrayRetainableByteBufferPool() : retainableByteBufferPool, retainableByteBufferPool == null);
+        addBean(_byteBufferPool, pool == null);
```
lorban: If you do not "forward" the `RetainableByteBuffer` bean for the server to the connector, you're changing the current behavior as all the `findOrAdapt()` use the connector as the container to look into.

You're also breaking the principle of least surprise as you can override the default non-retainable pool by adding a bean on the server but you can't for the retainable pool.

lorban (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
gregw: I don't think so.  The main byte buffer pool is keep hold of idle buffers only - ie ones that it has previously been acquired and released.   The retainableByteBufferPool is kind of just like a client that has acquired but not yet released the buffer.

So clear should not release retained buffers any more than it should "take back" buffers allocated to other users.    But perhaps it could trigger a shrinking of the retained pool to only buffers that are in use?   Is t hat semantic already supported?

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
gregw: Ah yes - should return `_retainableByteBufferPool`. standby....

jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferPool.java
gregw: Perhaps.    We can return a new instance here because it is a non allocating one - ie it is the null implementations. So would it be OK to say that if the pool returned does actually retain buffers then the pool returned must always be the same (or a wrapper of the same) retained pool.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
gregw: We are not adding it.    I'm just supportin the use-case where a user has added their own with a specific configuration.    If they also configure the main pool to have a retained buffers, then there is duplication.

Perhaps, since this is a major release, we don't support this backwards compatibility?

In which case, this method just become a call to `asRetainableByteBufferPool` and can be inlined.   Maybe that is simplest.

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
gregw: The super `asRetainableByteBufferPool` is just the default impl that creates a new null instance for every call.   At this abstract level, we do the same at construction, so at least the same instance is always returned.

Perhaps I get rid of the default impl and say that the same instance should always be returned?

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
gregw: @lorban I think we need to review exactly what clear does on a retainable pool.... which means we need to review what remove does on a Pool.  I don't think the semantic for that is correct, specially for removing active items.  Also for multi items, remove might return false, but the item is actually closed so will eventually be removed?? this makes accounting on the return value of remove difficult.   Needs review.   So for now, let's not wire up clear.

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
gregw: done

jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferPool.java
gregw: removed default impl and javadoc now says must be the same instance always

jetty-io/src/main/java/org/eclipse/jetty/io/MappedByteBufferPool.java
gregw: I don't think all pools should have a retained pool by default.... or maybe that is OK since it will start empty and only fill if used???

Need to consider defaults... also some strangeness with -1 vs 0 for disabled vs heuristic.   more review needed.

jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
gregw: removed and inlined

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
gregw: I've removed `findOrAdapt`, so a connector will always get it's retainable buffer pool from it's primary buffer pool.

However, I'm a little cautious about how we create buffer pools in each connector.  It might be better to always create a single pool in the server whenever a connector is created without a pool being passed.

gregw:
@lorban we need to think a bit harder about the default values.  Currently we have lots of parameters that are defined as:
 + > 0 use the value.
 + 0 use the default heuristic
 + -1 unlimited.

But I think we also want the ability to disable via these parameters.  Ie how do you say you don't want to pool any direct memory or to not retain any heap memory.

Perhaps it would be better to have:
 + Long.MAX_VALUE for unlimited
 + > 0 use the value
 + 0 for disabled (or essentially the 0 value)
 + -1 for the heuristic

We need to determine a single scheme and use it and document it in several places (all the javadoc in the constructors is currently different).  Pity you can't inherit constructor javadoc!

thoughts?

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
lorban: The `clear` contracts of both pools are slightly different, but not incompatible IMHO.

`clear` on BBP immediately stops holding onto its buffers.
`clear` on RBBP immediately stops holding onto the unused buffers it contains, and marks the ones still in use as to-be-removed-when-released.

If you clear the RBBP then clear the BBP you make RBBP give all its unused buffers back to the BBP then make BBP drop all its buffers. I think this is what we want.

jetty-io/src/main/java/org/eclipse/jetty/io/MappedByteBufferPool.java
lorban: The MappedByteBufferPool is primarily used by the client which benefits from a RBBP too, I don't see why we would disable the retainability by default.

That -1 vs 0 change you made breaks the pre-existing contract, perhaps you should revert to 0 for heuristic, -1 for unlimited and add -2 for non-retaining retainable pool?

jetty-server/src/main/config/etc/jetty-bytebufferpool.xml
lorban: Which is both very nice and desirable. :+1:

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
lorban: Right, the existing behavior is slightly broken as it creates one RBBP per connector by default. It should have registered it as a bean onto the server instead of just onto the connector.

But if a user registers a BBP or a RBBP onto the server, those pools are used and shared across all connectors, so you're breaking backward compatibility by not looking for a RBBP bean in the server.

lorban:
@gregw the fact that we have two distinct pools in 10/11 is a design mistake we made. BBP/RBBP should have been better integrated from the start, like what you're suggesting.

But do we want to break backward compatibility with the  0 / -1 values? If not, we should keep 0 for heuristic, -1 for unlimited and maybe add -2 to disable the retaining feature? But do we want to disable the retaining feature, knowing that `findOrAdapt()` was introduced because of existing Connector constructors only accepting BBP?

This is something that should be totally cleaned up in 12. I don't mind the queue-based pool algorithm, but maybe 12 should only use the RBBP API everywhere?

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
gregw: > `clear` on RBBP immediately stops holding onto the unused buffers it contains, and marks the ones still in use as to-be-removed-when-released.

Yeah I think that is the intention, but it is not the implementation!  So I think it needs review and javadocing before linking the clears together

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
gregw: I was looking for the RBBP bean in the server, but I thought you didn't like that, so I removed it. 

I'll ponder some more on the best solution that we want, and then see how much backwards compatibility can be achieved.

gregw:
@lorban I think this PR shows that having one BBP that contains a RBBP works pretty well and is the same to ThreadPool containing a ReservedThreadPool.

I think we do have some scope to clean up the -1, 0, values because they are mostly used internally... but perhaps best to leave to 12??? I'm on the fence.

I'm dubious we have the ability to remove the BBP/RBBP duality from jetty-12 entirely.  I also think that this PR shows that it is not strictly necessary.  The combined pool model can pretty much achieve most outcomes - ie we could configure to have a non zero RBBP, but a zero BBP if we wanted to.    It's really just a matter of how much each API is used.     I think we have more important things to work on rather than hunting down usages of BBP only to then find out days/weeks/months/years later that we have buffer leak as we were not handling all scenarios correctly.

lorban:
@gregw I like the idea of the combined BBP / RBBP, it's a strong +1 but we definitely have to resolve what we want to do with the -1, 0, values as those are public, they're even in the `bytebufferpool.mod` file.

Regarding 12, we could keep the leak-resilient pool impl by default and still replace all BBP usages with RBBP.

sbordet:
I like the idea, but I think the implementations should give up using queues in favor of `Pool`, which we know being much faster.

joakime:
Skipping this release, moving to 10.0.12 / 11.0.12 release.

gregw:
@sbordet it does use Pool for the RBBP.  The term queue size is used in a few places when it really means bucket size.

gregw:
@sbordet the usage of the Queue in ABBP gives a different semantic to the usage of Pool in ARBBP.  Specifically BBPs have an entry only for idle buffers and when a buffer is acquired it leaves the pool entirely and may never return.  But for RBBP, an acquired buffer is still in the pool in an active state, which is the semantic provided by Pool.

I'm OK with gradually replacing BBP API usage with RBBP usage, but I don't think it will always be appropriate, nor do I think we should hurry as we have already had several bugs due to RBBP leaks which were not a problem with BBP.

gregw:
@lorban I have improved a lot of the -2,-1,0 configuration and I think it is more consistent.
I also increased the default factor to 4096, as 1024 results in too many buckets.

I tried using the log pool by default, but that resulted in some tests failing (eg AsyncCompletionTest.testAsyncIOWrite.[66] AWCH{w=BYTE,cl=true,ir=true,f=true,c=true,d=330}) so if you have time, perhaps you could look at that?

jetty-server/src/main/config/etc/jetty-bytebufferpool-logarithmic.xml
```diff
@@ -7,5 +7,7 @@
     <Arg type="int"><Property name="jetty.byteBufferPool.maxQueueLength" default="-1"/></Arg>
```
lorban: This property should be renamed.

jetty-server/src/main/config/etc/jetty-bytebufferpool.xml
```diff
@@ -8,5 +8,7 @@
     <Arg type="int"><Property name="jetty.byteBufferPool.maxQueueLength" default="-1"/></Arg>
```
lorban: This property should be renamed.

jetty-server/src/main/config/modules/bytebufferpool-logarithmic.mod
```diff
@@ -23,8 +23,14 @@ etc/jetty-bytebufferpool-logarithmic.xml
...
 
-## Maximum heap memory retainable by the pool (0 for heuristic, -1 for unlimited)
+## Maximum heap memory held idle by the pool (0 for heuristic, -1 for unlimited).
 #jetty.byteBufferPool.maxHeapMemory=0
 
-## Maximum direct memory retainable by the pool (0 for heuristic, -1 for unlimited)
+## Maximum direct memory held idle by the pool (0 for heuristic, -1 for unlimited).
 #jetty.byteBufferPool.maxDirectMemory=0
+
+## Maximum heap memory retained whilst in use by the pool (0 for heuristic, -1 for unlimited).
```
lorban: document -2 for no retained memory?

jetty-server/src/main/config/modules/bytebufferpool-logarithmic.mod
```diff
@@ -23,8 +23,14 @@ etc/jetty-bytebufferpool-logarithmic.xml
...
+#jetty.byteBufferPool.retainedHeapMemory=0
+
+## Maximum direct memory retained whilst in use by the pool (0 for heuristic, -1 for unlimited).
```
lorban: document -2 for no retained memory?

jetty-server/src/main/config/modules/bytebufferpool.mod
```diff
@@ -1,5 +1,6 @@
 [description]
 Configures the ByteBufferPool used by ServerConnectors.
+Use module "bytebufferpool-logarithmic" for a more space efficient pool.
```
lorban: "more space efficient" is debatable. SSL requires buffers that are a few bytes over 16 KB large, meaning a log pool would waste nearly 50% of the memory it hands off in this scenario.

jetty-server/src/main/config/modules/bytebufferpool.mod
```diff
@@ -24,8 +25,14 @@ etc/jetty-bytebufferpool.xml
...
 
-## Maximum heap memory retainable by the pool (0 for heuristic, -1 for unlimited).
+## Maximum heap memory held idle by the pool (0 for heuristic, -1 for unlimited).
 #jetty.byteBufferPool.maxHeapMemory=0
 
-## Maximum direct memory retainable by the pool (0 for heuristic, -1 for unlimited).
+## Maximum direct memory held idle by the pool (0 for heuristic, -1 for unlimited).
 #jetty.byteBufferPool.maxDirectMemory=0
+
+## Maximum heap memory retained whilst in use by the pool (0 for heuristic, -1 for unlimited).
```
lorban: document -2 for no retained memory?

jetty-server/src/main/config/modules/bytebufferpool.mod
```diff
@@ -24,8 +25,14 @@ etc/jetty-bytebufferpool.xml
...
+#jetty.byteBufferPool.retainedHeapMemory=0
+
+## Maximum direct memory retained whilst in use by the pool (0 for heuristic, -1 for unlimited).
```
lorban: document -2 for no retained memory?

jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
lorban: ```addBean(_byteBufferPool, pool == null);``` is redundant with the previous `addBean` calls made in the `synchronized` block, isn't it?

lorban (changes requested):

lorban:
I've had a look at `AsyncCompletionTest.testAsyncIOWrite.[66]` and it seems that a bug/poor feature in `LogarithmicArrayByteBufferPool` triggers a pre-existing bug in `HttpOutput.write(int b)`. If the logic after `if (!flush)` is executed more than once, then the test fails.

This happens when `LogarithmicArrayByteBufferPool` returns a 128 bytes buffer to `acquireBuffer()`'s 99 bytes buffer request, and since that test tries to write over 300 bytes the buffer runs out, multiple flushes are needed and something goes wrong along the way. This does not happen with `ArrayByteBufferPool` as the returned buffer has a 8192 bytes capacity which is enough to reduce the flushes to a single one.

So `LogarithmicArrayByteBufferPool` (and its retainable counterpart) should be modified so that they don't create tiny buckets by default (buckets of 1, 2, 4, 8, 16, 32, 64... bytes are created by default) then the `HttpOutput.write(int b)` method that writes one single byte at a time should be investigated to figure out what's wrong with the flushes.

gregw:
@lorban isn't that a bug in the unit test? Small buffets might not be desirable, but it is bigger than asked for, so the test should not assume it can write 330 bytes without blocking

jetty-server/src/main/config/modules/bytebufferpool.mod
gregw: Well more space efficient in the pool, but may be less space efficient in the buffer itself.

For the sizes we are commonly using, I'm actually thinking that just increasing the default factor to 4K is pretty good..... except for tiny buffers.  Geting 1K when you only needed 6 bytes was bad enough, but 4K get's a bit ridiculous.

gregw:
hmmm so we appear to have a few tests that are relying on either getting extra space in a buffer or not getting extra space in a buffer.... so changing the buffer pool algorithm and/or defaults has broken a few tests.    Not good.  Fixing as I find them.

jetty-server/src/main/config/modules/bytebufferpool.mod
lorban: "less space efficient in the buffer" is indeed what my comment is about and your remark about tiny buffers is exactly my point, but it also applies to other scenarios.

When you consider that in a typical SSL scenario where the buffer pool is limited in size, configuring a logarithmic pool may very well be noticeably less efficient; i.e.: a pool configured to hold a maximum of 16 MB is only going to be able to hold at most ~500 SSL buffers when using a logarithmic divide while it could hold up to ~950 with a linear divide.

I'm just not sure if it's a real issue in practice, so I'm conservatively thinking that we should not change the default nor recommend alternatives until we have a clearer view.

lorban:
I've been thinking that we could add some statistics collection to the pools that would record the monitored buffer sizes requested vs the ones served for each bucket. This would help figuring out how to tune the bucket sizes.

gregw:
@lorban So my thinking on sizing/algorithm goes likes this:
 
1. 1024 factor with max 64K linear is not great:
    - 64 buckets per sub-pool for 256 buckets in total... lot's of scope for buffers to get lost/left in seldom used buckets. 
    - the 2 retained sub-pools are always fully populated, so that's 128 buckets, most of which will very rarely get used.
    - we do ask for small buffers (eg 16bytes for a HTTP/1.1 chunk), so 1024 is a lot of wasted space for that.
2. 4096 factor with max 64K linear is better and worse:
    - 16 buckets per sub-pool for 64 in total is more manageable.
    - 32 retained buckets is better
    - but 4096 for a small buffer is much worse
3. log buckets with min 0, max 64K gives buckets 1,2,4,8,16,32,64,128,256,512,1K,2K,4K,8K,16K,32K,64K 
    - that's 17 buckets per sub-pool for 68 in total - OK
    - 34 retained buckets are always created, which sucks a little
    - small buffers are well handled.... perhaps too well
    - mid size buffers can be wasteful, as in the SSL example of 16K+1 getting a 32K buffer
4.  log buckets with min 16, max 64K gives buckets 16,32,64,128,256,512,1K,2K,4K,8K,16K,32K,64K 
    - that's 13 buckets per sub-pool for 52 in total - OK
    - 26 retained buckets are always created, which sucks a little less
    - small buffers are handled OK .... perhaps too well
    - mid size buffers can be wasteful, as in the SSL example of 16K+1 getting a 32K buffer
5. We could consider custom buckets that are non-linear and non-log. For example: 16, 1K, 8K, 16K, 17K, 32K, 64K 
    - 7 buckets per sub-pool, 28 in total is good
    - only 14 retained buckets always created
    - small buffers are handled OK
    - mid size buffers can be optimised by hand
    - bucket lookup might be a little more expensive, but could be optimized with a lookup table.

Without stats, I think options 2, 3 or 4 are all OK, with a slight preference for 4.  We'd need stats to say if 5 is worth it or if yet another strategy would be better.

lorban:
@gregw with the default memory limit being `HeapSize / 4` I agree that options 4 is reasonable even if SSL would be a bit wasteful.

Option 5 is IIRC how kernel slab allocation is usually done as most sizes are known at compilation time. If we had stats, we could go that route but without such guidance, I'd prefer to rely on either linear or log.

jetty-io/src/main/java/org/eclipse/jetty/io/AbstractByteBufferPool.java
```diff
@@ -29,37 +34,79 @@
 @ManagedObject
 abstract class AbstractByteBufferPool implements ByteBufferPool
 {
+    public static final Integer DEFAULT_FACTOR = 4096;
+    public static final Integer DEFAULT_MAX_CAPACITY_BY_FACTOR = 16;
```
lorban: Should be `int`, shouldn't it?

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
```diff
@@ -57,6 +57,18 @@ public ArrayByteBufferPool()
...
+    /**
+     * Creates a new ArrayByteBufferPool with a default configuration.
+     * Both {@code maxHeapMemory} and {@code maxDirectMemory} default to 0 to use default heuristic.
+     * @param retainable True if {@link #asRetainableByteBufferPool()} return a real implementation.
+     */
+    public ArrayByteBufferPool(boolean retainable)
+    {
+        this(-1, -1, -1, -1, -1, -1,
+            retainable ? -1 : -2,
+            retainable ? -1 : -2);
```
lorban: Why are you changing the defaults from 0 (use heuristic) to -1 (unlimited)?

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
```diff
@@ -67,7 +79,7 @@ public ArrayByteBufferPool()
      */
     public ArrayByteBufferPool(int minCapacity, int factor, int maxCapacity)
     {
-        this(minCapacity, factor, maxCapacity, -1, 0, 0);
+        this(minCapacity, factor, maxCapacity, -1, -1, -1);
```
lorban: Why are you changing the defaults from 0 (use heuristic) to -1 (unlimited)?

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
```diff
@@ -81,7 +93,7 @@ public ArrayByteBufferPool(int minCapacity, int factor, int maxCapacity)
      */
     public ArrayByteBufferPool(int minCapacity, int factor, int maxCapacity, int maxQueueLength)
     {
-        this(minCapacity, factor, maxCapacity, maxQueueLength, 0, 0);
+        this(minCapacity, factor, maxCapacity, maxQueueLength, -1, -1);
```
lorban: Why are you changing the defaults from 0 (use heuristic) to -1 (unlimited)?

jetty-io/src/main/java/org/eclipse/jetty/io/MappedByteBufferPool.java
```diff
@@ -83,11 +83,39 @@ public MappedByteBufferPool(int factor, int maxQueueLength)
      * @param maxQueueLength the maximum ByteBuffer queue length
      * @param newBucket the function that creates a Bucket
      */
-    public MappedByteBufferPool(int factor, int maxQueueLength, Function<Integer, Bucket> newBucket)
+    private MappedByteBufferPool(int factor, int maxQueueLength, Function<Integer, Bucket> newBucket)
     {
-        this(factor, maxQueueLength, newBucket, 0, 0);
+        this(factor, maxQueueLength, newBucket, -1, -1, -1, -1);
```
lorban: Why changing the default from use heuristic to unlimited?

jetty-io/src/main/java/org/eclipse/jetty/io/NullByteBufferPool.java
```diff
@@ -19,6 +19,8 @@
 
 public class NullByteBufferPool implements ByteBufferPool
 {
+    private RetainableByteBufferPool _retainableByteBufferPool = RetainableByteBufferPool.from(this);
```
lorban: Please make it `final`.

lorban (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
gregw: Because I thought I had already changed it from -1 to 0.....  I'll double check what it was originally.

gregw: The original and current null constructor uses -1, so this new constructor also uses -1.   Other cases below are more interesting.

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
gregw: OK,  this was 0, which is different to the null constructor, so it was inconsistent.   I'll revert to 0 for now, but perhaps we should make consistent in jetty-12?

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
gregw: again will revert to 0, but this is inconsistent

jetty-io/src/main/java/org/eclipse/jetty/io/MappedByteBufferPool.java
gregw: reverted

jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTLSTest.java
```diff
@@ -708,22 +708,65 @@ protected int networkFill(ByteBuffer input) throws IOException
...
+        {
+            super(0, factor, maxCapacity, maxBucketSize, retainedHeapMemory, retainedDirectMemory);
+            _pool = pool;
+        }
+
+        @Override
+        protected ByteBuffer allocate(int capacity)
+        {
+            System.err.println("allocate " + capacity);
+            new Throwable().printStackTrace();
```
lorban: prints should go.

jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTLSTest.java
```diff
@@ -708,22 +708,65 @@ protected int networkFill(ByteBuffer input) throws IOException
...
+            return _pool.acquire(capacity, false);
+        }
+
+        @Override
+        protected ByteBuffer allocateDirect(int capacity)
+        {
+            System.err.println("allocateDirect " + capacity);
+            new Throwable().printStackTrace();
```
lorban: prints should go.

jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTLSTest.java
```diff
@@ -916,6 +963,7 @@ protected void service(String target, Request jettyRequest, HttpServletRequest r
 
         assertThrows(Exception.class, () -> client.newRequest("localhost", connector.getLocalPort()).scheme(HttpScheme.HTTPS.asString()).send());
 
+        byteBufferPool.asRetainableByteBufferPool().clear();
         await().atMost(5, TimeUnit.SECONDS).until(() -> leakedBuffers, is(empty()));
```
lorban, line 963: I just noticed something unrelated to your change: the `leakedBuffers` variable should have a `CopyOnWriteArrayList`.

jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTLSTest.java
```diff
@@ -834,6 +880,7 @@ protected boolean networkFlush(ByteBuffer output) throws IOException
...
```
lorban, line 880: I just noticed something unrelated to your change: the `leakedBuffers` variable should have a `CopyOnWriteArrayList`.

jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTLSTest.java
```diff
@@ -998,6 +1046,7 @@ protected void service(String target, Request jettyRequest, HttpServletRequest r
...
```
lorban, line 1046: I just noticed something unrelated to your change: the `leakedBuffers` variable should have a `CopyOnWriteArrayList`.

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
lorban: The implementation doesn't match what the javadoc says, and I checked what the current code does and it does default to use the heuristic in all constructors where the memory sizes aren't specified. The -1 in t he no-arg constructor do not refer to the memory sizes and used to call a constructor that defaulted them to 0.

lorban (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
gregw: Ah yeah.  Anyway, this is currently not used, so I'm deleting it.

lorban (approved):

PR #8177, opened 2022-06-20
exclude some infinispan dependencies to avoid problem with m-enforcer-p

lachlan-roberts:
the fix from #8113 didn't work for me on `9.4.x` but with these changes I can get a green build

- exclude `org.apache.logging.log4j:log4j-core-java9`
- exclude `org.apache.logging.log4j:log4j-api-java9`

### DIFF
--- a/jetty-infinispan/infinispan-remote-query/pom.xml
+++ b/jetty-infinispan/infinispan-remote-query/pom.xml
@@ -107,11 +107,7 @@
         <!-- this is depending on an old log4j version which have this issue https://issues.apache.org/jira/browse/LOG4J2-3241 -->
         <exclusion>
           <groupId>org.apache.logging.log4j</groupId>
-          <artifactId>log4j-core-java9</artifactId>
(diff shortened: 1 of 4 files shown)


jetty-infinispan/infinispan-remote-query/pom.xml
```diff
@@ -107,7 +107,11 @@
         <!-- this is depending on an old log4j version which have this issue https://issues.apache.org/jira/browse/LOG4J2-3241 -->
         <exclusion>
           <groupId>org.apache.logging.log4j</groupId>
-          <artifactId>log4j-core</artifactId>
+          <artifactId>log4j-core-java9</artifactId>
```
olamy: maybe just use a single exclusion line such
```
        <exclusion>
          <groupId>org.apache.logging.log4j</groupId>
          <artifactId>*</artifactId>
        </exclusion>
```

joakime (approved):

PR #8187, opened 2022-06-21
Fix test-distribution classpath re resolver

cstamas:
As currently it seems mixed versions of resolver artifacts are
present on classpath. Resolver 1.8.0 introduces a binary
breakage against 1.6.x.

Updated to bugfix 1.8.1 resolver

Anyway, this fixes the thing, as impl was actually coming in
as transitive dep of maven-resolver-provider, that is NOT
the same version as the rest of resolver dependencies.

Relying on transitive while fixing other versions is wrong.

image

### DIFF
--- a/tests/test-distribution/pom.xml
+++ b/tests/test-distribution/pom.xml
@@ -50,6 +50,11 @@
     </dependency>
     <dependency>
       <groupId>org.apache.maven.resolver</groupId>
+      <artifactId>maven-resolver-impl</artifactId>
+      <version>${maven.resolver.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven.resolver</groupId>
       <artifactId>maven-resolver-connector-basic</artifactId>
       <version>${maven.resolver.version}</version>
     </dependency>
--- a/tests/test-distribution/pom.xml
+++ b/tests/test-distribution/pom.xml
@@ -52,6 +52,13 @@
       <groupId>org.apache.maven.resolver</groupId>
       <artifactId>maven-resolver-impl</artifactId>
       <version>${maven.resolver.version}</version>
+      <exclusions>
+        <!-- Used when running in SISU container to manage beans lifecycle (of locking factories) -->
+        <exclusion>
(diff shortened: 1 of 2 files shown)


tests/test-distribution/pom.xml
```diff
@@ -48,6 +48,18 @@
       <artifactId>maven-resolver-spi</artifactId>
       <version>${maven.resolver.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.maven.resolver</groupId>
+      <artifactId>maven-resolver-impl</artifactId>
+      <version>${maven.resolver.version}</version>
+      <exclusions>
+        <!-- Used when running in SISU container to manage beans lifecycle (of locking factories) -->
```
joakime, line 56: I don't understand this comment.

cstamas: https://github.com/apache/maven-resolver/blob/master/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/DefaultSyncContextFactory.java#L84

Sisu container is able to manage lifecycle, and on container shutdown will invoke this method. For your case is not needee, as you craft resolver manually, but also do not use any special kind of named locks that would require this cleanup (you use defaults).

cstamas:
Strange, this PR now made module far-far before change to break and seems completely unrelated

```
[ERROR] Failures: 
[ERROR]   HttpClientRedirectTest.testRedirectFailed:303 Unexpected exception type thrown ==> expected: <java.util.concurrent.ExecutionException> but was: <java.util.concurrent.TimeoutException>
[ERROR]   HttpClientRedirectTest.testRedirectFailed:303 Unexpected exception type thrown ==> expected: <java.util.concurrent.ExecutionException> but was: <java.util.concurrent.TimeoutException>
[INFO] 
[ERROR] Tests run: 655, Failures: 2, Errors: 0, Skipped: 15
```

joakime (approved):
LGTM

cstamas:
FTR, created https://github.com/eclipse/sisu.inject/issues/58

PR #8200, opened 2022-06-23
Jetty 12.0.x eex jettymavenplugin

janbartel:
Make ee9 and ee10 jetty maven plugins work.

Some work still to do to see if we can refactor out common code to reduce duplication, but this is the first step to have working plugins.


jetty-ee10/jetty-ee10-maven-plugin/pom.xml
```diff
@@ -180,192 +114,93 @@
...
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-slf4j-impl</artifactId>
       <scope>test</scope>
     </dependency>
-    <!-- TODO
-    <dependency>
-      <groupId>org.eclipse.jetty</groupId>
-      <artifactId>jetty-home</artifactId>
```
gregw, line 257: Like it!

gregw (approved):
I bit hard to review as it is so large.  But Iike the dependency/build structure a lot better, so if you can get a green CI build, then LGTM!

Issue #8206, opened 2022-06-24
Stopping server from within AbstractConnector#accept fails and results in a partially stopped QueuedThreadPool


kohlschuetter:
**Jetty version(s)**
11.0.11 (also affecting older versions, such as 9.4.12)

**Java version/vendor** `(use: java -version)`
openjdk version "18" 2022-03-22
OpenJDK Runtime Environment Temurin-18+36 (build 18+36)
OpenJDK 64-Bit Server VM Temurin-18+36 (build 18+36, mixed mode)
and others

**OS type/version**
macOS 12.4

**Description**
A Server instance using `QueuedThreadPool` may shutdown incompletely (preventing the process from terminating) upon a call to Server#stop from within `AbstractConnector#accept`. This is due to the fact that two calls in `QueuedThreadPool#doStop` may throw `InterruptedException` (`joinThreads`), preventing the method to complete as expected.

**How to reproduce?**

```
public class QTPTest {
  public static void main(String[] args) throws Exception {
    Server s1 = newServer();

    try (Socket sock = new Socket()) {
      // triggers AbstractConnector#accept and thus, Server#stop.
      // see below.
      sock.connect(new InetSocketAddress("localhost", 12345));
    }

    s1.join(); // wait until Server stops

    System.out.println("The end."); // not reached without fix
  }

  private static Server newServer() throws Exception {
    Server server = new Server();

    ServerConnector connector = new ServerConnector(server) {

      @Override
      public void accept(int acceptorID) throws IOException {
        super.accept(acceptorID);

        // trigger stop from within accept

        Runnable stopRunnable = () -> {
          try {
            System.out.println("STOPPING: " + server);
            server.stop();
          } catch (Exception e) {
            e.printStackTrace();
          }
        };

        // stopRunnable.run(); // doesn't work
        // throws an InterruptedException via ManagedSelector.doStop(ManagedSelector.java:137)

        // works but creating new Threads may be discouraged by policy
        // new Thread(stopRunnable).start();

        // works but requires a new ExecutorService and shutdown logic
        // ExecutorService es = Executors.newSingleThreadExecutor();
        // es.execute(stopRunnable);
        // es.shutdown();

        getExecutor().execute(stopRunnable); // gets a bit farther, but (without fix) throws an
                                             // InterruptedException in
                                             // QueuedThreadPool.joinThreads(QueuedThreadPool.java:324)
      }

    };
    connector.setHost("localhost");
    connector.setPort(12345);
[...]

PR #8207, opened 2022-06-24
QueuedThreadPool: Prevent incomplete stop due to InterruptedException

kohlschuetter:
QueuedThreadPool#doStop may fail to execute completely if a thread
throws InterruptedException upon Thread#join.

Fixes https://github.com/eclipse/jetty.project/issues/8206

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
@@ -320,8 +320,13 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
             long canWait = TimeUnit.NANOSECONDS.toMillis(stopByNanos - System.nanoTime());
             if (LOG.isDebugEnabled())
                 LOG.debug("Waiting for {} for {}", thread, canWait);
-            if (canWait > 0)
-                thread.join(canWait);
+            if (canWait > 0) {
+                try {
+                    thread.join(canWait);
+                } catch (InterruptedException e) {
+                    // ignore
+                }
+            }
         }
     }
 


jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -320,8 +320,13 @@ private void joinThreads(long stopByNanos) throws InterruptedException
             long canWait = TimeUnit.NANOSECONDS.toMillis(stopByNanos - System.nanoTime());
             if (LOG.isDebugEnabled())
                 LOG.debug("Waiting for {} for {}", thread, canWait);
-            if (canWait > 0)
-                thread.join(canWait);
+            if (canWait > 0) {
+                try {
+                    thread.join(canWait);
+                } catch (InterruptedException e) {
+                    // ignore
```
joakime, line 327: It was interrupted, which means the thread still exists, and has the interrupted flag set.
This seems to bypass the `interrupted()` cleanup in the QueuedThreadPool.
Not 100% sold on this.
I appreciate that you have protection so that 1 bad `Thread.join()` doesn't prevent other threads from completing, but there's something off about this approach. (Can't put my finger on it)

I think, historically speaking, this is intended to be a speculative join, where it's allowed to fail.  But I could be misremembering things.

gregw: I'm onboard with handling interruptedException here, but I agree with @joakime that ignoring it does not look correct.
Perhaps we should just join on the thread again?   We could do this in a loop or once only?  

Eitherway, we'd need a unit test in this PR.  @kohlschuetter do you wish to keep working on this or do  you want us to take it over?

kohlschuetter: @gregw Feel free to take over from here, thanks!

gregw:
closing in favour of #8220

Issue #8211, opened 2022-06-27
Jetty-12 already complete exception


gregw:
**Jetty version(s)**
jetty-12

**Java version/vendor** `(use: java -version)`

**OS type/version**

**Description**

When building with tests, the following exception is produced many times from many tests.  This should either not happen, or if it is normal (??) then the log should be stackless at least.

```
java.lang.IllegalStateException: already completed
	at org.eclipse.jetty.server.internal.HttpChannelState$ChannelCallback.lockedOnComplete(HttpChannelState.java:1422)
	at org.eclipse.jetty.server.internal.HttpChannelState$ChannelCallback.failed(HttpChannelState.java:1367)
	at org.eclipse.jetty.server.internal.HttpChannelState$HandlerInvoker.run(HttpChannelState.java:666)
	at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.runTask(AdaptiveExecutionStrategy.java:412)
	at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.consumeTask(AdaptiveExecutionStrategy.java:381)
	at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.tryProduce(AdaptiveExecutionStrategy.java:268)
	at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.produce(AdaptiveExecutionStrategy.java:190)
	at org.eclipse.jetty.http2.internal.HTTP2Connection.produce(HTTP2Connection.java:209)
	at org.eclipse.jetty.http2.internal.HTTP2Connection.onFillable(HTTP2Connection.java:156)
	at org.eclipse.jetty.http2.internal.HTTP2Connection$FillableCallback.succeeded(HTTP2Connection.java:379)
	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:100)
	at org.eclipse.jetty.io.SelectableChannelEndPoint$1.run(SelectableChannelEndPoint.java:53)
	at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.runTask(AdaptiveExecutionStrategy.java:412)
	at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.consumeTask(AdaptiveExecutionStrategy.java:381)
	at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.tryProduce(AdaptiveExecutionStrategy.java:268)
[...]

PR #8215, opened 2022-06-28
Issue #8211 - fix ISE from HttpChannelState if failure during process

lachlan-roberts:
If a timeout or stream reset occurs during process, the `HttpChannelState` will be failed, this causes `callback.succeeded()` to throw ISE. This ISE is then caught in `HandlerInvoker` and the callback is failed, and because the callback was failed twice we get the `java.lang.IllegalStateException: already completed` seen in #8211.

This PR fixes this by 
- only failing the callback in `HandlerInvoker` if `HttpChannel` has not been completed.
- Do not throw from `callback.failed()` if the callback is failed with same cause as why the `HttpChannel` was failed.


jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -663,7 +663,15 @@ public void run()
...
-                request._callback.failed(failure);
+            {
+                try (AutoLock ignored = _lock.lock())
+                {
+                    if (request._httpChannel != null && !request._httpChannel._completed)
+                        request._callback.failed(failure);
+                    else if (LOG.isDebugEnabled())
+                        LOG.debug("Process failed", failure);
+                }
+            }
```
gregw: Potentially it would be neater to put this logic into a method:
```suggestion
            request._callback.ensureCompleted(failure);
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -1364,6 +1372,9 @@ public void failed(Throwable failure)
             boolean completeStream;
             try (AutoLock ignored = _request._lock.lock())
             {
+                request = _request;
+                if (request._httpChannel != null && request._httpChannel._error != null && TypeUtil.isAssociated(request._httpChannel._error.getCause(), failure))
+                    return;
```
gregw: Need a comment here about the fact we are being forgiving of lazy applications that doubly report a failure.
```suggestion
                // Be forgiving of applications that report a failure that is already known by treating as a noop.
                if (request._httpChannel != null && request._httpChannel._error != null && TypeUtil.isAssociated(request._httpChannel._error.getCause(), failure))
                    return;
```

gregw (approved):
I think this LG, but as I'm kind of a co-author, we should get @sbordet's thoughts as well.

gregw (dismissed):
oops I approved when I wanted to request some small changes.

jetty-core/jetty-http2/jetty-http2-tests/src/test/java/org/eclipse/jetty/http2/tests/IdleTimeoutTest.java
```diff
@@ -384,6 +386,77 @@ public boolean onIdleTimeout(Stream stream, Throwable x)
         assertFalse(((HTTP2Session)session).isDisconnected());
     }
 
+    @Test
+    public void testForgivingFailedClientEnforcingStreamIdleTimeout() throws Exception
+    {
+        final int idleTimeout = 1000;
```
sbordet: No need for `final`, here and below.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -1317,6 +1319,10 @@ public void succeeded()
             boolean completeStream;
             try (AutoLock ignored = _request._lock.lock())
             {
+                if (_completed)
+                    return;
+                _completed = true;
```
sbordet: I would not do this check here.
The call below `lockedOnComplete()` is the one responsible for setting `_completed` to `true`, and throwing if it's already set.
I would: A) make `lockedOnComplete()` return a `boolean`, and if it is `false` return; and B) remove the code in `lockedOnComplete()` that throws, which is now never executed.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -1377,6 +1383,10 @@ public void failed(Throwable failure)
...
```
sbordet: Ditto.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -663,8 +663,9 @@ public void run()
             {
                 failure = x;
             }
+
             if (failure != null)
-                request._callback.failed(failure);
+                _request._callback.ensureCompleted(failure);
```
sbordet: I am not sure I understand this. With the current modifications, the original code calling `failed()` is already doing the check on `_completed`, so this method does nothing different, and certainly does not "ensure" and it only fails the callback, while the name "ensureCompleted" seems to hint that it either succeeds or fails it.
I would restore the old code.

sbordet (changes requested):

sbordet:
@lachlan-roberts I have pushed changes that align to my comments, so I'll merge this PR once build passes.

PR #8217, opened 2022-06-28
Go back to simpler Jenkinsfile

joakime:
Lets use a simpler Jenkinsfile for jetty 12 now that we've gotten past most of the does-not-compile issues.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,132 +7,37 @@ pipeline {
   // save some io during the build
   options { durabilityHint('PERFORMANCE_OPTIMIZED') }
   stages {
-    stage("Checkout Jetty") {
-      steps {
-        container('jetty-build') {
-          dir("${env.WORKSPACE}/buildy") {
-            checkout scm
-          }
-        }
-      }
-    }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -7,132 +7,37 @@ pipeline {
...
-              timeout(time: 120, unit: 'MINUTES') {
-                dir("${env.WORKSPACE}/buildy") {
-                  //cleanup all projects
-                  mavenBuild("jdk17", "clean -T4", "maven3")
-                  mavenBuild("jdk17", "clean install -f jetty-ee8", "maven3")
-                }
+              timeout( time: 180, unit: 'MINUTES' ) {
+                mavenBuild( "jdk17", "clean install -Perrorprone", "maven3")
+                // Collect up the jacoco execution results (only on main build)
+//                jacoco inclusionPattern: '**/org/eclipse/jetty/**/*.class',
```
joakime, line 19: Jacoco needs to be fixed, but in a different PR.

gregw (approved):

Issue #7833, opened 2022-04-04
Jetty injects `Content-Length: 0` into HEAD responses


thecodewarrior:
**Jetty version(s)** 

10.0.7

**Java version/vendor** `(use: java -version)`

```
openjdk version "17.0.1" 2021-10-19 LTS
OpenJDK Runtime Environment Corretto-17.0.1.12.1 (build 17.0.1+12-LTS)
OpenJDK 64-Bit Server VM Corretto-17.0.1.12.1 (build 17.0.1+12-LTS, mixed mode, sharing)
```

**OS type/version**

macOS 12.3

**Description**

Jetty always sends a `Content-Length` header in `HEAD` requests. As per [RFC 7231 §4.3.2](https://httpwg.org/specs/rfc7231.html#HEAD) they should send the same headers as a `GET` request would have sent. Inferring the `Content-Length` makes the correct behavior impossible when the content length is indeterminate.

From my investigation I believe it's because [`HttpGenerator` doesn't consider them "no content responses"](https://github.com/eclipse/jetty.project/blob/cb12779/jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java#L388-L402), so it infers the `Content-Length` from the body size. If we can't access the request in that class when generating the response, at least providing a way to indicate that jetty shouldn't infer a content length (e.g. by setting the length to -2) would be sufficient. 

**How to reproduce?**
- Create a Jetty `Handler` which sends a large payload (larger than jetty's buffer, so it gets chunked) without setting a `Content-Length` header. Make sure to skip the actual payload when receiving a `HEAD` request.
- Perform a `GET` request and note that there is no `Content-Length` header in the response
- Perform a `HEAD` request and note that the response contains `Content-Length: 0`

joakime:
Section https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.2 doesn't mention anything about adding a `Content-Length` header if it's missing.

The use of `Transfer-Encoding: chunked` by a response should be preserved, and no `Content-Length` header should be required to be present on the HEAD response if the actual response is chunked.

I have verified this is a bug, and the behavior is present in branch `jetty-9.4.x` onwards (see commit https://github.com/eclipse/jetty.project/commit/08344162f9bea1ffda0b66aac3f91a2cc05030f1)

joakime:
Interesting...

If the `Transfer-Encoding: chunked` is determined via hinting, the test fails.
[...]

PR #8219, opened 2022-06-28
Issue #7833 - Chunked Transfer-Encoding and HEAD response test

joakime:
Tests for #7833

### DIFF
--- a/jetty-webapp/src/test/java/org/eclipse/jetty/webapp/HugeResourceTest.java
+++ b/jetty-webapp/src/test/java/org/eclipse/jetty/webapp/HugeResourceTest.java
@@ -421,6 +421,7 @@ public class HugeResourceTest
             try (InputStream input = resource.openStream())
             {
                 resp.setContentType("application/octet-stream");
+                resp.flushBuffer();
                 IO.copy(input, output);
             }
         }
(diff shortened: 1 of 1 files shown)


jetty-webapp/src/test/java/org/eclipse/jetty/webapp/HugeResourceTest.java
```diff
@@ -359,6 +411,21 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws I
...
+    {
+        @Override
+        protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException
+        {
+            URL resource = req.getServletContext().getResource(req.getPathInfo());
+            OutputStream output = resp.getOutputStream();
+            try (InputStream input = resource.openStream())
+            {
+                resp.setContentType("application/octet-stream");
+                IO.copy(input, output);
```
gregw: If you wanted to ensure chunking is used regardless of the relative size of the content to the buffer:

```suggestion
                resp.setContentType("application/octet-stream");
                resp.flushBuffers();
                IO.copy(input, output);
```

gregw (dismissed):
Other than a niggle... LGTM

jetty-webapp/src/test/java/org/eclipse/jetty/webapp/HugeResourceTest.java
joakime: Changed.

gregw (approved):

PR #8220, opened 2022-06-29
Interrupted QTP.stop #8206

### LINKED ISSUE #8206: Stopping server from within AbstractConnector#accept fails and results in a partially stopped QueuedThreadPool

gregw:
Fix #8206 
Replaces #8207
Made QTP stop resilient to spurious and self interrupts. 

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
@@ -243,9 +243,8 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
         {
             // Fill the job queue with noop jobs to wakeup idle threads.
             for (int i = 0; i < threads; ++i)
-            {
-                jobs.offer(NOOP);
-            }
+                if (!jobs.offer(NOOP))
+                    break;
 
             // try to let jobs complete naturally for half our stop time
             joinThreads(System.nanoTime() + TimeUnit.MILLISECONDS.toNanos(timeout) / 2);
@@ -255,6 +254,8 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
             // interrupt remaining threads
             for (Thread thread : _threads)
             {
+                if (thread == Thread.currentThread())
+                    continue;
                 if (LOG.isDebugEnabled())
                     LOG.debug("Interrupting {}", thread);
                 thread.interrupt();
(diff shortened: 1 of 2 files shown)


joakime (approved):
LGTM

PR #8223, opened 2022-06-29
Issue #8222 - Addressing NPE with JvmArgs and non-existent modules.

### LINKED ISSUE #8222: Jetty start.jar fails with NullPointerException when referencing a non existent module and using JVM args
**Jetty version(s)** 10.0.11

**Java version/vendor** `openjdk 11.0.15+10` 

**Description**

Running webapp with Jetty distribution fails with a null pointer exception:

```
> java -jar /home/chat/jetty/start.jar
java.lang.NullPointerException
	at org.eclipse.jetty.start.Main.lambda$start$1(Main.java:476)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:176)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
	at org.eclipse.jetty.start.Main.start(Main.java:478)
	at org.eclipse.jetty.start.Main.main(Main.java:75)

Usage: java -jar $JETTY_HOME/start.jar [options] [properties] [configs]
       java -jar $JETTY_HOME/start.jar --help  # for more information
```

We are upgrading from Jetty `9.4.45.v20220203` and reusing the same `.ini` files. It seems that `continuation` module has been removed. 

Reference: https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-start/src/main/java/org/eclipse/jetty/start/Main.java#L474:L476

My suspicion is that line 474 returns null because module is not found. It would be nice to throw an Exception when module is not found to help figuring out the problem.

**How to reproduce?**

1. Add continuation module to `start.ini`:
```
# ---------------------------------------
# Module: continuation
--module=continuation
```
2. Run Jetty distribution `java -jar /jetty/start.jar`

joakime:

### DIFF
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
@@ -188,7 +188,7 @@ public class Main
 
     public void invokeMain(ClassLoader classloader, StartArgs args) throws IllegalAccessException, InvocationTargetException, NoSuchMethodException, IOException
     {
-        if (args.getEnabledModules().isEmpty())
+        if (args.getSelectedModules().isEmpty())
         {
             if (Files.exists(getBaseHome().getBasePath("start.jar")))
                 StartLog.error("Do not start with ${jetty.base} == ${jetty.home}!");
@@ -326,7 +326,7 @@ public class Main
         modules.registerAll();
 
         // 4) Active Module Resolution
-        List<String> selectedModules = args.getEnabledModules();
+        List<String> selectedModules = args.getSelectedModules();
         List<String> sortedSelectedModules = modules.getSortedNames(selectedModules);
         List<String> unknownModules = new ArrayList<>(selectedModules);
         unknownModules.removeAll(sortedSelectedModules);
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
@@ -481,7 +481,7 @@ public class Main
             CommandLineBuilder cmd = args.getMainArgs(StartArgs.ALL_PARTS);
             cmd.debug();
 
-            List<String> execModules = args.getValidEnabledModulesStream()
+            List<String> execModules = args.getAllModules().getEnabled().stream()
                 // Keep only the forking modules.
                 .filter(module -> !module.getJvmArgs().isEmpty())
                 .map(Module::getName)
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
@@ -29,6 +29,7 @@ import java.net.Socket;
 import java.net.SocketTimeoutException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 import java.util.stream.Collectors;
@@ -325,7 +326,17 @@ public class Main
         modules.registerAll();
 
         // 4) Active Module Resolution
-        for (String enabledModule : modules.getSortedNames(args.getEnabledModules()))
+        List<String> enabledModules = args.getEnabledModules();
+        List<String> sortedEnabledModules = modules.getSortedNames(enabledModules);
+        List<String> unknownModules = new ArrayList<>(enabledModules);
+        unknownModules.removeAll(sortedEnabledModules);
+        if (unknownModules.size() >= 1)
+        {
+            throw new UsageException(UsageException.ERR_UNKNOWN, "Unknown module%s=[%s] List available with --list-modules",
+                unknownModules.size() > 1 ? 's' : "",
+                String.join(", ", unknownModules));
+        }
+        for (String enabledModule : sortedEnabledModules)
         {
             for (String source : args.getSources(enabledModule))
             {
@@ -470,8 +481,7 @@ public class Main
             CommandLineBuilder cmd = args.getMainArgs(StartArgs.ALL_PARTS);
             cmd.debug();
 
-            List<String> execModules = args.getEnabledModules().stream()
-                .map(name -> args.getAllModules().get(name))
+            List<String> execModules = args.getValidEnabledModulesStream()
                 // Keep only the forking modules.
                 .filter(module -> !module.getJvmArgs().isEmpty())
                 .map(Module::getName)
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
@@ -326,22 +326,22 @@ public class Main
         modules.registerAll();
 
         // 4) Active Module Resolution
-        List<String> enabledModules = args.getEnabledModules();
-        List<String> sortedEnabledModules = modules.getSortedNames(enabledModules);
-        List<String> unknownModules = new ArrayList<>(enabledModules);
-        unknownModules.removeAll(sortedEnabledModules);
+        List<String> selectedModules = args.getEnabledModules();
+        List<String> sortedSelectedModules = modules.getSortedNames(selectedModules);
+        List<String> unknownModules = new ArrayList<>(selectedModules);
+        unknownModules.removeAll(sortedSelectedModules);
         if (unknownModules.size() >= 1)
         {
             throw new UsageException(UsageException.ERR_UNKNOWN, "Unknown module%s=[%s] List available with --list-modules",
                 unknownModules.size() > 1 ? 's' : "",
                 String.join(", ", unknownModules));
         }
-        for (String enabledModule : sortedEnabledModules)
+        for (String selectedModule : sortedSelectedModules)
         {
-            for (String source : args.getSources(enabledModule))
+            for (String source : args.getSources(selectedModule))
             {
                 String shortForm = baseHome.toShortForm(source);
-                modules.enable(enabledModule, shortForm);
+                modules.enable(selectedModule, shortForm);
             }
         }
 
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
@@ -31,14 +31,12 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
 import java.util.StringTokenizer;
 import java.util.jar.Attributes;
 import java.util.jar.Manifest;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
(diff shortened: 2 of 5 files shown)


jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
```diff
@@ -646,11 +648,27 @@ public Classpath getClasspath()
...
     {
         return this.modules;
     }
 
+    /**
+     * Get the Stream of {@link Module} instances that have been enabled successfully.
+     * Filtering out any that are invalid (such as modules that were not, or cannot be enabled).
+     * @return the stream of {@link Module} instances
+     */
+    public Stream<Module> getValidEnabledModulesStream()
```
joakime: This is a convenience method that makes usage predictable.
This addressed the NPE directly.

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
```diff
@@ -646,11 +648,27 @@ public Classpath getClasspath()
         return classpath;
     }
 
+    /**
+     * The list of selected Modules to enable based on configuration.
+     * @return the list of selected modules (by name) that the configuration is requesting.
+     */
```
joakime, line 651: Added javadoc to explain what this list actually is.

jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
```diff
@@ -325,12 +326,22 @@ public StartArgs processCommandLine(String[] cmdLine) throws Exception
         modules.registerAll();
 
         // 4) Active Module Resolution
-        for (String enabledModule : modules.getSortedNames(args.getEnabledModules()))
+        List<String> selectedModules = args.getEnabledModules();
+        List<String> sortedSelectedModules = modules.getSortedNames(selectedModules);
+        List<String> unknownModules = new ArrayList<>(selectedModules);
+        unknownModules.removeAll(sortedSelectedModules);
```
joakime, line 332: The use of `modules.getSortedNames(List<String>)` actually left out unknown modules from the returned list.
This filtering to show unknown modules is how we now error out when an unknown module is selected to be enabled.

jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
```diff
@@ -470,8 +481,7 @@ else if (args.isCreateFiles() || !args.getStartModules().isEmpty())
             CommandLineBuilder cmd = args.getMainArgs(StartArgs.ALL_PARTS);
             cmd.debug();
 
-            List<String> execModules = args.getEnabledModules().stream()
-                .map(name -> args.getAllModules().get(name))
+            List<String> execModules = args.getValidEnabledModulesStream()
```
joakime: Direct prevention of the NPE.
This is preventative, ideally we shouldn't be able to reach here with bad module names, as it should error out with a UsageException for modules that don't exist earlier.

gregw: This feels a bit artificial.  I'd prefer that we just modified `getEnabledModules` to never return null.

jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
```diff
@@ -325,12 +326,22 @@ public StartArgs processCommandLine(String[] cmdLine) throws Exception
         modules.registerAll();
 
         // 4) Active Module Resolution
-        for (String enabledModule : modules.getSortedNames(args.getEnabledModules()))
+        List<String> selectedModules = args.getEnabledModules();
```
gregw: To reduce confusion, can we keep to one verb. Use either selected or enabled, but not both.
```suggestion
        List<String> enabledModules = args.getEnabledModules();
```

gregw (changes requested):
mostly good, just a few niggles

jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
joakime: That's just it.

`getEnabledModules()` returns what the user requested to be enabled. aka: what they selected.
As the javadoc now says, this is the list of all `--module=<name>` entries in their configuration.
This is what they have asked for, their configuration, not what has actually been enabled. (which is often this selected list and the transitive modules)

IMO, we should not change that behavior.
It's also useful for the error message later, to show the whole list of unknown module names.

joakime: This should have been called something else, "getSelectedModuleNames" or "getRequestedModuleNames".
It's not the list of enabled modules, never has been.
We don't even update this list for full list of modules that includes the transitive modules.
That's what the `Modules` object is for.

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
joakime: No longer present

jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
joakime: The new method `StartArgs.getValidEnabledModulesStream() `is no longer present.

jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
joakime: See prior comment about what this method `StartArgs.getEnabledModules()` actually does. (also see javadoc)

gregw: When this variable is used, it is always the list of enabled modules.   The `sortedSelectedModules` and the `unknownModules` collections are created from what should be called `enabledModules`.

jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
```diff
@@ -470,8 +481,7 @@ else if (args.isCreateFiles() || !args.getStartModules().isEmpty())
             CommandLineBuilder cmd = args.getMainArgs(StartArgs.ALL_PARTS);
             cmd.debug();
 
-            List<String> execModules = args.getEnabledModules().stream()
-                .map(name -> args.getAllModules().get(name))
+            List<String> execModules = args.getAllModules().getEnabled().stream()
```
gregw, line 484: How is `args.getAllModules().getEnabled().stream()` different from `args.getEnabledModules().stream()`?

If it is different, then we have a naming problem.  If it is not different then just use `getEnabledModules()`

gregw (changes requested):
@joakime I'm not suggesting that we change behaviour.   I'm saying that we need to cleanup the implementation and/or the naming.

What is the difference between a "selected" module and an "enabled" module?
What is the difference between a module returned by `startArgs.getEnabledModule` and a module in `getAllModules` that returns true from `isEnabled`?  If there is no difference, then we should avoid duplicating state in the implemenation.  If there is a difference, then naming needs to reflect that difference.

If we can determine the set of enabled modules by simply filtering all modules on `isEnabled`, then do we actually need to maintain a collection of enabled modules in StartArgs?   If so, shouldn't it be called `enabledModules` rather than just `modules` etc.

I know this is a bit of scope increase for fixing just a NPE, but whilst we are looking at this, let's do it properly.

joakime:
`StartArgs` holds the configuration, in other words, what's in the `start.ini` and `start.d/*.ini` and command line.

It could, for example only have in it's `.getEnabledModules()` a List of type String with 2 entries `[ "http", "deploy" ]`.

The `Modules` class holds what has been enabled.
It's `Modules.getEnabled()` method (pre-existing, already there and doing the right thing) is a list of `Module` classes, which will return the complete list of enabled modules, based on the limited set provider in `StartArgs`, including all transitively enabled modules.
For the above `StartArgs` (`[ "http", "deploy" ]`), the `Modules.getEnabled()` will return the list `[bytebufferpool, resources, logging/slf4j, logging-jetty, threadpool, server, security, servlet, webapp, deploy, http]`.

The `StartArgs.getEnabledModules()` is the wrong API to use to build out the JvmArgs in the `Main` codepath for forked JVMs, as it will exclude any module specified JvmArgs as well.

The old, bad code, used `StartArgs.getEnabledModules()` in a stream, then mapped to a `Module`, then grabbed the JvmArgs, then filtered based on if the Module has JvmArgs, then mapped back to the name, to then get a list of module names used in the warning message for why start.jar is forking the JVM.

The new code, uses existing API properly here. (in Main for start with a forked JVM)
AND it also fails start up (with a UsageException) if a module name that doesn't exist is referenced with a complete list of referenced module names. (in Main for processCommandLine)
This last part used to be there, in the `Modules.enable(String, String)`, but the introduction of `modules.getSortedNames()` broke that location from triggering the UsageException.

gregw:
@joakime don't explain here. Explain in the code!   If  `StartArgs.getEnabledModules()` is the wrong API, the remove it or rename name it or at least document how it is different from `getAllModules().getEnabled()`.

Perhaps it is the case that the modules from start.ini and start.d/*.ini and command line should be called "selected" or "configured" rather than "enabled".    A module become enabled because it is selected/configured or is a dependency of a selected/configured module.

As I keep saying, I think the new code and handling of unknown modules is good.... but let's finish cleaning up the old code as well rather than leave two different concepts of what "enabled" means.

joakime:
@gregw done, code changed, javadoc added.

jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
joakime: I don't see this as a naming difference, just a "source of truth" difference.   
The `Modules` vs `StartArgs` are difference classes, with different purposes.

jetty-start/src/test/java/org/eclipse/jetty/start/fileinits/MavenMetadataTest.java
```diff
@@ -71,7 +71,7 @@ public void testIsExpiredTimestampNextWeek()
     @Test
     public void testIsExpiredTimestampNow()
     {
-        LocalDateTime now = LocalDateTime.now();
+        LocalDateTime now = LocalDateTime.now(ZoneId.of("UTC"));
```
joakime, line 74: This was a flaky test.
I just fixed it here as part of this PR.
It's really unrelated to the purpose of this PR, just happens to be in the same `jetty-start` code tree.

gregw (approved):
Much better!  thanks

Issue #8216, opened 2022-06-28
OpenID logout / more extensibible OpenIdConfiguration


keller-j:
**Target Jetty version(s)**
`10.0.x`
`11.0.x`
`12.0.x`

**Additional Details**
I'm currently working with `jetty-11` and available `OpenIdAuthenticator`. My OpenID (SSO) server is running on `keycloak-18` while trying to logout via client application (jetty) I realized that this isn't working at the moment. While having a look at `OpenIdConfiguraion` I realized that no information about `end_session_endpoint` isn't even processed in `OpenIdConfiguration`:
https://github.com/eclipse/jetty.project/blob/5fe999caa25755dec57d21bc3d694f67de4e45ab/jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdConfiguration.java#L119-#L130
At the moment I'm `extending OpenIdConfiguration` and loading `end_session_endpoint` information with a separate call similar to https://github.com/eclipse/jetty.project/blob/5fe999caa25755dec57d21bc3d694f67de4e45ab/jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdConfiguration.java#L141-#L172 . Then in an extended `OpenIdAuthenticator` I'm reacting to a certain logout page with redirecting user to `end_session_endpoint` or (depending on class configuration) also providing `id_token_hint` (e.g., available via `HttpSession`) and `post_logout_redirect_uri` as stated in OpenID documentation: https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout

**Enhancement Description**
1. For more easy extensibility of `OpenIdConfiguration` would be great if results fetched from OpenID configuration endpoint could be further processed for own needs by simply extending `OpenIdConfiguration` and implementing / overriding protected method, e.g., `processMetadata(Map<?, ?> fetchedMetadata)` which can be called in `doStart` with information of `discoveryDocument`:
https://github.com/eclipse/jetty.project/blob/5fe999caa25755dec57d21bc3d694f67de4e45ab/jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdConfiguration.java#L119 . Since there are lot of configuration settings not yet considered for current implementation possibility to simply process and store them along the configuration could come in handy.
2. `OpenIdAuthenticator` could also support logout from SSO. Therefore additional information must be processed in `OpenIdConfiguration` and considered in `OpenIdAuthenticator`.

lachlan-roberts:
[...]

PR #8224, opened 2022-06-29
Issue #8216 - OpenID logout

keller-j:
provide `logout`to `OpenIdAuthenticator` via `end_session_endpoint` information retrieved from `OpenIdConfiguration`

### DIFF
--- a/jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
+++ b/jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
@@ -16,6 +16,7 @@ package org.eclipse.jetty.security.openid;
 import java.io.IOException;
 import java.io.Serializable;
 import java.math.BigInteger;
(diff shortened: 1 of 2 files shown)


keller-j:
NOTE: this is just a sample implementation of logout functionality. Constructors were just copied and extended with new attributes. Also not quite sure whether it is correct to send `Authentication.SEND_CONTINUE` to response for logout redirection:
https://github.com/keller-j/jetty.project/blob/5282ca37ef7a81789b3c1d0a41854259898bce96/jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java#L608

Furthermore, also an option to redirect to `end_session_endpoint` with `id_token_hint` parameter set without supplying the `redirectAfterLogout` would be possible:
https://github.com/keller-j/jetty.project/blob/5282ca37ef7a81789b3c1d0a41854259898bce96/jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java#L596

Issue #8212, opened 2022-06-27
Jetty-12 flakey HTTP2 test


gregw:
**Jetty version(s)**

Flakey HTTP2 spec test in jetty-12

**Description**

Building locally on the command line, this is a frequent error. 

```
[ERROR] Failed to execute goal org.mortbay.jetty:h2spec-maven-plugin:1.0.9:h2spec (h2spec) on project jetty-http2-tests: 
[ERROR] Failed test cases:
[ERROR]         [8.1.2.6 - Sends a HEADERS frame with the "content-length" header field which does not equal the DATA frame payload length] failed
[ERROR]  Expected: GOAWAY Frame (Error Code: PROTOCOL_ERROR)
[ERROR]  Actual: RST_STREAM Frame (Error Code: PROTOCOL_ERROR)
[ERROR] 
```

lorban:
The latest version of `h2spec`'s 8.1.2.6 test accepts `RST_STREAM` as a valid response: https://github.com/summerwind/h2spec/blob/master/http2/8_1_2_6_malformed_requests_and_responses.go#L69-L97
but according to the above error message, the test fails because `GOAWAY` was expected but `RST_STREAM` was received.

It seems we're using an out of date h2spec? @olamy could you please check that we're using the latest h2spec binary?

lorban:
Running the latest h2spec v2.6.0 binary, 8.1.2.6 sometimes fails because it responds to the invalid request with a `DATA` frame instead of a `RST_STREAM` one, see the attached [8_1_2_6_failure.pcap.gz](https://github.com/eclipse/jetty.project/files/8999408/8_1_2_6_failure.pcap.gz) file for a capture of the failing run.

```
./h2spec -p 9999 http2/8.1.2.6
Hypertext Transfer Protocol Version 2 (HTTP/2)
  8. HTTP Message Exchanges
    8.1. HTTP Request/Response Exchange
      8.1.2. HTTP Header Fields
        8.1.2.6. Malformed Requests and Responses
          ✔ 1: Sends a HEADERS frame with the "content-length" header field which does not equal the DATA frame payload length
          using source address 127.0.0.1:44348"content-length" header field which does not equal the sum of the multiple DATA frames payload length
          × 2: Sends a HEADERS frame with the "content-length" header field which does not equal the sum of the multiple DATA frames payload length
            -> The endpoint MUST treat this as a stream error of type PROTOCOL_ERROR.
[...]

PR #8228, opened 2022-06-30
Issue #8212 - Exclude spec 8.1.2.6

joakime:
Exclude spec "8.1.2.6. Malformed Requests and Responses" from executing (for now).

### DIFF
--- a/jetty-core/jetty-http2/jetty-http2-tests/pom.xml
+++ b/jetty-core/jetty-http2/jetty-http2-tests/pom.xml
@@ -23,6 +23,7 @@
           <reportsDirectory>${project.build.directory}/h2spec-reports</reportsDirectory>
           <excludeSpecs>
             <excludeSpec>3.5 - Sends invalid connection preface</excludeSpec>
+            <excludeSpec>8.1.2.6. Malformed Requests and Responses</excludeSpec>
           </excludeSpecs>
         </configuration>
         <executions>
--- a/jetty-core/jetty-http2/jetty-http2-tests/pom.xml
+++ b/jetty-core/jetty-http2/jetty-http2-tests/pom.xml
@@ -15,6 +15,7 @@
       <plugin>
         <groupId>org.mortbay.jetty</groupId>
         <artifactId>h2spec-maven-plugin</artifactId>
+        <version>1.0.10-SNAPSHOT</version>
         <configuration>
           <mainClass>org.eclipse.jetty.http2.tests.H2SpecServer</mainClass>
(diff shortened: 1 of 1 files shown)


lorban:
Well, it seems your change doesn't work too well as the PR build failed on h2spec test 8.1.2.6. :-)

I'm not sure if it's something about the syntax or if there's a genuine bug in the h2spec plugin as I've seen builds failing because of the already excluded h2spec test 3.5 too.

joakime:
@lorban I had to add some debug output to h2spec to figure out why.
See: https://github.com/jetty-project/h2spec-maven-plugin/commit/5c9d174b2d38a484453e06b3f06ce3a435ccea20

How h2spec works is that ALL spec tests are run, and then the failed cases are evaluated against the `excludeSpecs` configuration of the h2spec-maven-plugin, if it's configured to exclude, that FAILED case is marked as ignored for purposes of the maven build failure.
It's a strange way to do that, I would have expected the `specIdentifier` specified in the `excludeSpecs` to not even run.

Also, the 8.1.2.6 `specIdentifier` is super odd according to h2spec.
It's full name is ... `8.1.2.3 - 8.1.2.6. Malformed Requests and Responses` ?

Which is what this PR is now using.

jetty-core/jetty-http2/jetty-http2-tests/pom.xml
```diff
@@ -23,6 +24,7 @@
           <reportsDirectory>${project.build.directory}/h2spec-reports</reportsDirectory>
           <excludeSpecs>
             <excludeSpec>3.5 - Sends invalid connection preface</excludeSpec>
+            <excludeSpec>8.1.2.3 - 8.1.2.6. Malformed Requests and Responses</excludeSpec>
```
gregw, line 27: Well done to find the magic needed!     
But can we have a TODO to investigate and remove the exclusions... and/or an issue.

gregw (changes requested):

jetty-core/jetty-http2/jetty-http2-tests/pom.xml
joakime: We can use issue #8212, right?

joakime: @gregw *bump*

gregw: Ok, don't close it with this

gregw (approved):

PR #8201, opened 2022-06-23
Jetty 12.0.x MultiPart

sbordet:
Moved multipart test files to jetty-http.

Introduced jetty-http `MultiPart` class and made `MultiPart.Parser` non-blocking and lenient on newlines.
Made `MultiPartCaptureTest` to pass.

Moved o.e.j.server.MultiPartParserTest tests to o.e.j.http.MultiPartTest.

Renamed MultiPartFormInputStreamTest to MultiPartsTest.

Introduced class o.e.j.server.MultiParts, a CompletableFuture of Parts
that can be used asynchronously in Handlers and in general in core Jetty.

Implemented multi-part handling in ee10.

Implemented DelayedHandler.UntilMultiPart.

Introduced MultiPart.ContentSource as the multipart "generator".

Signed-off-by: Simone Bordet <simone.bordet@gmail.com>

PR #8263, opened 2022-07-04
Jetty-12 Improved Content testing and javadoc

gregw:
Some javadoc improvements and some minor refinements


jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -462,6 +463,20 @@ static Chunk from(ByteBuffer byteBuffer, boolean last, Runnable releaser)
...
+         *
+         * @param byteBuffer the ByteBuffer with the bytes of this Chunk
+         * @param last whether the Chunk is the last one
+         * @param releaser the code to run when this Chunk is released
+         * @return a new Chunk
+         */
+        static Chunk from(ByteBuffer byteBuffer, boolean last, Consumer<ByteBuffer> releaser)
+        {
+            // TODO make this non capturing
+            return new ByteBufferChunk(byteBuffer, last, () -> releaser.accept(byteBuffer));
```
lorban: I think you should add a null check for `releaser`.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceInputStream.java
```diff
@@ -21,6 +21,13 @@
 import org.eclipse.jetty.util.Blocking;
 import org.eclipse.jetty.util.IO;
 
+/**
+ * An {@link InputStream} that is backed by a {@link Content.Source}.
+ * <p>
+ * The read methods are implemented by calling {@link Content.Source#read()}.
+ * Any {@link Content.Chunk}s read are released once all their content
+ * has been read.
```
lorban, line 28: I have a feeling you just put your finger on something important we missed: what should happen when the `InputStream` is closed?

IIRC some (all?) `Content.Source` implementations expect that `read()` be called continuously until a last chunk is returned, or that `fail()` is called, otherwise some resources might be leaked. `PathContentSource` for instance is impacted by this problem.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceTransformer.java
```diff
@@ -17,6 +17,17 @@
...
+/**
+ * An abstract transformer of {@link Content.Chunk}s from a {@link Content.Source} as a {@link Content.Source}.
+ * <p>
+ * This abstract {@link Content.Source} wraps another {@link Content.Source} and implementors need only to provide
+ * the {@link #transform(Content.Chunk)} method, which is used to transform {@link Content.Chunk} read from the
+ * wrapped source.
+ * <p>
+ * The {@link #demand(Runnable)} conversation is passed directly to the wrapped {@link Content.Source}, which means
+ * that transformations that may fully consume bytes read can result in a null return from {@link Content.Source#read()}
+ * even after a callback to the demand {@link Runnable}.
```
lorban: This last paragraph is (should?) already be covered by the fact that `demand()` may lead to spurious callbacks. I'd either remove this paragraph or link to the one that speaks about `demand()`'s spurious behavior.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceTransformer.java
```diff
@@ -112,5 +123,22 @@ private Content.Chunk process(Content.Chunk rawChunk)
...
+     * Content chunk transformation method.
+     * <p>
+     * This method is called during a {@link Content.Source#read()} to transform a raw chunk to a chunk that
+     * will be returned from the read call.  The caller of {@link Content.Source#read()} method is always
+     * responsible for calling {@link Content.Chunk#release()} on the returned chunk, which may be:
+     * <ul>
+     * <li>the <code>rawChunk</code>. This is typically done for {@link Content.Chunk.Error}s,
+     *     when {@link Content.Chunk#isLast()} is true, or if no transformation is required.</li>
+     * <li>a new (or predefined) {@link Content.Chunk} derived from the <code>rawChunk</code>. The transform is
+     *     responsible for calling {@link Content.Chunk#release()} on the <code>rawChink</code>, either during the call
```
lorban: `<code>rawChink</code>` ?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/InputStreamContentSource.java
```diff
@@ -16,24 +16,42 @@
...
+import org.eclipse.jetty.io.NullByteBufferPool;
 import org.eclipse.jetty.util.IO;
 import org.eclipse.jetty.util.thread.AutoLock;
 import org.eclipse.jetty.util.thread.SerializedInvoker;
 
+/**
+ * A {@link Content.Source} that is backed by an {@link InputStream}.
+ * <p>
+ * Data is read from the {@link InputStream} into a buffer that is optionally acquired
+ * from a {@link ByteBufferPool}, and converted to a {@link Content.Chunk} that is
```
lorban, line 30: You should specify where the buffer pool is coming from when none is given during construction.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/OutputStreamContentSource.java
```diff
@@ -21,6 +21,14 @@
 import org.eclipse.jetty.util.FutureCallback;
 import org.eclipse.jetty.util.IO;
 
+/**
+ * A {@link Content.Source} backed backed by an {@link OutputStream}.
```
lorban: `backed backed` ?

lorban (changes requested):
Reading the english-written contracts makes it more apparent that there are things we overlooked and others that need to be improved.

The [MultiPart PR](https://github.com/eclipse/jetty.project/pull/8201) contains some related changes that probably should be reviewed together with this PR.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: I've instead implemented the non capturing approach, which handles null

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceInputStream.java
gregw: We are correctly releasing the chunk on a close, but I think we should set the next chunk to an `Error` if we closed with a non null non error chunk.

gregw:
@lorban @sbordet Can I get this (re)reviewed please?  I've reviewed the multipart changes to the javadoc and they are orthogonal/complimentary.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceInputStream.java
lorban: You are correctly releasing the `Chunk` contained in `ContentSourceInputStream` but you're not notifying the wrapped `Content.Source` to release its chunks.

lorban (changes requested):
The `ContentSourceInputStream` does not notify the wrapped `Content.Source` when it is closed, potentially making it leak its internal resources. @sbordet it feels like something is unclear about the `Content.Source` contract that makes this problem non-obvious.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceInputStream.java
gregw: @lorban I'm not sure there is any compulsion to read all the content from a Source, so close not reading to completion is not a big deal.
However, I have pushed a version  that if close is done and EOF has not been read, then not only is the chunk set to an Error, but the source is failed.
@sbordet your thoughts?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceInputStream.java
```diff
@@ -89,10 +96,21 @@ public int available() throws IOException
...
     {
+        if (chunk == null)
+            chunk = content.read();
+
+        if (chunk != null && chunk.isLast())
+        {
+            if (!chunk.isTerminal())
+                chunk.release();
+            return;
+        }
```
sbordet: What? We certainly don't want to do this!

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceInputStream.java
```diff
@@ -89,10 +96,21 @@ public int available() throws IOException
...
+
         if (chunk != null)
             chunk.release();
-        if (chunk == Content.Chunk.EOF || chunk instanceof Content.Chunk.Error)
-            return;
-        chunk = Content.Chunk.EOF;
+
+        Throwable closed = new IOException("closed before EOF");
+        chunk = Content.Chunk.from(closed);
+        content.fail(closed);
```
sbordet:
```suggestion
        if (chunk == Content.Chunk.EOF || chunk instanceof Content.Chunk.Error)
            return;
        Throwable closed = new IOException("closed before EOF");
        chunk = Content.Chunk.from(closed);
        content.fail(closed);
```
I think a combination of old code and new code is better.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/InputStreamContentSource.java
```diff
@@ -59,17 +78,17 @@ public Content.Chunk read()
 
         try
         {
-            byte[] buffer = new byte[getBufferSize()];
-            int read = inputStream.read(buffer);
+            ByteBuffer buffer = bufferPool.acquire(getBufferSize(), false);
+            int read = inputStream.read(buffer.array());
```
sbordet: Not sure you can do this... Shouldn't `arrayOffset`, etc. be used?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/InputStreamContentSource.java
```diff
@@ -59,17 +78,17 @@ public Content.Chunk read()
...
             if (read < 0)
             {
                 close();
                 return Content.Chunk.EOF;
             }
             else
             {
-                ByteBuffer byteBuffer = ByteBuffer.wrap(buffer, 0, read);
-                return Content.Chunk.from(byteBuffer, false);
+                buffer.position(read);
```
sbordet: Buffer should be flipped and sliced.

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/AsyncContentSourceTest.java
```diff
@@ -53,7 +59,7 @@ public void testCloseInvokesDemandCallback() throws Exception
 
         CountDownLatch latch = new CountDownLatch(1);
         async.demand(latch::countDown);
-        assertFalse(latch.await(500, TimeUnit.MILLISECONDS));
+        assertFalse(latch.await(100, TimeUnit.MILLISECONDS));
```
sbordet: Don't think only 100 ms is a good idea.

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/AsyncContentSourceTest.java
```diff
@@ -67,27 +73,52 @@ public void testCloseInvokesDemandCallback() throws Exception
...
-        assertNotNull(chunk);
+            Content.Chunk chunk = async.read();
+            assertNotNull(chunk);
 
-        CountDownLatch latch = new CountDownLatch(1);
-        async.demand(latch::countDown);
-        assertFalse(latch.await(500, TimeUnit.MILLISECONDS));
+            CountDownLatch latch = new CountDownLatch(1);
+            async.demand(latch::countDown);
+            assertFalse(latch.await(100, TimeUnit.MILLISECONDS));
```
sbordet: Too short.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java
```diff
@@ -683,5 +683,21 @@ public Request.Processor handle(Request request) throws Exception
...
+            }
+        }
+
+        public abstract static class NonBlocking extends Processor
+        {
+            public NonBlocking()
+            {
+                super(InvocationType.NON_BLOCKING);
+            }
+        }
```
sbordet: Why is this part of this PR?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/thread/Invocable.java
```diff
@@ -13,6 +13,8 @@
 
 package org.eclipse.jetty.util.thread;
 
+import org.eclipse.jetty.util.Callback;
+
```
sbordet: Unused import?

sbordet (changes requested):
Have you confronted this PR with the multipart PR, that has more extensive javadocs?
Why the need for this PR when the multipart was already doing it?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java
gregw: Ah it wasn't meant to be... but kind of is OK.  I'm working on a blog/tutorial and thus fixing missing documentation and making better examples.  Having these inner classes greatly improves those examples and usability.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/InputStreamContentSource.java
gregw: technically yes, but currently no buffer allocated by our pools will have a non zero arrayOffset.  But I will add.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/InputStreamContentSource.java
gregw: why is slice needed?  The buffer is not retained nor reused by this class, so once the consumer calls release it is put back into the pool.

Perhaps it could be held expecting a release?  But this is simpler for now unless there is a performance indication that we need to do better.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/thread/Invocable.java
gregw: Is used in javadoc, but I have made that fully qualified instead of an import.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceInputStream.java
gregw: @sbordet the problem with your version is that it tests for `== EOF`, thus it will not see a normal chunk with isLast()==true as end of file and will terminate with a failure.

So my version is trying to say, that if a last chunk has been read, then a close is perfectly OK even if that chunk has unconsumed content... we just have to release it.   
I'll try to make that more explicit....

gregw: Ah I didn't see that isTerminal checks for no remaining bytes, so my code is kind of wrong?   Or maybe not, should closing a non terminal last chunk be a failure?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/InputStreamContentSource.java
gregw: Hmmm ContentSourceTest tests all the failure modes, but does test actually reading the content!  So this code is a) wrong; b) not tested!  fixing....

gregw:
The ContentSourceTests were not testing very much and had errors.  I have improved the tests and fixed some bugs, but there are still some that need to be checked.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferPool.java
```diff
@@ -27,6 +27,8 @@
  */
 public interface ByteBufferPool
 {
+    ByteBufferPool NULL_POOL = new NullByteBufferPool();
```
sbordet: I'd prefer this to be called `NOOP_POOL` and `NoopByteBufferPool`, as "null" does not really convey what's doing.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -223,6 +224,38 @@ static void consumeAll(Source source, Callback callback)
...
+            source.demand(getNext);
+            try
+            {
+                return next.get();
+            }
+            catch (Exception e)
+            {
+                throw new RuntimeException(e);
+            }
+        }
```
sbordet: I don't like this infinitely-waiting method that is only used by tests be so public in the main interface.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -459,7 +492,24 @@ static Chunk from(ByteBuffer byteBuffer, boolean last)
          */
         static Chunk from(ByteBuffer byteBuffer, boolean last, Runnable releaser)
         {
-            return new ByteBufferChunk(byteBuffer, last, releaser);
+            return releaser == null
+                ? new ByteBufferChunk(byteBuffer, last)
+                : new ByteBufferChunk.ReleasedByRunnable(byteBuffer, last, releaser);
```
sbordet: Now that the non-releaser version directly allocates, I would mandate that the 3rd parameter must be non-null.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -459,7 +492,24 @@ static Chunk from(ByteBuffer byteBuffer, boolean last)
...
+         * @param byteBuffer the ByteBuffer with the bytes of this Chunk
+         * @param last whether the Chunk is the last one
+         * @param releaser the code to run when this Chunk is released
+         * @return a new Chunk
+         */
+        static Chunk from(ByteBuffer byteBuffer, boolean last, Consumer<ByteBuffer> releaser)
+        {
+            return releaser == null
+                ? new ByteBufferChunk(byteBuffer, last)
+                : new ByteBufferChunk.ReleasedByConsumer(byteBuffer, last, releaser);
```
sbordet: Now that the non-releaser version directly allocates, I would mandate that the 3rd parameter must be non-null.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/NullByteBufferPool.java
```diff
@@ -21,6 +21,9 @@ public class NullByteBufferPool implements ByteBufferPool
 {
     private final RetainableByteBufferPool _retainableByteBufferPool = RetainableByteBufferPool.from(this);
 
+    public NullByteBufferPool()
+    {}
+
```
sbordet: We don't need this (it's generated automatically), unless it was meant to be `private` since there is now the constant in the interface.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
```diff
@@ -27,6 +27,13 @@
 import org.eclipse.jetty.util.thread.AutoLock;
 import org.eclipse.jetty.util.thread.SerializedInvoker;
 
+/**
+ * <p>A {@link Content.Source} that is also a {@link Content.Sink}.
+ * Content written to the {@link Content.Sink} is converted to a {@link Content.Chunk}
+ * and made available to calls to the {@link #read()} method.  In necessary, any
```
sbordet: `In` -> `If`.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
```diff
@@ -27,6 +27,13 @@
...
+ * {@link Runnable} passed to the {@link #demand(Runnable)} method is invoked once
+ * content is written to the {@link Content.Sink}.</p?
```
sbordet: `</p?` typo.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceInputStream.java
```diff
@@ -89,10 +96,26 @@ public int available() throws IOException
...
+
+            // if the chunk was a last chunk (but not an instanceof EOF), then nothing more to do
+            if (chunk.isLast())
+                return;
+        }
+
+        // This is an abnormal close before EOF
+        Throwable closed = new IOException("closed before EOF");
+        chunk = Content.Chunk.from(closed);
+        content.fail(closed);
```
sbordet, line 116: Still think we should not do any `read()` from `close()`.
Previous logic plus the last three lines above to fail before EOF.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceTransformer.java
```diff
@@ -112,5 +124,22 @@ private Content.Chunk process(Content.Chunk rawChunk)
         }
     }
 
+    /**
+     * Content chunk transformation method.
+     * <p>
```
sbordet, line 129: Improve `<p>` wrapping (one per paragraph).

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/InputStreamContentSource.java
```diff
@@ -59,17 +77,17 @@ public Content.Chunk read()
 
         try
         {
-            byte[] buffer = new byte[getBufferSize()];
-            int read = inputStream.read(buffer);
+            ByteBuffer buffer = bufferPool.acquire(getBufferSize(), false);
+            int read = inputStream.read(buffer.array(), buffer.arrayOffset(), buffer.capacity());
```
sbordet, line 81:
```suggestion
            int read = inputStream.read(buffer.array(), buffer.arrayOffset(), buffer.remaining());
```
This is more correct. We have `MappedByteBufferPool.Tagged` that slices the buffers of the pool, so `remaining()` is correct, not `capacity()`.

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/AsyncContentSourceTest.java
```diff
@@ -32,18 +37,19 @@
     @Test
     public void testOfferInvokesDemandCallback() throws Exception
     {
-        AsyncContent async = new AsyncContent();
+        try (AsyncContent async = new AsyncContent())
```
sbordet, line 44: No need of try-with-resources.

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/AsyncContentSourceTest.java
```diff
@@ -32,18 +37,19 @@
...
+        {
+            CountDownLatch latch = new CountDownLatch(1);
+            async.demand(latch::countDown);
+            assertFalse(latch.await(250, TimeUnit.MILLISECONDS));
```
sbordet, line 48: Beats me why you want to use 250 ms given all the slowness of CI. I'd restore at least 500.

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTest.java
```diff
@@ -65,15 +69,21 @@ public static List<Content.Source> all() throws Exception
             asyncSource.write(Content.Chunk.from(UTF_8.encode("two"), false), Callback.NOOP);
         }
 
-        Content.Source byteBufferSource = new ByteBufferContentSource(UTF_8.encode("one"), UTF_8.encode("two"));
+        ByteBufferContentSource byteBufferSource = new ByteBufferContentSource(UTF_8.encode("one"), UTF_8.encode("two"));
 
-        Content.Source.Transformer transformerSource = new Content.Source.Transformer(byteBufferSource)
+        Content.Source.Transformer transformerSource = new Content.Source.Transformer(new ByteBufferContentSource(UTF_8.encode("one"), UTF_8.encode("two")))
```
sbordet, line 72: Duplicated, revert to old code.

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTest.java
```diff
@@ -65,15 +69,21 @@ public static List<Content.Source> all() throws Exception
...
             @Override
             protected Content.Chunk transform(Content.Chunk rawChunk)
             {
                 return rawChunk;
             }
+
+            @Override
+            public String toString()
+            {
+                return "transformerSource@%x".formatted(hashCode());
```
sbordet: Please no, let's use `Class.simpleName` instead of hardcoded strings that don't refactor in case of renamings.

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTest.java
```diff
@@ -83,31 +93,99 @@ protected Content.Chunk transform(Content.Chunk rawChunk)
...
+                            e.printStackTrace();
+                        }
+                    }).start();
+                }
+            }
+
+            @Override
+            public String toString()
+            {
+                return "OutputStreamContentSource@%x".formatted(hashCode());
```
sbordet: Use `Class.simpleName`.

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTest.java
```diff
@@ -83,31 +93,99 @@ protected Content.Chunk transform(Content.Chunk rawChunk)
...
+            }
+        };
+
+        return List.of(asyncSource, byteBufferSource, transformerSource, pathSource, inputSource, inputSource2, outputSource);
     }
 
     @ParameterizedTest
     @MethodSource("all")
-    public void testDemandReadDemandDoesNotRecurse(Content.Source source)
+    public void testRead(Content.Source source) throws Exception
```
sbordet, line 136: Rename to `testReadToEOF()`.

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTest.java
```diff
@@ -121,21 +199,19 @@ public void run()
                 assertTrue(recursion.compareAndSet(true, false));
             }
         });
-        assertTrue(processed.get());
+        assertTrue(processed.await(10, TimeUnit.SECONDS));
```
sbordet, line 196: The previous use of `AtomicBoolean` made this test stricter on purpose. Using a `CountDownLatch` is not exactly the same (there was no need to change it).

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTest.java
```diff
@@ -146,16 +222,17 @@ public void run()
                     // Second demand after the first must throw.
                     assertThrows(IllegalStateException.class, () -> source.demand(this));
                 }
+                processed.countDown();
             }
         });
-        assertTrue(processed.get());
+        assertTrue(processed.await(10, TimeUnit.SECONDS));
```
sbordet, line 222: The previous use of `AtomicBoolean` made this test stricter on purpose. Using a `CountDownLatch` is not exactly the same (there was no need to change it).

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTest.java
```diff
@@ -146,16 +222,17 @@ public void run()
...
     }
 
     @ParameterizedTest
     @MethodSource("all")
-    public void testReadFailReadReturnsError(Content.Source source)
+    public void testReadFailReadReturnsError(Content.Source source) throws Exception
     {
-        Content.Chunk chunk = source.read();
+        Content.Chunk chunk = Content.Source.nextChunk(source);
```
sbordet: Why this? All sources should return a non null initial chunk all the times.

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTest.java
```diff
@@ -199,9 +271,13 @@ public void testReadErrorDemandInvokesDemandCallback(Content.Source source) thro
...
     @MethodSource("all")
-    public void testDemandCallbackThrows(Content.Source source)
+    public void testDemandCallbackThrows(Content.Source source) throws Exception
     {
-        Content.Chunk chunk = source.read();
+        // TODO fix for OSCS
+        if (source instanceof OutputStreamContentSource)
+            return;
+
+        Content.Chunk chunk = Content.Source.nextChunk(source);
```
sbordet: Why this? All sources should return a non null initial chunk all the times.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpConnection.java
```diff
@@ -1005,12 +1005,16 @@ public boolean content(ByteBuffer buffer)
...
-            });
+
+            stream._chunk = (LOG.isDebugEnabled())
+                ? Content.Chunk.from(buffer, false, () ->
+                {
+                    retainable.release();
+                    if (LOG.isDebugEnabled())
+                        LOG.debug("release {}/{} for {}", BufferUtil.toDetailString(buffer), retainable, this);
+                })
+                : Content.Chunk.from(buffer, false, retainable::release);
```
sbordet: I'd revert to old code, as the lambda is allocating anyway. `retainable::release` is nice to write, but it's no different from `() -> retainable.release()` which captures `retainable` which allocates.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/TutorialTest.java
```diff
@@ -0,0 +1,196 @@
...
+import org.eclipse.jetty.io.Content;
+import org.eclipse.jetty.server.Handler;
+import org.eclipse.jetty.server.Request;
+import org.eclipse.jetty.server.Response;
+import org.eclipse.jetty.util.Blocker;
+import org.eclipse.jetty.util.BufferUtil;
+import org.eclipse.jetty.util.Callback;
+import org.eclipse.jetty.util.thread.SerializedInvoker;
+
+public class TutorialTest
```
sbordet: No. This should be converted to a documentation class, see all examples in the documentation, e.g. `ServerDocs.java`.

sbordet (changes requested):

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTest.java
gregw: @sbordet No the OutputStreamContentSource doesn't

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTest.java
gregw: Without the latch, the test is assuming that the demand callback will be called within the scope of the demand call. That's not necessarily true in general, even if content is available.  It is definitely not true for OutputStreamContentSource, which uses a different thread to add the content after a demand call.

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTest.java
gregw: OutputStreamContentSource doesn't

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpConnection.java
gregw: good point

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/TutorialTest.java
gregw: Thanks, as the comment said... it was looking for a good home... but documentation is not currently being built and needs a bit of work to update it to 12 layout.

sbordet (approved):

PR #8264, opened 2022-07-04
Fix errors in  Mapped pool and javadoc

gregw:
I just noticed this was wrong with our buffer pool javadoc

### DIFF
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/MappedByteBufferPool.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/MappedByteBufferPool.java
@@ -133,6 +133,12 @@ public class MappedByteBufferPool extends AbstractByteBufferPool implements Dump
         _newBucket = newBucket;
     }
 
+    @Override
+    protected RetainableByteBufferPool newRetainableByteBufferPool(int factor, int maxCapacity, int maxBucketSize, long retainedHeapMemory, long retainedDirectMemory)
+    {
+        return new Retained(factor, maxCapacity, maxBucketSize, retainedHeapMemory, retainedDirectMemory);
+    }
+
     private Bucket newBucket(int key, boolean direct)
     {
         return (_newBucket != null) ? _newBucket.apply(key) : new Bucket(capacityFor(key), getMaxBucketSize(), updateMemory(direct));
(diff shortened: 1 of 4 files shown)


gregw:
Also some real fixes now!

jetty-io/src/main/java/org/eclipse/jetty/io/MappedByteBufferPool.java
```diff
@@ -302,4 +308,30 @@ public String toString()
             getMaxBucketSize(),
             getCapacityFactor());
     }
+
+    protected class Retained extends ArrayRetainableByteBufferPool
```
lachlan-roberts, line 312: Seems strange to me that you would create a `MappedByteBufferPool` and ask it for its `RetainableBufferPool` and it gives you an array based implementation.

gregw: yeah, but not sure we want to create a MappedRetainableByteBuffer pool.    Prior to the combination of pool, this is exactly what we would have got if the user selected a MBBP, as we would have created an ARBBP.   So just a bit strange that they are together.

I'm not actually sure if we really want to keep MBBP. @sbordet @lorban why do we have it?

lorban: AFAICT it's only used in the client, but I can't see why `MBBP` was created in the first place and I lack history. So unless @sbordet has a good reason to keep it, that pool impl can go IMHO.

gregw: OK, but for this PR, I'd like to fix javadoc and MBBP to unblock other work.
We can consider deprecating MBBP later.

lorban: Sounds good to me, and the change looks reasonable enough: currently MBBP is used together with a ARBBP so this change just solidifies a behavior that already exist.

gregw: @lorban can I get a green tick then please?

gregw:
@lachlan-roberts @lorban Can I get this approved?

lorban (approved):

PR #8272, opened 2022-07-07
Restore stage separated Jenkinsfile

joakime:
Restore the Stage separated Jenkinsfile from before.

This has the side effect that our Jenkins/CI builds **does not follow the same rules as a maven command line build** (or IDE build).
Building locally will **not** produce the same results as Jenkins/CI.
It is extremely likely that CI will fail, but local build will not.
If you encounter that, blame this stage separated `Jenkinsfile`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,37 +7,146 @@ pipeline {
   // save some io during the build
   options { durabilityHint('PERFORMANCE_OPTIMIZED') }
   stages {
-    stage("Parallel Stage") {
-      parallel {
-        stage("Build / Test - JDK17") {
-          agent { node { label 'linux' } }
+    stage("Checkout Jetty") {
+      steps {
+        container('jetty-build') {
+          dir("${env.WORKSPACE}/buildy") {
+            checkout scm
+          }
+        }
+      }
+    }
+    stage("Build & Test - JDK17") {
+      stages {
+        stage("Setup") {
           steps {
             container('jetty-build') {
-              timeout( time: 180, unit: 'MINUTES' ) {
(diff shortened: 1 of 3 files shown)


gregw:
@joakime nice to have the staged view back in CI.  But the main thing I was asking for was the restoration of fail at end semantics.  Currently CI is stopping at the first test failure, which may be a flake, and thus we don't get to see the rest of the build.    

So does this PR restore fail-at-end behaviour as well?

olamy:
@gregw fail at end has been restored in jetty-12.0.x branch with this commit https://github.com/eclipse/jetty.project/commit/f5e3ba459a77ac9ea8ffe2cf743b666be8dcec61 
which means all tests are executed even if a module has a test failure

olamy:
@joakime LGTM maybe we could add `-Dmaven.test.failure.ignore=true` as well to run all tests?

joakime:
> @joakime LGTM maybe we could add `-Dmaven.test.failure.ignore=true` as well to run all tests?

Sure, but lets put it in the `mavenBuild` function instead of in every stage.

Jenkinsfile
```diff
@@ -103,7 +103,7 @@ pipeline {
               timeout(time: 120, unit: 'MINUTES') {
                 dir("${env.WORKSPACE}/buildy") {
                   //cleanup all projects
-                  mavenBuild("jdk17", "clean install -f jetty-home", "maven3")
+                  mavenBuild("jdk17", "clean install -pl :jetty-home", "maven3")
```
joakime, line 106: Why this change?

PR #8275, opened 2022-07-07
Updating for published CVES

joakime:


gregw (changes requested):
why is this needed as a PR?  Aren't we just merging forwards from 10 to 11 to 12?

PR #8106, opened 2022-06-03
rework Resource API

lorban:
Overhaul of the Resource API and implementation.

PR #8276, opened 2022-07-08
Bis: Jetty 12.0.x resource handler and default servlet

lorban:
Work done on top of https://github.com/eclipse/jetty.project/pull/8106, rebasing this PR will be needed once that base PR gets merged.

Supersedes https://github.com/eclipse/jetty.project/pull/8090


lorban:
@janbartel @gregw This PR contains the preliminary default servlet / default handler / resource service code that should be able to fulfill basic requests to serve files from the filesystem. It's based on top of the Resource PR, which means it should be rebased on top of 12.0.x once the Resource PR gets merged.

joakime:
@gregw @janbartel this branch / PR is now green and ready for review.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -14,241 +14,167 @@
...
+    public static Resource getDefaultStylesheet()
+    {
+        // TODO the returned path should point to the classpath.
+        // This points to a non-existent file '/jetty-dir.css'.
+        return Resource.newResource(Path.of("/jetty-dir.css"));
+    }
 
-        boolean endsWithSlash = (pathInfo == null ? (_pathInfoOnly ? "" : servletPath) : pathInfo).endsWith(URIUtil.SLASH);
-        boolean checkPrecompressedVariants = _precompressedFormats.length > 0 && !endsWithSlash && !included && reqRanges == null;
+    public void doGet(GenericRequest request, GenericResponse response, Callback callback, HttpContent content) throws Exception
```
gregw, line 135: I don't t hink there is need of `GenericRequest` and `GenericResponse`.  Rather than always wrap both core and servlet objects to call this service, we should use core APIs and just wrap the servlet objects to look like core request/responses.

The reason I prefer this is that ultimately we will need those wrappers anyway to do such  things as cross context dispatch, so they may as well be used for calling this service and then not complicate this service when being used as core handler

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -869,4 +831,44 @@ protected void putHeaders(HttpServletResponse response, HttpContent content, lon
          */
         String getWelcomeFile(String pathInContext) throws IOException;
     }
+
+    public interface GenericRequest
```
gregw, line 835: Use core Request

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -869,4 +831,44 @@ protected void putHeaders(HttpServletResponse response, HttpContent content, lon
...
+        long getHeaderDate(String name);
+
+        HttpURI getHttpURI();
+
+        String getPathInContext();
+
+        String getContextPath();
+    }
+
+    public interface GenericResponse
```
gregw, line 850: use core Response

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ResourceHandler.java
```diff
@@ -82,126 +56,67 @@
...
+                return null;
+
+            for (String welcome : _welcomes)
             {
-                _defaultStylesheet = getDefaultStylesheet();
+                // TODO GW: This logic needs to be extensible so that a welcome file may be a servlet (yeah I know it shouldn't
+                //          be called a welcome file then.   So for example if /foo/index.jsp is the welcome file, we can't
+                //          serve it's contents - rather we have to let the servlet layer to either a redirect or a RequestDispatcher to it.
+                //          Worse yet, if there was a servlet mapped to /foo/index.html, then we need to be able to dispatch to it
+                //          EVEN IF the file does not exist.
```
gregw, line 111: this will need to be implemented sooner rather than later

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ResourceHandler.java
```diff
@@ -223,861 +138,154 @@ public Request.Processor handle(Request request) throws Exception
...
-            if (!response.isCommitted())
-                Response.writeError(request, response, callback, HttpStatus.BAD_REQUEST_400);
-            throw iae;
-        }
-
-        return false;
-    }
-
-    protected void sendWelcome(HttpContent content, String pathInContext, boolean endsWithSlash, Request request, Response response, Callback callback) throws Exception
+    public Resource getResourceBase()
```
gregw, line 154: No that we have restored `Resource` I think we can probably rename this method everywhere back to `getBaseResource()`... but I'll do that in another PR.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ResourceHandler.java
```diff
@@ -1122,269 +329,172 @@ public void setPathInfoOnly(boolean pathInfoOnly)
...
+        _resourceService.setStylesheet(stylesheet);
     }
 
     public void setWelcomeFiles(List<String> welcomeFiles)
     {
         _welcomes = welcomeFiles;
     }
 
-    private class PathContentFactory implements HttpContent.ContentFactory
+    private static class ResourceHandlerGenericRequest implements ResourceService.GenericRequest
```
gregw, line 349: Should not be necessary to wrap in core. See above.

jetty-ee10/jetty-ee10-demos/jetty-ee10-demo-embedded/src/main/resources/fileserver.xml
```diff
@@ -25,10 +25,14 @@
                   <Array type="String"><Item>index.html</Item></Array>
                 </Set>
                 <Set name="baseResource">
```
gregw, line 27: interesting, this is still `baseResource`, so we've not renamed it everywhere... or this XML has not been tested.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -13,8 +13,357 @@
...
+    public void init(ServletConfig config) throws ServletException
+    {
+        ContextHandler contextHandler = initContextHandler(config.getServletContext());
+
+        _resourceService = new ServletResourceService();
+        MimeTypes mimeTypes = new MimeTypes();
+        CompressedContentFormat[] precompressedFormats = new CompressedContentFormat[0];
+        _resourceService.setContentFactory(new CachingContentFactory(new ResourceContentFactory(contextHandler.getResourceBase(), mimeTypes, precompressedFormats)));
+
+        // TODO init other settings
```
gregw, line 64: Let's not forget this one!

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -13,8 +13,357 @@
...
+        }
+    }
+
+    @Override
+    protected void doHead(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException
+    {
+        doGet(req, resp);
+    }
+
+    private static class ServletGenericRequest implements ResourceService.GenericRequest
```
gregw, line 115: This class should be a wrapper of the `HttpServletRequest` to a core `Request` as a top level class, that will have other uses.

gregw (approved):
There are still some vital TODOs in this PR.    Also I do not like the basic approach of creating a generic Request/Response interface.    Our core Request/Response interfaces are the generic request/response for our server, so we should use those.   There will be other uses where we need to wrap servlet request/responses as core request/responses so those wrappers will be otherwise useful.    So we really don't need to create a third (or is it 4th? 5th?) class of request/response.

However, as this is a green build and a step in the right direction, I'll approve now and we can iterate to fix these problems.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -13,8 +13,357 @@
...
+        else
+            return context.getContextHandler();
+    }
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException
+    {
+        boolean useOutputDirectByteBuffers = true;
+        if (resp instanceof ServletContextResponse.ServletApiResponse servletApiResponse)
+            useOutputDirectByteBuffers = servletApiResponse.getResponse().getWrapped().getRequest().getConnectionMetaData().getHttpConfiguration().isUseOutputDirectByteBuffers();
```
janbartel, line 87: Wow, that's a long journey!

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ResourceHandler.java
joakime: Perhaps welcome file at ResourceHandler doesn't support the full welcome file behaviors?
But DefaultServlet absolutely has to.

janbartel (approved):
Super difficult to review with so many changes. I am surprised to see still so much commented out code - we're at the stage where jetty-12 is not experimental, and we need to ensure good code quality. However, in the interests of moving forward, I'm going to approve this.

Issue #8281, opened 2022-07-11
Jetty-12 : Review slf4j behaviors with the jetty-ee#-maven-plugins


joakime:
**Jetty version(s)**
12

**Java version/vendor** `(use: java -version)`
17

**Description**

Two things have been noticed when working with the tests against the `jetty-ee#-maven-plugin` implementations in Jetty 12.

First, is that there seems to be no way to show output.

You'll get ...

```
SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#noProviders for further details.
```

I have not found a configuration of a `<packaging>war</packaging>` project that uses the `jetty-ee#-maven-plugin` that will allow output to show.

And secondly, is the `ELContextCleaner` triggers exceptions ...

```
[INFO] jetty-12.0.0-SNAPSHOT; built: 2022-07-11T18:25:42.282Z; git: 67d03825080c75831e682a2a2b64665901d4f139; jvm 17.0.3+7
[WARNING] Failed startup of context o.e.j.e.m.p.MavenWebAppContext@31834a2b{ROOT,/,[file:///home/joakim/code/jetty/jetty.project-12.0.x/tests/jetty-ee9-test-maven-plugin/target/it/jetty-start-mojo-multi-module-single-war-it/webapp-war/src/main/webapp/],false}{file:///home/joakim/code/jetty/jetty.project-12.0.x/tests/jetty-ee9-test-maven-plugin/target/it/jetty-start-mojo-multi-module-single-war-it/webapp-war/src/main/webapp/}
java.lang.NoClassDefFoundError: org/slf4j/spi/SLF4JServiceProvider
    at java.lang.ClassLoader.defineClass1 (Native Method)
    at java.lang.ClassLoader.defineClass (ClassLoader.java:1012)
    at java.security.SecureClassLoader.defineClass (SecureClassLoader.java:150)
    at java.net.URLClassLoader.defineClass (URLClassLoader.java:524)
    at java.net.URLClassLoader$1.run (URLClassLoader.java:427)
    at java.net.URLClassLoader$1.run (URLClassLoader.java:421)
    at java.security.AccessController.doPrivileged (AccessController.java:712)
    at java.net.URLClassLoader.findClass (URLClassLoader.java:420)
    at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClassFromSelf (ClassRealm.java:425)
[...]

PR #8282, opened 2022-07-12
Issue #8281 Add slf4j impls to forked jetty plugin it tests

janbartel:
See #8281 

Adds jetty-slf4j-impl to jetty maven plugin it test executions so jetty log output can be seen.

### DIFF
--- a/tests/jetty-ee10-test-maven-plugin/src/it/jetty-cdi-start-forked/pom.xml
+++ b/tests/jetty-ee10-test-maven-plugin/src/it/jetty-cdi-start-forked/pom.xml
@@ -69,6 +69,13 @@
       <plugin>
         <groupId>org.eclipse.jetty.ee10</groupId>
         <artifactId>jetty-ee10-maven-plugin</artifactId>
+        <dependencies>
+          <dependency>
+              <groupId>org.eclipse.jetty</groupId>
+              <artifactId>jetty-slf4j-impl</artifactId>
+              <version>${jetty.version}</version>
+          </dependency>
+        </dependencies>
         <executions>
           <execution>
             <id>start-jetty</id>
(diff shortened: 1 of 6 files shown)


tests/jetty-ee10-test-maven-plugin/src/it/jetty-cdi-start-forked/pom.xml
```diff
@@ -69,6 +69,13 @@
       <plugin>
         <groupId>org.eclipse.jetty.ee10</groupId>
         <artifactId>jetty-ee10-maven-plugin</artifactId>
+        <dependencies>
+          <dependency>
+              <groupId>org.eclipse.jetty</groupId>
+              <artifactId>jetty-slf4j-impl</artifactId>
+              <version>${jetty.version}</version>
```
joakime, line 76: Don't we have this as `@jetty.version@`?
Should we have this in `it-parent-pom` like the other deps we use in these IT tests?

joakime:
Copy the IT test `jetty-start-forked-mojo-it` into a new test.

Lets call it `jetty-start-forked-mojo-failed-init-it`.
Add this to the `jetty-start-forked-mojo-failed-init-it/jetty-simple-webapp/src/main/webapp/WEB-INF/web.xml`

``` xml
  <!-- intentionally bad init -->
  <servlet>
    <servlet-name>foo</servlet-name>
    <servlet-class>org.foo.Bogus</servlet-class>
  </servlet>
```

and make the `jetty-start-forked-mojo-failed-init-it/postbuild.groovy` have another test ...

``` groovy
assert outputLog.text.contains( 'java.lang.ClassNotFoundException: org.foo.Bogus')
```

gregw (approved):

PR #8286, opened 2022-07-12
Issue #8216 - OpenID Connect RP-Initiated Logout

lachlan-roberts:
## Issue #8216

see https://openid.net/specs/openid-connect-rpinitiated-1_0.html

- Redirect back to openid provider when `HttpServletRequest.logout()` is called.
- Store the `end_session_endpoint` in the `OpenIdConfiguration`.

### DIFF
--- a/jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
+++ b/jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
@@ -83,7 +83,7 @@ public class OpenIdAuthenticator extends LoginAuthenticator
     private final SecureRandom _secureRandom = new SecureRandom();
     private OpenIdConfiguration _openIdConfiguration;
     private String _redirectPath;
-    private String _logoutRedirectPath = "/";
+    private String _logoutRedirectPath;
     private String _errorPage;
     private String _errorPath;
     private String _errorQuery;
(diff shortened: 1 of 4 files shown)


jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
```diff
@@ -166,6 +179,22 @@ else if (!redirectPath.startsWith("/"))
         _redirectPath = redirectPath;
     }
 
+    public void setLogoutRedirectPath(String logoutRedirectPath)
+    {
+        if (logoutRedirectPath == null)
+        {
+            LOG.warn("redirect path must not be null, defaulting to /");
+            logoutRedirectPath = "/";
+        }
```
gregw, line 188: I'd prefer to allow null to be set here and have that mean "do not redirect on logout"

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
```diff
@@ -218,6 +247,7 @@ public UserIdentity login(String username, Object credentials, ServletRequest re
     @Override
     public void logout(ServletRequest request)
     {
+        attemptLogoutRedirect(request);
```
gregw, line 250: This should be conditional on there being a redirection URL

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
```diff
@@ -218,6 +247,7 @@ public UserIdentity login(String username, Object credentials, ServletRequest re
     @Override
     public void logout(ServletRequest request)
```
gregw, line 248: Perhaps add some javadoc saying what this does during logout

gregw (changes requested):

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
lachlan-roberts: This redirect path is actually for the OpenId provider redirecting back to the webapp, not for the webapp to redirect to the OpenID Provider on logout. 

> post_logout_redirect_uri
> OPTIONAL. URI to which the RP is requesting that the End-User's User Agent be redirected after a logout has been performed. This URI SHOULD use the https scheme and MAY contain port, path, and query parameter components; however, it MAY use the http scheme, provided that the Client Type is confidential, as defined in Section 2.1 of [OAuth 2.0](https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RFC6749) [RFC6749], and provided the OP allows the use of http RP URIs. The URI MAY use an alternate scheme, such as one that is intended to identify a callback into a native application. The value MUST have been previously registered with the OP, either using the post_logout_redirect_uris Registration parameter or via another mechanism. An id_token_hint is also RECOMMENDED when this parameter is included.

So the "do not redirect on logout" setting is actually defined in `OpenIdConfiguration` as the `endSessionEndpoint`. But since this is optional we could not include this parameter if it is not explicitly set. Either way I think this does need some javadoc/documentation to explain this.

jetty-openid/src/test/java/org/eclipse/jetty/security/openid/OpenIdProvider.java
```diff
@@ -285,12 +294,44 @@
...
+            String logoutRedirect = req.getParameter("post_logout_redirect_uri");
+            if (logoutRedirect == null)
+            {
+                resp.sendError(HttpServletResponse.SC_BAD_REQUEST, "no post_logout_redirect_uri");
+                return;
+            }
+
+            loggedInUsers.decrement();
+            resp.setContentType("text/plain");
+            resp.sendRedirect(logoutRedirect);
```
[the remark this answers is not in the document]
lachlan-roberts: @olamy any idea how to suppress this warning? this is just a test class so not an issue.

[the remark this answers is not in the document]
lachlan-roberts: Never mind think I figured it out.

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
gregw: I understand it is the redirect after the logout redirect.   If there is no `endSessionEndpoint`, you just redirect to the logout redirect, which you are currently insisting is non null.
But you could allow null and define that to mean no redirecting at all on logout, thus giving the old behaviour.

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdConfiguration.java
```diff
@@ -97,14 +99,38 @@ public OpenIdConfiguration(@Name("issuer") String issuer,
                                @Name("clientSecret") String clientSecret,
                                @Name("authMethod") String authMethod,
                                @Name("httpClient") HttpClient httpClient)
+    {
+        this(issuer, authorizationEndpoint, tokenEndpoint, null, clientId, clientSecret, authorizationEndpoint, httpClient);
```
keller-j: I think `authorizationEndpoint` is passed by accident instead of `authMethod` (`authorizationEndpoint` is passed twice)

keller-j (changes requested):

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdConfiguration.java
lachlan-roberts: Good catch, thanks! Fixed.

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
lachlan-roberts: Ok I have changed to logic to not redirect back to application if `LOGOUT_REDIRECT_PATH` is null.

And updated the javadoc of `attemptLogoutRedirect` to explain how it works.

gregw (approved):

PR #8289, opened 2022-07-12
Re-enabled jetty-ee9-proxy and jetty-ee10-proxy modules.

sbordet:
Introduced TunnelSupport to abstract out the tunnelling capabilities.


sbordet:
@gregw here's the work on the proxy modules.

Note that I left 4 tests disabled as idle timeout in core still does not work as intended, but the majority of the work for the proxy modules is done.

Also note that I had to modify `ServletApiRequest.getRequestURL()` to comply with the spec (was returning the query string but it should not).
Also implemented `ServletApiRequest.getCookies()`, but can be optimized.

Moved `ConnectHandler` to core.

In ee10's `ServletChannel.handle()`, case `SEND_ERROR` I had to use `resetContent()` instead of `reset()` as we want to retain any header that has been set by the application before writing the error response. This is in line with Jetty 11.

gregw (changes requested):
currently not compiling in CI.
I'll have a look today if I get time.

jetty-ee10/jetty-ee10-proxy/src/test/java/org/eclipse/jetty/ee10/proxy/ForwardProxyTLSServerTest.java
```diff
@@ -889,18 +886,42 @@ private static class ServerHandler extends Handler.Processor
...
+                if (request.getHttpURI().getQuery() != null)
+                {
+                    Fields fields = Request.extractQueryParameters(request);
+                    String body = fields.getValue("body");
+                    if (body != null)
+                        Content.Sink.write(response, true, body, callback);
+                    else
+                        callback.succeeded();
+                }
+                else if (MimeTypes.Type.FORM_ENCODED.is(request.getHeaders().get(HttpHeader.CONTENT_TYPE)))
```
joakime, line 901: Not required, but could be useful (in the future).

``` java
Request.hasFormEncoded(request)
Request.hasQuery(request)
Request.hasMultipartForm(request)
```

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextRequest.java
```diff
@@ -449,8 +450,23 @@ public String getAuthType()
...
+            // TODO: optimize this.
+            return Request.getCookies(getRequest()).stream()
+                .map(this::convertCookie)
+                .toArray(Cookie[]::new);
+        }
+
+        public Cookie convertCookie(HttpCookie cookie)
+        {
+            Cookie result = new Cookie(cookie.getName(), cookie.getValue());
+            // TODO: inbound (client-to-server) cookies don't have all these parameters.
```
joakime, line 465: I think this is right for the Request side.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextRequest.java
```diff
@@ -570,7 +586,7 @@ public String getRequestURI()
         @Override
         public StringBuffer getRequestURL()
         {
-            return new StringBuffer(ServletContextRequest.this.getHttpURI().asString());
+            return new StringBuffer(HttpURI.build(ServletContextRequest.this.getHttpURI()).query(null).asString());
```
joakime, line 592: Makes sense, need to return the RequestURL without the query portion.

joakime (approved):
LGTM

olamy (approved):

PR #8298, opened 2022-07-13
Fix #8294 push added cookie

### LINKED ISSUE #8294: java.lang.ClassCastException: class org.eclipse.jetty.http.HttpField cannot be cast to class org.eclipse.jetty.http.HttpCookie$SetCookieHttpField
**Jetty version(s)**
10.0.11

**Java version/vendor**
17

**Description**
After we have call the follow code:
``` java
String rfc6265cookie = ...;
response.addHeader( "Set-Cookie", rfc6265cookie );
```

we get the follow exception:

```
java.lang.ClassCastException: class org.eclipse.jetty.http.HttpField cannot be cast to class org.eclipse.jetty.http.HttpCookie$SetCookieHttpField (org.eclipse.jetty.http.HttpField and org.eclipse.jetty.http.HttpCookie$SetCookieHttpField are in unnamed module of loader com.inet.plugin.DependencyClassLoader @4b2bac3f)
	at org.eclipse.jetty.server.Request.newPushBuilder(Request.java:355)
	at javax.servlet.http.HttpServletRequestWrapper.newPushBuilder(HttpServletRequestWrapper.java:349)
	at com.inet.remote.gui.angular.AngularContentService.createTemplate(AngularContentService.java:428)
	at com.inet.remote.gui.angular.AngularContentService.serveTemplate(AngularContentService.java:307)
	at com.inet.remote.gui.modules.reporterror.ReportErrorHandler.handleGet(ReportErrorHandler.java:74)
	at com.inet.remote.gui.modules.reporterror.ReportErrorHandler.handle(ReportErrorHandler.java:41)
	at com.inet.remote.gui.angular.AngularContentService.handleError(AngularContentService.java:874)
	at com.inet.remote.gui.angular.RemoteGuiServletErrorHandlerProvider.sendErrorPage(RemoteGuiServletErrorHandlerProvider.java:49)
	at com.inet.http.error.ServletErrorHandler.sendErrorPage(ServletErrorHandler.java:55)
	at com.inet.http.PluginDispatcherServlet.service(PluginDispatcherServlet.java:157)
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:764)
	at org.eclipse.jetty.servlet.ServletHandler$ChainEnd.doFilter(ServletHandler.java:1665)
	at org.eclipse.jetty.websocket.servlet.WebSocketUpgradeFilter.doFilter(WebSocketUpgradeFilter.java:170)
	at org.eclipse.jetty.servlet.FilterHolder.doFilter(FilterHolder.java:202)
```

gregw:
Reparse cookie added with addCookie via the `java.net.HttpCookie`  for #8294

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
@@ -352,7 +352,10 @@ public class Request implements HttpServletRequest
             HttpHeader header = field.getHeader();
             if (header == HttpHeader.SET_COOKIE)
             {
-                HttpCookie cookie = ((SetCookieHttpField)field).getHttpCookie();
+                HttpCookie cookie = (field instanceof SetCookieHttpField)
+                    ? ((SetCookieHttpField)field).getHttpCookie()
+                    : new HttpCookie(field.getValue());
+
                 if (cookie.getMaxAge() > 0)
                     cookies.put(cookie.getName(), cookie.getValue());
                 else
--- a/jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
+++ b/jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
@@ -1967,7 +1967,7 @@ public class RequestTest
         request.getResponse().getHttpFields().add(new HttpCookie.SetCookieHttpField(new HttpCookie("bonza", "bewdy", 1), CookieCompliance.RFC6265));
         request.getResponse().getHttpFields().add(new HttpCookie.SetCookieHttpField(new HttpCookie("bad", "thumbsdown", 0), CookieCompliance.RFC6265));
         request.getResponse().getHttpFields().add(new HttpField(HttpHeader.SET_COOKIE, new HttpCookie("ugly", "duckling", 100).getSetCookie(CookieCompliance.RFC6265)));
-        request.getResponse().getHttpFields().add(new HttpField(HttpHeader.SET_COOKIE, new HttpCookie("flown", "away", 0).getSetCookie(CookieCompliance.RFC6265)));
+        request.getResponse().getHttpFields().add(new HttpField(HttpHeader.SET_COOKIE, "flow=away; Max-Age=0; Secure; HttpOnly; SameSite=None"));
         HttpFields.Mutable fields = HttpFields.build();
         fields.add(HttpHeader.AUTHORIZATION, "Basic foo");
         request.setMetaData(new MetaData.Request("GET", HttpURI.from(uri), HttpVersion.HTTP_1_0, fields));
--- a/jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
+++ b/jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
@@ -54,6 +54,7 @@ import org.eclipse.jetty.http.BadMessageException;
 import org.eclipse.jetty.http.CookieCompliance;
 import org.eclipse.jetty.http.HttpCompliance;
 import org.eclipse.jetty.http.HttpCookie;
+import org.eclipse.jetty.http.HttpField;
 import org.eclipse.jetty.http.HttpFields;
 import org.eclipse.jetty.http.HttpHeader;
 import org.eclipse.jetty.http.HttpTester;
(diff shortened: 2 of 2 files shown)


jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
```diff
@@ -1965,6 +1966,7 @@ public void testPushBuilder()
         request.getResponse().getHttpFields().add(new HttpCookie.SetCookieHttpField(new HttpCookie("good", "thumbsup", 100), CookieCompliance.RFC6265));
         request.getResponse().getHttpFields().add(new HttpCookie.SetCookieHttpField(new HttpCookie("bonza", "bewdy", 1), CookieCompliance.RFC6265));
         request.getResponse().getHttpFields().add(new HttpCookie.SetCookieHttpField(new HttpCookie("bad", "thumbsdown", 0), CookieCompliance.RFC6265));
+        request.getResponse().getHttpFields().add(new HttpField(HttpHeader.SET_COOKIE, new HttpCookie("ugly", "thumbsdown", 100).getSetCookie(CookieCompliance.RFC6265)));
```
joakime: We should get the example cookie that the OP is setting and use it here.
There might be an aspect of RFC6265 we've missed.

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -352,7 +352,10 @@ public PushBuilder newPushBuilder()
             HttpHeader header = field.getHeader();
             if (header == HttpHeader.SET_COOKIE)
             {
-                HttpCookie cookie = ((SetCookieHttpField)field).getHttpCookie();
+                HttpCookie cookie = (field instanceof SetCookieHttpField)
+                    ? ((SetCookieHttpField)field).getHttpCookie()
+                    : new HttpCookie(field.getValue());
```
joakime, line 357: This reparse should follow the `HttpConfiguration` setting for `response._responseCookieCompliance` too.

joakime (changes requested):
Reparse should follow the HttpConfiguration configuration for response cookie compliance.

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
gregw: Hmm maybe.  But that is about the format of the set-cookies that jetty will generate.  This is a set-cookie that the application has generated, probably to get around some compliance issue.
Also, we don't actually have a compliance specific Set-Cookie parser, as we currently only use that part for generation.  So that would be a bit of work.    

So I think this is a step in the right direction, if not perfect.

joakime: The normal cookie parsing step is done via the `CookieCutter` class, which has compliance modes, not the direct `org.eclipse.jetty.http.HttpCookie(String)` constructor, which uses the `java.net.HttpCookie` class to parse, which has limitations (namely it only supports RFC2109 and RFC2965).  

For example, if the OPs `Set-Cookie` has a Same-Site attribute the `org.eclipse.jetty.http.HttpCookie(String)` usage will not parse it.

jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
joakime: This usage, of having a `HttpField` with an `HttpCookie` object as a value isn't parsing like you think.
Add a Same-Site to that cookie, and see if the resulting cookie from the Push has that Same-Site attribute.

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
gregw: CookieCutter only parses Cookie values, not Set-Cookie values... and then it is mostly about ancient long deprecated RFCs.

I've extended the test to make sure we don't throw anything if the same-site attribute is set.  We don't.  It is ignored, but then all we need is max-age to be parsed correctly, which it is.
So I still think we are good.   We don't need to parse same-site, as by definition the cookie on a push request is going to the same site.

joakime: ok, works for me.

joakime (approved):

gregw:
@joakime  can you merge for me and then merge through 11 & 12.
Note where it ends up in 12... Ee9 or 10? Hopefully 9 and then copy to 10

joakime:
Merged to `jetty-12.0.x` at ee9 level.
But the ee10 `ServletContextRequest.ServletApiRequest` does not have `newPushBuilder()` implemented yet.

Issue #8296, opened 2022-07-13
SymlinkAllowedResourceAliasChecker is initialized after checkAlias is called resulting that access to resource is denied


jlindman:
**Jetty version(s)**

11.0.8-11.0.11

**Java version/vendor** `(use: java -version)`

openjdk version "17.0.3" 2022-04-19

**OS type/version**

Ubuntu 20.04.4 LTS 

**Description**

I have a JAX-WS application that reads the configuration file /WEB-INF/sun-jaxws.xml.

When the WEB-INF directory is inside a directory that is a symlink the sun-jaxws.xml file cannot be retrieved since is not allowed because the WEB-INF directory is protected.

This results in the following error:

`SEVERE: WSSERVLET11: failed to parse runtime descriptor: jakarta.xml.ws.WebServiceException: Runtime descriptor "/WEB-INF/sun-jaxws.xml" is missing`

Adding below works with Jetty 11.0.7 (where `webapp` is a symlink):

```
ContextHandler linkHandler = new ContextHandler();
linkHandler.setResourceBase("webapp");
webapp.addAliasCheck(new SymlinkAllowedResourceAliasChecker(linkHandler));
```

However, with Jetty 11.0.8 this stopped working. It looks like the initialization of the AllowedResourceAliasChecker is now done too late(?). The resource base seems to be null when `aliasCheck` is made.

With 11.0.8 the resource base is null for the SymlinkAllowedResourceAliasChecker objects (which was not the case for 11.0.7).

In ContentHandler.checkAlias():1967

```
getAliasChecks() ->
 result = {Collections$UnmodifiableRandomAccessList@7090}  size = 2
  0 = {SymlinkAllowedResourceAliasChecker@3192} "SymlinkAllowedResourceAliasChecker@54f6b629{base=null,protected=[/META-INF, /WEB-INF]}"
  1 = {SymlinkAllowedResourceAliasChecker@3827} "SymlinkAllowedResourceAliasChecker@4bc9ca97{base=null,protected=[]}"
```

Stacktrace:
```
checkAlias:1967, ContextHandler (org.eclipse.jetty.server.handler)
getResource:1941, ContextHandler (org.eclipse.jetty.server.handler)
getResource:411, WebAppContext (org.eclipse.jetty.webapp)
getResource:1445, WebAppContext$Context (org.eclipse.jetty.webapp)
[...]

Issue #8259, opened 2022-07-01
Symlinks cause 404 with DefaultServlet when its "resourceBase" is different from ContextHandler's


andrus:
**Jetty version(s)**

10.0.11 ... This worked on 9.x

**Java version/vendor** `(use: java -version)`

openjdk version "11.0.14.1" 2022-02-08

**OS type/version**

Mac OS 12.4

**Description**

The problem: `DefaultServlet` can't serve files located in symlinked directories if its resource base is different from its parent `ContextHandler`.

I am running Jetty with [Bootique Framework](https://github.com/bootique/bootique-jetty). `ContextHandler` has no "resourceBase" set. There is a `DefaultServlet` that has "resourceBase" set to say `/opt/p/c`. If `/opt/p/c` are all regular folders, I can serve static files from this path via the `DefaultServlet`. If `c` is a symlink to a folder outside of `p`, I am getting a 404. Debugging shows that the culprit is alias checking by `SymlinkAllowedResourceAliasChecker`. 

Workaround 1: setting `ContextHandler.resourceBase` to `/opt/p/c` makes things work again. This isn't always applicable, as I may have multiple DefaultServlets pointing to different folders on the filesystem.
Workaround 2: Using deprecated `AllowSymLinkAliasChecker` :

```java
contextHandler.setAliasChecks(List.of(new AllowSymLinkAliasChecker()));
```

Desired behavior: in `SymlinkAllowedResourceAliasChecker` make alias checking to work relative to the DefaultServlet base.

joakime:
If you have a ContextHandler (or WebAppContext) resourceBase set, then the actual DefaultServlet (on url-pattern `/` with name `default`) should be using that exclusively.

If you want another DefaultServlet, not on the usual servlet default requirement, then you can specify extra DefaultServlet's to then have a resourceBase declared for each.

There's examples of this at https://github.com/jetty-project/embedded-jetty-cookbook/blob/jetty-10.0.x/src/main/java/org/eclipse/jetty/cookbook/DefaultServletMultipleBases.java

andrus:
@joakime 

> If you want another DefaultServlet, not on the usual servlet default requirement, then you can specify extra DefaultServlet's to then have a resourceBase declared for each.

Yep, this is what I do (my DefaultServlet is mapped to "/asset/*", ContextHandler resource base is not set). But due to symlinks in the servlet's resource base, it stopped working after upgrading from Jetty 9 to 10. That's the problem that I described above.

[...]

PR #8315, opened 2022-07-18
Issue #8296 and #8259 -  AllowedResourceAliasChecker improvements

lachlan-roberts:
## Issue #8296 and #8259

- change `AllowedResourceAliasChecker` and `SymlinkAllowedResourceAliasChecker` to allow custom resource base directory, so we can more easily support use case of multiple default servlets with different resource bases from #8259
- allow `AllowedResourceAliasChecker` to be used when `ContextHandler` is still starting, by always extracting resource base when needed before `lifeCycleStarted` is called.
- resolve symlinks in the `baseResource` during `ContextHandler.doStart()` so that not every path from it is also an alias.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
@@ -861,9 +861,7 @@ public class ContextHandler extends ScopedHandler implements Attributes, Gracefu
         if (getBaseResource() != null && getBaseResource().isAlias())
         {
             // We may have symlink to baseResource, try to resolve symlink if possible.
-            File file = getBaseResource().getFile();
-            if (file != null)
-                _baseResource = Resource.newResource(file.toPath().toRealPath());
+            _baseResource = Resource.resolveAlias(_baseResource);
 
             LOG.warn("BaseResource {} is aliased to {} in {}. May not be supported in future releases.",
                 getBaseResource(), getBaseResource().getAlias(), this);
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
@@ -859,8 +859,15 @@ public class ContextHandler extends ScopedHandler implements Attributes, Gracefu
             throw new IllegalStateException("Null contextPath");
 
         if (getBaseResource() != null && getBaseResource().isAlias())
+        {
+            // We may have symlink to baseResource, try to resolve symlink if possible.
+            File file = getBaseResource().getFile();
+            if (file != null)
+                _baseResource = Resource.newResource(file.toPath().toRealPath());
+
             LOG.warn("BaseResource {} is aliased to {} in {}. May not be supported in future releases.",
                 getBaseResource(), getBaseResource().getAlias(), this);
+        }
 
         _availability.set(Availability.STARTING);
 
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
@@ -77,6 +77,26 @@ public abstract class Resource implements ResourceFactory, Closeable
         return __defaultUseCaches;
     }
 
+    public static Resource resolveAlias(Resource resource)
+    {
+        if (!resource.isAlias())
+            return resource;
+
+        try
+        {
+            File file = resource.getFile();
+            if (file != null)
+                return Resource.newResource(file.toPath().toRealPath());
+        }
+        catch (IOException e)
+        {
+            if (LOG.isDebugEnabled())
+                LOG.debug("resolve alias failed", e);
+        }
+
+        return resource;
+    }
+
     /**
      * Construct a resource from a uri.
      *
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
@@ -77,6 +77,11 @@ public abstract class Resource implements ResourceFactory, Closeable
         return __defaultUseCaches;
     }
 
+    /**
+     * Attempt to resolve the real path of a Resource to potentially remove any symlinks causing the Resource to be an alias.
+     * @param resource the resource to resolve.
+     * @return a new Resource resolved to the real path of the original Resource, or the original resource if it was not an alias.
+     */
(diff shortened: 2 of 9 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -859,8 +859,15 @@ protected void doStart() throws Exception
             throw new IllegalStateException("Null contextPath");
 
         if (getBaseResource() != null && getBaseResource().isAlias())
+        {
+            // We may have symlink to baseResource, try to resolve symlink if possible.
+            File file = getBaseResource().getFile();
+            if (file != null)
+                _baseResource = Resource.newResource(file.toPath().toRealPath());
+
```
gregw: I'm pretty sure the WebAppProvider in the deployer does something similar.   Perhaps wrap up in a utlility static method on Resource

tests/test-integration/src/test/java/org/eclipse/jetty/test/AliasCheckerMultipleResourceBasesTest.java
```diff
@@ -0,0 +1,161 @@
...
+        Files.delete(_altDir2Symlink);
+
+        _client.stop();
+        _server.stop();
+    }
+
+    @Test
+    public void test() throws Exception
+    {
+        System.err.println(_webRootPath.toAbsolutePath());
```
gregw: remove println

tests/test-integration/src/test/java/org/eclipse/jetty/test/AliasCheckerMultipleResourceBasesTest.java
```diff
@@ -0,0 +1,161 @@
...
+    {
+        Files.delete(_altDir1Symlink);
+        Files.delete(_altDir2Symlink);
+
+        _client.stop();
+        _server.stop();
+    }
+
+    @Test
+    public void test() throws Exception
```
gregw, line 120: this test will only work on FSs with symlink support, so you probably need some assumes

tests/test-integration/src/test/java/org/eclipse/jetty/test/AliasCheckerMultipleResourceBasesTest.java
```diff
@@ -0,0 +1,161 @@
...
+        {
+            for (ContextHandler.AliasCheck aliasCheck : aliasChecks)
+            {
+                _context.addAliasCheck(aliasCheck);
+            }
+        }
+    }
+
+    @BeforeAll
+    public static void beforeAll() throws Exception
```
gregw: I bit strange having a beforeAll and then only a single test.  Do you expect more tests?

tests/test-integration/src/test/java/org/eclipse/jetty/test/AliasCheckerWebRootIsSymlinkTest.java
```diff
@@ -0,0 +1,159 @@
...
+    {
+        // Try to delete all files now so that the symlinks do not confuse other tests.
+        Files.delete(_webrootSymlink);
+
+        _client.stop();
+        _server.stop();
+    }
+
+    @Test
+    public void test() throws Exception
```
gregw, line 115: why isn't this just another test case method in the previous test class? why two test classes, each with 1 test?

tests/test-integration/src/test/java/org/eclipse/jetty/test/AliasCheckerWebRootIsSymlinkTest.java
```diff
@@ -0,0 +1,159 @@
...
+    public static void afterAll() throws Exception
+    {
+        // Try to delete all files now so that the symlinks do not confuse other tests.
+        Files.delete(_webrootSymlink);
+
+        _client.stop();
+        _server.stop();
+    }
+
+    @Test
```
gregw, line 114: again, this needs to assume LINUX or system with symlinks

gregw (changes requested):
a few niggles

tests/test-integration/src/test/java/org/eclipse/jetty/test/AliasCheckerMultipleResourceBasesTest.java
lachlan-roberts: Most of our tests with symlinks don't have any assumes at all.

Windows has supported symlinks for a long time now so don't think its needed anyway.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
lachlan-roberts: I added a static method on `Resource` called `resolveAlias`.

tests/test-integration/src/test/java/org/eclipse/jetty/test/AliasCheckerMultipleResourceBasesTest.java
lachlan-roberts: done

tests/test-integration/src/test/java/org/eclipse/jetty/test/AliasCheckerMultipleResourceBasesTest.java
lachlan-roberts: I have changed these to use before each

tests/test-integration/src/test/java/org/eclipse/jetty/test/AliasCheckerWebRootIsSymlinkTest.java
lachlan-roberts: The setup is different enough so that it is difficult to merge into one test.

jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -77,6 +77,26 @@ public static boolean getDefaultUseCaches()
         return __defaultUseCaches;
     }
 
+    public static Resource resolveAlias(Resource resource)
```
gregw, line 85: javadoc

jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
gregw: yeah but it needs javadoc!

gregw (dismissed):
Approved... but add some javadoc to the new method before merging.

joakime (approved):

PR #8323, opened 2022-07-21
some playing around git

olamy:


olamy:
yup finally find a solution with a custom slave-jnlp image https://github.com/webtide/jenkins-slave/blob/master/slave-image/Dockerfile

PR #8324, opened 2022-07-21
skip test for CodeQL

olamy:


.github/workflows/codeql-analysis.yml
```diff
@@ -67,11 +67,8 @@ jobs:
           # Details on CodeQL's query packs refer to : https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs
           # queries: security-extended,security-and-quality
 
-
-      # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
-      # If this step fails, then you should remove it and run the build manually (see below)
-      - name: Autobuild
-        uses: github/codeql-action/autobuild@v2
+      - name: Clean install dependencies and build
+        run: mvn install -DskipTests
```
joakime: mvn **clean** install -DskipTests ?

joakime: Do we want to break the javascript scan?

Lets do this, as we only care about the java build.
```
- if: matrix.language == 'java' 
  name: Clean install dependencies and build
  run: |
    mvn clean install -DskipTests
```

From https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-the-codeql-workflow-for-compiled-languages#adding-build-steps-for-a-compiled-language

We could leave autobuild around for the javascript matrix.

joakime: What about the `-Pci` profile?  Does that include anything useful we should be scanning?

olamy: `ci` profile is only for static analysis.
I removed javascript https://github.com/eclipse/jetty.project/commit/39acb3c13f60c78d9f1ad8d952801a817ca8cf98
not sure we really need `clean` as it's a fresh checkout

joakime (approved):

PR #8325, opened 2022-07-21
Jetty 12.0.x retainable chunk

sbordet:
Made Chunk implement Retainable, and protocol implementations to link the RetainableByteBuffer all the way to the Chunk.

Extended Retainable to have both a retain() and a release() methods.

Removed HTTP/2's ISession and IStream, now just using HTTP2Session and HTTP2Stream.
Removed *.Adapter classes in favor of interfaces with default methods.

Changed the low-level HTTP/2 read code to be a demand() + onDataAvailable() model.


sbordet:
@gregw @joakime the main reason for this PR was to make `Chunk` a `Retainable`, because we needed a way to link the network buffer (a `Retainable`) that we use in HTTP/2 and HTTP/3 to the `Chunk` that we return from `Request.read()`, so that if the `Chunk` is retained (for example by the multipart implementation), then the retention propagates all the way back to the network buffer, without the need to copy data.

With the occasion, also `Data` classes in HTTP/2 and HTTP/3 have been made `Retainable` so that low-level HTTP/2 and HTTP/3 code can retain `Data` objects for later use.
Also with the occasion, the HTTP/3 way of reading `Data` objects has been ported to HTTP/2 as well.

Notable classes to review:
* `ServerFCGIConnection` -- where the network buffer is linked to the chunk
* `HTTP2Connection:408` -- where the network buffer is linked to the `Data` object
* `HTTP2Session:264` -- where the `Data` object is wrapped with an independent reference counter (please read javadocs of the `StreamData` class to know why)
* `HttpStreamOverHTTP2:146` -- where the `Data` object is converted to a `Chunk` and linked to it
*  `HTTP3StreamConnection:212` -- where the network buffer is linked to the Data object; differently from HTTP/2, we don't need to deal with the flow control window because it's handled by Quiche
* `HttpStreamOverHTTP3:146` -- where the `Data` object is converted to a `Chunk` and linked to it
* The classes in `jetty-io` where there are the changes to `Chunk`; `InputStreamContentSource` and `PathContentSource` that now use a `RetainableByteBufferPool`; `ContentSourceTransformer` semantic has been clarified
* `GzipRequest` -- decoding has been rewritten in light of the `ContentSourceTransformer` semantic changes
* `HttpConnection:1010` -- where the network buffer is linked to the `Chunk`

An unrelated change is the change in `ServletChannel` where the output is re-opened; this is necessary to be able to write the content of an error dispatch (Jetty 11 has similar code).

jetty-core/jetty-alpn/jetty-alpn-conscrypt-client/src/test/java/org/eclipse/jetty/alpn/java/client/ConscryptHTTP2ClientTest.java
```diff
@@ -69,29 +67,32 @@ public void testConscryptHTTP2Client() throws Exception
...
 
                 @Override
-                public void onData(Stream stream, DataFrame frame, Callback callback)
+                public void onDataAvailable(Stream stream)
                 {
-                    System.err.println(frame);
-                    callback.succeeded();
-                    if (frame.isEndStream())
+                    Stream.Data data = stream.readData();
+                    System.err.println(data);
```
joakime, line 92: Do you need System.err.println here?

jetty-core/jetty-alpn/jetty-alpn-java-client/src/test/java/org/eclipse/jetty/alpn/java/client/JDK9HTTP2ClientTest.java
```diff
@@ -55,30 +53,33 @@ public void testJDK9HTTP2Client() throws Exception
...
 
                 @Override
-                public void onData(Stream stream, DataFrame frame, Callback callback)
+                public void onDataAvailable(Stream stream)
                 {
-                    System.err.println(frame);
-                    callback.succeeded();
-                    if (frame.isEndStream())
+                    Stream.Data data = stream.readData();
+                    System.err.println(data);
```
joakime, line 79: Do you need System.err.println here?

jetty-core/jetty-http2/jetty-http2-tests/src/test/java/org/eclipse/jetty/http2/tests/DataDemandTest.java
```diff
@@ -99,61 +107,52 @@ public void onDataDemanded(Stream stream, DataFrame frame, Callback callback)
...
-        int loops = 0;
+        AtomicInteger count = new AtomicInteger(serverQueue.size());
         while (true)
         {
-            if (++loops > 100)
-                fail();
-
-            Thread.sleep(100);
-
+            serverStream.demand();
```
joakime, line 118: Missing `serverStream.release()` ?

jetty-core/jetty-http2/jetty-http2-tests/src/test/java/org/eclipse/jetty/http2/tests/DataDemandTest.java
```diff
@@ -99,61 +107,52 @@ public void onDataDemanded(Stream stream, DataFrame frame, Callback callback)
...
-        loops = 0;
+        count.set(clientQueue.size());
         while (true)
         {
-            if (++loops > 100)
-                fail();
-
-            Thread.sleep(100);
-
+            clientStream.demand();
```
joakime, line 154: Missing `clientStream.release()`?

joakime (approved):
Minor nits on System.err and seemingly missing release in test cases.
Not enough to not merge.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -41,6 +41,7 @@
 import org.eclipse.jetty.util.FuturePromise;
 import org.eclipse.jetty.util.IO;
 import org.eclipse.jetty.util.Promise;
+import org.eclipse.jetty.util.Retainable;
```
gregw: The javadoc for `Content.Souce#read` needs to be updated to acknowledge Retainable.  It needs to say that it returns a `Chunk` that has an implicit call to `retain` (or something similar).

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Retainable.java
```diff
@@ -13,7 +13,139 @@
...
+ * It may be retained more times (thus incrementing its reference count) and released
+ * (thus decrementing its reference count), until the reference count goes to zero.</p>
+ */
 public interface Retainable
 {
-    public void retain();
+    /**
+     * <p>Retains this resource, incrementing the reference count.</p>
+     */
+    public default void retain()
```
gregw: don't need public

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Retainable.java
```diff
@@ -13,7 +13,139 @@
...
+    }
+
+    /**
+     * <p>Releases this resource, decrementing the reference count.</p>
+     * <p>This method returns {@code true} when the reference count goes to zero,
+     * {@code false} otherwise.</p>
+     *
+     * @return whether the invocation of this method decremented the reference count to zero
+     */
+    public default boolean release()
```
gregw: don't need public

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Retainable.java
```diff
@@ -13,7 +13,139 @@
...
+    {
+        return true;
+    }
+
+    public static class Wrapper implements Retainable
```
gregw: don't need public static

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Retainable.java
```diff
@@ -13,7 +13,139 @@
...
+     * <p>A reference count implementation for a {@link Retainable} resource.</p>
+     * <p>The reference count is initialized to 1 when the resource is created,
+     * and therefore it is implicitly retained and needs a call to {@link #release()}.</p>
+     * <p>Additional calls to {@link #retain()} must be matched by correspondent
+     * calls to {@link #release()}.</p>
+     * <p>When the reference count goes to zero, the resource may be pooled.
+     * When the resource is acquired from the pool, {@link #acquire()} should be
+     * called to set the reference count to {@code 1}.</p>
+     */
+    public static class ReferenceCounter implements Retainable
```
gregw: don't need public static

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Retainable.java
```diff
@@ -13,7 +13,139 @@
 
 package org.eclipse.jetty.util;
```
gregw: should this be in jetty-io?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -477,6 +442,50 @@ static Chunk from(ByteBuffer byteBuffer, boolean last, Consumer<ByteBuffer> rele
...
+         *
+         * @param byteBuffer the ByteBuffer with the bytes of this Chunk
+         * @param last whether the Chunk is the last one
+         * @param retainable the Retainable this Chunk links to
+         * @return a new Chunk
+         */
+        static Chunk from(ByteBuffer byteBuffer, boolean last, Retainable retainable)
+        {
+            return new ByteBufferChunk.ReleasedByRetainable(byteBuffer, last, Objects.requireNonNull(retainable));
+        }
```
gregw: The javadoc on all the from methods needs to be more explicit that there is an implicit retain already done on the returned Chunk, so that a single release will do the releasing, unless retain has been called.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -477,6 +442,50 @@ static Chunk from(ByteBuffer byteBuffer, boolean last, Consumer<ByteBuffer> rele
             return new ByteBufferChunk.ReleasedByConsumer(byteBuffer, last, Objects.requireNonNull(releaser));
         }
 
+        /**
+         * <p>Creates a last/non-last Chunk with the given ByteBuffer, linked to the given Retainable.</p>
```
gregw, line 451: The text "last/non-last" is hard to read and gives too much detail in the text. The parameter description is enough.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -477,6 +442,50 @@ static Chunk from(ByteBuffer byteBuffer, boolean last, Consumer<ByteBuffer> rele
...
+         * <p>A typical example are Chunks decoded (e.g. gzip) via {@link ContentSourceTransformer}:
+         * the decoded Chunk is not linked to the encoded Chunk, but needs its own reference counter
+         * as the decoded Chunk may in turn be {@link #slice(Chunk, int, int, boolean) sliced}.</p>
+         *
+         * @param byteBuffer the ByteBuffer with the bytes of this Chunk
+         * @param last whether the Chunk is the last one
+         * @param releaser the code to run when this Chunk is released
+         * @return a new Chunk
+         */
+        static Chunk fromWithReferenceCount(ByteBuffer byteBuffer, boolean last, Consumer<ByteBuffer> releaser)
```
gregw: It is not obvious in the javadoc of the `static Chunk from(ByteBuffer byteBuffer, boolean last, Consumer<ByteBuffer> releaser)` method that it doesn't have a reference count.

In fact, I don't think we should ever have a chunk that does not have a real reference count (either directly or in a delegate field).

So my preference is to get rid of the other from method (which does not appear to be used) and rename this method to just `from`  (although I think `as` would be a better name for all the `from` methods here).

IF we are to have some chunks that do not have a real reference count in them, then retain must either return a boolean false OR throw an ISE.   I don't like either!

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Retainable.java
```diff
@@ -13,7 +13,139 @@
...
+ */
 public interface Retainable
 {
-    public void retain();
+    /**
+     * <p>Retains this resource, incrementing the reference count.</p>
+     */
+    public default void retain()
+    {
+    }
```
gregw: I find an empty method here a little bit dangerous.  You have several implementions that implement release to do something, but there is no actual reference count, so calls to retain are ignored.

If you want a default, then it should be:
```suggestion
    {
        throw new UnsupportedOperationException();
    }
```

But it is probably better to force implementations to implement this method, so that only the ones that have a noop release also have a noop retain.

At the very least, you need javadoc that says implementations that implement release, must implement retain and respect the contract.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/internal/ByteBufferChunk.java
```diff
@@ -102,11 +107,40 @@ public ReleasedByConsumer(ByteBuffer byteBuffer, boolean last, Consumer<ByteBuff
             this.releaser = new AtomicReference<>(releaser);
         }
 
-        public void release()
+        @Override
+        public boolean release()
+        {
+            boolean released = super.release();
```
gregw, line 147: This is a wrong implementation. A call to `retain` is a noop and thus a subsequent call to `release` will always release regardless of how many times retain has been called.  This violates the `Retainable` contract.

Do we need this?

gregw (changes requested):
A few niggles... but  also a show stopper.  

We should not have chunks that violate the `Retainable` contract.  A call to `retain` can only be a noop if a call to `release` is a noop.  Any `Retainable` with a `release` that does something must have a real reference counter somewhere counting calls to retain.

OR at the very least a call to retain needs to throw if it is ignored and release is not a noop.

jetty-core/jetty-alpn/jetty-alpn-conscrypt-client/src/test/java/org/eclipse/jetty/alpn/java/client/ConscryptHTTP2ClientTest.java
sbordet: Yes, it's useful when this class is run against an external server.

jetty-core/jetty-alpn/jetty-alpn-java-client/src/test/java/org/eclipse/jetty/alpn/java/client/JDK9HTTP2ClientTest.java
sbordet: Yes, it's useful when this class is run against an external server.

jetty-core/jetty-http2/jetty-http2-tests/src/test/java/org/eclipse/jetty/http2/tests/DataDemandTest.java
sbordet: implemented.

jetty-core/jetty-http2/jetty-http2-tests/src/test/java/org/eclipse/jetty/http2/tests/DataDemandTest.java
sbordet: Implemented.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: The javadoc mandates that the returned Chunk must be released. 
I think it's enough and does not overspecify anything about `retain()` -- it could be possible to have an implementation of `Source` that does not retain the chunks it returns.
The caller of `read()` only need to know that it has to call `release()`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Retainable.java
sbordet: Was originally (in Jetty 10) in `util`, but I agree it should be in `io`.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: Like for `Source.read()` I think it's better to specify that a `release()` must be done, rather than going into details about `retain()`. For example, for this particular method, there is no `retain()` as the `Chunk` just wraps the given `Retainable`.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: Agreed on the reference counting argument.
I modified the `from()` methods to return either a reference counted `Chunk`, or one that wraps another `Retainable`.
Now the only use of non-reference counted `ByteBufferChunk` is for the static constants `EMPTY` and `EOF`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Retainable.java
sbordet: I made `Retainable` methods abstract now, so every implementation must implement them.
However, `release()` cannot throw if `retain()` is empty, because special implementations such as `Chunk.Error` or `Chunk.EOF` must have a release method that does not throw.
However, should it return `true` or `false`? For now I decided for `false` (contrary to the previous implementation that returned `true`.
Typically callers don't care about the return value of `release()`, only wrappers do, so I guess either `false` or `true` is pretty much the same, but I now have a slight preference for `false`.

sbordet: Ok I changed my mind. I now prefer to throw `UnsupportedOperationException` from `retain()` and return `true` from `release()` for special implementations.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/internal/ByteBufferChunk.java
sbordet: This should be solved by having `retain()` throw and `release()` returning `true`.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: I think saying release must be called is a good start. But i think it should say more.  it needs to say that if retain is called then release needs to be called twice.  

Even if the chunk returned is not able to be further retained, it is still a Retainable and needs to be released. So it's really just a retain limited retainable.

So i think explain the contract without referenced to retainable, but then have extra info, perhaps in an @see clause

sbordet:
@gregw I think I addressed your concerns, please re-review.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: Done.

gregw: Now I think you've gone too far.  I'll resolve here and make a suggestion against the latest javadoc.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -271,6 +270,9 @@ default long getLength()
          * continue to return a last chunk (although the instance may be different).</p>
          * <p>Chunks of content that have been consumed by the content reader code must
          * be {@link Chunk#release() released}.</p>
+         * <p>The content reader code may perform additional calls to {@link Chunk#retain()}
+         * on the returned {@link Chunk}, that must be matched by a correspondent number
+         * of calls to {@link Chunk#release()}.</p>
```
gregw: How about:
```suggestion
         * <p>The content reader code must ultimately arrange for a call to {@link Chunk#release()} on the 
         returned {@link Chunk}.  Additionally, prior to the ultimate call to {@link Chunk#release()}, the reader
         may make additonal calls to {@link Chunk#retain()}, that must ultimately be matched by a 
         correspondent number of calls to {@link Chunk#release()} once all references to the returned chunk 
         are discarded.</p>
```

gregw (changes requested):

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -440,23 +407,24 @@ static void write(Sink sink, boolean last, String utf8Content, Callback callback
...
+         * <p>The returned Chunk must be {@link #release() released}.</p>
          *
          * @param byteBuffer the ByteBuffer with the bytes of this Chunk
          * @param last whether the Chunk is the last one
          * @return a new Chunk
          */
         static Chunk from(ByteBuffer byteBuffer, boolean last)
         {
-            return new ByteBufferChunk(byteBuffer, last);
+            return new ByteBufferChunk.WithReferenceCount(byteBuffer, last);
```
gregw, line 420: I don't think this one does need a reference count as there is no release action.  It can have a noop retain and a release that always returns false.

gregw (approved):

PR #8326, opened 2022-07-21
Re-enabled the documentation module.

sbordet:
Now -DskipTests will compile the documentation classes but not build the documentation, saving time when building.

Much more work to do to update the documentation for Jetty 12.


documentation/jetty-documentation/pom.xml
```diff
@@ -70,6 +70,7 @@
               <goal>process-asciidoc</goal>
             </goals>
             <configuration>
+              <skip>${skipTests}</skip>
```
olamy: to avoid repeating this, you should try add a single entry in the `<properties>` section such:
```
<properties>
  <asciidoctor.skip>${skipTests}</asciidoctor.skip>
</properties>
```

olamy (changes requested):

olamy (approved):

PR #8328, opened 2022-07-22
Jetty 12 - Cleanup MultiReleaseJarFile to use newFileSystem behaviors built into JVM

joakime:
Use new Resource.mount and FileSystem behaviors from JVM to simplify the MultiReleaseJarFile


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -143,13 +169,82 @@ public static Resource.Mount mount(URI uri) throws IOException
...
+            idxEnd = path.length();
+        // look for filename suffix
+        int idxSuffix = path.lastIndexOf('.', idxEnd);
+        if (idxSuffix < 0)
+            return false; // no suffix found, can't possibly be a jar/war
+        String suffix = path.substring(idxSuffix, idxEnd).toLowerCase(Locale.ENGLISH);
+        return suffix.equals(".jar") || suffix.equals(".war");
+    }
+
+    public static URI toJarRoot(URI uri)
```
gregw: There is something wierd with the return of this method, which I  think it is something like "jar:file:/some/path/foo.jar", which is not valid as it is missing the "!/".

It should return either "file:/some/path/foo.jar" OR "jar:file:/some/path/foo.jar!/".
I don't mind either, but have a slight preference for the later as it matches the name better.

Eitherway, javadoc is needed.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -108,18 +109,41 @@ public static boolean getDefaultUseCaches()
...
+            return null;
+        String scheme = uri.getScheme();
+        if (scheme == null)
+            return null;
+        if (!isJar(uri))
             return null;
         try
         {
-            return (uri.getScheme().equalsIgnoreCase("jar")) ? FileSystemPool.INSTANCE.mount(uri) : null;
+            if (scheme.equalsIgnoreCase("jar") || scheme.equalsIgnoreCase("file"))
```
gregw: why are you mounting file URIs?  Is it because somethings it has not been converted to a jar:file URI yet? In which case I'd prefer we always did that.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
```diff
@@ -186,6 +187,11 @@ public void deconfigure(WebAppContext context) throws Exception
         {
             _mountedResources.forEach(IO::close);
         }
+        Resource baseResource = context.getResourceBase();
+        if (baseResource instanceof Closeable closeableResource)
+        {
+            IO.close(closeableResource);
+        }
```
gregw: maybe we should have an `IO.close(Object)`?

gregw (dismissed):
This PR does a lot more than it says on the tin!
Can you update title and description.

I have a few questions...

joakime:
> This PR does a lot more than it says on the tin! Can you update title and description.
> 
> I have a few questions...

Absolutely!
This is where I left off on Friday.

I want to split out the ResourceCollection changes at the very least to a different PR on Monday.
There's probably a few other things that could be smaller PRs as well.

joakime:
Created PR #8337 with the finished Immutable `ResourceCollection` (with auto-mount) changes from this PR.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/FileSystemPool.java
```diff
@@ -43,6 +43,15 @@
     private static final Logger LOG = LoggerFactory.getLogger(FileSystemPool.class);
     public static final FileSystemPool INSTANCE = new FileSystemPool();
 
+    private static final Map<String, String> ENV_MULTIRELEASE_RUNTIME;
+
+    static
+    {
+        Map<String, String> env = new HashMap<>();
+        env.put("releaseVersion", "runtime");
```
lorban, line 58: Don't forget to add here a comment with the link to the doc covering this env parameter.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/FileSystemPool.java
```diff
@@ -62,12 +71,16 @@ public Resource.Mount mount(URI uri) throws IOException
...
             try
             {
-                fileSystem = FileSystems.newFileSystem(uri, Collections.emptyMap());
+                fileSystem = FileSystems.newFileSystem(uri, ENV_MULTIRELEASE_RUNTIME);
             }
             catch (FileSystemAlreadyExistsException fsaee)
             {
                 fileSystem = Paths.get(uri).getFileSystem();
             }
+            catch (ProviderNotFoundException pnfe)
```
lorban, line 91: Why do you want to move on if such exception is thrown? If there's no FS provider, we should not try to retain that filesystem and return a broken mount.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
lorban: If you want to mount the jar pointed to by a `file:` URI, I'd prefer to have a `mountJarIfNeeded(URI jarFileUri)` helper method that explicitly does that.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -130,6 +154,8 @@ public static Resource.Mount mountIfNeeded(URI uri)
      */
     public static Resource.Mount mount(URI uri) throws IOException
     {
+        if (!isJar(uri))
```
lorban: This is redundant with the following `if`, one has to go.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -143,13 +169,82 @@ public static Resource.Mount mount(URI uri) throws IOException
...
+        if (!isJar(path))
+            throw new IllegalArgumentException("Path is not a JAR: " + path);
         URI pathUri = path.toUri();
         if (!pathUri.getScheme().equalsIgnoreCase("file"))
             throw new IllegalArgumentException("not an allowed path: " + path);
         URI jarUri = URI.create(toJarRoot(pathUri.toString()));
         return FileSystemPool.INSTANCE.mount(jarUri);
     }
 
+    public static boolean isJar(Path path)
```
lorban: I'd prefer to keep such helper `private`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -143,13 +169,82 @@ public static Resource.Mount mount(URI uri) throws IOException
...
+    {
+        if (path == null)
+            return false;
+        return path.getFileName().toString().toLowerCase(Locale.ENGLISH).endsWith(".jar");
+    }
+
+    public static boolean isJar(URI uri)
```
lorban: I'd prefer to keep such helper `private`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
lorban: I'd prefer to keep such helper `private`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -30,17 +29,23 @@
...
- * Allows webapps to have multiple (static) sources.
+ * A collection of Resources.
+ * Allows webapps to have multiple sources.
  * The first resource in the collection is the main resource.
  * If a resource is not found in the main resource, it looks it up in
- * the order the resources were constructed.
+ * the order the provided in the constructors
  */
-public class ResourceCollection extends Resource
+public class ResourceCollection extends Resource implements AutoCloseable
```
lorban: No `Resource` should ever be `Closeable`.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/ResourceTest.java
```diff
@@ -393,4 +396,57 @@ public void testDotAlias() throws Exception
         assertNotNull(same);
         assertTrue(same.isAlias());
     }
+
+    @ParameterizedTest
+    @ValueSource(strings = {
+        "file:/home/user/.m2/repository/com/company/1.0/company-1.0.jar",
+        "jar:file:/home/user/.m2/repository/com/company/1.0/company-1.0.jar!/",
```
lorban: This URI looks wrong to me as it lacks a trailing `!/`.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/ResourceTest.java
```diff
@@ -393,4 +396,57 @@ public void testDotAlias() throws Exception
...
+        "jar:file:/home/user/.m2/repository/jakarta/servlet/jakarta.servlet-api/6.0.0/jakarta.servlet-api-6.0.0.jar!/META-INF/resources"
+    })
+    public void testIsJarUriTrue(String rawUri)
+    {
+        assertTrue(Resource.isJar(URI.create(rawUri)), "Should be detected as a JAR URI: " + rawUri);
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {
+        "jar:file:/home/user/project/with.jar/in/path/name",
```
lorban: This URI looks wrong to me as it lacks a trailing `!/`.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/ResourceTest.java
```diff
@@ -393,4 +396,57 @@ public void testDotAlias() throws Exception
...
+        assertFalse(Resource.isJar(URI.create(rawUri)), "Should be detected as a JAR URI: " + rawUri);
+    }
+
+    public static Stream<Arguments> toJarRootCases()
+    {
+        List<Arguments> cases = new ArrayList<>();
+
+        String expected = "jar:file:/path/company-1.0.jar!/";
+        cases.add(Arguments.of("file:/path/company-1.0.jar", expected));
+        cases.add(Arguments.of("jar:file:/path/company-1.0.jar", expected));
```
lorban: This URI looks wrong to me as it lacks a trailing `!/`.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/ResourceTest.java
```diff
@@ -393,4 +396,57 @@ public void testDotAlias() throws Exception
...
+        cases.add(Arguments.of("jar:file:/path/company-1.0.jar!/", expected));
+        cases.add(Arguments.of("jar:file:/path/company-1.0.jar!/META-INF/services", expected));
+
+        expected = "jar:file:/opt/jetty/webapps/app.war!/";
+        cases.add(Arguments.of("file:/opt/jetty/webapps/app.war", expected));
+        cases.add(Arguments.of("jar:file:/opt/jetty/webapps/app.war", expected));
```
lorban: This URI looks wrong to me as it lacks a trailing `!/`.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
lorban: No `Resource` should ever be `Closeable`. This need to close feels like a lack of mounting.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
```diff
@@ -711,6 +713,12 @@ public FileVisitResult visitFile(Path f, BasicFileAttributes attrs) throws IOExc
         return tlds;
     }
 
+    public static boolean isTldFile(Path path)
```
lorban: We're starting to get a lot of filters like this one scattered all around. We should consider grouping them into a util class.

lorban (changes requested):
The real annoyance is the `Closeable` `ResourceCollection`. But we've discussed the matter and we agreed a new static `Resource` method doing collection-mounting would solve this.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/FileSystemPool.java
joakime: Not an IAE

joakime:
@lorban look at the `URIUtil.fixBadJavaIoFileUrl` method. (awkward method name, got a better one?)

That's the heart of the issue I was having with the FileSystemPool retain/release.
Two different URIs.
The bad ones with `file:/foo.jar` from `URLClassLoader.getURLs` or `File.toURL` or `URL.toURI`.
And the to spec ones `file:///foo.jar` from `Path.toURI`, `FileSystem`, etc
We would wind up with a uri-to-spec version during `Configuration`, and when the same JARs went through the `ClassLoader` they got mangled to the out-of-spec ones.
This `URIUtil.fixBadJavaIoFileUrl` cleans them up as we go to be to spec compliant.

lorban:
@joakime I'm worried about the brittleness of `fixBadJavaIoFileUrl`. I've been scratching my head and figured we can take advantage of the existing logic in `FileSystemPool.mount` with minimal changes. Consider the following:

```java
URI uri1 = URI.create("jar:file:" + absoluteJarPath + "!/");
URI uri2 = URI.create("jar:file://" + absoluteJarPath + "!/");

FileSystem fileSystem1 = FileSystems.newFileSystem(uri1, Collections.emptyMap());
FileSystem fileSystem2;
try
{
    fileSystem2 = FileSystems.newFileSystem(uri2, Collections.emptyMap());
    fail("expected FileSystemAlreadyExistsException");
}
catch (FileSystemAlreadyExistsException e)
{
    fileSystem2 = Paths.get(uri2).getFileSystem();
}
assertSame(fileSystem1, fileSystem2);
assertEquals(fsToUri(fileSystem1), fsToUri(fileSystem2));
```

The above shows that the `FileSystems.newFileSystem` code already handles this problem. If we can find a reasonable way transform a `FileSystem` into a URI, we can take advantage of the fact that the JVM already takes care of that `file:` vs `file://` URI difference.

So I came up with:

```java
private static URI fsToUri(FileSystem fileSystem)
{
    Iterable<Path> rootDirectories = fileSystem.getRootDirectories();
    return StreamSupport.stream(rootDirectories.spliterator(), false)
        .sorted()
        .findFirst()
        .orElseThrow()
        .toUri();
}
```

which would then allow us to rewrite `FileSystemPool.mount` in this way (I've omitted the changes to `Mount` itself):

```diff
-            Mount mount = new Mount(uri);
-            retain(uri, fileSystem, mount);
+            URI fsUri = fsToUri(fileSystem);
+            Mount mount = new Mount(fsUri, Resource.newResource(uri));
+            retain(fsUri, fileSystem, mount);
```

WDYT?

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/URIUtilTest.java
```diff
@@ -699,39 +747,24 @@ public void testFileUriGetUriLastPathSegment(String basePath, String expectedNam
...
     }
 
-    public static Stream<Arguments> uriLastSegmentSource() throws URISyntaxException, IOException
+    public static Stream<Arguments> uriLastSegmentSource() throws IOException
     {
         final String TEST_RESOURCE_JAR = "test-base-resource.jar";
+        List<Arguments> arguments = new ArrayList<>();
         Path testJar = MavenTestingUtils.getTestResourcePathFile(TEST_RESOURCE_JAR);
-        URI uri = new URI("jar", testJar.toUri().toASCIIString(), null);
+        URI jarFileUri = Resource.toJarFileUri(testJar.toUri());
```
lorban, line 755: `Resource.toJarFileUri` looks useless now.

jetty-ee10/jetty-ee10-quickstart/src/main/java/org/eclipse/jetty/ee10/quickstart/PreconfigureQuickStartWar.java
```diff
@@ -95,7 +95,7 @@ public static void preconfigure(Resource war, Resource dir, Resource xml) throws
             if (!dir.exists())
                 Files.createDirectories(dir.getPath());
 
-            URI jarUri = URI.create("jar:" + war.getURI() + "!/");
+            URI jarUri = Resource.toJarFileUri(war.getURI());
```
lorban, line 98: `toJarFileUri` seems like it should not be part of `Resource` anymore. Or maybe should go away entirely.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -1644,30 +1647,100 @@ public static String addQueries(String query1, String query2)
...
-        if (!uri.startsWith("jar:"))
-            return uri;
-        int bangSlash = uri.indexOf("!/");
-        return (bangSlash >= 0) ? uri.substring(4, bangSlash) : uri.substring(4);
-    }
+        URL[] urls = urlClassLoader.getURLs();
+        return Stream.of(urls)
+            .filter(Objects::nonNull)
+            .map(URIUtil::getJarSource)
+            .map(URIUtil::fixBadJavaIoFileUrl);
```
lorban, line 1744: I'm not convinced calling `fixBadJavaIoFileUrl` here has any value.

lorban (approved):
A few nits, otherwise LGTM.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/URIUtilTest.java
joakime: Lets try to address this in the FileID/URIUtil second PR (not the first)

jetty-ee10/jetty-ee10-quickstart/src/main/java/org/eclipse/jetty/ee10/quickstart/PreconfigureQuickStartWar.java
joakime: Ditto, Lets try to address this in the FileID/URIUtil second PR (not the first)

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: This fixes non-Jetty URLClassLoaders to produce to-spec URIs.
Perhaps it should be `ensureToSpecFileUri()`? 
But we can address the naming in the FileID/URIUtil second PR (not the first)

PR #8334, opened 2022-07-25
codeql correctly setup maven to not depend on default installation

olamy:
to not depend on what is installed in the image

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -69,7 +69,7 @@ jobs:
           
       - name: Set up Maven
         run:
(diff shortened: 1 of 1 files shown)


joakime:
Seems odd to force a mvnw version like this.

It's just a workflow action.
Can't we just use the normal `setup-java` stuff?

ala ...

``` yml
      - name: Set up JDK
        uses: actions/setup-java@v2.3.1
        with:
          distribution: 'temurin'
          java-version: ${{ matrix.java }}
          cache: 'maven'
```

olamy:
yup something such (from https://github.com/stCarolas/setup-maven)
```
    - name: Set up Maven
      uses: stCarolas/setup-maven@v4
      with:
        maven-version: 3.8.6
```

olamy:
@joakime change done

joakime (approved):
LGTM, this action approach is probably better cached (within github) too.
No need to go out to the internet now.

Issue #8285, opened 2022-07-12
Jetty-12 can't serve resources from jar


gregw:
**Jetty version(s)**
Jetty-12

**Description**

After #8106 was merged, the ee9-demo-async-rest webapp is no longer able to server content from it's jar:
```
2022-07-12 12:34:17.703:WARN :oejen.HttpChannel:qtp1955915048-15: /ee9-demo-async-rest/asyncrest/red.png
java.lang.UnsupportedOperationException
        at jdk.zipfs/jdk.nio.zipfs.ZipFileSystem$1.map(ZipFileSystem.java:1102)
        at org.eclipse.jetty.util.BufferUtil.toMappedBuffer(BufferUtil.java:1072)
        at org.eclipse.jetty.util.BufferUtil.toMappedBuffer(BufferUtil.java:1065)
        at org.eclipse.jetty.ee9.nested.CachedContentFactory.getMappedBuffer(CachedContentFactory.java:344)
        at org.eclipse.jetty.ee9.nested.CachedContentFactory$CachedHttpContent.getBuffer(CachedContentFactory.java:550)
        at org.eclipse.jetty.ee9.nested.HttpOutput.sendContent(HttpOutput.java:1307)
        at org.eclipse.jetty.ee9.nested.HttpOutput.sendContent(HttpOutput.java:1186)
        at org.eclipse.jetty.ee9.nested.ResourceService.sendData(ResourceService.java:744)
        at org.eclipse.jetty.ee9.nested.ResourceService.doGet(ResourceService.java:300)
        at org.eclipse.jetty.ee9.servlet.DefaultServlet.doGet(DefaultServlet.java:459)
        at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:500)
        at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:587)
        at org.eclipse.jetty.ee9.servlet.ServletHolder$NotAsync.service(ServletHolder.java:1410)
        at org.eclipse.jetty.ee9.servlet.ServletHolder.handle(ServletHolder.java:764)
        at org.eclipse.jetty.ee9.servlet.ServletHandler.doHandle(ServletHandler.java:530)
        at org.eclipse.jetty.ee9.nested.ScopedHandler.handle(ScopedHandler.java:127)
        at org.eclipse.jetty.ee9.security.SecurityHandler.handle(SecurityHandler.java:578)
[...]

PR #8335, opened 2022-07-25
Improve Resource to mapped ByteBuffer logic

lorban:
Closes https://github.com/eclipse/jetty.project/issues/8285.
Expose a way to figure if a Resource supports mmap'ing or not, and add extra BufferUtil methods that make use of that new API.

### DIFF
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingContentFactory.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingContentFactory.java
@@ -209,15 +209,7 @@ public class CachingContentFactory implements HttpContent.ContentFactory
             if (_useFileMappedBuffer)
             {
                 // map the content into memory
-                // TODO this is assuming the resource can be mapped! Inefficient to throw to test this
-                try
-                {
-                    byteBuffer = BufferUtil.toMappedBuffer(httpContent.getResource().getPath(), 0, _contentLengthValue);
-                }
-                catch (Throwable t)
-                {
-                    LOG.trace("ignored", t);
-                }
+                byteBuffer = BufferUtil.toMappedBuffer(httpContent.getResource(), 0, _contentLengthValue);
             }
 
             if (byteBuffer == null)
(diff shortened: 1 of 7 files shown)


jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingContentFactory.java
```diff
@@ -209,15 +209,7 @@ private CachingHttpContent(String key, String precalculatedEtag, HttpContent htt
...
-                // TODO this is assuming the resource can be mapped! Inefficient to throw to test this
-                try
-                {
-                    byteBuffer = BufferUtil.toMappedBuffer(httpContent.getResource().getPath(), 0, _contentLengthValue);
-                }
-                catch (Throwable t)
-                {
-                    LOG.trace("ignored", t);
-                }
+                byteBuffer = BufferUtil.toMappedBuffer(httpContent.getResource(), 0, _contentLengthValue);
```
joakime, line 212: If we have a Resource, is the start/length necessary?

joakime (approved):
I like it. :+1:

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingContentFactory.java
gregw: Regardless of the rest of the API... the ability to avoid throwing/catching/ignoring an exception in normal handling is huge.  So I'm favourable towards this PR already.

However, for code simplicity get rid of the if above and jave a one arg version so you can just do:
```suggestion
            ByteBuffer byteBuffer = _useFileMappedBuffer ? BufferUtil.toMappedBuffer(httpContent.getResource()) : null;
```

gregw (approved):
Other than a niggle LGTM

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingContentFactory.java
lorban: Done.

PR #8336, opened 2022-07-25
Take care of `Resource` deprecated methods

lorban:
`getInputStream()` and `getReadableByteChannel()` are confusing in their names as they don't make it clear to the caller that he has responsibility of closing the channel / input stream => rename with `new` instead of `get` prefix and undeprecated them.

Refactor `BufferUtil.toBuffer(Resource resource, boolean direct)` to use `newReadableByteChannel` and throw if the resource isn't valid.

`EmptyResource` is unneeded => remove it.


joakime (approved):
:+1:

gregw (changes requested):
I'm not OK with this change as it moves us away from `Resource` being an abstraction that can handle multiple sources of content.  It means users have to assume that a resource can provide a Path that is valid for a Files call.    This is not the case for `EmptyResource` or `ResourceCollection`.     Also, who knows, in future we might have a BytBufferResource or similar.


However, I do agree that the methods should not be getters, as it doesn't convey the ownership.
What about renaming them to `createInputStream`?     

Also, shouldn't `EmptyResource` produce an input stream that is already at EOF?

lorban:
@gregw this is one of those things that should be cleared in the `Resource` API. Currently, `getPath()` can return null and in some places we check this condition, in some other we don't.

I'm okay keeping those two methods around, but we need to better define what they mean in the context of `EmptyResource`, `ResourceCollection` and potential future resources.

Let me modify this PR to keep those two methods and let's see where this leads us.

joakime:
Can't we eliminate `EmptyResource` now?
Most of our use cases for `EmptyResource` is the `EmptyResource.INSTANCE`.
And that's basically used as a `null` test in code anyway, that `EmptyResource` never seems to get exposed anywhere.

lorban:
@joakime I'd love to get rid of `EmptyResource`, and this sounds doable as it's not exposed but only used internally.

It's currently used in `ee.MetaData` as a cache key and in `ee.MetaInfConfiguration` as a cached value to specify that we're caching the absence of any resource. Plus in `ResourceCollection` but that looks like an unreachable codepath.

I've committed a tentative patch that removes it. I think I managed to rewrite the logic without using `EmptyResource` but I'd be curious to know what @gregw or @janbartel think of this.

joakime (approved):
I like it.
Feels easier to follow too.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
```diff
@@ -1029,7 +1028,10 @@ public static ByteBuffer toBuffer(Resource resource, boolean direct) throws IOEx
             BufferUtil.readFrom(resource.getPath(), buffer);
         else
         {
-            try (InputStream is = resource.getInputStream();)
+            // TODO what does it mean to create a buffer from a resource that is not backed by a Path?
+            //  For instance, passing an EmptyResource to this method allocates and returns an empty buffer
+            //  while we could test for if (path == null) return EMPTY_BUFFER.
```
gregw: I think this is the wrong question.  Can all resources with a path be `readFrom`?   What if it is a directory?

gregw (approved):
Other than a niggle with your TODO comment (correct to ask a question there, but I'm not sure we are yet asking the right question), this is good.

So let's get it merged and work on the TODO.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
joakime: A directory cannot be read in the traditional sense.

Depending on the OS, a read from a directory might return directory metadata.

We have `Resource.isDirectory()` now, and it should be used properly.

gregw: So either:
 + we throw ISE if you try to read a directory
 + we return an empty InputStream
 + we return an InputStream over the HTML listing of the directory.

I like ISE for now.

joakime: I like throwing ISE

PR #8337, opened 2022-07-25
Jetty-12 : Immutable ResourceCollection and Context based mount management

joakime:
* Cleaned up constructors.
* No more setters for the (now `final`) internal `List<Resource>` field. (It's only constructed and populated in the constructors)
* Internal `List<Resource>` is now immutable so that `.getResources()` cannot be modified.
* Improved testing.
* Adding `Resource.isArchive` to test for jar/war/zip
* Adding new `Resource.fromList` implementatinos to help with various String lists (csv, arrays, collections, etc) to `List<Resource>` creation
* Adding `StringUtil.tokenizerAsIterator` to help reduce code duplication with `Resource.fromList()` implementations


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -21,36 +21,31 @@
...
- * Allows webapps to have multiple (static) sources.
+ * A collection of Resources.
+ * Allows webapps to have multiple sources.
  * The first resource in the collection is the main resource.
  * If a resource is not found in the main resource, it looks it up in
- * the order the resources were constructed.
+ * the order the provided in the constructors
  */
-public class ResourceCollection extends Resource
+public class ResourceCollection extends Resource implements AutoCloseable
```
lorban: Don't make `Resource` subclasses closeable, as `Resource` should always be a light GC'able object like `java.nio.Path` is.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -21,36 +21,31 @@
...
-     * Instantiates an empty resource collection.
-     * <p>
-     * This constructor is used when configuring jetty-maven-plugin.
-     */
-    public ResourceCollection()
-    {
-        _resources = new ArrayList<>();
-    }
+    private final List<Resource> _resources;
+    private final List<Mount> _mounted = new ArrayList<>();
```
lorban: Mounting should be external to the resource.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -79,56 +77,41 @@ public ResourceCollection(Collection<Resource> resources)
...
-                assertResourceValid(resource);
-                _resources.add(resource);
-            }
 
-            if (_resources.isEmpty())
-            {
-                throw new IllegalArgumentException("resources cannot be empty or null");
+                try
+                {
+                    Resource.Mount mount = Resource.mountIfNeeded(r);
```
lorban: Do not mount here, but `Resource.mount()` and `Resource.mountIfNeeded()` probably need to be modified to mount all contained resources if the passed resource actually is a `ResourceCollection`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -130,6 +155,8 @@ public static Resource.Mount mountIfNeeded(URI uri)
      */
     public static Resource.Mount mount(URI uri) throws IOException
     {
+        if (!isArchive(uri))
+            throw new IllegalArgumentException("URI is not a JAR: " + uri);
```
joakime: Whoops, this a typo. should read "URI is not a mountable Archive"

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -143,27 +170,127 @@ public static Resource.Mount mount(URI uri) throws IOException
      */
     public static Resource.Mount mountJar(Path path) throws IOException
     {
+        if (!isArchive(path))
+            throw new IllegalArgumentException("Path is not a JAR: " + path);
```
joakime: Ditto, "Path is not a mountable Archive:"

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -139,7 +122,13 @@ public ResourceCollection(String[] resources)
      */
     public ResourceCollection(String csvResources) throws IOException
     {
-        setResources(csvResources);
+        this(Resource.fromList(csvResources, false));
+    }
+
+    @Override
+    public void close() throws Exception
```
lorban: This shouldn't exist.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/ResourceTest.java
```diff
@@ -393,4 +396,58 @@ public void testDotAlias() throws Exception
         assertNotNull(same);
         assertTrue(same.isAlias());
     }
+
+    @ParameterizedTest
+    @ValueSource(strings = {
+        "file:/home/user/.m2/repository/com/company/1.0/company-1.0.jar",
+        "jar:file:/home/user/.m2/repository/com/company/1.0/company-1.0.jar!/",
```
lorban, line 419: Should this one be considered an archive? IMHO it's not as an archive should be mountable.

lorban (changes requested):
The `Resource` javadoc is probably lacking as the fact that it's a lightweight object isn't as obvious as it should be.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
joakime: I disagree, as that puts the mounting logic everywhere.
It needs to be centralized and done properly.

Ideally, the `ContextHandler.baseResource` (and `ServletContextHandler`, and `WebAppContext`) will all be a `ResourceCollection` (like every `WebAppContext` instance is since Jetty 9.x).

We don't want each Configuration managing mounting, and the ContextHandler, and the ServletContextHandler, and the WebAppContext, and the DefaultServlet, and the ResourceService, and the ResourceHandler, and the QuickStart (and that's just to name a few).

It needs to be centralized.

There's also a different discussion on how `ContextHandler.getBaseResource()` behaves, as it's been broken for MultiRelease jar files since they were introduced as a concept in Java.

Take this example.

Call to `WebAppContext.getBaseResource().resolve("WEB-INF")` returns what?
(about 98% of the time this is a `ResourceCollection` courtesy of 3 different `Configuration` behaviors).
This could return a `WEB-INF/lib/foo.jar!/WEB-INF/` or even `WEB-INF/lib/foo.jar!/META-INF/versions/11/WEB-INF`.
For users of `getBaseResource().resolve(String)`, they need more control over what to search. (the first entry, which might be the WAR? or all entries?)
This also has impacts on TLD searches, and even QuickStart.

If we had `PathCollection` still as the base for a ContextHandler, then you can `pathCollection.stream().filter(FileID::isWar).filter(p->p.getFileName().endsWith(".tld"))...`
Which you cannot do in `Resource` or `ResourceCollection`.
Instead, you have to ask, every time you call `.getBaseResource()`, are you a collection? what's the first entry? is it a war? lets search that only.  Oops, if there was a TLD search, then the WAR might not be first, as that was created by a Set from one of the Configuration classes.
Wait, how is an entry in the `ResourceCollection` also another `ResourceCollection`? (this happens with the Configuration classes too!)

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
joakime: A `ResourceCollection` that contains another `ResourceCollection` is a horrible idea and puts far more burden on the users of `context.getBaseResource()` to do things right.
Which happens in the `Configurations` btw.

I need to either reject, or unwrap, any seen `ResourceCollection` during the construction of a `ResourceCollection`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
joakime: This needs to exist, perhaps not here, but somewhere centralized.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/ResourceTest.java
joakime: It's most definitely an archive, as this is a valid URI to pass into the Resource APIs.

A common use case is Embedded Jetty users and uber jars.
Where they reference a Classloader URL to a resource base in their own JAR.
That URL will be fully complete with `jar:file://path/to/app.jar!/content/base/` as a reference.

joakime:
After a conversation with @lorban I think that we should eliminate ResourceCollection entirely, as once the Multi-Release JAR files were introduced into Java, lots of assumptions we used to make are now no longer valid.
The concept of a collection of resources or paths (with mounting) should be kept entirely within `ContextHandler.Context`.
This simplifies the behaviors, and eliminates the need for the `Resource` interface entirely.
Also, since our current AliasCheckers don't use Resource to do their alias checks, this also makes the process easier. (in fact, the AliasCheckers are currently broken if the Resource is a ResourceCollection)

gregw:
@lorban @joakime you have already tried and failed to remove `ResourceCollection`.  What has changed now to make that possible?

`Resource`s are now light-weight immutable stateless instances with mounting etc. dealt with externally.  Thus `ResourceCollection` can also be a lightwieight immutable and stateless (might need a builder pattern?)

A `ResourceCollection` should never need to mount a `Resource` it has been given as whatever created that resource in the first place should have done the mounting.  In the example of `WebAppContext.getBaseResource().resolve("WEB-INF")`, it is whatever created the base resource that should do the mounting and the resolved "WEB-INF" resource need not worry about mounting.  Of course once the webapp is stopped an the baseResource unmounted, then the web-inf resource will become invalid.... but so it should!

gregw (dismissed):
Not in favour of this approach.
However, I'm OK for some changes that ensure ResourceCollection is immutable.

lorban:
@gregw The debate between @joakime and myself started about the fact that `ResourceCollection` should not be closeable and should not mount / unmount. I pointed out that if that isn't obvious, that's a deficiency in the `Resource` javadoc.
I'm neutral about that `ContextHandler.Context` idea as I don't clearly see the grand picture.

But I do agree that `ResourceCollection` should be immutable. And IMHO the builder should be in `Resource` much like all other ways to create a resource instead of scattering that code in multiple classes.

joakime:
Since this `ResourceCollection` in `ContextHandler.Context.getBaseResource()` exists for the life of the Context, and is used by multiple classes for the lifecycle of the Context, where do we put the mounting?
It either makes sense in the `ResourceCollection` itself, or in the `ContextHandler.Context`.
This `ResourceCollection` is also created by multiple classes (Configurations), so we have to deal with that as well.

joakime:
> Not in favour of this approach.
> However, I'm OK for some changes that ensure ResourceCollection is immutable.

What do you mean by "of this approach"?
The changes already present in this PR? or the proposed grander scale changes?

joakime:
> @lorban @joakime you have already tried and failed to remove `ResourceCollection`. What has changed now to make that possible?

I have not tried.

The only reason for `Resource` to still exist is the belief that we have to expose `ResourceCollection` concepts as a `Resource` (which was decided by yourself and @lorban), but that's still broken, and no changes have been made to address the issues around keeping `ResourceCollection` as a `Resource` in that decision.

We have many places that have side-effects with this decision.
And they had issue all the way back in Jetty 7.
Those side effects were minor and we could live with them.
But the introduction of MultiRelease JAR files exploded the side effects greatly.

I have TODO's scattered around in this PR to show some of the warts of these side effects.

* https://github.com/eclipse/jetty.project/pull/8337/commits/09a5d710ca131646640b8cfc4bda3e906098bfcc

Then there's also the issues that `ResourceCollection` causes to other places that haven't been addressed.
`AliasCheck` for example doesn't yet address what happens when a `ResourceCollection` is passed to it.
Which is possible, if you `ResourceCollection.resolve(String)` and there is more than 1 hit, you get all of the hits back as another `ResourceCollection`.

The `HttpContent` layer will treat this return as a directory (because `ResourceCollection.isDirectory()` is always true), even though it might contain only files, and will attempt to serve a directory contents.

Having `ContextHandler.Context.getBaseResource()` or `Resource.resolve(String)` return possibly a `ResourceCollection` isn't handled well in our code.
We don't check the return on those methods for collection types in the places that matter.

The API contract `Resource.resolve(String)` is one of the places that complicates things when used against a `ResourceCollection`.

My proposal ...

1. Eliminate the idea that `ResourceCollection` is available as a `Resource`.
2. `Resource` contract (javadoc) will be changed to only represent a single `Resource` (never a collection)
3. Establish something new to represent "a collection of resource sources" as a low level API below Resource.
4. This new "collection of resource sources" manages it's own mounts. (it could be a lifecycle), this also eliminates Resource.Mount, Resource.mount() concepts from the Resource API.
5. This new "collection of resource sources" is used within `ContextHandler.Context` as a field/bean (we simplify 90% of the mounting concerns this way)
6. This new "collection of resource sources" can identify the "main archive" of it's sources (eg: in WebAppContext this would be the war file/dir) this will help mitigate some of the existing bugs.
7. This new "collection of resource sources" has APIs that clearly identify intended use at each code point wanting to use it can decide if it needs a single entry, entry from main archive, the list of entries across single archive, matching entries from all archives, etc.
8. This new "collection of resource sources" is what is exposed from `ContextHandler.Context`, not a `Resource`, the results _from_ API calls can still be `Resource`, but that's not necessary for this low level API.   Now the calls from user/app space need Resource, and what it brings to the table, but that should be built on top of the low level API, not be the low level API.
9. Use this new "collection of resource sources" in other non-context default locations: quickstart generation, runner, DefaultServlet.initParam("resourceBase"), ResourceHandler (without a ContextHandler), etc...

This proposal doesn't even break our current AliasCheck behaviors, as they are all based on Path now, not Resource.
In fact, if we use Resource above the low level API, then the AliasCheck behaviors suddenly get easier to implement as we don't have to worry about ResourceCollection behaviors now.

To be clear, I'm _not_ proposing to eliminate `Resource` entirely.
Just have it _always_ represent a single `Resource`, and use `Resource` 100% of the time when dealing with behaviors from users/protocols.
It wouldn't be used for `Configuration` implementations, `WebAppClassloader`, or `ServletContext.getResource()` (and similar APIs) as those don't need the uri canonicalization or alias check behaviors that `Resource` brings to the table.
Classes like `DefaultServlet` / `ResourceService` / `HttpContent` will still use `Resource` even after this proposal.

Also, I don't want to do this proposal in this PR.

For this PR, let `ResourceCollection` be immutable and manage it's own mounts (like it is now).
Then I'll submit a new PR later to show the proposal.

joakime:
Setting to draft as I implement the proposal from @lorban 

_From @lorban_
> I'd avoid exposing ResourceCollection's constructor, and create some new static builders for collections in Resource
> How about something like `Mount m = Resource.mountCollection(URI... uris)`
> then `m.root()` can return the `ResourceCollection`
> `Mount m = Resource.mountCollection(String csv)` ?

This also means new tests for using a `ResourceCollection` from XML (which I currently cannot find)

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
joakime: Done

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
joakime: Done

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
joakime: Done

gregw:
> Since this `ResourceCollection` in `ContextHandler.Context.getBaseResource()` exists for the life of the Context, and is used by multiple classes for the lifecycle of the Context, where do we put the mounting? It either makes sense in the `ResourceCollection` itself, or in the `ContextHandler.Context`. This `ResourceCollection` is also created by multiple classes (Configurations), so we have to deal with that as well.

I believe it is already in `WebInfConfiguration` when the resource base is created from the war.    It the context is not configured with a war, then who ever sets the resource base will need to track the mount.

Perhaps let's write some mount leak detection code to see if we are leaking before we start preemptively adding mounts/closes in places not needed.

gregw:
> My proposal ...

@joakime I think we are too deep into this release cycle for any more such revolutionary changes.   We gave several months over to trying to find an alternative abstraction, but that failed, primarily because the big picture was not understood.     We've gained experience now and perhaps next time there will be a better way.   But I don't think it can be done in the jetty-12 cycle.

I think the current improvements to `Resource` are working well. Whilst we may have some mounting issues to clean up, I doubt they are any worse than the resource leaks that exist in previous releases.   By all means let's clean it up (and I kind of like @lorban suggestion of a ResourceCollection builder API in Resource itself).

But I'm strongly opposed to `ResourceCollection` managing mounts.   It is the creator of the resources that are put into the collection that must manage the mounts - ie those resources are not valid unless their FS is already mounted.   If the collection mounts them again, then it is reference counting so they may exist beyond the life span of the creator of the collection... which is precisely what we do not want. We want a context to be able to clean up it's mounts.

joakime:
> I think the current improvements to `Resource` are working well. Whilst we may have some mounting issues to clean up, I doubt they are any worse than the resource leaks that exist in previous releases. By all means let's clean it up (and I kind of like @lorban suggestion of a ResourceCollection builder API in Resource itself).

This PR, in its current state, implements the `Resource.mountCollection()` ideas from @lorban.

> 
> But I'm strongly opposed to `ResourceCollection` managing mounts. It is the creator of the resources that are put into the collection that must manage the mounts - ie those resources are not valid unless their FS is already mounted. If the collection mounts them again, then it is reference counting so they may exist beyond the life span of the creator of the collection... which is precisely what we do not want. We want a context to be able to clean up it's mounts.

This PR has also moved the mounts out of `ResourceCollection`.
Please review again.

joakime:
> I believe it is already in `WebInfConfiguration` when the resource base is created from the war. It the context is not configured with a war, then who ever sets the resource base will need to track the mount.

It's not just `WebInfConfiguration`, it's also `WebAppContext`, `MetaInfConfiguration`, and `Quickstart` that all build their own `ResourceCollection` and just wrap the existing `context.getBaseResource()` that was returned to set a new `context.setBaseResource(Resource)`.   (and if we are using the `jetty-maven-plugin` it can be 2 more levels of `ResourceCollection`)

gregw:
> It's not just `WebInfConfiguration`, it's also `WebAppContext`, `MetaInfConfiguration`, and `Quickstart` that all build their own `ResourceCollection` and just wrap the existing `context.getBaseResource()` that was returned to set a new `context.setBaseResource(Resource)`. (and if we are using the `jetty-maven-plugin` it can be 2 more levels of `ResourceCollection`)

Sure we may have multiple sources of mounted jars, but each has their own unique lifecycle and they must be responsible for creating the mounts if they are needed.     

Ultimately, `ResourceCollection` is just a class/method that is passed `Resource`s for it to use.  It, like any other class/method that is passed a `Resource`, is entitled to just use that `Resource` without having to worry about mounting/unmounting.     It is the creator of a resource that must deal with mounting/unmounting, not the user of a resource.

I'll have a go at a simpler `ResourceCollection` cleanup.... stand by....

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -108,18 +141,42 @@ public static boolean getDefaultUseCaches()
...
             throw new IllegalArgumentException(ioe);
         }
     }
 
+    /**
+     * @param resourceBase the base resource that might need to be mounted.
+     * @return the {@link Mount} if a mount is needed, null if one is not needed.
+     * @see #mountIfNeeded(URI)
+     */
+    public static Mount mountIfNeeded(Resource resourceBase)
```
lorban: That looks weird: you should not be able to create a `Resource` that needs mounting.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -98,8 +98,41 @@ public static boolean getDefaultUseCaches()
         return __defaultUseCaches;
     }
 
+    public static Resource.Mount mountCollection(Collection<Resource> resources)
```
lorban: That looks weird: you should not be able to create a Resource that needs mounting. I think this method should return `Resource`, be called `newResourceCollection` and assume everything is properly mounted.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/ResourceCacheTest.java
```diff
@@ -112,67 +114,75 @@ public void testMultipleSources1() throws Exception
...
-        List<Resource> r = rc.getResources();
-        MimeTypes mime = new MimeTypes();
+        List<Resource> resourceList = Stream.of("one", "two", "three")
+            .map(basePath::resolve)
+            .map(Resource::newResource)
+            .toList();
 
-        CachedContentFactory rc3 = new CachedContentFactory(null, r.get(2), mime, false, false, CompressedContentFormat.NONE);
-        CachedContentFactory rc2 = new CachedContentFactory(rc3, r.get(1), mime, false, false, CompressedContentFormat.NONE)
+        try (Resource.Mount rcMount = Resource.mountCollection(resourceList))
```
lorban: Since the `Resource`s could be created, there's no need for mounting.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/ResourceCacheTest.java
```diff
@@ -112,67 +114,75 @@ public void testMultipleSources1() throws Exception
...
-            Resource.newResource(basePath.resolve("two")),
-            Resource.newResource(basePath.resolve("three")));
+        List<Resource> resourceList = Stream.of("one", "two", "three")
+            .map(basePath::resolve)
+            .map(Resource::newResource)
+            .toList();
 
-        List<Resource> r = rc.getResources();
-        MimeTypes mime = new MimeTypes();
+        try (Resource.Mount rcMount = Resource.mountCollection(resourceList))
```
lorban: Since the `Resource`s could be created, there's no need for mounting.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java
```diff
@@ -779,12 +778,14 @@ public void multipleResourcesHandler() throws Exception
         ResourceHandler handler = new ResourceHandler();
 
         // For multiple directories, use ResourceCollection.
-        ResourceCollection directories = new ResourceCollection();
-        // TODO: how to add paths to ResourceCollection?
-//        directories.addPath("/path/to/static/resources/");
-//        directories.addPath("/another/path/to/static/resources/");
-
-        handler.setBaseResource(directories);
+        Resource.Mount mount = Resource.mountCollection(
```
lorban: Since the `Resource`s could be created, there's no need for mounting.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -21,55 +21,66 @@
...
-    private List<Resource> _resources;
+    private static final Logger LOG = LoggerFactory.getLogger(ResourceCollection.class);
 
-    /**
-     * Instantiates an empty resource collection.
-     * <p>
-     * This constructor is used when configuring jetty-maven-plugin.
-     */
-    public ResourceCollection()
+    public static class Mount implements Resource.Mount
```
lorban: This should not be public.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -21,55 +21,66 @@
...
     {
-        _resources = new ArrayList<>();
-    }
+        private final List<Resource.Mount> _mounts;
+        private final ResourceCollection _resourceCollection;
```
lorban: Rename to `_root`?

jetty-ee10/jetty-ee10-maven-plugin/src/main/java/org/eclipse/jetty/ee10/maven/plugin/MavenWebAppContext.java
```diff
@@ -219,16 +219,18 @@ public boolean getBaseAppFirst()
...
-        List<String> resources = new ArrayList<String>();
-        for (String rl : resourceBases)
+        try
         {
-            String[] rs = StringUtil.csvSplit(rl);
-            for (String r : rs)
-            {
-                resources.add(r);
-            }
+            List<Resource> resources = Resource.fromList(List.of(resourceBases), false);
```
lorban: This looks broken, as trying to create a `Resource` out of a URI that needs to be mounted throws.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebAppContext.java
```diff
@@ -803,10 +803,12 @@ public Resource getWebInf() throws IOException
             return null;
 
         // Iw there a WEB-INF directory?
-        Resource webInf = getResourceBase().resolve("WEB-INF/");
+        Resource webInf = getResourceBase().resolve("WEB-INF/"); // TODO: what does this do in a collection?
```
lorban, line 807: That's a very interesting TODO, and that contract needs to be clearly defined.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -98,8 +98,41 @@ public static boolean getDefaultUseCaches()
...
+        {
+            // can't create ResourceCollection.Mount, unmount and rethrow.
+            mounts.forEach(IO::close);
+            throw t;
+        }
+    }
+
+    public static Resource.Mount mountCollectionURIs(Collection<URI> uris)
+    {
+        return mountCollection(uris.stream().map(Resource::newResource).toList());
```
lorban: This won't work as if one URI does need mounting, trying to create a `Resource` from it will throw. This should call `mountIfNeeded(URI uri)` and return a combined mount.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -98,8 +98,41 @@ public static boolean getDefaultUseCaches()
...
+    }
+
+    public static Resource.Mount mountCollection(String csvResources) throws IOException
+    {
+        return mountCollection(Resource.fromList(csvResources, false));
```
lorban: This won't work as if one csv entry does need mounting, trying to create a `Resource` from it will throw.

lorban (changes requested):
The mounting contract isn't clear enough as this change doesn't use it right.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebAppContext.java
joakime: There's no way to satisfy the Servlet spec with the existing Resource.resolve() contract and ResourceCollection.
It's super easy to wind up with things that are not part of the main archive.
This would be a Servlet specific contract issue, not a Resource specific contract issue (as we cannot push Servlet behaviors down to jetty-core)

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime: ~That's not how `Resource` currently works.~

I'll write some unit tests.

jetty-ee10/jetty-ee10-maven-plugin/src/main/java/org/eclipse/jetty/ee10/maven/plugin/MavenWebAppContext.java
joakime: This `resourceBases` is user space provided.
It can be simple paths (absolute or relative), or URI references (or a combination of both)

joakime: I believe the latest cleanup has addressed this concern.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebAppContext.java
joakime: This needs to be punted to a future issue/PR.
Out of scope for this PR.
Leaving the TODO in place is good though.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime: Unit tests added, implementation greatly simplified.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime: Completely reworked.

Now `Resource.split(String)` (simpler replacement for old-school `Resource.fromList()`) handles most of this by converting only to URI.
Letting other code manage the URI to Resource mapping.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java
joakime: Greatly simplified

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
joakime: Done

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/ResourceCacheTest.java
joakime: Greatly simplified

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/ResourceCacheTest.java
joakime: Greatly simplified

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime: Method removed

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime: Method removed

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
joakime: Set to package scope (so that `Resource` can use it)

lorban (approved):
LGTM

joakime:
@lorban a few more updates.

* Increase reliability of `WebAppContextTests`
* Introduced `Resource.unwrapContainer` to help in servlet cases where the raw JAR path should be represented in a `ServletContext` attribute.
* Made `FileSystemPool.containerUri` just use new `Resource.unwrapContainer` (final cleanup in future PR around FileID)
* webapp `MetaData` updated to use URIs references to Libs (not File objects) which helps in later ResourceCollection mount requirements.
* jetty-ee#-maven-plugin use URIs for its classpath tracking to aid in mounting issues later
* webapp extraClasspath supports raw JAR references as well as glob now, supported by change in `Resource.split(String)` to help with [jetty-ee#-maven-plugin expectations](https://github.com/eclipse/jetty.project/blob/fix/jetty-12-resourcecollection/tests/jetty-ee9-test-maven-plugin/src/it/jetty-start-mojo-it/jetty-simple-webapp/pom.xml#L121).

https://github.com/eclipse/jetty.project/blob/3e5cccc1ebcbd40ca59ba25d5777b609abb8f393/tests/jetty-ee9-test-maven-plugin/src/it/jetty-start-mojo-it/jetty-simple-webapp/pom.xml#L121

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/StringUtil.java
```diff
@@ -981,6 +984,36 @@ public static String[] arrayFromString(String s)
...
     }
 
+    /**
+     * Present the results of a {@link StringTokenizer} as an {@link Iterator} of type {@link String}
+     * @param input the StringTokenizer input
+     * @param delim the StringTokenizer delim
+     * @return the resulting iterator for the StringTokenizer
+     * @see StringTokenizer#StringTokenizer(java.lang.String, java.lang.String)
+     */
+    public static Iterator<String> tokenizerAsIterator(String input, String delim)
```
joakime, line 994: I think I can get rid of this now.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -352,6 +352,29 @@ public static String toJarPath(String jarFile, String pathInJar)
...
 
+    /**
+     * Unwrap a URI to expose its container path reference.
+     *
+     * Take out the container archive name URI from a {@code jar:file:${container-name}!/} URI.
+     *
+     * @param uri the input URI
+     * @return the container String if a {@code jar} scheme, or just the URI untouched.
+     */
+    public static URI unwrapContainer(URI uri)
```
lorban, line 363: Since that's completely `Resource`-agnostic, this can probably be moved to `URIUtil`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -1329,8 +1352,12 @@ public Collection<Resource> getAllResources()
...
      *     Each part of the input string could be path references (unix or windows style), or string URI references.
      * </p>
+     * <p>
+     *     If the result of processing the input segment is a java archive, then its resulting URI will be a mountable URI as `jar:file:...!/`.
+     * </p>
      *
      * @param str the input string of references
+     * @see #toJarFileUri(URI)
      */
     public static List<URI> split(String str)
```
lorban, line 1362: Since that's completely `Resource`-agnostic, this can probably be moved to `URIUtil`.

jetty-ee10/jetty-ee10-maven-plugin/src/main/java/org/eclipse/jetty/ee10/maven/plugin/MavenWebAppContext.java
```diff
@@ -282,15 +283,21 @@ public void doStart() throws Exception
...
         // Set up the classpath
-        _classpathFiles = new ArrayList<>();
-        _classpathFiles.addAll(_webInfClasses);
-        _classpathFiles.addAll(_webInfJars);
+        _classpathUris = new ArrayList<>();
+        _webInfClasses.forEach(f -> _classpathUris.add(f.toURI()));
+        _webInfJars.forEach(f ->
+        {
+            // ensure our JAR file references are `jar:file:...` URI references
+            URI jarFileUri = Resource.toJarFileUri(f.toURI());
```
lorban, line 297: Since `Resource.toJarFileUri` is completely `Resource`-agnostic, it can probably be moved to `URIUtil`.

jetty-ee9/jetty-ee9-maven-plugin/src/main/java/org/eclipse/jetty/ee9/maven/plugin/MavenWebInfConfiguration.java
```diff
@@ -51,15 +51,15 @@ public void configure(WebAppContext context) throws Exception
...
         {
             if (LOG.isDebugEnabled())
                 LOG.debug("Setting up classpath ...");
-            WebAppClassLoader loader = (WebAppClassLoader)context.getClassLoader();
-            for (File classpath : jwac.getClassPathFiles())
+
+            for (URI uri : jwac.getClassPathUris())
             {
-                loader.addClassPath(classpath.getCanonicalPath());
+                loader.addClassPath(uri.toASCIIString());
```
lorban, line 62: Why are you using `toASCIIString` instead of `toString`?

jetty-ee10/jetty-ee10-maven-plugin/src/main/java/org/eclipse/jetty/ee10/maven/plugin/MavenWebInfConfiguration.java
```diff
@@ -52,14 +52,13 @@ public void configure(WebAppContext context) throws Exception
...
+        if (jwac.getClassPathUris() != null && context.getClassLoader() instanceof WebAppClassLoader loader)
         {
             if (LOG.isDebugEnabled())
                 LOG.debug("Setting up classpath ...");
-            WebAppClassLoader loader = (WebAppClassLoader)context.getClassLoader();
-            for (File classpath : jwac.getClassPathFiles())
+            for (URI uri : jwac.getClassPathUris())
             {
-                loader.addClassPath(classpath.getCanonicalPath());
+                loader.addClassPath(uri.toASCIIString());
```
lorban, line 61: It's probably worth writing a test that uses a `URI` pointing to a file that has a Swedish "å" character in its name to check that we can handle internationally-named jar files.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebAppClassLoader.java
```diff
@@ -282,11 +283,19 @@ public void addClassPath(String classPath)
...
     {
         int dot = file.lastIndexOf('.');
-        return dot != -1 && _extensions.contains(file.substring(dot));
+        return dot != -1 && _extensions.contains(file.substring(dot).toLowerCase(Locale.ENGLISH));
+    }
+
+    // TODO: move to FileID in later PR
+    private boolean isFileSupported(Path path)
+    {
+        return isFileSupported(path.getFileName().toString());
```
lorban, line 298: Did you know that `Path.of("/").getFileName()` returns `null`?

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebAppClassLoader.java
```diff
@@ -299,32 +308,35 @@ public void addJars(Resource lib)
...
-                        String fnlc = resource.getName().toLowerCase(Locale.ENGLISH);
-                        // don't check if this is a directory (prevents use of symlinks), see Bug 353165
-                        if (isFileSupported(fnlc))
-                        {
-                            String jar = URIUtil.encodeSpecific(resource.toString(), ",;");
-                            addClassPath(jar);
-                        }
+                            LOG.debug("addJar - {}", jar);
+                        URI jarUri = Resource.toJarFileUri(jar.toUri());
+                        addClassPath(jarUri.toASCIIString());
```
lorban, line 328: Another `toASCIIString` that worries me w.r.t files with non-ascii characters in their names.

lorban (changes requested):
We're getting very close!

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime: Absolutely, and is one of several static methods that will move with that future FileID/URIUtil PR.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime: Ditto, the future FileID/URIUtil PR.

jetty-ee10/jetty-ee10-maven-plugin/src/main/java/org/eclipse/jetty/ee10/maven/plugin/MavenWebAppContext.java
joakime: Ditto, the future FileID/URIUtil PR.

jetty-ee9/jetty-ee9-maven-plugin/src/main/java/org/eclipse/jetty/ee9/maven/plugin/MavenWebInfConfiguration.java
joakime: `toString()` returns the input string in most cases.

`toASCIIString()` returns a properly encoded (only if input contains characters in the unicode space that need encoding) string in all cases.

jetty-ee10/jetty-ee10-maven-plugin/src/main/java/org/eclipse/jetty/ee10/maven/plugin/MavenWebInfConfiguration.java
joakime: We have those tests already.
And the use of `uri.toASCIIString()` ensure they are properly URI+UTF-8 encoded (if the input string hasn't encoded them).

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebAppClassLoader.java
joakime: I'm aware of that.  
This is why the use of this is in a filter preceded by `.filter(Files:isRegularFile)`

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebAppClassLoader.java
joakime: Using `toASCIIString` doesn't strip characters, it ensure that the entire string is ASCII characters, even if the original input string contains unicode (by encoding).

lorban (approved):

PR #8338, opened 2022-07-25
Jetty-12 : Review MultiException in light of addSuppressed options

joakime:
After a conversation with @lorban it was pointed out that `MultiException` should be removed in favor of just using `addSuppressed(Throwable)` instead.

* Cleanup of `MultiException` to be easier to follow/understand by using Suppressed Throwables internally.
* Improved `MultiExceptionTest`
* Add `TypeUtil.withSuppressed(Throwable, List<Throwable> suppressed)` to make working with error lists easier.
* Converted `HandlerCollection` and `XmlConfiguration` to show how to use Error Lists instead.

### DIFF
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/MultiException.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/MultiException.java
@@ -13,9 +13,9 @@
 
 package org.eclipse.jetty.util;
 
-import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * Wraps multiple exceptions.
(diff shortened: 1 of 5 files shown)


jetty-core/jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlConfiguration.java
```diff
@@ -735,12 +733,8 @@ private void set(Object obj, XmlParser.Node node) throws Exception
...
             if (types != null)
                 message += ". Found setters for " + types;
-            NoSuchMethodException failure = new NoSuchMethodException(message);
-            for (int i = 0; i < me.size(); i++)
-            {
-                failure.addSuppressed(me.getThrowable(i));
-            }
-            throw failure;
+
+            throw TypeUtil.withSuppressed(new NoSuchMethodException(message), errors);
```
joakime, line 737: There's no test for `errors.isEmpty()` here as at this point we must throw an exception.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/MultiException.java
```diff
@@ -24,195 +24,130 @@
...
-                addSuppressed(t);
-        }
+        // Avoid filling in stack trace information.
+        super(message, null, true, false);
     }
 
     public void add(Throwable e)
     {
-        if (e instanceof MultiException)
+        if (e instanceof MultiException mex)
```
gregw, line 48: Before adding suppress we should check org.eclipse.jetty.util.TypeUtil#isAssociated.   Perhaps this entire method can just be:
```suggestion
        if (!TypeUtil#isAssociated(this,e))
            addSuppressed(e);
```

gregw (changes requested):
Rather than deprecate and remove, I think there is value is just making the a class with a few static utility methods that work more or less how the old lazy list did (ie starts with a null value).
For example, one of the destroy usages of MultiException could be changed to:
```java

    @Override
    public void destroy()
    {
        Throwable multi = null;
        if (_configurations != null)
        {
            for (Configuration configuration : _configurations)
            {
                try
                {
                    configuration.destroy(this);
                }
                catch (Exception e)
                {
                    multi = MultiException.add(multi, e);
                }
            }
        }
        _configurations = null;
        super.destroy();
        MultiException.ifExceptionalThrowRuntime(multi);
    }
```

Eitherway, let's not do an intermediary step. Let's decide the final form and go straight to it. If we step halfway, then we'll get stuck there for sure!

If we are really worried about backwards compatibility, perhaps we keep the old API as deprecated but use the new static version ourselves.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/MultiExceptionTest.java
```diff
@@ -31,93 +27,42 @@ public void testEmpty() throws Exception
...
+        assertEquals(run, re1);
+
+        MultiException mex = assertThrows(MultiException.class, me::ifExceptionThrowMulti);
+        assertEquals(me, mex);
+
+        RuntimeException re2 = assertThrows(RuntimeException.class, me::ifExceptionThrowRuntime);
+        assertEquals(run, re2);
     }
 
     private MultiException multiExceptionWithRtIo()
```
joakime, line 68: Should remove this unused method.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/TypeUtil.java
```diff
@@ -621,6 +621,19 @@ public static void dump(ClassLoader cl)
...
+    /**
+     * Decorate a Throwable with the suppressed errors and return it.
+     * @param t the throwable
+     * @param errors the list of errors
+     * @return the original throwable with suppressed errors
+     * @param <T> of type Throwable
+     */
+    public static <T extends Throwable> T withSuppressed(T t, List<Throwable> errors)
+    {
+        errors.forEach(t::addSuppressed);
```
gregw, line 633: This needs to check if the errors are associated with `TypeUtil.isAssociated`.

gregw (changes requested):

joakime:
Closing in favor of rework in PR #8340

PR #8340, opened 2022-07-26
Refactor MultiException into static ExceptionUtil

gregw:
@joakime @lorban this is an experiment  of 3 API styles:
 + static invocation
 + Builder style
 + Builder with `callAndCatch` helper

See javadoc on MultiException for examples of the styles.

thoughts?


lorban:
@gregw I like both styles 1 and 2 as they keep the original exception type and add further exceptions as suppressed, i.e.: never throw `MultiException` itself. So maybe it should not be a `Throwable` anymore and be renamed to `ExceptionUtil` to make it explicit?

Style 1 is more verbose but saves the builder's allocation. Style 2 is more elegant and readable. Styles 2 and 3 are basically the same, with 3 adding an extra bit of sugar on top.

My preference goes for style 2/3 as I don't think the builder allocation would matter in most cases where a loop-that-must-catch is used. But we could keep style 1 around to optimize those fast-path loops.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/MultiException.java
```diff
@@ -13,36 +13,149 @@
...
+        if (multi instanceof RuntimeException runtimeException)
+            throw runtimeException;
+        if (multi instanceof Error error)
+            throw error;
+        throw new RuntimeException(multi);
+    }
+
     private static final String DEFAULT_MESSAGE = "Multiple exceptions";
 
+    @Deprecated
```
joakime: If we are keeping `MultiException`, why are we deprecating?
Just remove the fields/methods, or set them private to disallow use.
Heck set the class itself to be final too.

joakime (changes requested):
Things I do like about the new code ...

* The `MultiException.ifExceptionalThrow()` methods now introduce a stracktrace that points to where it was thrown, which the old code didn't. (that was annoying at times)
* The `MultiException.callAndCatch()` method makes lambdas / streams / foreach easier to deal with. yay!
* The `MultiException` itself seems to no longer be thrown? (or is it popping out of `ifExceptionThrow(Throwable multi)` due to it being an `Exception` type when using the builder technique?)

All 3 styles don't address the needs to throw a specific exception in some cases.

* `NoSuchMethodException` - in `XMLConfiguration`
* `ServletException` - in `HandlerCollection`
* `UnavailableException` - deep in the `ServletContextHandler` layer (this one is spread out in potentially many classes)

And that's just the ones I've found so far.

Go ahead and revert the changes in my branch for `XmlConfiguration` and `HandlerCollection` and see how this new `MultiException` fits in.
Look at the `ServletHolderTest` for some examples of `UnavailableException` (in fact we even catch `MultiException` in a few places there)

How about, as a way to avoid `MultiException` itself being thrown  ..
* If there is 1 `Throwable` being tracked, throw that.
* If there are more than 1 `Throwable` being tracked, the first `Throwable` should have the rest of the `Throwables` be suppressed on it, and then thrown.

I also agree with @lorban 's renaming `MultiException` to `ExceptionUtil` and not having it be a `Throwable` itself.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/MultiException.java
```diff
@@ -13,36 +13,149 @@
...
+            }
+            catch (Throwable t)
+            {
+                add(t);
+            }
+            return this;
+        }
+
+
+    }
```
joakime: Checkstyle violation here

`jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/MultiException.java:105:9: '}' has more than 1 empty lines after. [EmptyLineSeparator]`

joakime:
@gregw did you want to fix the `TypeUtil.withSuppressed()` here? or just remove it in favor of this approach.

* https://github.com/eclipse/jetty.project/pull/8338#discussion_r929426394

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/MultiException.java
gregw: This current version keeps MultiException around in more or less it's current form as a deprecated backwards compatible API.
But if we are happy to break backward compatibility in major version (I am), then renaming to `ExceptionUtil` makes sense.... we can also move the new `TypeUtil.isAssociated` over to `ExceptionUtil`

gregw:
I've entirely rebased this PR to go direct to jetty-12.0.x and it now contains the `ExceptionUtil` class only and I have removed the old `MultiException`

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/ExceptionUtil.java
```diff
@@ -0,0 +1,289 @@
...
+     * @param <T> The type of the {@link Throwable} to be thrown if <code>throwable</code> is not null.
+     * @param type The type of the Throwable to be thrown. It must have a constructor that
+     * takes a {@link Throwable} as a cause.
+     * @param throwable A {@link Throwable} or null.
+     * @throws Error If the passed {@link Throwable} is an {@link Error}.
+     * @throws RuntimeException If the passed {@link Throwable} is a {@link RuntimeException}.
+     * @throws T Thrown in <code>throwable</code> is not null and neither an {@link Error} nor {@link RuntimeException}.
+     * @throws IllegalArgumentException if the passed <code>type</code> cannot be constructed with a cause.
+     */
+    public static <T extends Throwable> void ifExceptionThrowAs(Class<T> type, Throwable throwable) throws
```
joakime, line 118: Neat! :+1:

joakime (approved):
I like it!
Merge it.

joakime:
> I've entirely rebased this PR to go direct to jetty-12.0.x and it now contains the `ExceptionUtil` class only and I have removed the old `MultiException`

I'm good with closing #8338 then?
Meh, I'll close it anyway. (we can always reopen if need be, but I can't imagine why)

gregw:
@lorban I'll hold off merging this a few hours so you can have look.  Can you check this first thing.

lorban (approved):
That's quite elegant, LGTM.

PR #8341, opened 2022-07-26
fix some tests for Jetty 12 ee10 servlets

lachlan-roberts:
fix ee10 servlet tests associated with dispatching
fix some tests failing due to the Servlet 6 URI encoding changes


jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -669,7 +670,7 @@ public void setContextPath(String contextPath)
     {
         if (isStarted())
             throw new IllegalStateException(getState());
-        _contextPath = contextPath;
+        _contextPath = URIUtil.normalizePath(contextPath);
```
gregw: See #8343, as that will change this to canonicalizePath. ... race you to merge :)

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextResponse.java
```diff
@@ -134,11 +134,17 @@ public ServletApiResponse getServletApiResponse()
...
-        _httpOutput.reopen();
+        _httpServletResponse.resetBuffer();
         _outputType = OutputType.NONE;
     }
 
+    public void included()
+    {
+        if (_outputType == OutputType.WRITER)
+            _writer.reopen();
+        _httpOutput.reopen();
```
gregw, line 145: Why is this needed?  I thought the include response wrapper intercepts close, so it cannot be closed?
Is this for any sendErrors within an incliude?  A comment explaining would be good.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletHandler.java
```diff
@@ -775,7 +775,7 @@ public void addServletWithMapping(ServletHolder servlet, String pathSpec)
 
             ServletMapping mapping = new ServletMapping();
             mapping.setServletName(servlet.getName());
-            mapping.setPathSpec(pathSpec);
+            mapping.setPathSpec(URIUtil.normalizePath(pathSpec));
```
gregw: will be renamed to canonicalPath after #8343

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletHandler.java
```diff
@@ -887,7 +887,7 @@ public void addFilterWithMapping(FilterHolder holder, String pathSpec, EnumSet<D
 
             FilterMapping mapping = new FilterMapping();
             mapping.setFilterName(holder.getName());
-            mapping.setPathSpec(pathSpec);
+            mapping.setPathSpec(URIUtil.normalizePath(pathSpec));
```
gregw: ditto

gregw (changes requested):
I think we need to ponder our space handling for a little bit.
The servlet 6.0 spec does allow us to return some encoded characters in servletPath and pathInfo, but that is primarily intended for things like `%2F`.

We are currently leaving `%20` encoded because the JVM URI class barfs on space as an illegal character.  But perhaps we should decode for the servlet  API and then ensure we re-encode in the API - which is something that #8343 has added to some extent.

So whilst we ponder, can you add additional tests to this PR that test encoded %2F and %3F in servlet patterns.  That way we can later change our %20 handling and still test the encoding this PR is doing.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextResponse.java
lachlan-roberts: I was just replicating all the places where this was called from Jetty 11. So maybe this is no longer needed. But then this will just be a NOOP anyway.

joakime (approved):
I like it!

I'm anxious to get my hands on these changes for the various DefaultServlet tests i'm currently working on too.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextResponse.java
gregw: Be wary of just copying behavours from jetty-11 as much as changed around things like dispatch and content management.  So good to question anything that looks suspect.
I'm guessing it is for sendErrors... but good to verify that it does do something sometimes and to then add a comment.

gregw (approved):
Since this is blocking other work, I will approve.
But can you investigate if the reopen after included really is needed and add a comment explaining it (no need for PR).

PR #8343, opened 2022-07-26
Jetty 12.0.x canonical uri

gregw:
+ Correct the usage of the terms canonical and normal in `URIUtil`
+ canonical paths are always normalized
+ canonicalize rather than just normalize paths from the servlet API methods

### DIFF
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
@@ -740,7 +740,7 @@ public class URIUtil
         {
             Utf8StringBuilder builder = null;
             int end = offset + length;
-            boolean slash = false;
+            boolean slash = true;
             boolean normal = true;
             for (int i = offset; i < end; i++)
             {
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
@@ -586,7 +586,7 @@ public class URIUtil
 
     /**
      * Decode a URI path and strip parameters of UTF-8 path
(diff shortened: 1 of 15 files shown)


gregw:
The servlet 6.0 spec improved the handling of dangerous encoded characters in a URI path.  Specifically it now allows for some characters like '/' to remain encoded in the values returned from `getServletPath()` and `getPathInfo()`.    Thus the space is no longer a decoded path, but a minimal-for-safety-encoded path space.

Currently in Jetty-12 we have embraced this approach and now work with canonically encoded URI paths with have a well defined list of characters that are left encoded.  This includes "%2F" for `/`, but it also includes characters that are not accepted by the jvm `URI` API like space (%20).

Perhaps we should be less "safe" and also decode %20 to space, but at the cost that we would need to re-encode that before resolving against `Resource`.  This PR puts a re-canonicalization step in for all paths from the servlet API, so it is mostly OK to decode space.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -1181,37 +1200,24 @@ else if (slash)
...
-     * <p>
-     * All segments of "." and ".." are factored out.
-     * Null is returned if the path tries to .. above its root.
+     * <p>Normalize a URI path by factoring out all segments of "." and "..".
+     * Null is returned if the path is normalized above its root.
      * </p>
      *
      * @param path the decoded URI path to convert. Any special characters (e.g. '?', "#") are assumed to be part of
      * the path segments.
      * @return the canonical path, or null if path traversal above root.
```
lachlan-roberts: Comment here still refers to it as the canonical path, not the normalized path.

Same for `normalizePathQuery` above.

lachlan-roberts (changes requested):

joakime (approved):

lachlan-roberts (approved):

PR #8345, opened 2022-07-26
remove old Resource defaultUseCache mechanism

lorban:
Tentative of getting rid of the `Resource` `defaultUseCaches` mechanism. 
It control two things:

1) The parameter passed to `JarURLConnection.setUseCaches()`.
2) Whether or not the `JarFile` backing the `JarURLConnection` gets closed after browsing or not.

(1) isn't relevant anymore if we mount the JAR as a resource instead of using a `JarURLConnection` to browse it, and for (2) I'm not sure why `JarFile`s were left open as I don't think we're going to browse those JARs more than once, unless I missed something.

Let's keep this on hold until https://github.com/eclipse/jetty.project/pull/8328 gets merged.

### DIFF
--- a/jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
+++ b/jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
@@ -22,6 +22,7 @@ import java.net.URLClassLoader;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.ArrayList;
@@ -719,21 +720,29 @@ public class MetaInfConfiguration extends AbstractConfiguration
     public Collection<URL> getTlds(URI uri) throws IOException
     {
         HashSet<URL> tlds = new HashSet<>();
-        URI jarUri = uriJarPrefix(uri, "!/");
-        try (Resource.Mount mount = Resource.mount(jarUri);
+        try (Resource.Mount mount = Resource.mount(uriJarPrefix(uri, "!/"));
              Stream<Path> stream = Files.walk(mount.root().getPath()))
         {
-            Iterator<String> it = stream
-                .map(Path::toString)
-                .filter(filename -> filename.startsWith("META-INF") && filename.endsWith(".tld"))
+            Iterator<Path> it = stream
+                .filter(MetaInfConfiguration::isTldFile)
                 .iterator();
-            String prefix = jarUri.toString();
             while (it.hasNext())
             {
-                tlds.add(new URL(prefix + it.next()));
+                Path entry = it.next();
+                tlds.add(entry.toUri().toURL());
             }
         }
         return tlds;
+    }
+
+    private static boolean isTldFile(Path path)
+    {
+        if (path.getNameCount() < 2)
+            return false;
+        if (!path.getName(0).toString().equalsIgnoreCase("META-INF"))
+            return false;
+        String filename = path.getFileName().toString();
+        return filename.toLowerCase(Locale.ENGLISH).endsWith(".tld");
     }
 
     protected List<Resource> findClassDirs(WebAppContext context)
--- a/jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
+++ b/jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
@@ -15,7 +15,6 @@ package org.eclipse.jetty.ee10.webapp;
 
 import java.io.File;
 import java.io.IOException;
-import java.net.JarURLConnection;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
@@ -30,17 +29,16 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
-import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.jar.JarEntry;
-import java.util.jar.JarFile;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.eclipse.jetty.util.IO;
 import org.eclipse.jetty.util.PatternMatcher;
@@ -720,25 +718,21 @@ public class MetaInfConfiguration extends AbstractConfiguration
      */
     public Collection<URL> getTlds(URI uri) throws IOException
     {
-        HashSet<URL> tlds = new HashSet<URL>();
-
+        HashSet<URL> tlds = new HashSet<>();
         URI jarUri = uriJarPrefix(uri, "!/");
-        URL url = jarUri.toURL();
-        JarURLConnection jarConn = (JarURLConnection)url.openConnection();
-        jarConn.setUseCaches(Resource.getDefaultUseCaches());
-        JarFile jarFile = jarConn.getJarFile();
-        Enumeration<JarEntry> entries = jarFile.entries();
-        while (entries.hasMoreElements())
+        try (Resource.Mount mount = Resource.mount(jarUri);
+             Stream<Path> stream = Files.walk(mount.root().getPath()))
         {
-            JarEntry e = entries.nextElement();
-            String name = e.getName();
-            if (name.startsWith("META-INF") && name.endsWith(".tld"))
+            Iterator<String> it = stream
+                .map(Path::toString)
+                .filter(filename -> filename.startsWith("META-INF") && filename.endsWith(".tld"))
+                .iterator();
+            String prefix = jarUri.toString();
+            while (it.hasNext())
             {
-                tlds.add(new URL(jarUri + name));
+                tlds.add(new URL(prefix + it.next()));
             }
         }
-        if (!Resource.getDefaultUseCaches())
-            jarFile.close();
         return tlds;
     }
 
--- a/jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
+++ b/jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
@@ -720,8 +720,11 @@ public class MetaInfConfiguration extends AbstractConfiguration
     public Collection<URL> getTlds(URI uri) throws IOException
     {
         HashSet<URL> tlds = new HashSet<>();
-        try (Resource.Mount mount = Resource.mount(uriJarPrefix(uri, "!/"));
-             Stream<Path> stream = Files.walk(mount.root().getPath()))
+        Resource.Mount mount = Resource.mount(uriJarPrefix(uri, "!/"));
+        if (_mountedResources == null)
+            _mountedResources = new ArrayList<>();
+        _mountedResources.add(mount);
+        try (Stream<Path> stream = Files.walk(mount.root().getPath()))
         {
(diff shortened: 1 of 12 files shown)


jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
```diff
@@ -720,25 +718,21 @@ public FileVisitResult visitFile(Path f, BasicFileAttributes attrs) throws IOExc
...
-        JarURLConnection jarConn = (JarURLConnection)url.openConnection();
-        jarConn.setUseCaches(Resource.getDefaultUseCaches());
-        JarFile jarFile = jarConn.getJarFile();
-        Enumeration<JarEntry> entries = jarFile.entries();
-        while (entries.hasMoreElements())
-        {
-            JarEntry e = entries.nextElement();
-            String name = e.getName();
-            if (name.startsWith("META-INF") && name.endsWith(".tld"))
+        try (Resource.Mount mount = Resource.mount(jarUri);
```
joakime: This won't honor multi-release jars in this PR.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
```diff
@@ -720,25 +718,21 @@ public FileVisitResult visitFile(Path f, BasicFileAttributes attrs) throws IOExc
...
+        {
+            Iterator<String> it = stream
+                .map(Path::toString)
+                .filter(filename -> filename.startsWith("META-INF") && filename.endsWith(".tld"))
+                .iterator();
+            String prefix = jarUri.toString();
+            while (it.hasNext())
             {
-                tlds.add(new URL(jarUri + name));
+                tlds.add(new URL(prefix + it.next()));
```
joakime: This can be done entirely with Path and URI.
Don't `map(Path::toString())`.
Let the Stream/Iterator be Path.
Then just use `path.toURI().toURL()` here.
This will also keep the encoding proper the entire way through, which this current code doesn't do.

Btw, this change was done in the MultiReleaseJarFile PR too.

jetty-ee9/jetty-ee9-webapp/src/main/java/org/eclipse/jetty/ee9/webapp/MetaInfConfiguration.java
```diff
@@ -714,25 +712,21 @@ public FileVisitResult visitFile(Path f, BasicFileAttributes attrs) throws IOExc
...
+        {
+            Iterator<String> it = stream
+                .map(Path::toString)
+                .filter(filename -> filename.startsWith("META-INF") && filename.endsWith(".tld"))
+                .iterator();
+            String prefix = jarUri.toString();
+            while (it.hasNext())
             {
-                tlds.add(new URL(jarUri + name));
+                tlds.add(new URL(prefix + it.next()));
```
joakime: Ditto, don't map to String, use Path.toURI().toURL().
Also, the filename suffix needs to be case insensitive, it might end with `.TLD` or `.Tld`

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
```diff
@@ -720,25 +718,21 @@ public FileVisitResult visitFile(Path f, BasicFileAttributes attrs) throws IOExc
      */
     public Collection<URL> getTlds(URI uri) throws IOException
     {
-        HashSet<URL> tlds = new HashSet<URL>();
-
+        HashSet<URL> tlds = new HashSet<>();
         URI jarUri = uriJarPrefix(uri, "!/");
```
joakime: The `MultiReleaseJarFile` PR has a fix for this as well.
`URI Resource.toJarURI(URI)

joakime (changes requested):
The Files.walk usage needs some help.
Some of these changes have been made in the MultiReleaseJarFile PR, you could look there as well.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
```diff
@@ -720,25 +718,21 @@ public FileVisitResult visitFile(Path f, BasicFileAttributes attrs) throws IOExc
...
-        {
-            JarEntry e = entries.nextElement();
-            String name = e.getName();
-            if (name.startsWith("META-INF") && name.endsWith(".tld"))
+        try (Resource.Mount mount = Resource.mount(jarUri);
+             Stream<Path> stream = Files.walk(mount.root().getPath()))
+        {
+            Iterator<String> it = stream
+                .map(Path::toString)
+                .filter(filename -> filename.startsWith("META-INF") && filename.endsWith(".tld"))
```
joakime: Create a new static method in this class ...

``` java
    public static boolean isTldFile(Path path)
    {
        if (path.getNameCount() < 2)
            return false;
        if (!path.getName(0).toString().equalsIgnoreCase("META-INF"))
            return false;
        String filename = path.getFileName().toString();
        return filename.toLowerCase(Locale.ENGLISH).endsWith(".tld");
    }
```

And use it in place of this anonymous filter.

``` java
.filter(MetaInfConfiguration::isTldFile)
```

jetty-ee9/jetty-ee9-webapp/src/main/java/org/eclipse/jetty/ee9/webapp/MetaInfConfiguration.java
joakime: Copy/use the `isTldFile` method from ee10.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
lorban: Done.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
lorban: Done.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
lorban: Done.

jetty-ee9/jetty-ee9-webapp/src/main/java/org/eclipse/jetty/ee9/webapp/MetaInfConfiguration.java
lorban: Done.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
lorban: Yes, but hopefully the MultiReleaseJarFile PR will take care of that with minimal changes. Or maybe this PR should wait for the MultiReleaseJarFile PR to get merged?

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
```diff
@@ -720,28 +719,32 @@ public FileVisitResult visitFile(Path f, BasicFileAttributes attrs) throws IOExc
...
-        {
-            JarEntry e = entries.nextElement();
-            String name = e.getName();
-            if (name.startsWith("META-INF") && name.endsWith(".tld"))
+        HashSet<URL> tlds = new HashSet<>();
+        try (Resource.Mount mount = Resource.mount(uriJarPrefix(uri, "!/"));
+             Stream<Path> stream = Files.walk(mount.root().getPath()))
+        {
+            Iterator<Path> it = stream
+                .filter(MetaInfConfiguration::isTldFile)
```
joakime, line 730: Add another `.filter(Files::isRegularFile)` before this.
Otherwise this will match directories like `/META-INF/foo.tld`, which is odd.

jetty-ee9/jetty-ee9-webapp/src/main/java/org/eclipse/jetty/ee9/webapp/MetaInfConfiguration.java
```diff
@@ -714,28 +712,32 @@ public FileVisitResult visitFile(Path f, BasicFileAttributes attrs) throws IOExc
...
-        {
-            JarEntry e = entries.nextElement();
-            String name = e.getName();
-            if (name.startsWith("META-INF") && name.endsWith(".tld"))
+        HashSet<URL> tlds = new HashSet<>();
+        try (Resource.Mount mount = Resource.mount(uriJarPrefix(uri, "!/"));
+             Stream<Path> stream = Files.walk(mount.root().getPath()))
+        {
+            Iterator<Path> it = stream
+                .filter(MetaInfConfiguration::isTldFile)
```
joakime, line 723: Add another `.filter(Files::isRegularFile)` before this.
Otherwise this will match directories like `/META-INF/foo.tld`, which is odd.

joakime (changes requested):
Minor improvement to ignore directories.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
lorban: That test should be part of `isTldFile()`, shouldn't it?

jetty-ee9/jetty-ee9-webapp/src/main/java/org/eclipse/jetty/ee9/webapp/MetaInfConfiguration.java
lorban: That test should be part of `isTldFile()`, shouldn't it?

joakime: Typically no, if you stream filter by `Files:isRegularFile` then the other filters have less to check.
The `isTldFile` predicate only checks the name, not if it's a directory or file.

lorban: If you make the `Files.isRegularFile` check at the top of the `isTldFile` method, you pretty much have the same amount of logic to execute than if you have the check as an extra filter.

Plus it feels like `isTldFile` logically should check that the Path points to a file and not a folder.

joakime (approved):

joakime:
There's some test failures here.

```
java.nio.file.ClosedFileSystemException
	at jdk.zipfs/jdk.nio.zipfs.ZipFileSystem.ensureOpen(ZipFileSystem.java:1666)
	at jdk.zipfs/jdk.nio.zipfs.ZipFileSystem.newInputStream(ZipFileSystem.java:858)
	at jdk.zipfs/jdk.nio.zipfs.ZipPath.newInputStream(ZipPath.java:749)
	at jdk.zipfs/jdk.nio.zipfs.ZipFileSystemProvider.newInputStream(ZipFileSystemProvider.java:261)
	at java.base/java.nio.file.Files.newInputStream(Files.java:160)
	at org.eclipse.jetty.ee10.webapp.Descriptor.parse(Descriptor.java:42)
```

I believe I have this fixed in the immutable `ResourceCollection` PR #8337, lets wait for that to merge and update this PR from those changes.

lorban:
@joakime the failures are caused by the incorrect use of mounting in `getTlds`: it's not supposed to close the mounted jar upon return, but it should add it to its `_mountedResources` list so that unmounting occurs when the deconfig happens.

joakime (approved):
:+1:

PR #8347, opened 2022-07-27
Immutable ResourceCollection

gregw:
This is an alternative to #8337 that does no more than make ResourceCollection immutable.

### DIFF
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
@@ -35,6 +35,7 @@ import java.nio.file.StandardOpenOption;
 import java.nio.file.attribute.FileTime;
 import java.text.DateFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Base64;
 import java.util.Collection;
 import java.util.Collections;
@@ -1232,5 +1233,83 @@ public abstract class Resource implements ResourceFactory
          * @throws IOException Problem accessing resource
          */
         Resource root() throws IOException;
+    }
+
+    /**
+     * <p>Make a Resource containing a collection of other resources</p>
+     * @param resources multiple resources to combine as a single resource. Typically, they are directories.
+     * @return A Resource of multiple resources.
+     * @see ResourceCollection
+     */
+    public static ResourceCollection of(List<Resource> resources)
+    {
+        if (resources == null)
+            throw new IllegalArgumentException("No resources");
+
+        return new ResourceCollection(resources);
+    }
+
+    /**
+     * <p>Make a Resource containing a collection of other resources</p>
+     * @param resources multiple resources to combine as a single resource. Typically, they are directories.
+     * @return A Resource of multiple resources.
+     * @see ResourceCollection
+     */
+    public static ResourceCollection of(Resource... resources)
+    {
+        if (resources == null)
+            throw new IllegalArgumentException("No resources");
+
+        return new ResourceCollection(Arrays.asList(resources));
+    }
+
+    /**
+     * <p>Make a Resource containing a collection of other resources</p>
+     * @param resourceAsStrings multiple resources to combine as a single resource. Typically, they are directories.
+     * @return A Resource of multiple resources or null if no resources passed.
+     * @see ResourceCollection
+     * @throws IllegalArgumentException if a string cannot be converted to resource
+     */
+    public static ResourceCollection of(String... resourceAsStrings)
+    {
+        if (resourceAsStrings == null || resourceAsStrings.length == 0)
+            throw new IllegalArgumentException("No resources");
+
+        List<Resource> resources = new ArrayList<>();
+        for (String r : resourceAsStrings)
+        {
+            if (r == null)
+                continue;
+            r = r.trim();
+            if (StringUtil.isBlank(r))
+                continue;
+            try
+            {
+                resources.add(Resource.newResource(r));
+            }
+            catch (IOException e)
+            {
+                throw new IllegalArgumentException(e);
+            }
+        }
+        if (resources.size() == 0)
+            throw new IllegalArgumentException("No resources");
+
+        return new ResourceCollection(resources);
+    }
+
+    /**
+     * <p>Make a Resource containing a collection of other resources</p>
+     * @param csvResources multiple resources to combine as a single resource. Typically, they are directories.
+     * @return A Resource of multiple resources or null if no resources passed.
+     * @see ResourceCollection
+     * @throws IllegalArgumentException if a string cannot be converted to resource
+     */
+    public static ResourceCollection of(String csvResources)
+    {
+        if (StringUtil.isBlank(csvResources))
+            throw new IllegalArgumentException("No resources");
+
+        return of(StringUtil.csvSplit(csvResources));
     }
 }
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
@@ -21,14 +21,12 @@ import java.net.URI;
 import java.nio.channels.ReadableByteChannel;
 import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 
-import org.eclipse.jetty.util.IO;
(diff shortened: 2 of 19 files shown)


gregw:
OK - I've been a little bad and added a slightly unrelated commit.  I made the FileSystemPool dumpable and made the server dump include it, so we can monitor any leakage that way.
I've also added a test for sweep....
Actually, I'll cherry pick that directly to the main branch, so it should not be seen here soon.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -1233,4 +1234,82 @@ public interface Mount extends Closeable
...
+    }
+
+    /**
+     * <p>Make a Resource containing a collection of other resources</p>
+     * @param resourceAsStrings multiple resources to combine as a single resource. Typically, they are directories.
+     * @return A Resource of multiple resources or null if no resources passed.
+     * @see ResourceCollection
+     * @throws IllegalArgumentException if a string cannot be converted to resource
+     */
+    public static ResourceCollection of(String... resourceAsStrings)
```
joakime, line 1273: This needs to support glob, as it's a feature of WebAppContext.setExtraClasspath.

See the fromList code.

joakime, line 1273: This needs to support glob, and ResourceFactory, as it's a feature of WebAppContext.setExtraClasspath behaviors.

See the fromList code.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -42,185 +40,41 @@ public class ResourceCollection extends Resource
...
         for (Resource r : resources)
         {
-            if (r == null)
-            {
-                continue;
-            }
             if (r instanceof ResourceCollection)
             {
-                _resources.addAll(((ResourceCollection)r).getResources());
+                tmp.addAll(((ResourceCollection)r).getResources());
```
joakime, line 57: This accidentally adds all without a unique check like below.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -42,185 +40,41 @@ public class ResourceCollection extends Resource
...
             }
             else
             {
                 assertResourceValid(r);
```
joakime, line 61: No longer needed here.
Inline it.

jetty-ee10/jetty-ee10-maven-plugin/src/main/java/org/eclipse/jetty/ee10/maven/plugin/MavenWebAppContext.java
```diff
@@ -228,7 +228,7 @@ public void setResourceBases(String[] resourceBases)
                 resources.add(r);
             }
         }
-        setBaseResource(new ResourceCollection(resources.toArray(new String[resources.size()])));
+        setBaseResource(Resource.of(resources.toArray(new String[resources.size()])));
```
joakime, line 231: What?
`resources` is a List already.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -1233,4 +1234,82 @@ public interface Mount extends Closeable
...
+        for (String r : resourceAsStrings)
+        {
+            if (r == null)
+                continue;
+            r = r.trim();
+            if (StringUtil.isBlank(r))
+                continue;
+            try
+            {
+                resources.add(Resource.newResource(r));
```
joakime, line 1288: If the string is a URI reference `jar:file://foo.jar!/` from a user configuration, this won't work, as it hasn't been mounted yet.

joakime (changes requested):
Cleaner, but still has things to do.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -1233,4 +1234,82 @@ public interface Mount extends Closeable
...
+     * @return A Resource of multiple resources or null if no resources passed.
+     * @see ResourceCollection
+     * @throws IllegalArgumentException if a string cannot be converted to resource
+     */
+    public static ResourceCollection of(String csvResources)
+    {
+        if (StringUtil.isBlank(csvResources))
+            throw new IllegalArgumentException("No resources");
+
+        return of(StringUtil.csvSplit(csvResources));
```
joakime, line 1313: This doesn't support the old behavior of allowing `,` or `|` as delimiters in the cases where the filename has a comma of its own.

This is what fromList does.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime, line 1273: Input here can be path references or uri string references.
Path references can be absolute, or relative.
Path references can also be presented in windows or unix formats.
Path references from classpath/classloader style configurations can even include globs like `C:\other\libs\*` or `/opt/web/libs/*`, which would need to be expanded at some point.   Going through `Resource.newResource(String)` for globs is tricky, as you cannot create a glob File or Path object on Windows, you have to work with it as a String or a URI first, then expand from there.

It would make more sense to convert them all to URIs without the Resource layer involved first, then you know if there's a glob involved, and if there's mounting needing to be done.

gregw: I'm just going to keep using the existing `fromList` for now.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
gregw: Ah yes.  So I think the solution is to not have any String constructors to force the caller to do the resource conversion, and thus do the mounts if needed.
Working on it....

joakime: I have this fixed in the original ResourceCollection PR, which has taken much of your changes in this PR
it's smaller than it was before you created this PR.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime: I eliminated fromList in the original ResourceCollection PR as it had a chicken/egg issue.
It's been replaced with `List<URI> Resource.split(String)` in the original ResourceCollection PR.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
joakime: This is now fixed in the original ResourceCollection PR.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime: This API `Resource.of(String ... resourcesAsStrings)` was eliminated entirely in the original ResourceCollection PR.

The original ResourceCollection PR has many more tests around this kind of user provided configuration as well.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
joakime: This is also fixed now in the original ResourceCollection PR

jetty-ee10/jetty-ee10-maven-plugin/src/main/java/org/eclipse/jetty/ee10/maven/plugin/MavenWebAppContext.java
joakime: This is also fixed in the original ResourceCollection PR now.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime: These concerns (about going thru Resource.newResource() during glob calculation has been addressed in the original ResourceCollection PR

gregw:
closing for now.... working back in  #8337

PR #8350, opened 2022-07-28
Generalize the immutable ResourceCollection

gregw:
This is a generalization of the latest #8337.  Rather than have a special `ResourceCollection.Mount` and specific mount handling in `ContextHandler`, this approach introduces a `Mounts` collection that can be stored on any `Container`.  I am now using this in most places where a string is converted to a Resource, and can produce a log warning if a mount is leaked (in the places I have not yet updated).

However, I am now thinking that this can be further generalized, as even creating a Resource from a URI may need a mount.

So this is step 1. Standby for another PR with step 2.


joakime:
Closing in favor of replacement PR

* #8352

PR #8351, opened 2022-07-28
Embrance the ResourceFactory

gregw:
This is a  alternative to #8350, where the ResourceFactory has been embraced as the main source of resources and it is the factory lifecycle that will determine when mounts are closed.
I think this approach can be extended, so that by make a `ResourceFactory.ofJar(Resource jar)`, we can actually entirely hide mounts from users.

I'm working on that now... either for this PR or perhaps a redux3 branch.


joakime:
Closing in favor of replacement PR

* #8352

PR #8352, opened 2022-07-28
Jetty 12 : `ResourceFactory` redux 3

gregw:
This is a WIP alternative to both #8350 and #8351

This extends the embrace of the ResourceFactory concept to try to entirely hide the concept of a Mount from most users.


joakime:
I like the idea, and the syntax, the naming could use some help though.

Looks like it would be used like this ...

``` java
// Redux 3 using String
ResourceFactory factory = ResourceFactory.of(container);
Resource resource = factory.newResource(String);

// Redux 3 using URI
ResourceFactory factory = ResourceFactory.of(container);
Resource resource = factory.newResource(URI);

// Redux 3 using List<URI>
ResourceFactory factory = ResourceFactory.of(container);
Resource resource = factory.newResource(uriList); // don't care if it's a ResourceCollection
// in fact, if the list is only 1 entry, it could return a Resource.
```

And Embedded Jetty users would use it as ...

``` java
// Embedded Point of view 1
WebAppContext context = new WebAppContext();
ResourceFactory resourceFactory = ResourceFactory.of(context);
Resource base = resourceFactory.newResource(uriToJarFileLoc);
context.setBaseResource(base);
```

But some contexts would already have a ResourceFactory as mandatory, like `WebAppContext`, so why not expose it?

``` java
// Embedded Point of view 2
WebAppContext context = new WebAppContext();
Resource base = context.getResourceFactory().newResource(uriToJarFileLoc);
context.setBaseResource(base);
```

gregw:
> But some contexts would already have a ResourceFactory as mandatory, like `WebAppContext`, so why not expose it?
> 
> ```java
> // Embedded Point of view 2
> WebAppContext context = new WebAppContext();
> Resource base = context.getResourceFactory().newResource(uriToJarFileLoc);
> context.setBaseResource(base);
> ```


Ideally the `ResourceFactory.of(context)` API would return any factory that the context already has to have. So long as it is added as a Bean.   Hmmm perhaps the `of` method should not look for a specific type? 

So either style should end up identical.

lorban:
Superseded by https://github.com/eclipse/jetty.project/pull/8409

PR #8355, opened 2022-07-29
Jetty 12 : Remove `java.io.File` usage from `jetty-start`

joakime:
* This corrects our URL/URI/URLClassLoader behaviors
  to always produce to-spec URIs for `file://` references.


lorban:
Getting rid of all our `java.io.File` usages in favor of `java.nio.Path` is a very desirable change, so it gets a major :+1:.

But that shouldn't be used to workaround our `file:` vs `file://` handling issue in `FileSystemPool`.

lorban (approved):

PR #8359, opened 2022-07-29
Jetty 12 : Move `FileID` from `jetty-deploy` to `jetty-util`

joakime:
+ Move static methods from other places to `FileID` or `URIUtil`
  - `MultiReleaseJarFile`
  - `Resource`
  - `MetaInfConfiguration`
+ Improve testing of `URIUtil` and `FileID`


jetty-core/jetty-deploy/src/test/java/org/eclipse/jetty/deploy/MockAppProvider.java
```diff
@@ -60,9 +60,9 @@ public ContextHandler createContextHandler(App app) throws Exception
 
         String name = app.getPath().toString();
         name = name.substring(name.lastIndexOf("-")  + 1);
-        File war = new File(webappsDir, name);
+        Path war = new File(webappsDir, name).toPath();
```
lorban: Why don't you change the `File webappsDir` field instead?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/FileID.java
```diff
@@ -0,0 +1,380 @@
...
+{
+    /**
+     * Does the provided path have a directory segment with
+     * the configured name.
+     *
+     * @param path the path to search
+     * @param directoryName the directory name to look for (case insensitive lookup)
+     * @return true if the directory name exists in path, false if otherwise
+     */
+    public static boolean containsDirectory(Path path, String directoryName)
```
lorban: This implementation is broken when `path` is `/`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/FileID.java
```diff
@@ -0,0 +1,380 @@
...
+    /**
+     * Retrieve the basename of a path. This is the name of the
+     * last segment of the path, with any dot suffix (e.g. ".war") removed
+     *
+     * @param path The string path
+     * @return The last segment of the path without any dot suffix
+     */
+    public static String getBasename(Path path)
+    {
+        String basename = path.getFileName().toString();
```
lorban: You should test for `path.getFileName()` returning `null`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/FileID.java
```diff
@@ -0,0 +1,380 @@
...
+     */
+    public static String getExtension(Path path)
+    {
+        if (path == null)
+            return null; // no path
+
+        if (!Files.isRegularFile(path))
+            return null; // not a file
+
+        return getExtension(path.getFileName().toString());
```
lorban, line 64: You should test for `path.getFileName()` returning `null`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/FileID.java
```diff
@@ -0,0 +1,380 @@
...
+     * Predicate to skip {@code META-INF/versions/*} tree from walk/stream results.
+     *
+     * <p>
+     * This only works with a zipfs based FileSystem
+     * </p>
+     *
+     * @param path the path to test
+     * @return true if not in {@code META-INF/versions/*} tree
+     */
+    public static boolean skipMetaInfVersions(Path path)
```
lorban: Rename to `isNotMetaInfVersions`?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/FileID.java
```diff
@@ -0,0 +1,380 @@
...
+     * Predicate to skip {@code module-info.class} files.
+     *
+     * <p>
+     * This is a simple test against the last path segment using {@link Path#getFileName()}
+     * </p>
+     *
+     * @param path the path to test
+     * @return true if not a {@code module-info.class} file
+     */
+    public static boolean skipModuleInfoClass(Path path)
```
lorban: rename to `isNotModuleInfoClass`?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -1647,25 +1655,7 @@ public static String addQueries(String query1, String query2)
...
-        // Get SSP (retaining encoded form)
-        String s = uri.getRawSchemeSpecificPart();
-        int bangSlash = s.indexOf("!/");
-        if (bangSlash >= 0)
-            s = s.substring(0, bangSlash);
-        return URI.create(s);
-    }
-
-    public static URI fixBadJavaIoFileUrl(URI uri)
+    public static URI correctBadFileURI(URI uri)
```
lorban: Rename to `correctFileURI`?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -852,6 +719,7 @@ public Resource resolve(String subUriPath) throws IOException
         return newResource(resolvedUri);
     }
 
+    // TODO: move to URIUtil
     private static URI uriResolve(URI uri, String subUriPath) throws IOException
```
lorban: Move to `URIUtil` maybe? :-)

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationParser.java
```diff
@@ -917,7 +917,7 @@ public void resetParsedClasses()
      *
      * @param name the class file name
      * @return whether the class file name is valid
-     * TODO: does multiple things, we should be able to move this FileID in future PR
+     * TODO: does multiple things, we should be able to eliminate this in future PR by consolidating the ZipFS and Directory scanning behaviors into a single Files.walk() stream
```
lorban, line 920: Do you already have the shape of that future PR in your head?

lorban (changes requested):
There are a couple small bugs with `FileID` helpers' handling of `/` when `Path.getFilename()` is called, otherwise this looks good.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/FileID.java
joakime: This case is protected by the `path.getNameCount()` which returns 0, causing the for loop statements to not run.
I'll add a testcase for this anyway.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/FileID.java
joakime: Adding testcase

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/FileID.java
joakime: Adding testcase

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime: I'm going to get rid of this, as it doesn't make much sense anymore.

jetty-core/jetty-deploy/src/test/java/org/eclipse/jetty/deploy/MockAppProvider.java
joakime: Done.

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationParser.java
joakime: Yes.

I want to consolidate the WAR/JAR file scanning/walking with the Directory scanning/walking.
AFAICT, the only difference between the two is that the Directory walking technique also does a "is hidden file" check, which we have a static predicate for in `FileID`, and that same check in WAR walk is harmless.   That means the AnnotationParser gets even simpler.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/FileID.java
lorban: Wouldn't `containsDirectory(Path.get("/"), "tmp")` return false even if `/tmp` exists?

lorban (changes requested):
LGTM, besides one last tiny bug.

lorban (approved):

jetty-ee10/jetty-ee10-tests/jetty-ee10-test-quickstart/src/test/resources/jetty-logging.properties
```diff
@@ -1,3 +1,5 @@
 # Jetty Logging using jetty-slf4j-impl
 org.eclipse.jetty.LEVEL=INFO
+org.eclipse.jetty.xml.LEVEL=DEBUG
+org.eclipse.jetty.quickstart.LEVEL=DEBUG
```
joakime: This file was restored/rollback'd in later commit fdbeb2af7f407975b1143df08d2a5e0fbbb29c93

lorban (approved):

joakime:
Merged into `jetty-12.0.x`

PR #8360, opened 2022-07-31
Fixes #8007 - Support Loom.

### LINKED ISSUE #8007: Support Loom

sbordet:
Alternative implementation of support for virtual threads for HTTP/1.1, HTTP/2 and HTTP/3.

The virtual thread support is in AdaptiveExecutionStrategy.
When virtual threads are supported and enabled, reserved threads are disabled and
blocking tasks are run in a virtual thread instead that being executed by the Executor.

### DIFF
--- a/jetty-quic/quic-server/src/main/java/org/eclipse/jetty/quic/server/QuicServerConnector.java
+++ b/jetty-quic/quic-server/src/main/java/org/eclipse/jetty/quic/server/QuicServerConnector.java
@@ -40,7 +40,6 @@ import org.eclipse.jetty.quic.quiche.QuicheConfig;
 import org.eclipse.jetty.quic.quiche.SSLKeyPair;
 import org.eclipse.jetty.server.AbstractNetworkConnector;
 import org.eclipse.jetty.server.ConnectionFactory;
-import org.eclipse.jetty.server.Connector;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.util.IO;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
@@ -55,7 +54,7 @@ import org.eclipse.jetty.util.thread.Scheduler;
  *
  * @see QuicConfiguration
  */
-public class QuicServerConnector extends AbstractNetworkConnector implements Connector.VirtualThreadsConfigurable
+public class QuicServerConnector extends AbstractNetworkConnector
 {
     private final QuicConfiguration quicConfiguration = new QuicConfiguration();
     private final QuicSessionContainer container = new QuicSessionContainer();
@@ -69,7 +68,6 @@ public class QuicServerConnector extends AbstractNetworkConnector implements Con
     private int outputBufferSize = 2048;
     private boolean useInputDirectByteBuffers = true;
     private boolean useOutputDirectByteBuffers = true;
-    private boolean invokeApplicationWithVirtualThreads;
 
     public QuicServerConnector(Server server, SslContextFactory.Server sslContextFactory, ConnectionFactory... factories)
     {
@@ -143,25 +141,6 @@ public class QuicServerConnector extends AbstractNetworkConnector implements Con
     public void setUseOutputDirectByteBuffers(boolean useOutputDirectByteBuffers)
     {
         this.useOutputDirectByteBuffers = useOutputDirectByteBuffers;
-    }
-
-    @Override
-    public boolean isInvokeApplicationWithVirtualThreads()
-    {
-        return invokeApplicationWithVirtualThreads;
-    }
-
-    @Override
-    public void setInvokeApplicationWithVirtualThreads(boolean invokeApplicationWithVirtualThreads)
-    {
-        try
-        {
-            VirtualThreadsConfigurable.super.setInvokeApplicationWithVirtualThreads(invokeApplicationWithVirtualThreads);
-            this.invokeApplicationWithVirtualThreads = invokeApplicationWithVirtualThreads;
-        }
-        catch (UnsupportedOperationException ignored)
-        {
-        }
     }
 
     @Override
@@ -325,7 +304,7 @@ public class QuicServerConnector extends AbstractNetworkConnector implements Con
         @Override
         public Connection newConnection(SelectableChannel channel, EndPoint endpoint, Object attachment)
         {
-            ServerQuicConnection connection = new ServerQuicConnection(QuicServerConnector.this, endpoint, isInvokeApplicationWithVirtualThreads());
+            ServerQuicConnection connection = new ServerQuicConnection(QuicServerConnector.this, endpoint);
             connection.addEventListener(container);
             connection.setInputBufferSize(getInputBufferSize());
             connection.setOutputBufferSize(getOutputBufferSize());
--- a/jetty-quic/quic-server/src/main/java/org/eclipse/jetty/quic/server/QuicServerConnector.java
+++ b/jetty-quic/quic-server/src/main/java/org/eclipse/jetty/quic/server/QuicServerConnector.java
@@ -40,6 +40,7 @@ import org.eclipse.jetty.quic.quiche.QuicheConfig;
 import org.eclipse.jetty.quic.quiche.SSLKeyPair;
 import org.eclipse.jetty.server.AbstractNetworkConnector;
 import org.eclipse.jetty.server.ConnectionFactory;
+import org.eclipse.jetty.server.Connector;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.util.IO;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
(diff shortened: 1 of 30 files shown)


sbordet:
@gregw @lorban please review this alternative support for Loom, more in line with what suggested by @gregw.

It's not complete as we need to verify that all code paths that call application code are actually invoked with a virtual thread (so more testing needed), but it's functional enough to deserve a review to see if it is the right direction before doing more work.

joakime:
Keep in mind that the prior Loom support PR #8035 still exists and is still open

* #8035

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -134,20 +135,27 @@ public class AdaptiveExecutionStrategy extends ContainerLifeCycle implements Exe
...
     private boolean _pending;
 
     /**
-     * @param producer The produce of tasks to be consumed.
+     * @param producer The producer of tasks to be consumed.
      * @param executor The executor to be used for executing producers or consumers, depending on the sub-strategy.
      */
     public AdaptiveExecutionStrategy(Producer producer, Executor executor)
+    {
+        this(producer, executor, false);
```
gregw: I guess we could have a JVM wide system property (cough splutter gasp!) to configure this default?

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -134,20 +135,27 @@ public class AdaptiveExecutionStrategy extends ContainerLifeCycle implements Exe
...
+    }
+
+    public AdaptiveExecutionStrategy(Producer producer, Executor executor, boolean useVirtualThreads)
     {
         _producer = producer;
         _executor = executor;
-        _tryExecutor = TryExecutor.asTryExecutor(executor);
+        _useVirtualThreads = useVirtualThreads;
+        _tryExecutor = useVirtualThreads ? TryExecutor.NO_TRY : TryExecutor.asTryExecutor(executor);
```
gregw: yeah... but no!

If you don't want to use reserved threads, then configure 0 reserved threads.   If reserved threads are configured in the thread pool, then they should be used if available.

No need to do anything special here as the user already can configure a NO_TRY tryExecutor if they want it.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -462,7 +470,10 @@ private void execute(Runnable task)
     {
         try
         {
-            _executor.execute(task);
+            if (_useVirtualThreads)
+                VirtualThreads.startVirtualThread(task);
+            else
+                _executor.execute(task);
```
gregw: Wow, that's pretty simple...
Makes me wonder if we could do it all with just a change to the ThreadPool?   Hmmm but then we'd have virtual threads being selectors.... so nah!

jetty-util/src/main/java/org/eclipse/jetty/util/VirtualThreads.java
```diff
@@ -0,0 +1,97 @@
...
+//
+
+package org.eclipse.jetty.util;
+
+import java.lang.reflect.Method;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class VirtualThreads
```
gregw, line 30: javadoc for this class and its methods.

jetty-quic/quic-server/src/main/java/org/eclipse/jetty/quic/server/QuicServerConnector.java
```diff
@@ -143,6 +145,25 @@ public void setUseOutputDirectByteBuffers(boolean useOutputDirectByteBuffers)
         this.useOutputDirectByteBuffers = useOutputDirectByteBuffers;
     }
 
+    @Override
+    public boolean isInvokeApplicationWithVirtualThreads()
```
gregw: wrong name

jetty-server/src/main/java/org/eclipse/jetty/server/Connector.java
```diff
@@ -97,4 +98,21 @@ public interface Connector extends LifeCycle, Container, Graceful
      * @return The connector name or null.
      */
     public String getName();
+
+    public interface VirtualThreadsConfigurable
+    {
+        public default boolean isInvokeApplicationWithVirtualThreads()
```
gregw: This is not the right name.
Nor am I convinced if it should be a connection configuration.

gregw (changes requested):
This is looking much better. 
I've made a couple of comments / niggles.

But I also don't like the way `useVirtualThreads` boolean is passed around.  

Rather than configure connectors,  I'm wondering if it could be a filed/attribute/property associated with the thread pool?  The strategy could then look at the threadpool it has been given and work out how to act from that.

jetty-util/src/main/java/org/eclipse/jetty/util/VirtualThreads.java
```diff
@@ -0,0 +1,161 @@
...
+        return false;
+    }
+
+    /**
+     * <p>Implementations of this interface can be configured to use virtual threads.</p>
+     * <p>Whether virtual threads are actually used depends on whether the runtime
+     * supports virtual threads and, if the runtime supports them, whether they are
+     * configured to be used via {@link #setUseVirtualThreads(boolean)}.</p>
+     */
+    public interface Configurable
```
gregw, line 133: good name!

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -134,19 +135,21 @@ public class AdaptiveExecutionStrategy extends ContainerLifeCycle implements Exe
...
     /**
-     * @param producer The produce of tasks to be consumed.
+     * @param producer The producer of tasks to be consumed.
      * @param executor The executor to be used for executing producers or consumers, depending on the sub-strategy.
      */
     public AdaptiveExecutionStrategy(Producer producer, Executor executor)
     {
         _producer = producer;
         _executor = executor;
+        _useVirtualThreads = VirtualThreads.isUseVirtualThreads(executor);
```
gregw: Is grabbing this at construction too early?
I know it is nice to have as a final, but I think it is probably more correct to look at the value in `doStart` as it might be set on the executor after the strategy has been built (but not started).

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -462,7 +465,10 @@ private void execute(Runnable task)
     {
         try
         {
-            _executor.execute(task);
+            if (isUseVirtualThreads())
```
gregw, line 474: Should we make this method protected so that if somebody wants to be more exotic in their logic they can extend?

gregw (changes requested):
This looks pretty good.... but other than the niggles I've commented on, you also need to modify `jetty-threadpool.xml` and module to have a property for useVirtualThreads.

Or if you like, you could have a `loom.mod` that depended on thread pool and set the property.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
sbordet: Good point, done.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
sbordet: It's already `public`.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -462,7 +471,10 @@ private void execute(Runnable task)
...
+                VirtualThreads.startVirtualThread(task);
+            else
+                _executor.execute(task);
```
gregw, line 477: Should have a test in `AdaptiveExecutionStrategyTest` for this.  It should start a billion threads and be OK :)

gregw (approved):
LGTM other than a specific test for the execution strategy would be good to have.

lorban (approved):
LGTM

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
sbordet: Billions are a thing of the past. I'll use trillions, because that's what's needed these days with 0.75 CPUs in the cloud.

PR #8399, opened 2022-08-02
fix exclusion of groups stress and slow

olamy:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2325,7 +2325,6 @@
               <groupId>org.apache.maven.plugins</groupId>
               <artifactId>maven-surefire-plugin</artifactId>
               <configuration>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -2325,7 +2325,7 @@
               <groupId>org.apache.maven.plugins</groupId>
               <artifactId>maven-surefire-plugin</artifactId>
               <configuration>
-                <excludedGroups>external, large-disk-resource</excludedGroups>
+                <excludedGroups>external, large-disk-resource, stress, slow</excludedGroups>
```
joakime: Does nightly use the `ci` profile?
If so, this is not the right place for this.

joakime (commented):
Just a question

joakime (approved):
This looks promising.
:+1:

PR #8403, opened 2022-08-02
Jetty 12 : Remove `maxBuffer` from `ContentFactory.getContent(String)` parameters

joakime:
Per the TODO in `ContentFactory.getContent()` remove the `maxBuffer` argument

https://github.com/eclipse/jetty.project/blob/74a11a153f91b51b3c54efab3e5adf26aac0dc4d/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpContent.java#L74-L81


lorban (approved):

joakime:
Since this was a TODO added by @gregw lets have him review this change as well.

It started with removing the argument in `HttpContent` and the rest is a ripple effect of removing that argument.
Should let @gregw take a look at the whole scope of change for this.

joakime:
@sbordet this is the PR I mentioned.

sbordet (approved):

PR #8404, opened 2022-08-02
Jetty 12 : Simplify `AnnotationParser` scanning for class files

joakime:
+ Not using MultiReleaseJarFile
+ Walk of a Directory or a JAR is the same
+ Reduced parse method count
+ Fixed TestAnnotationParser
  Using proper fully qualified classnames now (they were not yet converted over to jetty-12 naming)


jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationParser.java
```diff
@@ -98,7 +90,7 @@ private static int asmVersion()
             .map(Integer::parseInt)
             .max(Integer::compareTo);
 
-        if (!asmVersion.isPresent())
+        if (asmVersion.isEmpty())
```
janbartel, line 94: I say potayto you say potahto.

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationParser.java
```diff
@@ -717,92 +559,66 @@ public void parse(final Set<? extends Handler> handlers, Resource r) throws Exce
...
-            {
-                scanClass(handlers, null, is);
-                return;
-            }
+            parseClass(handlers, null, r.getPath());
         }
 
         if (LOG.isDebugEnabled())
-            LOG.warn("Resource not scannable for classes: {}", r);
+            LOG.warn("Resource not able to be scanned for classes: {}", r);
```
janbartel, line 571: ????

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationParser.java
```diff
@@ -717,92 +559,66 @@ public void parse(final Set<? extends Handler> handlers, Resource r) throws Exce
...
      * Parse all classes in a directory
      *
      * @param handlers the set of handlers to look for classes in
-     * @param root the resource directory to look for classes
+     * @param baseResource the resource representing the baseResource being scanned (jar, dir, etc)
+     * @param dir the directory root to start the scanning of classes for
      * @throws Exception if unable to parse
      */
-    protected void parseDir(Set<? extends Handler> handlers, Resource root) throws Exception
+    private void parseDir(Set<? extends Handler> handlers, Resource baseResource, Path dir) throws Exception
```
janbartel: Why is there both `baseResource` and `Path`. Surely they are one and the same thing?

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationParser.java
```diff
@@ -717,92 +559,66 @@ public void parse(final Set<? extends Handler> handlers, Resource r) throws Exce
...
 
-        if (fullname.endsWith(".class"))
+        if (FileID.isClassFile(r.getPath()))
         {
-            try (InputStream is = Files.newInputStream(r.getPath()))
-            {
-                scanClass(handlers, null, is);
-                return;
-            }
+            parseClass(handlers, null, r.getPath());
```
janbartel, line 567: This should be `parseClassFile`, otherwise you're not recording the parsed  classes.

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationParser.java
```diff
@@ -812,173 +628,55 @@ protected void parseDir(Set<? extends Handler> handlers, Resource root) throws E
...
-                    LOG.debug("Scanning class from jar {}", location);
-                LOG.debug("Scanning class from jar {}", location);
-                scanClass(handlers, jar, is);
-            }
+            Path root = jarMount.root().getPath();
+            parseDir(handlers, jarResource, root);
         }
     }
 
+    private void parseClassFile(Set<? extends Handler> handlers, Resource jar, Path root, Path entry) throws IOException
```
janbartel: I don't think there should be both `parseClassFile` and `parseClass` methods, should just be one method.

janbartel (changes requested):

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationParser.java
joakime: This can be different in the case of a JAR.
The `baseResource`, is the JAR itself (the File)
The `dir` is the Directory (as a Path) being scanned, which is the root of the archive.

Ideally, I want to eliminate the tracking of the Resource, as it's a needless bit of trivia in the AnnotationParser IMO.

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationParser.java
joakime: Just reads better

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationParser.java
joakime: It's the same end result, but easier to read.

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationParser.java
joakime: There's a small difference.

The one with `Path root, Path entry` allows the "shortname" to be calculated properly.
Which takes something like ...

```
 root = /path/to/scan/
entry = /path/to/scan/org/eclipse/jetty/demo/Foo.class
```

And is able to create a proper shortname of `org.eclipse.jetty.demo.Foo`, which is used in the `_parsedClassNames` Map.
However, the raw/naked scan of a Path, without a root (something we used to support) difficult to generate a shortName from.

See Jetty 10 -> https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-annotations/src/main/java/org/eclipse/jetty/annotations/AnnotationParser.java#L741-L748

This code path cannot ever generate a proper classname.
All we have is a path, with no known cutoff for it.
Eg: `/path/to/scan/org/eclipse/jetty/demo/Foo.class`

So the old Jetty 10 behavior was to not track that class.
This old code path (raw class file parse) is the only reason for this difference in methods.
If you are ok with eliminating this code path in the one remaining public `parse` method, then we can simplify this difference in parseClass* methods down to 1 method.

Another approach is to not build the classname from the path string, but to take it from the ASM parse step, it knows the name.
Yeah, let me try that approach.

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationParser.java
joakime: Closing this comment, as it's covered by the one below.

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationParser.java
joakime: Turns out there is 1 reason for `Resource` being tracked as `baseResource`, and that's a single code block down in `org.eclipse.jetty.ee10.webapp.MetaData` in it's `addDiscoveredAnnotation(DiscoveredAnnotation annotation)` method.

https://github.com/eclipse/jetty.project/blob/jetty-12.0.x/jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaData.java#L360-L390

Oddly, if I intentionally break that code block with a null resource and run the build/test, the only tests that fail are 2 test methods in `org.eclipse.jetty.ee10.webapp.TestMetaData`.
(testAddDiscoveredAnnotation and testResolve).   No other tests fail?  Does this mean this requirement for tracking a Resource only exists to satisfy a testcase?

Something to investigate later, for now I'm leaving the tracking of Resource in place.

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationParser.java
joakime: Turns out using the ASM parse results for the classname is a better choice, much easier code.

joakime:
@janbartel this is read for a fresh review

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationParser.java
janbartel: No, the requirement for tracking the resource is used eg in Quickstart, where we set an attribute with the source (descriptor/annotation/programmatic) of each element that is generated in the quickstart-web.xml. Please keep it.

janbartel: Also, I still don't see why you need to pass in the `Path`, why can't you just ask the `baseResource` for it's `Path`? You're only ever passing in the root level of the jarfile, so it's not like this method is scanning some subdirectory of the `baseResource`, it's scanning the `baseResource` itself.

janbartel (changes requested):

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationParser.java
joakime: > No, the requirement for tracking the resource is used eg in Quickstart, where we set an attribute with the source (descriptor/annotation/programmatic) of each element that is generated in the quickstart-web.xml. Please keep it.

Can you point out where it's used?
I cannot find where the resource is ever used in quickstart, not even from a `.toString()`

As there is currently a difference between `baseResource` and `dir`, they point to different things when it's a JAR. (they point to the same thing if the resource is a directory. eg: `file://path/to/dir/`)

`baseResource` is the JAR itself, the File, aka `file://path/to/foo.jar`
`dir` is a directory, pointing to the root contents of the JAR, as in `jar:file://path/to/foo.jar!/`

This is an important distinction in Resource.
One is a File, no mount needed, it's treated like a file on disk, nothing more.
The other is a directory, with a mount being managed.

I need to know how quickstart uses this resource, to know if I should be passing in the `file://path/to/foo.jar` or `jar:file://path/to/foo.jar!/` reference.

joakime: Checking `jetty-10.0.x` and it's `test-quickstart` i see output like this ...

``` xml
  <context-param>
    <param-name>org.eclipse.jetty.resources</param-name>
    <param-value><![CDATA[
    "jar:${WAR.uri}/WEB-INF/lib/demo-web-fragment-10.0.12-SNAPSHOT.jar!/META-INF/resources"]]></param-value>
  </context-param>
```

Is this what you are referring to?

joakime: Looks like Quickstart generates optional the `origin="..."` attributes, that has references to resource.
These are informational, and nothing in our code uses those attributes.

Good, that means `parseDir()` can use just the Resource, even if it's `file://` or `jar:file://` without breaking quickstart.  

I'll make this change.

joakime: Change made in commit a5eeac1c786e1bb192da184cb9b4986e49eb730c

joakime:
@janbartel the changes you have asked for have been made.  Please review again.

jetty-ee10/jetty-ee10-annotations/src/test/java/org/eclipse/jetty/ee10/annotations/TestAnnotationInheritance.java
```diff
@@ -13,6 +13,12 @@
```
janbartel: I'm not sure this test has a lot of validity any more, as you removed the method that was walking the class hierarchy of a given class. All these tests are now copying in multiple classes to a directory, then passing that directory to the parser, so it's really now testing the `parseDir` method. Same comment for ee9.

janbartel (changes requested):
If you could fix up the TestAnnotationInheritance class, then we will be good to go.

jetty-ee10/jetty-ee10-annotations/src/test/java/org/eclipse/jetty/ee10/annotations/TestAnnotationInheritance.java
joakime: From commit: [Deleting TestAnnotationInheritance tests](https://github.com/eclipse/jetty.project/pull/8404/commits/f93e2268c3a2194fa10543f29697f69989e0a70e) 

+ We don't want the `AnnotationParser`
  creating new Resource objects for `Class.getSuperClass()` discovered classes
  resources that don't belong to the same `Resource` that the passed
  in class belongs to (as this is out of scope for the Mount management).

  Imagine if told to parse `ClassB`
  which extends from `ClassFoo`, which
  is in a different JAR than `ClassB`?

  This concept from Jetty 9/10/11 was only used in testcases and not
  in live runtime code, so it was  removed.

janbartel (approved):

PR #8406, opened 2022-08-02
Jetty 12 : XML namespace cleanup

joakime:
Adding Servlet 6 XML Entities to `WebDescriptor`
Updating all `jetty-ee10` based web XML to use new Servlet 6 namespace and XSDs


joakime:
@janbartel only 1 file was manually changed, that's the ee10 `WebDescriptor` (to add the Servlet 6.0 XML Entities, I also merged forward the recent changes in the ee9 WebDescriptor to ee10 first before this PR)
The rest of the changes, which are only XML namespace changes, are mechanical changes.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebDescriptor.java
```diff
@@ -162,6 +158,55 @@ void mapResources()
                         jsp30xsd = Loader.getResource("jakarta/servlet/jsp/resources/jsp_3_0.xsd");
                 }
 
+                // Servlet 5 Support
+                if (ServletContextHandler.ServletContextApi.SERVLET_MAJOR_VERSION >= 5)
```
janbartel: I don't understand the need for these if statements. Whenever a new version of the spec comes out, we just add the redirects for it.

jetty-ee10/jetty-ee10-demos/jetty-ee10-demo-spec/jetty-ee10-demo-spec-webapp/src/main/webapp/WEB-INF/web.xml
```diff
@@ -2,9 +2,9 @@
 <web-app 
    xmlns="https://jakarta.ee/xml/ns/jakartaee"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-   xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/web-app_5_0.xsd"
+   xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/web-app_6_0.xsd"
```
janbartel, line 5: If you're updating the version of the spec for the `demo-spec-webapp`, then please also update the `index.html` file with a heading of the newer version (currently it says **Servlet 5.0 Demo WebApp**)

janbartel (changes requested):

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebDescriptor.java
joakime: This is to keep the sections in line when it comes to ee9/ee8 usages (and the new tooling to translate things)
Don't want this list to be too different between the ee# versions.

In the case of the ee9 `WebDescriptor` it gets munged into ee8 version by tooling, but then attempts to load the resources that don't exist in the servlet-api.jar that ee8 sees.

Ideally, this ENTIRE section (registering the XML entities) can be a method in jetty-xml with a single arg ...

`org.eclipse.jetty.xml.XmlEntities.registerEEResources(int level)`

That way, it makes things consistent across the EE levels, and makes @olamy job with the ee8 effort easier, as he doesn't have to worry about the dozens of new exception for the `jakarta` to `javax` translation.

jetty-ee10/jetty-ee10-demos/jetty-ee10-demo-spec/jetty-ee10-demo-spec-webapp/src/main/webapp/WEB-INF/web.xml
joakime: Good catch.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebDescriptor.java
joakime: There was an API change in Resource to not throw IOException in many of the APIs.

It didn't make sense, and the overwhelming majority of the methods indicated as throwing IOException actually didn't throw them.

Turns out that 1 change revealed a ton of bugs that were being hidden by spurious `catch(IOException)` that were catching too much.

One issue that was revealed was that the `jakarta.servlet:jakarta-servlet-api:<ver>` artifacts didn't have the same list of resources that our `jetty-servlet-api` jars had.
But those errors were being consumed and ignored resulting in invalid states during Xml Parsing later.

Those missing XML Entities were added in commit e848dd08522b186f94ffe8c37397cc07acb3b3b1

The addition of the `Loader.getRequiredResource(String)` in that commit also revealed that the EE8 translation layer was modifying that list of XML Entities!
Both the path locations in `servlet-api-<ver>.jar` and the URLs were being modified by that EE8 tooling.
Had a talk with @olamy about it, and he's been adding exclusions left and right.
But that also revealed that various XML files themselves coming from ee9 have to be downgraded to what exists in ee8 as well (he's working on that too, it's a new requirement for that ee8 tooling)

joakime:
Changes made

joakime:
@janbartel this is ready for a fresh review

joakime:
@janbartel I went ahead and standardized the XML Entity registration process in commit https://github.com/eclipse/jetty.project/pull/8406/commits/a40c7bea5c79d55d5742858d2fe5bbf94739a7f2

Now we have a consistent XML Entity registration process across all EE levels.
This also fixed a bug in the EE9 (and EE8) levels, where the dtds for web-app 2.2 and 2.3 were not actually being registered (the resource was looked up, but not mapped to a system id or public id in those two EE levels)

Bonus side effect of this change is that the EE8 tooling now won't mess up the XML Entity registration process (by mangling the strings) anymore.
Hey @olamy look! --^

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebDescriptor.java
```diff
@@ -87,156 +88,10 @@ protected InputSource resolveEntity(String pid, String sid)
             {
                 if (!mapped)
```
janbartel: I'd rather see a solution based on the `Environment`. A `Context` knows what environment it is in - "ee8", "ee9", "ee10" etc. The `WebDescriptor` can have a simple static list of names of resources that match each environment, and do the `registerRedirect` as appropriate.

janbartel (changes requested):

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebDescriptor.java
joakime: I'll give that a whirl

jetty-core/jetty-xml/src/main/resources/org/eclipse/jetty/xml/catalog-configure.xml
```diff
@@ -0,0 +1,29 @@
...
+  <!-- 9.3 -->
+  <system systemId="http://jetty.mortbay.org/configure_9_3.dtd" uri="configure_9_3.dtd"/>
+  <system systemId="https://jetty.mortbay.org/configure_9_3.dtd" uri="configure_9_3.dtd"/>
+  <system systemId="http://jetty.eclipse.org/configure_9_3.dtd" uri="configure_9_3.dtd"/>
+  <system systemId="https://jetty.eclipse.org/configure_9_3.dtd" uri="configure_9_3.dtd"/>
+
+  <!-- 10.0 -->
+  <system systemId="http://jetty.eclipse.org/configure_10_0.dtd" uri="configure_10_0.dtd"/>
+  <system systemId="https://jetty.eclipse.org/configure_10_0.dtd" uri="configure_10_0.dtd"/>
+</catalog>
```
gregw: EOLN missing

jetty-core/jetty-xml/src/main/resources/org/eclipse/jetty/xml/catalog-org.w3.xml
```diff
@@ -0,0 +1,18 @@
...
+  <public publicId="XMLSchema.dtd" uri="XMLSchema.dtd"/>
+  <public publicId="xml.xsd" uri="xml.xsd" />
+
+  <!-- publicId names used within other existing dtd's -->
+  <public publicId="datatypes" uri="datatypes.dtd" />
+
+  <!-- systemIds for these resources -->
+  <system systemId="http://www.w3.org/2001/XMLSchema.dtd" uri="XMLSchema.dtd" />
+  <system systemId="https://www.w3.org/2001/XMLSchema.dtd" uri="XMLSchema.dtd" />
+</catalog>
```
gregw: EOLN missing

jetty-core/jetty-xml/src/main/resources/META-INF/services/java.net.spi.URLStreamHandlerProvider
```diff
@@ -0,0 +1 @@
+org.eclipse.jetty.xml.XmlParser$JettyLoaderURLStreamHandlerProvider
```
gregw: EOLN missing

jetty-core/jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlParser.java
```diff
@@ -854,4 +850,35 @@ public void remove()
...
+                        URL url = Loader.getResource(path);
+                        if (url == null)
+                            throw new FileNotFoundException("Unable to find Loader resource: " + path);
+                        return url.openConnection();
+                    }
+                };
+            }
+            return null;
+        }
+    }
```
gregw: This class needs better explanation.  I'm very cautious about URL and any JVM-wide classes like this.

jetty-core/jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlParser.java
```diff
@@ -854,4 +850,35 @@ public void remove()
...
+        {
+            if ("jetty-loader".equals(protocol))
+            {
+                return new URLStreamHandler()
+                {
+                    @Override
+                    protected URLConnection openConnection(URL u) throws IOException
+                    {
+                        String path = u.getPath();
+                        URL url = Loader.getResource(path);
```
gregw: This is a scary change of behaviour.  Previously the dtd/schemas were obtained by essentially `XmlConfiguration.class.getResource(String)`.   Now they are being directly loaded from the thread context classloader or the system classloader.

Are we sure that multiple versions of Jetty can no co-exist in the same JVM?    What happens if a future version wants to change the implementation of this URLStreamHandler?

gregw (changes requested):
Whilst the catalog is kind of interesting, I'm not sure we want to depend on an installed URLHandler.  I have concerns with multiple versions of jetty cohabitating the same JVM.

Moreover, I'm not exactly sure what problem we are fixing here?

I get it that there were some exception handling cleanups, but not sure why we have to switch to use the catalogs for that?

jetty-core/jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlParser.java
joakime: The Servlet descriptors, dtds/xsds, were loaded with the `Loader.getResource(String)` before.

This protocol handler `jetty-loader` preserves that.

jetty-core/jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlParser.java
joakime: Unlike pre-Java 9 days, a URL protocol handler is now on-demand, and not registered permanently across the entire JVM once and only once.

In the pre-java 9 days, if you wanted a URL protocol handler, you had 2 techniques.

1. A System Property that needed to be set at start of the JVM (`-Djava.protocol.handler.pkgs=<package-name>`)
2. Using the static method [`java.net.URL.setURLStreamHandlerFactory()`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/net/URL.html#setURLStreamHandlerFactory(java.net.URLStreamHandlerFactory)), which could  ever be called once.

Now, the protocol handlers are service loader based, and only are searched when a new protocol is encountered during an `openStream()` event.

jetty-core/jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlParser.java
gregw: Nope: https://github.com/eclipse/jetty.project/blob/jetty-11.0.11/jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlConfiguration.java#L1906-L1911

jetty-core/jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlParser.java
gregw: Hmmm maybe.... but what problem are we solving by switching to catalogs?  They are still baked into the jar, so we are just switching from java to xml that still has to be built.

So why do we need to introduce the extra complexity of a URL handler?  What is the benefit?

joakime:
> Whilst the catalog is kind of interesting, I'm not sure we want to depend on an installed URLHandler. I have concerns with multiple versions of jetty cohabitating the same JVM.

The same Jetty classes, but from different versions of Jetty in the same JVM is not currently supported in Jetty 7 thru Jetty 11, I don't see this a concern.

> Moreover, I'm not exactly sure what problem we are fixing here?

The EE8 tooling mangles the entire schema / dtd list.
The locations, the publicIds, the systemIds, etc are all destroyed in the transition to EE8.

This mangling was discovered during the other Resource API PRs.
I decided to try to fix this discovery in this PR.
What was discovered was also that the EE9 vs EE10 layer had different essential definitions and even missing mappings for these dtds/xsds.
First attempt was to just fix all of the errors, but it was quickly discovered that the EE8 layer is also using the resources from a fundamentally different base URI. (`<path-to-jar>!/javax/servlet/*` vs `<path-to-jar>!/jakarta/servlet/*`)

The next attempt, before I switched to XML Catalogs (an XML standard from 2005, with a standard Java API since Java 9), was to take the EE8 tooling out of the picture and move all of those definitions out of the EE9 tree and into to `jetty-xml` in the prior `XmlEntities` class, this also had the side effect of making the definitions, behaviors, and rules consistent across all of the Servlet levels, regardless of EE level.  We would only have 1 place to manage this list.

See: https://github.com/eclipse/jetty.project/blob/b6bfdb5a01e78d7f7b4f0cf8a23bca30a7de0b0c/jetty-core/jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlEntities.java

But Jan didn't like this centralized approach.

She encouraged me to research using the new Environment.
But going to the Environment approach brought back all of the broken behaviors that EE8 tooling introduces.

So I had to look for another approach.

> I get it that there were some exception handling cleanups, but not sure why we have to switch to use the catalogs for that?

The XML Catalogs are a standard (from 2005, with a Java standard API since Java 9).

This PR is currently in a transitional state, and I only committed to start the conversation about XML Catalog again, which is what we are doing, yay. :nerd_face: 

When you use a Catalog XML from Java, the Catalog itself becomes the "base uri" for all of the mentioned uri (relative) references.
That made the baseline xsd/dtds for XmlParser incredibly easy to implement, and is fully JPMS compliant as well.

See: 
* [`catalog-configure.xml`](https://github.com/eclipse/jetty.project/blob/fix/jetty-12-xml-namespace-cleanup/jetty-core/jetty-xml/src/main/resources/org/eclipse/jetty/xml/catalog-configure.xml) - the core `Configure.dtd` definitions
* [`catalog-org.w3.xml`](https://github.com/eclipse/jetty.project/blob/fix/jetty-12-xml-namespace-cleanup/jetty-core/jetty-xml/src/main/resources/org/eclipse/jetty/xml/catalog-org.w3.xml) - the core XML datatypes/xsd support
* [`CatalogTest.java`](https://github.com/eclipse/jetty.project/blob/fix/jetty-12-xml-namespace-cleanup/jetty-core/jetty-xml/src/test/java/org/eclipse/jetty/xml/CatalogTest.java) - showing how `Catalog` works with its lookups.

But when it came to the dtds/xsds for Servlet, we have a few oddities to worry about.

EE Level | Jar XSD / DTD are located | Owner of Jar | Base URI for DTD/XSD resources
---- | ----- | ---- | ----
ee8 | jetty-servlet-api-4.x.jar | Jetty | `<jar>!/javax/servlet/*`
ee9 | jakarta-servlet-api-5.x.jar | Jakarta EE | `<jar>!/jakarta/servlet/*`
ee10 | jakarta-servlet-api-6.x.jar | Jakarta EE | `<jar>!/jakarta/servlet/*`

This means that referencing the xsd/dtd jar resources from `jetty-ee#-servlet` across to the other JAR is now subject to JPMS rules on jar file access.
And that means the EE8 tooling that modifying the String references (that were in java code) from EE9 to EE8 has to do extra work to _only_ change the base URI to where those resources are located and not muck with the publicIds or systemIds (that were in java code).

Examples of mangling.
```
correct: https://jakarta.ee/xml/ns/jakartaee/jsp_3_0.xsd
mangled: https://javax.ee/xml/ns/javaxee/jsp_3_0.xsd

correct: http://xmlns.eclipse.org/xml/ns/jakartaee/web-app_4_0.xsd
mangled: http://xmlns.eclipse.org/xml/ns/javaxee/web-app_4_0.xsd
```

Using an XML catalog fixes the "java source" issues around EE9 -> EE8 tooling and moves the definitions to an XML file.
This means that only the base URI is different and needs to be adjusted now.
That's defined by the `xml:base="jetty-loader:jakarta/servlet/` entry in the catalog XML (something that the EE8 tooling should be able to do easily and reliably).

Regarding the `jetty-loader` URL protocol techniques, that's just to satisfy the cross module access requirements that we had before.

Ideally if we can get the Servlet spec to include a `catalog.xml` in the JAR file the `xml:base` and this entire `jetty-loader` URL protocol technique is no longer required.
I was going to suggest that to the servlet spec on monday. (I wonder if I could convince Eclipse to release a maintenance release of Servlet 5.x just for a catalog?)

gregw:
> The same Jetty classes, but from different versions of Jetty in the same JVM is not currently supported in Jetty 7 thru Jetty 11, I don't see this a concern.

We have users that fire up different versions of jetty in different classloaders in the same JVM.    This is why any JVM wide mechanisms such as system properties can be problematic.

So we have to at least consider how such a mechanism will work if multiple versions are required in the same JVM.  Perhaps URLHandlers are better than they once were, but let's confirm that before we commit to them.

jetty-core/jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlParser.java
joakime: Lets take a look at the history of the ee10 `WebDescriptor`.

https://github.com/eclipse/jetty.project/commits/jetty-12.0.x/jetty-ee9/jetty-ee9-webapp/src/main/java/org/eclipse/jetty/ee9/webapp/WebDescriptor.java

First it looked like this ...

https://github.com/eclipse/jetty.project/blob/04acdb72f041a15267bb79c36d8fc4d517beaadb/jetty-ee9/jetty-ee9-webapp/src/main/java/org/eclipse/jetty/ee9/webapp/WebDescriptor.java#L109-L114

Then I discovered the missing entries, bad entry names, bad JPMS setups, and it became like this ...

https://github.com/eclipse/jetty.project/blob/e848dd08522b186f94ffe8c37397cc07acb3b3b1/jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebDescriptor.java#L109-L114

You can find `Loader.getRequiredResource(String)` here

https://github.com/eclipse/jetty.project/blob/e848dd08522b186f94ffe8c37397cc07acb3b3b1/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Loader.java#L38-L61

As you can see, from the Servlet point of view, it's been Loader.getResource() the entire time.

The loading of the `jetty-xml` schemes never use the `jetty-loader` URL pattern (doesn't have to, as it's not a cross-jpms-module lookup) and is still using `XmlParser.class.getResource(String)` in the current codebase.

https://github.com/eclipse/jetty.project/blob/dd5aec090865c10390398d2829588f55775cb282/jetty-core/jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlParser.java#L94

and

https://github.com/eclipse/jetty.project/blob/dd5aec090865c10390398d2829588f55775cb282/jetty-core/jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlConfiguration.java#L1999

jetty-core/jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlParser.java
joakime: The way it exists now in Java 9+, this logic could even be isolated to only operate in a thread that is tagged by the `XmlParser`.

Eg:

* parse() - tag thread
* only resolve when tag is present on current thread
* done with parse() - tag removed
* all other uses are just a null reference

jetty-ee10/jetty-ee10-annotations/src/test/resources/web31false.xml
```diff
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <web-app 
-   xmlns="http://xmlns.jcp.org/xml/ns/javaee" 
+   xmlns="https://jakarta.ee/xml/ns/jakartaee"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-   xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/web-app_3_1.xsd"
+   xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/web-app_6_0.xsd"
    metadata-complete="false"
-   version="3.1"> 
+   version="6.0">
```
gregw, line 7: Why is a version=6.0 content inside a test file called web31?

jetty-ee10/jetty-ee10-annotations/src/test/resources/web31true.xml
```diff
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <web-app 
-   xmlns="http://xmlns.jcp.org/xml/ns/javaee" 
+   xmlns="https://jakarta.ee/xml/ns/jakartaee"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-   xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/web-app_3_1.xsd"
+   xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/web-app_6_0.xsd"
    metadata-complete="true"
-   version="3.1"> 
+   version="6.0">
```
gregw, line 7: again 6.0 content in a 3.1 file?

jetty-ee10/jetty-ee10-annotations/src/test/resources/web-fragment4true.xml
```diff
@@ -1,6 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<web-fragment xmlns="http://xmlns.jcp.org/xml/ns/javaee" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/web-fragment_4_0.xsd" metadata-complete="true" version="4.0">
+<web-fragment
+  xmlns="https://jakarta.ee/xml/ns/jakartaee"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/web-fragment_6_0.xsd"
+  metadata-complete="true"
+  version="6.0">
```
gregw, line 8: fragment4 suggests this should be a 4.0 file

gregw (changes requested):

gregw:
> Ideally if we can get the Servlet spec to include a `catalog.xml` in the JAR file the `xml:base` and this entire `jetty-loader` URL protocol technique is no longer required.
> I was going to suggest that to the servlet spec on monday. (I wonder if I could convince Eclipse to release a maintenance release of Servlet 5.x just for a catalog?)

How would that work with the `jetty-loader:` base URI thing?

joakime:
> > Ideally if we can get the Servlet spec to include a `catalog.xml` in the JAR file the `xml:base` and this entire `jetty-loader` URL protocol technique is no longer required.
> > I was going to suggest that to the servlet spec on monday. (I wonder if I could convince Eclipse to release a maintenance release of Servlet 5.x just for a catalog?)
> 
> How would that work with the `jetty-loader:` base URI thing?

The `xml:base` entry in the `catalog.xml` would not be required, and that would eliminate the need for `jetty-loader:` URL scheme.

As mentioned in my previous comment - https://github.com/eclipse/jetty.project/pull/8406#issuecomment-1207510835

> When you use a Catalog XML from Java, the Catalog itself becomes the "base uri" for all of the mentioned uri (relative) references.
> That made the baseline xsd/dtds for XmlParser incredibly easy to implement, and is fully JPMS compliant as well.

jetty-ee10/jetty-ee10-annotations/src/test/resources/web-fragment4true.xml
joakime: Dang, good point on all 3 of these.

This was a mechanical change, but I didn't consider the filenames themselves.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebDescriptor.java
```diff
@@ -87,156 +86,13 @@ protected InputSource resolveEntity(String pid, String sid)
...
                 {
-                    mapResources();
+                    URL url = WebDescriptor.class.getResource("catalog.xml");
```
gregw: should this be called `catalog-ee10.xml` for clarity?

jetty-ee10/jetty-ee10-annotations/src/test/resources/web31true.xml
joakime: This is a curious one.

Do we keep the version at 3.1, even tho while on Servlet 6.0 there is absolutely no support for Servlet 3.1 as it's across the "big bang" barrier?
Or do we update this "old" descriptor to at least Servlet 5.0?

jetty-ee10/jetty-ee10-annotations/src/test/resources/web-fragment4true.xml
joakime: The original XML (from base `jetty-12.0.x`) is from Servlet 4.0, that's not supported anymore while on Servlet 6.0.
Should this be upgraded to be "old" on Servlet 5.0 at least?

jetty-core/jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlParser.java
```diff
@@ -162,13 +156,37 @@ public boolean isValidating()
...
+    /**
+     * Load the specified URI as a catalog for entity mapping purposes.
+     *
+     * @param url the url to the catalog
+     */
+    public void addCatalog(URL url)
+    {
+        addCatalog(URI.create(url.toExternalForm()));
+    }
+
```
gregw: Do we really need the URL form?

joakime: The conversion from URL to URI needs to be somewhere, decided to keep it here.

Otherwise every `XmlParser.class.getResource(String)` style or `Loader.getResource(String)` style call (that returns URL) needs to do the conversion at that point.

joakime: The use of `URL.create(url.toExternalForm())` makes this a runtime exception if there is an issue.
vs the use of `URL.toURI()` which is a checked exception.

jetty-core/jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlParser.java
```diff
@@ -52,63 +64,45 @@
...
     private final AutoLock _lock = new AutoLock();
-    private Map<String, URL> _redirectMap = new HashMap<String, URL>();
+    // private Map<String, URL> _redirectMap = new HashMap<String, URL>();
     private SAXParser _parser;
     private Map<String, ContentHandler> _observerMap;
     private Stack<ContentHandler> _observers = new Stack<ContentHandler>();
     private String _xpath;
     private Object _xpaths;
     private String _dtd;
+    private List<CatalogResolver> _catalogs = new ArrayList<>();
```
gregw:
```suggestion
    private List<CatalogResolver> _catalogResolvers = new ArrayList<>();
```

jetty-ee10/jetty-ee10-webapp/src/main/resources/org/eclipse/jetty/ee10/webapp/catalog.xml
```diff
@@ -0,0 +1,110 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<catalog xmlns="urn:oasis:names:tc:entity:xmlns:xml:catalog" xml:base="jetty-loader:jakarta/servlet/">
```
gregw: It feels wrong to need to specify `jetty-loader:` in this file.  
If the servlet spec ultimately did include a catalog, surely it would just be a map of publicId/systemId to a uri path within the same jar?

So should it not be the resolver that knows how to turn "resources/web-app_2_2.dtd" into a `Some.class.getResource("resources/web-app_2_2.dtd");`?

gregw: Can't we do that now?   Use a catalog to convert a publicId/SystemId to a simple URI path, and then use our own resolver that knows how to turn that URI path into a reference into the specific class resource?

jetty-ee10/jetty-ee10-annotations/src/test/resources/web31false.xml
gregw: Note that if ee10 doesn't support 3.1 descriptors (@janbartel does it??) then we can probably delete the file and the test.   But 6.0 should be able to handle 5.0 descriptors at least, so we may need a new test/file

jetty-ee10/jetty-ee10-annotations/src/test/resources/web31true.xml
gregw: Ah you raised the same question I did above....   I'm thinking we update "old" to be 5.0.  @janbartel agree?

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebDescriptor.java
joakime: It can be any name we want it to be.
It's in the `org/eclipse/jetty/ee10/webapp/` directory already, does it really need to be called `catalog-ee10.xml` ?

jetty-ee10/jetty-ee10-webapp/src/main/resources/org/eclipse/jetty/ee10/webapp/catalog.xml
joakime: The `Catalog` is just an interface, we could just provide the `Catalog` implementation as an internal representation too.
It's a trivial class, I bet I can knock that out easily enough.
The base uri can then be whatever we want it to be, and not hardcoded into the XML (like it's spec'd to be)

It's the `CatalogResolver` that has lions share of effort (and we would just use it as-is), it just uses the `Catalog` as provided anyway.

joakime: > So should it not be the resolver that knows how to turn "resources/web-app_2_2.dtd" into a `Some.class.getResource("resources/web-app_2_2.dtd");`?

Not quite, as per JPMS rules, we cannot use something like `Servlet.class.getResource("resources/web-app_2_2.dtd")` (from within `jetty-ee#-servlet` classes) easily enough, that would require the Servlet jar to expose the `jakarta/servlet/resources` tree to us.

The `Loader.getResource("jakarta/servlet/resources/web-app_2_2.dtd")` usage allowed us to get around this (for now, but who knows what kind of new rules will creep up in the future).

gregw: But the module-info for the servlet API does have `opens jakarta.servlet.resources;`, so we should be able to access it?

Eitherway, the concept as I see it should be that catalog.xml transforms the ID into a local URI and then we can implement a resolver that knows how to get that URI out of a local jar file.  I think `Servlet.class.getResource("jakarta/servlet/resources/web-app_2_2.dtd")` should work, but if not then we use whatever JPMS approved method necessary.  But the point being, is that we don't need to have a `jetty-loader:` scheme in the XML (and probably not at all).

Essentially, we can use the catalog mechanism simple to replace the maps of ID that we were previously maintaining ourselves.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebDescriptor.java
gregw: I assume we will end up with an ee9 and ee8 version, so keeping the names different would probably be good... but not absolutely necessary.

joakime:
Latest commit has a new `XmlParser.addCatalog(URI catalogXml, Class<?> classBase)`

To allow the loading of a catalog which has its resources referenced via the `<class>.getResource(name)` technique, ensuring that the resource comes from the same archive location as the `classBase`.

Now all `WebDescriptor` implementations uses the `Environment` to load appropriate `catalog-<environment.getName()>.xml`

jetty-ee10/jetty-ee10-webapp/src/main/resources/org/eclipse/jetty/ee10/webapp/catalog-ee10.xml
```diff
@@ -0,0 +1,110 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<catalog xmlns="urn:oasis:names:tc:entity:xmlns:xml:catalog" xml:base="">
+  <!-- 2.2 -->
+  <public publicId="web-app_2_2.dtd" uri="resources/web-app_2_2.dtd" />
```
gregw: Shouldn't this be:
```suggestion
<catalog xmlns="urn:oasis:names:tc:entity:xmlns:xml:catalog" xml:base="jakarta/servlet">
  <!-- 2.2 -->
  <public publicId="web-app_2_2.dtd" uri="resources/web-app_2_2.dtd" />
```

or

```suggestion
<catalog xmlns="urn:oasis:names:tc:entity:xmlns:xml:catalog" xml:base="">
  <!-- 2.2 -->
  <public publicId="web-app_2_2.dtd" uri="jakarta/servlet/resources/web-app_2_2.dtd" />
```

jetty-core/jetty-xml/src/main/java/org/eclipse/jetty/xml/RemoteBaseCatalog.java
```diff
@@ -0,0 +1,216 @@
...
+ * of the catalog XML, allowing for runtime determination of the {@code xml:base}
+ * (such as pointing to the contents of a remote location)
+ * </p>
+ * <p>
+ * This is a temporary Catalog implementation, and should be removed once
+ * all of our usages of {@code servlet-api-<ver>.jar} have their own
+ * {@code catalog.xml} files.
+ * </p>
+ */
+public class RemoteBaseCatalog implements Catalog, EntityResolver
```
gregw: I like the class... but not the name.
How about `BaseClassCatalog`?

gregw (changes requested):
getting close....

jetty-ee10/jetty-ee10-webapp/src/main/resources/org/eclipse/jetty/ee10/webapp/catalog-ee10.xml
gregw: Aaah!! The URI is resolved relative to a specific class (eg Servlet.class).   Hmmm OK, but if we were to ask the servlet spec to include the catalog, it would probably be relative to the top of the jar?

Resolving against a class is good... but is there a way the contents of this file can be relative to the top level?

joakime: No, the `catalog.xml` that the `servlet-api-<ver>.jar` would live in the `jakarta/servlet/catalog.xml` location (for Servlet 5+)

That would make the URI to the catalog something like ...

```
jar:file://path/to/jakarta.servlet.api-6.0.1.jar!/jakarta/servlet/catalog.xml
```

Which would also be the baseURI for the catalog.
All non-absolute URI references are relative to that baseURI.

Basically doing a normal Base URI resolve techniques.

```
base  : jar:file://path/to/jakarta.servlet.api-6.0.1.jar!/jakarta/servlet/catalog.xml
ref   : resources/web-app_6_0.xsd
result: jar:file://path/to/jakarta.servlet.api-6.0.1.jar!/jakarta/servlet/resources/web-app_6_0.xsd
```

jetty-core/jetty-xml/src/main/java/org/eclipse/jetty/xml/RemoteBaseCatalog.java
joakime: Done

jetty-ee10/jetty-ee10-webapp/src/main/resources/org/eclipse/jetty/ee10/webapp/catalog-ee10.xml
gregw: ack

jetty-ee9/jetty-ee9-webapp/src/main/resources/org/eclipse/jetty/ee9/webapp/catalog-ee8.xml
```diff
@@ -0,0 +1,81 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
joakime: Move to jetty-ee8/jetty-ee8-webapp tree

gregw (approved):

gregw:
This broke XML validation. Running the demos I get:
```
2022-08-11 13:52:48.603:WARN :oatud.DigesterFactory:main: The XML schema [XMLSchema.dtd] could not be found. This is very likely to break XML validation if XML validation is enabled.
2022-08-11 13:52:48.603:WARN :oatud.DigesterFactory:main: The XML schema [datatypes.dtd] could not be found. This is very likely to break XML validation if XML validation is enabled.
2022-08-11 13:52:48.603:WARN :oatud.DigesterFactory:main: The XML schema [xml.xsd] could not be found. This is very likely to break XML validation if XML validation is enabled.
2022-08-11 13:52:48.606:WARN :oatud.DigesterFactory:main: The XML schema [web-jsptaglibrary_3_1.xsd] could not be found. This is very likely to break XML validation if XML validation is enabled.
2022-08-11 13:52:48.606:WARN :oatud.DigesterFactory:main: The XML schema [jakartaee_10.xsd] could not be found. This is very likely to break XML validation if XML validation is enabled.
2022-08-11 13:52:48.606:WARN :oatud.DigesterFactory:main: The XML schema [jsp_3_1.xsd] could not be found. This is very likely to break XML validation if XML validation is enabled.
```

joakime:
> oatud.DigesterFactory

That looks like some `org.apache.tomcat` stuff, they have their own registration process for those.
The changes in this PR focused on our `XmlParser` usage. (basically only used for the servlet descriptors and configure xmls)

We could look into submitting our Catalog to their layers in a future PR.

PR #8407, opened 2022-08-02
Jetty 12 : Remove unused Resource APIs

joakime:
+ If no src/main/java code used the API then it's removed.
+ APIs created just for test cases are also removed, and a similar technique is used in the test case.

Removed APIs

* `Resource.delete(String)`
* `Resource.renameTo(Resource)`
* `Resource.getFileName()`


joakime:
Once the `AnnotationParser` PR is merged (#8404) then two more APIs can be removed.

* `Resource.list()`
* `Resource.getAllResources()`

lorban (approved):

PR #8409, opened 2022-08-03
Jetty 12 : Rework `ResourceFactory` and `Resource`

lorban:
This is based on the [redux3 prototype](https://github.com/eclipse/jetty.project/pull/8352) but taken to its logical end.


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -72,135 +67,41 @@ public abstract class Resource implements ResourceFactory
...
-     * <p>
-     *     Use this if you are working with URIs from an unknown source,
-     *     such as a user configuration.  As some of the entries
-     *     might need mounting, but we cannot determine that yet.
-     * </p>
-     *
-     * @param uris collection of URIs to mount into a {@code ResourceCollection}
-     * @return the {@link Mount} with a root pointing to the {@link ResourceCollection}
-     */
-    public static Resource.Mount mountCollection(Collection<URI> uris)
```
joakime, line 83: Don't forget to remove the ResourceCollection.Mount too.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -72,135 +67,41 @@ public abstract class Resource implements ResourceFactory
...
      * <p>Mount a URI if it is needed.</p>
      *
      * @param uri The URI to mount that may require a FileSystem (e.g. "jar:file://tmp/some.jar!/directory/file.txt")
      * @return A reference counted {@link Mount} for that file system or null. Callers should call {@link Mount#close()} once
      * they no longer require any resources from a mounted resource.
      * @throws IllegalArgumentException If the uri could not be mounted.
-     * @see #mount(URI)
      */
-    public static Resource.Mount mountIfNeeded(URI uri)
+    static Resource.Mount mountIfNeeded(URI uri)
```
joakime: Can this live in ResourceFactory?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -72,135 +67,41 @@ public abstract class Resource implements ResourceFactory
...
-    }
-
     /**
      * <p>Make a Resource containing a collection of other resources</p>
      * @param resources multiple resources to combine as a single resource. Typically, they are directories.
      * @return A Resource of multiple resources.
      * @see ResourceCollection
      */
-    public static ResourceCollection of(List<Resource> resources)
+    public static ResourceCollection combine(Collection<Resource> resources)
```
joakime: List is important, as the search order is important within the ResourceCollection.
Make it only support `List<Resource>` please.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -245,42 +146,13 @@ public static String dump(Resource resource)
             .formatted(resource.toString(), resource.exists(), resource.isDirectory(), resource.lastModified());
     }
 
-    /**
-     * Construct a resource from a url.
-     *
-     * @param url A URL.
-     * @return A Resource object.
-     */
-    public static Resource newResource(URL url)
```
joakime, line 250: Yay!

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactory.java
```diff
@@ -15,29 +15,298 @@
...
+     * found, then the {@link Loader#getResource(String)} method is used.
+     * If it is still not found, then {@link ClassLoader#getSystemResource(String)} is used.
+     * Unlike {@link ClassLoader#getSystemResource(String)} this method does not check for normal resources.
+     *
+     * @param resource the relative name of the resource
+     * @return Resource or null
+     */
+    default Resource newClassPathResource(String resource)
+    {
+        URL url = Resource.class.getResource(resource);
```
joakime: ResourceFactory.class.getResource instead?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactory.java
```diff
@@ -15,29 +15,298 @@
...
+    {
+        return newResource(Resource.toURI(resource));
+    }
+
+    default Resource newResource(Path path)
+    {
+        return newResource(path.toUri());
+    }
+
+    default ResourceCollection newResource(Collection<URI> uris)
```
joakime: `List<URI>` please

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactory.java
```diff
@@ -15,29 +15,298 @@
...
+        LifeCycle factory = new LifeCycle();
+        org.eclipse.jetty.util.component.LifeCycle.start(factory);
+        return factory;
+    }
+
+    static ResourceFactory of(Resource baseResource)
+    {
+        Objects.requireNonNull(baseResource);
+
+        if (baseResource instanceof ResourceFactory resourceFactory)
```
joakime, line 207: Resource doesn't implement ResourceFactory now

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactory.java
```diff
@@ -15,29 +15,298 @@
...
+                return baseResource.resolve(resource);
+            }
+        };
+    }
+
+    static ResourceFactory of(Container container)
+    {
+        Objects.requireNonNull(container);
+
+        LifeCycle factory = container.getBean(LifeCycle.class);
```
joakime, line 232: Worry about thread safe execution?
What if this `of(Container)` is called from multiple threads?
Something like `container.calculateIfAbsent`?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactory.java
```diff
@@ -15,29 +15,298 @@
...
+        {
+            factory = lifecycle();
+            container.addBean(factory, true);
+            final LifeCycle finalFactory = factory;
+            container.addEventListener(new org.eclipse.jetty.util.component.LifeCycle.Listener()
+            {
+                @Override
+                public void lifeCycleStopped(org.eclipse.jetty.util.component.LifeCycle event)
+                {
+                    container.removeBean(finalFactory);
```
joakime, line 244: Remove the listener itself too?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactory.java
```diff
@@ -15,29 +15,298 @@
...
+
+        if (baseResource instanceof ResourceFactory resourceFactory)
+            return resourceFactory;
+
+        return new ResourceFactory()
+        {
+            @Override
+            public Resource newResource(URI resource)
+            {
+                // TODO add an optimized pathway that keeps the URI and doesn't go via String
```
joakime, line 215: mountIfNeeded here? and this resource factory is tracking the mount to close?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactory.java
```diff
@@ -15,29 +15,298 @@
...
+    }
+
+    static ResourceFactory.LifeCycle lifecycle()
+    {
+        LifeCycle factory = new LifeCycle();
+        org.eclipse.jetty.util.component.LifeCycle.start(factory);
+        return factory;
+    }
+
+    static ResourceFactory of(Resource baseResource)
```
joakime, line 203: We should try to remove this in a future PR.
Right now, leave this in place on this PR.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/ssl/SslContextFactory.java
```diff
@@ -95,7 +96,7 @@
  * and {@link Client} to configure HTTP or WebSocket clients.</p>
  */
 @ManagedObject
-public abstract class SslContextFactory extends AbstractLifeCycle implements Dumpable
+public abstract class SslContextFactory extends ContainerLifeCycle implements Dumpable
```
joakime, line 99: Seems like a big change.  AbstractLifeCycle to ContainerLifeCycle

joakime (changes requested):
Looks promising. 
A few changes and updates needed.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactory.java
joakime: might just make sense to synchronize here (on the container?)

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/CachedContentFactory.java
```diff
@@ -67,7 +66,7 @@ public class CachedContentFactory implements HttpContent.ContentFactory
      * Constructor.
      *
      * @param parent the parent resource cache
-     * @param factory the resource factory
+     * @param factory the resource
```
joakime: strange edit, this param is a ResourceFactory.

jetty-core/jetty-xml/src/test/java/org/eclipse/jetty/xml/XmlConfigurationTest.java
```diff
@@ -109,77 +127,80 @@ public void testPassedObject(String configure) throws Exception
...
-        XmlConfiguration configuration = new XmlConfiguration(Resource.newResource(url));
-        TestConfiguration tc = new TestConfiguration("tc");
-        configuration.getProperties().putAll(properties);
-        configuration.configure(tc);
+        try (ResourceFactory.Closeable resourceFactory = ResourceFactory.closeable())
+        {
+            XmlConfiguration configuration = new XmlConfiguration(resourceFactory.newResource(url));
+            TestConfiguration tc = new TestConfiguration("tc");
+            configuration.getProperties().putAll(properties);
+            configuration.configure(tc);
```
joakime, line 137: Once this line is run, the try-with-resources could be finished.

jetty-core/jetty-xml/src/test/java/org/eclipse/jetty/xml/XmlConfigurationTest.java
```diff
@@ -193,82 +214,85 @@ public void testNewObject(String configure) throws Exception
...
+                        ((TestConfiguration)object).setTestString("NEW DEFAULT");
+                    }
                 }
-            }
-        };
-        configuration.getProperties().putAll(properties);
-        TestConfiguration tc = (TestConfiguration)configuration.configure();
+            };
+            configuration.getProperties().putAll(properties);
+            TestConfiguration tc = (TestConfiguration)configuration.configure();
```
joakime: Once this line is run, the try-with-resources could be finished.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
lorban: Done.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
lorban: Done.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactory.java
lorban: Done.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactory.java
lorban: Excellent catch!

This is not easy as it looks as there is nothing to rely on that offers an atomic `compute` semantic, and using a lock here might be overkill.

I'll leave this as-is for now, with a TODO.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactory.java
lorban: Done.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactory.java
lorban: Agreed.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/ssl/SslContextFactory.java
lorban: Is it? @gregw made the exact same change in the original redux 3 PR so I thought this was okay.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/CachedContentFactory.java
lorban: It's probably a leftover of the many refactorings I made around this class. Let me revert this change.

jetty-core/jetty-xml/src/test/java/org/eclipse/jetty/xml/XmlConfigurationTest.java
lorban: Let me change this.

jetty-core/jetty-xml/src/test/java/org/eclipse/jetty/xml/XmlConfigurationTest.java
lorban: Let me change this.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactory.java
lorban: Done.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactory.java
lorban: That's not needed as the returned `ResourceFactory` is working against an already-mounted `Resource` and will only create sub-resources. This cannot end up requiring a new mount.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/ssl/SslContextFactory.java
joakime: Ah, he did? oh well, disregard my earlier comment.

joakime: Perhaps `SslContextFactory` should only take `Path` and/or `Resource` as configurations, not `String`, then we can skip the whole `ResourceFactory` / Mount management effort within this class.

Let the users of the class manage it.

The only impact I see is how to make that work from a jetty XML configuration.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/ssl/X509Test.java
```diff
@@ -138,7 +154,7 @@ public void testClientClassWithSni() throws Exception
     {
         SslContextFactory clientSsl = new SslContextFactory.Client();
         Path keystorePath = MavenTestingUtils.getTestResourcePathFile("keystore_sni.p12");
-        clientSsl.setKeyStoreResource(Resource.newResource(keystorePath));
+        clientSsl.setKeyStoreResource(ResourceFactory.root().newResource(keystorePath));
```
joakime, line 157: I'm starting to wonder about `.root()`.
What if we hid `.root()` (as private) and had instead a more clear method name on `ResourceFactory`?

``` java
Resource res = ResourceFactory.newUnmanagedResource(path);
```

lorban: There isn't a single static method on `ResourceFactory` that returns `Resource` as that is reserved for `ResourceFactory` instances.

This proposal breaks that principle.

joakime: It could be `Resource.newUnmanagedResource(path)` too, doesn't have to be `ResourceFactory`.

Internally, the implementation would still be ...

``` java
public static Resource newUnmanagedResource(path)
{
    return ResourceFactory.root().newResource(path);
}
```

That prevents users from using the returned `ResourceFactory` from the `.root()` call in other methods elsewhere that take a `ResourceFactory`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.concurrent.CopyOnWriteArrayList;
+
+import org.eclipse.jetty.util.IO;
+import org.eclipse.jetty.util.component.AbstractLifeCycle;
+import org.eclipse.jetty.util.component.Dumpable;
+import org.eclipse.jetty.util.component.DumpableCollection;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class ResourceFactoryInternals
```
joakime, line 29: Do we want this somewhere, package namespace wise, where we setup the JPMS module-info.class to not expose this to other modules?

Like we do with other ".internal." packages?

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/ssl/X509Test.java
lorban: Only `ResourceFactory` instances can create `Resource` instances. Your proposal is breaking that other principle. :smiley:

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
lorban: Yes, I've made this class package-protected for this purpose but I did not consider JPMS. Maybe it should go to some `.internal.` package instead?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
lorban: This went to `ResourceFactoryInternals`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -55,148 +46,22 @@
  * a file, a URL or an entry in a jar file.
  * </p>
  */
-public abstract class Resource implements ResourceFactory
+public abstract class Resource
```
joakime, line 50: Can't this become an interface again?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -55,148 +46,22 @@
  * a file, a URL or an entry in a jar file.
  * </p>
```
joakime, line 48: Alternative javadoc ...

```
<p>
A Resource is a {@link Path} object pointing to a File or Directory that can be represented by a Java {@link FileSystem}.
</p>
<p>
Supports real FileSystems, and also <a href="https://docs.oracle.com/en/java/javase/17/docs/api/jdk.zipfs/module-summary.html">ZipFS</a>.
</p>
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
lorban: I've personally never seen `Resource` being an interface. I'd consider that only after all the other cleanups have been done, i.e.: that's a job for another PR.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
lorban: I've used your javadoc but slightly adapted it.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime: Hmm, I must be imaging things, I could have sworn that it briefly was an interface over the past couple of months, but I can find no proof of that.

PR #8410, opened 2022-08-03
Jetty 12 : Remove static Resource.isContainedIn API

joakime:
Turns out the static `Resource.isContainedIn(Resource,Resource)` was superfluous.
We have a normal (non-static) `.isContainedIn()` method already.


lorban (approved):
Good riddance!!!

PR #8411, opened 2022-08-03
Jetty 12 : Renamed FutureFormFields to just FormFields, as it is mostly used

sbordet:
as a CompletableFuture, and the "Future" prefix is soo Java 5 :)

Improved releasing of chunks in FormFields implementation.

Improved configuration of FormFields for max number of fields and max form content length.


joakime:
@sbordet test failure ...

```
[ERROR]   DelayedHandlerTest.testDelayedFormFields:428 
Expected: a string containing "null"
     but: was "]"
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
```diff
@@ -103,44 +99,33 @@ protected Request.Processor delayed(Request request, Request.Processor processor
...
     private static class UntilFormFieldsProcessor implements Request.Processor, BiConsumer<Fields, Throwable>
     {
         private final Request _request;
         private final Request.Processor _processor;
-        private final Charset _charset;
         private Response _response;
         private Callback _callback;
 
-        public UntilFormFieldsProcessor(Request request, Request.Processor processor, Charset charset)
+        public UntilFormFieldsProcessor(Request request, Request.Processor processor)
```
joakime, line 121: Removed `Charset charset`?
How is the HTML special form field `_charset_` handled now?

joakime (commented):
Generally like it.
Just curious if the `_charset_` (hidden) form field behavior is still covered?

Also, lets get a green CI build first. (there's a test failure right now)

sbordet:
@joakime fixed test.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
sbordet: I'm not aware of `_charset_` for `application/x-www-form- urlencoded`. Do you have a reference?
I only saw it for multipart.

As for the removal of the charset, it is a pre-check to avoid delaying if the content-type is not `application/x-www-form- urlencoded`. If it is, the charset will be computed by `FormFields`.
I think it's wrong to be able to create an instance of `UntilFormFieldsProcessor` with a `Charset`, because it may not match the one present in the request headers.

joakime (approved):
:+1:

PR #8412, opened 2022-08-03
Jetty 12 : Fixing Mount leak in `MetaInfConfiguration`

joakime:
This is a small (potential) patch for the resource leak in branch `jetty-12.0.x-rework-resource-factory` (aka redux 3).

Turns out the URI matching done in MetaInfConfiguration against CONTAINER_JAR_PATTERN and WEBINF_JAR_PATTERN introduced an extra unnecessary `newResource()` call.

The fix here was to avoid converting a `List<Resource>` to `List<URI>` to just match against the provided patterns, which then does a `resourceFactory.newResource(URI)` on the pre-existing Resource.

+ Introducing `UriPatternPredicate` to replace `PatternMatcher`
+ Creating `ResourceUriPatternPredicate` specific implementation to to allow filtering of resources by their URI
+ Reworked `MetaInfConfiguration` to filter lists of URI or Resources in a way to avoid recreating the Resource object.

### DIFF
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/FileID.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/FileID.java
@@ -46,6 +46,46 @@ public class FileID
     }
 
     /**
+     * Retrieve the extension of a URI path.
+     * This is the name of the last segment of the URI path with a substring
+     * for the extension (if any), including the dot, lower-cased.
+     *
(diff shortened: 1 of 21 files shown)


lorban:
@joakime I merged then reverted this PR because while it fixes the mount leaks, it introduces new bugs in `QuickStartTest`. I think you touched too many things so we need to reduce the scope of these changes.

PR #8413, opened 2022-08-04
Jetty 12 : Strengthen WebSocket upgrade contract and other improvements

lachlan-roberts:
- Strengthen contract of WebSocket upgrade methods by defining in javadoc that an error response needs to be generated by them if they are not returning a WebSocket Object.
- Remove `WebSocketNegotiation` from the public API, use (req, resp, cb) style method in `WebSocketNegotiator` instead.
- Factor out a `isWebSocketUpgradeRequest` method for the `Handshaker` so that the Upgrade Servlet/Filters can check it is really an upgrade attempt before allocating the Blocking Callback and other resources.
- Add `Attachment` interface onto the Servlet Request objects so the wrapped `HttpServletRequest` and `HttpServletResponse` can be stored there and retrieved once `websocket-core` calls back the API layer.


jetty-core/jetty-websocket/websocket-core-server/src/main/java/module-info.java
```diff
@@ -19,4 +19,5 @@
     requires transitive org.eclipse.jetty.websocket.core.common;
 
     exports org.eclipse.jetty.websocket.core.server;
+    exports org.eclipse.jetty.websocket.core.server.internal;
```
sbordet: No, we can't export internal classes, that's why they are internal.

jetty-core/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/CreatorNegotiator.java
```diff
@@ -11,19 +11,13 @@
 // ========================================================================
 //
 
-package org.eclipse.jetty.websocket.core.server.internal;
+package org.eclipse.jetty.websocket.core.server;
```
sbordet: This class should be an implementation detail... why was moved out of internal?

jetty-core/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/Handshaker.java
```diff
@@ -29,5 +29,24 @@ static Handshaker newInstance()
         return new HandshakerSelector();
     }
 
+    boolean isWebSocketUpgradeRequest(Request request);
```
sbordet: Javadocs for this too.

jetty-core/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/Handshaker.java
```diff
@@ -29,5 +29,24 @@ static Handshaker newInstance()
...
+
+    /**
+     * This will attempt to upgrade a request to WebSocket.
```
sbordet:
```suggestion
     * <p>Attempts to upgrade a request to WebSocket.</p>
```

jetty-core/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/Handshaker.java
```diff
@@ -29,5 +29,24 @@ static Handshaker newInstance()
...
+     *
+     * <p>This method returns true if a WebSocket upgrade was attempted in which case this method takes responsibility for
+     * completing the callback and generating a response, the request may be upgraded to WebSocket or some error response
+     * will be sent. If this method returns false the WebSocket upgrade was not accepted and the caller is still responsible
+     * for completing the callback and generating a response.</p>
```
sbordet:
```suggestion
     * <p>Returns {@code true} if the WebSocket upgrade is successful and a successful response is generated and the callback eventually completed,
     * or if the WebSocket upgrade failed and a failure response is generated and the callback eventually completed.
     * Returns {@code false} if a response is not generated and the caller is responsible for generating a response and completing the callback.</p>
```

jetty-core/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/Handshaker.java
```diff
@@ -29,5 +29,24 @@ static Handshaker newInstance()
...
+     *
+     * @param negotiator the negotiator
+     * @param request the request
+     * @param response the response
+     * @param callback the callback
+     * @param components the WebSocket components
+     * @param defaultCustomizer the customizer
+     * @return true if the WebSocket upgrade was attempted
```
sbordet:
```suggestion
     * @return true if a response was generated, false if a response is not generated
```

jetty-core/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/WebSocketCreator.java
```diff
@@ -15,23 +15,26 @@
 
 import org.eclipse.jetty.util.Callback;
 
-// TODO: improve javadoc.
 /**
  * Abstract WebSocket creator interface.
```
sbordet: All javadocs paragraphs must be wrapped with `<p>...</p>`.

jetty-core/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/WebSocketCreator.java
```diff
@@ -15,23 +15,26 @@
...
     /**
      * Create a websocket from the incoming request.
      *
-     * @param req the request details
-     * @param resp the response details
-     * @param callback callback
-     * @return a websocket object to use, or null if no websocket should be created from this request.
+     * <p>If the creator returns null it is responsible for completing the {@link Callback} and sending a response.
+     * But if the creator intends to return non-null WebSocket object, it MUST NOT write content to the response or
+     * complete the {@link Callback}.</p>
```
sbordet: Can it modify the response headers?

jetty-core/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/WebSocketMappings.java
```diff
@@ -229,6 +228,21 @@ public WebSocketNegotiator getMatchedNegotiator(String target, Consumer<PathSpec
         return negotiator;
     }
 
+    /**
+     * This will attempt to find a WebSocket mapping and upgrade a request to WebSocket.
```
sbordet: Javadoc guidelines suggest to use 3rd person to write javadocs, so rather than "This will do something" use "Does something".
Here in particular, "Attempts to upgrade a request to WebSocket, etc.".

jetty-core/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/WebSocketMappings.java
```diff
@@ -239,13 +253,25 @@ public boolean upgrade(Request request, Response response, Callback callback, Co
...
-        if (negotiator == null)
-            return false;
-
-        // We have an upgrade request
-        return handshaker.upgradeRequest(negotiator, request, response, callback, components, defaultCustomizer);
+        return upgrade(negotiator, request, response, callback, defaultCustomizer);
     }
 
+    /**
+     * This will attempt to upgrade a request to WebSocket.
```
sbordet: Ditto.

jetty-core/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/WebSocketNegotiator.java
```diff
@@ -13,49 +13,31 @@
...
+import org.eclipse.jetty.util.Callback;
 import org.eclipse.jetty.websocket.core.Configuration;
 import org.eclipse.jetty.websocket.core.FrameHandler;
-import org.eclipse.jetty.websocket.core.server.internal.CreatorNegotiator;
 
 public interface WebSocketNegotiator extends Configuration.Customizer
 {
-    FrameHandler negotiate(WebSocketNegotiation negotiation) throws IOException;
+    /**
+     * Create a {@link FrameHandler} from the incoming request.
```
sbordet: Use 3rd person and wrap with `<p>`.

Is it worthwhile mentioning when this method is called with respect to upgrade methods and creation methods?

jetty-core/jetty-websocket/websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/internal/RFC6455Handshaker.java
```diff
@@ -32,27 +32,29 @@
...
-                LOG.debug("not upgraded method!=GET {}", request);
+                LOG.debug("not upgraded version!=1.1 {}", request);
             return false;
         }
 
-        if (!HttpVersion.HTTP_1_1.is(request.getConnectionMetaData().getProtocol()))
+        if (!super.isWebSocketUpgradeRequest(request))
+            return false;
+
+        if (!HttpMethod.GET.is(request.getMethod()))
```
sbordet: As this method is tested for every request, I think the test for GET was first because it is the cheapest.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextResponse.java
```diff
@@ -85,7 +86,8 @@ public class ServletContextResponse extends ContextResponse
     private ResponseWriter _writer;
 
     private long _contentLength = -1;
-    
+    final AtomicReference<Object> _attachment = new AtomicReference<>();
```
sbordet: `private` !!!

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextRequest.java
```diff
@@ -122,6 +124,7 @@ public static ServletContextRequest getBaseRequest(ServletRequest request)
     final HttpInput _httpInput;
     final String _pathInContext;
     Charset _queryEncoding;
+    final AtomicReference<Object> _attachment = new AtomicReference<>();
```
sbordet: `private` !!!

But, if you use attributes, do you need attachments?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextRequest.java
```diff
@@ -85,7 +86,8 @@
 public class ServletContextRequest extends ContextRequest implements Runnable
 {
     public static final String __MULTIPART_CONFIG_ELEMENT = "org.eclipse.jetty.multipartConfig";
-
+    public static final String WEBSOCKET_WRAPPED_REQUEST_ATTRIBUTE = "org.eclipse.jetty.websocket.wrappedRequest";
+    public static final String WEBSOCKET_WRAPPED_RESPONSE_ATTRIBUTE = "org.eclipse.jetty.websocket.wrappedResponse";
```
sbordet: I don't think we want to have "websocket"-named constant in this class, feels like the wrong place.
If they are used by the `Handshaker` implementation, they should be a constant there.

jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jetty-server/src/main/java/org/eclipse/jetty/ee10/websocket/server/JettyWebSocketCreator.java
```diff
@@ -25,6 +25,10 @@
     /**
      * Create a websocket from the incoming request.
```
sbordet: Use 3rd person and wrap with `<p>`.

jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jetty-tests/src/test/java/org/eclipse/jetty/ee10/websocket/tests/SimpleEchoTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.eclipse.jetty.server.ServerConnector;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class SimpleEchoTest
```
sbordet, line 35: Don't we have a million of these already? Why do we need this one?

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/ContextHandler.java
```diff
@@ -2452,6 +2452,9 @@ public ClassLoader getCallerClassLoader(int depth)
 
     public static class CoreContextRequest extends ContextRequest
     {
+        public static final String WEBSOCKET_WRAPPED_REQUEST_ATTRIBUTE = "org.eclipse.jetty.websocket.wrappedRequest";
+        public static final String WEBSOCKET_WRAPPED_RESPONSE_ATTRIBUTE = "org.eclipse.jetty.websocket.wrappedResponse";
+
```
sbordet: I would not taint with "WebSocket" anything in this module.
Can these constants be moved in some class in the websocket modules?

sbordet (changes requested):

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextRequest.java
lachlan-roberts: They are not used in `Handshaker`, and is not used at all by websocket-core.

This is a replacement for the use of the `Attachment` interface which greg didn't like.

This is used from the `JakartaWebSocketServerContainer`, `JettyWebSocketServerContainer`, `JettyWebSocketServlet` and `WebSocketUpgradeFilter`.

jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jetty-tests/src/test/java/org/eclipse/jetty/ee10/websocket/tests/SimpleEchoTest.java
lachlan-roberts: We didn't have a single test which is just a simple echo, and its useful to have this for debugging things, or just for using it as a base for other tests.

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/ContextHandler.java
lachlan-roberts: I have moved these into `WebSocketConstants` for now, still seems a weird place to put them because websocket-core doesn't use them ever.

sbordet (approved):

PR #8423, opened 2022-08-05
10.0.x: Fix drainTo when `head == tail` but the queue isn't empty

### LINKED ISSUE #8414: BlockingArrayQueue drops all contents on drain
**Jetty version(s)**
11.0.11 (latest) and older

openjdk version "11.0.15" 2022-04-19

**OS type/version**
Linux: Ubuntu 20.04

**Description**
In specific scenarios, `org.eclipse.jetty.util.BlockingArrayQueue` loses all elements on call to `drainTo`.

**How to reproduce?**
- Create a BlockingArrayQueue with `64 * N` elements, where `N > 1` (64 matches `DEFAULT_GROWTH`)
- Make a call to method `int drainTo(Collection<? super E> c, int maxElements)` with any `maxElements` value
- The queue is now empty and no items were added to the collection it was supposed to drain to

The example works as expected if an additional element is added (and then removed), likely by making the queue grow.

Minimal example:
```java
public static void main(String[] args) {
  BlockingQueue<Integer> queue = new BlockingArrayQueue<>();
  for (int i = 0; i < 192; i++) {   // 192 = 3 * 64
    queue.add(i);
  }

  // uncomment this and the example works as expected
//  queue.add(5);
//  queue.poll();

  System.out.println("Source queue size before draining: " + queue.size());
  List<Integer> list = new ArrayList<>();
  queue.drainTo(list, 1);
  System.out.println("Source queue size after draining: " + queue.size() + ", dest list size: " + list.size());
}
```
Output:
```
Source queue size before draining: 192
Source queue size after draining: 0, dest list size: 0
```

lorban:
Fixes https://github.com/eclipse/jetty.project/issues/8414

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/BlockingArrayQueue.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/BlockingArrayQueue.java
@@ -608,13 +608,19 @@ public class BlockingArrayQueue<E> extends AbstractList<E> implements BlockingQu
             _headLock.lock();
             try
             {
+                if (_size.get() == 0)
+                    return 0;
+
                 final int head = _indexes[HEAD_OFFSET];
                 final int tail = _indexes[TAIL_OFFSET];
                 final int capacity = _elements.length;
 
                 int i = head;
-                while (i != tail && elements < maxElements)
+                while (elements < maxElements)
                 {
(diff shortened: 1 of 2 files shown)


joakime (dismissed):
LGTM

gregw (approved):
LGTM

Issue #8397, opened 2022-08-01
Jetty 12 - is `Resource.copyTo(Path)` for file to file? or dir to dir?


joakime:
**Jetty version(s)**
Jetty 12

**Description**
The API `Resource.copyTo(Path)` has a contract of File to File.
All the way back to Jetty 7.x version of Resource.
However, in Jetty 7, the specific `JarResource` changes the contract to `Resource.copyTo(directory)` allowing the copying of a JAR contents to a directory somewhere else.

Now that we are Path based, we should just drop the `Resource.copyTo()` method and use the `java.nio.file.Files` APIs instead.
The users of the resource know if  it's a file, or directory, or an archive and decide the best path there.

janbartel:
But if you're handed a Resource, you don't know what it is, so you might not know to copy it.

joakime:
> But if you're handed a Resource, you don't know what it is, so you might not know to copy it.

Unfortunately, as the user of the Resource you need to be sure that your Resource is even capable of being copied, or even needs to be copied.
The behaviors in Jetty 9 thru 11 in this regard is also quietly broken in many places (as copyTo() doesn't do anything in a few cases i looked into)

In Jetty 12 if you are handed a resource (which is always a Path btw)

For Jar/Zip/War files, you have 2 ways to represent them in a Resource.
1. As the Jar/Zip/War itself, which is just a File, and is no more special then any other File.
2. As the root directory of the Jar/Zip/War archive (represented via the zipfs layer), this is a Directory, and is no different than any other Directory.

The logic that we need to settle on as a result of this issue ...

If the `Resource` _____  then copyTo(Path) means what?

**Is a File**  

   1. The destination is a directory then you want the file copied to the destination directory (with the same name?)
   2. The destination is a file, then you want the file contents copied to the destination file.
   3. The destination doesn't exist, then what? do we create a directory? or a file? or fail the copyTo?

**Is a Directory**

1. The destination is a directory, then you want the source directory copied to the destination directory.
    What to do with links? (not just unix symlinks, but any links from any filesystem type, even ntfs)
2. The destination is a file, then what?  Do you want to create a JAR/ZIP?  Concat all of the contents of the source into the file?
3. The destination doesn't exist, then what?  do we create a file? or a directory? or fail the copyTo?

**does not exist**

1. This case should throw an exception, as this points to a bug in the user of a Resource, it should never have reached the point of a blind copyTo(Path)

**Is a ResourceCollection**

[...]

PR #8427, opened 2022-08-08
Jetty 12 : eliminate `Resource.list()` and `Resource.getAllResources()` methods

joakime:
Now that `Resource` is based on `Path`, we can eliminate the `Resource.list()` and `Resource.getAllResources()` methods.

This now puts the users of those methods in control of what they get/want from the `Resource` they are working with.
(eg: how deep of a walk. filter by type/id/name. ignore hidden. etc)

This also puts the user of the `Resource` to decide how they want to deal with the resulting list/stream of `Paths`.
Use them directly as `Path`, or use a `ResourceFactory` that they control to create new Resources from results of their usage.

The biggest user of these APIs prior to this PR was the `AnnotationParser`, which as of PR #8404  does not use these APIs anymore.


jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/JarResourceTest.java
```diff
@@ -72,8 +73,8 @@ public void testJarFile()
         {
             Resource r = resourceFactory.newResource(uri);
 
-            Set<String> entries = new HashSet<>(r.list());
-            assertThat(entries, containsInAnyOrder("alphabet", "numbers", "subsubdir/"));
+            Set<String> entries = Files.list(r.getPath()).map(Path::getFileName).map(Path::toString).collect(Collectors.toSet());
+            assertThat(entries, containsInAnyOrder("alphabet", "numbers", "subsubdir"));
```
lorban: IIRC the trailing slash is needed when directory listing is served by the `DefaultServlet`.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/JarResourceTest.java
```diff
@@ -234,25 +218,33 @@ public void testEncodedFileName()
...
+                    .map(Path::getFileName)
+                    .map(Path::toString)
+                    .sorted()
+                    .toList();
+
+                String[] expected = new String[]{
+                    "one",
+                    "aaa",
+                    "bbb",
+                    "oddities",
```
lorban: I really think we should keep the trailing slashes, at least in this PR to avoid conflating multiple subjects.

lorban (changes requested):

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/JarResourceTest.java
joakime: I bet we have test cases for that.
I'll check.

joakime: Yep, we have testcases for `DefaultServlet`.

In EE10, it uses the `ResourceService`, which in turn uses the `ResourceListing` class, which manages the trailing slash itself.
I'll check EE9/EE8 next.

joakime: The EE9/EE8 `DefaultServlet` uses the `ResourceListing` class from `jetty-core/jetty-server` as well.
No trailing slash problem there too.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/JarResourceTest.java
joakime: Incoming PR shortly

lorban (approved):
LGTM

gregw (changes requested):
Something has gone wrong here!  A directory Resource might not be based on Path.  It may be a ResourceCollection or some maven aggregate structure or an future in memory mechanism.

As of today, we are required to list all of the combined resources within a resource collection, so I think we must have lost that functionality in the previous PRs as it cannot be assumed that a directory request resolves to a single Path

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
```diff
@@ -66,7 +66,6 @@ public static String getAsHTML(Resource resource, String base, boolean parent, S
         List<Path> listing = null;
```
gregw: line 63 above is wrong.  Not all resources are Paths.

gregw (changes requested):

gregw:
Note also that we are required to support the [ServletContext#getResourcePaths](https://docs.oracle.com/javaee/6/api/javax/servlet/ServletContext.html#getResourcePaths(java.lang.String)) API, which requires list semantics that will combine file system resources and potentially resources from within fragment jars.

This is the second time we have inadvertently removed support for this in jetty-12.  I think we need and explicit test for this functionality so that we stop deleting it!

joakime:
> Something has gone wrong here! A directory Resource might not be based on Path. It may be a ResourceCollection or some maven aggregate structure or an future in memory mechanism.
> 
> As of today, we are required to list all of the combined resources within a resource collection, so I think we must have lost that functionality in the previous PRs as it cannot be assumed that a directory request resolves to a single Path

I took a look through the usages of `Resource.list()`, and here's what I found with the non-test usages.

**org.eclipse.jetty.ee10.runner.Runner**

There could be a ResourceCollection from the command line arguments
`--lib=<entry>` and `--jar=<entry>`, but that wasn't supported in any release of `Runner`.

The changes committed to this branch now supports that, and even has a bonus side effect of supporting globs.

**ServletContextHandler.getResourcePaths(String)**
**and ContextHandler.getResourcePaths(String)**

This collected a `Set<String>` of entries consisting of `<input-path>/<classified-filename>`.
This would actually fail if one entry in the `ResourceCollection` didn't exist (the entire method would be skipped)
or if one entry wasn't a directory (this could actually crash the `ResourceCollection.list()` implementation)

The changes committed to this branch now supports a `ResourceCollection` that has empty/non-existent/non-directory entries, without creating throwaway `Resource` objects to just get a filename.

**MetaInfConfiguration.findWebInfLibJars(WebAppContext context)**

This should not ever support `ResourceCollection` as this would mean `WebAppContext.getWebInf()` is returning content that doesn't belong to the main artifact.
Such as `WEB-INF/lib/foo.jar!/WEB-INF` or `META-INF/versions/#/WEB-INF` and similar things.

I've added more TODOs to address this bug in `WebAppContext.getWebInf()`


That's it, that's all that's left of `Resource.list()` usage before this PR.

gregw:
> That's it, that's all that's left of `Resource.list()` usage before this PR.

Exactly!  that's my point.   A usage of list from the ResourceService has been removed and not detected.   We need to restore that mechanism and add a test in to be sure it is not accidentally deleted again.

joakime:
> > That's it, that's all that's left of `Resource.list()` usage before this PR.
> 
> Exactly! that's my point. A usage of list from the ResourceService has been removed and not detected. We need to restore that mechanism and add a test in to be sure it is not accidentally deleted again.

I was originally confused as to where you got that `ResourceService` had `Resource.list()` usage ...

https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java

There's no `Resource.list()` usage in Jetty 10.
The only thing remotely `list()` like in Jetty 10's ResourceService is the `resource.getListHTML` call, which would have used `Resource.list()` internally back in Jetty 10.

In Jetty 12, that's handled by the `ResourceListing` component in jetty-core/jetty-server.
Interestingly, the existing `ResourceCollection` tests that result in a Directory Index/Listing are passing.
I'll beef up the testing of `ResourceListing` directly, and add some `ResourceCollection` behaviors.

gregw:
Commit e0a9c21 added tests showing that Resource.list() was still necessary.  It reverted some of the changes to list as HTML to pass the tests.  I'll soon do the same for getting resources from the context that has a resource collection.

So I'm going to close this PR for now, as its basic premise is not correct.  I think there is some goodness/cleanups that are incidental in the branch, but I'll let @joakime harvest those into a new branch/PR.

joakime:
Didn't have to close this, I can do the requested changes in this branch still.

gregw:
> Didn't have to close this, I can do the requested changes in this branch still.

What request changes?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
```diff
@@ -94,13 +99,18 @@ public static String getAsXHTML(Resource resource, String base, boolean parent,
         }
 
         // Perform sort
-        Comparator<? super Resource> sort = switch (sortColumn)
+        if (sortColumn.equals("M"))
```
gregw, line 102: Why switching back to old style?

gregw (changes requested):
This is looking like an omnibus PR with lots of unrelated changes with little motivation.

Can this be broken up into smaller motivated PRs

joakime:
> This is looking like an omnibus PR with lots of unrelated changes with little motivation.
> 
> Can this be broken up into smaller motivated PRs

Lets start with PR #8471

joakime:
Closing, lets review fundamentals in Issue #8397

PR #8428, opened 2022-08-08
Jetty 12 : EE10 implement default servlet serving jetty-dir.css

lorban:
Make ee10 `DefaultServlet` able to serve jetty-dir.css


jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -73,7 +74,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class DefaultServlet extends HttpServlet
+public class DefaultServlet extends HttpServlet implements ResourceFactory
```
joakime: Not sure this is correct.
ResourceFactory is for creating new Resources objects.
DefaultServlet needs to resolve Resources from an existing ResourceBase-ish.
Ew, the ResourceService should be handling this, and the overridden ResourceService in the DefaultServlet should be handling this ee10 specific behavior.

lorban: As you suggested, the easiest way would be to combine the CSS and base resources into a `ResourceCollection`. But this requires https://github.com/eclipse/jetty.project/pull/8435 to be merged first.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -130,6 +125,15 @@ public Resource resolve(String subUriPath)
         Resource addedResource = null;
         for (Resource res : _resources)
         {
+            // since we allow non-directory resources, we need a way to resolve them
+            // -> match if the subUriPath contains no / and matches the resource's filename.
+            // TODO review this
+            String subUriStringNotPrependedBySlash = subUriPath;
+            if (subUriStringNotPrependedBySlash.startsWith("/"))
+                subUriStringNotPrependedBySlash = subUriStringNotPrependedBySlash.substring(1);
```
gregw, line 134: This as never been necessary before.  Resource has always treated leading slash and no leading slash the same, so we should not have to do this.... and if we did, it should not be done in the loop.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -130,6 +125,15 @@ public Resource resolve(String subUriPath)
...
+            if (!res.isDirectory() && res.exists() && !subUriStringNotPrependedBySlash.contains("/") && res.getPath().getFileName().toString().equals(subUriStringNotPrependedBySlash))
+                return res;
```
gregw, line 136: I don't understand this code?  resolving against a non directory should just result in null.  Perhaps we just skip non directory resources?

gregw (changes requested):
Other than the ResourceCollection changes, this looks good.   If you remove those, you can merge the other fixes and we can discuss collection in a different PR

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
lorban: The idea is that if you as a ResourceCollection to resolve `abc.txt` and it happens to contain a non-directory resource named `abc.txt` then it's going to serve that resource.

I'm not saying this mechanism is a good idea, but if we want to allow ResourceCollection to contain non-directory resources, we must have some form of resolving logic that can match the non-dir resources.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
lorban: This is a quick & ugly hack needed to demonstrate what it takes to be able to serve non-dir resources added to a ResourceCollection.

In case we think this mechanism is a good idea, I'll come up with a clean implementation to replace this horrible hack.

lorban:
We have to decide if we like this mechanism or not. If we do, ResourceCollection needs a way to be able to return the non-dir resources it contains from a call to resolve.

If we don't, jetty-dir.css needs to be served in some other way which involves making the `ResourceContentFactory` or its parent serve jetty-dir.css to avoid bypassing the `CachingContentFactory` that wraps `ResourceContentFactory`, i.e.: that logic cannot be in `ResourceService` unless we want to bypass `CachingContentFactory`.

ee9 is implemented by making `DefaultServlet` being able to serve jetty-dir.css, implement `ResourceFactory` and be wrapped by `ResourceContentFactory`.

lorban:
Oh and favicon went away because it's supposed to be served by DefaultHandler, not by any DefaultServlet nor ResourceHandler.

lorban:
See https://github.com/eclipse/jetty.project/issues/8443 and https://github.com/eclipse/jetty.project/issues/8456

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
gregw: Why do we want `ResourceCollection` to contain non-directory resources?    What is the use-case?

gregw:
> We have to decide if we like this mechanism or not. If we do, ResourceCollection needs a way to be able to return the non-dir resources it contains from a call to resolve.

I don't understand? `ResourceCollection` has always been able to resolve non-dir resources.     The complications in this PR appear to be from an effort to make a collection containing non-dir resources? what is the use-case for that?


> If we don't, jetty-dir.css needs to be served in some other way which involves making the `ResourceContentFactory` or its parent serve jetty-dir.css to avoid bypassing the `CachingContentFactory` that wraps `ResourceContentFactory`, i.e.: that logic cannot be in `ResourceService` unless we want to bypass `CachingContentFactory`.

`jetty-dir.css` definitely needs to be served in some other way than having it part of a resource collection.  It is a synthetic resource that "appears" to exist in every directory even if it is not there.  It is best handled, as it was previously, with an interception of the URI before resolving the resource (or perhaps after a not found resource).

gregw:
See commit e0a9c21, which added a directory test with commented out jetty-dir.css checks

joakime:
This can be closed.

PR #8433, opened 2022-08-09
improve performance of alias checking

lachlan-roberts:
We don't want to create the unmodifiableList every time we check an alias, so use `_aliasChecks` directly instead of using the getter.

<s>Also I don't think we need the `CopyOnWriteArrayList` because we do not allow modifications after the `ContextHandler` has been started.</s>

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
@@ -228,7 +228,7 @@ public class ContextHandler extends ScopedHandler implements Attributes, Gracefu
     private final List<ContextScopeListener> _contextListeners = new CopyOnWriteArrayList<>();
     private final Set<EventListener> _durableListeners = new HashSet<>();
     private Index<ProtectedTargetType> _protectedTargets = Index.empty(false);
-    private final List<AliasCheck> _aliasChecks = new CopyOnWriteArrayList<>();
+    private final List<AliasCheck> _aliasChecks = new ArrayList<>();
 
     public enum Availability
     {
@@ -1967,7 +1967,7 @@ public class ContextHandler extends ScopedHandler implements Attributes, Gracefu
                 LOG.debug("Aliased resource: {}~={}", resource, resource.getAlias());
 
             // alias checks
-            for (AliasCheck check : getAliasChecks())
+            for (AliasCheck check : _aliasChecks)
             {
                 if (check.check(path, resource))
                 {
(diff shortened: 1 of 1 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -1967,7 +1967,7 @@ public boolean checkAlias(String path, Resource resource)
                 LOG.debug("Aliased resource: {}~={}", resource, resource.getAlias());
 
             // alias checks
-            for (AliasCheck check : getAliasChecks())
```
joakime, line 1970: The reason this existed this way was to support folks that wanted to have adjustable (at runtime) list of alias checks.

lachlan-roberts: Hmm I thought alias checks could only be added before start but looks like that is not the case.

So maybe we do still need the `CopyOnWriteArrayList`, but we don't need to use `getAliasChecks()` for the unmodifiable list here.

joakime: I don't have a strong opinion on the solution you are working towards.
Just explaining the history of why things are the way they are in Jetty 11 and earlier.
Up to you to decide if this behavior is worth keeping around or changing for Jetty 12 tho.

In regards to Jetty 12 effort though ...
To add to this, there once was a desire to have different Alias checks depending on url-pattern.
So that you could reject aliases for say `/archives/*` and allow aliases for `/avatars/*`.

Frankly, only the Jetty 12 `ResourceService` should have the alias check lists, as that's where all of our serve-static-content lies now (for all EE levels).   The uses of say `ServletContext.getResource(String)` should allow anything that is requested from it as that's programmatic from Servlet APIs, not something that a user request can trigger.   It's up to the webapp developer to decide how to sanitize / evaluate if sane the user request before they call `ServletContext.getResource(String)`.   (Example: Jersey and Spring serve static content themselves, not via the DefaultServlet / ResourceHandler.  But spark uses the ResourceHandler.)

gregw: I'm not convinced we need to allow getAliasChecks to be overridable, as I don't really see the use-case for a dynamic list, nor is there any context passed on which to make that choice.  So I'm OK with this change..... but equally we could assign a single immutable list as a field and return that.

gregw (approved):

PR #8435, opened 2022-08-09
Jetty 12 : Avoid recreating `Resource` during container/jar pattern filtering in `MetaInfConfiguration`

joakime:
This fixes a few things with the URI matching done in `MetaInfConfiguration` against the regex from `CONTAINER_JAR_PATTERN` and `WEBINF_JAR_PATTERN`.

1. The old `PatternMatcher` usage introduced an extra unnecessary `newResource()` call
2. The regex patterns need to work against the container of a `jar:file://` URI, not the `jar:file:` URI itself.

The fix here was to avoid converting a `List<Resource>` to `List<URI>` and then `URI[]` to just match against the provided patterns, which then does a `resourceFactory.newResource(URI)` on the pre-existing Resource to then add it to another `List<Resource>`.

Introducing `UriPatternPredicate` to standardize the selection logic from `PatternMatcher`, and have `PatternMatcher` use it.
Creating `ResourceUriPatternPredicate` specific implementation to to allow filtering of resources by their `URI`
Reworked `MetaInfConfiguration` to filter Lists of `Resources` using these new classes to test againt the archive container, and to avoid recreating the `Resource` object.


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/FileID.java
```diff
@@ -149,35 +189,10 @@ public static boolean isArchive(String filename)
...
-        }
-        String path = uri.getPath();
-        // look for `!/` split
-        int jarEnd = path.indexOf("!/");
-        if (jarEnd >= 0)
-        {
-            return isArchive(path.substring(0, jarEnd));
-        }
-        return isArchive(path);
+        return (ext.equals(".jar") || ext.equals(".war") || ext.equals(".zip"));
```
lorban, line 195: Don't you want to use `.equalsIgnoreCase()`?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/FileID.java
```diff
@@ -255,6 +270,17 @@ public static boolean isHidden(Path base, Path path)
...
 
+    /**
+     * Is the URI pointing to a Java Archive (JAR) File (not directory)
+     *
+     * @param uri the uri to test.
+     * @return True if a .jar file.
+     */
+    public static boolean isJavaArchive(URI uri)
+    {
+        return ".jar".equals(getExtension(uri));
```
lorban, line 281: Don't you want to use `.equalsIgnoreCase()`?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -1779,6 +1779,25 @@ public static List<URI> split(String str)
...
+    /**
+     * Optionally wrap a URI with `jar:file:` (zipfs) syntax if the input URI
+     * is an archive and isn't wrapped.
+     *
+     * @param uri the input URI
+     * @return the input URI optionally wrapped with `jar:` if it's an archive.
+     * @see FileID#isJavaArchive(Path)
+     * @see URIUtil#toJarFileUri(URI)
+     */
+    public static URI wrapArchiveToJarFileUri(URI uri)
```
lorban: Doesn't `toJarFileUri()` already provide the exact same functionality?

lorban (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/FileID.java
joakime: `getExtension()` returns it in lowercase already.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: Almost, it the `toJarFileURI()` will return null if the input URI is not a java archive, it's a subtle difference.

I wonder if we can just adjust `toJarFileURI` instead?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/FileID.java
joakime: `getExtension()` returns it in lowercase already.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/PatternMatcher.java
```diff
@@ -14,90 +14,20 @@
 package org.eclipse.jetty.util;
 
 import java.net.URI;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.regex.Pattern;
 
+// TODO: replace with UriPatternPredicate or ResourceUriPatternPredicate to avoid Resource -> URI -> Resource flows.
```
janbartel: If you're introducing these new predicates, I don't see the point in retaining the PatternMatcher class.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -81,10 +81,6 @@ private static void gatherUniqueFlatResourceList(List<Resource> unique, List<Res
                     throw new IllegalArgumentException("Does not exist: " + r);
                 }
 
-                if (!r.isDirectory())
```
janbartel, line 84: Why is this line removed in this pr?

janbartel (changes requested):
Fully remove PatternMatcher.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
joakime: Basically this restriction was bogus.

The `ResourceCollection` objects used in `MetaInfConfiguration` would be unable to hold the TLD lists then.
This restriction for only directory was a new thing added in Jetty 12, and wasn't present in Jetty 11.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/PatternMatcher.java
joakime: If nothing else is using it, i'll remove it.

joakime: Removed in commit 1066ae04f37ac0b585ad8705826cadade90e9e4c

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: URIUtil.wrapArchiveToJarFileUri has been removed.

lorban (approved):

janbartel (approved):

PR #8436, opened 2022-08-09
Jetty 12 : More testing for Resource alias

joakime:
Reworked the `PathResource.checkAliasPath` ...

If the URI is a `jar:file:` based, the URI only check is skipped and the path segment checks still apply.
Also, the symlink check is skipped (as it's not possible to have symlinks in zipfs)


joakime:
I need more test cases!

What is a combination of ..

1. A zipfs PathResource
2. A request path used in `resolve(String)` that resolves to a resource that exists, but can be viewed as an alias.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -50,15 +50,17 @@ private Path checkAliasPath()
     {
         Path abs = path;
 
+        boolean isJarArchive = uri.toASCIIString().startsWith("jar:file:");
```
gregw: @lorban I remain convinced that we should have a `JarPathResource` class to reduce these conditionals and to do other optimizations so that both Path and Jar Path resources are faster.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -50,15 +50,17 @@ private Path checkAliasPath()
...
-         * from an #addPath(String) is normalized away during
+         * from an #resolve(String) is normalized away during
          * the creation of a Path object reference.
-         * If the URI is different then the Path.toUri() then
+         * If the URI is different from the Path.toUri() then
          * we will just use the original URI to construct the
          * alias reference Path.
          */
-        if (!URIUtil.equalsIgnoreEncodings(uri, path.toUri()))
+        if (!isJarArchive && !URIUtil.equalsIgnoreEncodings(uri, path.toUri()))
```
gregw: I don't understand why jars are being excluded.  Please explain in the comment above.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -85,7 +87,7 @@ private Path checkAliasPath()
 
         try
         {
-            if (Files.isSymbolicLink(path))
+            if (!isJarArchive && Files.isSymbolicLink(path))
```
gregw: please explain in a comment

gregw (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: 1. When a path in a `jar:file:` is a directory, the URI produced will not end in a slash.
2. There is no support for symlinks in zipfs, so those tests are moot and can be skipped.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: We have test cases for this now. See point 1 in comment above. (and in code)

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: See point 2 in comment above.  (and in code)

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/PathResourceTest.java
```diff
@@ -140,4 +158,204 @@ public void testSame()
...
+
+        Map<String, String> env = new HashMap<>();
+        env.put("create", "true");
+
+        URI jarUri = URIUtil.uriJarPrefix(testJar.toUri(), "!/");
+        try (FileSystem zipfs = FileSystems.newFileSystem(jarUri, env))
+        {
+            Path root = zipfs.getPath("/");
+
+            Path file0 = root.resolve("test.txt\0");
```
joakime, line 270: On ZipFS, this `Path.resolve("test.txt\0")` results in a `InvalidPathException`...

```
java.nio.file.InvalidPathException: Path: nul character not allowed: test.txt?
	at jdk.zipfs/jdk.nio.zipfs.ZipPath.normalize(ZipPath.java:501)
	at jdk.zipfs/jdk.nio.zipfs.ZipPath.normalize(ZipPath.java:474)
	at jdk.zipfs/jdk.nio.zipfs.ZipPath.normalize(ZipPath.java:516)
	at jdk.zipfs/jdk.nio.zipfs.ZipPath.resolve(ZipPath.java:370)
	at jdk.zipfs/jdk.nio.zipfs.ZipPath.resolve(ZipPath.java:56)
	at org.eclipse.jetty.util@12.0.0-SNAPSHOT/org.eclipse.jetty.util.resource.PathResourceTest.testNullCharEndingFilename(PathResourceTest.java:299)
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -50,15 +50,25 @@ private Path checkAliasPath()
     {
         Path abs = path;
 
+        // Boolean to track if Path is a ZipFS based
+        boolean isZipFs = uri.toASCIIString().startsWith("jar:file:");
```
gregw: If zipFS is to have exceptions and/or different handling, then let's create a `ZipPathResource` class and remove these conditionals from the `PathResource` class

gregw (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: Nah, not needed.
I got rid of that boolean/test with the latest code.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
lorban: @gregw `MountedPathResource` actually is managing only mounted JAR resources. I tried at first to make it more generic but that proved useless as the implementation progressed. So we already have the equivalent to that `JarPathResource` and I'm not against renaming it.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -694,29 +694,60 @@ public static String decodePath(String path, int offset, int length)
         }
     }
 
+    /**
+     * Test if codepoint is safe to decode for URI
```
gregw:
```suggestion
     * Test if codepoint is safe and unamigous to be given to parsed by the URI.create method
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -694,29 +694,60 @@ public static String decodePath(String path, int offset, int length)
...
+     *
+     * @param code the codepoint code to test
+     * @return true if safe to decode, otherwise false;
```
gregw, line 704:
```suggestion
     * @return true if the decoded value is safe to pass as input to URI, otherwise false;
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -1642,20 +1677,45 @@ else if (!uriA.getAuthority().equals(uriB.getAuthority()))
...
+     * Add a sub path to an existing URI.
+     *
      * @param uri A URI to add the path to
      * @param path A decoded path element
+     * @param encodePath true to encode provided path, false to leave it alone in resulting URI
      * @return URI with path added.
+     * @see #addPaths(String, String)
      */
-    public static URI addPath(URI uri, String path)
+    public static URI addPath(URI uri, String path, boolean encodePath)
```
gregw: I'd prefer 2 methods rather than one with a boolean arg

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -1642,20 +1677,45 @@ else if (!uriA.getAuthority().equals(uriB.getAuthority()))
...
     {
-        String base = uri.toASCIIString();
-        StringBuilder buf = new StringBuilder(base.length() + path.length() * 3);
+        Objects.requireNonNull(uri, "URI");
+
+        if (path == null)
+            return uri;
+
+        // collapse any "//" paths in the path portion
+        path = compactPath(path);
```
gregw, line 1794: I wonder if we really should do this.... but leave it for now.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -1642,20 +1677,45 @@ else if (!uriA.getAuthority().equals(uriB.getAuthority()))
...
+     * @see #addPaths(String, String)
      */
-    public static URI addPath(URI uri, String path)
+    public static URI addPath(URI uri, String path, boolean encodePath)
     {
-        String base = uri.toASCIIString();
-        StringBuilder buf = new StringBuilder(base.length() + path.length() * 3);
+        Objects.requireNonNull(uri, "URI");
+
+        if (path == null)
```
gregw: or if path is just "" or "/" return uri as well

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -52,13 +52,13 @@ private Path checkAliasPath()
...
-         * from an #addPath(String) is normalized away during
+         * from an #resolve(String) is normalized away during
          * the creation of a Path object reference.
-         * If the URI is different then the Path.toUri() then
+         * If the URI is different from the Path.toUri() then
          * we will just use the original URI to construct the
          * alias reference Path.
          */
-        if (!URIUtil.equalsIgnoreEncodings(uri, path.toUri()))
+        if (!URIUtil.equalsIgnoreEncodings(uri, normalize(path)))
```
gregw: why do we need the expense of normalization here? `equalsIgnoreEncoding` already handles the /// differences.
What case is normalization catching?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -1545,7 +1576,8 @@ public static void appendSchemeHostPort(StringBuffer url, String scheme, String
         }
     }
 
-    public static boolean equalsIgnoreEncodings(String uriA, String uriB)
+    // Only URIUtil is using this method
+    static boolean equalsIgnoreEncodings(String uriA, String uriB)
```
gregw, line 1588: I don't see a fix in this method for handling multi-byte encodings.
Perhaps the changes in the URI method might avoid calling this method with such encoding differences, but we should still fix this method to avoid future bugs

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -1614,6 +1646,9 @@ public static boolean equalsIgnoreEncodings(URI uriA, URI uriB)
         if (uriA.equals(uriB))
             return true;
 
+        if (uriA.toASCIIString().equals(uriB.toASCIIString()))
```
gregw, line 1747: when does this return true, but the condition above doesn't?

gregw (dismissed):
A few things need fixing.

Overall, I'm really concerned about the amount of garbage that we are going to create as we frequently convert between Strings, URIs and Paths, then make corrected versions of them then throw them away.      But let's go for correctness first I guess.

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/pathmap/ServletPathSpecOrderTest.java
```diff
@@ -81,7 +81,7 @@ public static Stream<Arguments> data()
         mappings.put(new ServletPathSpec("/"), "default"); // 8
         // 9 was the old Jetty ":" spec delimited case (no longer valid)
         mappings.put(new ServletPathSpec(""), "root"); // 10
-        mappings.put(new ServletPathSpec("/\u20ACuro/*"), "money"); // 11
+        mappings.put(new ServletPathSpec("/%E2%82%ACuro/*"), "money"); // 11
```
lachlan-roberts: The test currently doesn't do a `canonicalPath` before trying the match so this won't work.

The registering of the mapping does do a `canonicalPath` in the constructor of `ServletPathSpec`.

lachlan-roberts (changes requested):
Few tests are now failing. 

If the resource is an alias using `Resource.getPath()` will throw `NoSuchFileException` when used, but the `Resource.exists()` returns true. So everywhere it is used you would have to use a pattern like `resource.isAlias() ? resource.getAlias() : resource.getPath()`, so maybe instead `resource.getPath()` should return the alias path?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: Updated

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: Updated

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: I'll just get rid of the boolean.
There's nothing in mainline code right now using it as true anyway.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: Doing if `"".equals(path)` is fine, but can't do `"/".equals(path)` as that would break usages like ...

```
URIUtil.addPath(URI.create("file:///tmp/aaa"), "/")
// Expected to be:  file:///tmp/aaa/
```

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/pathmap/ServletPathSpecOrderTest.java
joakime: Lets fix that in a different PR.
Meanwhile, I'll revert this change in `ServletPathSpecOrderTest` (i don't think it's needed anymore)

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: The `normalize(Path)` corrects ...

* `file:/path/to/foo.dat` to `file:///path/to/foo.dat`
* `jar:file:/path/to/foo.jar!/bar.dat` to `jar:file:///path/to/foo.jar!/bar.dat`
* `file:///path/to/dir` to `file:///path/to/dir/`

So that the `equalsIgnoreEncoding` is happy.
While the first 2 corrections are now handled by `equalsIgnoreEncoding` the last one isn't

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: The `equalsIgnoreEncoding(String, String)` is not used outside of `equalsIgnoreEncoding(URI, URI)`, so it's essentially protected by the checks already done there.

I started down the path of writing support for multi-byte UTF-8 sequences properly in `equalsIgnoreEncoding(String, String)` and quickly discovered that we would have to decode the entire input `String` into a `Utf8StringBuilder` properly, and not check character by character like we do now in this implementation.

This is because we can easily have multiple multi-byte sequences in a row.

Examples

```
// Expecting true as a result
URIUtil.equalsIgnoreEncoding("/path/to/bä€ãm/", "/path/to/b%C3%A4%E2%82%AC%C3%A3m/");

// Expecting false as a result
URIUtil.equalsIgnoreEncoding("/path/to/bãm/", "/path%2Fto/b%C3%A3m/");
URIUtil.equalsIgnoreEncoding("/path/bãm/or/bust", "/path/b%C3%A3m/or%2Fbust");
URIUtil.equalsIgnoreEncoding("/path/to/bä€ãm/", "/path%2Fto/b%C3%A4%E2%82%AC%C3%A3m/");
```

At first, I thought this implementation would literally turn into something similar to ...

``` java
    static boolean equalsIgnoreEncodings(String uriA, String uriB)
    {
        return canonicalPath(uriA).equals(canonicalPath(uriB));
    }
```

But that would strip path params and also decode `%2F` sequences, so I didn't do that.

As a result of this test, I added a testcase for path params in the false tests.

Can't use `canonicalPath(String)` either, as that uses the `isSafe(int)` checks for encoding characters (or not), I need the reverse.

To implement this properly, I would need to decode everything that is safe to decode.
Oh-kay, i'll implement a `URIUtil.decodeSafe(String)`

joakime: Turned out to be `String URIUtil.ensureSafeEncoding(String)` instead (with test cases)

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: Example: both `uriA` and `uriB` are opaque.
If `uriA` has a raw unicode, and `uriB` is utf-8 percent encoded of the same unicode.

The comparison in `URI.equals(URI)` will perform a simple string comparison against `URI.getRawSchemeSpecificPart()` of both.
This `URI.equals(URI)` comparison will fail.
But the `uriA.toASCIIString().equals(uriB.toASCIIString())` will return true.

joakime: An example that doesn't pass `uriA.equals(uriB)`, but does pass `uriA.toASCIIString().equals(uriB.toASCIIString())`

```
uriA = "jar:file:///path/to/test.jar!/bãm/"
uriB = "jar:file:///path/to/test.jar!/b%C3%A3m/"
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
gregw: Is 2nd correction correct?

joakime: Good catch, typo.
Edited the comment.

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/pathmap/ServletPathSpecOrderTest.java
joakime: Reverted in commit a40002fb11db739501dd6c4bebad930f7bc601c8

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: The prior response is now a bit out of date, the `normalize(Path)` method doesn't do the `file:/` behavior, only the trailing slash behavior now in this PR.

gregw: But back to my question... Why normalize a temporary URI when equalsIgnoreEncoding doesn't notice /// or /?  It compares the authority and not the URI string.

joakime: The old `normalize(Path)` did the `file:/...` to `file:///`.

The new `normalize(Path)` doesn't modify the `file:///` portion anymore, as the `URI.equals(URI)` check is capable of evaluating that successfully.

Both the old and new `normalize(Path)` still enforces the trailing slash on the URI if the path is a directory.
This is done to match up with the forced trailing slash done in the `PathResource(URI)` constructor.

joakime: This trailing slash behavior depends on the Path existing and being a directory.

One quirk of this behavior (and not the only one) ...
If the Resource.uri has `.../Unknown/` as a result of `Resource.resolve("Unknown/")` and the directory doesn't exist, then this results in a case where the input URI has a trailing slash, and the `path.toUri()` doesn't, even with the `normalize(Path)`.

gregw: I still don't understand why we are normalizing here.   That evades aliases rather than detects them.

joakime: I think you are getting caught up on the word "normalize" here.

Let me rename it, as that's NOT what it's doing.

joakime: Renamed. with more comments about it's use in commit dce2da81daaa6b7a535093f8865f267deb1d9fc0

lachlan-roberts:
@joakime  Can you look at the disabled test `org.eclipse.jetty.ee10.servlet.DefaultServletTest#testListingFilenamesOnlyUrlResource` in relation to the changes in this PR. 

Is using `jar:file` URIs and in `12.0.x` was failing because it was being classified as an alias because of trailing slash. But now on this PR the test it is throwing `UnsupportedOperationException`.

```
java.lang.UnsupportedOperationException
	at jdk.zipfs/jdk.nio.zipfs.ZipPath.toFile(ZipPath.java:669)
	at org.eclipse.jetty.server/org.eclipse.jetty.server.ResourceListing.getAsHTML(ResourceListing.java:207)
	at org.eclipse.jetty.server/org.eclipse.jetty.server.ResourceService.sendDirectory(ResourceService.java:530)
	at org.eclipse.jetty.server/org.eclipse.jetty.server.ResourceService.sendWelcome(ResourceService.java:438)
	at org.eclipse.jetty.server/org.eclipse.jetty.server.ResourceService.doGet(ResourceService.java:166)
	at org.eclipse.jetty.ee10.servlet/org.eclipse.jetty.ee10.servlet.DefaultServlet.doGet(DefaultServlet.java:388)
	at jakarta.servlet@6.0.0/jakarta.servlet.http.HttpServlet.service(HttpServlet.java:527)
	at jakarta.servlet@6.0.0/jakarta.servlet.http.HttpServlet.service(HttpServlet.java:614)
	at org.eclipse.jetty.ee10.servlet/org.eclipse.jetty.ee10.servlet.ServletHolder.handle(ServletHolder.java:742)
```

lorban:
This is related to https://github.com/eclipse/jetty.project/issues/8462 as well.

lachlan-roberts (approved):

Issue #8426, opened 2022-08-08
Jetty 12 - Review and fix Alias Checking


lachlan-roberts:
Jetty 12 alias checking needs to be reviewed and fixed.

AliasCheck is an interface defined in jetty-core but is only used in the ee10 servlet ContextHandler and is not used in the core ContextHandler, which also has no way to define protected targets.

joakime:
@lachlan-roberts the `jetty-12.0.x` branch build was red/broken as a result of your last merge (commit 19e2e7babc0b2d22253123cf237051b5b0046af9)

I went ahead and reverted a small hack you had in your last merge.
See commit e7e570f4bbab151729644dde580305f79ed6d2db

This brought the build back to green.

lorban:
While you're a it, I think there is a bug in ee9 `ContextHandler.checkAlias`: if the resource is an alias but there are no alias checks then `false` is returned while I think `true` should be returned.

gregw:
@lachlan-roberts can we add a test for that.

lachlan-roberts:
> While you're a it, I think there is a bug in ee9 ContextHandler.checkAlias: if the resource is an alias but there are no alias checks then false is returned while I think true should be returned.

This is not a bug, we do not allow aliases unless there is an alias check in place to specifically allow them. We have even deprecated `ApproveAliases` (which always returns true) and warn if it is used because using it would introduce security issues if you used it without a very good reason.

lorban:
Something doesn't feel right, he's why I believe `ContextHandler.checkAlias` is buggy in its current form: if one creates a ee9 `DefaultServlet` to serve the contents of a JAR file then it's not possible to serve the directory listings of any subfolder of the jar because `jar:file:///some/file.jar!/foo/` is considered an alias of `jar:file:///some/file.jar!/foo`. Maybe @joakime has some insights to share about this?

I've created the https://github.com/eclipse/jetty.project/commits/jetty-12.0.x-enable-ee9-demo-embedded branch to illustrate the problem I found.

I added this test that configures the ee9 DefaultServlet against a jar file's contents as it base resource: https://github.com/eclipse/jetty.project/blob/jetty-12.0.x-enable-ee9-demo-embedded/jetty-ee9/jetty-ee9-demos/jetty-ee9-demo-embedded/src/test/java/org/eclipse/jetty/ee9/demos/JarServerTest.java

[...]

PR #8449, opened 2022-08-11
Issue #8426 - Review of alias checks for Jetty 12 

lachlan-roberts:
## Issue #8426

- ee9 and ee10 servlets should use the core alias checking mechanism
- do the alias check in the resourceService so it is not done on programmatic calls to getResource()


lachlan-roberts:
Still has some test failures which I need to investigate.

tests/test-integration/src/test/resources/altDir1/file1
```diff
@@ -0,0 +1 @@
+file 1 contents
```
gregw, line 1: no EOLN on this file and several below

gregw (approved):
LGTM

Issue #8443, opened 2022-08-10
Jetty 12: review `resolve()` and `getPath()` contracts in `ResourceCollection` and `MemoryResource`


lorban:
**Jetty version(s)**
12

**Description**
Review how `resolve()` and `getPath()` are supposed to behave when directory and non-directory resources are part of a `ResourceCollection`, and for `MemoryResource` in general. Or if we want those `Resource` implementations at all.

lorban:
My personal opinion is that neither `resolve()` nor `getPath()` should ever return null, but I'm open to different opinions.

gregw:
If we could arrange for `getPath()` to never return null, then we would not need the resource class and could just use `Path`.    However, that fails on the `ResourceCollection` issue that we are required to provide aggregate base resources and to do so without necessarily extracting.

However, we could have that a non-directory resource would always return a path.

Having resolve never return null is doable, but it may be that we would need to return an empty or bad resource from some resolves (or /dev/null if we must have a path).

PR #8451, opened 2022-08-11
MemoryResource

gregw:
Added MemoryResource to simplify favicon handling.
Using the ResourceFactory mechanism results in an entire jar mounted as a ZipFS in memory that is only needed for favicon.ico

### DIFF
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/MemoryResource.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/MemoryResource.java
@@ -1,0 +1,117 @@
+//
+// ========================================================================
+// Copyright (c) 1995-2022 Mort Bay Consulting Pty Ltd and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.util.resource;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.nio.channels.Channels;
+import java.nio.channels.ReadableByteChannel;
+import java.nio.file.Path;
+import java.util.Objects;
+
+import org.eclipse.jetty.util.IO;
+
+/**
+ * <p>An in memory Resource created from a {@link URL}</p>
+ */
+public class MemoryResource extends Resource
+{
+
+    private final URI _uri;
+    private final long _created = System.currentTimeMillis();
+    private final byte[] _bytes;
+
+    public MemoryResource(URL url)
+    {
+        try
+        {
+            _uri = Objects.requireNonNull(url).toURI();
+            _bytes = IO.readBytes(url.openStream());
+        }
+        catch (IOException | URISyntaxException e)
+        {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public Path getPath()
+    {
+        return null;
+    }
+
+    @Override
+    public boolean isContainedIn(Resource r)
+    {
+        return false;
+    }
+
+    @Override
+    public URI getURI()
+    {
+        return _uri;
+    }
+
+    @Override
+    public String getName()
+    {
+        return _uri.getPath();
+    }
+
+    @Override
+    public long lastModified()
+    {
+        return _created;
+    }
+
+    @Override
+    public long length()
+    {
+        return _bytes.length;
+    }
+
+    @Override
+    public InputStream newInputStream() throws IOException
+    {
+        return new ByteArrayInputStream(_bytes);
+    }
+
+    @Override
+    public ReadableByteChannel newReadableByteChannel() throws IOException
+    {
+        return Channels.newChannel(newInputStream());
+    }
+
+    @Override
+    public Resource resolve(String subUriPath)
+    {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public boolean exists()
+    {
+        return true;
+    }
+
+    @Override
+    public String toString()
+    {
+        return getName();
+    }
+}
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/MemoryResource.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/MemoryResource.java
@@ -31,17 +31,19 @@ import org.eclipse.jetty.util.IO;
  */
 public class MemoryResource extends Resource
 {
-
     private final URI _uri;
     private final long _created = System.currentTimeMillis();
     private final byte[] _bytes;
 
-    public MemoryResource(URL url)
+    MemoryResource(URL url)
     {
         try
         {
             _uri = Objects.requireNonNull(url).toURI();
-            _bytes = IO.readBytes(url.openStream());
+            try (InputStream in = url.openStream())
+            {
+                _bytes = IO.readBytes(in);
+            }
         }
         catch (IOException | URISyntaxException e)
         {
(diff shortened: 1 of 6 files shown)


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/MemoryResource.java
```diff
@@ -0,0 +1,117 @@
...
+    @Override
+    public ReadableByteChannel newReadableByteChannel() throws IOException
+    {
+        return Channels.newChannel(newInputStream());
+    }
+
+    @Override
+    public Resource resolve(String subUriPath)
+    {
+        throw new UnsupportedOperationException();
```
lorban: `resolve()` should never throw, otherwise we're going to have to add try / catch blocks everywhere we call it. I'm not even sure we should let it return `null`, so it feels like the right thing to do here is to return `this`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/MemoryResource.java
```diff
@@ -0,0 +1,117 @@
...
+        catch (IOException | URISyntaxException e)
+        {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public Path getPath()
+    {
+        return null;
```
lorban: So far, code has been written that assumes `Resource.getPath()` cannot return null, and maybe there's some code that still calls `File.newInputStream(resource.getPath())` instead of `resource.getInputStream()`.

I think we should arrange for this implementation of `getPath()` to never return null, maybe by creating a simple in-memory filesystem?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/MemoryResource.java
```diff
@@ -0,0 +1,117 @@
...
+ * <p>An in memory Resource created from a {@link URL}</p>
+ */
+public class MemoryResource extends Resource
+{
+
+    private final URI _uri;
+    private final long _created = System.currentTimeMillis();
+    private final byte[] _bytes;
+
+    public MemoryResource(URL url)
```
lorban: No resource should have a public constructor IMHO. So far they're all factored with `ResourceFactory` or `Resource.combine()` (the latter should be moved to `ResourceFactory`).

Adding `MemoryResource loadToMemory(Resource r)` would allow us to write:
```java
MemoryResource favicon;
try (ResourceFactory.Closeable rf = ResourceFactory.closeable())
{
    favicon = ResourceFactory.loadToMemory(rf.newResource(favIconUrl));
}
```

which could be wrapped in a `MemoryResource loadToMemory(URL url)` helper:
```java
MemoryResource favicon = ResourceFactory.loadToMemory(favIconUrl);
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/MemoryResource.java
```diff
@@ -0,0 +1,117 @@
...
+    {
+        try
+        {
+            _uri = Objects.requireNonNull(url).toURI();
+            _bytes = IO.readBytes(url.openStream());
```
lorban: You're leaking a file descriptor.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/MemoryResourceTest.java
```diff
@@ -0,0 +1,34 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class MemoryResourceTest
+{
+    @Test
+    public void testJettyLogging() throws Exception
+    {
+        Resource resource = new MemoryResource(Loader.getResource("jetty-logging.properties"));
+        assertTrue(resource.exists());
+        String contents = IO.toString(resource.newInputStream());
```
lorban, line 31: You're leaking a file descriptor.

lorban (changes requested):
I like the idea, but I'm very concerned by this implementation's shortcomings. They're solvable, so I think we should not cut corners.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/MemoryResource.java
joakime: > maybe by creating a simple in-memory filesystem?

heh, uh, this is a BIG undertaking.

joakime:
Whatever you decide for `favicon.ico` should also be done for `jetty-dir.css` they exist in the same jar and have the same basic issues you are attempting to solve here.

The way it's handled in Jetty 12 now is that those exist as mounts on the Server object, which already has the `jetty-server-<ver>.jar` mounted for other things. (it gets accessed for every webapp for the bytecode scanning steps for example)

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/MemoryResource.java
gregw: @lorban Well every time I see anybody saying that `getPath` cannot return null, I've said "oh yes it can".   If getPath could always return non-null then we would have succeeded in replacing Resource with Path as there would be a 1:1 mapping.   But there is not.     ResourceCollection is one such resource that cannot always return a Path.

I've been on the lookout for code that makes that kind of assumption (eg getting the Path and using it to get last modified, rather than just asking the resource).   We need to remove such assumptions. We should javadoc that getPath can return null.

gregw: > > maybe by creating a simple in-memory filesystem?
> 
> heh, uh, this is a BIG undertaking.

Yeah - that's why I want this to be just a single resource - no lists, no directories, no resolve.

gregw:
> Whatever you decide for `favicon.ico` should also be done for `jetty-dir.css` they exist in the same jar and have the same basic issues you are attempting to solve here.

I'm doing this for `Server.newResource`, so we can change the mechanism. I think the stylesheet already uses that same mechanism.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/MemoryResource.java
gregw: I believe IO closes it, by I'll use a try with resources.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/MemoryResource.java
gregw: I think `this` is the wrong thing to return.
We don't want these resources slipping into general usage (not unless we implement an in memory filesystem... and hey that is what ZipFS is), so I'm OK with an unchecked exception.   If one of these resources ever ends up being used for a resolve, then that is a bad state and should fail.

gregw: It has a URI, so I'm now just defering to the default resolve.  You wont get back a memory resource, but that is fine.

joakime:
Wouldn't it be easier to copy those to the work dir and serve them via a normal file?

gregw:
> Wouldn't it be easier to copy those to the work dir and serve them via a normal file?

Potentially, but remember our start times are important as well.  We don't want to mount a zip file, extract favicon, unmount it, mount it again, extract style sheet, unmount etc.  So that ends up with the result we want but will take a lot of CPU cycles at start time.

These are classloader resources already available via the classloader getResource mechanism... unfortunately via a URL.  So we just want to quickly grab the contents via the existing JVM mechanism and make them available.

joakime:
The unmount only occurs at server.stop() for those resources

gregw:
@joakime if we use the zipFS bases mechanism to get these resources, then either we leave it mounted whilst running (wasting memory) or we unmount after extracting (wasted CPU).   There is already a perfectly good (modulo URL API) mechanism to get classpath resources.  We should use that and not re-crack open the jar file.      The zipFS mechanism behind Resource is great and we should use it... but only when appropriate (eg the deployer has asked to not unpack a war file).   We should not use it when a better mechanism already exists.

gregw:
@joakime @lorban I'm inclined to merge this, even if not perfect.  The current favicon stuff has been breaking the build for a few days due to mount leaks, so this will at least avoid that problem.

lorban:
This is adequate for now, but we should come up with an agreement for the long term. That discussion should happen in https://github.com/eclipse/jetty.project/issues/8443.

Issue #8442, opened 2022-08-10
Jetty 12 : `Handler.Collection` does not propagate `Server` to handlers in expected way


joakime:
**Jetty version(s)**
Jetty 12

**Description**

In Jetty 12, if you use the jetty-core `org.eclipse.jetty.server.Handler$Collection` the `.setHandlers(List)` does not propagate the `Server` correctly.

Example code.

``` java
WebAppContext context = new WebAppContext();
context.setContextPath("/");
context.setBaseResource(context.getResourceFactory().newResource(staticBase));
server.setHandler(new Handler.Collection(context, new DefaultHandler()));
server.start();
```

The `DefaultHandler.setServer(Server)` never gets called with a valid Server instance.

This is because the internal `Handler.Collection` processing does it's `handler.setServer(server)` far too early.
AFAICT, this seems to be the only Handler implementation with this bug.

gregw:
fixed by a0d6f42

lorban:
@gregw sorry, I've had to revert these changes in https://github.com/eclipse/jetty.project/pull/8452 because they broke the build.

gregw:
@lorban no probs.... any hint to how the build was broken?

gregw:
This was fixed during Handler restructure in 12.

PR #8452, opened 2022-08-11
Revert #8442 changes

lorban:
The changes in #8442 broke the build.


gregw:
I think the broken build is unrelated to #8442.  I think it is the favicon stuff that #8451 fixes

gregw:
I don't think this is needed now

PR #8454, opened 2022-08-12
Jetty 12.0.x module renaming

sbordet:


gregw (approved):
Demos work (or as well as they did before)

PR #8457, opened 2022-08-12
restore mount leak detection

lorban:


gregw (approved):
I'm cautious about this, as it broke the build for several days... but it has a green build now, so let's give it a go again and see...

lorban:
The mechanism was detecting a genuine leak that you fixed, this is why re-enabling it keeps the build green.

PR #8458, opened 2022-08-12
Code cleanup, removing deprecations, for class Pool,

sbordet:
ConnectionPool and its implementations.

Removed DuplexHttpDestination and MultiplexHttpDestination.


jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/AbstractConnectionPool.java
```diff
@@ -121,24 +113,25 @@ public void setMaxDuration(long maxDurationInMs)
...
 
-    protected void setMaxMultiplex(int maxMultiplex)
+    protected void setInitialMaxMultiplex(int initialMaxMultiplex)
     {
-        pool.setMaxMultiplex(maxMultiplex);
+        this.initialMaxMultiplex = initialMaxMultiplex;
     }
 
-    protected int getMaxUsageCount()
+    @ManagedAttribute(value = "The maximum amount of times a connection is used before it gets closed")
```
lorban, line 135: The fact that it defaults to 0 and that `0-` values disable the mechanism should be documented.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
```diff
@@ -177,163 +180,29 @@ public int getMaxEntries()
...
-
     /**
      * <p>Retrieves the max multiplex count for the given pooled object.</p>
      *
      * @param pooled the pooled object
      * @return the max multiplex count for the given pooled object
-     * @deprecated Multiplex functionalities will be removed
      */
-    @Deprecated
     protected int getMaxMultiplex(T pooled)
```
lorban: Since you added a constructor that takes a `ToIntFunction`, there is no need to keep this method as protected anymore, it can be made private.

lorban (approved):
LGTM except for a couple niggles.

Issue #8456, opened 2022-08-12
Fix ee10 and ResourceHandler to make them serve jetty-dir.css resource


lorban:
**Jetty version(s)**
12

**Description**

Here's the current state about which of the static content serving mechanisms (ee9 `DefaultServlet`, ee10 `DefaultServlet` and `ResourceHandler`) can serve jetty-dir.css:

|                            |  jetty-dir.css |
| ---------------------- |  --------------- |
| ee9                      | Y                 |
| ee10                    | N                |
| resource handler | N                |

As a side note, by design, `DefaultHandler` is meant to serve favicon while `ResourceHandler` and both `DefaultServlet`s are supposed to serve only jetty-dir.css.

lachlan-roberts:
done

PR #8463, opened 2022-08-15
Issue #8456 - serve jetty-dir.css as resource in jetty-core ResourceService

lachlan-roberts:
## Issue #8456

For the jetty-core `ResourceService` now has special handling for the `jetty-dir.css`. This is set on the `ResourceService` through API on the `ResourceHandler` and the ee10 `DefaultServlet`. The ee9 code does not use this mechanism as it uses the old `ResourceService` and has logic on the ee9 `ResourceHandler` and `DefaultServlet` to get servers default and set custom values for the stylesheet.

Similar PR to #8428


gregw:
Does this replace #8428?

gregw (approved):
LGTM

lachlan-roberts:
Test failure is one pre-existing in 12.0.x branch.

lachlan-roberts:
> Does this replace https://github.com/eclipse/jetty.project/pull/8428?

Yes but that PR does things in a different way by using `ResourceCollection`, so maybe that is preferable for some reason, but I will merge this as it works and leave 8428 open for now.

PR #8465, opened 2022-08-15
Fixes #8007 - Support Loom.

### LINKED ISSUE #8007: Support Loom

sbordet:
Now using Executors.newVirtualThreadPerTaskExecutor() to execute
tasks, so the executor is tracked by the runtime for thread dumps, etc.


sbordet:
@AlanBateman I went on and used an `Executor` as you suggested.

I did not use `Runtime.version().feature()` as I still need that bit of reflection and I have to actually invoke a preview API that would throw if `--enable-preview` is missing, so the test on whether I could get the executor should be enough.

gregw (approved):
nicer

PR #8470, opened 2022-08-16
Jetty 12 : Simplify favicon.ico and jetty-dir.css behaviors

joakime:
* Remove `Server.getDefaultFavicon()`
* Remove `Server.getDefaultStylesheet()`
* Remove `MemoryResource`
* `favicon.ico` is only managed by the `DefaultHandler` now
* `DefaultHandler` loads favicon as ByteBuffer now
* Introduce new `ResourceService` APIs.
  - `void .loadStylesheet(Resource)` - loads an arbitrary resource to be the contents of all `jetty-dir.css` requests
  - `void .loadDefaultStylesheetIfNotPresent()` - loads default stylesheet if not already present (done because ResourceService is not a LifeCycle)
  - `boolean .isStylesheetRequest(String pathInContext)` - to test if pathInContext is for a the stylesheet
  - `boolean .sendStylesheet(String pathInContext, Request request, Response response, Callback callback)` - to send stylesheet if appropriate
* Add Stylesheet testing in EE10 `DefaultServletTest`
* Add Stylesheet testing in EE9/EE8 `DefaultServletTest`
* Add Stylesheet testing in core `ResourceHandler`


jetty-core/jetty-server/src/main/java/module-info.java
```diff
@@ -34,4 +34,9 @@
 
     // TODO required for testing ????
     exports org.eclipse.jetty.server.ssl;
+
+    // expose jetty-dir.css to ee9/ee8
+    opens org.eclipse.jetty.server to
```
joakime, line 39: This can be removed once ee9/ee8 uses the core ResourceService.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -77,26 +83,73 @@
...
+    public void loadDefaultStylesheetIfNotPresent()
     {
-        return _stylesheet;
+        if (_stylesheetBuffer != null)
+            return;
+
+        URL url = ResourceService.class.getResource("jetty-dir.css");
+        if (url == null)
+        {
+            LOG.warn("Unable to find resource: %s/jetty-dir.css".formatted(TypeUtil.toClassReference(ResourceService.class.getPackageName())));
```
joakime, line 138: This can happen with badly uber-jar'd projects.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -191,6 +189,7 @@ public void init() throws ServletException
                     LOG.warn("Unable to use stylesheet: {} - {}", stylesheetParam, e.toString());
             }
         }
+        _resourceService.loadDefaultStylesheetIfNotPresent();
```
joakime, line 192: If `ResourceService` was a `LifeCycle`, this could be done in the `doStart()` method instead.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -77,26 +83,73 @@
...
+        byte[] buf = Files.readAllBytes(path);
+        _stylesheetBuffer = ByteBuffer.wrap(buf);
     }
 
     /**
-     * @return Returns the stylesheet as a Resource.
+     * Load the default Stylesheet if one has not been previously with a call to {@link #loadStylesheet(Resource)}
      */
-    public Resource getStylesheet()
+    public void loadDefaultStylesheetIfNotPresent()
```
lachlan-roberts, line 130: Could this just call the other `loadStyleSheet` method?

```java
if (_stylesheetBuffer != null)
    return;

try
{
    URL url = ResourceService.class.getResource("jetty-dir.css");
    if (url == null)
    {
        LOG.warn("Unable to find resource: %s/jetty-dir.css".formatted(TypeUtil.toClassReference(ResourceService.class.getPackageName())));
        return;
    }
    
    loadStylesheet(ResourceFactory.root().newResource(url));
}
catch (IOException e)
{
    LOG.warn("Unable to load default stylesheet", e);
}
```

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -338,9 +337,24 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
         {
-            HttpContent content = _resourceService.getContent(pathInContext, ServletContextRequest.getBaseRequest(req));
+            ServletContextRequest servletContextRequest = ServletContextRequest.getBaseRequest(req);
+            HttpContent content = _resourceService.getContent(pathInContext, servletContextRequest);
             if (content == null || !content.getResource().exists())
             {
+                try (Blocker.Callback callback = Blocker.callback())
+                {
+                    ServletContextResponse servletContextResponse = ServletContextResponse.getBaseResponse(resp);
+                    if (_resourceService.sendStylesheet(pathInContext, servletContextRequest, servletContextResponse, callback))
```
lachlan-roberts, line 347: Why are we sending the stylesheet here?

lachlan-roberts (commented):
I don't like how the serving of the stylesheet is sometimes handled within `ResourceService` and sometimes outside of it.

But if it works it that's fine, just a few questions.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -77,26 +83,73 @@
...
     }
 
     /**
-     * @param stylesheet The location of the stylesheet to be used as a String.
+     * Load the stylesheet buffer
+     *
+     * @param stylesheet The location of the stylesheet to be used as-is.
      */
-    public void setStylesheet(Resource stylesheet)
+    public void loadStylesheet(Resource stylesheet) throws IOException
```
gregw, line 100: I this `setStyleSheet` is a better name.  The fact that it needs to be loaded to be used, is implementation detail.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
gregw, line 130: why is this needed as a public API?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -160,6 +213,15 @@ public void doGet(Request request, Response response, Callback callback, HttpCon
 
         try
         {
+            if (isStylesheetRequest(pathInContext))
```
gregw, line 216: Is this the right place for this? Perhaps we should first look for the real resource and only check if it is a stylesheet request if it misses.  This has better performance for other resource hits and allows the stylesheet to be overriden is specific directories.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -160,6 +213,15 @@ public void doGet(Request request, Response response, Callback callback, HttpCon
...
+            {
+                if (sendStylesheet(pathInContext, request, response, callback))
```
gregw, line 218: why is this a boolean return?  Do we expect failure to be frequent? Can't it just throw and let normal exception handling deal with it?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DefaultHandler.java
```diff
@@ -64,42 +65,23 @@ public class DefaultHandler extends Handler.Processor
...
-        catch (Exception e)
-        {
-            LOG.warn("Unable to find default favicon", e);
+            LOG.warn("Unable to find resource: %s/favicon.ico".formatted(TypeUtil.toClassReference(DefaultHandler.class.getPackageName())));
         }
-        finally
+        else
         {
-            _favicon = BufferUtil.toBuffer(favbytes);
+            try (InputStream in = url.openStream();
```
gregw, line 75: Isn't using URL going to use the internal URL cache that will load the jar into memory again and then not use it?
Can't we use getResourceAsStream instead?

gregw (changes requested):

joakime:
I'm closing this.

The recent proposed work for the HttpContent.Factory hierarchy would make this effort far easier.

I can see a `StaticHttpContentFactory` that sits after `ResourceContentFactory` and just returns a pre-configured `StaticHttpContent` for requests ending in `/jetty-dir.css`.
A similar process to how it works currently in `ResourceService`.

https://github.com/eclipse/jetty.project/blob/a3f226876669e31fd990a18db83452de09c86f6e/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java#L109-L124

PR #8471, opened 2022-08-17
Jetty 12 : `ResourceListing` produces XHTML (and is validated in test cases)

joakime:
Building from the work on PR #5142

The ResourcListing output is now XHTML compliant.
XML well formedness verified during testing.


jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
```diff
@@ -84,12 +94,12 @@ public static String getAsHTML(Resource resource, String base, boolean parent, S
         }
 
         // Perform sort
-        Comparator<? super Resource> sort = switch (sortColumn)
-        {
-            case "M" -> ResourceCollators.byLastModified(sortOrderAscending);
-            case "S" -> ResourceCollators.bySize(sortOrderAscending);
-            default -> ResourceCollators.byName(sortOrderAscending);
-        };
+        Comparator<? super Path> sort = switch (sortColumn)
```
gregw: Path cannot be used here as all directory entries mighty not have paths. A subdir might itself be a resource collection. Currently that is erroneously returning the path of the first item in the collection, that may have a totally unrelated name.

gregw (changes requested):

gregw (changes requested):
We have to stop externalizing methods that can be done on resource. If there is a good implementation in Path, then let PathResource provide that rather than externalizing behavior that not so resources may have!

joakime:
> We have to stop externalizing methods that can be done on resource. If there is a good implementation in Path, then let PathResource provide that rather than externalizing behavior that not so resources may have!

That API is fundamentally bad.

* It should not return String. (as that just causes another `newResource` to use the string properly)
* It should not return Resource. (as that causes a pointless allocation and even an alias lookup. For the 3 places that `.list()` was actually used, there's no need for any of that)
* It should not coalescence names that are the same (like the `ResourceCollection.list` does), as that eliminates names that just happen to be the same but are different locations to search for things (like libs)

It should return Path, and all Paths that are found.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
joakime: This comment is on a `Comparator`, what specific thing are you wanting to point out?

This is a `ResourceListing`, so it should only be presenting 1 level of information deep.

joakime: > Path cannot be used here as all directory entries mighty not have paths.

A `ResourceCollection` can have non-directory entries, correct.
But every one of those entries is a `PathResource` implementation, hence it has a `Path`.

I just added another test for this scenario (a `ResourceCollection` consisting of file, then dir, then a different dir)

gregw:
Don't tell me the api is bad. Fix the api. Just we tried replacing it with Path and that failed.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
```diff
@@ -36,24 +43,27 @@
...
+     * Convert the Resource directory into an XHTML directory listing.
      *
      * @param resource the resource to build the listing from
      * @param base The base URL
      * @param parent True if the parent directory should be included
      * @param query query params
      * @return the HTML as String
      */
-    public static String getAsHTML(Resource resource, String base, boolean parent, String query)
+    public static String getAsXHTML(Resource resource, String base, boolean parent, String query)
```
gregw, line 55: Good idea to make it xhtml

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
```diff
@@ -200,57 +210,125 @@ public static String getAsHTML(Resource resource, String base, boolean parent, S
...
+            return;
+
+        if (resource instanceof ResourceCollection resourceCollection)
+        {
+            for (Resource child: resourceCollection.getResources())
+            {
+                collectListing(listing, child);
+            }
+            return;
+        }
```
gregw: This is an anti-pattern.   We should not test to see if it is a ResourceCollection and then collect the listing differently.  We should just ask the resource for it's listing and we don't care how it is done or what type of resource it is.

This recursing for loop is exactly what is inside `ResourceCollection.list()`.  I agree that returning a `List<String>` is probably wrong, but a list of Resources would be fine.  If you are worried about allocation, then cache the sub-resources in the resource collection..... but creating resources should be cheap... even more so if done by a Resource method (e.g. it can use private constructors that bypass alias checking)

gregw: There is also a bug here.  If multiple resources in the resource collection have the same name in them, then you will get duplicate entries in the listing.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/ResourceListingTest.java
```diff
@@ -0,0 +1,276 @@
...
+import org.xml.sax.SAXException;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class ResourceListingTest
+{
+    @Test
+    public void testBasicResourceXHtmlListingRoot(WorkDir workDir) throws IOException
```
gregw, line 56: We already have directory listing tests in `ResourceHandlerTest`.  Let's not duplicate tests nor spread them in multiple places.    Either move all the listing tests here OR just test listings in the `ResourceHandlerTest`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
```diff
@@ -200,57 +210,125 @@ public static String getAsHTML(Resource resource, String base, boolean parent, S
...
+            {
                 name += URIUtil.SLASH;
+            }
 
             // Name
             buf.append("<tr><td class=\"name\"><a href=\"");
-            // TODO should this be a relative link?
-            String path = URIUtil.addEncodedPaths(encodedBase, URIUtil.encodePath(name));
-            buf.append(path);
+
```
gregw: You have deleted my TODO?  Do you know why we are using absolute rather than relative links here?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
```diff
@@ -200,57 +210,125 @@ public static String getAsHTML(Resource resource, String base, boolean parent, S
...
+            try
+            {
+                FileTime lastModified = Files.getLastModifiedTime(item, LinkOption.NOFOLLOW_LINKS);
+                buf.append(dfmt.format(new Date(lastModified.toMillis())));
+            }
+            catch (IOException ignore)
+            {
+                // do nothing (lastModifiedTime not supported by this file system)
+            }
```
gregw: This is exactly the kind of boiler plate that can be hidden inside a `Resource` class.  Besides, I assume the lastModifiedTime for a resource collection would be the latest date of all its contained resources, so again this fails because it assumes there is a 1:1 between resource and path.

gregw (dismissed):
More detailed review.  This is definitely a step in the wrong direction (other than the xhtml)

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
```diff
@@ -84,12 +94,12 @@ public static String getAsHTML(Resource resource, String base, boolean parent, S
...
-            case "S" -> ResourceCollators.bySize(sortOrderAscending);
-            default -> ResourceCollators.byName(sortOrderAscending);
-        };
+        Comparator<? super Path> sort = switch (sortColumn)
+            {
+                case "M" -> PathCollators.byLastModified(sortOrderAscending);
+                case "S" -> PathCollators.bySize(sortOrderAscending);
+                default -> PathCollators.byName(sortOrderAscending); // sorts by filename only (not full path)
+            };
         listing.sort(sort);
```
gregw: `Resource.list()` has already done a sort by the natural order.   We should not double sort.  Either list returns unsorted OR we optionally pass in a comparator to list.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
joakime: This doesn't assume ResourceCollection vs Resource, this is guaranteed to always be a Path at this point.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
joakime: This is how its has been since Jetty 7.x thru to Jetty 11.x.

Original change in our repo (for Jetty 7.0.0):

https://github.com/eclipse/jetty.project/commit/456a00161b47b8a3a13fc253c3a94ccacd69ea3b#diff-42810570fa3d8b562ad62d94a4ae857667aeb10b17f4ef9e93af43411af3d0aaR457

Current Jetty 11.0.x HEAD:

https://github.com/eclipse/jetty.project/blob/jetty-11.0.x/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java#L694

It's survived multiple updates and fixes since it was introduced.

To change from absolute links to relative links is possible, but out of scope for this PR. (don't want to evolve this to an omnibus PR again)
I'll reintroduce the TODO, but change the text to be more clear.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
gregw: There can be more than one path at this point. If multiple items in a resource collection contain the same sub directory, then the listed item should be a resource collection and the last modified date should be the most recent of its items.

sbordet: I agree with @gregw on the boilerplate. You don't want to remember, every single time you need a last modified time, to use the right `LinkOption` and lookup the date formatting APIs (why are we not using `java.time` here, but the old `java.text`?).

I also feel that the compositeness of `Resource` is still necessary, and would reduce other boilerplate.
For example, we don't want to rewrite all the times a recursive `collectListing()` (should it not be just shallow for directory listing?) every time we need children resources.

To me, `list()` has the wrong semantic. It should be `resource.getChildren()` and, only if really needed, `resource.getDescendants()`, which are easily implemented in both the "single" case and the "composite" case.
This would make crystal clear that `Resource` has a tree structure, and to display a directory listing you obviously want to call `getChildren()` (not `getDescendants()`), and then do the sorting externally (as dictated by the requirements of the directory listing request, e.g. query parameters).

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ResourceHandlerTest.java
```diff
@@ -2711,134 +2697,40 @@ public void testListingContextBreakout(ResourceHandlerTest.Scenario scenario) th
...
-
         /*
          * Intentionally bad request URI. Sending a non-encoded URI with typically
-         * encoded characters '<', '>', and '"'.
+         * encoded characters '<', '>', and '"', using the path-param feature of the
+         * URI spec to still produce a listing.  This path-param value should not make it
+         * down to the ResourceListing.getAsXHTML() method.
          */
+        // TODO: investigate why ResourceHandler isn't producing a directory listing when asked here.
+        // TODO: this request currently produces a 404
```
joakime: Fixing this is out of scope for this PR (we don't want omnibus PRs)

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
joakime: Ok, several things being discussed here.

_First, `lastModified` ..._

What would you see the API as?
`Resource.getLastModifiedMillis()` ?

The formatting of millis to date was not updated in this PR, this is the technique used since Jetty 7.0.0

See: https://github.com/eclipse/jetty.project/commit/456a00161b47b8a3a13fc253c3a94ccacd69ea3b#diff-42810570fa3d8b562ad62d94a4ae857667aeb10b17f4ef9e93af43411af3d0aaR471

If we want to use the new `java.time` APIs, this would be a good time to update that.

_Next, the proposed `Resource` listing APIs ..._

`Resource.getChildren()` returns what?
I would hope it's `Stream<Path>` as that's going to produce TONS less garbage than `List<Resource>` (and I can prove it in a jmh test).
Even if the API is `Stream<Resource>` it will produce less garbage overall when it's used.
Even the implementation on `ResourceCollection.getChildren()` could handle that.

If we cared about garbage, then the alias handling would be 100% optional and only trigger in the code paths that require it (currently only things using ResourceService or DefaultServlet).  It should not trigger for Configuration handling, or ResourceListing, or ServletContext.getResource(String).
It would have been a better solution to have `Alias Resource.getAlias()` as the only API for alias on `Resource`, and it only executes when asked for, returning the information needed by AliasChecker to evaluate if the resource is allowed or not.

The `Resource.getDescendants()` is not a good API for the modern era with webapps and/or resourceBases that have 10,000+ entries (super common, especially with spring-boot applications that use javascript libs).  It made sense in the past, but not today.

As a replacement for `Resource.getDescendants()`, how about a `List<Resource> Resource.find(Predicate predicate)` and let the implementations find what is being asked for.

_Next, the concept of a Resource a as a tree structure._

This is the source of many bugs, even in Jetty 7 thru Jetty 11.
We shouldn't have hidden `ResourceCollection` behind the `Resource` interface.
If the API can, possible, even remotely, be a collection, then it should indicate that in it's API, use the ResourceCollection object, not Resource.

Take, for example, the `Resource WebAppContext.getWebInf()`.
This API, at face value, can be a `ResourceCollection`, but in practice, it's assumed to not be one in many places in the Configuration flows.
Take this simple example, `EnvConfiguration` and it's `jetty-env.xml` (there are more if you look, this one is just easy to explain).
If we have a `ResourceCollection` with multiple `WEB-INF` directory entries (as returned by `WebAppContext.getWebInf()`), then it has the potential for returning multiple `jetty-env.xml` entries.    This entry is eventually passed to the `XmlConfiguration(Resource)` to load the XML.  `XmlConfiguration(Resource)` does not support `ResourceCollection` objects.

The current `ResourceCollection.resolve("jetty-env.xml")` punts on this and only returns the first one.   
The user has to be aware of this and ensure their collection order is sane for it to function in a predictable way.
But in reality, the `jetty-env.xml` is now a composite of multiple `jetty-env.xml` entries found potentially in many places. (just like web fragments)

This means `ResourceCollection.resolve(String)` should not return "first hit", but "all hits" in this case.
Or we should have a `ResourceColllection ResourceCollection.resolveAll(String)`, but that means being `ResourceCollection` aware again, as `Resource.resolveAll(String)` doesn't make sense.

When working with `Resource` and the possibility of if being a `ResourceCollection` and having to access the `Resource` it represents (read contents, last modified, size, etc) the API breaks again.   You have to unwrap the `Resource` if it's a `ResourceCollection` to get those APIs to function properly.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
joakime: The API for `Resource.list()` specifically states "Ordering is unspecified".

https://github.com/eclipse/jetty.project/blob/1c2290dca7a553a33933f33580c66cc0ac187925/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java#L265-L276

joakime:
Opened Issue to discussion review at #8397

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
gregw: This is no different to Path itself.  It can represent either a directory or a file (or other things).  If you use API in the wrong way with the wrong Path it "breaks". ie listing entries for a file or asking for the length of a directory.  This is nothing new.  It is a duality with File Descriptors, Files, Paths and Resources.    It is a very common pattern we just have to live  with.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
joakime: Reverting Path handling in this PR, in favor of using the work from PR #5142 
Keeping this PR's XHTML changes though.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
joakime: Reverting Path handling in this PR, in favor of using the work from PR #5142 
Keeping this PR's XHTML changes though.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ResourceHandlerTest.java
```diff
@@ -2784,9 +2790,26 @@ public void testListingWithQuestionMarks() throws Exception
...
+     * Tests to ensure that when requesting a legit directory listing, you
+     * cannot arbitrarily include XSS in the output via careful manipulation
+     * of the request path.
+     * </p>
+     * <p>
+     * This is mainly a test of how the raw request details evolve over time, and
+     * migrate through the ResourceHandler before it hits the
+     * ResourceListing.getAsXHTML for output production
+     * </p>
+     */    @Test
```
lorban: You probably want a newline for the test annotation.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ResourceHandlerTest.java
```diff
@@ -2798,8 +2821,12 @@ public void testListingXSS() throws Exception
 
         /*
          * Intentionally bad request URI. Sending a non-encoded URI with typically
-         * encoded characters '<', '>', and '"'.
+         * encoded characters '<', '>', and '"', using the path-param feature of the
+         * URI spec to still produce a listing.  This path-param value should not make it
+         * down to the ResourceListing.getAsXHTML() method.
          */
+        // TODO: investigate why ResourceHandler isn't producing a directory listing when asked here.
```
lorban: Please open an issue about this problem.

lorban (changes requested):

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ResourceHandlerTest.java
joakime: Fixed.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ResourceHandlerTest.java
joakime: This is now a bogus TODO.
This was actually fixed in other PRs already.
Removing both TODOs here.

lorban (approved):

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletTest.java
```diff
@@ -465,13 +465,13 @@ public void testListingFilenamesOnlyUrlResource() throws Exception
 
         assertThat(response.getStatus(), is(HttpStatus.OK_200));
         body = response.getContent();
-        assertThat(body, containsString(">#hashcode&nbsp;<")); // text on page
+        assertThat(body, containsString(">#hashcode&#160;<")); // text on page
```
gregw: why is the numeric value better than the slightly descriptive `nbsp`?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
```diff
@@ -197,6 +199,7 @@ public static String getAsHTML(Resource resource, String base, boolean parent, S
         {
             // Name
             buf.append("<tr><td class=\"name\"><a href=\"");
+            // TODO This produces an absolute link from the /context/<listing-dir> path, investigate if we can use relative links reliably now
```
gregw, line 202: +1

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/ResourceListingTest.java
```diff
@@ -0,0 +1,453 @@
...
+    private static boolean isValidXHtml(String content)
+    {
+        // we expect that our generated output conforms to text/xhtml is well formed
+        try (ByteArrayInputStream inputStream = new ByteArrayInputStream(content.getBytes(StandardCharsets.UTF_8)))
+        {
+            DocumentBuilderFactory xmlDocumentBuilderFactory = DocumentBuilderFactory.newInstance();
+            xmlDocumentBuilderFactory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+            DocumentBuilder db = xmlDocumentBuilderFactory.newDocumentBuilder();
+            // We consider this content to be XML well formed if these 2 lines do not throw an Exception
+            Document doc = db.parse(inputStream);
```
gregw, line 449: +1

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -285,7 +285,10 @@ public List<Resource> list()
         List<Resource> result = new ArrayList<>();
         for (Resource r : _resources)
         {
-            result.addAll(r.list());
+            if (r.isDirectory())
+                result.addAll(r.list());
+            else
+                result.add(r);
```
gregw, line 291: This is a strange extra case.  Do we ever have a use-case for a resource collection where it's members are not all directories?  Perhaps this is a restriction that can be put into ResourceCollection (not in this PR), so if we intend to do that, this is not required.

gregw (commented):
This is pretty much good to go, but I have a few niggles.   I won't request changes nor approve as I'll be out of contact, but I'm happy for this go go forward... even happier if niggles fixed.

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletTest.java
joakime: `&nbsp;` is an HTML named entity that isn't valid on XHTML.

Well ... technically it's not valid on XML.
XML only had 5 named entities (`&lt;`, `&gt;`, `&amp;`, `&apos;` and `&quot;`).
I'll see if I can find a valid xsd/dtd that could possibly bring in more named entities and remain standards supported.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
joakime: Yes, a testcase has that example (some entries are dirs, some are files).
But a general review of this is a good idea.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
```diff
@@ -145,7 +147,7 @@ public static String getAsHTML(Resource resource, String base, boolean parent, S
             }
         }
 
-        buf.append("<tr><th class=\"name\"><a href=\"?C=N&O=").append(order).append("\">");
+        buf.append("<tr><th class=\"name\"><a href=\"?C=N&amp;O=").append(order).append("\">");
```
sbordet, line 150: This would be a legit use of naked `&`, it should not be `&amp;` since it's a query parameter separator.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
```diff
@@ -165,7 +167,7 @@ public static String getAsHTML(Resource resource, String base, boolean parent, S
             }
         }
 
-        buf.append("<th class=\"lastmodified\"><a href=\"?C=M&O=").append(order).append("\">");
+        buf.append("<th class=\"lastmodified\"><a href=\"?C=M&amp;O=").append(order).append("\">");
```
sbordet, line 170: Ditto.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
```diff
@@ -184,7 +186,7 @@ public static String getAsHTML(Resource resource, String base, boolean parent, S
                 arrow = ARROW_DOWN;
             }
         }
-        buf.append("<th class=\"size\"><a href=\"?C=S&O=").append(order).append("\">");
+        buf.append("<th class=\"size\"><a href=\"?C=S&amp;O=").append(order).append("\">");
```
sbordet, line 189: Ditto.

sbordet (changes requested):

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletTest.java
joakime: I added the DTD for xhtml and have a testcase that validates against the DTD now.
I have restored the `&nbsp;` usage.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
joakime: Not recommended to have naked ampersands in urls / hrefs / links in HTML or XHTML documents.

See https://www.w3.org/International/questions/qa-escapes#ampersands

> **Single ampersands.** Although HTML user agents have tended to turn a blind eye, you should never have a single ampersand (`&`) in your HTML document. You should pay particular attention to URIs that include parameters. For example, your document should contain `http://example.org/my-script.php?class=guest&amp;name=user`, rather than `http://example.org/my-script.php?class=guest&name=user`.

joakime: A naked ampersand in a url / href / link also fails both HTML5 and XHTML 1.1 validation

sbordet: Doh! Leave it as `&amp;` then.

sbordet (approved):

PR #8472, opened 2022-08-17
Restore jetty-ee8-maven-plugin mvove back ITs to jetty-eeX-maven-plugin modules

olamy:
- add back jetty-ee8-maven-plugin
- add back jetty-ee8-jspc-maven-plugin
- remove not used site plugin and simplify code
- moving back ee9 maven plugin ITs to the module
- back of ee10 maven ITs test to the plugin module


janbartel:
I just don't see how it is going to work to move the plugin it tests back into the `jetty-eeX-maven-plugin` module. The reason I moved them out is because they depend on artifacts produced by building `jetty-home`, and `jetty-home` depends on building `jetty-eeX`.  If you look at the distro tests for the ee10 plugin you see: `<modules>ee10-apache-jsp,ee10-glassfish-jstl,ee10-testmod,resources</modules>` . That is an unexpressed dependency on `jetty-home`.  Before I moved all of the integration tests out of the `jetty-eeX-maven-plugin`, the build would fail intermittently because it was all dependent on the order of the build whether it worked or not: building from the top level sometimes worked, sometimes not and the same went for building just the `jetty-eeX-maven-plugins`. This was particularly obvious when I cleaned out my local repo of all jetty-12 artifacts and started fresh.

So if you do this change, how are you ensuring that there is no circular dependency on `jetty-home` and `jetty-eeX`, and that building from the top level with a empty repo will work?

olamy:
@janbartel `jetty-home` do not depend on `jetty-eeX-maven-plugin`. `jetty-eeX` is not a dependency.
The change add dependency on jetty-home for each `jetty-eeX-maven-plugin` such https://github.com/eclipse/jetty.project/pull/8472/files#diff-5f85754f7acf88c40067098bb1ddb6088c6b8d1914915226dadac2dfd927fcaaR163
Because we need to tell maven `jetty-eeX-maven-plugin` depends on it!

janbartel:
> @janbartel `jetty-home` do not depend on `jetty-eeX-maven-plugin`. `jetty-eeX` is not a dependency. 

OK so the circuit breaker in here was the introduction of the `jetty-eeX-home` modules, and having `jetty-home` depending on those. None of the `jetty-eeX-home` modules depends on any of the `jetty-eeX-maven-plugin` modules, so there is no loop. Good.  

What's the advantage of doing more work to move these integration tests back into the `jetty-eeX-maven-plugin` modules? We have other integration tests that are external to the module. Is it a problem to generate the `ee8` module?

olamy:
> > @janbartel `jetty-home` do not depend on `jetty-eeX-maven-plugin`. `jetty-eeX` is not a dependency.
> 
> OK so the circuit breaker in here was the introduction of the `jetty-eeX-home` modules, and having `jetty-home` depending on those. None of the `jetty-eeX-home` modules depends on any of the `jetty-eeX-maven-plugin` modules, so there is no loop. Good.
> 

no the circuit breaker was the introduction of the Jenkinsfile building the project in multiple steps giving this visual impression about the build order in a graphical way. And by the way that's why you got issues when you had first the it in the plugin sources. Jenkins was failing because of the Jenkinsfile using the `-f` option which doesn't care about the build plan and dependencies between reactor modules. 
A Maven may not respect the `<modules>` declaration in various poms. Maven will look at dependencies 

> What's the advantage of doing more work to move these integration tests back into the `jetty-eeX-maven-plugin` modules? We have other integration tests that are external to the module. Is it a problem to generate the `ee8` module? 
btw the work is done.

because it's easier and more natural to run the tests and  IT tests in just one go with something such `mvn verify -pl :jetty-ee9-maven-plugin` or `mvn verify -f jetty-ee9/etty-ee9-maven-plugin` (even if I don't like the `-f` as this ignore so many parts from the reactor...


session is a bit different as it's a sort of tck and dependencies etc.. need to be different between each session provider.

PR #8475, opened 2022-08-18
Implemented ProxyHandler in new module jetty-core/jetty-proxy.

sbordet:
Moved ConnectHandler to org.eclipse.jetty.server.handler, where it should have been from the start.

Cleaned up and corrected behavior of SecureRequestCustomizer.
Now the response is wrapped with isSecure=true if it arrived over a secure transport.
The request URI scheme indicates whether the request is secure towards the origin server.


jmcc0nn3ll (approved):
Thanks for the walk-through, looks great!

PR #8486, opened 2022-08-24
backward compat for ee10-security module which is now in jetty-ee10-servlet

olamy:


olamy:
@gregw jetty-10 have mixed together jetty-servlet and jetty-security
so the ee10-security need to use jetty-ee10-servlet jar

gregw:
@olamy Can you check the status of this and ask for review if it is still needed

olamy:
@gregw it depends what we want to do exactly. 
ee9 has a Maven module jetty-ee9-security as so the correspond Jetty module ee9-security.mod
ee10 doesn't have the same Maven module jetty-ee10-security so do we want/need to keep having a ee10-security.mod which should only have `jetty-ee10-servlet-${jetty.version}.jar` in it.
the file ee10-security.mod will here only for a sort of backward compat so users can still use `--add-module=ee9-security` then  `--add-module=ee10-security` where they do not need it.
I don't really have strong opinion here

gregw:
@lachlan-roberts can you think about this one when you consider core security

gregw:
@olamy I think this has been addressed with the core security changes?   Can we close?

olamy:
yes

PR #8487, opened 2022-08-26
add some public MessageHandlers to websocket-core-common

lachlan-roberts:
There were MessageHandlers outside of the `internal` package for websocket-core. So this PR adds `MessageHandler` and `DemandingMessageHandler` to public packages in `websocket-core-common`.

`DemandingMessageHandler` is able to use the `ByteBufferCallbackAccumulator` so can be more efficient with its aggregation of binary messages.


jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/AbstractMessageHandler.java
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * A utility implementation of FrameHandler that de-fragments text frames and binary frames into a whole messages before
+ * calling {@link #onText(String, Callback)} or {@link #onBinary(ByteBuffer, Callback)}.
+ * This is a demanding frame handler so
+ * Flow control is by default automatic, but an implementation
+ * may extend {@link #isDemanding()} to return true and then explicityly control
+ * demand with calls to {@link CoreSession#demand(long)}
+ */
+public abstract class AbstractMessageHandler implements FrameHandler
```
sbordet: This class is still `internal`, but maybe it'll be useful to expose it?

For example, to use a different accumulation method, or to asynchronously proxy frames to other components (hence no accumulation needed), I would need to extend from this class to reuse most/all of its code, but I can't if it's internal.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ByteArrayMessageSink.java
```diff
@@ -79,16 +79,17 @@ public void accept(Frame frame, Callback callback)
                 if (out == null)
                     out = new ByteBufferCallbackAccumulator();
                 out.addEntry(payload, callback);
+                callback = Callback.NOOP;
             }
 
             // If the methodHandle throws we don't want to fail callback twice.
```
sbordet: Is this comment misplaced now?

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ByteBufferMessageSink.java
```diff
@@ -77,10 +77,10 @@ public void accept(Frame frame, Callback callback)
                 if (out == null)
                     out = new ByteBufferCallbackAccumulator();
                 out.addEntry(payload, callback);
+                callback = Callback.NOOP;
             }
 
             // If the methodHandle throws we don't want to fail callback twice.
```
sbordet: Comment misplaced now?

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/DemandingMessageHandler.java
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * <p>A utility implementation of FrameHandler that de-fragments text frames and binary frames into a whole messages before
+ * calling {@link #onText(String, Callback)} or {@link #onBinary(ByteBuffer, Callback)}.</p>
+ * <p>This is a demanding frame handler so flow control is by not automatic, an implementation of {@link #onText(String, Callback)}
+ * and {@link #onBinary(ByteBuffer, Callback)} must call {@link CoreSession#demand(long)} to receive the next WebSocket frame.
+ * Because this is a demanding FrameHandler it can be more efficient for binary messages by avoiding copies
+ * for aggregation until the full message has been received.</p>
+ */
+public class DemandingMessageHandler extends AbstractMessageHandler
```
sbordet, line 47: `DemandingMessageHandler` means to me that this class demands, but apparently it's not the case.
I would rename the class to just `MessageHandler`.

Also, I don't particularly like that to demand the next _message_ you have to use an API that demands _frames_.
However, I understand this is a low-level utility class, so I guess users have to know about the low-levelness.

Please review the javadoc, contains typos and says it's a demanding FrameHandler, but it's not (or it is? but if it is, why all the methods need to call `demand(1)`?).

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/MessageHandler.java
```diff
@@ -0,0 +1,175 @@
...
+import org.eclipse.jetty.websocket.core.exception.BadPayloadException;
+import org.eclipse.jetty.websocket.core.exception.MessageTooLargeException;
+import org.eclipse.jetty.websocket.core.internal.AbstractMessageHandler;
+
+/**
+ * <p>A utility implementation of FrameHandler that de-fragments text frames and binary frames into a whole messages before
+ * calling {@link #onText(String, Callback)} or {@link #onBinary(ByteBuffer, Callback)}.</p>
+ * <p>This is a demanding frame handler so flow control is by default automatic when the callback is succeeded.</p>
+ */
+public class MessageHandler extends AbstractMessageHandler
```
sbordet: Similar comment to the `DemandingMessageHandler`, as this is the demanding one, so naming and javadocs need review.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/MessageHandler.java
```diff
@@ -0,0 +1,175 @@
...
+            if (frame.hasPayload())
+            {
+                if (_binaryMessageBuffer == null)
+                    _binaryMessageBuffer = new ByteArrayOutputStream();
+                BufferUtil.writeTo(frame.getPayload(), _binaryMessageBuffer);
+            }
+
+            if (frame.isFin())
+            {
+                onBinary(BufferUtil.toBuffer(_binaryMessageBuffer.toByteArray()), callback);
```
sbordet: I feel the usage of `ByteArrayOutputStream` forces too many copies of the data.
Why you did not use `ByteBufferAccumulator`?

sbordet (changes requested):

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/AbstractMessageHandler.java
lachlan-roberts: I have moved this outside internal package to `org.eclipse.jetty.websocket.core.util`.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ByteArrayMessageSink.java
lachlan-roberts: removed

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ByteBufferMessageSink.java
lachlan-roberts: removed

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/DemandingMessageHandler.java
lachlan-roberts: It is named `DemandingMessageHandler` because `FrameHandler.isDemanding()` returns true. I have now renamed the other one to `AutoDemandingMessageHandler` to make it more clear.

I have also added a `demand()` method onto `DemandingMessageHandler`.

Not sure what these typos are you are referring to but I have added to the javadoc.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/MessageHandler.java
lachlan-roberts: It is not possible to use this with the `AutoDemandingMessageHandler`. We cannot receive the next message until the callback is succeeded and to succeed the callback means we no longer own the buffer, so we need to copy the buffer.

If you want more efficient binary message aggregation you should use the `DemandingMessageHandler`.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/DemandingMessageHandler.java
sbordet: @lachlan-roberts sorry don't understand. `isDemanding()==true` means that we have to explicitly demand, but `isDemanding()==false` means "auto-demanding"? Makes no sense.

Perhaps you should rename `FrameHandler.isDemanding()` to something that is clearer, so far it's just confusing to no end.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/MessageHandler.java
sbordet: `ByteBufferAccumulator` does exactly that, it copies the buffer, so you can use it.
Right now you're doing the same job as `ByteBufferAccumulator`, only using `ByteArrayOutputStream`.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/AutoDemandingMessageHandler.java
```diff
@@ -0,0 +1,183 @@
...
+ * This means that flow control is by automatic when the callback is succeeded.
+ * It is preferable to use {@link DemandingMessageHandler} as the aggregation for binary messages can be done more
+ * efficiently if the frame handler can manage its own demand.
+ * </p>
+ * @see DemandingMessageHandler
+ */
+public class AutoDemandingMessageHandler extends AbstractMessageHandler
+{
+    private Utf8StringBuilder _textMessageBuffer;
+    private ByteArrayOutputStream _binaryMessageBuffer;
```
sbordet: `ByteBufferAccumulator` does copy the buffer, so you can use it instead of `ByteArrayOutputStream`.

sbordet (changes requested):

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/DemandingMessageHandler.java
lachlan-roberts: @sbordet I don't think we can be just renaming public api methods like that and breaking anyone implementing `FrameHandler`.

It is currently explained what demanding means in the javadoc of `FrameHandler`. We could change the name of this for 12 if we have a better name for it. @gregw what do you think?

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/MessageHandler.java
lachlan-roberts: Yep agreed, I was thinking of `ByteBufferCallbackAccumulator`.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/DemandingMessageHandler.java
joakime: As far as method name changes, those can be for Jetty 12.
The classnames of the new classes can be changed now, but not the old ones.

lachlan-roberts:
@sbordet can you please re-review.

jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferAccumulator.java
```diff
@@ -187,10 +187,15 @@ public void writeTo(OutputStream out) throws IOException
         }
     }
 
-    @Override
-    public void close()
+    public void reset()
```
sbordet, line 190: Why adding this method?
It's just for clarity because it's better to `reset()` and the reuse it, rather than `close()` and reuse it?

jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferCallbackAccumulator.java
```diff
@@ -96,4 +96,14 @@ public void fail(Throwable t)
         _entries.clear();
         _length = 0;
     }
+
+    public void close()
+    {
+        for (Entry entry : _entries)
+        {
+            entry.callback.succeeded();
+        }
```
sbordet:
```suggestion
        _entries.forEach(entry -> entry.callback.succeeded());
```

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ByteArrayMessageSink.java
```diff
@@ -79,16 +79,16 @@ public void accept(Frame frame, Callback callback)
...
-            // If the methodHandle throws we don't want to fail callback twice.
-            callback = Callback.NOOP;
             if (frame.isFin())
             {
                 byte[] buf = out.takeByteArray();
                 methodHandle.invoke(buf, 0, buf.length);
             }
 
+            callback.succeeded();
             session.demand(1);
```
sbordet, line 92: Should you demand if `frame.isFin()`?

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ByteBufferMessageSink.java
```diff
@@ -97,6 +96,7 @@ public void accept(Frame frame, Callback callback)
                 }
             }
 
+            callback.succeeded();
             session.demand(1);
```
sbordet, line 100: Should you demand if `frame.isFin()`?

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/AutoDemandingMessageHandler.java
```diff
@@ -0,0 +1,186 @@
...
+import org.eclipse.jetty.websocket.core.exception.BadPayloadException;
+import org.eclipse.jetty.websocket.core.exception.MessageTooLargeException;
+
+/**
+ * <p>
+ * A utility implementation of FrameHandler that de-fragments text frames and binary frames into a whole messages before
+ * calling {@link #onText(String, Callback)} or {@link #onBinary(ByteBuffer, Callback)}.
+ * </p>
+ * <p>
+ * This is a demanding frame handler, meaning {@link FrameHandler#isDemanding()} returns false.
```
sbordet: OMG! This sentence is nuts, just remove it.

Can you explain what this class does without going into these much confusing details?
Something like "This class implicitly demands more frames when the `Callback` passed to either `onText()` or `onBinary()` is succeeded."

Please open an issue to rename this insanity in Jetty 12 😬

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/DemandingMessageHandler.java
```diff
@@ -0,0 +1,222 @@
...
+import org.eclipse.jetty.websocket.core.exception.BadPayloadException;
+import org.eclipse.jetty.websocket.core.exception.MessageTooLargeException;
+
+/**
+ * <p>
+ * A utility implementation of FrameHandler that de-fragments text frames and binary frames into a whole messages before
+ * calling {@link #onText(String, Callback)} or {@link #onBinary(ByteBuffer, Callback)}.
+ * </p>
+ * <p>
+ * This is a demanding frame handler, meaning {@link FrameHandler#isDemanding()} returns true.
```
sbordet: Ditto. Remove this sentence and explain what this class does in other words.

sbordet (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferAccumulator.java
lachlan-roberts: Yes, just because it looks wrong to do `close()` and then reuse it.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/ByteBufferMessageSink.java
lachlan-roberts: Yes we still need to demand, there is no demand in the Jetty WebSocket API.

PR #8491, opened 2022-08-29
jetty 12.0.x error unwrap servlet exception

olamy:
@gregw maybe time to finally do it with jetty 12?


joakime (approved):
LGTM.   It's simple enough to revert if we discover issues with it during testing.

gregw (approved):

Issue #8493, opened 2022-08-30
Review HTTP client feature `setRemoveIdleDestinations`


lorban:
**Jetty version(s)**
9.4.x, 10.0.x+

**Description**
The `removeIdleDestinations` feature always was racy, but with the advent of #5218 the race went from benign to problematic, because the `Pool` object does not give back connections when it is closed.

Also review why we use both `HttpDestination.close()` and `HttpDestination.doStop()` to manage lifecycle, and why `HttpDestination` is both added as lifecycle bean and in a `Map` within `HttpClient`.

sbordet:
@lorban here's an idea for fixing this issue.

`HttpClient.resolveDestination()` is implemented to call a new method, `HttpDestination.stale()` (not just a getter because it has a side effect).

```java
boolean HttpDestination::stale() {
  lock {
    boolean stale = _stale;
    if (!stale) {
      _staleCount = 0;
    }
    return stale;
  }
}
```

Then, if `removeIdleDestination==true`, we use a `Sweeper` to periodically sweep destinations:

```java
boolean HttpDestination::sweep() {
  boolean remove = false;
  lock {
    boolean stale = _exchanges.isEmpty() && _connectionPool.isEmpty();
    if (stale)
      ++_staleCount;
    else
      _staleCount = 0;
    if (_staleCount == 4) {
      _stale = true;
      remove = true;
    }
  }
  if (remove) {
    getHttpClient().removeDestination(this);
    LifeCycle.stop(this);
  }
  return remove;
}
```

The sweeper's attempt to remove/stop the destination gets pushed into the future every time `resolveDestination()` is called, because `_staleCount` is reset, so the sweeper loses the race with the call to `stale()`.
If the sweeper wins the race, then `_stale==true` and `resolveDestination()` will create a new destination object for the same origin.

`HttpClient.removeDestination()` needs to be implemented by calling `Map.remove(key, value)` to be sure to remove the stale destination and not the new one possibly created by `resolveDestination()`.

At this point we know that the stale destination cannot be referenced for new requests, _and_ it is not processing any request (as it has been stale for 4 sweeper cycles) therefore it can be safely stopped.

PR #8495, opened 2022-08-31
Jetty 9.4.x : fix client remove idle destinations

lorban:
Solves #8493 for 9.4.x by implementing the solution proposed by @sbordet.

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
@@ -148,7 +148,7 @@ public class HttpClient extends ContainerLifeCycle
     private boolean tcpNoDelay = true;
     private boolean strictEventOrdering = false;
     private HttpField encodingField;
-    private boolean removeIdleDestinations = false;
+    private long destinationIdleTimeout;
     private boolean connectBlocking = false;
     private String name = getClass().getSimpleName() + "@" + Integer.toHexString(hashCode());
     private HttpCompliance httpCompliance = HttpCompliance.RFC7230;
@@ -1088,13 +1088,43 @@ public class HttpClient extends ContainerLifeCycle
     }
 
     /**
+     * @return the delay in ms before idle destinations should be removed
+     * @see #setDestinationIdleTimeout(long)
+     */
+    @ManagedAttribute("The delay in ms before idle destinations are removed, disabled when zero or negative")
+    public long getDestinationIdleTimeout()
+    {
+        return destinationIdleTimeout;
+    }
+
+    /**
+     * Whether destinations that have no connections (nor active nor idle) should be removed
+     * after a certain delay.
+     * <p>
+     * Applications typically make request to a limited number of destinations so keeping
+     * destinations around is not a problem for the memory or the GC.
+     * However, for applications that hit millions of different destinations (e.g. a spider
+     * bot) it would be useful to be able to remove the old destinations that won't be visited
+     * anymore and leave space for new destinations.
+     *
+     * @param destinationIdleTimeout the delay in ms before idle destinations should be removed
+     * @see org.eclipse.jetty.client.DuplexConnectionPool
+     */
+    public void setDestinationIdleTimeout(long destinationIdleTimeout)
+    {
+        this.destinationIdleTimeout = destinationIdleTimeout;
+    }
+
+    /**
      * @return whether destinations that have no connections should be removed
      * @see #setRemoveIdleDestinations(boolean)
+     * @deprecated replaced by {@link #getDestinationIdleTimeout()}
      */
+    @Deprecated
     @ManagedAttribute("Whether idle destinations are removed")
     public boolean isRemoveIdleDestinations()
     {
-        return removeIdleDestinations;
+        return destinationIdleTimeout > 0L;
     }
 
     /**
@@ -1108,10 +1138,12 @@ public class HttpClient extends ContainerLifeCycle
      *
      * @param removeIdleDestinations whether destinations that have no connections should be removed
      * @see org.eclipse.jetty.client.DuplexConnectionPool
+     * @deprecated replaced by {@link #setDestinationIdleTimeout(long)}, calls the latter with a value of 5000 ms.
      */
+    @Deprecated
     public void setRemoveIdleDestinations(boolean removeIdleDestinations)
     {
-        this.removeIdleDestinations = removeIdleDestinations;
+        setDestinationIdleTimeout(removeIdleDestinations ? 5000 : 0);
     }
 
     /**
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
@@ -71,7 +71,6 @@ import org.eclipse.jetty.util.annotation.ManagedAttribute;
 import org.eclipse.jetty.util.annotation.ManagedObject;
 import org.eclipse.jetty.util.component.ContainerLifeCycle;
 import org.eclipse.jetty.util.component.DumpableCollection;
-import org.eclipse.jetty.util.component.LifeCycle;
 import org.eclipse.jetty.util.log.Log;
 import org.eclipse.jetty.util.log.Logger;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
@@ -155,7 +154,7 @@ public class HttpClient extends ContainerLifeCycle
     private String name = getClass().getSimpleName() + "@" + Integer.toHexString(hashCode());
     private HttpCompliance httpCompliance = HttpCompliance.RFC7230;
     private String defaultRequestContentType = "application/octet-stream";
-    private Sweeper destinationIdleTimeoutSweeper;
+    private Sweeper destinationSweeper;
 
     /**
      * Creates a {@link HttpClient} instance that can perform requests to non-TLS destinations only
@@ -241,21 +240,6 @@ public class HttpClient extends ContainerLifeCycle
         if (scheduler == null)
             setScheduler(new ScheduledExecutorScheduler(name + "-scheduler", false));
 
-        destinationIdleTimeoutSweeper = new Sweeper(scheduler, 1000L);
-        // Bind the start of the destinationIdleTimeoutSweeper to the scheduler's start as we cannot add the former as a bean
-        // because HttpDestination.doStart() expects to find a different sweeper by calling getBean() on the HttpClient to be
-        // used for connection pool sweeping. That sweeper cannot be used for destinationIdleTimeout sweeping, so we need to
-        // maintain a second one on the client that isn't a bean and is dedicated to that task.
-        scheduler.addLifeCycleListener(new LifeCycle.Listener()
-        {
-            @Override
-            public void lifeCycleStarted(LifeCycle event)
-            {
-                LifeCycle.start(destinationIdleTimeoutSweeper);
-                scheduler.removeLifeCycleListener(this);
-            }
-        });
-
         if (resolver == null)
             setSocketAddressResolver(new SocketAddressResolver.Async(executor, scheduler, getAddressResolutionTimeout()));
 
@@ -270,7 +254,14 @@ public class HttpClient extends ContainerLifeCycle
         cookieStore = cookieManager.getCookieStore();
 
         transport.setHttpClient(this);
+
         super.doStart();
+
+        if (getDestinationIdleTimeout() > 0L)
+        {
+            destinationSweeper = new Sweeper(scheduler, 1000L);
+            destinationSweeper.start();
+        }
     }
 
     private CookieManager newCookieManager()
@@ -281,7 +272,11 @@ public class HttpClient extends ContainerLifeCycle
     @Override
     protected void doStop() throws Exception
     {
-        LifeCycle.stop(destinationIdleTimeoutSweeper);
+        if (destinationSweeper != null)
+        {
+            destinationSweeper.stop();
+            destinationSweeper = null;
+        }
 
         decoderFactories.clear();
         handlers.clear();
@@ -338,9 +333,9 @@ public class HttpClient extends ContainerLifeCycle
         return cookieManager;
     }
 
-    Sweeper getDestinationIdleTimeoutSweeper()
+    Sweeper getDestinationSweeper()
     {
-        return destinationIdleTimeoutSweeper;
+        return destinationSweeper;
     }
 
     /**
@@ -1113,30 +1108,36 @@ public class HttpClient extends ContainerLifeCycle
     }
 
     /**
-     * @return the delay in ms before idle destinations should be removed
+     * The default value is 0
+     * @return the time in ms after which idle destinations are removed
      * @see #setDestinationIdleTimeout(long)
      */
-    @ManagedAttribute("The delay in ms before idle destinations are removed, disabled when zero or negative")
+    @ManagedAttribute("The time in ms after which idle destinations are removed, disabled when zero or negative")
     public long getDestinationIdleTimeout()
     {
         return destinationIdleTimeout;
     }
 
     /**
-     * Whether destinations that have no connections (nor active nor idle) should be removed
-     * after a certain delay.
      * <p>
-     * Applications typically make request to a limited number of destinations so keeping
-     * destinations around is not a problem for the memory or the GC.
-     * However, for applications that hit millions of different destinations (e.g. a spider
-     * bot) it would be useful to be able to remove the old destinations that won't be visited
-     * anymore and leave space for new destinations.
+     * Whether destinations that have no connections (nor active nor idle) and no exchanges
+     * should be removed after the specified timeout.
+     * </p>
+     * <p>
+     * If the specified {@code destinationIdleTimeout} is 0 or negative, then the destinations
+     * are not removed.
+     * </p>
+     * <p>
+     * Avoids accumulating destinations when applications (e.g. a spider bot or web crawler)
+     * hit a lot of different destinations that won't be visited again.
+     * </p>
      *
-     * @param destinationIdleTimeout the delay in ms before idle destinations should be removed
-     * @see org.eclipse.jetty.client.DuplexConnectionPool
+     * @param destinationIdleTimeout the time in ms after which idle destinations are removed
      */
     public void setDestinationIdleTimeout(long destinationIdleTimeout)
     {
+        if (isStarted())
+            LOG.warn("Calling setDestinationIdleTimeout() while started has no effect");
         this.destinationIdleTimeout = destinationIdleTimeout;
     }
 
@@ -1168,7 +1169,7 @@ public class HttpClient extends ContainerLifeCycle
     @Deprecated
     public void setRemoveIdleDestinations(boolean removeIdleDestinations)
     {
-        setDestinationIdleTimeout(removeIdleDestinations ? 10_000 : 0);
+        setDestinationIdleTimeout(removeIdleDestinations ? 10_000L : 0L);
     }
 
     /**
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
@@ -1138,12 +1138,12 @@ public class HttpClient extends ContainerLifeCycle
      *
      * @param removeIdleDestinations whether destinations that have no connections should be removed
      * @see org.eclipse.jetty.client.DuplexConnectionPool
-     * @deprecated replaced by {@link #setDestinationIdleTimeout(long)}, calls the latter with a value of 5000 ms.
+     * @deprecated replaced by {@link #setDestinationIdleTimeout(long)}, calls the latter with a value of 10000 ms.
      */
     @Deprecated
     public void setRemoveIdleDestinations(boolean removeIdleDestinations)
     {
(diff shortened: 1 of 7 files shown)


jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
```diff
@@ -104,21 +109,72 @@ public HttpDestination(HttpClient client, Origin origin)
...
 
+    public boolean stale()
+    {
+        try (Locker.Lock l = staleLock.lock())
+        {
+            boolean stale = this.stale;
+            if (!stale)
+                this.staleCount = 0;
+            if (LOG.isDebugEnabled())
+                LOG.debug("Stale check on {} done, result: {}", this, stale);
```
sbordet: Just a nit. Switch to log first the result, then the destination, as its `toString()` can be really long.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
```diff
@@ -104,21 +109,72 @@ public HttpDestination(HttpClient client, Origin origin)
...
+            LOG.debug("Sweep check on {} in progress", this);
+        boolean remove = false;
+        try (Locker.Lock l = staleLock.lock())
+        {
+            boolean stale = exchanges.isEmpty() && connectionPool.isEmpty();
+            if (stale)
+                this.staleCount++;
+            else
+                this.staleCount = 0;
+            if (this.staleCount == 4)
```
sbordet: This hardcoded value `4` means that we only have a 4 s safe window for the sweeper. 
If the `stale()` thread stalls for more than that, we can still sweep a destination that it is about to be used.

Maybe a better solution would be to have a `destinationIdleTimeout`, and rather than counting to `4` we compare the time between now and first time the destination became stale with the destination idle timeout.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
```diff
@@ -104,21 +109,72 @@ public HttpDestination(HttpClient client, Origin origin)
...
+                LOG.debug("Stale check on {} done, result: {}", this, stale);
+            return stale;
+        }
+    }
+
+    @Override
+    public boolean sweep()
+    {
+        if (LOG.isDebugEnabled())
+            LOG.debug("Sweep check on {} in progress", this);
```
sbordet:
```suggestion
            LOG.debug("Sweep check in progress on {}", this);
```

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
```diff
@@ -104,21 +109,72 @@ public HttpDestination(HttpClient client, Origin origin)
...
+                remove = true;
+            }
+        }
+        if (remove)
+        {
+            getHttpClient().removeDestination(this);
+            LifeCycle.stop(this);
+        }
+        if (LOG.isDebugEnabled())
+            LOG.debug("Sweep check on {} done, result: {}", this, remove);
```
sbordet: As above, log the destination as last.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
```diff
@@ -104,21 +109,72 @@ public HttpDestination(HttpClient client, Origin origin)
...
     @Override
     protected void doStart() throws Exception
     {
         this.connectionPool = newConnectionPool(client);
         addBean(connectionPool, true);
         super.doStart();
         Sweeper sweeper = client.getBean(Sweeper.class);
+        if (sweeper == null && getHttpClient().isRemoveIdleDestinations())
+        {
+            sweeper = new Sweeper(getHttpClient().getScheduler(), 1000L);
```
sbordet: Let's use a private `Sweeper` in all cases.

jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTest.java
```diff
@@ -538,6 +540,47 @@ public void handle(String target, org.eclipse.jetty.server.Request baseRequest,
         }
     }
 
+    @ParameterizedTest
+    @ArgumentsSource(ScenarioProvider.class)
+    public void testRetryWithRemoveIdleDestinationsEnabled(Scenario scenario) throws Exception
+    {
+        start(scenario, new AbstractHandler()
```
sbordet: Use `EmptyServerHandler`.

jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTest.java
```diff
@@ -538,6 +540,47 @@ public void handle(String target, org.eclipse.jetty.server.Request baseRequest,
...
+                .send();
+        }
+
+        // Wait for the sweeper to remove the idle HttpDestination.
+        await().atMost(10, TimeUnit.SECONDS).until(() ->
+        {
+            Collection<HttpDestination> destinations = client.getBeans(HttpDestination.class);
+            return destinations.isEmpty();
+        });
+    }
```
sbordet, line 575: Good test, but add a test that tests the sweeping of destination in `HttpDestinationOverHTTPTest`.

sbordet (changes requested):

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
lorban: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
lorban: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
lorban: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
lorban: Done.

jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTest.java
lorban: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
lorban: Done.

jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientTest.java
lorban: Those already exist: `testDestinationIsRemoved` and `testDestinationIsRemovedAfterConnectionError`, including the check that the `HttpDestination` instances are different after the idle timeout.
I just needed to modify them slightly to adapt them to the new `DestinationIdleTimeout` mechanism.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
```diff
@@ -65,7 +68,11 @@ public abstract class HttpDestination extends ContainerLifeCycle implements Dest
     private final ClientConnectionFactory connectionFactory;
     private final HttpField hostField;
     private final RequestTimeouts requestTimeouts;
+    private final Locker staleLock = new Locker();
     private ConnectionPool connectionPool;
+    private boolean stale;
+    private long staleTs;
```
sbordet: Rename to `staleNanos`, so we know what unit it is.
Also, this is more like `activeNanos`.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
```diff
@@ -65,7 +68,11 @@ public abstract class HttpDestination extends ContainerLifeCycle implements Dest
...
+    private Sweeper idleDestinationsSweeper;
```
sbordet: Uhm. Now we have a sweeper per destination, which is not efficient.
I think we should initialize it in `HttpClient.doStart()` add it as a bean in `HttpClient`, and from here just look it up.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
```diff
@@ -104,23 +111,83 @@ public HttpDestination(HttpClient client, Origin origin)
         hostField = new HttpField(HttpHeader.HOST, host);
     }
 
+    public boolean stale()
+    {
+        try (Locker.Lock l = staleLock.lock())
+        {
+            boolean stale = this.stale;
+            if (!stale)
+                this.staleTs = 0;
```
sbordet: I don't think this is right. Should it not be `System.nanoTime()` instead of `0`?

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
```diff
@@ -530,16 +577,38 @@ public String asString()
...
+        {
+            return this.stale;
+        }
+    }
+
     @Override
     public String toString()
     {
-        return String.format("%s[%s]@%x%s,queue=%d,pool=%s",
+        return String.format("%s[%s]@%x%s,state=%s,queue=%d,pool=%s,stale=%s,staleDelay=%d",
```
sbordet: Don't like the word `delay`, as it's not really a delay.
How about just `stale=%b,idle=%d` where `idle` is filled with `System.nanoTime() - staleTs` in ms, which then can also be exposed via JMX?

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
```diff
@@ -1101,10 +1138,12 @@ public boolean isRemoveIdleDestinations()
      *
      * @param removeIdleDestinations whether destinations that have no connections should be removed
      * @see org.eclipse.jetty.client.DuplexConnectionPool
+     * @deprecated replaced by {@link #setDestinationIdleTimeout(long)}, calls the latter with a value of 5000 ms.
      */
+    @Deprecated
     public void setRemoveIdleDestinations(boolean removeIdleDestinations)
     {
-        this.removeIdleDestinations = removeIdleDestinations;
+        setDestinationIdleTimeout(removeIdleDestinations ? 5000 : 0);
```
sbordet: I'd bump it to 10 s.

sbordet (changes requested):

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
lorban: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
lorban: Renamed.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
lorban: Good catch, fixed.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
lorban: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
lorban: This is going to clash with the sweeper used for the connection pool's sweeping.

lorban: After discussion it was decided that `HttpClient` is going to keep that extra sweeper as a field and not added as a bean.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
```diff
@@ -238,6 +241,21 @@ protected void doStart() throws Exception
...
+        // maintain a second one on the client that isn't a bean and is dedicated to that task.
+        scheduler.addLifeCycleListener(new LifeCycle.Listener()
+        {
+            @Override
+            public void lifeCycleStarted(LifeCycle event)
+            {
+                LifeCycle.start(destinationIdleTimeoutSweeper);
+                scheduler.removeLifeCycleListener(this);
+            }
+        });
```
sbordet: I think it should be much simpler.

```java
if (getDestinationIdleTimeout() > 0)
{
    destinationIdleTimeoutSweeper = new Sweeper(scheduler, 1000L);
    scheduler.addBean(destinationIdleTimeoutSweeper);
}
```

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
```diff
@@ -263,6 +281,8 @@ private CookieManager newCookieManager()
     @Override
     protected void doStop() throws Exception
     {
+        LifeCycle.stop(destinationIdleTimeoutSweeper);
```
sbordet: To avoid exception wrapping, this should be:

```java
if (destinationIdleTimeoutSweeper != null)
    destinationIdleTimeoutSweeper.stop();
```

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
```diff
@@ -1080,14 +1112,44 @@ public void setStrictEventOrdering(boolean strictEventOrdering)
...
+     * @see #setDestinationIdleTimeout(long)
+     */
+    @ManagedAttribute("The delay in ms before idle destinations are removed, disabled when zero or negative")
+    public long getDestinationIdleTimeout()
+    {
+        return destinationIdleTimeout;
+    }
+
+    /**
+     * Whether destinations that have no connections (nor active nor idle) should be removed
```
sbordet: no connections and no exchanges queued.

Please use Javadoc standards of wrapping every paragraph with <p> and </p> (let's not use the old style of unclosed <p> as if it was a newline).

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
```diff
@@ -1080,14 +1112,44 @@ public void setStrictEventOrdering(boolean strictEventOrdering)
...
+     * after a certain delay.
```
sbordet:
```suggestion
     * after the specified timeout.
```

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
```diff
@@ -1080,14 +1112,44 @@ public void setStrictEventOrdering(boolean strictEventOrdering)
...
+     * <p>
+     * Applications typically make request to a limited number of destinations so keeping
+     * destinations around is not a problem for the memory or the GC.
```
sbordet: Remove the 2 lines above.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
```diff
@@ -1080,14 +1112,44 @@ public void setStrictEventOrdering(boolean strictEventOrdering)
...
+     * However, for applications that hit millions of different destinations (e.g. a spider
+     * bot) it would be useful to be able to remove the old destinations that won't be visited
+     * anymore and leave space for new destinations.
```
sbordet:
```suggestion
     * Avoids accumulating destinations when applications (e.g. a spider bot or web crawler) 
     * hit a lot of different destinations that won't be visited again.
```

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
```diff
@@ -1080,14 +1112,44 @@ public void setStrictEventOrdering(boolean strictEventOrdering)
...
+     *
+     * @param destinationIdleTimeout the delay in ms before idle destinations should be removed
```
sbordet:
```suggestion
     * @param destinationIdleTimeout the time in ms after which idle destinations are removed
```

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
```diff
@@ -1080,14 +1112,44 @@ public void setStrictEventOrdering(boolean strictEventOrdering)
...
+     * @see org.eclipse.jetty.client.DuplexConnectionPool
```
sbordet: Why this reference? I think it's unrelated.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
```diff
@@ -104,23 +110,81 @@ public HttpDestination(HttpClient client, Origin origin)
...
         this.connectionPool = newConnectionPool(client);
         addBean(connectionPool, true);
         super.doStart();
-        Sweeper sweeper = client.getBean(Sweeper.class);
-        if (sweeper != null && connectionPool instanceof Sweeper.Sweepable)
-            sweeper.offer((Sweeper.Sweepable)connectionPool);
+        Sweeper connectionPoolSweeper = client.getBean(Sweeper.class);
+        if (connectionPoolSweeper != null && connectionPool instanceof Sweeper.Sweepable)
+            connectionPoolSweeper.offer((Sweeper.Sweepable)connectionPool);
+        getHttpClient().getDestinationIdleTimeoutSweeper().offer(this);
```
sbordet:
```suggestion
        Sweeper destinationSweeper = getHttpClient().getDestinationIdleTimeoutSweeper();
        if (destinationSweeper != null)
            destinationSweeper.offer(this);
```

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
```diff
@@ -104,23 +110,81 @@ public HttpDestination(HttpClient client, Origin origin)
...
     }
 
     @Override
     protected void doStop() throws Exception
     {
-        Sweeper sweeper = client.getBean(Sweeper.class);
-        if (sweeper != null && connectionPool instanceof Sweeper.Sweepable)
-            sweeper.remove((Sweeper.Sweepable)connectionPool);
+        getHttpClient().getDestinationIdleTimeoutSweeper().remove(this);
```
sbordet: Null check as above.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
```diff
@@ -530,16 +574,39 @@ public String asString()
...
+        {
+            return this.stale;
+        }
+    }
+
     @Override
     public String toString()
     {
-        return String.format("%s[%s]@%x%s,queue=%d,pool=%s",
+        return String.format("%s[%s]@%x%s,state=%s,queue=%d,pool=%s,stale=%s,idle=%d",
```
sbordet: `stale=%b`

sbordet (changes requested):

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
lorban: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
lorban: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
lorban: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
lorban: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpDestination.java
lorban: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
lorban: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
lorban: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
lorban: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
lorban: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
lorban: Done.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
lorban: Done.

sbordet (dismissed):

sbordet (approved):

PR #8535, opened 2022-09-02
Fixes #8532 - Review System.nanoTime() usages.

### LINKED ISSUE #8532: Review System.nanoTime() usages
**Jetty version(s)**
10+

**Description**
`System.nanoTime()`, as specified in its javadocs, can roll-over.
Because of that, it's wrong to use the `<` operator between values, as well as `Math.min()` etc.

This problem has already been reported in CometD, see https://github.com/cometd/cometd/issues/1215.

The whole codebase should be reviewed so that `System.nanoTime()` is properly used.

sbordet:
Introduced `o.e.j.util.NanoTime` class to deal with nanoTimes.

Now NanoTime.now() should be used instead of System.nanoTime(),
and various `<unit>elapsed*()` methods to calculate elapsed times.

Furthermore, comparing 2 nanoTimes should be done via `isBefore()`,
rather than using the `<` operator, which is wrong as specified in
the System.nanoTime() javadocs.

### DIFF
--- a/jetty-server/src/test/java/org/eclipse/jetty/server/StressTest.java
+++ b/jetty-server/src/test/java/org/eclipse/jetty/server/StressTest.java
@@ -18,7 +18,6 @@ import java.net.Socket;
 import java.util.Queue;
 import java.util.Random;
 import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
@@ -26,6 +25,7 @@ import javax.servlet.http.HttpServletResponse;
 
 import org.eclipse.jetty.server.handler.HandlerWrapper;
 import org.eclipse.jetty.util.IO;
+import org.eclipse.jetty.util.NanoTime;
 import org.eclipse.jetty.util.thread.QueuedThreadPool;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
@@ -345,11 +345,11 @@ public class StressTest
     {
         if (persistent)
         {
-            long start = TimeUnit.NANOSECONDS.toMillis(System.nanoTime());
+            long start = NanoTime.now();
             Socket socket = new Socket("localhost", _connector.getLocalPort());
             socket.setSoTimeout(30000);
 
-            long connected = TimeUnit.NANOSECONDS.toMillis(System.nanoTime());
+            long connected = NanoTime.now();
 
             for (int i = 0; i < __tests.length; i++)
             {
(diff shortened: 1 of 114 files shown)


jetty-util/src/main/java/org/eclipse/jetty/util/NanoTime.java
```diff
@@ -0,0 +1,180 @@
...
+        return elapsed(beginNanoTime, now());
+    }
+
+    /**
+     * <p>Calculates the nanoseconds elapsed between the current nanoTime and an end nanoTime.</p>
+     *
+     * @param endNanoTime the end nanoTime
+     * @return the nanoseconds elapsed between the current nanoTime and the given end nanoTime
+     */
+    public static long elapsedTo(long endNanoTime)
```
lorban: This name is confusing as `endNanoTime` is supposed to be in the future but the past tense in the method name suggests that it's in the past.
Something like `remaningTo` for instance sounds better to me.

lorban (changes requested):

jetty-util/src/main/java/org/eclipse/jetty/util/NanoTime.java
sbordet: I like `remainingTo`, updated.

lorban (dismissed):

gregw (changes requested):
Quick review whilst I have network.
Mostly good, but can the methods of NanoTime be renamed:
- elapsed(long, long); since(long); until(long);
- millisElapsed(long, long); millisSince(long); millisUntil(long);
- secondsElapsed(long, long); secondsSince(long); secondsUntil(long);

I'll check more as I go....

jetty-server/src/test/java/org/eclipse/jetty/server/StressTest.java
```diff
@@ -458,24 +458,19 @@ private static class TestHandler extends HandlerWrapper
...
-            long received = baseRequest.getTimeStamp();
 
             _handled.incrementAndGet();
-            long delay = received - start;
-            if (delay < 0)
-                delay = 0;
+            long delay = NanoTime.millisElapsed(start, now);
             _latencies[2].add(delay);
-            _latencies[3].add((now - start));
+            _latencies[3].add(delay);
```
gregw, line 467: I don't think this is the same logic?

gregw (changes requested):
I've looked through all the files and other than the one query noted, I've not seen any issues other than better names for NanoTime methods.

jetty-util/src/main/java/org/eclipse/jetty/util/NanoTime.java
gregw: `remainingTo` is not good phrasing.  It should be `remainingUntil` or just `until`.   I think using `elapsed`, `since` and `until` is clear.

jetty-server/src/test/java/org/eclipse/jetty/server/StressTest.java
sbordet: It's not exactly the same logic, but before it was a garbage value, since `baseRequest.getTimeStamp()` returns an epoch millis time, and `delay` was calculated mixing up epoch millis with current nanoTime converted to millis -- a bogus number that could have even be negative.

Maybe `baseRequest.getTimeStamp()` should return a nanoTime, but `CustomRequestLog` relies on it to be a UTC dateTime -- feels like we should either use `Instant` or have 2 values (the nanoTime and the dateTime).

lorban (approved):
I like the new names better!

gregw (approved):

jetty-server/src/test/java/org/eclipse/jetty/server/StressTest.java
gregw: Ok. It's only a test anyway, so if it is a nonsense value, perhaps delete it.

Issue #8540, opened 2022-09-02
Maven pom is not correct for `org.eclipse.jetty/infinispan-embedded` and `org.eclipse.jetty/infinispan-remote`


MoonLord-LM:
**Jetty version(s)**
10.0.x
11.0.x
12.0.x

**Java version/vendor** 

**OS type/version**

**Description**
【org.eclipse.jetty/infinispan-embedded】
【org.eclipse.jetty/infinispan-remote】 

404 Not Found:
https://repo1.maven.org/maven2/org/eclipse/jetty/infinispan-embedded/10.0.11/infinispan-embedded-10.0.11.jar
https://repo1.maven.org/maven2/org/eclipse/jetty/infinispan-remote/10.0.11/infinispan-remote-10.0.11.jar

~These projects are not 【 `<packaging>jar</packaging>` 】 but 【`<packaging>pom</packaging>`】.~
~These should be marked as 【`<type>pom</type>`】，like other things.~
~Such as 【`<type>war</type>`】 and 【`<type>zip</type>`】.~

https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-infinispan/infinispan-embedded/pom.xml
https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-infinispan/infinispan-remote/pom.xml

**How to reproduce?**

MoonLord-LM:
Oh, I see.
It is about the classifier 【`<classifier>config</classifier>`】.

https://repo1.maven.org/maven2/org/eclipse/jetty/infinispan-embedded/10.0.11/
https://repo1.maven.org/maven2/org/eclipse/jetty/infinispan-remote/10.0.11/

The correct jar file link is:
https://repo1.maven.org/maven2/org/eclipse/jetty/infinispan-embedded/10.0.11/infinispan-embedded-10.0.11-config.jar
https://repo1.maven.org/maven2/org/eclipse/jetty/infinispan-remote/10.0.11/infinispan-remote-10.0.11-config.jar

MoonLord-LM:
I will fix this.

PR #8541, opened 2022-09-02
Issue #8540, add classifier for jetty-10.0.x

MoonLord-LM:
fix：https://github.com/eclipse/jetty.project/issues/8540

### DIFF
--- a/jetty-bom/pom.xml
+++ b/jetty-bom/pom.xml
@@ -210,6 +210,7 @@
       <dependency>
         <groupId>org.eclipse.jetty</groupId>
         <artifactId>infinispan-remote</artifactId>
+        <classifier>config</classifier>
         <version>10.0.12-SNAPSHOT</version>
       </dependency>
       <dependency>
@@ -220,6 +221,7 @@
       <dependency>
         <groupId>org.eclipse.jetty</groupId>
         <artifactId>infinispan-embedded</artifactId>
+        <classifier>config</classifier>
         <version>10.0.12-SNAPSHOT</version>
       </dependency>
       <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -1360,6 +1360,7 @@
       <dependency>
         <groupId>org.eclipse.jetty</groupId>
         <artifactId>infinispan-embedded</artifactId>
+        <classifier>config</classifier>
         <version>${project.version}</version>
       </dependency>
       <dependency>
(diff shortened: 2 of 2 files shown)


MoonLord-LM:
@olamy

olamy (approved):

jetty-bom/pom.xml
```diff
@@ -210,6 +210,7 @@
       <dependency>
         <groupId>org.eclipse.jetty</groupId>
         <artifactId>infinispan-remote</artifactId>
+        <classifier>config</classifier>
```
joakime, line 213: What?

MoonLord-LM: https://repo1.maven.org/maven2/org/eclipse/jetty/infinispan-remote/10.0.11/

[https://repo1.maven.org/maven2/org/eclipse/jetty/infinispan-remote/10.0.11/infinispan-remote-10.0.11-<b>config</b>.jar](https://repo1.maven.org/maven2/org/eclipse/jetty/infinispan-remote/10.0.11/infinispan-remote-10.0.11-config.jar)

joakime: Depending on the `config` classifier (only meaningful to `jetty-home`) in a bom is just super odd, and feels wrong.
Why not just have `<type>pom</type>` instead?

MoonLord-LM: Yes, I agree, `<type>pom</type>` is better.

joakime: @olamy can you think of any maven technical reason for not just using `<type>pom</type>` instead?

olamy: the bom is mostly made (e.g only useful) for embedded users.
So TBH I don't even understand why we have `<artifactId>infinispan-remote</artifactId>` in the bom. This module contains no code but only configuration files.
So except the `config` jar which contains the jetty configuration module files it's a totally useless artifact :) 
Either we remove it from the bom or we "expose" the only useless artifact which is the `config` but frankly I even don't understand why :) 
wdy? Remove the entry from the bom?

joakime: > Remove the entry from the bom?

I'd vote +1 for this approach.

olamy: FTR https://github.com/eclipse/jetty.project/pull/8554

PR #8547, opened 2022-09-06
re-enabling and fixing some tests for Jetty-12

lachlan-roberts:
re-enabling and fixing some tests for Jetty-12

### DIFF
--- a/jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/ResponseHeadersTest.java
+++ b/jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/ResponseHeadersTest.java
@@ -15,7 +15,6 @@ package org.eclipse.jetty.ee10.servlet;
 
 import java.io.IOException;
 import java.io.PrintWriter;
-import java.net.URLDecoder;
 import java.nio.ByteBuffer;
 
 import jakarta.servlet.ServletException;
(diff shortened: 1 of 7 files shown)


jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/ResponseHeadersTest.java
```diff
@@ -204,9 +201,7 @@ public void testMultilineResponseHeaderValue() throws Exception
         assertThat("Response Code", response.getStatus(), is(200));
         assertThat("Response Header Content-Type", response.get("Content-Type"), is("text/plain;charset=UTF-8"));
 
-        String expected = StringUtil.replace(actualPathInfo, "%0A", " "); // replace OBS fold with space
-        expected = URLDecoder.decode(expected, "utf-8"); // decode the rest
-        expected = expected.trim(); // trim whitespace at start/end
+        String expected = URIUtil.canonicalPath(actualPathInfo);
```
joakime: Looks like you changed the expectations to make the test pass.
Why?

joakime (changes requested):
Expectation change bothers me.

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/ResponseHeadersTest.java
lachlan-roberts: Because it is testing the value for `pathInfo` which used to be decoded but now is returning the "canonically encoded" value.

joakime: Looking back at the history of this test, this is a test to prevent a regression.

There was a time where a response header value could (accidentally?) trigger a response header formatted with the old "obsolete fold" http syntax.
This test was to ensure that this obsolete fold format was not used.

lachlan-roberts: I will fully decode the `pathInfo` in the servlet and then send it back as a header. 
Then we can keep this test expectation the same.

joakime (approved):
LGTM

Issue #8474, opened 2022-08-17
Jetty 12 : Resource API Review


joakime:
# org.eclipse.jetty.util.resource.Resource Proposal

## High level

The `Resource` abstract class should be made into an Interface, with minimal default implementations (if any).  
Existing static methods behaviors stay.

The `ResourceCollection` should be renamed.  
It's ultimately how we composite a set of Resources together.

Its current implementation is a List, so calling it `ResourceList` is tempting.  
However, the current implementation flattens and uniques the incoming Resources, so it cannot have a nested composite inside of it.
This was done in early iterations when ResourceCollection was managing the resources (mounts).
It was felt that having multiple composites, possibly from different management scopes (now called `ResourceFactory`) could be a recipe for shooting your whole leg off.

This concern is not valid anymore IMO.
We could allow a composite to have other composites.
So lets call it `CompositeResource`.

My argument here is that `List` implies the same entry type, and is flat, but `Composite` makes it clear that it structured.

The `Resource` interface should also expose a `Iterable<Resource>` (more on that below).

### Use Case: Create Resource from multiple Paths and pass it around

Simple creation, but used as a composite

``` java
Resource resourceBase = resourceFactory.newResource(listOfUris);

contextHandler.setResourceBase(resourceBase);
```

Allowed URIs can reference locations that do not exist.  
A composite with zero entries should throw an exception.

### Use Case: Create Resource from single path and pass it around

Simple creation, used as a simple resource

``` java
Resource resourceBase = resourceFactory.newResource(uri);

contextHandler.setResourceBase(resourceBase);
```

Allowed URI or Path can reference locations that do not exist.

### Use Case: combine multiple resources into a single Resource

Combine Resources together to form a composite Resource.

``` java
ResourceCollection rc1 = Resource.combine(
    List.of(
        resourceFactory.newResource(one), // simple directory
        resourceFactory.newResource(two), // simple directory
        resourceFactory.newResource(three) // simple directory
    )
);

ResourceCollection rc2 = Resource.combine(
    List.of(
        // the original ResourceCollection
        rc1,
        // a duplicate entry
        resourceFactory.newResource(two), // simple directory
        // a new entry
        resourceFactory.newResource(dirFoo) // simple directory
    )
);

rc2.getResources(); // has 4 entries [one, two, three, dirFoo]
```

Constructing a composite resource from other resources, even other composite resources.
Duplicate entries are eliminated.
[...]

PR #8551, opened 2022-09-06
Issue #8474 - Jetty 12 - Introduce `Iterable<Resource>` to base `Resource`

joakime:
Per Issue #8474 - base `Resource` now implements `Iterable<Resource>`.
This eliminates a big need for code to be aware of `ResourceCollection`.

A future PR will use this more appropriately elsewhere in the code (as its expected to be a largish PR)

The purpose of this is to evaluate the implementation for `PathResource` and `ResourceCollection` now, in a small PR.

### DIFF
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
@@ -34,7 +34,6 @@ import java.util.Base64;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
-import java.util.NoSuchElementException;
 
 import org.eclipse.jetty.util.IO;
 import org.eclipse.jetty.util.URIUtil;
@@ -166,30 +165,7 @@ public abstract class Resource implements Iterable<Resource>
     @Override
     public Iterator<Resource> iterator()
     {
-        return new Iterator<>()
-        {
-            private boolean next = true;
-
-            @Override
-            public boolean hasNext()
-            {
-                return next;
-            }
-
-            @Override
-            public Resource next()
-            {
-                if (next)
-                {
-                    next = false;
-                    return Resource.this;
-                }
-                else
-                {
-                    throw new NoSuchElementException("No more Resource entries to iterate");
-                }
-            }
-        };
+        return List.of(this).iterator();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -152,6 +154,44 @@ public boolean isSame(Resource resource)
...
+     * <p>
+     *     This is meaningful if you have a Composite Resource, otherwise it will be a single entry Iterator.
+     * </p>
+     *
+     * @return the iterator of Resources.
+     */
+    @Override
+    public Iterator<Resource> iterator()
+    {
+        return new Iterator<>()
```
sbordet: Uhm. `List.of(this).iterator()` seems a lot simpler.

`List.of(this)` could be cached as a field if worried about the extra allocation.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/PathResourceTest.java
```diff
@@ -140,4 +141,15 @@ public void testSame()
...
         }
     }
+
+    @Test
+    public void testIterable()
+    {
+        Path rpath = MavenTestingUtils.getTestResourcePathFile("resource.txt");
+        Resource resource = ResourceFactory.root().newResource(rpath);
+        int count = 0;
+        for (Resource r: resource)
```
sbordet: Add space before `:`.

sbordet (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime: Not worried about allocation, as calling  `.iterator()` will allocate anyway.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/PathResourceTest.java
joakime: Done.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime: Done.

sbordet (approved):

PR #8552, opened 2022-09-06
Issue #8474 - Jetty 12 - Eliminate `Resource.isMemoryMappable`

joakime:
Per Issue #8474 - Remove Resource.isMemoryMappable()

This was only used for BufferUtil.toMappedBuffer()

The determination should reside there, along with possibly an alternative API that can return a non-mapped `ByteBuffer` if the size is smallish.
See [javadoc on FileChannel.map()](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/channels/FileChannel.html#map(java.nio.channels.FileChannel.MapMode,long,long)) for reference.
> For most operating systems, mapping a file into memory is more expensive than reading or writing a few tens of kilobytes of data via the usual [read](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/channels/FileChannel.html#read(java.nio.ByteBuffer)) and [write](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/channels/FileChannel.html#write(java.nio.ByteBuffer)) methods. From the standpoint of performance it is generally only worth mapping relatively large files into memory.

### DIFF
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
@@ -1071,14 +1071,16 @@ public class BufferUtil
 
     public static ByteBuffer toMappedBuffer(Resource resource) throws IOException
     {
-        if (!resource.isMemoryMappable())
+        Path path = resource.getPath();
+        if (path == null)
             return null;
         return toMappedBuffer(resource.getPath());
     }
 
     public static ByteBuffer toMappedBuffer(Resource resource, long pos, long len) throws IOException
     {
-        if (!resource.isMemoryMappable())
+        Path path = resource.getPath();
+        if (path == null)
             return null;
         return toMappedBuffer(resource.getPath(), pos, len);
     }
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
@@ -1074,7 +1074,7 @@ public class BufferUtil
         Path path = resource.getPath();
         if (path == null)
             return null;
-        return toMappedBuffer(resource.getPath());
+        return toMappedBuffer(path);
     }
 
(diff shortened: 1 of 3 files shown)


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
```diff
@@ -1071,14 +1071,16 @@ public static ByteBuffer toMappedBuffer(Path filePath, long pos, long len) throw
 
     public static ByteBuffer toMappedBuffer(Resource resource) throws IOException
     {
-        if (!resource.isMemoryMappable())
+        Path path = resource.getPath();
+        if (path == null)
             return null;
         return toMappedBuffer(resource.getPath());
```
sbordet: Use the `path` variable.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
```diff
@@ -1071,14 +1071,16 @@ public static ByteBuffer toMappedBuffer(Path filePath, long pos, long len) throw
...
     }
 
     public static ByteBuffer toMappedBuffer(Resource resource, long pos, long len) throws IOException
     {
-        if (!resource.isMemoryMappable())
+        Path path = resource.getPath();
+        if (path == null)
             return null;
         return toMappedBuffer(resource.getPath(), pos, len);
```
sbordet: Use the `path` variable.

sbordet (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
joakime: Done.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
joakime: Done.

sbordet (approved):

PR #8553, opened 2022-09-06
Issue #8474 - Jetty 12 - Eliminate `Resource.isSame`

joakime:
Per Issue #8474 - Eliminate `Resource.isSame()` as it is not used outside of test cases.


sbordet (changes requested):
The changes are ok, but can you please modernize equals() and hashCode() to use `Objects.equals(path, other.path)` and `Objects.hashCode(path)`.
And move the 2 close to each other, so we know they're overridden as a pair.

sbordet (approved):

PR #8554, opened 2022-09-07
remove infinispan-remote from bom as it doesn't have to be here

olamy:


joakime (changes requested):
Also ...

``` xml
      <dependency>
        <artifactId>infinispan-embedded</artifactId>
        <classifier>config</classifier>
        <version>10.0.12-SNAPSHOT</version>
      </dependency>
```

olamy:
@joakime arghh! done as well

joakime (approved):

PR #8573, opened 2022-09-08
Jetty 12 - Delay Resource Alias calculation until it's requested/needed

joakime:
Per Issue #8474 - Delay the calculation of Resource Alias until it's requested/needed


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -44,8 +44,9 @@ public class PathResource extends Resource
         .build();
 
     private final Path path;
-    private final Path alias;
     private final URI uri;
+    private boolean aliasChecked = false;
```
sbordet: I'd rename this field to `resolved`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -44,8 +44,9 @@ public class PathResource extends Resource
...
+    private Path alias;
 
     private Path checkAliasPath()
```
sbordet: I'd rename this method to `resolve()` or `resolveRealURI()`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -352,9 +352,10 @@ public boolean isAlias()
...
 
     /**
-     * The canonical Alias for the Resource as a URI.
+     * If this Resource is an alias pointing to a different location,
+     * the resulting location as URI is returned pointing to the real location.
      *
-     * @return The canonical Alias of this resource or null if none.
+     * @return The real URI location of this resource or null if not different.
      */
     public URI getAlias()
```
sbordet: The name of this method is confusing, because it does not return an alias, but the real location.

Where this method is used, the variable name is typically `realURI` so perhaps it should be renamed to `getResolvedRealURI()`, with the `Resolved` part indicating that if there was no resolution because the resource was already a real location, then `null` is returned.

I could also live with `getRealURI()` always return the real URI and for the 3-4 places this method is used rather than a null check do an equal check: `if (!resource.getURI().equals(resource.getRealURI()))`, or equivalently a call to `isAlias()`.
I like this latter solution better because it will be evident what you're testing for, rather than doing a null check and having to read the javadocs to understand that `null` means that the resource URI is already a real URI.

sbordet (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: `resolve()` is the name of a different method, the main one that gets a new Resource from this one.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime: I don't like the name "get*RealURI()" as it conflicts with the `getURI()`, and has a name that indicates it will always return a value.
It _only returns a value_ only if there is a difference from what `getURI()` would return.

It's almost a `getAliasLocationURI()`, but it still stems from how backwards this API is.

Currently, it is .. Alias is `this`, and the `.getAlias()` shows the real file location / URI.

joakime: Wonder if we can use the language from symlinks?

```
$ man ln

NAME
       ln - make links between files

SYNOPSIS
       ln [OPTION]... [-T] TARGET LINK_NAME
       ln [OPTION]... TARGET
       ln [OPTION]... TARGET... DIRECTORY
       ln [OPTION]... -t DIRECTORY TARGET...

DESCRIPTION
       In  the  1st form, create a link to TARGET with the name LINK_NAME.  In
       the 2nd form, create a link to TARGET in the current directory.  In the
       3rd  and  4th  forms, create links to each TARGET in DIRECTORY.  Create
       hard links by default, symbolic links  with  --symbolic.   By  default,
       each  destination  (name  of  new link) should not already exist.  When
       creating hard links, each TARGET must exist.  Symbolic links  can  hold
       arbitrary  text;  if  later resolved, a relative link is interpreted in
       relation to its parent directory.
```

In this language, the `PathResource` itself is the `LINK_NAME` and the `.getAlias()` is the `TARGET`.

How about `.getTargetURI()` ?

joakime: Changed to `.getTargetURI()` (it reads better too)

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -325,9 +254,87 @@ public boolean isContainedIn(Resource r)
...
+        {
+            alias = resolveTargetAliasPath();
+            targetResolved = true;
+        }
+        if (alias == null)
+            return null;
+        return alias.toUri();
+    }
+
+    private Path resolveTargetAliasPath()
```
sbordet: Should be renamed to just `resolveTargetPath()` as _this_ `Resource` is the alias.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/ssl/KeyStoreScanner.java
```diff
@@ -53,10 +54,12 @@ public KeyStoreScanner(SslContextFactory sslContextFactory)
         if (Files.isDirectory(monitoredFile))
             throw new IllegalArgumentException("expected keystore file not directory");
 
-        if (keystoreResource.getAlias() != null)
+        // Use real location of keystore (if different), so that change monitoring can work properly
+        URI realUri = keystoreResource.getTargetURI();
+        if (realUri != null)
         {
             // this resource has an alias, use the alias, as that's what's returned in the Scanner
```
sbordet: Remove this comment that is now obsoleted by the one above.

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/ContextHandler.java
```diff
@@ -1404,7 +1404,7 @@ public boolean checkAlias(String path, Resource resource)
         if (resource.isAlias())
         {
             if (LOG.isDebugEnabled())
-                LOG.debug("Aliased resource: {}~={}", resource, resource.getAlias());
+                LOG.debug("Aliased resource: {}~={}", resource, resource.getTargetURI());
```
sbordet:
```suggestion
                LOG.debug("Alias resource {} for {}", resource, resource.getTargetURI());
```

sbordet (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -48,79 +48,9 @@ public class PathResource extends Resource
...
-            LOG.trace("IGNORED", e);
-        }
-        catch (Exception e)
-        {
-            LOG.warn("bad alias ({} {}) for {}", e.getClass().getName(), e.getMessage(), path);
-        }
-        return null;
-    }
+    private boolean targetResolved = false;
+    private Path alias;
```
sbordet: Since now everything refers to `target`, rename this field to `target` or `targetPath`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/ssl/KeyStoreScanner.java
```diff
@@ -53,11 +54,10 @@ public KeyStoreScanner(SslContextFactory sslContextFactory)
         if (Files.isDirectory(monitoredFile))
             throw new IllegalArgumentException("expected keystore file not directory");
 
-        if (keystoreResource.getAlias() != null)
-        {
-            // this resource has an alias, use the alias, as that's what's returned in the Scanner
-            monitoredFile = Paths.get(keystoreResource.getAlias());
-        }
+        // Use real location of keystore (if different), so that change monitoring can work properly
+        URI realUri = keystoreResource.getTargetURI();
```
sbordet: Rename variable to `targetURI`?

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebInfConfiguration.java
```diff
@@ -291,12 +291,13 @@ public void unpack(WebAppContext context) throws IOException
             if (webApp == null)
                 throw new IllegalStateException("No resourceBase or war set for context");
 
-            // Accept aliases for WAR files
-            if (webApp.isAlias())
+            // Use real location (if different) for WAR file, so that change/modification monitoring can work.
+            URI realUri = webApp.getTargetURI();
```
sbordet: Rename variable to `targetURI`?

sbordet (approved):
LGTM, please fix a couple of renamings.

PR #8577, opened 2022-09-09
add a step to push tag to a fork

olamy:


.github/ISSUE_TEMPLATE/release-template.md
```diff
@@ -27,11 +27,14 @@ This release process will produce releases:
 - [ ] Verify target [project(s)](https://github.com/eclipse/jetty.project/projects) are complete.
 - [ ] Verify that branch `jetty-10.0.x` is merged to branch `jetty-11.0.x`.
 - [ ] Assign issue to "build manager", who will stage the releases.
+  + [ ] create a branch release/target_jetty_version then work from the branch
   + [ ] Ensure `VERSION.txt` additions for each release will be meaningful, descriptive, correct text.
   + [ ] Stage 9.4 release with Java 11.
   + [ ] Stage 10 release with Java 17.
   + [ ] Stage 11 release with Java 17.
   + [ ] Edit a draft release (for each Jetty release) in GitHub (https://github.com/eclipse/jetty.project/releases). Content is generated with the "changelog tool".
+  + [ ] Push tag to the fork https://github.com/jetty-project/jetty.project
```
joakime: "to the fork" ?


How about ...

The tags need to be pushed before you create the draft release (this step needs to reference the tag, otherwise this draft release will _create_ the tag, which we do not want)

```
+ [ ] Push release branches `release/<ver>` to to https://github.com/eclipse/jetty.project
+ [ ] Push release tags `jetty-<ver>` to https://github.com/eclipse/jetty.project
+ [ ] Edit a draft release (for each Jetty release) in GitHub (https://github.com/eclipse/jetty.project/releases). Content is generated with the "changelog tool".
```

.github/ISSUE_TEMPLATE/release-template.md
```diff
@@ -27,11 +27,14 @@ This release process will produce releases:
 - [ ] Verify target [project(s)](https://github.com/eclipse/jetty.project/projects) are complete.
 - [ ] Verify that branch `jetty-10.0.x` is merged to branch `jetty-11.0.x`.
 - [ ] Assign issue to "build manager", who will stage the releases.
+  + [ ] create a branch release/target_jetty_version then work from the branch
```
joakime: How about ..

```
+ [ ] Create and use branches `release/<ver>` to perform version specific release work from.
```

joakime (changes requested):
Order and language changes

.github/ISSUE_TEMPLATE/release-template.md
olamy: you mean pushing to eclipse/jetty.project? 
perso I don't like it much as some users may take it as a real release and will eventually fetch the tag which can be a project for tag such jetty-10.0.12 (as it will be the same tag name)

.github/ISSUE_TEMPLATE/release-template.md
olamy: Definitely better english ;) thanks!

.github/ISSUE_TEMPLATE/release-template.md
joakime: Yes, that's what I meant.
Edited my comment to fit.

olamy: you do not need to reference a tag to create a draft release. the draft just need to have the potential target release in the name.
I still find a dangerous to push non validated (e.g only staged) tags to mainstream repo as users may take this as real release. especially tags now (>= 10) doesn't have anymore timestamp in their name. And so those users may get the tag in their fork.

joakime: > I still find a dangerous to push non validated (e.g only staged) tags to mainstream repo as users may take this as real release. especially tags now (>= 10) doesn't have anymore timestamp in their name. And so those users may get the tag in their fork.

A concern for sure, but the fact that these staged releases show up on `oss.sonatype.org` in a public way already is enough to show that this issue exists already, no matter how much we want to prevent/avoid this.
Projects that want early releases can already access those from oss.sonatype.org, and we have many such projects on github that do that now.   The existence (or not) of the tags doesn't change that.

joakime: At least with the changes in github over the past few months, the meaning of "tag" vs "release" has been separated.

No longer does a "tag" automatically indicate a "release".
Also, if we push the tag, and then build a "draft release" against said tag, that tag also shows as not having a release (yet).
That existence of a new tag also doesn't automatically stomp on the "release" page (like it used to in the past).

olamy: changes done. this will make our life easier

joakime (approved):

Issue #8578, opened 2022-09-09
`getRequestURL` can append "null" if `getRequestURI` is unspecified in an authority-form request-target


electrum:
**Jetty version(s)**

9.4.x
10.x

**Description**

When upgrading to Jetty 9.4.48, we see that `AbstractProxyServlet.rewriteTarget()` calls `clientRequest.getRequestURL()` which results in `https://localhost:50393null` due to `getRequestURI()` returning null (since the URI path is null). While this `Request` behavior seems to have existed for ~10 years and is not the root issue, it still seems to be a bug.

joakime:
Opened PR #8579 to solve for `jetty-9.4.x`.

Attempting to replicate in `jetty-10.0.x` now ...

joakime:
Opened PR #8580 to solve for `jetty-10.0.x` (this will be merged to `jetty-11.0.x` once approved/merged)

joakime:
Complete.

Merged from 9.4.x through to 12.0.x

PR #8579, opened 2022-09-09
Issue #8578 - `getRequestURL` can append "null" if `getRequestURI` is unspecified in an authority-form request-target

joakime:
When working with CONNECT headers, the old implementation of `HttpURI.getPath()` would return the authority component.
The changes in #8014 broke this backward compatibility behavior, causing unintended behaviors on `HttpServletRequest.getRequestURL()` and `HttpServletRequest.getRequestURI()`.

### DIFF
--- a/jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
+++ b/jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
@@ -876,10 +876,12 @@ public class RequestTest
     @Test
     public void testConnectRequestURL() throws Exception
     {
-        final AtomicReference<String> result = new AtomicReference<>();
+        final AtomicReference<String> resultRequestURL = new AtomicReference<>();
+        final AtomicReference<String> resultRequestURI = new AtomicReference<>();
         _handler._checker = (request, response) ->
         {
-            result.set("" + request.getRequestURL());
+            resultRequestURL.set("" + request.getRequestURL());
+            resultRequestURI.set("" + request.getRequestURI());
             return true;
         };
 
@@ -890,7 +892,8 @@ public class RequestTest
                 "\n");
         HttpTester.Response response = HttpTester.parseResponse(rawResponse);
         assertThat(response.getStatus(), is(HttpStatus.OK_200));
-        assertThat(result.get(), is("http://myhost:9999"));
+        assertThat(resultRequestURL.get(), is("http://myhost:9999"));
+        assertThat(resultRequestURI.get(), is("myhost:9999"));
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
```diff
@@ -872,6 +873,29 @@ public boolean check(HttpServletRequest request, HttpServletResponse response)
...
+        _handler._checker = (request, response) ->
+        {
+            resultRequestURL.set("" + request.getRequestURL());
+            resultRequestURI.set("" + request.getRequestURI());
+            return true;
+        };
+
+        String rawResponse = _connector.getResponse(
+            "CONNECT myhost:9999 HTTP/1.1\n" +
+                "Host: myhost:9999\n" +
```
sbordet, line 891: What happens when `Host` is different from the authority-form just after the `CONNECT` method?
Can we write another test case to document how we should behave in that case?

sbordet (changes requested):

jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
joakime: Right now in Jetty 9.4.48, if the request is say ....

```
CONNECT myhost:9999 HTTP/1.1
Host: otherhost:8888
Connection: close
```

The two API calls become ...

* `HttpServletRequest.getRequestURL()` becomes `http://otherhost:8888`
* `HttpServletRequest.getRequestURI()` becomes `otherhost:8888`

The information `myhost:9999` actually gets lost.
When the request gets parse, the request HttpURI starts out with `myhost:9999` (authority) when the `CONNECT` line is parsed.
Then gets overridden with `otherhost:8888` when the `Host` header is parsed.

Prior to PR #8014, the details in `myhost:9999` was actually stored in `HttpURI.getPath()` so that the two different authorities could be tracked separately.

sbordet: RFC 7230, section 5.4 states:

> If the target URI includes an authority component, then a
   client MUST send a field-value for Host that is identical to that
   authority component

That does not explicitly say "the request target is an authority-form", but perhaps could be inferred.

Also (same section):

> When a proxy receives a request with an absolute-form of
   request-target, the proxy MUST ignore the received Host header field
   (if any) and instead replace it with the host information of the
   request-target.

So the `Host` header value can be ignored, as long as there is another authority.
Again, it does not explicitly say the behavior for the authority-form, but talks about proxies and I guess we could have the same behavior (i.e. in case of authority-form, we ignore `Host` and use the authority-form).

If we change `getServerName()` to take that behavior into account (i.e. in case of absolute-form and authority-form we don't use the `Host` header, but the given authority instead), do we break much?

sbordet (approved):

PR #8581, opened 2022-09-11
Allow to access underlyng HttpServletRequest from JakartaServerUpgradeRequest

reta:
We are migrating to Jakarta WebSockets (from Jetty WebSockets) and run into some API limitations. In certain context we need access to underlying `HttpServletRequest` associated with particular `UpgradeRequest`. 

From the API perspective, both `ServerUpgradeRequest` and `JettyServerUpgradeRequest`  do allow to access the underlying `HttpServletRequest` , but `JakartaServerUpgradeRequest` does not. The change is pretty trivial and hopefully non-controversial, matching other `UpgradeRequest` implementation on the server side.

### DIFF
--- a/jetty-websocket/websocket-jakarta-server/src/main/java/org/eclipse/jetty/websocket/jakarta/server/internal/JakartaServerUpgradeRequest.java
+++ b/jetty-websocket/websocket-jakarta-server/src/main/java/org/eclipse/jetty/websocket/jakarta/server/internal/JakartaServerUpgradeRequest.java
@@ -16,6 +16,7 @@ package org.eclipse.jetty.websocket.jakarta.server.internal;
 import java.net.URI;
 import java.security.Principal;
 
+import jakarta.servlet.http.HttpServletRequest;
 import org.eclipse.jetty.websocket.core.server.ServerUpgradeRequest;
 import org.eclipse.jetty.websocket.jakarta.common.UpgradeRequest;
 
@@ -44,5 +45,13 @@ public class JakartaServerUpgradeRequest implements UpgradeRequest
     public String getPathInContext()
     {
         return servletRequest.getPathInContext();
+    }
+
+    /**
+     * @return Immutable version of {@link HttpServletRequest}
+     */
+    public HttpServletRequest getHttpServletRequest()
+    {
+        return servletRequest.getHttpServletRequest();
     }
 }


joakime:
What specifically do you need from the `HttpServletRequest`?

Keep in mine that the Jakarta WebSocket API is actually _designed_ to NOT have access to the Jakarta Servlet API as an intentional decision.

reta:
> What specifically do you need from the `HttpServletRequest`?

Thanks @joakime , we do need for some context retrieval operations [1]
 
> Keep in mine that the Jakarta WebSocket API is actually _designed_ to NOT have access to the Jakarta Servlet API as an intentional decision.

`JakartaServerUpgradeRequest`  is a Jetty specific wrapper, no need to modify / alter / change semantics of the Jakarta WebSocket APIs, but I understand your point. From this perspective, replacing hard `HttpServletRequest` with API-neutral getters could be an option to consider?

```
Object getAttribute(String name);
String getContextPath();
```

Thank you.

[1] https://github.com/apache/cxf/blob/main/rt/transports/websocket/src/main/java/org/apache/cxf/transport/websocket/jetty9/Jetty9WebSocketDestination.java#L340

joakime:
> String getContextPath();

You are expected to use `jakarta.websocket.server.HandshakeRequest.getRequestURI()` for this information (what is arriving on the request to upgrade).
Using the above with combination of the `ServerEndpointConfig.Builder` and it's path (what was configured within that context, for that specific websocket endpoint) lets you resolve back to the context path.
Some people go a step further and intentionally use a URI Template based path to allow breaking the path apart even further.

> Object getAttribute(String name);

This is a dangerous assumption when working with `javax.websocket` or `jakarta.websocket`.
There are implementations that will perform websocket upgrade on Servlet after Security, and before Filters are run.

If you are using `javax.websocket` (or `jakarta.websocket`) you cannot expected to have access to any meaningful Attributes (as those are likely never populated by a Filter and/or a HttpServletRequestWrapper)
This behavior was chosen to specifically avoid HTTP behaviors from FIlters and HttpServletRequestWrappers destroying the possibility of being able to Upgrade a WebSocket connection. 
In fact, there's only 1 way for `javax.websocket` (or `jakarta.websocket`) to reliably have request Attributes, and that is to not use a Server based container, but to use a `HttpServletRequest.upgrade()` based implementation of `jakarta.websocket` (like Tyrus), but this has the side effect that all of your Filters (even third party!) has to be aware of the potential for WebSocket upgrade and not do things that could harm the upgrade (like accessing request streams, changing response headers, using any API that can result in a response commit, wrapping the request or response objects, etc).

Be careful here, just because the Jetty implementation is done with a `WebSocketUpgradeFilter` doesn't mean you should rely on that fact, the `javax.websocket` spec and API does not mandate that kind of behavior.  We could easily have websocket upgrade occur between security and the start of the filter chain as well.

This is not unique to Attributes btw, even folks using `HttpSession` hit the same issue, as the only reliable information in the `HttpSession` is usually from a previous, non-websocket request, not the same request that is doing the websocket upgrade.

joakime:
Another common approach is to have the `jakarta.websocket.server.ServerEndpointConfig` (and it's `Configurator`) populated dynamically during Servlet init, passing in values for that upgrade request flow based on configuration and environment that has nothing to do with websocket.    This pre-populated config is then used to `addEndpoint(ServerEndpointConfig)` with all of the information needed to setup the eventual WebSocket endpoint properly for that environment.


Yet another approach is to use ThreadLocal to bring information into the WebSocket upgrade (but I personally don't like this one).

reta:
Thanks a lot for the number of options @joakime , I would explore those as a better way to move forward with migration to Jakarta WebSockets

PR #8595, opened 2022-09-15
Jetty 12 : precompressed content support for `ResourceService`

joakime:
Reenable test cases for precompressed content.

Fix implementation to support precompressed content.


jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/EtagUtil.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /**
+     * Calculate the weak etag for the provided {@link Path} reference.
+     *
+     * @param path the path to calculate from
+     * @param suffix the suffix for the ETag
+     * @return the weak etag
+     */
+    public static String calcWeakEtag(Path path, String suffix)
```
lorban: Shouldn't this method accept a `Resource` instead of a `Path`?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/EtagUtil.java
```diff
@@ -0,0 +1,194 @@
...
+     *
+     * <p>
+     * This supports relative path references as well.
+     * Which can be useful to establish a reliable etag if the base changes.
+     * </p>
+     *
+     * @param path the path to calculate from
+     * @return the weak etag
+     */
+    public static String calcWeakEtag(Path path)
```
lorban: Shouldn't this method accept a `Resource` instead of a `Path`?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/ResourceHttpContent.java
```diff
@@ -119,13 +119,13 @@ public HttpField getETag()
     @Override
     public String getETagValue()
     {
-        return _resource.getWeakETag();
+        return EtagUtil.calcWeakEtag(_resource.getPath());
```
lorban: This would only work with path-backed resources. Shouldn't we be capable of calculating the etag of a potential `MemoryResource`?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/CachedContentFactory.java
```diff
@@ -405,7 +406,7 @@ public class CachedHttpContent implements HttpContent
 
             _lastAccessed = Instant.now();
 
-            _etag = CachedContentFactory.this._etags ? new PreEncodedHttpField(HttpHeader.ETAG, resource.getWeakETag()) : null;
+            _etag = CachedContentFactory.this._etags ? new PreEncodedHttpField(HttpHeader.ETAG, EtagUtil.calcWeakEtag(resource.getPath())) : null;
```
lorban: This would only work with path-backed resources. Shouldn't we be capable of calculating the etag of a potential `MemoryResource`?

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ResourceHandlerTest.java
```diff
@@ -750,8 +749,35 @@ public void testBigger() throws Exception
...
+            \r
+            """);
+        HttpTester.Response response = HttpTester.parseResponse(rawResponse);
+        assertThat(response.toString(), response.getStatus(), is(HttpStatus.OK_200));
+        assertThat(response, containsHeaderValue(HttpHeader.CONTENT_LENGTH, "11"));
+        assertThat(response, containsHeaderValue(HttpHeader.CONTENT_TYPE, "text/plain"));
+        assertThat(response, containsHeaderValue(HttpHeader.VARY, "Accept-Encoding"));
+        assertThat(response, containsHeaderValue(HttpHeader.CONTENT_ENCODING, "br"));
+        String body = response.getContent();
+        assertThat(body, containsString("fake br"));
```
lorban: Why only `fake br`?

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ResourceHandlerTest.java
```diff
@@ -1892,35 +1912,6 @@ public void testDirectoryOfCollections() throws Exception
         assertThat(response.getStatus(), is(HttpStatus.OK_200));
     }
 
-    @Test
-    public void testEtagIfMatchAlwaysFailsDueToWeakEtag() throws Exception
```
lorban, line 1896: Why did you remove this test?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/PrecompressedHttpContent.java
```diff
@@ -51,20 +51,19 @@ public HttpField getETag()
     @Override
     public String getETagValue()
     {
-        //return _content.getResource().getWeakETag(_format.getEtagSuffix());
-        return null;
+        return EtagUtil.calcWeakEtag(_content.getResource().getPath(), _format.getEtagSuffix());
```
lorban: This would only work with path-backed resources. Shouldn't we be capable of calculating the etag of a potential `MemoryResource`?

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/CachedContentFactory.java
```diff
@@ -403,7 +404,7 @@ public class CachedHttpContent implements HttpContent
 
             _lastAccessed = Instant.now();
 
-            _etag = CachedContentFactory.this._etags ? new PreEncodedHttpField(HttpHeader.ETAG, resource.getWeakETag()) : null;
+            _etag = CachedContentFactory.this._etags ? new PreEncodedHttpField(HttpHeader.ETAG, EtagUtil.calcWeakEtag(resource.getPath())) : null;
```
lorban: This would only work with path-backed resources. Shouldn't we be capable of calculating the etag of a potential `MemoryResource`?

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/CachedContentFactory.java
```diff
@@ -603,7 +604,7 @@ public class CachedPrecompressedHttpContent extends PrecompressedHttpContent
             _content = content;
             _precompressedContent = precompressedContent;
 
-            _etag = (CachedContentFactory.this._etags) ? new PreEncodedHttpField(HttpHeader.ETAG, _content.getResource().getWeakETag(format.getEtagSuffix())) : null;
+            _etag = (CachedContentFactory.this._etags) ? new PreEncodedHttpField(HttpHeader.ETAG, EtagUtil.calcWeakEtag(_content.getResource().getPath(), format.getEtagSuffix())) : null;
```
lorban: This would only work with path-backed resources. Shouldn't we be capable of calculating the etag of a potential `MemoryResource`?

lorban (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/EtagUtil.java
joakime: The weak etag is calculated from the path name + last modified timestamp.

Not all resources have them (eg: MemoryResource has neither).
Can't use a ResourceCollection here either, as the path name and last modified timestamp won't always represent the same path object.
It's better to use Path, as even with ResourceCollection in the mix, we'll get a Path then use it.
This is the same Path that is used by the HttpContent to serve the same content.

BTW, Don't get hung up on MemoryResource, the HttpContentFactory reorg PR coming after this will make the case to remove MemoryResource entirely.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/CachedContentFactory.java
joakime: There's no ETag support for MemoryResource as it exists currently.

The HttpContentFactory reorg PR (after this one) will remove MemoryResource in favor of a StaticHttpContent object fallback instead (which can hold the etag based on the URI and JAR file last timestamp)

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ResourceHandlerTest.java
joakime: The contents don't actually have to be compressed to satisfy the test.
The ResourceService doesn't compress it anyway, it just serves the content as-is from the Resource Base.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ResourceHandlerTest.java
joakime: It's a bad/duplicate test.

There are other tests that test the same condition.
testIfEtag() for example.
And those expect a 200 OK, not a 412 precondition failure for the same request.

The origin of this test is unknown.
It is not present in jetty-11.0.x.
I can trace as far back as https://github.com/eclipse/jetty.project/commit/04acdb72f041a15267bb79c36d8fc4d517beaadb but not before then.
It seems to have been created during the hackathon.
But etags and preconditions were not working back then.

Interestingly, the github search also shows nothing for this test case...
https://github.com/eclipse/jetty.project/search?q=testEtagIfMatchAlwaysFailsDueToWeakEtag&type=code

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/PrecompressedHttpContent.java
joakime: Fixing etag support

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/CachedContentFactory.java
joakime: Fixing.

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/CachedContentFactory.java
joakime: Fixing.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/CachedContentFactory.java
joakime: Fixed in this PR anyway.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/ResourceHttpContent.java
joakime: Fixed in this PR.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/EtagUtil.java
joakime: Different method supports `Resource` now, this method still exists.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingContentFactory.java
```diff
@@ -285,18 +295,24 @@ public ByteBuffer getBuffer()
 
         public boolean isValid()
         {
-            try
+            if (_delegate.getResource() != null)
             {
-                FileTime lastModifiedTime = Files.getLastModifiedTime(_delegate.getResource().getPath());
-                if (lastModifiedTime.equals(_lastModifiedValue))
+                if (!_delegate.getResource().exists())
+                    return false;
```
sbordet: Aren't these redundant checks now that the constructor checks them?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CompressedContentFormat.java
```diff
@@ -15,20 +15,10 @@
...
-     * So etag for compressed resource that normally has an etag of <code>W/"28c772d6"</code>
-     * is <code>W/"28c772d6--gzip"</code>.  The separator may be changed by the
-     * "org.eclipse.jetty.http.CompressedContentFormat.ETAG_SEPARATOR" System property. If changed, it should be changed to a string
-     * that will not be found in a normal etag or at least is very unlikely to be a substring of a normal etag.
-     */
-    public static final String ETAG_SEPARATOR = System.getProperty(CompressedContentFormat.class.getName() + ".ETAG_SEPARATOR", "--");
-
     public static final CompressedContentFormat GZIP = new CompressedContentFormat("gzip", ".gz");
     public static final CompressedContentFormat BR = new CompressedContentFormat("br", ".br");
     public static final CompressedContentFormat[] NONE = new CompressedContentFormat[0];
```
sbordet, line 24: Why there is not one for `zip`? Tests are continuously creating the `zip` format when there could be a constant here.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/EtagUtil.java
```diff
@@ -0,0 +1,251 @@
...
+import java.util.Objects;
+
+import org.eclipse.jetty.util.QuotedStringTokenizer;
+import org.eclipse.jetty.util.StringUtil;
+import org.eclipse.jetty.util.resource.Resource;
+
+/**
+ * Utility classes for Etag behaviors
+ */
+public class EtagUtil
```
sbordet: Contains more than one utility, so I'd prefer `EtagUtils`.

Also, add a `private` empty constructor so it cannot be instantiated.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/EtagUtil.java
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    /**
+     * Calculate the weak etag for the provided {@link Path} reference.
+     *
+     * @param path the path to calculate from
+     * @param suffix the suffix for the ETag
+     * @return the weak etag
+     */
+    public static String calcWeakEtag(Path path, String suffix)
```
sbordet: Prefix `calc` implies calculations, which to me implies math operations only.
I think something like `compute` or `resolve` instead of `calc` is more appropriate.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceContentFactory.java
```diff
@@ -91,10 +91,14 @@ private HttpContent load(String pathInContext, Resource resource)
...
-                    compressedResource.length() < resource.length())
-                    compressedContents.put(format,
-                        new ResourceHttpContent(compressedResource, _mimeTypes.getMimeByExtension(compressedPathInContext)));
+                if (compressedResource == null || !compressedResource.exists())
+                    continue; // skip, doesn't exist
+                if (compressedResource.lastModified().isBefore(resource.lastModified()))
+                    continue; // skip, compressed resource modified before actual resource
+                if (compressedResource.length() >= resource.length())
+                    continue; // skip, not smaller than actual resource
+                // add found precompressed format
```
sbordet: Remove the comments, they are just describing what the code does, and the code is very clear.
Less chances that the comments rot.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/EtagUtil.java
```diff
@@ -0,0 +1,251 @@
...
+        catch (IOException ignored)
+        {
+            // TODO: should we ignore the inability to read last modified or size?
+        }
+        b.append(encoder.encodeToString(longToBytes(lastModifiedMillis ^ lhash)));
+        b.append(encoder.encodeToString(longToBytes(size ^ lhash)));
+        if (suffix != null)
+            b.append(suffix);
+        b.append('"');
+        return b.toString();
```
sbordet: Is there a reference for this algorithm? Is it defined by some spec? If so, please add a reference to it.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/EtagUtil.java
sbordet: The original implementation in `Resource` was not using `Path` at all, so I have the same concerns: this method should take a `Resource`, not a `Path`.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ResourceHandlerTest.java
sbordet: But it confuses the reader the fact that you test for `fake br` only, while at the beginning of the test the whole content is `fake brotli`. 
I would test for the whole content, since it's so small, to avoid such scratching head moment from other readers of this test ("why only `fake br`"?).

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/EtagUtil.java
```diff
@@ -0,0 +1,251 @@
...
+
+    /**
+     * Test a match of an etag against an etagWithOptionalSuffix.
+     *
+     * @param etag An etag without a compression suffix
+     * @param etagWithOptionalSuffix An etag optionally with a compression suffix.
+     * @return True if the tags are equal.
+     * @see <a href="https://www.rfc-editor.org/rfc/rfc9110#section-8.8.3.2">RFC9110: Section 8.8.3.2 - Etag Comparison</a>.
+     */
+    public static boolean match(String etag, String etagWithOptionalSuffix)
```
sbordet: I would call this method `matches` because it returns a boolean.
It's kind of a query/question rather than a command.

sbordet (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CompressedContentFormat.java
joakime: zip is an oddball one.
It's not a supported standard for Accept-Encoding / Content-Encoding

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/EtagUtil.java
joakime: The `Resource` version didn't handle non-path/file-system based Resources properly.

Btw, there is a `Resource` method in this class for working with `Resource` to `HttpField`.

joakime: There's also the problem when working with a `ResourceCollection` there is a difference in `.getPath()` and `.getLastModified()` which doesn't match the Resource actually being served in the `HttpContent`

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/EtagUtil.java
joakime: The Etag is just a string.
The bits and pieces we use are just our take on how to create that string.
I'll add the link to the RFC9110 section in the top level javadoc.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingContentFactory.java
```diff
@@ -285,18 +295,24 @@ public ByteBuffer getBuffer()
...
+
+                try
                 {
-                    _lastAccessed = NanoTime.now();
-                    return true;
+                    FileTime lastModifiedTime = Files.getLastModifiedTime(_delegate.getResource().getPath());
```
lorban: Since this makes caching only work with path-backed resources, shouldn't there be a check in the constructor to prevent accepting those?

lorban (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingContentFactory.java
joakime: Good point, let me fix this code block.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingContentFactory.java
joakime: Done

lorban (approved):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CompressedContentFormat.java
sbordet: Fair enough.

sbordet (approved):

PR #8597, opened 2022-09-16
Jetty 12 - Issue #8474 - return null on resolve() if resource doesn't exist

joakime:
* makes `Resource.resolve(String)` return null if the resulting `Resource` would not exist.
* makes `ResourceFactory.newResource()` (and similar methods) return null if the resulting `Resource` would not exist.


joakime:
@sbordet here's the minimal change to have PathResource.resolve() behave the same as ResourceCollection.resolve().

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -318,6 +318,48 @@ public int hashCode()
...
+        // Sub-paths are always resolved under the given URI,
+        // we compensate for input sub-paths like "/subdir"
+        // where default resolve behavior would be to treat
+        // that like an absolute path.
+        while (subUriPath.startsWith(URIUtil.SLASH))
+        {
+            // TODO XXX this appears entirely unnecessary and inefficient.  We already have utilities
+            //      to handle appending path strings with/without slashes.
+            subUriPath = subUriPath.substring(1);
+        }
```
gregw, line 280: I also think we have rewrite rules for removing // in user supplied URIs.   I really don't think we need to do this in resolve and I'd like to see what tests break if we remove it.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -318,6 +318,48 @@ public int hashCode()
...
+
+        URI uri = getURI();
+        URI resolvedUri = URIUtil.addPath(uri, subUriPath);
+        Path path = Paths.get(resolvedUri.normalize());
```
gregw, line 284: This is probably needed.... but I'd still be interested to know exactly why.  i.e what fails if we don't do this.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -318,6 +318,48 @@ public int hashCode()
         return Objects.hashCode(path);
     }
 
+    @Override
+    public Resource resolve(String subUriPath)
```
gregw, line 259: I like having a resolve in PathResource rather than try to be generic in Resource

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -295,30 +295,8 @@ public List<Resource> list()
...
-            // TODO XXX this appears entirely unnecessary and inefficient.  We already have utilities
-            //      to handle appending path strings with/without slashes.
-            subUriPath = subUriPath.substring(1);
-        }
-
-        URI uri = getURI();
-        URI resolvedUri = URIUtil.addPath(uri, subUriPath);
-        return create(resolvedUri);
+        // abstract returns null always.
+        return null;
```
gregw: Can it be abstract?

gregw (changes requested):
This is a bit of an omnibus PR.  There are too many changes in it that are unrelated to the premise (of a null return).

There is goodness in this PR, but it is too hard to review when there are 70 files changed with renames, reformatting, various fixes and cleanups.

Can you break this out into multiple PRs, so that ultimately there is a single PR with the null return change.

joakime:
Thank you for the preliminary review.

For now, hold off on any further reviews until I can move some of the oddball changes out of this PR into separate PRs (as once those are merged, it will cause a rebase in this PR to clean out those changes)

joakime:
The removal of `exist()` has not been done on this PR.

gregw:
> Thank you for the preliminary review.
> 
> For now, hold off on any further reviews until I can move some of the oddball changes out of this PR into separate PRs (as once those are merged, it will cause a rebase in this PR to clean out those changes)

For anything like renaming of variables / fields etc.   I think you can probably do them without a PR if they are not part of the public API.  Just check with another developer and if they are OK, push directly to the branch.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/ResourceTest.java
```diff
@@ -252,30 +259,25 @@ public static Stream<Arguments> scenarios() throws Exception
 
     public WorkDir workDir;
 
+    @Disabled
```
joakime, line 262: Need to revisit this

joakime:
From latest run https://jenkins.webtide.net/blue/organizations/jenkins/jetty.project/detail/PR-8597/10/tests

**org.eclipse.jetty.tests.distribution.DemoModulesTests.testJspDump([1] ee8) Test Failure**

```
java.lang.IllegalArgumentException: KeyStore Path does not exist: /home/jenkins/agent/workspace/jetty.project_PR-8597
  /tests/test-distribution/test-distribution-common/target/tests/test-bases/jetty_base_5828043306936187437/etc/test-keystore.p12
```

I wonder if this should be throwing, or not.
I'm leaning towards throwing.  (and fixing the testcase to actually have a keystore)

@sbordet and @lorban what do you think?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -126,8 +126,8 @@ public Resource resolve(String subUriPath)
         for (Resource res : _resources)
         {
             addedResource = res.resolve(subUriPath);
-            if (!addedResource.exists())
-                continue;
+            if (addedResource == null)
```
gregw, line 129: I think more changes below  are needed for code clarity.

specifically if resources == null, can we just explicitly return null?

jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ScanningAppProvider.java
```diff
@@ -226,7 +226,7 @@ protected void doStart() throws Exception
         List<Path> files = new ArrayList<>();
         for (Resource resource : _monitored)
         {
-            if (resource.exists() && Files.isReadable(resource.getPath()))
+            if (resource != null && Files.isReadable(resource.getPath()))
```
lorban, line 229: We are not adding `null`s into `_monitored` for non-existing resources, so this null check can go away.

jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ScanningAppProvider.java
```diff
@@ -339,7 +339,11 @@ public void setDeploymentManager(DeploymentManager deploymentManager)
     public void setMonitoredResources(List<Resource> resources)
     {
         _monitored.clear();
-        _monitored.addAll(resources);
+        for (Resource resource: resources)
+        {
+            if (resource != null)
+                _monitored.addAll(resources);
```
lorban, line 345: Why adding the `resources` collection for each resource?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -720,7 +720,18 @@ public void setBaseResource(Resource resourceBase)
...
     {
-        setBaseResource(path == null ? null : ResourceFactory.root().newResource(path));
+        if (path == null)
+        {
+            // allow user to unset variable
+            setBaseResource((Resource)null);
+            return;
+        }
+
+        Resource resource = ResourceFactory.root().newResource(path);
```
lorban, line 730: The `Path` may require mounting, right? So this may lead to a mount leak if I'm not mistaken.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/MountedPathResource.java
```diff
@@ -26,6 +28,14 @@
  */
 public class MountedPathResource extends PathResource
 {
+    public static MountedPathResource of(URI uri) throws IOException
```
lorban, line 31: Is `public` needed?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -47,6 +47,14 @@ public class PathResource extends Resource
         .with("jrt")
         .build();
 
+    public static PathResource of(URI uri) throws IOException
```
lorban, line 50: Is `public` needed?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/MountedPathResource.java
```diff
@@ -34,6 +44,18 @@ public class MountedPathResource extends PathResource
         containerUri = URIUtil.unwrapContainer(getURI());
     }
 
+    MountedPathResource(Path path, URI uri)
```
lorban, line 47: Is the other constructor still needed?

lorban (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: This is addressed in PR #8631

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: There is a behavior different with regards what is remembered by `Path` when you use `Paths.get(URI)` or `Paths.of(URI)` vs `Paths.of(String)`.
Path segments in a URI like `/../` are remembered and will impact behavior like `Files.exists(path)`.
This seems to be because a URI is seen as canonical and the raw `/../` is passed all the way down to the real filesystem.
Be happy to demonstrate with you later.

joakime:
Closing this large (in commit count) PR in favor of #8702

PR #8610, opened 2022-09-22
Jetty 12 : ResourceFactory.of(WebAppContext) to WebAppContext.getResourceFactory()

joakime:
Use `WebAppContext.getResourceFactory()` instead of `ResourceFactory.of(container)` where possible.

Taken from PR #8597


lorban (approved):
LGTM

PR #8611, opened 2022-09-22
Jetty 12 : `Descriptor` cleanup

joakime:
+ Sanity check Resource being provided
+ Descriptor toString changed to include full URI to descriptor so that it can be used consistently in Logging and Exception messages.
+ New Descriptor.toURI() for use in Servlet Source and other logging messages
+ Servlet Source value is now the URI of the Descriptor
+ Cleanup of test cases

Taken from PR #8597

### DIFF
--- a/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/Source.java
+++ b/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/Source.java
@@ -13,6 +13,10 @@
 
 package org.eclipse.jetty.ee10.servlet;
 
+import java.util.Objects;
+
+import org.eclipse.jetty.util.resource.Resource;
+
 /**
  * Source
  *
@@ -20,8 +24,8 @@ package org.eclipse.jetty.ee10.servlet;
  */
 public class Source
 {
-    public static final Source EMBEDDED = new Source(Origin.EMBEDDED, null);
-    public static final Source JAVAX_API = new Source(Origin.JAKARTA_API, null);
+    public static final Source EMBEDDED = new Source(Origin.EMBEDDED);
+    public static final Source JAVAX_API = new Source(Origin.JAKARTA_API);
 
     public enum Origin
     {
@@ -29,19 +33,49 @@ public class Source
         JAKARTA_API, DESCRIPTOR, ANNOTATION
     }
 
-    public Origin _origin;
-    public String _resource;
+    public final Origin _origin;
+    public final String _name;
+    public Resource _resource;
+
+    /**
+     * A Source without a name/location.
+     *
+     * @param o the Origin of the artifact (servlet, filter, mapping etc)
+     */
+    public Source(Origin o)
+    {
+        this(o, (String)null);
+    }
+
+    /**
+     * @param o the Origin of the artifact (servlet, filter, mapping etc)
+     * @param clazz the class where the artifact was declared
+     */
+    public Source(Origin o, Class<?> clazz)
+    {
+        this(o, clazz.getName());
+    }
 
     /**
      * @param o the Origin of the artifact (servlet, filter, mapping etc)
      * @param resource the location where the artifact was declared
      */
-    public Source(Origin o, String resource)
+    public Source(Origin o, Resource resource)
+    {
+        this(o, Objects.requireNonNull(resource).getURI().toASCIIString());
+        _resource = resource;
+    }
+
+    /**
+     * @param o the Origin of the artifact (servlet, filter, mapping etc)
+     * @param name the name of the location where the artifact was declared (not a {@link Resource})
+     */
+    public Source(Origin o, String name)
     {
         if (o == null)
             throw new IllegalArgumentException("Origin is null");
         _origin = o;
-        _resource = resource;
+        _name = name;
     }
 
     /**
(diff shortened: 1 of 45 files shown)


jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/Source.java
```diff
@@ -13,35 +13,69 @@
...
+    }
 
     /**
      * @param o the Origin of the artifact (servlet, filter, mapping etc)
      * @param resource the location where the artifact was declared
      */
-    public Source(Origin o, String resource)
+    public Source(Origin o, Resource resource)
+    {
+        this(o, Objects.requireNonNull(resource).getURI().toASCIIString());
```
sbordet: No need for `Objects.requireNonNull()` -- if it is `null` it will throw NPE like `Objects.requireNonNull()` does.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/FragmentConfiguration.java
```diff
@@ -24,6 +24,9 @@
  */
 public class FragmentConfiguration extends AbstractConfiguration
 {
+    // Holds a Map<Resource, Resource> .
+    //  key: Resource to the Jar
+    //  value: Resource to the web fragment xml
     public static final String FRAGMENT_RESOURCES = "org.eclipse.jetty.webFragments";
```
sbordet: This constant should now have a `ee10` specifier, since there is an equivalent class for `ee9`.
I would use the package name, eg. `org.eclipse.jetty.ee10.webapp.webFragments`.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/StandardDescriptorProcessor.java
```diff
@@ -176,7 +176,7 @@ public void visitContextParam(WebAppContext context, Descriptor descriptor, XmlP
                 if (descriptor instanceof FragmentDescriptor)
                 {
                     if (!((String)context.getInitParams().get(name)).equals(value))
-                        throw new IllegalStateException("Conflicting context-param " + name + "=" + value + " in " + descriptor.getResource());
+                        throw new IllegalStateException("Conflicting context-param " + name + "=" + value + " in " + descriptor);
```
sbordet: I'd use `descriptor.getURI()` here, since it's better than having in the exception message the descriptor's `toString()` which contains className, hashCode, etc.
See ee10's `MetaData.addFragmentDescriptor(...)`.

jetty-ee9/jetty-ee9-plus/src/main/java/org/eclipse/jetty/ee9/plus/webapp/PlusDescriptorProcessor.java
```diff
@@ -142,7 +142,7 @@ public void visitEnvEntry(WebAppContext context, Descriptor descriptor, XmlParse
             case WebFragment:
             {
                 //ServletSpec p.75. No declaration in web.xml, but in multiple web-fragments. Error.
-                throw new IllegalStateException("Conflicting env-entry " + name + " in " + descriptor.getResource());
+                throw new IllegalStateException("Conflicting env-entry " + name + " in " + descriptor);
```
sbordet: I'd use `descriptor.getURI()` for better exception messages.

jetty-ee9/jetty-ee9-webapp/src/main/java/org/eclipse/jetty/ee9/webapp/FragmentConfiguration.java
```diff
@@ -24,6 +24,9 @@
  */
 public class FragmentConfiguration extends AbstractConfiguration
 {
+    // Holds a Map<Resource, Resource> .
+    //  key: Resource to the Jar
+    //  value: Resource to the web fragment xml
     public static final String FRAGMENT_RESOURCES = "org.eclipse.jetty.webFragments";
```
sbordet: Same comment as ee10's version of this class.

jetty-ee9/jetty-ee9-webapp/src/main/java/org/eclipse/jetty/ee9/webapp/StandardDescriptorProcessor.java
```diff
@@ -175,7 +175,7 @@ public void visitContextParam(WebAppContext context, Descriptor descriptor, XmlP
                 if (descriptor instanceof FragmentDescriptor)
                 {
                     if (!((String)context.getInitParams().get(name)).equals(value))
-                        throw new IllegalStateException("Conflicting context-param " + name + "=" + value + " in " + descriptor.getResource());
+                        throw new IllegalStateException("Conflicting context-param " + name + "=" + value + " in " + descriptor);
```
sbordet: Use `getURI()` for better exception message (also in the rest of the class).

sbordet (changes requested):

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/StandardDescriptorProcessor.java
joakime: ok, but the Descriptor.toString() won't contain classname / hashcode / etc.

It's the Simple classname (last segment) and the name of the identifier it's tracking (classname, string, resource, etc)

sbordet (approved):

PR #8612, opened 2022-09-22
Jetty 12 : QuickStart Generation uses Path instead of Resource

joakime:
This change is in preparation for the resolve() returning null if doesn't exist PR #8597 

+ QuickStart will use a Path object to handle creation / regeneration of the quickstart-web.xml
+ Updated maven plugin usage as well.
+ Depends on path based Descriptor impl

Note: this PR based on PR #8611 (and will require it to function)

Taken from PR #8597


joakime:
Will rework against `jetty-12.0.x` instead

Issue #8613, opened 2022-09-22
Update build for Java 19


sbordet:
**Jetty version(s)**
10+

**Java version/vendor** `(use: java -version)`
19

**Description**
In Java 18, some API that we use was incubating but available without preview.
In Java 19, these APIs have moved to preview, so we need to update the POM files with the right profiles, especially the `quic-quiche` module that relies on the `foreign` APIs.

@lorban @olamy can you please look into this?

lorban:
Regarding the Foreign binding of Quiche, it's not going to be an easy change. The package changed from `jdk.incubator.foreign` to `java.lang.foreign` but there were other breaking API changes too.

I'm going to run `jextract` again on the quiche header file and update our current binding to make it work with JDK 19. But this new binding should replace the existing one; we don't want to keep both the incubator and final bindings, right?

lorban:
Interesting: jextract is now an [open-source project](https://github.com/openjdk/jextract/) but I could not find a pre-built binary. Building it myself is going to need a slightly bigger time investment.

sbordet:
@lorban yes we want to keep the Java 17 version, which I think is also valid for 18, and the Java 19 version.

They will be different modules under `jetty-quic-quiche` parent.

We have done this for years for ALPN for almost every JDK version, we can do 2 modules for foreign with the proper POM config and Jetty module config.

lorban:
@sbordet unfortunately, our JDK 17 impl is not compatible with JDK 18. Maybe adapting it wouldn't take much but it definitely isn't source nor binary compatible.

olamy:
we are now building with 21 :)

PR #8617, opened 2022-09-23
Disable building quiche foreign incubator binding builds on JDKs > 17

lorban:
The current quiche-foreign-incubator implementation only works with JDK 17, so disable it on other JDKs. See #8613.

This is a quick stopgap measure to allow building Jetty 12 with JDKs > 17. A proper rework of this quiche binding will be done at a latter stage.

### DIFF
--- a/jetty-core/jetty-quic/jetty-quic-quiche/pom.xml
+++ b/jetty-core/jetty-quic/jetty-quic-quiche/pom.xml
@@ -20,7 +20,7 @@
     <profile>
       <id>jdk17</id>
       <activation>
-        <jdk>[17,)</jdk>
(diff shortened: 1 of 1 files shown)


joakime (approved):
Is there a plan to make it compatible with JDKs past 17 in a different way?

lorban:
My plan is to rework our Foreign binding to use the final API as published with JDK 19 and give up supporting the foreign API on any earlier JDK. See https://github.com/eclipse/jetty.project/issues/8618.

sbordet:
@lorban does the Java 17 implementation work with Java 18?

lorban:
@sbordet no, unfortunately it does not.

PR #8620, opened 2022-09-26
Introduced [Callback|Promise]Completable.with(Consumer) to simplify u…

sbordet:
…sage of CompletableFuture APIs.

Updated code to use the new APIs.

### DIFF
--- a/jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/MultiPartCaptureTest.java
+++ b/jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/MultiPartCaptureTest.java
@@ -244,9 +244,8 @@ public class MultiPartCaptureTest
             List<MultiPart.Part> charSetParts = allParts.get("_charset_");
             if (charSetParts != null)
             {
-                Promise.Completable<String> promise = new Promise.Completable<>();
-                Content.Source.asString(charSetParts.get(0).getContent(), StandardCharsets.US_ASCII, promise);
-                defaultCharset = promise.get();
+                defaultCharset = Promise.Completable.<String>with(p -> Content.Source.asString(charSetParts.get(0).getContent(), StandardCharsets.US_ASCII, p))
+                    .get();
             }
 
             for (NameValue expected : partContainsContents)
--- a/jetty-core/jetty-proxy/src/test/java/org/eclipse/jetty/proxy/InterimResponseProxyTest.java
+++ b/jetty-core/jetty-proxy/src/test/java/org/eclipse/jetty/proxy/InterimResponseProxyTest.java
@@ -44,20 +44,10 @@ public class InterimResponseProxyTest extends AbstractProxyTest
             public void process(Request request, Response response, Callback callback)
             {
                 CompletableFuture<Void> completable = response.writeInterim(HttpStatus.CONTINUE_100, HttpFields.EMPTY)
-                    .thenCompose(ignored ->
-                    {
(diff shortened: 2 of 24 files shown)


jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/MultiPartCaptureTest.java
```diff
@@ -244,9 +244,8 @@ private void assertParts(Map<String, List<MultiPart.Part>> allParts) throws Exce
             List<MultiPart.Part> charSetParts = allParts.get("_charset_");
             if (charSetParts != null)
             {
-                Promise.Completable<String> promise = new Promise.Completable<>();
-                Content.Source.asString(charSetParts.get(0).getContent(), StandardCharsets.US_ASCII, promise);
-                defaultCharset = promise.get();
+                defaultCharset = Promise.Completable.<String>with(p -> Content.Source.asString(charSetParts.get(0).getContent(), StandardCharsets.US_ASCII, p))
```
lorban, line 247: Too bad we can't have `Content.Source.asString(Source, Charset)` that returns a `CompletableFuture`.

jetty-core/jetty-proxy/src/test/java/org/eclipse/jetty/proxy/InterimResponseProxyTest.java
```diff
@@ -44,20 +44,10 @@ public void testInterimResponses() throws Exception
             public void process(Request request, Response response, Callback callback)
             {
                 CompletableFuture<Void> completable = response.writeInterim(HttpStatus.CONTINUE_100, HttpFields.EMPTY)
-                    .thenCompose(ignored ->
-                    {
-                        Promise.Completable<String> p = new Promise.Completable<>();
-                        Content.Source.asString(request, StandardCharsets.UTF_8, p);
-                        return p;
-                    })
+                    .thenCompose(ignored -> Promise.Completable.<String>with(p -> Content.Source.asString(request, StandardCharsets.UTF_8, p)))
```
lorban, line 47: It really show here that a `Content.Source.asString` that returns a `CompletableFuture` would be more elegant.

jetty-core/jetty-proxy/src/test/java/org/eclipse/jetty/proxy/InterimResponseProxyTest.java
```diff
@@ -44,20 +44,10 @@ public void testInterimResponses() throws Exception
...
                     .thenCompose(content -> response.writeInterim(HttpStatus.PROCESSING_102, HttpFields.EMPTY).thenApply(ignored -> content))
                     .thenCompose(content -> response.writeInterim(HttpStatus.EARLY_HINT_103, HttpFields.EMPTY).thenApply(ignored -> content))
-                    .thenCompose(content ->
-                    {
-                        Callback.Completable c = new Callback.Completable();
-                        Content.Sink.write(response, true, content, c);
-                        return c;
-                    });
+                    .thenCompose(content -> Callback.Completable.with(c -> Content.Sink.write(response, true, content, c)));
```
lorban, line 50: A `Content.Sink.write` that returns a `CompletableFuture` would be more elegant here too.

lorban (approved):
`Callback.Completable.with()` is a really nice improvement as it allows using the `CompletableFuture` API fluently.

The only small annoyance is that using `with()` in a fluent call chain lowers readability a bit, I wish we always provided a method that returns `CompletableFuture` as it would make the fluent call chains crystal clear.

PR #8621, opened 2022-09-27
Fix Caching ContentFactories in Jetty-12

lachlan-roberts:
This PR removes the old `CachedContentFactory` which was only used in tests, and brings over some of its functionality to the `CachingContentFactory`.

Fixes the `CachingContentFactory` to make it pass tests, and re-enable the tests.


jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -365,8 +365,10 @@ protected boolean passConditionalHeaders(Request request, Response response, Htt
                         String matched = matchesEtag(etag, ifnm);
                         if (matched != null)
                         {
+                            // If we do sendError it clears the ETag header.
+                            response.setStatus(HttpStatus.NOT_MODIFIED_304);
                             response.getHeaders().put(HttpHeader.ETAG, matched);
-                            writeHttpError(request, response, callback, HttpStatus.NOT_MODIFIED_304);
+                            response.write(true, null, callback);
```
joakime: This needs to be handed to `writeHttpError` so that the DefaultServlet can override and use the Servlet `HttpServletResponse.sendError()` API.

Otherwise application provided filters, security, wrappers, etc are not executed.

joakime (dismissed):
Restore the writeHttpError usage.

Note that the removal of a header from the response has to be done through the Servlet API (in case of wrapping), so that raw ETAG removal also needs to be done at the Servlet level.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -365,8 +365,10 @@ protected boolean passConditionalHeaders(Request request, Response response, Htt
                         String matched = matchesEtag(etag, ifnm);
                         if (matched != null)
                         {
+                            // If we do sendError it clears the ETag header.
+                            response.setStatus(HttpStatus.NOT_MODIFIED_304);
                             response.getHeaders().put(HttpHeader.ETAG, matched);
```
joakime: This removal has to happen at the Servlet level when called from the DefaultServlet using the `HttpServletResponse.setHeader("ETag", null)`

lachlan-roberts: The header should not be removed, we have tests checking that the ETag header is received in the response. The sendError eventually does a resetContent on the response which clears out specific headers including ETag. 

In Jetty 11 we were setting the status, setting the header and flushing the buffer, so this never went through sendError at all. But I can revert my change and put specific logic in `ServletContextResponse.resetContent()` so this can still go through sendError and keep the ETag header.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingContentFactory.java
```diff
@@ -163,6 +163,30 @@ public void flushCache()
         }
     }
 
+    /**
+     * @param httpContent the resource to test
+     * @return whether the httpContent is cacheable. The default implementation tests the cache sizes.
```
sbordet: The javadoc description is not accurate.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/PrecompressedHttpContent.java
```diff
@@ -139,12 +139,12 @@ public Map<CompressedContentFormat, HttpContent> getPrecompressedContents()
...
     {
-        return _content.getBuffer();
+        return _precompressedContent.getBuffer();
     }
 
     @Override
     public void release()
     {
-        _content.release();
+        _precompressedContent.release();
```
sbordet, line 148: Should it not release both `_content` and `_precompressedContent`?

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/ResourceContentFactory.java
```diff
@@ -98,6 +98,17 @@ private HttpContent load(String pathInContext, Resource resource)
...
+     * <p>Utility to compare {@link Resource#lastModified()} of two resources.</p>
+     * @param resource1 the first resource to test.
+     * @param resource2 the second resource to test.
+     * @return true if modified time of resource1 is newer or equal to that of resource2.
+     */
+    static boolean newerThanOrEqual(Resource resource1, Resource resource2)
+    {
+        return !resource2.lastModified().isAfter(resource1.lastModified());
+    }
+
```
sbordet, line 111: I found the original version more readable, but ok.
Rename to `isNewerThanOrEqual()`.

sbordet (commented):
Minor things, please fix them, but LGTM.

sbordet (approved):

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/ResourceContentFactory.java
lachlan-roberts: The original version was wrong and was causing many test failures.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/PrecompressedHttpContent.java
lachlan-roberts: I don't know. The `_content` will be accessible separately to the pre-compressed version and will have separate entry in the cache, so maybe this would result in a double release?

At the moment it looks there are no implementations of `HttpContent.release()` which actually do anything, so this currently has no effect.

lorban (approved):
`HttpContent` should implement `Retainable` otherwise we may leak resources or release them while they're in use. But that should be addressed in another PR IMHO.

PR #8630, opened 2022-09-29
Jetty 12 - Restrict `ResourceCollection` to only hold directory entries

joakime:
This is the experiment to see what happens when `ResourceCollection` is only allowed to hold directories.


gregw (changes requested):
The only tests failing are the newly added ones that explictly add a non directory resource.   So this looks good to me.... once those tests are modified.

gregw (approved):

PR #8631, opened 2022-09-30
Jetty 12 - Simplification of aliases in `PathResource`

lachlan-roberts:
- The `resolveTargetPath()` method has been simplified.
- The method `getTargetPath()` now always returns the target resource with all aliases resolved.
- The check `isAlias()` should be used to determine whether it is an alias instead of `getTargetPath() != null`.


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -142,8 +142,8 @@ public static boolean isSameName(Path pathA, Path pathB)
 
     PathResource(URI uri, boolean bypassAllowedSchemeCheck)
     {
-        // normalize to referenced location, Paths.get() doesn't like "/bar/../foo/text.txt" style references
-        // and will return a Path that will not be found with `Files.exists()` or `Files.isDirectory()`
+        // Normalize to referenced location, Paths.get() doesn't like "/bar/../foo/text.txt" style references
+        // and will return a Path that will not be found with `Files.exists()` or `Files.isDirectory()`.
         this(Paths.get(uri.normalize()), uri, bypassAllowedSchemeCheck);
```
lachlan-roberts, line 148: This `Paths.get(uri.normalize())` will both normalize and decode the URI. This means it could already be an alias for the true URI passed in.

Because of this we need the check at line 283 `else if (!URIUtil.equalsIgnoreEncodings(uri, toUri(path)))`.

I don't like this, but it makes you able to use the `PathResource.getPath()` to access the file, because otherwise using it you could get `FileNotFoundException` if it is an unnormalized path.

If we always used methods like `getTargetUri()` and `getTargetPath()` to access the file then this would not be an issue as we'd be accessing the resolved location of the resource.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/FileSystemResourceTest.java
```diff
@@ -1143,21 +1134,24 @@ public void testResolveWindowsSlash() throws Exception
...
-            assertThat("Target URI: " + basePath, base, hasNoTargetURI());
+            assertThat("Target URI: " + basePath, base, isNotAlias());
 
             Resource r = base.resolve("aa%5C/foo.txt");
-            assertThat("getURI()", r.getPath().toString(), containsString("aa\\/foo.txt"));
-            assertThat("getURI()", r.getURI().toASCIIString(), containsString("aa%5C/foo.txt"));
-
             if (org.junit.jupiter.api.condition.OS.WINDOWS.isCurrentOs())
             {
+                assertThat("getURI()", r.getPath().toString(), containsString("aa\\foo.txt"));
```
lachlan-roberts, line 1153: This test was already failing on 12.0.x branch when run on windows. 
So this isn't directly related to the other changes.

joakime: Why is it `getURI()` and then using path to validate the string?

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/FileSystemResourceTest.java
```diff
@@ -1357,7 +1351,7 @@ public void testUncPath() throws Exception
         Resource resource = base.resolve("WEB-INF/");
         assertThat("getURI()", resource.getURI().toASCIIString(), containsString("path/WEB-INF/"));
         assertThat("isAlias()", resource.isAlias(), is(false));
-        assertThat("getTargetURI()", resource.getTargetURI(), nullValue());
+        assertThat(resource, isNotAlias());
```
joakime:
``` java
assertThat("Target URI: " + resource, resource, isNotAlias());
```

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/FileSystemResourceTest.java
```diff
@@ -1288,13 +1282,13 @@ public void testResolveDoubleSlash() throws Exception
...
             assertThat("Exists: " + basePath, base.exists(), is(true));
-            assertThat("Target URI: " + basePath, base, hasNoTargetURI());
+            assertThat("Target URI: " + basePath, base, isNotAlias());
 
             Resource r = base.resolve("aa//foo.txt");
             assertThat("getURI()", r.getURI().toASCIIString(), containsString("aa/foo.txt"));
 
             assertThat("isAlias()", r.isAlias(), is(false));
-            assertThat("getTargetURI()", r.getTargetURI(), nullValue());
+            assertThat(r, isNotAlias());
```
joakime:
``` java
assertThat("Target URI: " + r, r, isNotAlias());
```

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/FileSystemResourceTest.java
```diff
@@ -1256,13 +1250,13 @@ public void testResolveInitialDoubleSlash() throws Exception
...
             assertThat("Exists: " + basePath, base.exists(), is(true));
-            assertThat("Target URI: " + basePath, base, hasNoTargetURI());
+            assertThat("Target URI: " + basePath, base, isNotAlias());
 
             Resource r = base.resolve("//foo.txt");
             assertThat("getURI()", r.getURI().toASCIIString(), containsString("/foo.txt"));
 
             assertThat("isAlias()", r.isAlias(), is(false));
-            assertThat("getTargetURI()", r.getTargetURI(), nullValue());
+            assertThat(r, isNotAlias());
```
joakime:
``` java
assertThat("Target URI: " + r, r, isNotAlias());
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -256,75 +267,45 @@ public boolean isContainedIn(Resource r)
...
         {
-            if (Files.isSymbolicLink(path))
-                return path.getParent().resolve(Files.readSymbolicLink(path));
-            if (Files.exists(path))
-            {
-                Path real = abs.toRealPath();
-                if (!isSameName(abs, real))
-                    return real;
-            }
+            return path.normalize().toRealPath();
```
joakime, line 307: `normalize()` shouldn't be necessary here.
The `toRealPath()` should do that for you.
Got an example of where it doesn't?

joakime (changes requested):
Questions about normalize, and some nits about consistency in test cases.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: Also, if `path` does not exist, this will thrown an exception.

Example:
``` java
        Path nonexistent = Paths.get("/tmp/does/not/exist");
        System.out.println("nonexistent.exists() = " + Files.exists(nonexistent));
        Path real = nonexistent.toRealPath();
        System.out.println("real = " + real);
        System.out.println("real.exists() = " + Files.exists(real));
```

would result in ...

```
nonexistent.exists() = false
Exception in thread "main" java.nio.file.NoSuchFileException: /tmp/does/not/exist
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
	at java.base/sun.nio.fs.UnixPath.toRealPath(UnixPath.java:825)
	at fs.RealPathTest.main(RealPathTest.java:14)
```

This should not return the `path.normalize()` like you do in line 319, as that's a bad reference.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -256,75 +267,45 @@ public boolean isContainedIn(Resource r)
...
         {
-            try
-            {
-                // Use normalized path to get past navigational references like "/bar/../foo/test.txt"
-                Path ref = Paths.get(uri.normalize());
-                return ref.toRealPath();
-            }
-            catch (IOException ioe)
+            targetPath = resolveTargetPath();
+            if (Files.exists(targetPath))
```
joakime: If `targetPath` exists, how can it ever be non-existent?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -256,75 +267,45 @@ public boolean isContainedIn(Resource r)
...
-         * we will just use the original URI to construct the
-         * alias reference Path.
-         *
-         * We use the method `toUri(Path)` here, instead of
-         * Path.toUri() to ensure that the path contains
-         * a trailing slash if it's a directory, (something
-         * not all FileSystems seem to support)
-         */
-        if (!URIUtil.equalsIgnoreEncodings(uri, toUri(path)))
+        if (targetPath == null)
```
joakime: The targetPath can sometimes not be obtained.
FileSystem issues, IO issues, etc.
So this could be null.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -256,75 +267,45 @@ public boolean isContainedIn(Resource r)
...
+                     * the creation of a Path object reference.
+                     * If the URI is different from the Path.toUri() then
+                     * we will just use the original URI to construct the
+                     * alias reference Path.
+                     *
+                     * We use the method `toUri(Path)` here, instead of
+                     * Path.toUri() to ensure that the path contains
+                     * a trailing slash if it's a directory, (something
+                     * not all FileSystems seem to support)
+                     */
```
joakime: Perhaps add an example here.

``` 
// On windows
PathResource resource         = PathResource("C:/temp");
PathResource child            = resource.resolve("aa./foo.txt");
child.exists()                == true
child.isAlias()               == true
child.toUri()                 == "file:///C:/temp/aa./foo.txt"
child.getPath().toUri()       == "file:///C:/temp/aa/foo.txt"
child.getTargetPath().toUri() == "file:///C:/temp/aa/foo.txt"
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: So if the Resource does not exist, and the targetPath also does not exist (or can be determined).
Is it safe to have this `resource.isAlias()` return false?

What if the resource gets created later?

What if the resource is a symlink that doesn't point to anything? (this would be an exception in `.toRealPath()` as well)
And then it does point to something after the resource is created?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -170,7 +170,11 @@ public static boolean isSameName(Path pathA, Path pathB)
     @Override
     public boolean exists()
     {
-        return Files.exists(targetPath != null ? targetPath : path);
+        if (Files.exists(path))
+            return true;
+        if (isAlias())
+            return Files.exists(targetPath);
+        return false;
```
sbordet: `Files.exists()` implicitly take a `FOLLOW_SYMLINKS`, so the first at line 173 check should be enough.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -256,75 +267,45 @@ public boolean isContainedIn(Resource r)
...
-            targetResolved = true;
-        }
-        if (targetPath == null)
-            return null;
+        checkAlias();
         return targetPath.toUri();
     }
 
-    private Path resolveTargetPath()
+    private void checkAlias()
```
sbordet, line 268: Rename to `resolveAlias()`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
sbordet: And we don't want to resolve again, so we need a special object to an additional boolean.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -334,7 +315,8 @@ private Path resolveTargetPath()
         {
             LOG.warn("bad alias ({} {}) for {}", e.getClass().getName(), e.getMessage(), path);
         }
-        return null;
+
+        return path.normalize();
```
sbordet: If there was an exception, we want to return `null`, as we could not resolve the target path.

sbordet (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
lachlan-roberts: Not if the path has unresolved `..` segments, then `Files.exists()` returns false even if it does exist.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
lachlan-roberts: If you look at my implementation of `resolveTargetPath()` you'll see that it never returns null.
It only attempts to resolve aliases and if it can't be done it will return the same path it was given.

So the `targetPath` field acts as the flag to determine whether we have checked for aliases.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
lachlan-roberts: Not sure what you are asking here.

But I don't think we need this check, I will review why I have put it there.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
lachlan-roberts: > Got an example of where it doesn't?

If you give a path with un-normalized dot dot segments it will report file does not exist and `toRealPath()` will throw, but if you normalize first it resolves correctly to the existing file.
I encountered this during the testing of this change.

> This should not return the path.normalize() like you do in line 319, as that's a bad reference.

The reason I return `path.normalize()` on line 319 is so that a path like `/foo/bar/../something` would resolve as an alias to `/foo/something` without having to exist. I don't think paths with unresolved `.` or `..` segments should ever be classified as not an alias.

> What if the resource gets created later?

This is a good point, perhaps anything that does not exist should be reported as being an alias.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
lachlan-roberts: See I have changed the javadoc in `Resource.getTargetUri()` and changed the default implementation to `getUri()`.

So now the `getTargetUri()` is now only different to `getUri()` if there was an alias. Otherwise it will return the same value. And where `resolveTargetPath()` is used, we say it is an alias if the `targetPath` and `path` are different in any way.

So if path has unresolved dot segments we should resolve these here to show it is different to `path` (and should be an alias).

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: Lets say it this way ...

If `targetPath` is not null, because it got returned from `resolveTargetPath()` how can it not pass `Files.exists(targetPath)`?
It would seem this condition is always true.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: I think we (sbordet and I) both think `resolveTargetPath()` shouldn't return a Path on failure

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -50,7 +50,7 @@ public class PathResource extends Resource
     private final Path path;
     private final URI uri;
     private boolean alias = false;
-    private Path targetPath;
+    private Path canonicalPath;
```
joakime: Why "canonical"?
This is confusing to me.

In symlink terms, this PathResource is the requested resource.
The symlink target is the destination.

In windows extensionless terms, this PathResource is the requested resource.
The target is the actual file that would be opened in the FileSystem and FileStore objects.

In unicode normalization terms, this PathResource is the requested resource.
The target is the actual file that would be opened in the FileSystem and FileStore objects.

"canonical" is _not a term_ used by Path, FileSystem, or FileStore.
"Target" is, especially so with `toRealPath()` usage.

"canonical" is for `File` or `URI`, which `PathResource` and especially this field is not.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: I agree with sbordet

If `this.path` was created with one of the `URI` based methods for creating a `Path` (eg: `Paths.of(URI)` or `Paths.get(URI)`) then your statement is true.
As there is an expectation that the `URI` it is given is absolute, canonical, and normalized.
But if `this.path` was created from a `String` or other `Path` object, even the `/../` segments have no impact on the `Files.exists()` logic and will return true.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: I agree with @sbordet, this is a `resolveAlias()` or `resolveTarget()` name.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -329,12 +329,12 @@ public boolean isAlias()
     }
 
     /**
-     * The target URI of the resource. If this Resource is an alias pointing to a different location,
+     * The canonical URI of the resource. If this Resource is an alias pointing to a different location,
```
joakime: s/canonical/target/

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -329,12 +329,12 @@ public boolean isAlias()
...
      * this will resolve the alias to return the true target URI of the resource.
      *
-     * @return The target URI location of this resource, with any aliases resolved.
+     * @return The canonical URI location of this resource, with any aliases resolved.
```
joakime: s/canonical/target/

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -329,12 +329,12 @@ public boolean isAlias()
...
      */
-    public URI getTargetURI()
+    public URI getCanonicalURI()
```
joakime: s/canonical/target/

joakime (changes requested):
I disagree with the name change from `target` to `canonical` as that's not appropriate for `Path` (or `FileSystem` or `FileStore`).

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
gregw: Several reasons:
 + "Target" implies that it is a symlink that has a target.    Hopefully most PathResources will not be symlinks.
 + There are other types of alias other than symlink (case insensitive, 8.3 name, normalization). Target means nothing for these type of alias.
 + The changes here have this field being non null for all resources, i.e. even a non aliased resource will have this field set to the same value as path.  This is definitely not a target.
 + The very definition of an alias is when the passed path does not equal the canonical path (or passed URI does not equal the canonical URI), thus we get the good code `aliased = !this.path.equals(this.canonicalPath)`

I was happy with the previous name of `getAliased` which I don't think should have been changed to the symlink only `getTarget`.  But now that it is always non null, `getCanonical` is really what it is.

gregw: More over, "canonical" is definitely a term used in relation to the File system.  In Jetty 6,7,8,9,10,11 the definition of an alias has been if `file.getCanonicalFile() != file`.  Using the name canonical here is a good link back to previous versions so we can see that we have the same meaning for what an alias is.

lachlan-roberts: @gregw after recent changes it is not always true that `getCanonical` is always not null.
It is now null in the case where the file does not exist, or with a broken simlink, but if it returns null the resource is classified as an alias.

So `Resource.getCanonicalURI() == null` is now not equivalent to `isAlias()` like it was before this PR.
And if `Resource.getCanonicalURI() == null` then `Resource.exists()` is always false.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
lachlan-roberts: > But if this.path was created from a String or other Path object, even the /../ segments have no impact on the Files.exists() logic and will return true.

This is not correct.

The following test passes and shows that Files.exists does not work with `/../` segments. And since you can create `PathResource` with a `Path` directly we cannot assume that the path does not contain `/../` segments.
```java
Path resourcePath = testDir.resolve("resource.txt");
assertTrue(Files.exists(resourcePath));

Path aliasPath = testDir.resolve("foo/../resource.txt");
assertFalse(Files.exists(aliasPath));
assertTrue(Files.exists(aliasPath.normalize()));
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
gregw: @lachlan-roberts that doesn't make sense?  How can `resource.getCanonicalURI() == null` be equiv to isAlias? Do you mean  `resource.getCanonicalURI() != null` is equivalent to `isAlias==true`?
Does `getCanonicalURI()` ever return non-null for a non-aliased resource?   If not, then can we rename it back to `getAlias` or `getAliasedURI`?

lachlan-roberts: > How can resource.getCanonicalURI() == null be equiv to isAlias?

I meant `resource.getCanonicalURI() != null` is NOT equivalent to `isAlias==true`.
Before this PR it was used as an equivalent check.

> Does getCanonicalURI() ever return non-null for a non-aliased resource?

Yes as long as the resource exists then `getCanonicalURI()` will be non-null. If it doesn't exist we say the resource is an alias and `getCanonicalURI()` is null because we couldn't resolve it.

gregw: OK, so "Canonical" still works.  It is the correct representation of that resource... which is null for a non existent resource.

joakime: "Canonical" (when used in conjunction with File Systems) is an old term that no longer has relevance.

Note that "target" is the new term, as it refers to both "symlink" and "as storage preferred name" (eg: unicode normalization, case insensitive, escaped names, etc)

The techniques were to follow every target (symlink or as storage preferred name), in every component of the given name recursively, and ____

1. all but the last component must exist (and be accessible by user).
2. all components must exist (and be accessible by user).
3. without requirements of component existence or accessibility.

When java introduced `File.getCanonicalPath()` (in Java 1.1 back in 1997) and `File.getCanonicalFile()` (in Java 1.2 back in 1998) is actually a forced technique 3, which was fine at the time, but is now at odds with modern FileSystems that do not support that technique.
Tooling and other APIs was rampant for how to "canonicalize" a file system object too.
But that's not an API or tooling concern or technique, that's a FileSystem concern (how to get the real file system object name).

So things started to change around 2011.

I'll pick 2 notable events (out of many)

Back in 2012.
The gnu coreutils dropped "canonicalize" as a description of behaviors for filesystems when they moved from `readlink` (now deprecated) to `realpath` (the recommended approach).
You can still find "canonicalize" in the modern gnu coreutils documentation, but it's now historical and refers to in API techniques for resolution, not how the filesystem actually works. (if you look at an old version of coreutils documentation from before the introduction of `realpath` and you'll find "canonical" this and that everywhere in the documentation)

Back in late 2011.
The `File.getCanonicalFile()` / `File.getCanonicalPath()` APIs around Java 1.7 were updated, and is not a recommended API anymore, as it doesn't convey the truth of how FileSystems work anymore. (both of those APIs document and call out the existence of `Path.toRealPath()` now)

Java introduced the `Path` APIs in Java 1.7, and it was following the industry trend for trusting the filesystem, and not doing everything at the API layer.
The entire concept of canonicalization was updated to be "real path" now (as it referred to both symlink and "as storage preferred name" as well).
The default behavior of `UnixPath` (and `WindowsPath`) for `toRealPath()` is to defer to the FileSystem behavior (you can override this by using the option to not following links, and then it falls back to in-API behavior technique 2 in the above list).

In fact you cannot find "canonical" as a term in the javadoc on `java.nio.file.*` now.

```
[joakim@hyperion jdk-17-srcs]$ grep -RHin "canonical" java.base/java/io/* | wc -l
79
[joakim@hyperion jdk-17-srcs]$ grep -RHin "canonical" java.base/java/nio/file/* | wc -l
0
```

joakime: Don't introduce old, now-deprecated, terms.
Keep "target" as it applies to the modern era, not the past.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -241,90 +259,52 @@ public URI getURI()
...
         {
-            if (Files.isSymbolicLink(path))
-                return path.getParent().resolve(Files.readSymbolicLink(path));
-            if (Files.exists(path))
-            {
-                Path real = abs.toRealPath();
-                if (!isSameName(abs, real))
-                    return real;
-            }
+            return path.normalize().toRealPath();
```
joakime, line 307: This usage of `Path.toRealPath()` (with no arguments) is returning the FileSystem real name.

If you want the in-API version (not the true FileSystem name) then you have to pass in the NO_FOLLOW_LINKS argument.

Since you are dead set on using `canonicalPath` as the name of the field, at least use the API properly to actually get the old-school in-API canonicalization techniques out of this API.

I *really* don't like this.

joakime (changes requested):
Canonical Path is the wrong name for multiple reasons.

Either correct the name to be what it represents (a real name, or target), or correct the API usage to use the old school canonicalization techniques that you apparently want to stick with.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
lachlan-roberts: I don't care if it's called `canonicalPath` or `targetPath`, the only reason I changed it is because Greg asked me to. The naming of this method is not the purpose of the PR so I'd be happy to change the name back if agreed.

But we definitely do not want to use the `NO_FOLLOW_LINKS` argument here because we need to resolve symlinks to determine whether it is an alias.

lachlan-roberts: I have reverted the name back to `getTargetURI` for now.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
gregw: "Canonical" still is an english word with meaning.  But `getRealURI()` works for me.   "Target" is still just wrong in many circumstances.

gregw: Note also that we have recently added `getCanonicalPath` to `HttpURI`.  So we do have historical and current ties with the concept of canonical URIs and Paths. 

Moreover, one of the whole points of the `Resource` abstraction is that we simply cannot adopt an approach of "trusting the Filesystem".   It is precisely because of Filesystems forgiving nature that we have problems with aliasing that result in security issues.   So we explicitly do not defer resolution to the Filesystem.  We precisely work out a canonical URI, which we then wish to apply to a base Resource in a process that will reliably tell us if we have requested the origin/canonical/real resource or an alias of that.  

So we simply cannot "defer to the FileSystem" or JVM methods.   We have tried that several times before and every time that has resulted in security problems.  We have to stop repeating that mistake over and over.   

So we have three choices:

1. Go back to the `getAlias` or `getAliased` style names.   However, I think that @lachlan-roberts has expressed a preference for the ability to get a Path/URI that is non-null (if resource exists) regardless of if it is an alias or not.
2. Live with `getCanonicalURI` due to it's historic and current links with the process we have for detecting and handling aliases.
3. Find a new name ("Target" is not it) that can be applied to both `HttpURI` and `Resource`.   I could live with `getRealURI()`

@lachlan-roberts your call as to which if the above you go with.

joakime: Canonical makes sense for URI, not for anything related to FileSystems.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -371,6 +352,25 @@ private static URI toUri(Path path)
...
+    public boolean equals(Object obj)
+    {
+        if (this == obj)
+            return true;
+        if (obj == null)
+            return false;
+        if (getClass() != obj.getClass())
+            return false;
+        PathResource other = (PathResource)obj;
+        return Objects.equals(path, other.path) && Objects.equals(uri, other.uri);
```
joakime, line 365: Why do we need equals/hashCode at all?

The comparison on `uri` is odd, that's a user-space provided field.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/FileSystemResourceTest.java
```diff
@@ -622,13 +616,18 @@ public void testNonExistantSymlink() throws Exception
...
 
-        assertThat("resource.targetURI", resFoo, hasNoTargetURI());
-        assertThat("resource.uri.targetURI", ResourceFactory.root().newResource(resFoo.getURI()), hasNoTargetURI());
-        assertThat("resource.file.targetURI", ResourceFactory.root().newResource(resFoo.getPath()), hasNoTargetURI());
-
-        assertThat("targetURI", resBar, isTargetFor(resFoo));
-        assertThat("uri.targetURI", ResourceFactory.root().newResource(resBar.getURI()), isTargetFor(resFoo));
-        assertThat("file.targetURI", ResourceFactory.root().newResource(resBar.getPath()), isTargetFor(resFoo));
+        // This is an alias because the file does not exist.
+        assertFalse(resFoo.exists());
```
joakime, line 620: What!?
If the file doesn't exist, it can never be an alias.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/FileSystemResourceTest.java
```diff
@@ -622,13 +616,18 @@ public void testNonExistantSymlink() throws Exception
...
+        assertTrue(resFoo.isAlias());
```
joakime, line 621: This should be false.  Something is wrong here.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/FileSystemResourceTest.java
```diff
@@ -1067,10 +1066,14 @@ public void testSingleQuoteInFileName() throws Exception
             "URI[b] = " + refB;
         assertThat(msg, refA.equals(refB), is(false));
 
-        // now show that Resource.equals() does work
+        // These resources are not equal because they have different URIs.
```
joakime, line 1069: This feels dangerous for our mounting logic and also our caches.
The user-space provided URI should not be part of our equals/hashcode behaviors.
I don't see a use case in this PR that makes this change necessary.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/FileSystemResourceTest.java
joakime: This was fixed in PR #8706 btw (you should merge from `origin/jetty-12.0.x` again into this branch)

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/PathResourceTest.java
```diff
@@ -413,6 +416,44 @@ public void testSymlink(WorkDir workDir) throws Exception
         }
     }
 
+    @Test
+    public void testBrokenSymlink(WorkDir workDir) throws Exception
+    {
+        Path testDir = workDir.getEmptyPathDir();
+        Path resourcePath = testDir.resolve("resource.txt");
+        IO.copy(MavenTestingUtils.getTestResourcePathFile("resource.txt").toFile(), resourcePath.toFile());
+        Path symlinkPath = Files.createSymbolicLink(testDir.resolve("symlink.txt"), resourcePath);
```
joakime, line 425: Need protection for users that execute this test on filesystems that don't support symlink (don't check the OS here, check the success/failure of the createSymbolicLink method instead, like we do elsewhere)

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/PathResourceTest.java
```diff
@@ -413,6 +416,44 @@ public void testSymlink(WorkDir workDir) throws Exception
...
+
+        PathResource fileResource = new PathResource(resourcePath);
+        assertTrue(fileResource.exists());
+        PathResource symlinkResource = new PathResource(symlinkPath);
+        assertTrue(symlinkResource.exists());
+
+        // Their paths are not equal but not their canonical paths are.
```
joakime, line 432: This is not a canonical check.
This is provided path vs real path difference.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -170,36 +174,43 @@ public static boolean isSameName(Path pathA, Path pathB)
...
     public boolean exists()
     {
-        return Files.exists(targetPath != null ? targetPath : path);
-    }
-
-    @Override
-    public boolean equals(Object obj)
-    {
-        if (this == obj)
+        if (aliasResolved)
```
joakime, line 177: This feels super fishy.
Why are we checking the target/alias path first? or even at all?
What is a scenario where `this.path` does not exist, but `this.targetPath` does?

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/PathResourceTest.java
```diff
@@ -239,7 +242,7 @@ public void testJarFileIsAliasDirectory(WorkDir workDir) throws IOException
             // Resolve file using extension-less directory
             testText = archiveResource.resolve("/dir./test.txt");
             assertFalse(testText.exists());
-            assertFalse(testText.isAlias());
+            assertTrue(testText.isAlias());
```
joakime, line 245: Nope.  If it doesn't exist, it can never be an alias.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/PathResourceTest.java
```diff
@@ -413,6 +416,44 @@ public void testSymlink(WorkDir workDir) throws Exception
...
+        // Their paths are not equal but not their canonical paths are.
+        assertThat(fileResource.getPath(), not(equalTo(symlinkResource.getPath())));
+        assertThat(fileResource.getPath(), equalTo(symlinkResource.getTargetPath()));
+        assertFalse(fileResource.isAlias());
+        assertTrue(symlinkResource.isAlias());
+        assertTrue(fileResource.exists());
+        assertTrue(symlinkResource.exists());
+
+        // After deleting file the Resources do not exist even though symlink file exists.
+        Files.delete(resourcePath);
```
joakime, line 441: This delete needs protection for users on filesystems that don't immediately delete.
Test that the `resourcePath` was actually deleted in an assumption.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
lachlan-roberts: I'm not sure why we need to compare `PathResource`, I know there are tests for the `equals()` but not sure if we actually use it anywhere. But to compare these correctly you need to take the `uri` into account.

You can come up with examples where two different `PathResource`s have the same `path` field but one is classified as an alias and the other is not. So comparing only on `path` isn't quite correct.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/ResourceTest.java
```diff
@@ -398,7 +398,7 @@ public void testDotAliasDirDoesNotExist(WorkDir workDir)
         Resource dot = resource.resolve(".");
         assertNotNull(dot);
         assertFalse(dot.exists());
-        assertFalse(dot.isAlias(), "Reference to '.' is not an alias as directory doesn't exist");
+        assertTrue(dot.isAlias(), "Reference to '.' is an alias as directory doesn't exist");
```
joakime, line 401: Again, if the resource doesn't exist, it cannot have an alias.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/ResourceTest.java
```diff
@@ -428,7 +428,7 @@ public void testDotAliasFileDoesNotExists(WorkDir workDir) throws IOException
         Resource dot = resource.resolve(".");
         assertNotNull(dot);
         assertFalse(dot.exists());
-        assertFalse(dot.isAlias(), "Reference to '.' is not an alias as file doesn't exist");
+        assertTrue(dot.isAlias(), "Reference to '.' is an alias as file doesn't exist");
```
joakime, line 431: Again, if the resource doesn't exist, it cannot have an alias.

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/ContextHandler.java
```diff
@@ -1379,11 +1379,7 @@ public Resource getResource(String pathInContext) throws MalformedURLException
             // addPath with accept non-canonical paths that don't go above the root,
             // but will treat them as aliases. So unless allowed by an AliasChecker
             // they will be rejected below.
-            Resource resource = baseResource.resolve(pathInContext);
-
-            if (checkAlias(pathInContext, resource))
-                return resource;
-            return null;
+            return baseResource.resolve(pathInContext);
```
joakime, line 1382: Don't think you can remove the alias checking here.

jetty-ee9/jetty-ee9-servlet/src/main/java/org/eclipse/jetty/ee9/servlet/DefaultServlet.java
```diff
@@ -457,8 +457,6 @@ protected Resource resolve(String subUriPath)
             if (_baseResource != null)
             {
                 r = _baseResource.resolve(subUriPath);
```
joakime, line 459: Lets move these alias checking changes in ee9 out to a different PR.

joakime (changes requested):
The new equals/hashcode logic is not needed, nor is is critical for any step you have in this PR.
Also, if a Path doesn't exist, then it cannot have an alias.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -142,8 +143,8 @@ public static boolean isSameName(Path pathA, Path pathB)
 
     PathResource(URI uri, boolean bypassAllowedSchemeCheck)
     {
-        // normalize to referenced location, Paths.get() doesn't like "/bar/../foo/text.txt" style references
-        // and will return a Path that will not be found with `Files.exists()` or `Files.isDirectory()`
+        // Normalize to referenced location, Paths.get() doesn't like "/bar/../foo/text.txt" style references
+        // and will return a Path that will not be found with `Files.exists()` or `Files.isDirectory()`.
         this(Paths.get(uri.normalize()), uri, bypassAllowedSchemeCheck);
```
joakime, line 148: When using `Paths.get(URI)` the URI has to be sane, as path segments like `/../` are treated as valid segments to resolve on the filesystem of the path.
If we were using `Paths.get(String)` then those segments (`/../`) are treated as navigation segments and will work.

The behavior here is because we use URI vs String when creating the Path object.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: The `Files.exists()` does work with `/../` if each element of the path exists.
In your case `foo` does not exist on your filesystem.

This is part of the filesystem specific normalization techniques.
Some filesystems will find it (some will not).
Don't write your code assuming that all filesystems behave the same in this regard.

Example code (tested on Linux with ext4 filesystem)

``` java
package fs;

import java.io.IOException;
import java.net.URI;
import java.nio.file.Files;
import java.nio.file.Path;
import java.nio.file.Paths;

import org.eclipse.jetty.toolchain.test.FS;

public class PathsGetExample
{
    public static void main(String[] args) throws IOException
    {
        boolean withFoo = true;
        System.out.printf("withFoo = %b%n", withFoo);

        // Create some content
        Path testDir = Paths.get("/tmp/paths-get-examples");
        FS.ensureDirExists(testDir);
        Path fooDir = testDir.resolve("foo");
        if (withFoo)
            FS.ensureDirExists(testDir.resolve("foo"));
        else
            Files.deleteIfExists(fooDir);
        FS.touch(testDir.resolve("test.txt"));

        URI uri = URI.create(testDir.toUri().toASCIIString() + "foo/../test.txt");
        Path a = Paths.get(uri);
        System.out.printf("%s -> %s -> %b%n", uri, a, Files.exists(a));

        URI urin = uri.normalize();
        Path b = Paths.get(urin);
        System.out.printf("%s -> %s -> %b%n", urin, b, Files.exists(b));

        String str = "/tmp/paths-get-examples/foo/../test.txt";
        Path c = Paths.get(str);
        System.out.printf("%s -> %s -> %b%n", str, c, Files.exists(c));
    }
}
```

Executing with "foo"

```
withFoo = true
file:///tmp/paths-get-examples/foo/../test.txt -> /tmp/paths-get-examples/foo/../test.txt -> true
file:/tmp/paths-get-examples/test.txt -> /tmp/paths-get-examples/test.txt -> true
/tmp/paths-get-examples/foo/../test.txt -> /tmp/paths-get-examples/foo/../test.txt -> true
```

Executing without "foo"

```
withFoo = false
file:///tmp/paths-get-examples/foo/../test.txt -> /tmp/paths-get-examples/foo/../test.txt -> false
file:/tmp/paths-get-examples/test.txt -> /tmp/paths-get-examples/test.txt -> true
/tmp/paths-get-examples/foo/../test.txt -> /tmp/paths-get-examples/foo/../test.txt -> false
```

joakime: This is the same behavior you get on linux with `ls` ...

```
[joakim@hyperion ~]$ ls -la /tmp/paths-get-examples/
total 512
drwxrwxr-x   2 joakim joakim   4096 Oct 14 07:45 ./
drwxrwxrwt 221 root   root   516096 Oct 14 07:42 ../
-rw-rw-r--   1 joakim joakim      0 Oct 14 07:42 test.txt

[joakim@hyperion ~]$ ls -la /tmp/paths-get-examples/foo/../test.txt
ls: cannot access '/tmp/paths-get-examples/foo/../test.txt': No such file or directory

[joakim@hyperion ~]$ mkdir /tmp/paths-get-examples/foo

[joakim@hyperion ~]$ ls -la /tmp/paths-get-examples/
total 516
drwxrwxr-x   3 joakim joakim   4096 Oct 14 07:45 ./
drwxrwxrwt 221 root   root   516096 Oct 14 07:42 ../
drwxrwxr-x   2 joakim joakim   4096 Oct 14 07:45 foo/
-rw-rw-r--   1 joakim joakim      0 Oct 14 07:42 test.txt

[joakim@hyperion ~]$ ls -la /tmp/paths-get-examples/foo/../test.txt
-rw-rw-r-- 1 joakim joakim 0 Oct 14 07:42 /tmp/paths-get-examples/foo/../test.txt
```

joakime: The way I see it we have a few options here.

1. We have all `Resource` implementations do the URI normalization, removing segments like `/../` before creating the `Path` object.
2. We have all `Resource` implementations honor the as-submitted `URI`, with no `Resource` implementation normalization, and let the FileSystem suss out if the content can be referenced or not.

We are doing 1 currently, but should we?
Is it such a terrible idea to reject requests for content like `dir-does-not-exist/../test.dat` and allow `dir-that-exists/../test.dat`?
I could easily be convinced that the `Resource` implementations should not perform URI normalization.
It just means a few types of torturous requests we used to receive in the past we no longer support.
I wonder, is supporting `/dir-does-not-exist/` segments hiding or revealing extra information about the filesystem we shouldn't be doing? (need more time to ponder this)
What does supporting `/dir-does-not-exist/` segments mean for ResourceCollection too? (if anything)

Or, if we want to support these requests still, it means that the normalization occurs at the layer above `Resource.resolve(String)` (which kinda makes sense, and I suspect it would simplify the protected paths issues too)

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: This will impact the ResourceService cache though.

The same Resource will exist on multiple locations once this new equals/hashcode exists.
Is that an acceptable behavior?

joakime:
Closing in favor of #8734

PR #8675, opened 2022-10-01
Rollup of all of this month's `jetty-10.0.x` dependabots

joakime:
A single build/PR for all of the 10.0.x dependabots.


joakime:
Turns out, this is not a good idea.

PR #8676, opened 2022-10-01
Jetty 12 conditional handler

gregw:
During the hackathon, there was some discussion of factoring out the include/exclude conditions that exist in multiple handlers.
This is one way this could be done, with a wrapper that applies conditions (list of predicates) to the following handler.   If the predicates pass, then the next handler is called normally. If a predicate fails then either null is returned; or if skipNext is true, just the next handler is skipped and the subsequent wrapped handler called (this is like a GzipHandler not being applied, but it's wrapped handlers are).

This would allow most, but not all include/excludes to be removed from GzipHandler.  The mime type ones cannot be handled this way, as they are conditions on the response.

It is also a bit strange to be setting predicates on the previous handler.   An alternate approach is to have a `setConditions(List<Predicate<Request>>)` on all abstract handlers... but that is a little bit of taxation without representation for the many handlers that will not use it.

So currently, I'm not entirely sure that this is worth doing.   But I'll get some feedback on this draft and if we find a way we like, I will then try removing the conditional handling from various other handlers and seeing how it looks in code and XML config.


jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -600,6 +603,36 @@ public Request getWrapped()
...
+    static Predicate<Request> forPath(Collection<String> included, Collection<String> excluded)
+    {
+        IncludeExclude<String> paths = new IncludeExclude<>(PathSpecSet.class);
+        if (included != null)
+            paths.include(included.toArray(new String[0]));
+        if (excluded != null)
+            paths.exclude(excluded.toArray(new String[0]));
+        return r -> paths.test(r.getPathInContext());
+    }
+
```
sbordet, line 653: @gregw I think `Handler.Conditional` should be created like a builder, something like:

```java
Handler handler = Handler.Conditional
    .forMethod("GET")
    .forPaths("/include/*")
    .forHeader("X-Custom")
    .forHeaderValue("Expect", "100-Continue")
    .build();
```

In this way, the predicates will go out of `Request`.

Keep a generic constructor `Conditional(List<Predicate>)`, but prefer using the builder pattern.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java
```diff
@@ -690,4 +692,83 @@ public NonBlocking()
...
+                throw new IllegalStateException(getState());
+            _conditions.add(condition);
+        }
+
+        public boolean remove(Predicate<Request> condition)
+        {
+            if (isStarted())
+                throw new IllegalStateException(getState());
+            return _conditions.remove(condition);
+        }
```
sbordet, line 740: I don't think we want to enable `add()/remove()` predicates... I would just pass them to constructor and be a `final List`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
gregw: So in XML this could look like:
```XML
  <New class="org.eclipse.jetty.server.handler.Handler$ConditionalBuilder">
    <Call name="forMethod">
      <Arg>GET</Arg>
    </Call>
    <Call name="forPaths">
      <Arg>
        <Array type="String">
          <Item>"/include/*"</Item>
        </Array>
      </Arg>
    </Call>
    <Call id="ConditionalHandler" name="build">
      <Call name="setHandler">
        ...
      </Call>
    </Call>
```
Which reads OK, but cannot itself be used as an Arg. The resulting handler needs to be saved as an ID and then the reference used to set the handler.   So I'm not so convinced that the builder style works in XML.... perhaps we can add a `<Build class="org.eclipse.jetty.server.handler.Handler$Conditional">...</Build>` element to better support it?   Do we have other examples of builders used in XML?

But perhaps just doing constructor injection and getting rid of the setters would be simplest?

sbordet:
```xml
<Call id="conditionalHandler" class="Handler.Conditional" name="forMethod">
  <Arg>GET</Arg>
  <Call name="forPaths">
    <Arg>/include/*</Arg>
    <Call name="forHeader">
      <Arg>X-Custom</Arg>
      ...
    </Call>
  </Call>
</Call>
```

The `<Call>` elements are nested rather than sequential.
I don't see any problem to assign an ID and it can be used as an `Arg`, so I think we're already ok.

gregw: The calls can either be nested or sequential as builder methods just return this.  It should always be the same to do `builder.foo().bar();` or `builder.foo(); builder.bar()`

I'll make the changes to GzipHandler to use a Conditional Handler and see how it looks..... stand by ....

gregw: I've pondered the builder style for this, and decided against it.    All our other handlers are built by normal dependency injection, not constructor injection, so I don't think we should change style for just this one handler.

More over, our module configuration mechanism is additive, so we frequently will look for an entity as a XmlConfiguration ID and then add configuration to it.  In this case, I can imagine modules adding predicates to an already installed ConditionalHandler.

gregw:
@joakime @sbordet @lorban I think we should keep thinking about something like this, so we can move the repeated complexity out of things like `GzipHandler`.
This is still draft, but I'd like feedback on the approach and/or alternatives.

joakime:
@gregw is this still relevant?  can we close it?

gregw:
There is still some value here... although I think it is probably too late for 12.0.0.

gregw:
Closing for now.  Still something in the idea, but needs to be thought about and restarted off latest HEAD.

PR #8679, opened 2022-10-03
Fixes a test tagged as "external" that

sbordet:
fails during local builds with Java 17+.


joakime (approved):
Looks good to me.

You fixed this in Jetty 12 btw.

https://github.com/eclipse/jetty.project/blame/jetty-12.0.x/jetty-ee10/jetty-ee10-demos/jetty-ee10-demo-proxy-webapp/src/test/java/org/eclipse/jetty/ee10/demos/ProxyWebAppTest.java#L89

PR #8682, opened 2022-10-04
expand jetty properties when generating dry-run command line

lachlan-roberts:
When using `--dry-run`, the command line printed was not resolving jetty properties.

For example using `logging-jetty` adds `${jetty.home.uri}/lib/logging/` to server classes. 

However this is never resolved with the value of the `jetty.home.uri` property, and this is what is what ends up in server classes of the webapp when Jetty is started:
```
|  |     +> Serverclasses XWiki@c05fddc size=22
|  |     |  +> ${jetty.home.uri}/lib/logging/
```

this results in the `ServiceConfigurationError` seen in https://github.com/eclipse/jetty.docker/issues/118.

### DIFF
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
@@ -817,7 +817,7 @@ public class StartArgs
             {
                 for (Prop p : properties)
                 {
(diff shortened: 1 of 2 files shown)


joakime (approved):
Simple enough.

sbordet (approved):

PR #8685, opened 2022-10-04
[Jetty 12] Made WebSocket over HTTP/2 work.

sbordet:
Re-enabled tests, and restored HTTP2StreamEndPoint, as well as implemented getTunnelSupport() for HTTP/2.

Removed from HttpStream methods for upgrade that are not necessary anymore.


jetty-core/jetty-http2/jetty-http2-common/src/main/java/org/eclipse/jetty/http2/internal/HTTP2StreamEndPoint.java
```diff
@@ -354,9 +363,14 @@ public void fillInterested(Callback callback) throws ReadPendingException
     @Override
     public boolean tryFillInterested(Callback callback)
     {
-        boolean result = readCallback.compareAndSet(null, callback);
+       boolean result = readCallback.compareAndSet(null, callback);
```
lorban: This line looks misaligned.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingContentFactory.java
```diff
@@ -299,6 +298,18 @@ private CachingHttpContent(String key, HttpContent httpContent, String etagValue
             _lastAccessed = NanoTime.now();
         }
 
+        private ByteBuffer toMappedBuffer(Resource resource, long length)
```
lorban, line 301: `BufferUtil.toMappedBuffer` shouldn't throw, this shouldn't be necessary.

lorban (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingContentFactory.java
sbordet: Maybe, but changing `BufferUtil.toMappedBuffer()` to return `null` has a non-trivial scope better addressed in another PR.

lorban (approved):

PR #8688, opened 2022-10-05
Fix traversal bug on extraction of bad jar/zip files during jetty-start

joakime:
The jetty-start extraction of zip/jar files doesn't take into account traversal bugs in entries found in zip/jar files.

This PR fixes this for the jetty-start use case.

We already have similar protections in place for `JarResource` and `SelectiveJarResource` in jetty proper.


olamy (approved):

sbordet (approved):

MarcelRaschke (commented):
-s

Issue #8690, opened 2022-10-06
Confirm `merge-12-needed` labeled PRs have been merged


joakime:
**Jetty version(s)**
12.0.x

**Description**
Verify all remaining [`merge-12-needed`](https://github.com/eclipse/jetty.project/pulls?q=is%3Apr+label%3Amerge-12-needed+is%3Aclosed) labeled PRs have actually been merged to `jetty-12.0.x` branch.

Please verify each one, and check it off (also remove the `merge-12-needed` label)

## For @sbordet

 - [x] #7522
 - [x] #7563
 - [x] #7574
 - [x] #7637
 - [x] #7778
 - [x] #7800
 - [x] #7938
 
## For @janbartel
 
 - [x] #7532
 - [x] #7631
 - [x] #7639
 
## For @lachlan-roberts
 
 - [x] #7546
 - [x] #7579
 
## For @olamy 
 
 - [x] #7580
 - [x] #7687
 - [x] #7803
 
## For @lorban
 
 - [x] #7674
 

## Completed

 - [x] #7180
 - [x] #7184
 - [x] #7374
 - [x] #7417
 - [x] #7421
 - [x] #7441
 - [x] #7467
 - [x] #7506
 - [x] #7510
 - [x] #7516
 - [x] #7521
 - [x] #7525
 - [x] #7530
 - [x] #7547
 - [x] #7576
 - [x] #7596
 - [x] #7614
 - [x] #7618
 - [x] #7626
 - [x] #7633
 - [x] #7640
 - [x] #7656
 - [x] #7658
 - [x] #7669
 - [x] #7670
 - [x] #7678
 - [x] #7679
 - [x] #7682
 - [x] #7696
 - [x] #7764
 - [x] #7769
 - [x] #7776
 - [x] #7864
 - [x] #7867
 - [x] #8035

joakime:
I'm taking #7547 as that spelling mistake is in more places than the original PR

joakime:
Opened PR #8691 (inspired by for #7547)

joakime:
Yay. all done. closing this task.

PR #8691, opened 2022-10-06
Jetty 12 - Fix typo "octect" -> "octet"

joakime:
Inspired from PR #7547


joakime:
This is a task for #8690

sbordet (approved):

Issue #8695, opened 2022-10-07
Update quiche to 0.16.0


joakime:
**Jetty version(s)**
Jetty 10, Jetty 11, and Jetty 12

**Description**
Time to update quiche again.

Version 0.15.0 and 0.16.0 were just released.

https://github.com/cloudflare/quiche/releases/tag/0.15.0
https://github.com/cloudflare/quiche/releases/tag/0.16.0

joakime:
@lorban once this is updated for `jetty-12.0.x` branch, your task on Issue #8690 can be checked off.

lorban:
Quiche 0.16.0 has been released, updating straight to that version.

PR #8699, opened 2022-10-10
Jetty 10.0.x: update quiche to version 0.16.0

lorban:
The jetty-quiche-native artifact version 0.16.0 has now been deployed to central.

See #8695 (solved when this PR gets merged into 10, 11 and 12).

### DIFF
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
@@ -21,6 +21,7 @@ import java.net.SocketException;
 import java.net.SocketOption;
 import java.net.StandardProtocolFamily;
 import java.net.StandardSocketOptions;
+import java.nio.channels.DatagramChannel;
 import java.nio.channels.NetworkChannel;
 import java.nio.channels.SelectableChannel;
 import java.nio.channels.SelectionKey;
@@ -500,6 +501,13 @@ public class ClientConnector extends ContainerLifeCycle
             int sendBufferSize = getSendBufferSize();
             if (sendBufferSize >= 0)
                 setSocketOption(channel, StandardSocketOptions.SO_SNDBUF, sendBufferSize);
+        }
+        if (selectable instanceof DatagramChannel)
+        {
+            // QUIC must know the local address, but it is non-null on datagram sockets only if it has been bound,
+            // so implicitly bind to 0.0.0.0:0 when no bind address has been set.
+            if (getBindAddress() == null)
+                setBindAddress(new InetSocketAddress("0.0.0.0", 0));
         }
     }
 
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
@@ -21,7 +21,6 @@ import java.net.SocketException;
 import java.net.SocketOption;
 import java.net.StandardProtocolFamily;
 import java.net.StandardSocketOptions;
-import java.nio.channels.DatagramChannel;
 import java.nio.channels.NetworkChannel;
 import java.nio.channels.SelectableChannel;
 import java.nio.channels.SelectionKey;
@@ -501,13 +500,6 @@ public class ClientConnector extends ContainerLifeCycle
             int sendBufferSize = getSendBufferSize();
             if (sendBufferSize >= 0)
                 setSocketOption(channel, StandardSocketOptions.SO_SNDBUF, sendBufferSize);
-        }
-        if (selectable instanceof DatagramChannel)
-        {
-            // QUIC must know the local address, but it is non-null on datagram sockets only if it has been bound,
-            // so implicitly bind to 0.0.0.0:0 when no bind address has been set.
-            if (getBindAddress() == null)
-                setBindAddress(new InetSocketAddress("0.0.0.0", 0));
         }
     }
 
--- a/jetty-quic/quic-client/pom.xml
+++ b/jetty-quic/quic-client/pom.xml
@@ -53,8 +53,12 @@
     <profile>
       <id>jdk17</id>
       <activation>
-        <jdk>[17,)</jdk>
+        <jdk>17</jdk>
       </activation>
+      <!--
+      This profile makes sure the Foreign binding is used for tests when running exactly on JDK 17.
+      Older and newer JDKs will revert to the JNA binding.
+      -->
       <dependencies>
         <dependency>
           <groupId>org.eclipse.jetty.quic</groupId>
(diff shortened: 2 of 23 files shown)


joakime:
This branch is having issues.

```
java.lang.StackOverflowError
	at org.eclipse.jetty.quic.quiche@10.0.13-SNAPSHOT/org.eclipse.jetty.quic.quiche.QuicheConnection.connect(QuicheConnection.java:59)
	at org.eclipse.jetty.quic.quiche.jna.JnaQuicheBinding.connect(JnaQuicheBinding.java:62)
	at org.eclipse.jetty.quic.quiche@10.0.13-SNAPSHOT/org.eclipse.jetty.quic.quiche.QuicheConnection.connect(QuicheConnection.java:59)
	at org.eclipse.jetty.quic.quiche.jna.JnaQuicheBinding.connect(JnaQuicheBinding.java:62)
	at org.eclipse.jetty.quic.quiche@10.0.13-SNAPSHOT/org.eclipse.jetty.quic.quiche.QuicheConnection.connect(QuicheConnection.java:59)
	at org.eclipse.jetty.quic.quiche.jna.JnaQuicheBinding.connect(JnaQuicheBinding.java:62)
	at org.eclipse.jetty.quic.quiche@10.0.13-SNAPSHOT/org.eclipse.jetty.quic.quiche.QuicheConnection.connect(QuicheConnection.java:59)
	at org.eclipse.jetty.quic.quiche.jna.JnaQuicheBinding.connect(JnaQuicheBinding.java:62)
	at org.eclipse.jetty.quic.quiche@10.0.13-SNAPSHOT/org.eclipse.jetty.quic.quiche.QuicheConnection.connect(QuicheConnection.java:59)
	at org.eclipse.jetty.quic.quiche.jna.JnaQuicheBinding.connect(JnaQuicheBinding.java:62)
... etc ...
```

joakime (dismissed):

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
```diff
@@ -501,6 +502,13 @@ protected void configure(SelectableChannel selectable) throws IOException
             if (sendBufferSize >= 0)
                 setSocketOption(channel, StandardSocketOptions.SO_SNDBUF, sendBufferSize);
         }
+        if (selectable instanceof DatagramChannel)
+        {
+            // QUIC must know the local address, but it is non-null on datagram sockets only if it has been bound,
+            // so implicitly bind to 0.0.0.0:0 when no bind address has been set.
+            if (getBindAddress() == null)
+                setBindAddress(new InetSocketAddress("0.0.0.0", 0));
+        }
```
sbordet: Uh, no. Why is this needed? Certainly not "QUIC" the protocol.

jetty-quic/quic-client/pom.xml
```diff
@@ -61,6 +61,43 @@
...
+                @{argLine}
+                ${jetty.surefire.argLine}
+                --add-modules=jdk.incubator.foreign
+                --add-opens jdk.incubator.foreign/jdk.incubator.foreign=ALL-UNNAMED
+                --enable-native-access ALL-UNNAMED
+              </argLine>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
```
sbordet, line 84: If you add this, then you must restrict the profile to just `17`, not `[17,)`.

sbordet (changes requested):

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
joakime: Also, for IPv6 specific runtimes this would be a different address.

The quiche examples bind to these addresses always, just to satisfy things on OSX and BSD.

https://github.com/cloudflare/quiche/blob/21af572b4ae1ce906fbac7b3048409c935e59662/apps/src/client.rs#L73-L79

``` rs
    // Bind to INADDR_ANY or IN6ADDR_ANY depending on the IP family of the
    // server address. This is needed on macOS and BSD variants that don't
    // support binding to IN6ADDR_ANY for both v4 and v6.
    let bind_addr = match peer_addr {
        std::net::SocketAddr::V4(_) => format!("0.0.0.0:{}", args.source_port),
        std::net::SocketAddr::V6(_) => format!("[::]:{}", args.source_port),
    };
```

jetty-quic/quic-client/pom.xml
joakime: This needs to happen.

The new JDK 19 builds are failing without this.

```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[2022-10-12T22:11:58.242+0000][info][gc] Using G1
[ERROR] Error occurred during initialization of boot layer
[ERROR] java.lang.module.FindException: Module jdk.incubator.foreign not found, required by org.eclipse.jetty.quic.quiche.foreign.incubator
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[WARNING] Corrupted channel by directly writing to native stream in forked JVM 1. See FAQ web page and the dump file /home/jenkins/agent/workspace/jetty.project_PR-8707/jetty-core/jetty-quic/jetty-quic-server/target/surefire-reports/2022-10-12T21-58-24_538-jvmRun1.dumpstream
```

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
lorban: My understanding is that QUIC allows `Connection Migration`. This was only implemented in Quiche 0.15.0 and required a quiche API change ([see here](https://github.com/cloudflare/quiche/releases/tag/0.15.0)).

This means migrating a connection from one path (e.g.:IP1 <=> IP2) to another path (e.g.: IP3 <=> IP2) now requires quiche to know about the source and destination IP addresses. AFAICT the only way to get the right IP and port is to bind the local address otherwise querying the socket always returns null.

Is there another way, or did I miss something, or could we move that code somewhere else? Maybe, I'm open to discussing this.

lorban:
FYI, quiche 0.16.0 is now out, has been built with the native binaries published to central, and has no breaking API change so we could upgrade immediately to 0.16.0 by modifying this PR.

jetty-quic/quic-client/pom.xml
lorban: Done.

jetty-io/src/main/java/org/eclipse/jetty/io/ClientConnector.java
lorban: After a discussion with @sbordet we figured this binding is needed only the client needs this extra one, and it can figure the IP address if `null` is passed to `bind()` so this was moved to the quic client connector configurator.

joakime:
> FYI, quiche 0.16.0 is now out, has been built with the native binaries published to central, and has no breaking API change so we could upgrade immediately to 0.16.0 by modifying this PR.

Yeah, lets skip 0.15.0 and go straight to 0.16.0.
No need to make a new PR (if you don't want to).
Just update this one (and change the title)

sbordet (approved):

PR #8702, opened 2022-10-10
Jetty 12 - Resource `resolve()` and `newResource()` return null on resources that do not exist

joakime:
Cleaned up version of PR #8597


jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ScanningAppProvider.java
```diff
@@ -226,7 +226,7 @@ protected void doStart() throws Exception
         List<Path> files = new ArrayList<>();
         for (Resource resource : _monitored)
         {
-            if (resource.exists() && Files.isReadable(resource.getPath()))
+            if (resource != null && Files.isReadable(resource.getPath()))
```
gregw: So `resource != null` doesn't ensure the file exists, but then Files.isReadable() does an existence check, so that is OK.  But `getPath` can return null, so really this should be:

```suggestion
            if (resource != null && resource.getPath() != null && Files.isReadable(resource.getPath()))
```

Moreover, there is nothing stopping the monitored resource being a composite, so it may definitely have a null Path, but we probably do need to check directories for readability.  Thus it still would be better to just be:

```suggestion
            if (resource != null && resource.isReadable())
```

jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ScanningAppProvider.java
```diff
@@ -339,7 +339,11 @@ public void setDeploymentManager(DeploymentManager deploymentManager)
     public void setMonitoredResources(List<Resource> resources)
     {
         _monitored.clear();
-        _monitored.addAll(resources);
+        for (Resource resource: resources)
+        {
+            if (resource != null)
+                _monitored.addAll(resources);
+        }
```
gregw: You need to protect again null resources:
```suggestion
        if (resources != null)
            resources.stream().filter(Objects::nonNull).forEach(_monitored::add);
```
or maybe:

```suggestion
        Optional.ofNullable(resources).stream().flatMap(List::stream)
            .filter(Objects::nonNull)
            .forEach(_monitored::add);
```

jetty-core/jetty-keystore/src/main/config/modules/test-keystore.mod
```diff
@@ -6,7 +6,7 @@ DO NOT USE IN PRODUCTION!!!
 demo
 ssl
 
-[depend]
+[before]
```
gregw: why?
why in this PR?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -720,7 +720,18 @@ public void setBaseResource(Resource resourceBase)
...
+        if (path == null)
+        {
+            // allow user to unset variable
+            setBaseResource((Resource)null);
+            return;
+        }
+
+        Resource resource = ResourceFactory.root().newResource(path);
+        if (resource == null)
+            throw new IllegalArgumentException("Base Resource does not exist: " + path);
```
gregw: maybe:
```suggestion
        if (resource == null && resource.isReadable())
            throw new IllegalArgumentException("Base Resource does not exist or is not readable: " + path);
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/TypeUtil.java
```diff
@@ -695,17 +695,7 @@ public static URI getClassLoaderLocation(Class<?> clazz, ClassLoader loader)
...
-                    {
-                        uriStr = uriStr.substring(4);
-                        int idx = uriStr.indexOf("!/");
-                        if (idx > 0)
-                        {
-                            return URI.create(uriStr.substring(0, idx));
-                        }
-                    }
-                    return uri;
+                    return URIUtil.unwrapContainer(uri);
```
gregw: Just commit cleanups like this to the branch.... best to keep them out of PRs that need to be slowly considered

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -247,9 +255,53 @@ public int hashCode()
...
+        // Sub-paths are always resolved under the given URI,
+        // we compensate for input sub-paths like "/subdir"
+        // where default resolve behavior would be to treat
+        // that like an absolute path.
+        while (subUriPath.startsWith(URIUtil.SLASH))
+        {
+            // TODO XXX this appears entirely unnecessary and inefficient.  We already have utilities
+            //      to handle appending path strings with/without slashes.
+            subUriPath = subUriPath.substring(1);
+        }
```
gregw, line 297: Can we try removing this and see what breaks?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -131,8 +131,8 @@ public Resource resolve(String subUriPath)
         for (Resource res : _resources)
         {
             addedResource = res.resolve(subUriPath);
-            if (!addedResource.exists())
-                continue;
+            if (addedResource == null)
+                continue; // skip, doesn't exist
```
gregw: The javadoc for this method needs to be updated for the null return.

for me, this is the primary reason for returning null, as a non null return from a composite in this circumstance makes no sense... although I guess it could return a `BadResource` or some other non-null resource that represented the fact that the resource doesn't exist and can't easily be created.

jetty-ee10/jetty-ee10-maven-plugin/src/main/java/org/eclipse/jetty/ee10/maven/plugin/AbstractUnassembledWebAppMojo.java
```diff
@@ -176,7 +176,7 @@ protected void configureUnassembledWebApp() throws IOException
             if (webXml != null)
             {
                 Resource r = webApp.getResourceFactory().newResource(webXml.toPath());
-                if (r.exists() && !r.isDirectory())
+                if (r != null && !r.isDirectory())
```
gregw: Maybe readability would be better if we had some static methods that protected for null values:

```suggestion
                if (Resource.isDirectory(r))
```

We could avoid scattering lots of null checks all over the code then.

jetty-ee10/jetty-ee10-webapp/pom.xml
```diff
@@ -40,7 +40,7 @@
           </argLine>
           <useManifestOnlyJar>false</useManifestOnlyJar>
           <additionalClasspathElements>
-            <additionalClasspathElement>${basedir}/src/test/resources/module/foo-bar-janb.jar</additionalClasspathElement>
+            <additionalClasspathElement>${basedir}/src/test/resources/mods/foo-bar-janb.jar</additionalClasspathElement>
```
gregw: why?
why this PR?

gregw (changes requested):
Generally I'm still not convinced that we really want to scatter null checks all over the code base. I also really don't like the implied suggestion that a non null resource does exist.  That might be true soon after resolved, but it might not be true for a resource taken out of a collection some time after the resolution.

So I think regardless of returning null or not, I think we need some static methods that will do the null checking:

```java
    interface Resources
    {
        public static boolean exists(Resource r)
        {
            return r != null && r.exists();
        }
        public static boolean isDirectory(Resource r)
        {
            return r != null && r.isDirectory();
        }
    }
```

This style mimics the statics in `Paths` and `Files`   (although I don't mind if they are on `Resource` rather than a `Resources` interface).


I'm also wondering if perhaps an Optional would be better than a null return.  So resolve would always return a non null resource (even if it is a BadResource), but then we'd have a new method:

```java
    public Optional<Resource> resolveExisting(String subPath)
    {
        Resource resource = resolve(subPath);
        if (resource.exists())
            return Optional.of(resource);
        return Optional.empty();
    }
```

jetty-core/jetty-keystore/src/main/config/modules/test-keystore.mod
joakime: The test-keystore was out of order.
returning null from resolve() showed the bug.
fixing the order was necessary.
I'll move this change out.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/TypeUtil.java
joakime: Will move out

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: This is being done in PR #8631

jetty-ee10/jetty-ee10-webapp/pom.xml
joakime: This was another bug that manifested once we returned null.
That file doesn't exist in that location, at all, but we relied on it.
The directory name was changed a long time ago.
But the testcase was not testing the correct thing.
I'll move this out too.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
joakime: Done

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -247,6 +258,96 @@ public int hashCode()
...
+
+        URI uri = getURI();
+        URI resolvedUri = URIUtil.addPath(uri, subUriPath);
+        Path path = Paths.get(resolvedUri.normalize());
```
gregw: why is the normalize here?   won't that hide some aliases?   @lachlan-roberts will it be sufficient to detect aliases that the URI from the resulting path will differ from the `resolvedUri`?

joakime: This is because of differences when using `Paths.get(URI)` (super strict) vs `Paths.get(String)` (very flexible).
I'll be happy to show you.

joakime: This won't hide aliases, as the user provided URI is still tracked separately from the Path in the PathResource.
The URI comparison later, during the alias check will still show the difference we are looking for.

joakime: We should review all normalize within the Resource layer (inputs, outputs, comparison, etc)

See the other PR about this -> https://github.com/eclipse/jetty.project/pull/8631#discussion_r994087669

jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ScanningAppProvider.java
joakime: Using new `Resources` utility class instead

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
```diff
@@ -257,6 +257,13 @@ public boolean isDirectory()
         return true;
     }
 
+    @Override
+    public boolean isReadable()
+    {
+        // always a directory, never readable
+        return false;
+    }
```
gregw: A directory can be readable, in that it's contents can be listed.

I can see an interpretation of `isReadable` that means there is content available, but this is a change in semantics that I don't think we need.    It is a valid check to ask a directory if it is readable or not.   Perhaps this check could be made in the constructor?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resources.java
```diff
@@ -0,0 +1,82 @@
...
+     * @return true if resource is non-null, exists, and is a directory
+     * @see Resource#exists()
+     * @see Resource#isDirectory()
+     */
+    public static boolean isDirectory(Resource resource)
+    {
+        if (resource == null)
+            return false;
+
+        return resource.exists() && resource.isDirectory();
```
gregw: You don't need the exists check as `isDirectory` has an implicit existence check.
```suggestion
        return resource != null && resource.isDirectory();
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resources.java
```diff
@@ -0,0 +1,82 @@
...
+     * @param resource the resource to test
+     * @return true if resource is non-null and exists
+     * @see Resource#exists()
+     */
+    public static boolean exists(Resource resource)
+    {
+        if (resource == null)
+            return false;
+
+        return resource.exists();
```
gregw: I kind of prefer:
```suggestion
        return resource != null && resource.exists();
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resources.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * True if the resource is missing.
+     *
+     * @param resource the resource to test
+     * @return true if resource is null or doesn't exist
+     * @see Resource#exists()
+     */
+    public static boolean missing(Resource resource)
```
gregw, line 40: needed? whats wrong with !exists?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resources.java
```diff
@@ -0,0 +1,82 @@
...
+     * @return true if resource is non-null, exists, and is readable
+     * @see Resource#exists()
+     * @see Resource#isReadable()
+     */
+    public static boolean isReadable(Resource resource)
+    {
+        if (resource == null)
+            return false;
+
+        return resource.exists() && resource.isReadable();
```
gregw:
```suggestion
        return resource != null && resource.isReadable();
```

gregw (changes requested):
I like the statics... but a few niggles

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceCollection.java
joakime: Asking "isReadable()" should be different than "isDirectory()".
Meaning if you are `isDirectory() == true` then you have no need to call `isReadable()` anyway.

If you look at usage, the `isReadable()` checks are typically for places where we expect a non-directory that can be accessed and be read (eg: a file)

gregw: I know that we used to check some directories for readability.   I don't really see the need for changing this semantic.   What does a directory that exists but is not actually readable on the file system look like?  If all our directory handling code expects directories to not be readable, then we won't be able to distinguish this case?

gregw: perhaps rename to `hasContent` or `hasReadableContent` if you want your semantic, to distinguish from the `Files` method that can apply to a directory.

joakime: Perhaps it should be `canBeOpenForRead()` instead.
As it's not meant for "can we access" or "do we have read permissions".

gregw (approved):

PR #8705, opened 2022-10-12
Jetty 12 - Re-enabled RequestLog tests.

sbordet:
Re-implemented features that were commented out.


jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/CustomRequestLogTest.java
```diff
@@ -220,209 +246,301 @@ public void testLogAddress() throws Exception
...
-        assertThat(connect.getResponse(), containsString("200 OK"));
-
-        assertThat(_entries.poll(5, TimeUnit.SECONDS), is("KeepAliveRequests: 1"));
-        assertThat(_entries.poll(5, TimeUnit.SECONDS), is("KeepAliveRequests: 2"));
-        assertThat(_entries.poll(5, TimeUnit.SECONDS), is("KeepAliveRequests: 3"));
-    }
-
-    @Disabled
-    @Test
-    public void testLogKeepAliveRequestsHttp2() throws Exception
```
joakime, line 339: I don't see this test anymore.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/CustomRequestLogTest.java
```diff
@@ -494,172 +595,145 @@ public void testLogLatencySeconds() throws Exception
...
+                trailers.put("trailerName", "42");
+                callback.succeeded();
+            }
+        });
 
-    protected Socket newSocket(String host, int port) throws Exception
-    {
-        Socket socket = new Socket(host, port);
-        socket.setSoTimeout(10000);
-        socket.setTcpNoDelay(true);
```
joakime, line 588: I think this existed for Windows, i'll have to double-check on my laptop.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -781,14 +782,27 @@ void onTrailers(HttpFields trailers)
...
+        {
+            Authentication authentication = apiRequest.getAuthentication();
+            if (authentication instanceof Authentication.User userAuthentication)
+                _request.setAttribute(CustomRequestLog.USER_NAME, userAuthentication.getUserIdentity().getUserPrincipal().getName());
+
+            String realPath = apiRequest.getServletContext().getRealPath(_request.getPathInContext());
+            _request.setAttribute(CustomRequestLog.REAL_PATH, realPath);
+
+            String servletName = _request.getServletName();
+            _request.setAttribute(CustomRequestLog.HANDLER_NAME, servletName);
```
joakime, line 803: This all seems like something we used to do in a Customizer.

joakime (commented):
minor niggles.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/CustomRequestLogTest.java
sbordet: Correct, because we are in `jetty-server` we cannot have the http2 dependencies.
I have no idea why this test was added in the first place, but it's disabled in every Jetty version for the reasons above.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/CustomRequestLogTest.java
sbordet: I don't think it's necessary.
Worst case the server will close the connection after 30 seconds.
I can re-add them if you want.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
sbordet: All these attributes can only be set after the request has been handled, just before exiting the Servlet handling.
In a `Customizer` we don't have access to the Servlet name (nor the other values we store in these attributes), so they must really be done here.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/CustomRequestLogTest.java
sbordet: Re-added.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
joakime: Fair enough

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/CustomRequestLogTest.java
joakime: Fair enough

joakime (approved):

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -781,14 +782,27 @@ void onTrailers(HttpFields trailers)
...
+        ServletContextRequest.ServletApiRequest apiRequest = _request.getServletApiRequest();
         if (LOG.isDebugEnabled())
-            LOG.debug("onCompleted for {} written={}", httpServletRequest.getRequestURI(), getBytesWritten());
+            LOG.debug("onCompleted for {} written={}", apiRequest.getRequestURI(), getBytesWritten());
 
         long idleTO = _configuration.getIdleTimeout();
         if (idleTO >= 0 && getIdleTimeout() != _oldIdleTimeout)
             setIdleTimeout(_oldIdleTimeout);
 
+        if (getServer().getRequestLog() != null)
```
gregw, line 793: Should this be an instanceof for CustomRequestLog check?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -781,14 +782,27 @@ void onTrailers(HttpFields trailers)
...
+        {
+            Authentication authentication = apiRequest.getAuthentication();
```
gregw, line 795: Can we add a TODO here for only adding the attributes if the custom request logger has these things in the format string?   Better yet, in this PR, implement that check in the doStart method and have an EnumSet for if user name; real path; handlerName are needed, then only set the attributes if those booleans are set.

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/HttpChannel.java
```diff
@@ -945,12 +946,25 @@ protected boolean checkAndPrepareUpgrade()
...
     {
         if (LOG.isDebugEnabled())
-            LOG.debug("onCompleted for {} written={}", getRequest().getRequestURI(), getBytesWritten());
+            LOG.debug("onCompleted for {} written={}", _request.getRequestURI(), getBytesWritten());
 
         long idleTO = _configuration.getIdleTimeout();
         if (idleTO >= 0 && getIdleTimeout() != _oldIdleTimeout)
             setIdleTimeout(_oldIdleTimeout);
 
+        if (getServer().getRequestLog() != null)
```
gregw, line 955: Ditto comments from ServletChannel

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/CustomRequestLog.java
```diff
@@ -318,11 +319,13 @@
...
 public class CustomRequestLog extends ContainerLifeCycle implements RequestLog
 {
-    protected static final Logger LOG = LoggerFactory.getLogger(CustomRequestLog.class);
-
     public static final String DEFAULT_DATE_FORMAT = "dd/MMM/yyyy:HH:mm:ss ZZZ";
     public static final String NCSA_FORMAT = "%{client}a - %u %t \"%r\" %s %O";
     public static final String EXTENDED_NCSA_FORMAT = NCSA_FORMAT + " \"%{Referer}i\" \"%{User-Agent}i\"";
+    public static final String HANDLER_NAME = CustomRequestLog.class.getName() + ".handlerName";
+    public static final String REAL_PATH = CustomRequestLog.class.getName() + ".realPath";
+    public static final String USER_NAME = CustomRequestLog.class.getName() + ".userPrincipal";
```
gregw, line 327: Make these an `enum` and then an EnumSet can be used to record which of them are required.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -781,14 +782,27 @@ void onTrailers(HttpFields trailers)
...
+            if (authentication instanceof Authentication.User userAuthentication)
+                _request.setAttribute(CustomRequestLog.USER_NAME, userAuthentication.getUserIdentity().getUserPrincipal().getName());
```
gregw, line 797: Ultimately, it might be best to have an `EnumMap<CustomRequestLog.ATTR,String>` as part of the request, as a more efficient way to pass these items for requests that don't use the attributes.

This probably needs to be tested... but if we do decide that just using attributes is best/simplest, then perhaps we give up on the whole lazy creation of the attributes map underneath as needless complexity.

i.e. either we continue to strive to avoid routine request attributes or we remove lazy request attribute map.  It is silly to be lazy and routine.

gregw (changes requested):
Good that it is working now, but I'd like to see some more optimizations.  I could be convinced to do those in a later PR... but perhaps we can do now?

sbordet:
@gregw your comments are too specific to a particular implementation of `RequestLog`, but we must be more generic.
I have recorded your concerns in #8715, so for now I'll merge this PR.

PR #8711, opened 2022-10-13
ResourceCollection should not have a path

gregw:
A `ResourceCollection` should have a null path and URI.  It is by definition a collection of different paths and URIs, so it cannot just pick one of them and return it.   This is hiding lots of bugs in the code where `resource.getPath()` is called and assumes it will always get a non-null value that holds all the required resources.  However, when scanning for descriptors, annotations, TLDs etc. the underlying resource is often a collection, thus code that uses getPath is currently only scanning the first resource in the collection.

This PR fixes `ResourceCollection`, but will almost certainly break some other tests.   I think we want to use other PRs to fix those uses of `getPath` first before considering completing this PR.


joakime:
If we go this path, then I think `ResourceCollection` should also not have  ...

* `lastModified()`
* `length()`
* `getURI()`
* `getFileName()`
* `getName()`

The changes you are doing here also apply to `MemoryResource` btw

joakime:
The code overall would be so much easier with a `List<>` instead of hiding this collection behind an interface that is for a single entry.

gregw:
> If we go this path, then I think `ResourceCollection` should also not have ...
> 
> * `lastModified()`

lastModified has a sensible and meaningful interpretation for composite directories: the most recently modified time of all the composites

> * `length()`

Agreed.

> * `getURI()`

Yep, should return null... I think this PR already does.

> * `getFileName()`

I'm on the fence about this one.... if for example you have a composite base resource and resolve "WEB-INF", then you may get back a composite that contains `WEB-INF` directories from multiple resources in the base composite.   In that case, since they are all called `WEB-INF`, I think it is reasonable to return "WEB-INF" for the filename.    If they don't agree, then return null. That's how I have implemented it for now.

Flip side, is that if it always returned null, I don't think it would be a big problem... except perhaps for debugging.... maybe when we are naming contexts etc.   So I think on balance going with the impl in this PR is OK.

> * `getName()`

Yep.

> The changes you are doing here also apply to `MemoryResource` btw

Oh I thought we got rid of that class?  Is it still used? If not, let's kill it.    If we are still using it, then I'll also review.

gregw:
> The code overall would be so much easier with a `List<>` instead of hiding this collection behind an interface that is for a single entry.

Absolutely not!  

We have a bunch of methods that can apply to a composite collection of resources that we have to think carefully about how to interpret them against a collection: is it the first; is it a combination; is it null; is it allowed at all?    With a `ResourceCollection` class we can put all those difficult decisions in one place and make them consistent.  We can police the creation of the collection to only be non-null directory resources.

If we pass around a `List<Resource>` instead, then
 - every usage of that list will have to make the difficult choices:  is it the first; is it a combination; is it null; is it allowed at all?  Inevitably different choices will be made in different places so we will have inconsistency... a file might be visible for annotation scanning but not for TLD scanning!
- We will have duplicated code iterating over the lists in multiple places... or perhaps we decide to have static utility methods into which we pass the list? In which case we have just re-invented object oriented programming.
- Since we could not enforce the construction of the `List<Resource>` to contained only non-null directory resources, then we would need to spread sanity and null checking code everywhere.  We'd need to check on every usage of the collection rather than just check on the creation of the collection.  So instead of doing those checks once, we could end up doing them hundreds of thousands of times per second on every request.
- A directory is itself just a collection of other resources and we have many places where a single such collection is simple to deal with (e.g. scanning WEB-INF).  Turning that simple abstraction of a collection of files into a collection of collections of files is needlessly complex. 
- The readability of the code will be reduced.  Instead of `web-inf.lastModified()` we will have `Instance last = null; for (Resource r : list) { Instant l = r.lastModified(); if (last == null || l.isAfter(last)) last = l; } return last;`... repeated.... in many places.... ?!?!?!?

Note that there are valid usages for `List<Resource>` and we have many examples in the code.   We might have multiple descriptors, or multiple jar files or multiple webapps etc.   In these cases, the list is not trying to represent a single resource, but is genuinely representing multiple resources that the caller needs to explicitly handle in whatever capacity it is doing.  These are not use-cases for `ResourceCollection` because they are not abstracting a single resource.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -230,7 +230,10 @@ public long length()
      */
     public InputStream newInputStream() throws IOException
     {
-        return Files.newInputStream(getPath(), StandardOpenOption.READ);
+        Path path = getPath();
+        if (path == null)
+            return null;
```
lachlan-roberts, line 205: I would expect this to throw `IOException` not return null from the Javadoc. 
So I think the javadoc should be updated as well.

joakime:
> > If we go this path, then I think `ResourceCollection` should also not have ...
> > 
> > * `lastModified()`
> 
> lastModified has a sensible and meaningful interpretation for composite directories: the most recently modified time of all the composites

I'm not sure that's a valid use case, where do you see this being used like this?

> > * `length()`
> 
> Agreed.
> 
> > * `getURI()`
> 
> Yep, should return null... I think this PR already does.
> 
> > * `getFileName()`
> 
> I'm on the fence about this one.... if for example you have a composite base resource and resolve "WEB-INF", then you may get back a composite that contains `WEB-INF` directories from multiple resources in the base composite. In that case, since they are all called `WEB-INF`, I think it is reasonable to return "WEB-INF" for the filename. If they don't agree, then return null. That's how I have implemented it for now.

Interesting, I'll have to see if we use this from a Directory based Resource.

> Flip side, is that if it always returned null, I don't think it would be a big problem... except perhaps for debugging.... maybe when we are naming contexts etc. So I think on balance going with the impl in this PR is OK.

Most debugging should be using either `getName()` or `getURI()` as it's more clear.

> > * `getName()`
> 
> Yep.
> 
> > The changes you are doing here also apply to `MemoryResource` btw
> 
> Oh I thought we got rid of that class? Is it still used? If not, let's kill it. If we are still using it, then I'll also review.

The original proposal was in (the now closed, un-merged) PR #8470

Now, the work Lachlan is doing for HttpContent.Factory can easily get rid of MemoryResource, as it can serve a static HttpContent for `jetty-dir.css` as a fallback.
Lets wait for that work to complete and it'll be easy to remove MemoryResource.

jetty-ee9/jetty-ee9-webapp/src/main/java/org/eclipse/jetty/ee9/webapp/WebAppClassLoader.java
```diff
@@ -112,7 +113,7 @@ public interface Context extends ClassVisibilityChecker
          */
         boolean isParentLoaderPriority();
 
-        ResourceCollection getExtraClasspath();
+        Resource getExtraClasspath();
```
gregw: @lorban @joakime @sbordet  I believe it was a miss-use of `ResourceCollection` when this was changed from a `List<Resource>`.    We should very VERY rarely ever explicitly use a `ResourceCollection`, because the whole point of that class is to look like a singe resource.   If anything needs to iterate over the contained resources, then it probably should be using a `List<Resources>` rather than a `ResourceCollection`.

However, for not, I have only partially reverted to try how it looks if extra classpath is just a `Resource`, which may or may not be a collection.   This requires the `addClassPath` method below to be somewhat `ResourceCollection` away, as it needs to extract URLs.  I've done this for now with a static stream method.... but not sure I like.   Reverting all the way back to `List<Resource>` may be the best thing to do here.

jetty-ee9/jetty-ee9-webapp/src/main/java/org/eclipse/jetty/ee9/webapp/WebAppClassLoader.java
```diff
@@ -223,23 +219,30 @@ public Context getContext()
...
-     * @param resources The resources to add to the classpath
-     * @throws IOException if unable to add classpath from resource
+     * @param resource The resources to add to the classpath
      */
-    public void addClassPath(Resource resources)
-        throws IOException
+    public void addClassPath(Resource resource)
     {
-        for (Resource resource: resources)
+        ResourceCollection.stream(resource).forEach(r ->
```
gregw: Since any `Resource` might be a collection, if we need URLs, URIs or Paths from a resource, then we should use this stream abstraction to get back the one or more values.  The stream can then be used to pick the first, collect into a list or do forEach handling like I am below:

jetty-ee9/jetty-ee9-webapp/src/main/java/org/eclipse/jetty/ee9/webapp/WebAppClassLoader.java
gregw: maybe `ResourceCollection` should be renamed to `MultiResource` to emphasise that it is-a `Resource` and should not be thought of as a `Collection` (which is an implementation detail).

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactory.java
```diff
@@ -35,13 +35,15 @@
     /**
      * <p>Make a Resource containing a collection of other resources</p>
      * @param resources multiple resources to combine as a single resource. Typically, they are directories.
-     * @return A Resource of multiple resources.
+     * @return A Resource of multiple resources or a single resource if only 1 is passed, or null if none are passed
      * @see ResourceCollection
      */
-    static ResourceCollection combine(List<Resource> resources)
+    static Resource combine(List<Resource> resources)
```
gregw, line 41: `combine` should return a `Resource` and not a `ResourceCollection`, since if the passed list has only a single entry, then a collection is not needed.    The caller should not need to know if it is a collection or not.

jetty-ee9/jetty-ee9-webapp/src/main/java/org/eclipse/jetty/ee9/webapp/WebAppClassLoader.java
gregw: Better name options include:
 + `MultiResource`
 + `CompositeDirectoryResource`
 + `CombinedResource`
 + `CombinedDirectoryResource`

gregw:
@joakime nudge

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ErrorProcessor.java
```diff
@@ -110,7 +110,7 @@ public void process(Request request, Response response, Callback callback)
 
             try
             {
-                generateAcceptableResponse(request, response, code, message, cause, callback);
+                generateResponse(request, response, code, message, cause, callback);
```
joakime: Unrelated to this PR.
Plus, this has "acceptable" in it's name as it's generating based on the `Accept` header.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ReHandlingErrorProcessor.java
```diff
@@ -0,0 +1,146 @@
...
+import org.eclipse.jetty.util.Callback;
+import org.eclipse.jetty.util.ExceptionUtil;
+import org.eclipse.jetty.util.URIUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * An ErrorProcessor that can re-handle a request at an error page location.
+ */
+public abstract class ReHandlingErrorProcessor extends ErrorProcessor
```
joakime: Also unrelated to the scope of this PR,
Can you pull this out into a different PR?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/AttributeNormalizer.java
```diff
@@ -11,7 +11,7 @@
 // ========================================================================
 //
 
-package org.eclipse.jetty.ee9.quickstart;
+package org.eclipse.jetty.util.resource;
```
joakime, line 14: Different PR for this too

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/CombinedResource.java
```diff
@@ -26,36 +23,52 @@
...
 
 import org.eclipse.jetty.util.URIUtil;
 
 /**
- * A collection of Resources.
- * Allows webapps to have multiple sources.
- * The first resource in the collection is the main resource.
- * If a resource is not found in the main resource, it looks it up in
- * the order the provided in the constructors
+ * Multiple resource directories presenteda as a single Resource.
```
joakime: Typo `presented as`

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/CombinedResource.java
```diff
@@ -26,36 +23,52 @@
...
+public class CombinedResource extends Resource
 {
-    private final List<Resource> _resources;
+    /**
+     * Stream a resource to its component resources (if any).
+     * @param resource The resource to stream
+     * @return If the passed resource is a ResourceCollection, then it's contained
+     *         resources are streamed, otherwise the passed resource itself if streamed
+     */
+    public static Stream<Resource> stream(Resource resource)
```
joakime: Wrong place for this.
Perhaps this should be in `Resource` or `Resources` ?

joakime: Why not just have `.stream()` on the `Resource.stream()` interface instead?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/CombinedResource.java
```diff
@@ -26,36 +23,52 @@
...
-     * Instantiates a new resource collection.
-     *
-     * @param resources the resources to be added to collection
+     * <p>Make a Resource containing a combination of other resources</p>
+     * @param resources multiple resources to combine as a single resource. Typically, they are directories.
+     * @return A Resource of multiple resources or a single resource if only 1 is passed, or null if none are passed
+     * @see CombinedResource
      */
-    ResourceCollection(List<Resource> resources)
+    static Resource combine(List<Resource> resources)
```
joakime, line 40: Isn't this on `ResourceFactory.combine()` already?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/CombinedResource.java
```diff
@@ -127,127 +153,67 @@ public Resource resolve(String subUriPath)
...
-            }
-        }
         return null;
     }
 
     @Override
     public String getFileName()
     {
+        String filename = null;
+        // return a non-null filename only if all resources agree on the same name.
```
joakime, line 186: Perhaps we should make `getFileName()` not be the last segment name in case of Directory?   Is there a use case were in we care about, and use `getFileName()` in a directory case?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -198,22 +198,28 @@ public long length()
...
     }
 
     /**
      * Readable ByteChannel for the resource.
      *
-     * @return an readable bytechannel to the resource or null if one is not available.
+     * @return a readable {@link java.nio.channels.ByteChannel} to the resource or null if one is not available.
      * @throws IOException if unable to open the readable bytechannel for the resource.
      */
     public ReadableByteChannel newReadableByteChannel() throws IOException
```
joakime, line 215: This should probably be a `SeekableByteChannel` instead.
(Would be more generally useful in the static byte-range support as well)

joakime (changes requested):
My not-yet-fully-reviewed review

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime: The `Files.newByteChannel()` already returns `SeekableByteChannel`

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/CombinedResource.java
gregw: The style we have been adopting in jetty-12 is to avoid adding methods to interfaces if they can be implemented in a static like this.

However, this is not necessarily the correct place for this static.... and there is already the iterator API that does the same thing.... so I will review....

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/CombinedResource.java
gregw: It is, which calls this version, which has access to the private `gatherUniqueFlatResourceList` method which then calls the private constructor, that no longer does the sanity checks, as they are not needed when creating a CombinedResource from resolve.   So this allows us to avoid needless iterations and checks.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/CombinedResource.java
gregw: We can use the fileName to generate a context name.  I'm not 100% sure we still use this method for that, but it is a reasonable use-case to ask a directory for the name of it's last segment.

This method exists and I think this is a reasonable interpretation of it for a combined resource.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
gregw: probably... I'll do that in another PR.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/AttributeNormalizer.java
gregw: Actually this change is for this PR.  It turns out that the `AttributeNormalizer` has for a while been assuming that a Path is available, so the main change here is to make it use the stream API for CombinedResources to generate multiple URI to resource mappings.

But as the Normalizer code was identical in ee9 and ee10, I also moved to resource to avoid duplicating the change. 

So I think the change does belong here.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/CombinedResource.java
gregw: Note that this method  is only package scope, so it is not public API.  It is just to get the implementation in the right file rather than in ResourceFactory.java

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
joakime: Opened as PR #8772

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/CombinedResource.java
```diff
@@ -30,32 +27,34 @@
...
-     * Instantiates a new resource collection.
-     *
-     * @param resources the resources to be added to collection
+     * <p>Make a Resource containing a combination of other resources</p>
+     * @param resources multiple resources to combine as a single resource. Typically, they are directories.
+     * @return A Resource of multiple resources or a single resource if only 1 is passed, or null if none are passed
+     * @see CombinedResource
      */
-    ResourceCollection(List<Resource> resources)
+    static Resource combine(List<Resource> resources)
```
lachlan-roberts, line 40: Should this also have a `Resource combine(Resource... resources)` same as the `ResourceFactory`?

jetty-ee10/jetty-ee10-maven-plugin/src/test/java/org/eclipse/jetty/ee10/maven/plugin/TestWebAppPropertyConverter.java
```diff
@@ -152,10 +152,10 @@ public void testFromProperties() throws Exception
         assertEquals(true, webApp.isPersistTempDirectory());
         assertEquals(war.getAbsolutePath(), webApp.getWar());
         assertEquals(webXml.getAbsolutePath(), webApp.getDescriptor());
-        assertThat(webApp.getBaseResource(), instanceOf(ResourceCollection.class));
+        assertThat(webApp.getBaseResource(), instanceOf(CombinedResource.class));
 
-        ResourceCollection resourceCollection = (ResourceCollection)webApp.getBaseResource();
-        List<URI> actual = resourceCollection.getResources().stream().filter(Objects::nonNull).map(Resource::getURI).toList();
+        CombinedResource combinedResource = (CombinedResource)webApp.getBaseResource();
+        List<URI> actual = combinedResource.getResources().stream().filter(Objects::nonNull).map(Resource::getURI).toList();
```
lachlan-roberts: Probably better to use `Resource.stream()`.

lachlan-roberts (approved):
Looks good to me but there are some test failures.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ReHandlingErrorProcessor.java
gregw: synced with head.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/CombinedResource.java
gregw: moved to Resource

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/CombinedResource.java
gregw: It's not a public API.  It is just there to move the implementation into the correct file.   So if the ... is not needed, no need to add it.

Issue #8716, opened 2022-10-14
Multiple Host header values handled poorly


joakime:
**Jetty version(s)**
9.4.x thru 12.0.x

**Java version/vendor** `(use: java -version)`
All

**OS type/version**
All

**Description**
When a request is received with multiple host headers it's handled poorly.

This was pointed out by @Hexcles at https://github.com/eclipse/jetty.project/issues/7278#issuecomment-1279207755

Example using the jetty 11 demos.

```
$ curl -vvvv -H"Host: localhost, foo, bar, zed" http://localhost:8080/test/dump/info
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET /test/dump/info HTTP/1.1
> Host: localhost, foo, bar, zed
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Set-Cookie: visited=yes
< Expires: Thu, 01 Jan 1970 00:00:00 GMT
< Set-Cookie: JSESSIONID=node013ntsnazjuutz1gpikvehddsnk1.node0; Path=/test
< Content-Type: text/html;charset=utf-8
< Vary: Accept-Encoding
< Content-Length: 10477
< Server: Jetty(11.0.11)
< 
<html>
<body>
<h1>Dump Servlet</h1>
...(snip)...
<th align="right">getRequestURI:&nbsp;</th><td>/test/dump/info</td></tr><tr>
<th align="right">getRequestURL:&nbsp;</th><td>http://localhost, foo, bar, zed/test/dump/info</td></tr><tr>
<th align="right">getScheme:&nbsp;</th><td>http</td></tr><tr>
<th align="right">getServerName:&nbsp;</th><td>localhost, foo, bar, zed</td></tr><tr>
...(snip)...
```

So the key bits in the above example are ...

* HTTP/1.1 was used
* The request header **`Host: localhost, foo, bar, zed`** was sent
* Jetty parsed it without error, and it resulted in odd API behavior
  * `HttpServletRequest.getRequestURL()` is **`http://localhost, foo, bar, zed/test/dump/info`**
  * `HttpServletRequest.getServerName()` is **`localhost, foo, bar, zed`**

joakime:
Looks like Jetty interprets the all field values, including commas, as a single "Host" (or Server Name), complete with commas.

Hexcles:
[...]

PR #8717, opened 2022-10-14
Issue #8716 - Handle bad host/authority headers better

joakime:
This is a minimal version of code to reject bad Host/authority header usage.

The more robust validation can be found in the closed (and never merged) PR #7279 


jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
```diff
@@ -1028,6 +1028,8 @@ else if (_endOfContent == EndOfContent.CHUNKED_CONTENT)
                         break;
 
                     case HOST:
+                        if (_host)
+                            throw new BadMessageException(HttpStatus.BAD_REQUEST_400, "Bad Host: multiple headers");
```
joakime, line 1032: This prevents duplicate `Host` headers in the request

jetty-http/src/test/java/org/eclipse/jetty/http/HttpParserTest.java
```diff
@@ -2040,12 +2040,20 @@ public void testHostPort()
...
-    public void testHostBadPort()
+    @ParameterizedTest
+    @ValueSource(strings = {
+        "Host: whatever.com:xxxx\r\n",
+        "Host: myhost:testBadPort\r\n",
+        "Host: a b c d\r\n",
+        "Host: hosta, hostb, hostc\r\n",
+        "Host: hosta,hostb,hostc\r\n",
+        "Host: hosta\r\nHost: hostb\r\nHost: hostc\r\n"
+    })
```
sbordet, line 2053: Avoid the terminal `\r\n` in these parameters, and add it on the request string below.

jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
```diff
@@ -783,8 +785,16 @@ public void testEncodedNotParams() throws Exception
...
-    public void testInvalidHostHeader() throws Exception
+    @ParameterizedTest
+    @ValueSource(strings = {
+        "Host: whatever.com:xxxx\r\n",
+        "Host: myhost:testBadPort\r\n",
+        "Host: a b c d\r\n",
+        "Host: hosta, hostb, hostc\r\n",
+        "Host: hosta,hostb,hostc\r\n",
+        "Host: hosta\r\nHost: hostb\r\nHost: hostc\r\n"
+    })
```
sbordet, line 796: Avoid the terminal `\r\n` in these parameters, and add it on the request string below.

jetty-util/src/main/java/org/eclipse/jetty/util/HostPort.java
```diff
@@ -96,6 +100,29 @@ else if (authority.charAt(0) == '[')
         }
     }
 
+    /**
+     * Performs some safety checks on the authority.
+     *
+     * @param authority the authority to test
+     * @return true if the authority passes as valid
+     */
+    private boolean isValidAuthority(String authority)
```
sbordet: Rename to `isAuthorityValid()`?

The `Host` value is defined in [RFC 7230, sec. 5.4](https://datatracker.ietf.org/doc/html/rfc7230#section-5.4), which delegates to [RFC 3986, sec. 3.2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2).

Besides IPv4 and IPv6 representations, it is:

```
host        = IP-literal / IPv4address / reg-name
reg-name    = *( unreserved / pct-encoded / sub-delims )
unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
pct-encoded = "%" HEXDIG HEXDIG
sub-delims  = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="
```

In particular, `,` is allowed.

jetty-util/src/main/java/org/eclipse/jetty/util/HostPort.java
```diff
@@ -96,6 +100,29 @@ else if (authority.charAt(0) == '[')
...
+            int codepoint = authority.codePointAt(i);
+            if (codepoint == ',')
+                return false;
+            if (Character.isISOControl(codepoint))
+                return false;
+            if (Character.isWhitespace(codepoint))
+                return false;
+        }
+        return true;
+    }
```
sbordet, line 130: Besides the `,` that is allowed, I think for [IDN](https://unicode.org/faq/idn.html) we need to review `isWhitespace()`, because IDN allows zero-width joiners, etc.
Also true that it would be a very rare case.

jetty-util/src/main/java/org/eclipse/jetty/util/HostPort.java
```diff
@@ -81,6 +81,10 @@ else if (authority.charAt(0) == '[')
                 }
                 else
                 {
+                    if (!isValidAuthority(authority))
+                    {
+                        throw new IllegalArgumentException("Bad Authority");
+                    }
```
sbordet: Drop braces for consistency with rest of the code.

sbordet (changes requested):

jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
joakime: I have a 3 line param, see line 795

jetty-http/src/test/java/org/eclipse/jetty/http/HttpParserTest.java
joakime: I have a 3 line param, see line 2050

jetty-util/src/main/java/org/eclipse/jetty/util/HostPort.java
joakime: Good point.
I added the check for `,` just to catch multi-value http fields.
But that was obviously a mistake.

jetty-util/src/main/java/org/eclipse/jetty/util/HostPort.java
joakime: We cannot have raw unicode in an http field (at least for HTTP/1.1 that is).
Also, I would have expected either a percent encoded utf-8, or puny-code for IDN support.

joakime: I'll add a few test cases for IDN with ZWJ

joakime: Tests added

jetty-util/src/main/java/org/eclipse/jetty/util/HostPort.java
joakime: Doing check in HostPortHttpField properly now.

jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
joakime: Changed.

jetty-http/src/test/java/org/eclipse/jetty/http/HttpParserTest.java
joakime: Changed.

sbordet:
@joakime @gregw I'm on the fence about this PR.

I think we should fix the double `Host` header, but we should not attempt to do any check or parsing of the `Host` value besides validity of host:port format (already done by `HostPort`).

If someone sends `Host: a, b, c` then the authority is `a, b, c` and so we report.

Looking for `,` or doing `QuotedCSV` parsing, etc. I think it's too much with no additional benefit and a possible performance impact on all the valid cases.

jetty-security/src/test/java/org/eclipse/jetty/security/ConstraintTest.java
```diff
@@ -1134,7 +1134,6 @@ public void testNonFormPostRedirectHttp11() throws Exception
 
         response = _connector.getResponse("POST /ctx/auth/info HTTP/1.1\r\n" +
             "Host: test\r\n" +
-            "Host: localhost\r\n" +
```
joakime, line 1137: Gotcha!  bad testcase with two Host headers.

jetty-util/src/test/java/org/eclipse/jetty/util/HostPortTest.java
```diff
@@ -31,14 +31,10 @@ public static Stream<Arguments> validAuthorityProvider()
...
         return Stream.of(
             Arguments.of("", "", null),
-            Arguments.of(":80", "", "80"),
             Arguments.of("host", "host", null),
             Arguments.of("host:80", "host", "80"),
             Arguments.of("10.10.10.1", "10.10.10.1", null),
             Arguments.of("10.10.10.1:80", "10.10.10.1", "80"),
-            Arguments.of("[0::0::0::1]", "[0::0::0::1]", null),
-            Arguments.of("[0::0::0::1]:80", "[0::0::0::1]", "80"),
-            Arguments.of("0:1:2:3:4:5:6", "[0:1:2:3:4:5:6]", null),
```
joakime, line 41: All of the above examples were moved from "valid" to "invalid" as they are not seen as valid by Java itself.

They all fail in `InetAddress`, and `URI` usages for various reasons (see `invalidAuthorityProvider` comments for specific reasons)

jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -1122,6 +1122,56 @@ public static boolean hasScheme(String uri)
...
+                else
+                    return false;
+            }
+            // sub-delims
+            if (c == '!' || c == '$' || c == '&' || c == '\'' || c == '(' || c == ')' ||
+                c == '*' || c == '+' || c == ',' || c == ';' || c == '=')
+            {
+                continue;
+            }
+            return false;
```
gregw: Rather than 4 if, 3 of them with multiple comparisons.... it would be faster to have token lookup array.... in fact you could probably use `HttpTokens.TOKENS` and switch on the `Token`?    if not, then perhaps a secondary lookup in the same class?

gregw (changes requested):
Looks good... but we can make the isRegName more efficient with a lookup table.

jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -1122,6 +1122,56 @@ public static boolean hasScheme(String uri)
         return false;
     }
 
+    /**
+     * True if token is a <a href="https://www.rfc-editor.org/rfc/rfc3986">RFC3986</a> {@code reg-name}
+     *
+     * @param token the to test
+     * @return true if the token passes as a valid reg-name
+     */
+    public static boolean isRegName(String token)
```
sbordet: While "regName" is what the spec calls it, I would change it to `isValidHost()`.
I have no idea what "reg" means in this context, and I don't want to look it up (lazy), so if it's named `isValidHost()` I'll know what it is.

sbordet (changes requested):

jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: Per https://www.rfc-editor.org/rfc/rfc3986

This is `reg-name` only, and not `host` as this method does not validate `IP-literal` or `IPv4address` (which is also defined in `host`)

jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -41,6 +42,24 @@
...
+    private static final String UNRESERVED = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ-._~";
+    private static final String SUBDELIMS = "!$&'()*+,;=";
+    private static final String REGNAME = UNRESERVED + SUBDELIMS;
+
+    // Allowed characters in https://www.rfc-editor.org/rfc/rfc3986 reg-name
+    private static final boolean[] REGNAME_ALLOWED;
+
+    static
+    {
+        REGNAME_ALLOWED = new boolean[127];
```
sbordet: Should be 128.

jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -41,6 +42,24 @@
...
+        Arrays.fill(REGNAME_ALLOWED, false);
+        for (char c : REGNAME.toCharArray())
+        {
+            REGNAME_ALLOWED[c] = true;
+        }
+    }
+
     // Use UTF-8 as per http://www.w3.org/TR/html40/appendix/notes.html#non-ascii-chars
     public static final Charset __CHARSET = StandardCharsets.UTF_8;
```
sbordet, line 63: Make `private` and possibly remove the constant altogether.

sbordet (changes requested):

jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: Cool, lemme do this in a different PR

joakime: Actually, can't do this for Jetty 10.0.x (where this PR is).

BUT I will do it for Jetty 12.0.x

joakime: Took care of this (along with other URIUtil cleanup) in `jetty-12.0.x` PR #8861

sbordet (approved):

gregw (approved):

PR #8718, opened 2022-10-16
Moved FastCGIProxyServlet in ee9 to FastCGIProxyHandler in core.

sbordet:
Moved TryFilesFilter in ee9 to TryPathsHandler in core. Update Jetty modules to properly setup the class-path in case of proxy usage.

Updated and expanded test cases.


jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpTester.java
```diff
@@ -504,6 +510,14 @@ public ByteBuffer generate()
             }
         }
 
+        public byte[] generateBytes()
+        {
+            ByteBuffer byteBuffer = generate();
+            byte[] bytes = new byte[byteBuffer.remaining()];
+            byteBuffer.get(bytes);
+            return bytes;
```
joakime: Don't think this method is needed.

This can be shortened to ...

`return BufferUtil.toArray(generate());`

Which is how it's used in test cases.

Either as a String ...

``` java
socket.write(BufferUtil.toString(request.generate()));
```

or as a byte array ...

``` java
socket.write(BufferUtil.toArray(request.generate()));
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/TryPathsHandler.java
```diff
@@ -0,0 +1,107 @@
...
+        Request.WrapperProcessor result = new Request.WrapperProcessor(new TryPathsRequest(request, interpolated));
+        return result.wrapProcessor(super.handle(result));
+    }
+
+    // maint.html, $path, /ancient/$path, index.php?p=$path
+
+    // ctxh // rb = /srv/www/cometd.org/static
+    //  `tryh
+    //    `resh
+    //      `phph
```
joakime: Do we need these comments still?

jetty-core/jetty-fcgi/jetty-fcgi-proxy/src/main/config/modules/fcgi-proxy.mod
```diff
@@ -0,0 +1,13 @@
...
+[tags]
+fcgi
+proxy
+
+[depend]
+fcgi
+proxy
+
+[lib]
+lib/fcgi/jetty-fcgi-proxy-${jetty.version}.jar
```
joakime, line 13: do we ship with this class in the jetty-home archive?
(wondering if we should be downloading here or not)

jetty-core/jetty-fcgi/jetty-fcgi-server/src/main/config/modules/fcgi.mod
```diff
@@ -1,16 +1,13 @@
...
 
 [lib]
-lib/jetty-proxy-${jetty.version}.jar
-lib/fcgi/*.jar
-
-[ini-template]
-## For configuration of FastCGI contexts, see
-## https://www.eclipse.org/jetty/documentation/current/fastcgi.html
+lib/fcgi/jetty-fcgi-client-${jetty.version}.jar
+lib/fcgi/jetty-fcgi-server-${jetty.version}.jar
```
joakime, line 13: Same here, have we verified these exist in the jetty-home archive?

tests/test-distribution/test-distribution-common/src/test/java/org/eclipse/jetty/tests/distribution/DistributionTests.java
```diff
@@ -1174,4 +1174,107 @@ public void testDryRunProperties() throws Exception
...
+    @Test
+    public void testFastCGIProxying() throws Exception
+    {
+        String jettyVersion = System.getProperty("jettyVersion");
+        JettyHomeTester distribution = JettyHomeTester.Builder.newInstance()
+            .jettyVersion(jettyVersion)
+            .mavenLocalRepository(System.getProperty("mavenRepoPath"))
+            .build();
+
+        List<String> args1 = List.of("--add-modules=resources,http,fcgi-proxy,core-deploy");
```
joakime: This one tests `fcgi-proxy`, but what about `fcgi` module itself?

Also, what's all of the "fcgi" stuff in the `testBeforeDirectiveInModule` method in this class about? (it's also a `@Disabled` test)

joakime (changes requested):
I don't think the new stuff on `HttpTester` is needed.
A few stray comments that can probably be gotten rid of.
Also a few questions about the *.mod testing.

jetty-core/jetty-fcgi/jetty-fcgi-proxy/src/main/config/modules/fcgi-proxy.mod
sbordet: Not sure what class you refer to. 
But yes the `jar` is in `jetty-home`, I have a test in `DistributionTests` that prove it.

jetty-core/jetty-fcgi/jetty-fcgi-server/src/main/config/modules/fcgi.mod
sbordet: Yes they exist, proved by the test in `DistributionTests`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpTester.java
sbordet: Good point, removed.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/TryPathsHandler.java
sbordet: Good catch, I'll remove them, since now that structure is in a test case in `TryPathsHandlerTest`.

tests/test-distribution/test-distribution-common/src/test/java/org/eclipse/jetty/tests/distribution/DistributionTests.java
sbordet: The server `fcgi` module is really only for testing.
It is available as a `*.mod` and just add the relevant jars to the class-path, but has no XML file associated, again because it's mostly for testing.

The disabled tests will be re-enabled in a different PR I'm working on.

jetty-core/jetty-fcgi/jetty-fcgi-server/src/main/config/modules/fcgi.mod
joakime: AFAICT, this module (just `fcgi`) does not have a `DistributionTests` method.

joakime (approved):

PR #8722, opened 2022-10-18
Jetty 12 - Re-enabled distribution tests that were disabled.

sbordet:

### DIFF
--- a/tests/test-distribution/test-distribution-common/src/test/java/org/eclipse/jetty/tests/distribution/DistributionTests.java
+++ b/tests/test-distribution/test-distribution-common/src/test/java/org/eclipse/jetty/tests/distribution/DistributionTests.java
@@ -13,20 +13,22 @@
 
 package org.eclipse.jetty.tests.distribution;
 
-import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileWriter;
 import java.net.URI;
-import java.nio.charset.StandardCharsets;
+import java.net.http.WebSocket;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.nio.file.StandardOpenOption;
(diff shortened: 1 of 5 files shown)


tests/test-distribution/test-distribution-common/src/test/java/org/eclipse/jetty/tests/distribution/DistributionTests.java
```diff
@@ -200,26 +191,27 @@ public void testSimpleWebAppWithJSPandJSTL(String env) throws Exception
...
     }
 
-    @Disabled //TODO runtime can't find org.objectweb.asm on module path?
     @ParameterizedTest
-    @ValueSource(strings = {"ee9", "ee10"})
+    @ValueSource(strings = {"ee10"})
     public void testSimpleWebAppWithJSPOnModulePath(String env) throws Exception
     {
+        // Testing with env=ee9 is not possible because jakarta.transaction:1.x
+        // does not have a proper module-info.java, so JPMS resolution will fail.
```
joakime, line 199: Sounds like a new issue needs to be filed for this

joakime (approved):
looks good, just open an issue about that bad ee9 dependency from jakarta with jpms

tests/test-distribution/test-distribution-common/src/test/java/org/eclipse/jetty/tests/distribution/DistributionTests.java
sbordet: A new issue to what project?

I don't think it can be fixed, as `jakarta.transaction:1.x` should remain Java 8 compatible (so no `module-info.java`) as per JEE 9.
It only has an `Automatic-Module-Name` but that is not enough to run with full JPMS.

I don't think the JakartaEE project will modify the jar, since they have already fixed it in `jakarta.transaction:2.x` (and in fact, our ee10 tests pass without problems).

PR #8731, opened 2022-10-18
Jetty 10 - #8723: Provide a thread-safe way to modify HttpClient proxies

### LINKED ISSUE #8723: Provide a thread-safe way to modify HttpClient proxies at runtime
**Jetty version(s)**
11.0.12

**Enhancement Description**
`HttpClient.getProxyConfiguration().getProxies()` is backed by a `ArrayList` which is not thread-safe.

Currently, I am forced to externally lock each time I make an HTTP request, or modifying proxies. I am making thousands of requests per second, which means that I need to acquire a lock thousands of times per second.

**Possible solutions**: 

1. Force users to externally-synchronize access.
2. Replace `ArrayList` by `LinkedBlockingQueue`.
3. Others???

cowwoc:
…at runtime.

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/ProxyConfiguration.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/ProxyConfiguration.java
@@ -16,11 +16,13 @@ package org.eclipse.jetty.client;
 import java.net.URI;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
 
 import org.eclipse.jetty.http.HttpScheme;
 import org.eclipse.jetty.io.ClientConnectionFactory;
+import org.eclipse.jetty.util.BlockingArrayQueue;
 import org.eclipse.jetty.util.HostPort;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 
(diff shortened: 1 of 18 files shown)


cowwoc:
Moving from `ArrayList` to `CopyOnWriteList` adds thread-safety while retaining backwards compatibility.

cowwoc:
I'm not sure what's causing `continuous-integration` to fail. Can someone help?

jetty-client/src/main/java/org/eclipse/jetty/client/ProxyConfiguration.java
```diff
@@ -37,16 +37,45 @@
...
+     * Adds a proxy.
+     *
+     * @param proxy a proxy
+     * @throws NullPointerException if {@code proxy} is null
+     */
+    public void addProxy(Proxy proxy)
+    {
+        if (proxy == null)
+            throw new NullPointerException("proxy may not be null");
+        proxies.add(proxy);
```
sbordet:
```suggestion
        proxies.add(Objects.requireNonNull(proxy));
```

sbordet (changes requested):
Would you mind to update this PR by replacing the calls to `getProxies().add()` with the new `addProxy()` like you have done in 12? In this way the 2 codebases have less differences.

Also, I won't mind you fixing the `ProxyServletTest` here too (see my comment in the other PR).

cowwoc:
@sbordet How about this?

I couldn't update the v12 branch because the main branch has broken the build.

sbordet:
@cowwoc thanks!
This PR has been merged to Jetty 10, 11 and 12.

PR #8732, opened 2022-10-18
Jetty 11 - #8723 provide a thread safe way to modify proxies at runtime

cowwoc:


cowwoc:
Moving from `ArrayList` to `CopyOnWriteList` adds thread-safety while retaining backwards compatibility.

PR #8733, opened 2022-10-18
Jetty 12 - #8723: Provide a thread-safe way to modify HttpClient proxies

cowwoc:
…at runtime.

Unlike the v10, v11 implementations I opted to use `BlockingQueue` instead of `CopyOnWriteList`. This should improve performance.


jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/ProxyConfiguration.java
```diff
@@ -30,23 +32,42 @@
...
- * proxyConfig.getProxies().add(new HttpProxy(proxyHost, 8080));
+ * proxyConfig.addProxy(new HttpProxy(proxyHost, 8080));
  * </pre>
  *
  * @see HttpClient#getProxyConfiguration()
  */
 public class ProxyConfiguration
 {
-    private final List<Proxy> proxies = new ArrayList<>();
+    private final BlockingQueue<Proxy> proxies = new LinkedBlockingQueue<>();
```
sbordet, line 42:
```suggestion
    private final List<Proxy> proxies = new BlockingArrayQueue<>();
```

jetty-ee9/jetty-ee9-proxy/src/test/java/org/eclipse/jetty/ee9/proxy/ProxyServletTest.java
```diff
@@ -123,6 +126,7 @@ public static Stream<Arguments> impls()
     }
 
     private HttpClient client;
+    private final List<Proxy> clientProxies = new ArrayList<>();
```
sbordet: This field can be just a `Proxy proxy`.

jetty-ee9/jetty-ee9-proxy/src/test/java/org/eclipse/jetty/ee9/proxy/ProxyServletTest.java
```diff
@@ -197,16 +201,18 @@ private void startClient() throws Exception
 
     private void startClient(Consumer<HttpClient> consumer) throws Exception
     {
-        client = prepareClient(consumer);
+        clientProxies.add(new HttpProxy("localhost", proxyConnector.getLocalPort()));
+        client = prepareClient(consumer, clientProxies);
```
sbordet: Just initialize the `proxy` field, no need to pass it around, as `prepareClient()` will have access to it.

sbordet (changes requested):

sbordet:
@cowwoc thanks!
Resolved, squashed and merged with 67c36207c211a163b07187bbfc1e8c01696a8030.

PR #8734, opened 2022-10-18
Jetty 12 - Simplification of aliases in `PathResource` (Take 2)

joakime:
This is an update of PR #8631 to address the PathResource alias checking when it comes to bad resources.

PathResource cleanup

+ Adding comments about class fields.
+ Removing normalization from input/output/comparison flows.
+ Collapsing `resolveTargetPath` into `resolveAlias` to react accordingly to the exceptions that can flow out of `Path.toRealPath()`.
+ Failure on `Path.toRealPath()` is never an alias, as the resource cannot ever be served anyway.
+ More comments in `resolveAlias()`
+ Failed / Bad / Nonexistent / Inaccessible resources are not aliases to anything.

The changes in PR #8702 would impact these changes as we would never even have a Resource that is Failed / Bad / Nonexistent / Inaccessible resources anyway, rendering the extra checks in the `resolveAlias()` flows irrelevant for these kinds of bad resources (they are still relevant for good / accessible resources).


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -47,9 +47,15 @@ public class PathResource extends Resource
...
 
+    // The path object represented by this instance
     private final Path path;
+    // The as-requested URI for this path object
     private final URI uri;
-    private boolean targetResolved = false;
+    // True if this instance is an alias
+    private boolean alias = false;
+    // True to indicate that the alias resolution has been performed
+    private boolean aliasResolved = false;
```
gregw: @lachlan-roberts does this PR clash with yours?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -255,88 +277,61 @@ public boolean isContainedIn(Resource r)
...
+             * // On Windows
+             *  PathResource resource         = PathResource("C:/temp");
+             *  PathResource child            = resource.resolve("aa./foo.txt");
+             *  child.exists()                == true
+             *  child.isAlias()               == true
+             *  child.toUri()                 == "file:///C:/temp/aa./foo.txt"
+             *  child.getPath().toUri()       == "file:///C:/temp/aa/foo.txt"
+             *  child.getTargetURI()          == "file:///C:/temp/aa/foo.txt"
+             */
+            alias = !isSameName(path, targetPath) || !Objects.equals(uri, toUri(targetPath));
```
gregw: I like this. Similar to what I saw in @lachlan-roberts PR.  Makes it very clear what an alias is....  but I'm on fence if it is clearer as:

```suggestion
            alias = !(isSameName(path, targetPath) && Objects.equals(uri, toUri(targetPath)));
```

thoughts?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -373,6 +368,25 @@ private static URI toUri(Path path)
...
+    public boolean equals(Object obj)
+    {
+        if (this == obj)
+            return true;
+        if (obj == null)
+            return false;
+        if (getClass() != obj.getClass())
+            return false;
+        PathResource other = (PathResource)obj;
+        return Objects.equals(path, other.path) && Objects.equals(uri, other.uri);
```
gregw, line 454: Isn't it enough just to check uri?  When would the URI be the same but that path different?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -373,6 +368,25 @@ private static URI toUri(Path path)
...
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(path, uri);
```
gregw, line 460: Isn't it sufficient to hash the uri?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/Resource.java
```diff
@@ -297,19 +297,18 @@ public Resource resolve(String subUriPath)
...
     public boolean isAlias()
     {
-        return getTargetURI() != null;
+        return false;
     }
 
     /**
-     * If this Resource is an alias pointing to a different location,
-     * return the target location as URI.
+     * The target URI of the resource. If this Resource is an alias pointing to a different location,
```
gregw: Can we settle on "real URI" rather than target?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -47,9 +47,15 @@ public class PathResource extends Resource
...
     private final Path path;
+    // The as-requested URI for this path object
     private final URI uri;
-    private boolean targetResolved = false;
+    // True if this instance is an alias
+    private boolean alias = false;
+    // True to indicate that the alias resolution has been performed
+    private boolean aliasResolved = false;
+    // The targetPath of the alias, representing the real-path of this PathResource instance.
     private Path targetPath;
```
gregw:
```suggestion
    // The realPath of the resource, representing the un-aliased path of this PathResource instance.
    private Path realPath;
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -142,26 +148,34 @@ public static boolean isSameName(Path pathA, Path pathB)
 
     PathResource(URI uri, boolean bypassAllowedSchemeCheck)
     {
-        // normalize to referenced location, Paths.get() doesn't like "/bar/../foo/text.txt" style references
-        // and will return a Path that will not be found with `Files.exists()` or `Files.isDirectory()`
-        this(Paths.get(uri.normalize()), uri, bypassAllowedSchemeCheck);
+        this(Paths.get(uri), uri, bypassAllowedSchemeCheck);
```
gregw, line 149: Nice!!! no normalize in the constructor!!!

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -170,36 +184,43 @@ public static boolean isSameName(Path pathA, Path pathB)
...
+            // Avoid resolving alias unless necessary to determine existence.
+            if (Files.exists(path))
+                return true;
+            if (isAlias() && targetPath != null)
+                return Files.exists(targetPath);
         }
-        PathResource other = (PathResource)obj;
-        return Objects.equals(path, other.path);
+
+        return false;
```
gregw: don't need the else here as the previous block always returns.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
gregw: would a Boolean alias work better here?   null for unresolved and `Boolean.TRUE` or `Boolean.FALSE` once resolved.  That is correctly 3 states rather than using 3 of 4 possible states:
```suggestion
    // True/False if this instance is/isn't an alias. Null if alias resolution has not been performed.
    private Boolean alias;
```

gregw (changes requested):
Is this replacing @lachlan-roberts's #8631 ?  Oh I see the description says that. OK then...

a couple of niggles....

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: At first blush looks fine.
The bad / broken / non-existent / no-permission flows could be problematic until we merge PR #8702 .

Here's a few example scenarios.

path | targetPath | uri-as-requested | uri-from-target-path | alias
---- | ---- | ---- | ---- | ----
`C:/temp/aa./foo.txt` | `C:/temp/aa/foo.txt` | `file:///C:/temp/aa./foo.txt` | `file:///C:/temp/aa./foo.txt` | true
`/tmp/foo-symlink` | `/tmp/bar.txt` | `file:///tmp/foo-symlink` | `file:///tmp/bar.txt` | true
`C:/temp/aa.txt` | `C:/temp/AA.txt` | `file:///C:/temp/aa.txt` | `file:///C:/temp/AA.txt` | true
`/tmp/bar-exists/../foo.txt` | `/tmp/foo.txt` | `file:///tmp/bar-exists/../foo.txt` | `file:///tmp/foo.txt` | true
`/tmp/doesnt-exist.txt` | null (does not exist) | `file:///tmp/doesnt-exist.txt` | null (does not exist) | false
`/tmp/doesnt-exist/../foo.txt` | null (intermediate does not exist) | `file:///tmp/doesnt-exist/../foo.txt` | null (intermediate does not exist) | false
`/var/protected/config.xml` | null (no permissions) | `file:///var/protected/config.xml` | null (no permission) | false
`/tmp/foo-symlink` | null (broken symlink, doesn't point to anything) | `file:///tmp/foo-symlink` | null (broken symlink, doesn't point to anything) | false
`C:/temp/cannot:be:referenced` | null (illegal filename) | `file:///C:/temp/cannot:be:referenced` | null (illegal filename) | false

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: I've had this thought too.
I was even wondering why we are holding onto the original Path.
Couldn't the PathResource work with only have the "real-path" and "as-requested-uri"?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: Nor normalization in output of filenames, or normalization in the resolveAlias() either.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: Using `Boolean` could work. (I'll give it a go)

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
joakime: If the alias hasn't been resolved, and the original path (or target path) does not exist, then this is the final return.
This entire if/else block can be rewritten better.
Esp in light of the `Boolean` change.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -142,26 +148,34 @@
 
     PathResource(URI uri, boolean bypassAllowedSchemeCheck)
     {
-        // normalize to referenced location, Paths.get() doesn't like "/bar/../foo/text.txt" style references
-        // and will return a Path that will not be found with `Files.exists()` or `Files.isDirectory()`
-        this(Paths.get(uri.normalize()), uri, bypassAllowedSchemeCheck);
+        this(Paths.get(uri), uri, bypassAllowedSchemeCheck);
```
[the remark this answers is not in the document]
gregw: Is there an annotation or comment we can put into the code to turn off these warnings when we know we are taking uncontrolled data (which is pretty much all of the server)?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
gregw: punt to another PR

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
gregw: punt to another PR

lachlan-roberts (approved):

PR #8735, opened 2022-10-18
ResourceHandler set base resource as string

### LINKED ISSUE #8727: Jetty 12 - Improve usability for ResourceHandler
**Jetty version(s)**
12+

**Description**
For usage from XML, would be nice if `ResourceHandler.setBaseResource()` takes a `String` or a `Path`, because currently only takes a `Resource` which is cumbersome to create in XML.

gregw:
Added convenience method to set resource from a string. Closes #8727.

### DIFF
--- a/jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/ContextHandler.java
+++ b/jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/ContextHandler.java
@@ -1102,6 +1102,7 @@ public class ContextHandler extends ScopedHandler implements Attributes, Gracefu
     /**
      * @return Returns the resourceBase.
      */
+    @ManagedAttribute("document root for context")
     public Resource getBaseResource()
     {
         return _coreContextHandler.getBaseResource();
@@ -1109,8 +1110,9 @@ public class ContextHandler extends ScopedHandler implements Attributes, Gracefu
 
     /**
      * @return Returns the base resource as a string.
+     * @deprecated use #getBaseResource()
      */
-    @ManagedAttribute("document root for context")
+    @Deprecated
     public String getResourceBase()
     {
         Resource resourceBase = _coreContextHandler.getBaseResource();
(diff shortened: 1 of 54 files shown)


jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/ContextHandler.java
```diff
@@ -1102,15 +1102,17 @@ public void setDisplayName(String displayName)
...
     {
         return _coreContextHandler.getBaseResource();
     }
 
     /**
      * @return Returns the base resource as a string.
+     * @deprecated use #getBaseResource()
      */
-    @ManagedAttribute("document root for context")
+    @Deprecated
```
joakime, line 1115: Can't we just delete this method now?

joakime (approved):
Lots of mechanical changes.
Great cleanup of the Jetty XMLs. (even in the commented out projects/modules)
One odd deprecation. (IMO, just delete it)

I dismissed the CodeQL alert on Zip Slip (as that code is protected already)

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/ContextHandler.java
gregw: I think we should keep it in ee9 as we are trying to be as compatible as possible with jetty-10.
We can remove in 12.1.0

joakime: We already changed the contract from jetty-10 though.

`Resource.newResource(String)` no longer exists.
Have to use `ResourceFactory.newResource(String)` now.

gregw: yeah, there are changes, but it costs us nothing much to hold onto a deprecated method for at least a few release cycles.
We could remove it, but it can be a confusing name change, so leaving it as a sign post is a good idea.

PR #8736, opened 2022-10-18
Implement ErrorPageErrorProcessor to resolve #8726

### LINKED ISSUE #8726: Jetty 12 - Improve ErrorProcessor to handle error pages
**Jetty version(s)**
12+

**Description**
It would be nice if `ErrorProcessor` (or a subclass) can be configured with a mapping between the HTTP status code and an error page, something like:

```java
errorProcessor.addPage(404, "/404.html");
```

gregw:
Implement ErrorPageErrorProcessor to resolve #8726


gregw:
@sbordet something like this?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ErrorPageErrorProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+import org.eclipse.jetty.server.Response;
+import org.eclipse.jetty.util.Callback;
+import org.eclipse.jetty.util.URIUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * An ErrorProcessor that can re-handle a request at an error page location.
+ */
+public abstract class ErrorPageErrorProcessor extends ErrorProcessor
```
sbordet: This class does not deal with pages, so I'd call it `RedispatchingErrorProcessor` or `RehandlingErrorProcessor` or `DelegatingErrorProcessor` or `ForwardingErrorProcessor`, or similar.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ErrorPageErrorProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public InvocationType getInvocationType()
+    {
+        return _handler.getInvocationType();
+    }
+
+    @Override
+    protected void generateAcceptableResponse(Request request, Response response, int code, String message, Throwable cause, Callback callback) throws IOException
```
sbordet: Given this new implementation, I would rename this method of the base class, since it does not deal with the `Accept` header, to just `generateResponse()`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ErrorPageErrorProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+    {
+        String pathInContext = getErrorPathInContext(request, code, cause);
+
+        if (pathInContext != null)
+        {
+            HttpURI uri = HttpURI.build(request.getHttpURI()).path(URIUtil.addPaths(request.getContext().getContextPath(), pathInContext)).asImmutable();
+            Request.Wrapper wrapper = new Request.Wrapper(request)
```
sbordet: Maybe a properly named inner class to have good logging.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ErrorPageErrorProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+                Request.Processor processor = _handler.handle(wrapper);
+                if (processor != null)
+                {
+                    processor.process(wrapper, response, callback);
+                    return;
+                }
+            }
+            catch (Exception e)
+            {
+                LOG.warn("Unable to process error {}", wrapper, e);
```
sbordet: I'd stay at DEBUG level here, it's not something we want to pop-up in monitoring consoles.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ErrorPageErrorProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+            }
+        }
+
+        super.generateAcceptableResponse(request, response, code, message, cause, callback);
+    }
+
+    protected abstract String getErrorPathInContext(Request request, int code, Throwable cause);
```
sbordet: I'd rename this method accordingly to the rename of this class, so perhaps `getForwardingPath()`, or `getRehandlingPath()` or similar.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ErrorPageErrorProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * An ErrorPageErrorProcessor that uses a map of error codes to select a page.
+     */
+    public static class Mapped extends ErrorPageErrorProcessor
```
sbordet: Rename to `Pages`, or `Paths` so e..g in XML will become `ForwardingErrorProcessor.Paths.mapPath(404, "/404.html")`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ErrorPageErrorProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+    {
+        private final Map<Integer, String> _pageMap = new HashMap<>();
```
sbordet: I'd make it concurrent for safety.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ErrorPageErrorProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+            super(handler);
+        }
+
+        @Override
+        protected String getErrorPathInContext(Request request, int code, Throwable cause)
+        {
+            return _pageMap.get(code);
+        }
+
+        public String addPage(int code, String pathInContext)
```
sbordet: Rename to `mapPath()` and maybe provide also `unmapPath()` and `clearPaths()`?

sbordet (changes requested):

joakime:
The build is green, this can be non-draft now.

gregw:
> The build is green, this can be non-draft now.

I'll do the renames and then switch

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ReHandlingErrorProcessor.java
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * An ErrorProcessor that can re-handle a request at an error page location.
+ */
+public abstract class ReHandlingErrorProcessor extends ErrorProcessor
+{
+    private static final Logger LOG = LoggerFactory.getLogger(ReHandlingErrorProcessor.class);
+
+    private final Handler _handler;
+
+    protected ReHandlingErrorProcessor(Handler handler)
```
sbordet, line 39: I wonder whether we should be more strict here and only accept `Server`.

Otherwise this class can be configured with a `ContextHandler("/foo")`, but the rehandling code in `generateResponse()` rebuild the URI with the current request `contextPath` which may be different.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ReHandlingErrorProcessor.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    @Override
+    protected void generateResponse(Request request, Response response, int code, String message, Throwable cause, Callback callback) throws IOException
+    {
+        String pathInContext = getReHandlePathInContext(request, code, cause);
+
+        if (pathInContext != null)
+        {
+            HttpURI uri = HttpURI.build(request.getHttpURI()).path(URIUtil.addPaths(request.getContext().getContextPath(), pathInContext)).asImmutable();
```
sbordet: See comment above. If `_handler` has a different context, we are re-handling to a URI we don't expect?
If it results in a 404, we may end up in an infinite loop?

I mean the following:

request for: `/foo/notExist.html`
reHandlePathInContext is: `/404.html`
`_handler` is `ContextHandler("/bar")
`_handler.handle(reHandleRequest)` does not match the `contextPath`, which produces a 404, which loops.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ReHandlingErrorProcessor.java
```diff
@@ -0,0 +1,142 @@
...
+
+        super.generateResponse(request, response, code, message, cause, callback);
+    }
+
+    protected abstract String getReHandlePathInContext(Request request, int code, Throwable cause);
+
+    /**
+     * An ErrorPageErrorProcessor that uses a map of error codes to select a page.
+     */
+    public static class ByCode extends ReHandlingErrorProcessor
```
sbordet: Rename to `ByStatusCode` or `ByHttpStatusCode`?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ReHandlingErrorProcessor.java
```diff
@@ -0,0 +1,142 @@
...
+    {
+        private final Map<Integer, String> _pageMap = new ConcurrentHashMap<>();
```
sbordet: rename to `_codeMap` or `_pathMap`?

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/ErrorProcessorTest.java
```diff
@@ -689,6 +702,25 @@ public void process(Request request, Response response, Callback callback)
...
+                
+                """);
+
+        HttpTester.Response response = HttpTester.parseResponse(rawResponse);
+
+        assertThat(response.getStatus(), is(200));
+        assertThat(response.getField(HttpHeader.CONTENT_LENGTH).getIntValue(), greaterThan(0));
+        assertThat(response.getContent(), containsString("/ok/badMessage Error 400 : No Host"));
+    }
+
```
sbordet, line 723: Please write a test case for a possible infinite loop, where the re-handle generates the same error that it is trying to handle.

sbordet (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ReHandlingErrorProcessor.java
gregw: ContextHandlers can have their own ErrorProcessors, so in that case, we are only rewriting the pathInContext and redispatching back to the same context.
I guess there are also strange use-cases that might have errors handled in an entirely different context or maybe just a handler with no context at all.
We rewrite the HttpURI, so it is fine.   I'll add a context test case.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ReHandlingErrorProcessor.java
```diff
@@ -0,0 +1,146 @@
...
+                request.setAttribute(ReHandlingErrorProcessor.class.getName(), pathInContext);
+                HttpURI uri = HttpURI.build(request.getHttpURI()).path(URIUtil.addPaths(request.getContext().getContextPath(), pathInContext)).asImmutable();
+                Request.Wrapper wrapper = new ReHandleRequestWrapper(request, uri, uri.getCanonicalPath());
+
+                try
+                {
+                    Request.Processor processor = _handler.handle(wrapper);
+                    if (processor != null)
+                    {
+                        response.setStatus(200);
```
sbordet, line 68: Why setting the status? I think the processor we are invoking on the next line will do whatever it wants, so no point setting the status here.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ReHandlingErrorProcessor.java
```diff
@@ -0,0 +1,146 @@
...
+                        processor.process(wrapper, response, callback);
+                        return;
+                    }
+                }
+                catch (Exception e)
+                {
+                    if (LOG.isDebugEnabled())
+                        LOG.debug("Unable to process error {}", wrapper, e);
+                    if (cause != null && ExceptionUtil.areNotAssociated(cause, e))
+                        cause.addSuppressed(e);
```
sbordet, line 78: Not sure I want to chain this exception, but then call `super`. I would just log it and avoid the chaining.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ReHandlingErrorProcessor.java
```diff
@@ -0,0 +1,146 @@
...
+                    response.setStatus(code);
```
sbordet, line 79: Why setting the status here? We are calling `super` so it will do it.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ReHandlingErrorProcessor.java
```diff
@@ -0,0 +1,146 @@
...
+
+        public ByHttpStatus(Handler handler)
+        {
+            super(handler);
+        }
+
+        @Override
+        protected String getReHandlePathInContext(Request request, int code, Throwable cause)
+        {
+            return _statusMap.get(code);
```
sbordet: Call `get(code)` instead.

sbordet (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ReHandlingErrorProcessor.java
gregw: Ah it is no harm.  It is after all a suppressed exception.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ReHandlingErrorProcessor.java
gregw: Because we just dispatched to a handler with status code 200 trying to serve a page normally.  That failed, so we revert to the original error code.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ReHandlingErrorProcessor.java
gregw: Because we are dispatching to a normal page, perhaps even just a fixed html page, so we need an initial 200 status like any other dispatch.   This is NOT an error processor at this point, which would be generating an error response with an 4XX or 5XX code.  This is just serving a normal page that might say "you had a 4XX or 5XX error".

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ReHandlingErrorProcessor.java
gregw: @sbordet I added the context error test case.  OK?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ReHandlingErrorProcessor.java
gregw:
```suggestion
            return get(code);
```

sbordet (approved):

PR #8741, opened 2022-10-19
Jetty 12 - Remove start slash cleanup from `resolve(String)`

joakime:
Cleanup PathResource.resolve(String) a bit.
Remove the code that removes the starting slash, as it seems to not be needed.


joakime:
Turns out that use of `URIUtil.addPath(uri, subUriPath);` is taking care of this for us.

PR #8748, opened 2022-10-20
Jetty 12 - Introduce PathMappingsHandler

joakime:
New Handler allowing sub-handlers to be selected via a PathMappings configuration.

### DIFF
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
@@ -1,0 +1,60 @@
+//
+// ========================================================================
+// Copyright (c) 1995-2022 Mort Bay Consulting Pty Ltd and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.server.handler;
+
+import org.eclipse.jetty.http.pathmap.MatchedResource;
+import org.eclipse.jetty.http.pathmap.PathMappings;
+import org.eclipse.jetty.http.pathmap.PathSpec;
+import org.eclipse.jetty.server.Handler;
+import org.eclipse.jetty.server.Request;
+
+/**
+ * A Handler that delegates to other handlers through a configured {@link PathMappings}.
+ */
+public class PathMappingsHandler extends Handler.Wrapper
+{
+    private final PathMappings<Handler> mappings = new PathMappings<>();
+
+    public void addMapping(PathSpec pathSpec, Handler handler)
+    {
+        if (isStarted())
+            throw new IllegalStateException("Cannot add mapping: " + this);
+
+        mappings.put(pathSpec, handler);
+    }
+
+    @Override
+    protected void doStart() throws Exception
+    {
+        mappings.getMappings().forEach((mapped) -> addBean(mapped.getResource()));
+        super.doStart();
+    }
+
+    @Override
+    protected void doStop() throws Exception
+    {
+        super.doStop();
+        mappings.getMappings().forEach((mapped) -> removeBean(mapped.getResource()));
+    }
+
+    @Override
+    public Request.Processor handle(Request request) throws Exception
+    {
+        String pathInContext = request.getPathInContext();
+        MatchedResource<Handler> matchedResource = mappings.getMatched(pathInContext);
+        if (matchedResource == null)
+            return super.handle(request);
+        return matchedResource.getResource().handle(request);
+    }
+}
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
@@ -26,7 +26,8 @@ import org.slf4j.LoggerFactory;
  */
 public class PathMappingsHandler extends Handler.Wrapper
 {
-    private final static Logger LOG = LoggerFactory.getLogger(PathMappingsHandler.class);
+    private static final Logger LOG = LoggerFactory.getLogger(PathMappingsHandler.class);
+
     private final PathMappings<Handler> mappings = new PathMappings<>();
 
     public void addMapping(PathSpec pathSpec, Handler handler)
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
@@ -30,7 +30,7 @@ import org.slf4j.LoggerFactory;
  * A Handler that delegates to other handlers through a configured {@link PathMappings}.
  */
 
-public class PathMappingsHandler extends Handler.Wrapper
+public class PathMappingsHandler extends Handler.Abstract
 {
     private static final Logger LOG = LoggerFactory.getLogger(PathMappingsHandler.class);
 
@@ -83,7 +83,7 @@ public class PathMappingsHandler extends Handler.Wrapper
         {
             if  (LOG.isDebugEnabled())
                 LOG.debug("No match on pathInContext of {}", pathInContext);
-            return super.handle(request);
+            return null;
         }
         if (LOG.isDebugEnabled())
             LOG.debug("Matched pathInContext of {} to {} -> {}", pathInContext, matchedResource.getPathSpec(), matchedResource.getResource());
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
@@ -18,12 +18,15 @@ import org.eclipse.jetty.http.pathmap.PathMappings;
 import org.eclipse.jetty.http.pathmap.PathSpec;
 import org.eclipse.jetty.server.Handler;
 import org.eclipse.jetty.server.Request;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * A Handler that delegates to other handlers through a configured {@link PathMappings}.
  */
 public class PathMappingsHandler extends Handler.Wrapper
 {
+    private final static Logger LOG = LoggerFactory.getLogger(PathMappingsHandler.class);
     private final PathMappings<Handler> mappings = new PathMappings<>();
 
     public void addMapping(PathSpec pathSpec, Handler handler)
@@ -54,7 +57,13 @@ public class PathMappingsHandler extends Handler.Wrapper
         String pathInContext = request.getPathInContext();
         MatchedResource<Handler> matchedResource = mappings.getMatched(pathInContext);
         if (matchedResource == null)
+        {
+            if  (LOG.isDebugEnabled())
+                LOG.debug("No match on pathInContext of {}", pathInContext);
             return super.handle(request);
+        }
+        if (LOG.isDebugEnabled())
+            LOG.debug("Matched pathInContext of {} to {} -> {}", pathInContext, matchedResource.getPathSpec(), matchedResource.getResource());
         return matchedResource.getResource().handle(request);
     }
 }
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
@@ -14,8 +14,8 @@
 package org.eclipse.jetty.server.handler;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.List;
+import java.util.function.Supplier;
 
 import org.eclipse.jetty.http.pathmap.MappedResource;
 import org.eclipse.jetty.http.pathmap.MatchedResource;
@@ -23,7 +23,6 @@ import org.eclipse.jetty.http.pathmap.PathMappings;
 import org.eclipse.jetty.http.pathmap.PathSpec;
 import org.eclipse.jetty.server.Handler;
 import org.eclipse.jetty.server.Request;
-import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.util.component.Dumpable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public Request.Processor handle(Request request) throws Exception
+    {
+        String pathInContext = request.getPathInContext();
+        MatchedResource<Handler> matchedResource = mappings.getMatched(pathInContext);
+        if (matchedResource == null)
+            return super.handle(request);
+        return matchedResource.getResource().handle(request);
```
sbordet, line 101: I would add some DEBUG logging about what request came in, and whether there was a match or not.

sbordet (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
joakime: Done.

sbordet (approved):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
```diff
@@ -0,0 +1,69 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A Handler that delegates to other handlers through a configured {@link PathMappings}.
+ */
+public class PathMappingsHandler extends Handler.Wrapper
+{
+    private final static Logger LOG = LoggerFactory.getLogger(PathMappingsHandler.class);
+    private final PathMappings<Handler> mappings = new PathMappings<>();
```
sbordet, line 38: Separate static fields from instance fields with an empty line.

sbordet (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
joakime: Done.

joakime:
If we are ok with the idea of this, I'll write up some better javadoc for that class. (add some better tests for contexts other than `/` too)

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/PathMappingsHandlerTest.java
```diff
@@ -0,0 +1,242 @@
...
+     */
+    @Test
+    public void testOneMappingAndWrapper() throws Exception
+    {
+        ContextHandler contextHandler = new ContextHandler();
+        contextHandler.setContextPath("/");
+
+        PathMappingsHandler pathMappingsHandler = new PathMappingsHandler();
+        pathMappingsHandler.setHandler(new SimpleHandler("WrapperFoo Hit"));
+        pathMappingsHandler.addMapping(new ServletPathSpec("*.php"), new SimpleHandler("PhpExample Hit"));
```
joakime: Example 1 (of 2) that might be useful for FCGI (to PHP) Proxy.
This one relies on no-match behavior to fallback to the wrapper (which would be the ResourceHandler in the FCGI case)

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/PathMappingsHandlerTest.java
```diff
@@ -0,0 +1,242 @@
...
+    public void testSeveralMappingAndNoWrapper(String requestPath, int expectedStatus, String expectedResponseBody) throws Exception
+    {
+        ContextHandler contextHandler = new ContextHandler();
+        contextHandler.setContextPath("/");
+
+        PathMappingsHandler pathMappingsHandler = new PathMappingsHandler();
+        pathMappingsHandler.addMapping(new ServletPathSpec("/"), new SimpleHandler("FakeResourceHandler Hit"));
+        pathMappingsHandler.addMapping(new ServletPathSpec("/index.html"), new SimpleHandler("FakeSpecificStaticHandler Hit"));
+        pathMappingsHandler.addMapping(new ServletPathSpec("*.php"), new SimpleHandler("PhpHandler Hit"));
+        contextHandler.setHandler(pathMappingsHandler);
```
joakime, line 156: Example 2 (of 2) that might be useful for FCGI (to PHP) Proxy.
This one relies on default servlet pathspec behavior to use the resource handler for non-php requests.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
```diff
@@ -0,0 +1,70 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(PathMappingsHandler.class);
+
+    private final PathMappings<Handler> mappings = new PathMappings<>();
+
+    public void addMapping(PathSpec pathSpec, Handler handler)
+    {
+        if (isStarted())
+            throw new IllegalStateException("Cannot add mapping: " + this);
+
+        mappings.put(pathSpec, handler);
```
gregw: The mapped handlers should still be added to this handler as a bean, so we can dump them even in not started.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
```diff
@@ -0,0 +1,70 @@
...
+{
+    private static final Logger LOG = LoggerFactory.getLogger(PathMappingsHandler.class);
+
+    private final PathMappings<Handler> mappings = new PathMappings<>();
+
+    public void addMapping(PathSpec pathSpec, Handler handler)
+    {
+        if (isStarted())
+            throw new IllegalStateException("Cannot add mapping: " + this);
+
```
gregw, line 45: maybe some loop detection would be good?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
```diff
@@ -0,0 +1,70 @@
...
+        mappings.put(pathSpec, handler);
+    }
+
+    @Override
+    protected void doStart() throws Exception
+    {
+        mappings.getMappings().forEach((mapped) -> addBean(mapped.getResource()));
```
gregw: It is too late to add beans here. We want a dump to be meaningful even for a stopped or failed server.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
```diff
@@ -0,0 +1,70 @@
...
+        super.doStart();
+    }
+
+    @Override
+    protected void doStop() throws Exception
+    {
+        super.doStop();
+        mappings.getMappings().forEach((mapped) -> removeBean(mapped.getResource()));
```
gregw: Keep beans up to date as they are added/removed

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
```diff
@@ -0,0 +1,70 @@
...
+import org.eclipse.jetty.http.pathmap.PathSpec;
+import org.eclipse.jetty.server.Handler;
+import org.eclipse.jetty.server.Request;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A Handler that delegates to other handlers through a configured {@link PathMappings}.
+ */
+public class PathMappingsHandler extends Handler.Wrapper
```
gregw: I don't think this is a `Handler.Wrapper`, rather it is a `Handler.Container`

gregw: Hmmm I guess it can be both?

gregw: Eitherway, it needs to respect the `getHandlers()` and `getDescendants()` contracts.  These might be simpler to consider if this was just a `Container` and not a `Wrapper`.  You can always use the "/" or "/*" mappings to get a default mapping.

gregw (changes requested):
This is not a bad idea, but needs some work.

Maybe there should be a common ancestor of this and `ContextHandlerCollection` that handles the whole `Handler.Container` conversation about descendants and bean management.  Then some abstract methods can be used to make it either a `ContextHandlerCollection` or a `PathMapHandlerCollection`?

Maybe `Handler.Collection` is already that common ancestor?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
gregw: Yeah, I think this should extend `Handler.Collection` and be named `PathMappedHandlerCollection`

gregw:
See also the vaguely related #8676

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
joakime: What about making PathMappings itself be lifecycle-able (for the start/stop of the individual Handlers)?
After all PathMappings is already Dumpable.

joakime: What about making PathMappings the dump point, and only having doStart/doStop handle the start/stop of the handlers that the PathMappings is tracking? (not as individual beans)

gregw:
Perhaps we should look at the feedback I've given in this PR and put it into a "How to write a Handler" guide documentation?   Calling all tech writers!!!!

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
joakime: What kind of loop detection would fit here?
Got a scenario?

Or do you mean a Handler from earlier in the tree being mapped under here and possibly being a loop?
```
 server.handler
   MyHandler@0xBEEF
      PathMappingHandler
         mapping("/*", MyHandler@0xBEEF)
```

If that's the case, do we have a generic version of this kind of loop detection?
As it could be useful in many scenarios (even with just the existing Handler.Collection / Handler.Wrapper impls)

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
sbordet: @gregw `Handler.Collection` has a TODO to be renamed `Handler.List`. 

I think this new class should extend `Handler.AbstractContainer` because it's not a list nor a collection (it's more like a map) and just be called `PathMappingsHandler` or perhaps `PathMappingsHandlerContainer`, but nothing related to `List` or `Collection`.

sbordet (changes requested):

gregw:
+1 for extending `AbstractContainer` 
How about `Handler.Sequence` rather than `Handler.Collection` or `Handler.List` ?

joakime:
Not a fan of `Sequence` here.

Mainly because the word "Sequence" here has meaning that the PathMappingsHandler isn't doing.
Also the next JDK is supposed to actually implement the "Sequenced Collections" JEP.

https://openjdk.org/jeps/431

gregw:
> Not a fan of `Sequence` here.
> 
> Mainly because the word "Sequence" here has meaning that the PathMappingsHandler isn't doing. Also the next JDK is supposed to actually implement the "Sequenced Collections" JEP.
> 

But `PathMappingsHandler` would not be a `Handler.Sequence`.  The suggestion is for it to extend `Handler.AbstractContainer`.
The existing `Handler.Collection` would separately be renamed to `Handler.Sequence`, because it does have sequential behavior.   We know a JVM Sequence is coming, but it will be at least initially less common than `List` and `Collection`.   But open to other names that totally avoid clashing with a JVM class.... but I can't think of any good ones.   Maybe "Serial"?

joakime:
Extending from `Handler.AbstractContainer` doesn't work, as that would require implementing the following methods (that have no meaning under a `PathMappingsHandler`)

``` java
public void addHandler(Handler handler);
public void addHandler(Supplier<Handler> supplier);
public List<Handler> getHandlers();
public List<Handler> getDescendants();
public <T extends Container> T getContainer(Handler handler, Class<T> type);
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
joakime: Done (differently)

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
joakime: Done (differently)

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
joakime: Still don't understand this.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
joakime: Done.  Dump of PathMappings is now sane, regardless of start/stop state.

joakime:
Also, why does `Handler.Wrapper` extend `Handler.AbstractContainer` ?

gregw:
> Extending from `Handler.AbstractContainer` doesn't work, as that would require implementing the following methods (that have no meaning under a `PathMappingsHandler`)
> 
> ```java
> public void addHandler(Handler handler);
> public void addHandler(Supplier<Handler> supplier);
> public List<Handler> getHandlers();
> public List<Handler> getDescendants();
> public <T extends Container> T getContainer(Handler handler, Class<T> type);
> ```

They can all have reasonable meanings for your handler:
 + addHandler could add a handler at the default mapping.... or it could be unsupported operation exception
 + ditto for the add supplier
 + getHandlers certainly has a meaning, it returns all the nested handlers - ie the maps value set
 + ditto for descendants, it returns all the descendant handlers of all contained handlers. Default impl will work fine
 + ditto for getContainer, default impl will work fine

Handler.Wrapper is a Container because it is a `Handler` that contains other `Handler`s.  See how it extends `Handler.Nested` and has a meaningful interpretation of `addHandler`   

See `ContextHandlerCollection` (although maybe that really should be `ContextMapHandler extends AbstractContainer`.. but that's another PR and some history... plus it does iterate if there is no obvious match).   It's job is slightly easier for addHandler, as contexts carry their context paths with them. 

Having all this stuff correct is important, so we can dump, find, manipulate handlers in a consistent way.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
joakime: Done

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
```diff
@@ -0,0 +1,103 @@
...
+        {
+            Handler entryHandler = entry.getResource();
+
+            if (entryHandler == this ||
+                entryHandler == handler ||
+                (entryHandler instanceof Handler.Container container && container.getDescendants().contains(this)))
+                throw new IllegalStateException("addMapping loop detected: " + handler);
+        }
+
+        mappings.put(pathSpec, handler);
```
gregw, line 78: Can this return an old handler that was replaced?  If so, then that needs to be removed as a bean and probably returned from this method?

gregw (approved):
I've got one niggle, but will approve as I'm out of office for rest of today.   
Niggle is make sure you remove the bean of any replaced handlers.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/PathMappingsHandler.java
joakime: There's no replacement, the mappings are a `Set` or `Index`.

joakime: Opened Issue #8764 to address this later

PR #8751, opened 2022-10-20
Jetty 12 - Nested / Deep ContextHandler support

### LINKED ISSUE #8749: Jetty 12 - `request.getContext().getContextPath()` should return full context, not just last one
**Jetty version(s)**
12

**Description**
The `request.getContext().getContextPath()` only returns the last context seen, not the full context path up to this point.

``` java
    @Test
    public void testNestedThreeDeepContextHandler() throws Exception
    {
        ContextHandler contextHandlerA = new ContextHandler();
        contextHandlerA.setContextPath("/a");
        ContextHandler contextHandlerB = new ContextHandler();
        contextHandlerB.setContextPath("/b");
        ContextHandler contextHandlerC = new ContextHandler();
        contextHandlerC.setContextPath("/c");

        contextHandlerA.setHandler(contextHandlerB);
        contextHandlerB.setHandler(contextHandlerC);
        contextHandlerC.setHandler(new Handler.Processor()
        {
            @Override
            public void process(Request request, Response response, Callback callback)
            {
                response.getHeaders().put(HttpHeader.CONTENT_TYPE, "text/plain; charset=utf-8");
                String msg = "contextPath=" + request.getContext().getContextPath();
                response.write(true, BufferUtil.toBuffer(msg), callback);
            }
        });

        startServer(contextHandlerA);

        String rawRequest = """
            GET /a/b/c/d HTTP/1.1\r
            Host: local\r
            Connection: close\r
                        
            """;
        HttpTester.Response response = HttpTester.parseResponse(connector.getResponse(rawRequest));
        assertEquals(HttpStatus.OK_200, response.getStatus());
        assertEquals("contextPath=/a/b/c", response.getContent());
    }
```

In this scenario, where we have 3 `ContextHandler` nested inside of each other, the final Handler only sees the last contextPath.
(The assertion fails, as it expected `/a/b/c` but only got `/c`)

This makes it hard to rewrite / redirect / change the URI without the knowledge of the full context.
(for example, putting a `RewriteHandler` after a `ContextHandler`)

joakime:
This is just a place to experiment with Nested / Deep ContextHandler being discussed in Issue #8749

### DIFF
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Context.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Context.java
@@ -35,7 +35,10 @@ import org.eclipse.jetty.util.resource.Resource;
 public interface Context extends Attributes, Decorator, Executor
 {
     /**
+     * <p>Get the context path of this Context.</p>
+     * <p>This may differ from the {@link Request#getContextPath()} in the uncommon case of nested contexts.</p>
      * @return The URI path prefix of the context, which may be null for the server context, or "/" for the root context.
+     * @see Request#getContextPath()
      */
     String getContextPath();
 
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Context.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Context.java
@@ -36,7 +36,7 @@ public interface Context extends Attributes, Decorator, Executor
 {
     /**
      * <p>Get the context path of this Context.</p>
-     * <p>This may differ from the {@link Request#getContextPath()} in the uncommon case of nested contexts.</p>
+     * <p>This may differ from the {@link Request#getContextPath()} in the case of nested contexts.</p>
      * @return The URI path prefix of the context, which may be null for the server context, or "/" for the root context.
      * @see Request#getContextPath()
      */
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
@@ -150,6 +150,12 @@ public interface Request extends Attributes, Content.Source
      */
     Context getContext();
 
+    default String getContextPath()
+    {
+        Context context = getContext();
+        return context == null ? null : context.getContextPath();
+    }
+
     /**
      * TODO see discussion in #7713, as this path should probably be canonically encoded - ie everything but %25 and %2F decoded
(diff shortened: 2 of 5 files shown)


jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java
```diff
@@ -712,7 +712,7 @@ public DateField(long seconds, HttpField dateField)
 
     private static class DynamicErrorProcessor extends ErrorProcessor {}
 
-    private class ServerContext extends Attributes.Wrapper implements Context
+    public class ServerContext extends Attributes.Wrapper implements Context
```
joakime: I made this public to make the new code work, we should review this.
Perhaps package-private is sufficient?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Context.java
```diff
@@ -35,7 +35,10 @@
 public interface Context extends Attributes, Decorator, Executor
 {
     /**
+     * <p>Get the context path of this Context.</p>
+     * <p>This may differ from the {@link Request#getContextPath()} in the uncommon case of nested contexts.</p>
```
joakime: I think you can drop "uncommon" from this and it still be valid.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java
joakime: Yeah, package-private is enough.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -151,8 +151,27 @@ public interface Request extends Attributes, Content.Source
     Context getContext();
 
     /**
-     * TODO see discussion in #7713, as this path should probably be canonically encoded - ie everything but %25 and %2F decoded
-     * @return The part of the decoded path of the URI after any context path prefix has been removed.
+     * Get the full context path.
+     * @return The full contextPath of the request. In the common case, this is equivalent to
```
gregw, line 155: maybe reword to get rid of "common" here as well:
```suggestion
     * @return The full contextPath of the request. Unless using nested contexts, this is equivalent to
```

gregw:
This wasn't meant to be merged. It was replaced by #8758

sbordet:
@gregw I did not merge it, not sure what happened, looks it was done automatically when I merged #8758.
Did you base #8758 on top of this one?

Issue #8753, opened 2022-10-21
Starting HttpClient with destinationIdleTimeout set throws NPE.


janvojt:
**Jetty version(s)**
\>= 11.0.x

**Java version/vendor** `(use: java -version)`
any

**OS type/version**
any

**Description**
If `HttpClient` is initialized with a default (implicit) scheduler and at the same time `destinationIdleTimeout` is set, then `HttpClient` cannot start and throws `NullPointerException`.

**How to reproduce?**

```java
    @Test
    public void testDestinationIdleTimeout() throws Exception {
        HttpClient client = new HttpClient();
        client.setDestinationIdleTimeout(1000);
        client.start();
    }
```

Throws:
```
java.lang.NullPointerException: Cannot invoke "org.eclipse.jetty.util.thread.Scheduler.schedule(java.lang.Runnable, long, java.util.concurrent.TimeUnit)" because "this.scheduler" is null

	at org.eclipse.jetty.util@11.0.13-SNAPSHOT/org.eclipse.jetty.util.thread.Sweeper.activate(Sweeper.java:151)
	at org.eclipse.jetty.util@11.0.13-SNAPSHOT/org.eclipse.jetty.util.thread.Sweeper.doStart(Sweeper.java:88)
	at org.eclipse.jetty.util@11.0.13-SNAPSHOT/org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:93)
	at org.eclipse.jetty.client/org.eclipse.jetty.client.HttpClient.doStart(HttpClient.java:233)
	at org.eclipse.jetty.util@11.0.13-SNAPSHOT/org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:93)
	at org.eclipse.jetty.client/org.eclipse.jetty.client.HttpClientTest.testDestinationIdleTimeout(HttpClientTest.java:2005)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
	at java.base/java.lang.reflect.Method.invoke(Method.java:578)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:727)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
[...]

PR #8754, opened 2022-10-21
Issue #8753 - Fix HttpClient initialization with destinationIdleTimeout set.

janvojt:
The fix is also needed in version 12.

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
@@ -195,6 +195,7 @@ public class HttpClient extends ContainerLifeCycle
         {
             QueuedThreadPool threadPool = new QueuedThreadPool();
             threadPool.setName(name);
+            executor = threadPool;
             setExecutor(threadPool);
         }
         int maxBucketSize = executor instanceof ThreadPool.SizedThreadPool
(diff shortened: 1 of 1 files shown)


sbordet:
@janvojt please issue PRs against branch `jetty-10.0.x`. We will take care of merging to Jetty 11 and 12.

Also, please comply with https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/CONTRIBUTING.md, otherwise we cannot accept your PR.

sbordet:
@janvojt I would not mind if you could fix a similar issue with local variable `executor` few lines above the ones you modified for the `scheduler` local variable.

janvojt:
@sbordet The issue is not present in `jetty-10.0.x`, that is why I request merging into `jetty-11.0.x`.

I read `CONTRIBUTING.md`, signed-off commit, created new Eclipse account, agreed with ECA, revalidated ECA sign-off, but the check still failed. My ECA form has the following, so I suppose it is valid:
```
The Eclipse Contributor Agreement 3.1.0 that you signed is valid for a period of three years from date of signing and is set to expire on 2025-10-21.
```

I fixed the `executor` as well.

joakime:
> @sbordet The issue is not present in `jetty-10.0.x`, that is why I request merging into `jetty-11.0.x`.

Considering that we only really work in Jetty 10.0.x, and them merge to 11.0.x, this would be surprising!

janvojt:
I'm sorry, you are correct. I did not update my v10 branch. I am closing this and will create a new PR.

> > @sbordet The issue is not present in `jetty-10.0.x`, that is why I request merging into `jetty-11.0.x`.
> 
> Considering that we only really work in Jetty 10.0.x, and them merge to 11.0.x, this would be surprising!

PR #8762, opened 2022-10-25
Experiment with lock free AdaptiveExecutionStrategy

gregw:
Removed the lock from the AdaptiveExecutionStrategy in favor of a BiInteger.

@sbordet pointed out that the AES lock is the most contended in the server when running cometd benchmarks.  He pointed out that it could be implemented without a lock, so I've done that in this PR. @lorban can you benchmark this branch to see if there is any significant benefit?  I'm hopeful as many paths through the CaS loops don't look at all the state and some don't need to change state.

The other thing to consider, is if the pending mechanism is needed at all? Perhaps the reserved thread pool is sufficient to stop stampedes?  But as it is, pending handling is pretty simple... once you realize that it is now a counter rather than a boolean and it may go > 1 (two threads both start a pending at the same time) or < 0 (a started pending producer runs before the pending count is actually incremented).

### DIFF
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
@@ -20,13 +20,13 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.atomic.LongAdder;
 
+import org.eclipse.jetty.util.AtomicBiInteger;
 import org.eclipse.jetty.util.IO;
 import org.eclipse.jetty.util.VirtualThreads;
 import org.eclipse.jetty.util.annotation.ManagedAttribute;
 import org.eclipse.jetty.util.annotation.ManagedObject;
 import org.eclipse.jetty.util.annotation.ManagedOperation;
 import org.eclipse.jetty.util.component.ContainerLifeCycle;
-import org.eclipse.jetty.util.thread.AutoLock;
 import org.eclipse.jetty.util.thread.ExecutionStrategy;
 import org.eclipse.jetty.util.thread.Invocable;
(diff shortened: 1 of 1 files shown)


gregw:
@lorban an ETA for some benchmarks on this?

lorban:
@gregw I've updated [jetty-perf](https://github.com/jetty-project/jetty-perf) and [jetty-load-generator](https://github.com/jetty-project/jetty-load-generator) to work with the latest 12.0.x changes.

You can expect some results in the coming hours.

lorban:
I expected this change to have no measurable impact in the HTTP case, and I was right: you can like this change for code clarity but performance wise, this is equivalent to the locked version.

What I did not consider and surprised me is that this makes a measurable impact in the H2 case: the p99 latency goes down by 30% at the expense of a 25% higher max latency. And this made https://github.com/eclipse/jetty.project/issues/8810 surface.

Overall, I think this is a good change that brings meaningful improvement to H2.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -98,12 +98,9 @@ public class AdaptiveExecutionStrategy extends ContainerLifeCycle implements Exe
     /**
      * The production state of the strategy.
      */
-    private enum State
-    {
-        IDLE,       // No tasks or producers.
-        PRODUCING,  // There is an active producing thread.
-        REPRODUCING // There is an active producing thread and demand for more production.
-    }
+    static final int IDLE = 0;        // No tasks or producers.
```
lorban, line 101: These constants should be private.

lorban (approved):
LGTM, minus one small nit.

gregw:
@lorban @sbordet worth back porting this?

lorban:
@gregw I'd say yes since it's a rather small and contained change.

sbordet:
@gregw yes please backport to 10.

PR #8767, opened 2022-10-26
Refactor and improvements to HttpContent Factories

lachlan-roberts:
- flatten cache in HttpCachingHttpContentFactory
- remove old CachedContentFactory
- HttpContent uses RetainableByteBuffer
- use buffer pool in CachingHttpContentFactory
- add configuration to reduce/prevent file system accesses for cached HttpContent
- add separate HttpContent factory for file mapping


gregw:
@lachlan-roberts are those test failures flakes?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
```diff
@@ -0,0 +1,540 @@
...
+import org.eclipse.jetty.util.StringUtil;
+import org.eclipse.jetty.util.resource.Resource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * {@link HttpContent.Factory} implementation that wraps any other {@link HttpContent.Factory} instance
+ * using it as a caching authority.
+ * Only HttpContent instances whose path is not a directory are cached.
+ */
```
gregw, line 56: You need to document shrink and evict policy (even if delegated to another class)

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
```diff
@@ -0,0 +1,540 @@
...
+
+        if (_maxCachedFiles <= 0)
+            return false;
+
+        // Will it fit in the cache?
+        long len = httpContent.getContentLengthValue();
+        if (len <= 0)
+            return false;
+        if (httpContent instanceof FileMappedHttpContentFactory.FileMappedContent)
+             return true;
```
gregw: This is very specific.   Perhaps a tag interface would be better?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
```diff
@@ -0,0 +1,540 @@
...
+    public void setMaxCachedFiles(int maxCachedFiles)
+    {
+        _maxCachedFiles = maxCachedFiles;
+        shrinkCache();
+    }
+
+    private void shrinkCache()
+    {
+        // While we need to shrink
+        while (_cache.size() > 0 && (_cache.size() > _maxCachedFiles || _cachedSize.get() > _maxCacheSize))
```
gregw: get `_cache.size()` only once and assign to variable.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
```diff
@@ -0,0 +1,540 @@
...
+            AtomicBoolean wasAdded = new AtomicBoolean(false);
+            cachingHttpContent = _cache.computeIfAbsent(path, p ->
+            {
+                wasAdded.set(true);
+                CachingHttpContent cachingContent = (httpContent == null) ? newNotFoundContent(p) : newCachedContent(p, httpContent);
+                _cachedSize.addAndGet(cachingContent.getContentLengthValue());
+                return cachingContent;
+            });
+
+            if (wasAdded.get())
```
gregw: Is it worthwhile having the `wasAdded` atomic?  Why not just call `shrinkCache` anyway and that will do a volitile lookup of the size, which is the same cost.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
```diff
@@ -0,0 +1,540 @@
...
+            // TODO: do all the following lazily and asynchronously.
+            HttpField etagField = httpContent.getETag();
+            String eTagValue = httpContent.getETagValue();
+            if (StringUtil.isNotBlank(eTagValue))
+            {
+                etagField = new PreEncodedHttpField(HttpHeader.ETAG, eTagValue);
+            }
+            _etagField = etagField;
+
+            // Map the content into memory if possible.
```
gregw:
```suggestion
            // Read the content into memory if possible.
```

What if the buffer was a `FileMappedHttpContentFactory.FileMappedContent`, wont you also read it?

joakime:
The `TryPathsHandlerTest` failures seem to indicate a change in either Request.Processor behavior, or ResourceHandler behavior.
Not a flaky test.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
```diff
@@ -0,0 +1,540 @@
...
+        {
+            return null;
+        }
+
+        @Override
+        public String getCharacterEncoding()
+        {
+            return null;
+        }
+
```
gregw, line 531: I think these should all be default impls on the interface.

gregw: and most of the other methods in this class

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/EvictingCachingContentFactory.java
```diff
@@ -0,0 +1,165 @@
...
+ * A value of -1 means that the cached {@link HttpContent} will never be checked if it is still valid,
+ * a value of 0 means it is checked on every request,
+ * a positive value indicates the number of milliseconds of the minimum time between validation checks.
+ * </p>
+ * <p>
+ * This also remember a missed entry for the time set by {@code validationTime}ms. After this has
+ * elapsed the entry will be invalid and will be evicted from the cache at the next access.
+ * </p>
+ */
+public class EvictingCachingContentFactory extends CachingHttpContentFactory implements Runnable
```
gregw: The naming is not exactly correct because it is not this class that does the evicting. That is done by `CachingHttpContentFactory` if the content is not valid.    Perhaps `ValidatingCachingContentFactory`?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/EvictingCachingContentFactory.java
```diff
@@ -0,0 +1,165 @@
...
+
+    public EvictingCachingContentFactory(@Name("authority") HttpContent.Factory authority,
+                                         @Name("validationTime") long validationTime)
+    {
+        this(authority, validationTime, null, -1, -1);
+    }
+
+    public EvictingCachingContentFactory(@Name("authority") HttpContent.Factory authority,
+                                         @Name("validationTime") long validationTime,
+                                         @Name("scheduler") Scheduler scheduler,
```
gregw: need to better document the optional sweeper

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/EvictingCachingContentFactory.java
```diff
@@ -0,0 +1,165 @@
...
+        ConcurrentMap<String, CachingHttpContent> cache = getCache();
+        for (Map.Entry<String, CachingHttpContent> entry : cache.entrySet())
+        {
+            CachingHttpContent value = entry.getValue();
+            if (_maxCacheIdleTime > 0 && NanoTime.since(value.getLastAccessedNanos()) > TimeUnit.MILLISECONDS.toNanos(_maxCacheIdleTime))
+                removeFromCache(value);
+            else if (!value.isValid())
+                removeFromCache(value);
+        }
+        schedule();
```
gregw: maybe put in a try-finally... just in case.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
```diff
@@ -0,0 +1,72 @@
...
+                ByteBuffer buffer = BufferUtil.toMappedBuffer(content.getResource().getPath());
+                return new RetainableByteBuffer(buffer, b -> {});
+            }
+            catch (Throwable t)
+            {
+                if (LOG.isDebugEnabled())
+                    LOG.debug("Error getting Mapped Buffer", t);
+            }
+
+            return super.getBuffer();
```
gregw: This feels wrong.  If the resource cannot be mapped, then we should not create a `FileMappedContent`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
```diff
@@ -0,0 +1,72 @@
...
+            super(content);
+            this.content = content;
+        }
+
+        @Override
+        public RetainableByteBuffer getBuffer()
+        {
+            try
+            {
+                ByteBuffer buffer = BufferUtil.toMappedBuffer(content.getResource().getPath());
```
gregw: Isn't this re-mapping the buffer on every call to getBuffer()?  Surely it should be mapped already and just sliced?

gregw (dismissed):
Only a partial review so far.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
joakime: Here's some cases where BufferUtil.toMappedBuffer(Path) will not work.

* Extremely large files (I'm talking huge, like 200GB+)
* Real OS FileSystems based on dynamic content. (filesystems that have content that doesn't exist until it's read from, and can be different with each read, many such things exist on Windows)
* Emulated OS FileSystems. (eg: a Windows machine, in the WSL layer, reading a windows drive via the WSL layer)
* Some OS based Network FileSystems.
* Java FileSystems based on virtual concepts (eg:zipfs)

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
joakime: Here's some cases where BufferUtil.toMappedBuffer(Path) will not work.

* Extremely large files (I'm talking huge, like 200GB+)
* Real OS FileSystems based on dynamic content. (filesystems that have content that doesn't exist until it's read from, and can be different with each read, many such things exist on Windows)
* Emulated OS FileSystems. (eg: a Windows machine, in the WSL layer, reading a windows drive via the WSL layer)
* Some OS based Network FileSystems.
* Java FileSystems based on virtual concepts (eg:zipfs)

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
lachlan-roberts: Done.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
lachlan-roberts: Not sure what you mean by this.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
lachlan-roberts: Ok sure, but will also need to reassign for each iteration of the loop.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
lachlan-roberts: `shrinkCache` will have to do 2 volatile lookups, `_cache.size()` and `_cachedSize.get()`.

Ideally we wouldn't need the AtomicBoolean here, it is just because we can't know if `computeIfAbsent` created a new entry or not.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
lachlan-roberts: Yes, but I don't think we should be reading it. The check in `CachingContentFactory.isCacheable`
```java
if (httpContent instanceof FileMappedHttpContentFactory.FileMappedContent)
             return true;
```
indicates that we don't count these towards the size so we shouldn't be copying it into indirect buffer.

I will change this to 
```java
// Read the content into memory if the HttpContent does not already have a buffer.
RetainableByteBuffer buffer = httpContent.getBuffer();
if (buffer != null)
{
    ...
```

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
lachlan-roberts: Why? this is the only implementation which ever returns null and it is only an internal implementation which is never returned by `getContent()`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/EvictingCachingContentFactory.java
lachlan-roberts: Done.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/EvictingCachingContentFactory.java
lachlan-roberts: Done.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/EvictingCachingContentFactory.java
lachlan-roberts: Done.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
lachlan-roberts: Is there a test we can do to see if it is mappable? If so this should be done in `getContent()`.

Or I could try and call `BufferUtil.toMappedBuffer` in `getContent()` and only wrap if this does not throw. But if the file changes on disk it could become unmappable even after this `FileMappedContent` has been created.

If we cache a mapped buffer the size of the buffer could change if the file is changed on disk, so should we not cache any `FileMappedContent`.

lachlan-roberts:
@gregw @joakime I assumed `TryPathsHandlerTest` was flakey because it was not failing in this current branch, but looks like there are some new changes in `12.0.x` which cause it to fail with this PR.

I will investigate.

lachlan-roberts:
It was failing because the `ResourceHandler` in the `TryPathsHandlerTest` was caching the "not found" entries. So when they were created they were still reported as not found.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
joakime: I think the easiest way, in a testcase, is to have a zipfs (mounted jar) and try to use BufferUtil.toMappedBuffer(Path), which should fail in 100% of use cases.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
```diff
@@ -0,0 +1,72 @@
...
+    public FileMappedHttpContentFactory(HttpContent.Factory factory)
+    {
+        _factory = Objects.requireNonNull(factory);
+    }
+
+    @Override
+    public HttpContent getContent(String path) throws IOException
+    {
+        HttpContent content = _factory.getContent(path);
+        if (content != null && content.getContentLengthValue() > 16 * 1024)
```
lachlan-roberts: This `16 * 1024` value should be configurable.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
lachlan-roberts: Add method on `HttpContent` called `boolean getBufferOccupancy()`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
gregw: Because:
 + in future there may be other implementations.
 + this is an impl that actually does something, so best to not crowd the code with boiler plate.  Having the null impl in the interface is a better place that crowding this file.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
```diff
@@ -0,0 +1,110 @@
...
+public class FileMappedHttpContentFactory implements HttpContent.Factory
+{
+    private static final Logger LOG = LoggerFactory.getLogger(FileMappedHttpContentFactory.class);
+
+    private final HttpContent.Factory _factory;
+    private final int _minFileSize;
+
+    public FileMappedHttpContentFactory(HttpContent.Factory factory)
+    {
+        this(factory, 16 * 1024);
```
lorban: This constant should be javadoc'ed.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public HttpContent getContent(String path) throws IOException
+    {
+        HttpContent content = _factory.getContent(path);
+        if (content != null && content.getContentLengthValue() > _minFileSize)
+            return new FileMappedContent(content);
+        return content;
+    }
+
+    public static class FileMappedContent extends HttpContentWrapper
```
lorban: This should be `private` or maybe package-protected.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
```diff
@@ -38,7 +38,7 @@ public class RetainableByteBuffer extends Retainable.ReferenceCounter
     private final Consumer<RetainableByteBuffer> releaser;
     private final AtomicLong lastUpdate = new AtomicLong(NanoTime.now());
 
-    RetainableByteBuffer(ByteBuffer buffer, Consumer<RetainableByteBuffer> releaser)
+    public RetainableByteBuffer(ByteBuffer buffer, Consumer<RetainableByteBuffer> releaser)
```
lorban: It doesn't look like that is still needed.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/TryPathsHandlerTest.java
```diff
@@ -94,7 +96,16 @@ public void dispose()
     @Test
     public void testTryPaths() throws Exception
     {
-        ResourceHandler resourceHandler = new ResourceHandler();
+        ResourceHandler resourceHandler = new ResourceHandler()
+        {
+            @Override
+            protected HttpContent.Factory setupContentFactory()
+            {
+                // We don't want to cache not found entries for this test.
```
lorban, line 104: There should be a more straightforward way to achieve this. Perhaps a config setting?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
```diff
@@ -595,6 +595,21 @@ public static void readFrom(Path path, ByteBuffer buffer) throws IOException
...
+        int pos = BufferUtil.flipToFill(byteBuffer);
+        try (ReadableByteChannel channel = newReadableByteChannel)
+        {
+            int read = 0;
+            while (read != contentLengthValue)
+                read += channel.read(byteBuffer);
+        }
+        finally
+        {
+            BufferUtil.flipToFlush(byteBuffer, pos);
```
lorban, line 630: Why flipping in a `finally` block?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
```diff
@@ -595,6 +595,21 @@ public static void readFrom(Path path, ByteBuffer buffer) throws IOException
         }
     }
 
+    public static void readFrom(ReadableByteChannel newReadableByteChannel, long contentLengthValue, ByteBuffer byteBuffer) throws IOException
+    {
+        int pos = BufferUtil.flipToFill(byteBuffer);
+        try (ReadableByteChannel channel = newReadableByteChannel)
```
lorban: This is going to close the `ReadableByteChannel`, and this should not happen here: the usual convention is to always close an IO resource in the same code context it was opened.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
```diff
@@ -595,6 +595,21 @@ public static void readFrom(Path path, ByteBuffer buffer) throws IOException
...
+        {
+            int read = 0;
+            while (read != contentLengthValue)
```
lorban: This loop might spin. Not only should it call `Thread.onSpinWait()` when it does (i.e.: when `read` returns 0) but that should also be clearly stated in the method's javadoc.

jetty-ee9/jetty-ee9-servlet/src/main/java/org/eclipse/jetty/ee9/servlet/DefaultServlet.java
```diff
@@ -308,6 +300,17 @@ public void init()
             LOG.debug("resource base = {}", _baseResource);
     }
 
+    private static ByteBufferPool getByteBufferPool(ContextHandler contextHandler)
```
lorban, line 303: I think I saw the same piece of code duplicated in other classes. Can't it be factored out?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
```diff
@@ -0,0 +1,571 @@
...
+ */
+public class CachingHttpContentFactory implements HttpContent.Factory
+{
+    private static final Logger LOG = LoggerFactory.getLogger(CachingHttpContentFactory.class);
+
+    private final HttpContent.Factory _authority;
+    private final ConcurrentMap<String, CachingHttpContent> _cache = new ConcurrentHashMap<>();
+    private final AtomicLong _cachedSize = new AtomicLong();
+    private final ByteBufferPool _byteBufferPool;
+    private int _maxCachedFileSize = 128 * 1024 * 1024;
```
lorban: Those default values should be javadoc'ed.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
```diff
@@ -0,0 +1,571 @@
...
+        private final long _contentLengthValue;
+        private final AtomicLong _lastAccessed = new AtomicLong();
+        private final Set<CompressedContentFormat> _compressedFormats;
+        private final String _lastModifiedValue;
+        private final String _characterEncoding;
+        private final MimeTypes.Type _mimeType;
+        private final HttpField _contentLength;
+        private final Instant _lastModifiedInstant;
+        private final HttpField _lastModified;
+        private final AtomicInteger _referenceCount = new AtomicInteger(1);
```
lorban: You probably should use `Retainable.ReferenceCounter` instead.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
```diff
@@ -0,0 +1,571 @@
...
+        void retain();
+    }
+
+    protected class CachedHttpContent extends HttpContentWrapper implements CachingHttpContent
+    {
+        private final ByteBuffer _buffer;
+        private final String _cacheKey;
+        private final HttpField _etagField;
+        private final long _contentLengthValue;
+        private final AtomicLong _lastAccessed = new AtomicLong();
```
lorban: Looks like a `volatile long` is all you need.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
```diff
@@ -0,0 +1,571 @@
...
+        @Override
+        public boolean isValid()
+        {
+            return true;
+        }
+    }
+
+    protected static class NotFoundHttpContent implements CachingHttpContent
+    {
+        private final AtomicLong _lastAccessed = new AtomicLong();
```
lorban: Looks like a `volatile long` is all you need.

lorban (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
lachlan-roberts: done

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
lachlan-roberts: I made this private now

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/TryPathsHandlerTest.java
lachlan-roberts: I think there would be too many settings to introduce config for.
And this doesn't have init parameters like the `DefaultServlet`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
lachlan-roberts: I think Greg suggested this in previous review because `channel.read` can throw.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
lachlan-roberts: Good point, I will move this try with resources to where this method is used.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
lachlan-roberts: I have updated this loop/javadoc also putting in a case for (read < 0) so please re-review.

jetty-ee9/jetty-ee9-servlet/src/main/java/org/eclipse/jetty/ee9/servlet/DefaultServlet.java
lachlan-roberts: Don't thinks so, its duplicated across ee9 and ee10 which also use different `ContextHandler` classes.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
lachlan-roberts: done

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
lachlan-roberts: done

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
lachlan-roberts: done

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
lachlan-roberts: done

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
lachlan-roberts: reverted

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
lachlan-roberts: I have updated this loop/javadoc also putting in a case for (read < 0) so please re-review.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
```diff
@@ -0,0 +1,124 @@
...
+    {
+        _factory = Objects.requireNonNull(factory);
+        _minFileSize = minFileSize;
+    }
+
+    @Override
+    public HttpContent getContent(String path) throws IOException
+    {
+        HttpContent content = _factory.getContent(path);
+        if (content != null && content.getContentLengthValue() > _minFileSize)
```
sbordet: Should have a check about max size too (cannot be greater than Integer.MAX_VALUE as per `FileChannel.map()` javadocs).

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
```diff
@@ -0,0 +1,124 @@
...
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Objects;
+
+import org.eclipse.jetty.util.BufferUtil;
+import org.eclipse.jetty.util.thread.AutoLock;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class FileMappedHttpContentFactory implements HttpContent.Factory
```
sbordet: I'd call this class `FileMappingHttpContentFactory`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpContentWrapper.java
```diff
@@ -24,13 +25,18 @@
  */
 public class HttpContentWrapper implements HttpContent
```
sbordet: Can we move this class as an inner class of `HttpContent`? To be consistent with all other wrappers we have.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
```diff
@@ -0,0 +1,124 @@
...
+        public FileMappedContent(HttpContent content)
+        {
+            super(content);
+            this._content = content;
+        }
+
+        @Override
+        public ByteBuffer getByteBuffer()
+        {
+            try (AutoLock lock = _lock.lock())
```
sbordet: At least let's use double-checked locking.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
```diff
@@ -0,0 +1,124 @@
...
+                }
+
+                return (_buffer == null) ? super.getByteBuffer() : _buffer;
+            }
+        }
+
+        @Override
+        public long getBytesOccupied()
+        {
+            try (AutoLock lock = _lock.lock())
```
sbordet: Don't duplicate code, call `getByteBuffer()` instead.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpContentWrapper.java
```diff
@@ -110,20 +122,26 @@ public Resource getResource()
     }
 
     @Override
-    public Map<CompressedContentFormat, ? extends HttpContent> getPrecompressedContents()
+    public ByteBuffer getByteBuffer()
```
sbordet: Missing override for `getBytesOccupied()`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
```diff
@@ -0,0 +1,586 @@
...
+            // Read the content into memory if the HttpContent does not already have a buffer.
+            ByteBuffer byteBuffer = httpContent.getByteBuffer();
+            if (byteBuffer == null)
+            {
+                try
+                {
+                    long contentLengthValue = getContentLengthValue();
+                    if (contentLengthValue > _maxCachedFileSize)
+                    {
+                        byteBuffer = _byteBufferPool.acquire((int)contentLengthValue, false);
```
sbordet: This cast to `int` should be handled better, as we don't want to blindly cast, but check if `contentLengthValue < Integer.MAX_VALUE` and only then cache it.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/ResourceHttpContentFactory.java
```diff
@@ -82,29 +72,7 @@ private HttpContent load(String pathInContext, Resource resource)
         if (resource.isDirectory())
             return new ResourceHttpContent(resource, _mimeTypes.getMimeByExtension(resource.toString()));
```
sbordet: `Resource.toString()` is not implemented in `Resource`.
I would not assume "oh, but this is a PathResource so toString() is implemented`.
Plus `toString()` may be changed to a different format and totally break this code.
Why can it not be `pathInContext` like the line below?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/ValidatingCachingContentFactory.java
```diff
@@ -0,0 +1,193 @@
...
+     * @param byteBufferPool the {@link org.eclipse.jetty.io.ByteBufferPool} to use.
+     * @param scheduler scheduler to use for the sweeper, can be null to not use sweeper.
+     * @param sweepDelay time between runs of the sweeper in ms (if 0 never sweep for invalid entries).
+     * @param maxCacheIdleTime amount of time in ms an entry can be unused before evicted by the sweeper (if 0 never evict unused entries).
+     */
+    public ValidatingCachingContentFactory(@Name("authority") HttpContent.Factory authority,
+                                           @Name("validationTime") long validationTime,
+                                           @Name("byteBufferPool") ByteBufferPool byteBufferPool,
+                                           @Name("scheduler") Scheduler scheduler,
+                                           @Name("sweepDelay") long sweepDelay,
```
sbordet: Rename to `sweepPeriod`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/ValidatingCachingContentFactory.java
```diff
@@ -0,0 +1,193 @@
...
+                                           @Name("maxCacheIdleTime") long maxCacheIdleTime)
```
sbordet: Rename to `idleTimeout` like everywhere else.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -65,15 +67,15 @@
     // TODO: see if we can set this to private eventually
     public static final int USE_KNOWN_CONTENT_LENGTH = -2;
 
-    private List<CompressedContentFormat> _precompressedFormats = new ArrayList<>();
-    private WelcomeFactory _welcomeFactory;
+    private final List<CompressedContentFormat> _precompressedFormats = new ArrayList<>();
+    private final Map<String, List<String>> _preferredEncodingOrderCache = new ConcurrentHashMap<>();
+    private final List<String> _preferredEncodingOrder = new ArrayList<>();
```
sbordet: Remove newline.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -101,34 +103,55 @@ public Resource getStylesheet()
...
+            // TODO: can this go in a "StaticContentFactory" that goes after ResourceContentFactory?
             if ((_stylesheet != null) && (path != null) && path.endsWith("/jetty-dir.css"))
                 content = new ResourceHttpContent(_stylesheet, "text/css");
         }
 
         return content;
     }
 
-    public HttpContent.ContentFactory getContentFactory()
+    public HttpContent.Factory getContentFactory()
```
sbordet: Rename to `getHttpContentFactory()`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -101,34 +103,55 @@ public Resource getStylesheet()
...
     {
         return _contentFactory;
     }
 
-    public void setContentFactory(HttpContent.ContentFactory contentFactory)
+    public void setContentFactory(HttpContent.Factory contentFactory)
```
sbordet: Rename to `setHttpContentFactory()`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ResourceHandler.java
```diff
@@ -61,31 +67,49 @@ public ResourceHandler()
...
+        if (context == null)
+            return new NoopByteBufferPool();
+        Server server = context.getContextHandler().getServer();
+        if (server == null)
+            return new NoopByteBufferPool();
+        ByteBufferPool byteBufferPool = server.getBean(ByteBufferPool.class);
+        return (byteBufferPool == null) ? new NoopByteBufferPool() : byteBufferPool;
+    }
+
+    protected HttpContent.Factory setupContentFactory()
```
sbordet: Rename to `setupHttpContentFactory()`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ResourceHandler.java
```diff
@@ -120,7 +144,7 @@ public Request.Processor handle(Request request) throws Exception
     }
 
     // for testing only
-    HttpContent.ContentFactory getContentFactory()
+    HttpContent.Factory getContentFactory()
```
sbordet: Rename to `getHttpContentFactory()`.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/TryPathsHandlerTest.java
sbordet: I would have expected `ResourceHandler` to have a `setHttpContentFactory(HttpContent.Factory)` so that here I could have set just the `ResourceHCF`.

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/ResourceHandler.java
```diff
@@ -97,12 +105,33 @@ public void doStart() throws Exception
...
+    }
+
+    protected HttpContent.Factory setupContentFactory()
+    {
+        HttpContent.Factory contentFactory = new ResourceHttpContentFactory(this, _mimeTypes);
+        contentFactory = new PreCompressedHttpContentFactory(contentFactory, _resourceService.getPrecompressedFormats());
+        contentFactory = new FileMappedHttpContentFactory(contentFactory);
+        contentFactory = new ValidatingCachingContentFactory(contentFactory, Duration.ofSeconds(1).toMillis(), _byteBufferPool);
+        return contentFactory;
+    }
```
sbordet, line 139: Please also add `get/setHttpContentFactory()` so that it's easier to configure before `start()` (does not require subclassing).

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/ValidatingCachingContentFactory.java
```diff
@@ -0,0 +1,193 @@
...
+     *
+     * @param authority the wrapped {@link HttpContent.Factory} to use.
+     * @param validationTime time between filesystem checks in ms to see if an {@link HttpContent} is still valid (-1 never validate, 0 always validate).
+     * @param byteBufferPool the {@link org.eclipse.jetty.io.ByteBufferPool} to use.
+     * @param scheduler scheduler to use for the sweeper, can be null to not use sweeper.
+     * @param sweepDelay time between runs of the sweeper in ms (if 0 never sweep for invalid entries).
+     * @param maxCacheIdleTime amount of time in ms an entry can be unused before evicted by the sweeper (if 0 never evict unused entries).
+     */
+    public ValidatingCachingContentFactory(@Name("authority") HttpContent.Factory authority,
+                                           @Name("validationTime") long validationTime,
```
sbordet: Rename to `validationPeriod`.

sbordet (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
```diff
@@ -0,0 +1,124 @@
...
+        _factory = Objects.requireNonNull(factory);
+        _minFileSize = minFileSize;
+    }
+
+    @Override
+    public HttpContent getContent(String path) throws IOException
+    {
+        HttpContent content = _factory.getContent(path);
+        if (content != null && content.getContentLengthValue() > _minFileSize)
+            return new FileMappedContent(content);
```
gregw: So we don't map the buffer here because we are not sure if something will call `getByteBuffer`?
What exactly is the setup that could cause that to happen?

sbordet: We don't map it here because it's a heavy operation that blocks and takes time, which we don't want to spend in the supposedly fast and light `getHttpContent()`.

sbordet (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
lachlan-roberts: done

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
lachlan-roberts: done

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpContentWrapper.java
lachlan-roberts: done

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpContentWrapper.java
lachlan-roberts: fixed

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
lachlan-roberts: I think this check should instead be `(contentLengthValue <= _maxCachedFileSize)`, and `_maxCachedFileSize` is an `int` so it will be <= `Integer.MAX_VALUE`.

But we also do this check in `isCacheable` so there should be no way to get here with a contentLengthValue greater than `Integer.MAX_VALUE`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/ResourceHttpContentFactory.java
lachlan-roberts: Not sure, this is pre-existing logic which has been maintained here.

So I'm not sure why directories need to be handled differently here, there is `resource.getFileName()` which is better than the `toString()`, but I also think `pathInContext` is good enough.

lachlan-roberts: I have now removed this the `resource.isDirectory` block.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/ValidatingCachingContentFactory.java
lachlan-roberts: done

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/ValidatingCachingContentFactory.java
lachlan-roberts: done

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
lachlan-roberts: done

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
lachlan-roberts: done

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
lachlan-roberts: done

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ResourceHandler.java
lachlan-roberts: done

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ResourceHandler.java
lachlan-roberts: done

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/ValidatingCachingContentFactory.java
lachlan-roberts: done

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/ResourceHandler.java
lachlan-roberts: I have now created these methods, and `setupHttpContentFactory()` is only called if there was no `HttpContentFactory` set manually.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
lachlan-roberts: Changed it to do that now, I had to introduce a `SENTINEL_BUFFER` static final field. And remove the `_processed` boolean.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
gregw: This feels so wrong.  The fast path here is for a file that has already been mapped long ago, so we should just be able to use it directly without a lock and probably without even a volatile lookup.

Firstly are we really sure that mapping a file is a heavy operation?  I would have thought it would be cheaper than reading the file as it just makes an entry into the memory map of the process.   But let's assume that it is expensive....

Then the cache/factory should be able to deal with this in one of several ways:
 1. just block and map it.  Let's see what the cost of this slow path is.
 2. return the original HC until such time as the mapped buffer is loaded, at which time the entry in the cache is replaced with the fast mapped HC
 3. return a HC that waits (preferably threadlessly... but that is another issue) until the cache has been updated with the fast mapped HC.

In short, we need to stop putting policy into `getByteBuffer` that is based on a-priori  knowledge of how it is probably being used.  The `getByteBuffer` method on a `FileMappedHttpContent` should just return the mapped file buffer and leave the policy optimization to be done elsewhere.

lachlan-roberts: Note that when an `HttpContent` is cached it is wrapped by `CachedHttpContent` which will extract out the buffer using `getByteBuffer()`. It is then stored as a final field in `CachedHttpContent` and then the usage of this cached version never hits the volatile just the final field.

If you do not use the `CachingHttpContentFactory`, then using `FileMappedHttpContent.getBuffer()` directly will be hitting this volatile and potentially locking to generate the mapped ByteBuffer.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
sbordet: I would not restrict `maxCachedFileSize` to an `int`, given there may be systems with a lot of memory.
A 2 GiB file is not that large these days.
This is even more so because `getContentLengthValue()` is a `long` already.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpContent.java
```diff
@@ -78,4 +84,136 @@
          */
         HttpContent getContent(String path) throws IOException;
     }
+
+    /**
+     * HttpContent Wrapper.
+     */
+    class HttpContentWrapper implements HttpContent
```
sbordet: Just call it `Wrapper` like all the others.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
gregw: Good point, except now we have just transferred the delay/work from the `getByteBuffer` call on the FMHC to the `getHttpContent` on the CCF.    I.e any work done in getting the buffer in a HC that might be cached will end up being work done in the `getHttpContent` on the cache, so we have the same global lock and/or stampede problems there.

Note that I'm a bit dubious if we should ever map a file that is not being cached, as it is a big investment to make for a once off file.

I'm starting to think that file mapping should actually just be an aspect of the cache, that when it decides to load a resource it can do so either as a mapped file or as a byte buffer.  The cache entry can initially be one that either blocks; or calls the non cached HC;  or eventually queues requests until the real HC is ready and replaces it in the cache.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ResourceHandler.java
```diff
@@ -61,31 +67,60 @@ public ResourceHandler()
...
-//            _mimeTypes = _context == null ? new MimeTypes() : _context.getMimeTypes();
+        // TODO: _mimeTypes = _context == null ? new MimeTypes() : _context.getMimeTypes();
         if (_mimeTypes == null)
             _mimeTypes = new MimeTypes();
 
-        setupContentFactory();
-
+        _byteBufferPool = getByteBufferPool(context);
+        if (_resourceService.getHttpContentFactory() == null)
+            _resourceService.setHttpContentFactory(setupHttpContentFactory());
```
sbordet:
```suggestion
        if (getHttpContentFactory() == null)
            setHttpContentFactory(setupHttpContentFactory());
```

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ResourceHandlerTest.java
```diff
@@ -655,7 +660,19 @@ public void before() throws Exception
...
+            protected HttpContent.Factory setupHttpContentFactory()
+            {
+                // For testing the cache should be configured to validate the entry on every request.
+                HttpContent.Factory contentFactory = new ResourceHttpContentFactory(ResourceFactory.of(getBaseResource()), getMimeTypes());
+                contentFactory = new PreCompressedHttpContentFactory(contentFactory, getPrecompressedFormats());
+                contentFactory = new FileMappingHttpContentFactory(contentFactory);
+                contentFactory = new ValidatingCachingHttpContentFactory(contentFactory, 0, _local.getByteBufferPool());
+                return contentFactory;
+            }
+        };
```
sbordet, line 675: Use `ResourceHandler.setHttpContentFactory(...)` instead of subclassing.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ResourceHandlerTest.java
```diff
@@ -1941,6 +1957,9 @@ public void testEtagIfNoneMatchModifiedFile() throws Exception
         }
         while (Files.getLastModifiedTime(testFile).equals(before));
 
+        // The cached Resource will only go to fileSystem for expiryTime once per second.
```
sbordet: `expiryTime` is now called differently?
Also, I would not assume default values. Explicitly configure the period in `ValidatingCachingHttpContentFactory` with a large-ish value (e.g. `period = 2000`), and here wait for `1.5 * period`, so we don't make this test flaky for 100 ms.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/TryPathsHandlerTest.java
```diff
@@ -94,7 +96,16 @@ public void dispose()
...
-        ResourceHandler resourceHandler = new ResourceHandler();
+        ResourceHandler resourceHandler = new ResourceHandler()
+        {
+            @Override
+            protected HttpContent.Factory setupHttpContentFactory()
+            {
+                // We don't want to cache not found entries for this test.
+                return new ResourceHttpContentFactory(ResourceFactory.of(getBaseResource()), getMimeTypes());
+            }
+        };
```
sbordet, line 107: Use the setter now, don't subclass.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/TryPathsHandlerTest.java
```diff
@@ -160,7 +171,15 @@ public void process(Request request, Response response, Callback callback)
...
```
sbordet, line 182: Use the setter don't subclass.

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/ResourceHandler.java
```diff
@@ -97,12 +105,44 @@ public void doStart() throws Exception
         if (_mimeTypes == null)
             _mimeTypes = _context == null ? new MimeTypes() : _context.getMimeTypes();
 
-        _resourceService.setContentFactory(new ResourceContentFactory(this, _mimeTypes, _resourceService.getPrecompressedFormats()));
+        _byteBufferPool = getByteBufferPool(_context);
+        if (_resourceService.getHttpContentFactory() == null)
+            _resourceService.setHttpContentFactory(setupHttpContentFactory());
```
sbordet:
```suggestion
        if (getHttpContentFactory() == null)
            setHttpContentFactory(setupHttpContentFactory());
```

sbordet (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
sbordet: File mapping should not be a cache issue. 
If I want to map a file because an external process also maps it and changes it, I would like to return a mapped buffer even if there is no cache in the front.

lachlan-roberts: > we have just transferred the delay/work from the getByteBuffer call on the FMHC to the getHttpContent on the CCF. I.e any work done in getting the buffer in a HC that might be cached will end up being work done in the getHttpContent on the cache, so we have the same global lock and/or stampede problems there.

Because we do the file mapping in `getByteBuffer()` we don't have a stampede of threads all mapping the file. 

Each thread gets a `FileMappedHttpContent` from the factory, they all try to add them to the cache, but the creation of the `CachingHttpContent` (which calls `getByteBuffer()`) is done atomically inside the `_cache.computeIfAbsent` so only one thread will ever do the buffer mapping of the file, the rest of the `FileMappedHttpContent` instances will be discarded because they did not win the race to add it into the cache and they will not also map the file because `getByteBuffer()` was never called.

gregw: Maybe, but that would be yet another case where the implementation of `getByteBuffer` has to somehow telepathically know the intentions of the caller  - does it really want a mapped buffer or not?  is it calling it to cache it or not?  It's just another indication that we don't have the API exactly right yet... together with the fact that we've just had several cycles debating if it is best to do the work in the getContent or the getBuffer, only to realise that if it is cached then all the work is done in the getContent anyway!    We just can't tell why `getByteBuffer` is being called, so we can't optimise it or apply any reasonable policy that holds generally.  

There is a lot of goodness in this PR and I think the things I don't like are due to the API not being correct yet.   I think we need to get the design right (in another PR) and then focus back in on micro optimising the implementations.  So let's get this merged and then consider the API.

gregw: > Each thread gets a `FileMappedHttpContent` from the factory, they all try to add them to the cache, but the creation of the `CachingHttpContent` (which calls `getByteBuffer()`) is done atomically inside the `_cache.computeIfAbsent` so only one thread will ever do the buffer mapping of the file, the rest of the `FileMappedHttpContent` instances will be discarded because they did not win the race to add it into the cache and they will not also map the file because `getByteBuffer()` was never called.

And this is yet another example of a fragile design.   The cache is only not stampeding because we "known" that all the expensive stuff is done in `getByteBuffer`.

I'd much rather a cache design where the element initially put into the cache could make an explicit choice when other requests arrived for the same resource to block them waiting; queue them asynchronously; or dispatch them to the uncached content.     We would have explicit behavior encoded rather than emergent behavior from disciplined implementations of cheap constructors and expensive getBuffers.

sbordet: > Maybe, but that would be yet another case where the implementation of getByteBuffer has to somehow telepathically know the intentions of the caller

That is not true. The implementation does not care at all. It just returns a mapped buffer.
Whether it is the same one mapped once, or many mapped multiple times, the caller does not know and does not care.
You can argue that it is not efficient to map multiple times, but it has nothing to do with the API design.

> We just can't tell why getByteBuffer is being called, so we can't optimise it or apply any reasonable policy that holds generally.

I still don't understand what you are talking about here.
Not a single implementation cares about why or when `getByteBuffer()` is being called.
It is called, it returns a `ByteBuffer` if it can, done.

I still think we are looking for problems that don't exist?

> I'd much rather a cache design where the element initially put into the cache could make an explicit choice when other requests arrived for the same resource to block them waiting; queue them asynchronously; or dispatch them to the uncached content.

We have said repeatedly that this is a better job for a 3rd party cache implementation, not us.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
```diff
@@ -595,6 +596,39 @@ public static void readFrom(Path path, ByteBuffer buffer) throws IOException
...
+     * @param contentLengthValue the length of the content to read.
+     * @param byteBuffer the buffer to read into.
+     * @throws IOException if an I/O error occurs.
+     */
+    public static void readFrom(ReadableByteChannel readableByteChannel, long contentLengthValue, ByteBuffer byteBuffer) throws IOException
+    {
+        int pos = BufferUtil.flipToFill(byteBuffer);
+        try
+        {
+            int totalRead = 0;
```
lorban: Should be a `long`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
gregw: > > Maybe, but that would be yet another case where the implementation of getByteBuffer has to somehow telepathically know the intentions of the caller
> 
> That is not true. The implementation does not care at all. It just returns a mapped buffer. 

Exactly - a "mapped buffer" because it some how has to know that "an external process also maps it and changes it".

We should just have a HC that serves content as best it can.  If the caller wants a more efficient mechanism, then it is their responsibility to buffer that content or perhaps create a MappedFIleHC.

 
> We have said repeatedly that this is a better job for a 3rd party cache implementation, not us.

Except that we don't have an API that is powerful enough to use a 3rd party cache.  We couldn't use anything that needs to look at cache-control headers for example, as they are not available in getByteBuffer?


I don't understand why you are fighting against this proposal so hard, it can do everything that the getByteBuffer style can and I'm not proposing that initially we change much from what we are currently doing.  But by having a fully powerful API, it allows us to do more powerful things in the future.   It also means that we leverage our existing produce API rather than invent yet another API?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
lachlan-roberts: So you're saying if the `maxCachedFileSize > Integer.MAX_VALUE` the entry should be added to the cache but there should be no buffer in the cache? So we will be caching `lastModified` `contentLength` etc.. but not caching the content.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
gregw: I don't think so as it is size limited by the ByteBuffer, which can only be int sized.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ResourceHandlerTest.java
lachlan-roberts: Its already configured at 0, just forgot to remove these sleeps.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
lachlan-roberts: done

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpContent.java
lachlan-roberts: done

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappedHttpContentFactory.java
gregw: Closing this discussion to move it elsewhere and not distract from this PR.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java
gregw: Thinking about it, it doesn't matter either way. If the method throws, the buffer cannot be trusted.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ResourceHandlerTest.java
gregw: Is there a reason to prefer one technique over the other?   

They are slightly different in that subclassing will create the HCF during start, when all configuration is set/known, and that it will then remove that instance during stop, so a new instance is created on next start cycle, using latest configuration.    The setter will set a HCF that will be durable over stop/start cycles, so it will not see config changes in the RH.

So I kind of prefer the subclass style when the configuration of the HCF is entirely dependent on the RH.  The setter style is good for when the HCF has other external configuration.

We should have tests that use both style to make sure they work.   In these specific tests, it doesn't really matter one way or the other.

lachlan-roberts: After a hangout with Simone we decided to not add the setter method for HttpContentFactory at all as it makes the configuration more difficult.

gregw:
@lachlan-roberts @sbordet @lorban can we all work together to land this one.  It has been outstanding too long.

gregw (approved):
I'm OK for t his to be merged as an interim step.

PR #8774, opened 2022-10-27
Added SizeLimitHandler

gregw:
Add SizeLimitHandler

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/handler/SizeLimitHandler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/handler/SizeLimitHandler.java
@@ -66,10 +66,8 @@ public class SizeLimitHandler extends HandlerWrapper
         if (_requestLimit >= 0)
         {
             long contentLength = baseRequest.getContentLengthLong();
-            if (contentLength < 0)
-                contentLength = baseRequest.getHttpFields().getLongField("X-Content-Encoding");
             if (contentLength > _requestLimit)
-                throw new BadMessageException(413, "Request Body too large");
+                throw new BadMessageException(413, "Request Body too large: " + contentLength + ">" + _requestLimit);
             else if (contentLength < 0)
             {
                 baseRequest.getHttpInput().addInterceptor(new HttpInput.Interceptor()
@@ -110,6 +108,7 @@ public class SizeLimitHandler extends HandlerWrapper
                             throw new BadMessageException(413, "Response body too large: " +
                                 _written + ">" + _responseLimit);
                     }
+                    getNextInterceptor().write(content, last, callback);
                 }
 
                 @Override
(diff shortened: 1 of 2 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/handler/SizeLimitHandler.java
```diff
@@ -0,0 +1,131 @@
...
+import org.eclipse.jetty.server.HttpOutput;
+import org.eclipse.jetty.server.Request;
+import org.eclipse.jetty.util.Callback;
+import org.eclipse.jetty.util.IncludeExcludeSet;
+import org.eclipse.jetty.util.InetAddressSet;
+import org.eclipse.jetty.util.log.Log;
+import org.eclipse.jetty.util.log.Logger;
+
+/**
+ * InetAddress Access Handler
```
joakime: Need to fix the javadoc

jetty-server/src/main/java/org/eclipse/jetty/server/handler/SizeLimitHandler.java
```diff
@@ -0,0 +1,131 @@
...
+    @Override
+    public void handle(String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response) throws IOException, ServletException
+    {
+        if (_requestLimit >= 0)
+        {
+            long contentLength = baseRequest.getContentLengthLong();
+            if (contentLength < 0)
+                contentLength = baseRequest.getHttpFields().getLongField("X-Content-Encoding");
+            if (contentLength > _requestLimit)
+                throw new BadMessageException(413, "Request Body too large");
```
joakime: Order of interceptors is important.
If this is after the GzipHandler then you are basing the limits on uncompressed request/response.
Before the GzipHandler and you are basing the limits on raw network bytes.
Right?
Seems like an important detail for the javadoc.

gregw: yeah,  I removed the special handling for X-Content-Encoding, as that muddied the picture.  I'll add some javadoc.

jetty-server/src/main/java/org/eclipse/jetty/server/handler/SizeLimitHandler.java
```diff
@@ -0,0 +1,209 @@
...
+    private void checkRequestLimit(long size)
+    {
+        if (_requestLimit >= 0 && size > _requestLimit)
+            throw new BadMessageException(413, "Request body is too large: " + size + ">" + _requestLimit);
+    }
+
+    private void checkResponseLimit(long size)
+    {
+        if (_responseLimit >= 0 && size > _responseLimit)
+            throw new BadMessageException(413, "Response body is too large: " + size + ">" + _responseLimit);
```
lachlan-roberts: Should these be extensible so the error can be customized. 

Still think 413 is a weird code here because I thought that was only for the request size.

gregw:
@lachlan-roberts can you re-review

lachlan-roberts (approved):

joakime:
Don't forget to add PRs and Issues to projects.

PR #8775, opened 2022-10-27
Fixes #8770 - Review whether to send request body in redirects.

### LINKED ISSUE #8770: Review whether to send request body in redirects
**Jetty version(s)**
10+

**Enhancement Description**
`HttpRedirector` should probe the request content to know whether it's reproducible like e.g. `AuthenticationProtocolHandler` does.

sbordet:
Now the original request body is re-sent only if the redirect status code is 307 or 308.

In the other cases, it is a redirect to a GET method, so the Location is followed without resending the body.

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/HttpRedirector.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/HttpRedirector.java
@@ -28,6 +28,7 @@ import org.eclipse.jetty.client.api.Request;
 import org.eclipse.jetty.client.api.Response;
 import org.eclipse.jetty.client.api.Result;
 import org.eclipse.jetty.client.util.BufferingResponseListener;
+import org.eclipse.jetty.http.HttpHeader;
 import org.eclipse.jetty.http.HttpMethod;
 import org.eclipse.jetty.http.HttpStatus;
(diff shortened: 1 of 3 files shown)


jetty-client/src/main/java/org/eclipse/jetty/client/HttpRedirector.java
```diff
@@ -85,11 +87,11 @@ public boolean isRedirect(Response response)
     {
         switch (response.getStatus())
```
joakime, line 88: replace with `return HttpStatus.isRedirection(response.getStatus())`

joakime (changes requested):
Small improvement to use existing code in HttpStatus, otherwise it looks good.

jetty-client/src/main/java/org/eclipse/jetty/client/HttpRedirector.java
sbordet: Can't do that, because certain 3XX codes are not really redirects that can be handled here, such as `300 Multiple Choices`, `304 Not Modified`, `305 Use Proxy`, etc.
~I can change it into an expression switch though.~ I cannot, it's not Java 17 😅

joakime (approved):

PR #8777, opened 2022-10-27
Jetty 12 - Example of using `TryPathsHandler` with PHP

joakime:
+ This utilizes the matching logic of the `PathMappingsHandler`
+ Also fixed a paths entry with query string bug in `TryPathsRequest`


sbordet:
Closed in favor of #8781.

PR #8780, opened 2022-10-27
Issue #8779 - CompactPathRule should not drop query section from modified URI

### LINKED ISSUE #8779: CompactPathRule drops query section on use
**Jetty version(s)**
Jetty 10.0.12

**Description**
Using the `CompactPathRule` will result in the query section of the input URI being stripped.

joakime:
+ Simply use the original URI path+query when applying the rule.

### DIFF
--- a/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/CompactPathRule.java
+++ b/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/CompactPathRule.java
@@ -35,7 +35,7 @@ public class CompactPathRule extends Rule implements Rule.ApplyURI
     @Override
     public void applyURI(Request request, String oldURI, String newURI) throws IOException
     {
-        String uri = request.getRequestURI();
+        String uri = oldURI;
         if (uri.startsWith("/"))
             uri = URIUtil.compactPath(uri);
         request.setHttpURI(HttpURI.build(request.getHttpURI(), uri));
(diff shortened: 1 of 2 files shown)


jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/CompactPathRule.java
```diff
@@ -35,7 +35,7 @@ public CompactPathRule()
     @Override
     public void applyURI(Request request, String oldURI, String newURI) throws IOException
     {
-        String uri = request.getRequestURI();
+        String uri = oldURI;
```
joakime: This is the key change, don't use the servlet `getRequestURI()` as that doesn't include the query section, but instead use the passed-in `oldURI` which is the path+query.

jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/CompactPathRule.java
```diff
@@ -35,7 +35,7 @@ public CompactPathRule()
...
         if (uri.startsWith("/"))
             uri = URIUtil.compactPath(uri);
```
joakime, line 40: Reminder, the `URIUtil.compactPath(String)` API preserves the query section of the input string, and only operates on the path portion.

sbordet (dismissed):

sbordet (approved):

PR #8785, opened 2022-10-28
Jetty 12 - Reworking ResourceService.dirAllowed to DirBehavior

joakime:
Experiment to see what a ResourceService would look like if Directory serving had an option to SKIP serving if not possible to serve as a Welcome File.
This will skip the creation of the Directory Listing, and also the Forbidden Response.

This is to see if we can integrate TryPathsHandler + PathMappingsHandler + ResourceHandler in the same series of requests.

### DIFF
--- a/jetty-core/jetty-tests/jetty-test-core-integration/src/test/java/org/eclipse/jetty/test/core/integration/ResourceHandlerAndPhpTest.java
+++ b/jetty-core/jetty-tests/jetty-test-core-integration/src/test/java/org/eclipse/jetty/test/core/integration/ResourceHandlerAndPhpTest.java
@@ -1,0 +1,283 @@
+//
+// ========================================================================
+// Copyright (c) 1995-2022 Mort Bay Consulting Pty Ltd and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.test.core.integration;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.eclipse.jetty.http.HttpHeader;
+import org.eclipse.jetty.http.HttpStatus;
+import org.eclipse.jetty.http.HttpTester;
+import org.eclipse.jetty.http.pathmap.ServletPathSpec;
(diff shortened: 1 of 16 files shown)


jetty-core/jetty-tests/jetty-test-core-integration/src/test/java/org/eclipse/jetty/test/core/integration/ResourceHandlerAndPhpTest.java
```diff
@@ -0,0 +1,283 @@
...
+import org.junit.jupiter.api.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class ResourceHandlerAndPhpTest
```
joakime, line 55: @sbordet I mocked up a complete example here of the scenario we've talked about.

gregw:
What is the motivation for the SKIP semantic?

sbordet:
> What is the motivation for the SKIP semantic?

To make the `ResourceHandler` delegate to another handler, or possibly return `null` to indicate that it does not handle the directory (e.g. it does not exist) but without producing a 404, because it may be handled by another `Handler`.

This is an experiment needed by FastCGI proxying, but still not sure it's the right way to go.

We need this:

```
/ -> rewritten to /index.php -> hits fcgi proxy
/wp-admin/ -> rewritten to /wp-admin/index.php -> hits fcgi proxy
/<custom>/ -> rewritten to /index.php?p=/<custom>/ -> hits fcgi proxy
/wp-content/.../some.js -> no rewrite -> hits ResourceHandler
```

We need a way to distinguish between an existing directory `<dir>` (such as `/` or `wp-admin/`) and rewrite it to `/<dir>/index.php` (like we do with welcome files), and a "virtual" directory that needs a completely different rewrite to `/index.php?p=<dir>`.

Not sure this PR can do it, it was an idea worth exploring, but probably not enough.

gregw:
It kind of feels arbitrary to skip directories because you know a later handler wishes to handle them.  The same could be said for *.jsp files, or any special file.

The resource handler should handle a resource if it exists.  If it doesn't exist it can pass through to another handler (or 404).
If you want special handling of some resources, then you need to put some kind of conditional handler before the resource handler.  Trying to make it generic enough to handle all special cases feels like the wrong way to go.

Besides, if you want to make some resources transparent to the resource handler/service, you just give it a factory that makes those resources invisible.  I.E. if you want to skip directories, then give it a resource factory that always returns null for directories.    Then if you want it to ignore *.jsp files, the same approach works without modification to the resource service/handler

sbordet:
Closed in favor of https://github.com/eclipse/jetty.project/pull/8856.

PR #8787, opened 2022-10-31
Issue #8786 - add configuration for KeyStoreScanner to not resolve aliases

### LINKED ISSUE #8786: KeyStoreScanner is not able to monitor a symlink file and always resolves to the target.
**Jetty version(s)**
9.4.x, 10.0.x, 11.0.x

**Description**
`KeyStoreScanner` is not able to monitor a symlink file and always resolves to the target.

There is a test `org.eclipse.jetty.test.KeyStoreScannerTest#testReloadChangingSymbolicLink` which is incorrectly testing that this use case works, but it is currently not possible to configure in `KeyStoreScanner`.

lachlan-roberts:
## Issue #8786

Allow the `KeyStoreScanner` to monitor the symlink file itself instead of the target of the symlink.


sbordet (changes requested):
@lachlan-roberts why is this necessary? See also comment on the issue.

jetty-util/src/main/java/org/eclipse/jetty/util/ssl/KeyStoreScanner.java
```diff
@@ -43,6 +43,11 @@ public class KeyStoreScanner extends ContainerLifeCycle implements Scanner.Discr
     private final Scanner _scanner;
 
     public KeyStoreScanner(SslContextFactory sslContextFactory)
+    {
+        this(sslContextFactory, false);
```
sbordet: The default should be `true` because that was the default behavior prior to this change.

sbordet (changes requested):
"KeyStore" has the capital "S", so in the test class this should respected in variable names and pretty much everywhere.

Also, I won't mind to have the `jetty-ssl-context-reload.xml` modified to support this additional property, and Jetty module `ssl-reload` as well.

Also, please update the documentation `module-ssl-reload.adoc` with a small paragraph after the module properties that explains with an example the difference between the 2 cases.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-ssl-reload.adoc
```diff
@@ -22,3 +22,5 @@ The module properties are:
 ----
 include::{JETTY_HOME}/modules/ssl-reload.mod[tags=documentation]
 ----
+
+The `resolveAlias` property is used to specify whether aliases should be resolved in the path of the KeyStore. If set to false and the path of the KeyStore is a symbolic link, the scanner will monitor the symbolic link file for changes instead of its target.
```
sbordet: I would rename the property to `followLinks` because that's how it is named in the JDK (and used by `Scanner`), it's easier to understand, and it is not related to our own internal concept of "alias".

Also, every sentence on a new line.

tests/test-integration/src/test/java/org/eclipse/jetty/test/KeyStoreScannerTest.java
```diff
@@ -61,7 +61,7 @@
     public WorkDir testdir;
     private Server server;
     private Path keystoreDir;
```
sbordet, line 63: Capital `S` since you changed the others.

tests/test-integration/src/test/java/org/eclipse/jetty/test/KeyStoreScannerTest.java
```diff
@@ -124,7 +129,7 @@ public void testKeystoreHotReload() throws Exception
 
         // Switch to use newKeystore which has a later expiry date.
         useKeystore("newKeystore");
```
sbordet: Capital `S`.

sbordet (changes requested):

sbordet (approved):
LGTM.
If you want to fix all keystore occurrences with keyStore would be great, given that you changed 80% of them but not all.

Issue #8791, opened 2022-11-01
Jetty-12 Should Request.Processors be reusable?


gregw:
Currently in Jetty-12, we do not specify in the `Request.Processor` contract/javadoc if it can be reused or not.  i.e, once a Processor has been obtained from `Handle.handle`, can it be used over and over with many different request/response/callback tupples?

This is analogous to `RequestDispatchers` in the servlet API, which currently may be reused multiple times.

In Jetty-12, many of our processor implementations will work fine if reused.  However, we also have many examples of `Handler`s that wrap a request in the `handle` method and then reuse that wrapper in the `process` method rather than re-wrapping. In fact, often it is the request wrapper that is also the processor.

At the very least, we should javadoc if processors may not be re-used or not.

However, I think there are use-cases for re-using processor, so perhaps we should consider either: making all our processors reusable (could be some cost in duplicate wrapping); or making processors declare if they can be reused or not.

joakime:
No longer relevant, we don't have Request.Processors anymore.

PR #8793, opened 2022-11-01
Reusable Request.Processor

gregw:
Experiment for #8791 to see what is involved to have a reusable `Request.Processor`


gregw:
This is not so bad.  The code is a lot more readable that what we currently have.
However, we are currently wrapping twice, so we need a mechanism to save the wrapper.   Either a request can know it's last wrapper (which might be useful elsewhere) OR we could just use request attributes?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -644,11 +644,29 @@ public Request.Processor handle(Request request) throws Exception
         if (processor != null)
             return processor;
 
-        // The contextRequest is-a Supplier<Processor> that calls effectively calls getHandler().handle(request).
-        // Call this supplier in the scope of the context.
-        Request.Processor contextScopedProcessor = _context.get(contextRequest, contextRequest);
-        // Wrap the contextScopedProcessor with a wrapper that uses the wrapped request
-        return contextRequest.wrapProcessor(contextScopedProcessor);
+        Context lastContext = __context.get();
```
lachlan-roberts, line 664: Can we use some method in `Context` to do all this instead of doing it here.
Should be something like:
```java
processor = _context.get(() -> getHandler().handle(contextRequest), contextRequest);
```

but I think that one doesn't quite work because it doesn't handle the exception from handle.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -1240,4 +1258,54 @@ public String toString()
...
+
+        public ContextProcessor(Request.Processor processor)
+        {
+            _processor = processor;
+        }
+
+        @Override
+        public void process(Request request, Response response, Callback callback) throws Exception
+        {
+            Request contextRequest = wrap(request); // TODO reuse any previous wrapper
```
lachlan-roberts: I think this method can be simplified to:
```java
Request contextRequest = wrap(request); // TODO reuse any previous wrapper
Response contextResponse = new ContextResponse(_context, contextRequest, response);
_context.run(() ->
{
    try
    {
        _processor.process(contextRequest, contextResponse, callback);
    }
    catch (Throwable th)
    {
        Response.writeError(contextRequest, contextResponse, callback, th);
    }
}, contextRequest);
```

lachlan-roberts (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
gregw: We were doing that, but it is creating a capturing lambda in a hot path.   So yeah it is a little bit of duplicated behaviour, but it is all in the same code area.  Maybe we need to link with comments so we know where to update if anything changes...

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
gregw: yeah... but I want to avoid creating lambdas on the hot path.

gregw:
@sbordet @lorban This experiment still has a lot more work to do.... but I'm definitely think we need to do something.  The WrapperProcessor class is really horrid and we have used it incorrectly in a number of places.
So this is a heads up that I'd like you to tune into this PR sooner or later.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
gregw: I've factored out the duplicate code into protected methods

gregw:
@sbordet @lorban @lachlan-roberts @joakime This PR has grown a little bit bigger than intended, making it hard to review.   So it would be good to get a review now, before I complete the process of replacing `Request.WrapperProcessor` (which will touch a lot of files).

The key idea in this PR is to make processors re-usable and/or re-entrant.  This is to mimic the functionality of servlet RequestDispatchers and will increase our flexibility when we implement features like re-dispatching welcome files.

The secondary purpose of this PR is to get rid of `Request.WrapperProcessor` and similar overly complex "optimizations" that are based on object allocation phobia.   The classic example is [ContextHandler](https://github.com/eclipse/jetty.project/blob/jetty-12.0.0.alpha2/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java#L635-L644) which is incredibly hard to understand with its supplier of processor and runnables etc.      I think the clean up of that code is probably worth it in itself.  So this may be the primary reason for this PR and the reentrant stuff is just bonus.

So having a look at `ContextHandler` is a good place to start.

To replace `Request.WrapperProcessor`, I now have `Request.ReWrappingProcessor` and `Handler.ProcessingWrapper`.   The naming is not  that great, but I hope it will be less confusing once `WrapperProcessor` is gone.   The key concept here is a processor wrapper that rewraps requests passed to 'process', either by re-using the initial wrapper created in 'handle' or by making a new wrapper (needed if the processor is reused).

The `ProcessingWrapper` is just a convenience class that maps the abstracts of `ReWrappingProcessor` to protected methods on the handler.

The biggest refactor as a result of these changes has been to `GzipHandler`, as most of the work that was done in `handle` now has to be done in `process` so subsequent requests can be processed.   I think the code reads better.      

Big question is, does this approach cost anything in performance?  It probably allocates some light wrappers a bit more, but then the calling paths are simpler and might be better optimized?

gregw:
@joakime @sbordet One issue I have not solved, is how to do the rewrite handler, which is heavily dependent on the the deprecated `WrapperProcessor` class.  Your input would be appreciated to work out how to do that a different way.

sbordet:
> One issue I have not solved, is how to do the rewrite handler, which is heavily dependent on the the deprecated WrapperProcessor class.

Wait, what? Why is `WrapperProcessor` deprecated?
Are we not putting the horses in front of the chariot here?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java
```diff
@@ -690,4 +690,83 @@ public NonBlocking()
             }
         }
     }
+
+    /**
+     * <p>A {@link Handler.Wrapper} that supports wrapping of the request and optional wrapping of the response.
+     * The process methods can be implemented on this handler class and take the wrapped request object so that down casting
+     * is not necessary.</p>
+     */
+    abstract class ProcessingWrapper<W extends Request.Wrapper> extends Wrapper
```
sbordet: Not a good idea. We are now forced to make all request wrappers proper classes and `public`.

sbordet:
Sorry, can you please detail what the problem is?

gregw:
> Sorry, can you please detail what the problem is?

As described in issue #8791, we don't actually specify the lifecycle of a `Request.Processor`, specially with regards to reuse and/or reentrancy.       Your work on redispatching welcome files and also my pondering on the resource service made me realise that their might be use-cases for code to call `Handler.handle(Request)` and then to hold onto the resulting `Request.Processor` and use it for multiple subsequent requests. 

Now we could document that the lifecycle is that you must always use a processor once and only once after getting it from handle and that would resolve #8791,   but enforcing behaviour in javadoc is not a strong defence.   Users that have been exposed to servlet `RequestDispatcher`s will be used to the ability to get them once (expensive) and then use them many times (cheap), so it is plausible that somebody will try that with our processors.

So I started an experiment to see what would be involved to make our processors re-usable.   It turns out that many of them are already reusable and only a few specific usages are not really reusable.  These are mostly the usages of `WrapperProcessor` or similar classes like `ContextHandler`.  These usages are rather convoluted and hard to understand, with strange outcomes like having a processor based on a lambda that captures the wrapped request  and then ignores the actual request passed to the processor.     This is fragile behaviour that will only work if we call processors in exactly the way that our existing handlers do.  If we (or users) start using the processors in different ways, then our code will not be robust. 

I found that by doing this work, a lot of code was greatly simplified as it is able to call specific process methods with the wrapped request correctly type in the interface.  This avoids the need to ignore requests in favour of captured wrapped requests.

I think the result is more robust and more flexible code.   It is also a lot more consistent about what kinds of work are done in the blocking `handle` method, vs what is done in the async `process` method... with more work being done in `process` (which is good).  Perhaps it allocates a tiny bit more, but there may be savings from being less complex.

However, I've not yet solved the rewrite rules use-case, whose implementation is currently heavily based on `WrapperProcessor`... even though that is a processor that doesn't come from a call to `Handler.handle` so it already breaks any "javadoc rules" we might write and shows that there are use-cases to use processors in different ways.     Perhaps rewrite is a case that does need  `WrapperProcessor`, so maybe we don't deprecate it, or maybe move it to the rewrite handler.  But I also think we have to carefully consider what does re-dispatch mean for the rewrite handler.

So our code works fine without this.  I just think it can work better with it and be more robust, more flexible and easier to maintain at the same time.

gregw:
@joakime @sbordet I've "fixed" the rewrite issues in this PR by effectively moving the old `WrapperProcessor` to `Rule.RequestProcessor` and given it the ability to rewrap requests if it's processors are reused.

So this is PR is now ready for review.  The CI failures appear to be CI flakes.

Before review, have a good read of my two longer comments above (https://github.com/eclipse/jetty.project/pull/8793#issuecomment-1307938216 & https://github.com/eclipse/jetty.project/pull/8793#issuecomment-1308572168 ). Starting with the changes to `ContextHandler` are a good way into this PR.

Then perhaps look at `ProxiedRequestHandler`. I'm not sure exactly sure that class is used, but it is a simple example of the style.

Once you understand what is happening, then look at `GzipHandler`, which has the most significant changes (for the better I think), which would only get even better if we ultimate factor out all the path/method conditions to something like #8676 

Finally, I'm not exactly sure how to thoroughly test the processor reuse mechanism.  I have done test runs with the rewrapping disabled in the ContextHandler, so processed requests are always rewrapped with a different wrapper. This found lots of down stream processors that could handle reuse and they have been fixed.

gregw:
@lorban can I get a performance measurement on this branch for both CPU and garbage?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -638,17 +653,30 @@ public Request.Processor handle(Request request) throws Exception
...
         // Does this handler want to process the request itself?
         Request.Processor processor = processByContextHandler(contextRequest);
         if (processor != null)
             return processor;
 
-        // The contextRequest is-a Supplier<Processor> that calls effectively calls getHandler().handle(request).
-        // Call this supplier in the scope of the context.
-        Request.Processor contextScopedProcessor = _context.get(contextRequest, contextRequest);
-        // Wrap the contextScopedProcessor with a wrapper that uses the wrapped request
-        return contextRequest.wrapProcessor(contextScopedProcessor);
```
gregw, line 651: This is a prime example of confusing code that this PR replaces.   It is entirely unclear what line 649 is doing, specifically that it is ultimately doing a `processor = getHandler().handle(contextRequest)` and that the `Processor ContextRequest#get()` method should be looked at for the flow of control.

It is further unclear that line 651 returns the contextRequest as the wrapped processor, but with the normal process method extended, so that the request is ultimately used as a `Runnable` passed to `context.run(Runnable)`

This is very convoluted code, just to avoid some allocation of cheap wrappers.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/StatisticsHandler.java
```diff
@@ -422,19 +437,6 @@ public Content.Chunk read()
...
-            {
-                _statisticsRequest = (StatisticsRequest)processor;
-                return super.wrapProcessor(processor);
-            }
-
-            @Override
-            public void process(Request ignored, Response response, Callback callback) throws Exception
-            {
-                super.process(ignored, new MinimumDataRateResponse(this, response), callback);
-            }
```
gregw, line 437: Another good example of confusing code this PR resolves.   The current usage style often results in two copies of the request in scope, with one needing to be ignored.   This is confusing and in this case has resulted in a wrong name for a request that is obviously not ignored.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/SessionHandler.java
```diff
@@ -717,41 +717,47 @@ public void setSessionTrackingModes(Set<SessionTrackingMode> sessionTrackingMode
...
-            (servletContextRequest == null ? null : servletContextRequest.getServletApiRequest());
-        if (servletApiRequest == null)
-            throw new IllegalStateException("Request is not a valid ServletContextRequest");
-
         Request.Processor processor = getHandler().handle(request);
         if (processor == null)
             return null;
 
-        addSessionStreamWrapper(request);
+        return new SessionProcessor(processor);
```
lachlan-roberts: Is there anyway we could avoid creating this every time similar to the `ServletHandler.PROCESSOR`?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java
lachlan-roberts: Could this be something like 
```java
abstract class ProcessingWrapper<REQUEST extends Request, RESPONSE extends Response> extends Wrapper
```

so we could get the extended type of the response in process as well as the request, and use types which do not directly extend the Wrappers.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java
```diff
@@ -690,4 +690,83 @@ public NonBlocking()
...
+         * wrapper.
+         * @throws Exception If the nextProcessor throws.
+         */
+        protected void process(W wrappedRequest, Response wrappedResponse, Callback callback, Request.Processor nextProcessor)
+            throws Exception
+        {
+            nextProcessor.process(wrappedRequest, wrappedResponse, callback);
+        }
+
+        private class ProcessingWrapperProcessor extends Request.ReWrappingProcessor<W>
```
lachlan-roberts: Same question for the `Request.ReWrappingProcessor<W>`.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/SessionHandler.java
gregw: No, because we don't wrap the request, so we have nowhere to hold the next processor.    This is a very light/cheap wrapper, so let's see what the performance testing says.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java
gregw: yep. done.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java
gregw: yep

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/SessionHandler.java
gregw: Hmm it can be done with a setter/getter on ServletContextRequest.  A bit ugly, but have a look and tell me what you think

gregw:
@sbordet @lachlan-roberts @lorban Can we get this reviewed and merged this week.  It is really difficult to stop this one going stale.
@lorban any progress on performance testing?

lachlan-roberts (approved):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -408,10 +407,20 @@ public boolean removeEventListener(EventListener listener)
         return false;
     }
 
+    protected ClassLoader enterScope(Request contextRequest)
+    {
+        ClassLoader lastLoader = Thread.currentThread().getContextClassLoader();
+        __context.set(_context);
```
joakime, line 413: Isn't it possible for enterScope to occur twice in the same thread?

joakime (approved):

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/Rule.java
```diff
@@ -55,4 +58,101 @@ public String toString()
...
+                wrapped = wrapped instanceof Request.Wrapper w ? w.getWrapped() : null;
+            }
+
+            // We need a new instance of the wrapper
+            Request wrapper = wrap(request);
+            process(wrap(wrapper, response), callback);
+        }
+    }
+
+    public static class UriRequestProcessor extends RequestProcessor
```
sbordet, line 129: Rename to `HttpURIRequestProcessor`.

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/Rule.java
```diff
@@ -55,4 +58,101 @@ public String toString()
...
+        protected Request wrap(Request request)
+        {
+            return new Request.Wrapper(request)
+            {
+                @Override
+                public HttpURI getHttpURI()
+                {
+                    return _uri;
+                }
+            };
```
sbordet, line 155: Why not just returning `this` here? Do we really need to wrap this wrapper, with the same overridden method, returning the same `HttpURI`?

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/Rule.java
```diff
@@ -55,4 +58,101 @@ public String toString()
...
+                {
+                    process(wrap(this, response), callback);
+                    return;
+                }
+                wrapped = wrapped instanceof Request.Wrapper w ? w.getWrapped() : null;
+            }
+
+            // We need a new instance of the wrapper
+            Request wrapper = wrap(request);
+            process(wrap(wrapper, response), callback);
```
sbordet: I would rewrite this method in a way that it is evident that it cares about reuse.
Something like:

```java
boolean notReused = unwrapUntil(wrapped -> wrapped == request);
Request wrappedRequest = notReused ? this : wrap(request);
Response wrappedResponse = wrap(wrappedRequest, response);
process(wrappedRequest, wrappedResponse, callback);
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java
```diff
@@ -111,14 +111,17 @@
...
+     * </p>
      * <p>Otherwise, the HTTP request is not handled by this {@code Handler}
      * (for example, the HTTP request's URI does not match those handled
-     * by this {@code Handler}), and this method must return {@code null}.</p>
-     * <p>This method may inspect the HTTP request with the following rules:</p>
-     * <ul>
+     * by this {@code Handler}), and this method must return {@code null}.
+     * </p>
+     * <p>This method may inspect the HTTP request with the following rules:
+     * </p><ul>
```
sbordet, line 124: All closing `</p>` are now messed up in a new line?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -658,68 +677,114 @@ static long getContentBytesRead(Request request)
...
+     * <p>A {@link Request.Processor} that wraps another {@link Request.Processor} and supports
+     * use of {@link Request.Wrapper}s and {@link Response.Wrapper}s.
+     * The typed {@link #process(Request, Response, Callback, Processor)} method is called from
+     * {@link #process(Request, Response, Callback)} with the wrapped request and an optionally
+     * wrapped response. </p>
+     * @param <REQ> The type of the optionally wrapped request.
+     * @param <RES> The type of the optionally wrapped response.
      */
-    class WrapperProcessor extends Wrapper implements Processor
+    class ReWrappingProcessor<REQ extends Request, RES extends Response> implements Request.Processor
```
sbordet, line 705: I'd rename to `Rewrapping` (lowercase `w`).

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -1240,4 +1216,55 @@ public String toString()
...
+                return contextResponse;
+            return contextRequest.wrap(response);
+        }
+
+        @Override
+        protected void process(ContextRequest contextRequest, ContextResponse contextResponse, Callback callback, Request.Processor next) throws Exception
+        {
+            Context lastContext = __context.get();
+            if (lastContext == _context)
+                _processor.process(contextRequest, contextResponse, callback);
```
sbordet, line 1251: Surround with braces.
Also, why you do not need to `catch` like you do below?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
```diff
@@ -57,42 +57,33 @@ protected Request.Processor delayed(Request request, Request.Processor processor
...
-            }
+                try
+                {
+                    _processor.process(request, response, callback);
+                }
+                catch (Throwable t)
+                {
+                    Response.writeError(request, response, callback, t);
+                }
+            });
```
sbordet, line 86: Ugh. We went from "does not allocate a lambda, but does not work if reused concurrently" with "maybe rarely we reuse this processor, but we always allocate the lambda".
@lorban we need to benchmark this.

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/Rule.java
```diff
@@ -55,4 +58,101 @@ public String toString()
...
+         * @param processor the {@code Processor} to wrap
+         * @return this instance
+         */
+        public Processor wrapProcessor(Processor processor)
+        {
+            _processor = processor;
+            return processor == null ? null : this::wrappedProcess;
+        }
+
+        protected Request wrap(Request request)
```
lorban: Nit: rename to `wrapRequest()` maybe?

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/Rule.java
```diff
@@ -55,4 +58,101 @@ public String toString()
...
+        {
+            return request;
+        }
+
+        protected Response wrap(Request request, Response response)
```
lorban: Nit: rename to `wrapResponse()` maybe?

jetty-ee10/test-ee10-sessions/test-ee10-sessions-common/src/test/java/org/eclipse/jetty/ee10/session/CreationTest.java
```diff
@@ -285,7 +285,8 @@ public void testSessionCreateInvalidateCreate() throws Exception
             ContentResponse response = client.GET(url);
             assertEquals(HttpServletResponse.SC_OK, response.getStatus());
 
-            //check that the session does not exist
+            //check that the session exists
+            System.err.println("servlet._id " + servlet._id);
```
lorban: Please remove that `System.err.println()`, or make it a log.

lorban (approved):
A  few nits, otherwise LGTM.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
```diff
@@ -154,54 +145,61 @@ protected Request.Processor delayed(Request request, Request.Processor processor
...
-
-            String contentTypeValue = HttpField.valueParameters(contentType, null);
-            if (!MimeTypes.Type.MULTIPART_FORM_DATA.is(contentTypeValue))
-                return processor;
-            String boundary = MultiPart.extractBoundary(contentType);
-            if (boundary == null)
-                return processor;
-
-            return new Processor(request, processor, boundary);
+            return new DelayedMultiPartProcessor(processor);
```
sbordet, line 148: I think the previous logic should be reinstated here, rather than allocating `DelayedMultiPartProcessor`, only to not use it if the logic does not apply (e.g. content type is not multipart, there is no boundary, etc.).

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ProxiedRequestHandler.java
```diff
@@ -57,21 +57,43 @@ public HostPort getServerAuthority()
...
+            return null;
+
+        return new Request.ReWrappingProcessor<>(processor, wrapper)
+        {
             @Override
-            public ConnectionMetaData getConnectionMetaData()
+            protected Request.Wrapper wrap(Request request)
             {
-                return proxiedFor;
+                return new ProxiedWrapper(request, proxiedFor);
```
sbordet, line 71: This anonymous class captures `proxiedFor`.
However, in the context of reusability, it may be that the `ConnectionMetaData` is different (e.g. request coming on different connection), but this processor won't "refresh" it.
It starts to becoming more complicated to try to write a correct reusable processor than to just decide that they are non-reusable.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/StatisticsHandler.java
sbordet: I agree with this, but I think an improved API on `Request.WrapperProcessor` is enough to cover these confusing cases.
Like you have done, a `protected void process(Response res, Callback cbk, Request.Processor next)` should be enough to solve this particular issue.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/TryPathsHandler.java
```diff
@@ -184,4 +188,36 @@ public HttpURI getHttpURI()
...
+            _interpolated = interpolated;
+            _wrapper = wrapper;
+        }
+
+        @Override
+        public void process(Request request, Response response, Callback callback) throws Exception
+        {
+            TryPathsRequest wrapper = _wrapper;
+            if (wrapper != null && request == wrapper.getWrapped())
+                _wrapper = null;
```
sbordet, line 210: I think it needs to be an `AtomicReference.getAndSet(null)`, otherwise when used concurrently it's subject to races.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipHandler.java
```diff
@@ -528,119 +530,33 @@ public Request.Processor handle(Request request) throws Exception
...
-        if (alreadyGzipped)
-        {
-            LOG.debug("{} already intercepting {}", this, request);
-            HeaderWrappingRequest wrappedRequest = new HeaderWrappingRequest(request, newFields);
-            return wrappedRequest.wrapProcessor(super.handle(wrappedRequest));
-        }
+        if (!isPathInflatable(path) && !(isPathGzipable(path) && isPathMimeTypeGzipable(path)))
+            // Bypass wrapping
+            return getHandler().handle(request);
+        return super.handle(request);
```
sbordet, line 536: Another case of "allocate the wrapper first, then decide whether to use the wrapper at all" on the hot path, for sake of possible reuse of the processor (now the logic has to be run elsewhere in case the request is different) -- I don't like it much, but let's see what benchmark says.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -1301,7 +1302,7 @@ private MetaData.Response lockedPrepareResponse(HttpChannelState httpChannel, bo
         }
     }
 
-    private static class ChannelCallback implements Callback
+    static class ChannelCallback implements Callback
```
sbordet, line 1305: Should remain `private`.

sbordet (changes requested):
Overall I get the feeling that an explicit "processors cannot be reused" is simpler for implementers, less surprising (code goes where it is expected to be) and equivalent or better in terms of allocation (but needs confirmation).
However, I like the ideas of simpler APIs and simplifications, so my take would be to give up reusability but keep the simplifications.

gregw:
> Overall I get the feeling that an explicit "processors cannot be reused" is simpler for implementers, less surprising (code goes where it is expected to be) and equivalent or better in terms of allocation (but needs confirmation). 

I'm the complete opposite. I think processors that can sometimes only be used once is much more surprising.  Specially those that fail to be reused because they have held onto the request passed to handle.   If we want to not be reusable, then we shouldn't pass the request again into Processor, as that is what results in the confusion - two sources of the request that are mostly the same, but could be different depending on how we get there. 

> However, I like the ideas of simpler APIs and simplifications, so my take would be to give up reusability but keep the simplifications.

Some of the simplifications can be achieved without the reusability, but many of them are the result of changing to not hold onto the request from handle and respecting the request passed into process.  

I'll try to factor some of them out into different, less controversial PRs and see what we get left here.... stand by....

gregw:
I'm closing this PR.   I think there is some goodness in it, but I'll harvest that in other smaller PRs.    I also think there are bigger issues that it is trying to solve, but we need to discuss in #8929

PR #8813, opened 2022-11-01
jetty 10.0.x use default container

olamy:
- do not switch container
- debug
- remove debug


joakime (approved):
happy to see use of the default container

olamy:
> happy to see use of the default container



> happy to see use of the default container

well yes and no :) 
as this make the image more complicated to maintain as mixing jenkins agent image and our need.
see https://github.com/webtide/jenkins-slave/blob/master/slave-image/Dockerfile
but this will avoid having 2 containers running in the same pod and forcing Jenkins to maintain some connections between both..

PR #8856, opened 2022-11-02
Jetty 12 rehandle welcome files

sbordet:
Introduced WelcomeMode so that now welcome files can be redirected, served, or rehandled.

Updated ResourceHandler and DefaultServlet accordingly.


jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -467,62 +463,127 @@ protected void sendWelcome(HttpContent content, String pathInContext, boolean en
...
     }
 
-    public enum WelcomeActionType
+    /**
+     * <p>How welcome paths should be processed.</p>
+     */
+    public enum WelcomeMode
     {
+        /**
+         * The welcome path is used in an HTTP redirect
```
joakime: We use "welcome path", and "welcome target", and "welcome target path", and "welcome target URI path" in different places.

We should settle on one way of saying this.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ResourceHandler.java
```diff
@@ -335,8 +330,34 @@ public void setWelcomeFiles(List<String> welcomeFiles)
...
+                @Override
+                public HttpURI getHttpURI()
+                {
+                    return newHttpURI;
+                }
+            };
+
+            Request.Processor processor = getServer().handle(newRequest);
+            if (processor == null)
+                return false;
```
joakime: Don't see a testcase for this code path (where we return null on the processor)

joakime (changes requested):
missing a testcase at least.
might reveal an odd behavior if we return null/false on the rehandle flows

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
gregw: I'd like to keep "welcome file", to emphasis that this mechanism is only for files that exist within a directory and is not for any virtual resources.

gregw: Or perhaps "welcomeResource"?  But either way, the name that we have which results in:
```java
 Resource welcomePath = ...
```
is wrong.  Resources are not `Path`s. They may have 0, 1 or many `Path`s

sbordet: We never use `Resource` so I am sticking with "welcome target", also because this is how it is named in `WelcomeFactory`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ResourceHandler.java
sbordet: Test case added.

joakime (approved):

joakime (approved):
LGTM

joakime (approved):

gregw (changes requested):
Can you update after lachlans PR merge and then I'll review

janbartel:
@sbordet nudge

janbartel (commented):
It would be good to do the same conversion to enum to the `_welcomeExactServlets` and `_welcomeServlets` variables.

sbordet:
@gregw I merged @lachlan-roberts work, and updated the code with @janbartel suggestion for `_welcomeServlets`.
Please re-review if you can.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -910,109 +914,117 @@ private class ServletResourceService extends ResourceService implements Resource
...
+                    if (_welcomeServletMode != WelcomeServletMode.NONE && welcomeTarget == null)
                     {
                         ServletHandler.MappedServlet entry = _servletContextHandler.getServletHandler().getMappedServlet(welcomeInContext);
-                        if (entry != null && entry.getServletHolder().getServletInstance() != DefaultServlet.this &&
-                            (_welcomeServlets || (_welcomeExactServlets && entry.getPathSpec().getDeclaration().equals(welcomeInContext))))
-                            welcomeServlet = welcomeInContext;
+                        // Is there a different Servlet that may serve the welcome resource?
+                        if (entry != null && entry.getServletHolder().getServletInstance() != DefaultServlet.this)
+                        {
+                            if (_welcomeServletMode == WelcomeServletMode.MATCH || entry.getPathSpec().getDeclaration().equals(welcomeInContext))
```
joakime, line 1058: I see handling for WelcomeServletMode.NONE and .MATCH, but not .EXACT.
Do we have a testcase for that scenario?

joakime (commented):
Having a hard time following the .EXACT mode handling.

joakime (approved):

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
sbordet: Yes, there are a couple of tests, see `DefaultServletTest.testWelcomeExactServlet()`.

gregw:
@sbordet can you resolve the conflicts in this one.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -469,66 +466,130 @@ protected void sendWelcome(HttpContent content, String pathInContext, boolean en
...
+     * @param request the request
+     * @param response the response
+     * @param callback the callback to complete
+     * @param welcomeTarget the welcome target to serve
+     * @return true when the welcome target will be served and the callback completed,
+     * false otherwise
+     * @throws Exception if serving the welcome target fails
+     */
+    protected boolean serveWelcome(Request request, Response response, Callback callback, String welcomeTarget) throws Exception
+    {
```
gregw, line 555: Should this method check the conditional headers as per the TODO above?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -469,66 +466,130 @@ protected void sendWelcome(HttpContent content, String pathInContext, boolean en
...
     {
+        /**
+         * The welcome target is used in an HTTP redirect
+         * as the value of the {@code Location} header.
+         */
         REDIRECT,
-        SERVE
+        /**
+         * The welcome target is served either as a real file
+         * or forwarded to be processed again.
```
gregw: what does "or forwarded to be processed again" mean?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -469,66 +466,130 @@ protected void sendWelcome(HttpContent content, String pathInContext, boolean en
...
         if (welcomeAction == null)
             return false;
 
-        handleWelcomeAction(request, response, callback, welcomeAction);
-        return true;
+        return handleWelcomeAction(request, response, callback, welcomeAction);
     }
 
-    protected void handleWelcomeAction(Request request, Response response, Callback callback, WelcomeAction welcomeAction) throws IOException
+    protected boolean handleWelcomeAction(Request request, Response response, Callback callback, WelcomeAction welcomeAction) throws Exception
```
gregw: It is a bit unclear what the boolean return means.  If a welcome file has been found but this returns false, then do we try to serve the directory or do a 404?

i.e. Is there a difference between a welcome file being found but not being able to be served vs not finding a welcome file?

gregw (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
sbordet: Perhaps, but in another PR.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
sbordet: `ServletResourceService` overrides `serveWelcome()` to forward the welcome file, so it is re-processed as it goes again through the filters, etc.
I'm fine improving the text, you have suggestions?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
sbordet: The return value is used in `sendWelcome()`: if `false` the directory is served.

The return value may be `false` in case of re-handling, where the welcome URI goes again through the handlers, and none handles it.

gregw: I understand how the boolean return is used.  What is unclear is the two different meanings of a false return from serveWelcome: 1) No welcome file found; 2) a welcome file was found but could not be served.

If a welcome file is found but cannot be served (for some unknown reason), is it correct that a directory is served?   I think perhaps not.  Maybe a 500 would be more appropriate?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -469,66 +466,130 @@ protected void sendWelcome(HttpContent content, String pathInContext, boolean en
...
-    public enum WelcomeActionType
+    /**
+     * <p>How welcome targets should be processed.</p>
+     */
+    public enum WelcomeMode
     {
+        /**
+         * The welcome target is used in an HTTP redirect
+         * as the value of the {@code Location} header.
+         */
```
gregw:
```suggestion
        /**
         * The welcome target is used as the location for a redirect response, sent
         * by the {@link #redirectWelcome(Request, Response, Callback, String)} method.
         */
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
gregw:
```suggestion
         * The welcome target is served by  the 
         * {@link #serveWelcome(Request, Response, Callback, String)} method.
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -469,66 +466,130 @@ protected void sendWelcome(HttpContent content, String pathInContext, boolean en
...
         REDIRECT,
-        SERVE
+        /**
+         * The welcome target is served either as a real file
+         * or forwarded to be processed again.
+         */
+        SERVE,
+        /**
+         * The welcome target is wrapped in a new request
+         * that is re-handled.
```
gregw:
```suggestion
         * The welcome target is re-handledre-handled by the
         * {@link #rehandleWelcome(Request, Response, Callback, String)} method.
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
gregw: Can you open an issue for that.  Serving /index.html is a very frequent target and will often have conditional headers.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -85,14 +86,12 @@
 public class DefaultServlet extends HttpServlet
```
janbartel, line 187: I would like to see the javadoc restored here asap. As you're editing the class in this PR, I think now is the time to do it.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -186,19 +185,25 @@ public void init() throws ServletException
 
         _resourceService.setAcceptRanges(getInitBoolean("acceptRanges", _resourceService.isAcceptRanges()));
         _resourceService.setDirAllowed(getInitBoolean("dirAllowed", _resourceService.isDirAllowed()));
-        _resourceService.setRedirectWelcome(getInitBoolean("redirectWelcome", _resourceService.isRedirectWelcome()));
+        boolean redirectWelcome = getInitBoolean("redirectWelcome", false);
+        _resourceService.setWelcomeMode(redirectWelcome ? ResourceService.WelcomeMode.REDIRECT : ResourceService.WelcomeMode.SERVE);
```
janbartel, line 290: So how would I set `WelcomeMode.REHANDLE`? Or is this inappropriate in the servlet environment - if so, a javadoc to that effect would be good.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -1155,4 +1167,27 @@ public String getCharacterEncoding()
...
+         * <p>The welcome resource must exist as a file on the filesystem.</p>
+         */
+        NONE,
+        /**
+         * <p>Welcome resources may be served by Servlets, but only if the
+         * welcome resource does not exist as a file on the filesystem.</p>
+         */
+        MATCH,
+        /**
+         * <p>Welcome resources may be served by Servlets, but only if the
```
janbartel: I don't think that's correct. If it's `EXACT` we don't check for the existence of the file on disk, only that there is an exact servlet mapping. The text as it is implies that we will check for the file and it must not exist.

janbartel (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
sbordet: It has the same semantic of `Handler.handle()`, as specified in the javadocs of `serveWelcome()`, i.e. response sent + callback completed -> true, otherwise false.
For Servlets, we may return false in case we can't find a `RequestDispatcher` for the welcome target.
In Jetty 11, in the same case we would just return (which I think it's wrong as we do not modify the response, so I think it defaults to 200 with empty body).
In 12, if we want to change the behavior and return 500, fine. Want me to do it in this PR?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
sbordet: #9632.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
gregw: Yeah. Sending a 500 is a one liner

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
sbordet: @janbartel the code in ee9's `DefaultServlet.getWelcomeFile()` and in ee10's `ServletResourceService.getWelcomeTarget()` does check for file existence, so the javadoc reports the current code behavior.
Perhaps the code is wrong, then?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
sbordet: No `REHANDLE` in Servlets, it's equivalent to `SERVE`.
As this parameter is, and always was, a `boolean`, I don't think there is need to mention `REHANDLE` which is not applicable to Servlets anyway.
To be clear, I don't think it's worth mentioning a 3rd mode, just to say that you cannot use it.
Plus the parameter name would need to be changed, as it contains "redirect".

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -469,66 +466,130 @@ protected void sendWelcome(HttpContent content, String pathInContext, boolean en
...
+     * @param response the response
+     * @param callback the callback to complete
+     * @param welcomeTarget the welcome target to rehandle to
+     * @return true when the welcome target will be rehandled and the callback completed,
+     * false otherwise
+     * @throws Exception if the rehandling fails
+     */
+    protected boolean rehandleWelcome(Request request, Response response, Callback callback, String welcomeTarget) throws Exception
+    {
+        return false;
```
gregw: send a 500 here.
make it a void method and extended implementations should send a 500 if the handlers return false.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -469,66 +466,130 @@ protected void sendWelcome(HttpContent content, String pathInContext, boolean en
...
         if (welcomeAction == null)
             return false;
 
-        handleWelcomeAction(request, response, callback, welcomeAction);
-        return true;
+        return handleWelcomeAction(request, response, callback, welcomeAction);
     }
 
-    protected void handleWelcomeAction(Request request, Response response, Callback callback, WelcomeAction welcomeAction) throws IOException
+    protected boolean handleWelcomeAction(Request request, Response response, Callback callback, WelcomeAction welcomeAction) throws Exception
```
gregw: I think this should be a void method.  A welcome file has been found and there is no option but to serve it somehow.  If the configuration of the server does not allow it to be served, then it is a 500.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -469,66 +466,130 @@ protected void sendWelcome(HttpContent content, String pathInContext, boolean en
...
+     *
+     * @param request the request to redirect
+     * @param response the response
+     * @param callback the callback to complete
+     * @param welcomeTarget the welcome target to redirect to
+     * @return true when the request will be redirected and the callback completed,
+     * false otherwise
+     * @throws Exception if the redirection fails
+     */
+    protected boolean redirectWelcome(Request request, Response response, Callback callback, String welcomeTarget) throws Exception
```
gregw: void method

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -469,66 +466,130 @@ protected void sendWelcome(HttpContent content, String pathInContext, boolean en
...
+     *
+     * @param request the request
+     * @param response the response
+     * @param callback the callback to complete
+     * @param welcomeTarget the welcome target to serve
+     * @return true when the welcome target will be served and the callback completed,
+     * false otherwise
+     * @throws Exception if serving the welcome target fails
+     */
+    protected boolean serveWelcome(Request request, Response response, Callback callback, String welcomeTarget) throws Exception
```
gregw: void method

gregw (changes requested):

gregw (approved):

PR #8859, opened 2022-11-03
Fixes #8858 - Jetty 12 Review MovedContextHandler.

### LINKED ISSUE #8858: Jetty 12 - Review MovedContextHandler
**Jetty version(s)**
12+

**Description**
`MovedContextHandler` is not tested and it's currently broken.
Methods are still named after Servlets, but should be renamed to a more generic naming.

sbordet:
Updated method names to avoid references to Servlet concepts. Added test cases.
Updated XML files.

### DIFF
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/MovedContextHandler.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/MovedContextHandler.java
@@ -23,74 +23,126 @@ import org.eclipse.jetty.util.Callback;
 import org.eclipse.jetty.util.URIUtil;
 
 /**
- * Moved ContextHandler.
- * This context can be used to replace a context that has changed
- * location.  Requests are redirected (either to a fixed URL or to a
- * new context base).
+ * <p>A {@code ContextHandler} with a child {@code Handler}
+ * that redirects to a configurable URI.</p>
  */
 public class MovedContextHandler extends ContextHandler
 {
-    final Redirector _redirector;
-    String _newContextURL;
-    boolean _discardPathInfo;
-    boolean _discardQuery;
-    boolean _permanent;
-    String _expires;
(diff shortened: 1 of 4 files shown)


jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/MovedContextHandler.java
```diff
@@ -23,122 +23,168 @@
...
 
-    /**
-     * @param expires the expires header value or null if no expires header
-     */
-    public void setExpires(String expires)
-    {
-        _expires = expires;
+            String cacheControl = getCacheControl();
+            if (cacheControl != null)
+                response.getHeaders().put(HttpHeader.CACHE_CONTROL, cacheControl);
```
gregw: If we wanted to optimise, then when the cache control is set, we could construct a `PreEncodedHttpField` and then put that rather than build a new HttpField on every request.

gregw (approved):
Other than a minor performance niggle...

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/MovedContextHandler.java
```diff
@@ -13,132 +13,180 @@
...
  */
 public class MovedContextHandler extends ContextHandler
 {
-    final Redirector _redirector;
-    String _newContextURL;
-    boolean _discardPathInfo;
-    boolean _discardQuery;
-    boolean _permanent;
-    String _expires;
+    private int _statusCode = HttpStatus.SEE_OTHER_303;
```
joakime, line 33: Should document this default value in the javadoc somewhere.

In `getStatusCode()` perhaps?

joakime (changes requested):
Looks good, just want to see some javadoc telling folks about the default statuscode.

joakime (approved):

PR #8861, opened 2022-11-03
Jetty 12 - General cleanup of URIUtil

joakime:
+ Removal of __CHARSET
+ Removal of unused methods
+ Using new switch/case concepts

### DIFF
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/TypeUtil.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/TypeUtil.java
@@ -518,21 +518,6 @@ public class TypeUtil
         return true;
     }
 
-    public static void toHex(char c, Appendable buf)
-    {
-        try
-        {
-            int d = 0xf & ((0xF0 & c) >> 4);
-            buf.append((char)((d > 9 ? ('A' - 10) : '0') + d));
-            d = 0xf & c;
-            buf.append((char)((d > 9 ? ('A' - 10) : '0') + d));
-        }
-        catch (IOException e)
-        {
-            throw new RuntimeException(e);
-        }
-    }
-
     public static void toHex(byte b, Appendable buf)
     {
         try
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/TypeUtil.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/TypeUtil.java
@@ -518,6 +518,21 @@ public class TypeUtil
         return true;
     }
 
+    public static void toHex(char c, Appendable buf)
+    {
+        try
+        {
+            int d = 0xf & ((0xF0 & c) >> 4);
+            buf.append((char)((d > 9 ? ('A' - 10) : '0') + d));
+            d = 0xf & c;
+            buf.append((char)((d > 9 ? ('A' - 10) : '0') + d));
+        }
+        catch (IOException e)
+        {
+            throw new RuntimeException(e);
+        }
+    }
+
     public static void toHex(byte b, Appendable buf)
     {
         try
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
@@ -54,14 +54,10 @@ public final class URIUtil
         .with("jar:")
         .build();
 
-    public static final String SLASH = "/";
-    public static final String HTTP = "http";
-    public static final String HTTPS = "https";
-
     /**
      * The characters that are supported by the URI class and that can be decoded by {@link #canonicalPath(String)}
      */
-    public static final boolean[] __uriSupportedCharacters = new boolean[]
+    private static final boolean[] URI_SUPPORTED_CHARACTERS = new boolean[]
     {
         false, // 0x00 is illegal
         false, // 0x01 is illegal
@@ -680,7 +676,7 @@ public final class URIUtil
     {
         // Allow any 8-bit character (as it's likely unicode).
         // or any character labeled with true in __uriSupportedCharacters static
-        return (code >= __uriSupportedCharacters.length || __uriSupportedCharacters[code]);
+        return (code >= URI_SUPPORTED_CHARACTERS.length || URI_SUPPORTED_CHARACTERS[code]);
     }
 
     /**
@@ -969,7 +965,7 @@ public final class URIUtil
 
         if (buf.charAt(split - 1) == '/')
         {
-            if (p2.startsWith(URIUtil.SLASH))
+            if (p2.startsWith("/"))
             {
                 buf.deleteCharAt(split - 1);
                 buf.insert(split - 1, p2);
@@ -979,7 +975,7 @@ public final class URIUtil
         }
         else
         {
-            if (p2.startsWith(URIUtil.SLASH))
+            if (p2.startsWith("/"))
                 buf.insert(split, p2);
             else
             {
@@ -1011,8 +1007,8 @@ public final class URIUtil
         if (p2 == null || p2.length() == 0)
             return p1;
 
-        boolean p1EndsWithSlash = p1.endsWith(SLASH);
-        boolean p2StartsWithSlash = p2.startsWith(SLASH);
+        boolean p1EndsWithSlash = p1.endsWith("/");
+        boolean p2StartsWithSlash = p2.startsWith("/");
 
         if (p1EndsWithSlash && p2StartsWithSlash)
         {
@@ -1025,15 +1021,15 @@ public final class URIUtil
         StringBuilder buf = new StringBuilder(p1.length() + p2.length() + 2);
         buf.append(p1);
 
-        if (p1.endsWith(SLASH))
+        if (p1.endsWith("/"))
         {
-            if (p2.startsWith(SLASH))
+            if (p2.startsWith("/"))
                 buf.setLength(buf.length() - 1);
         }
         else
         {
-            if (!p2.startsWith(SLASH))
-                buf.append(SLASH);
+            if (!p2.startsWith("/"))
+                buf.append("/");
         }
         buf.append(p2);
 
@@ -1099,7 +1095,7 @@ public final class URIUtil
      */
     public static String parentPath(String p)
     {
-        if (p == null || URIUtil.SLASH.equals(p))
+        if (p == null || "/".equals(p))
             return null;
         int slash = p.lastIndexOf('/', p.length() - 2);
         if (slash >= 0)
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
@@ -206,180 +206,155 @@ public final class URIUtil
         if (path == null || path.length() == 0)
             return path;
 
-        StringBuilder buf = encodePath(null, path);
-        return buf == null ? path : buf.toString();
-    }
-
-    /**
-     * Encode a URI path.
-     *
-     * @param path The path to encode
-     * @param buf StringBuilder to encode path into (or null)
-     * @return The StringBuilder or null if no substitutions required.
-     */
-    public static StringBuilder encodePath(StringBuilder buf, String path)
-    {
+        StringBuilder buf = null;
         byte[] bytes = null;
-        if (buf == null)
-        {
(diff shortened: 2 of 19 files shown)


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -59,9 +58,6 @@
     public static final String HTTP = "http";
     public static final String HTTPS = "https";
```
sbordet: Apparently, only SLASH is used, which is weird anyway (should be inlined everywhere it's used).
The other 2 are not used.
I would remove all 3.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -59,9 +58,6 @@
...
 
-    // Use UTF-8 as per http://www.w3.org/TR/html40/appendix/notes.html#non-ascii-chars
-    public static final Charset __CHARSET = StandardCharsets.UTF_8;
-
     /**
      * The characters that are supported by the URI class and that can be decoded by {@link #canonicalPath(String)}
      */
```
sbordet, line 59: Array below should be `private` with upper case name.

sbordet (changes requested):

gregw (approved):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -206,233 +198,163 @@ private URIUtil()
...
-     * Encode a raw URI String and convert any raw spaces to
-     * their "%20" equivalent.
-     *
-     * @param str input raw string
-     * @return output with spaces converted to "%20"
-     */
-    public static String encodeSpaces(String str)
-    {
-        return StringUtil.replace(str, " ", "%20");
+        return buf == null ? path : buf.toString();
```
sbordet: Wow the logic of this method is really hard to follow, and I think unnecessary.

First, we loop once over the path; if we find a reserved character, we exit immediately, to enter the second loop where we restart from index 0; in the second loop, if we find a CTL or non-ASCII character we convert the path to UTF-8 bytes for the second time, exit and enter a 3rd loop where we restart from index 0.

I understand we want to optimize the common path with no allocations (so just one loop over all the chars).
But the less common case should be written better, iterating only once.

The new algorithm should be something like:
iterate over the chars; if a special is found, allocate a `StringBuilder`, copy all the non-reserved chars so far, and then append the % representation of the reserved char. If we find a CTL or non-ASCII, we keep what we have so far, convert to UTF-8 bytes the rest, and continue the loop until the end.

sbordet (changes requested):
Also all the javadocs `<code>` should be converted to `{@code}`

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: I simplified encodePath() a great deal.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -197,242 +189,92 @@
...
+     * the '{@code /}' character is not encoded.
      *
-     * @param path The path the encode
+     * @param path The path to encode
      * @return The encoded path
      */
     public static String encodePath(String path)
     {
         if (path == null || path.length() == 0)
             return path;
```
sbordet, line 221: Use `StringUtil.isEmpty(path)`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -197,242 +189,92 @@
...
+    static
+    {
+        ENCODE_PATH_NEEDS_ENCODING = new boolean[128];
+        // Special characters
+        for (char c: "%?;#\"'<> [\\]^`{|}".toCharArray())
+            ENCODE_PATH_NEEDS_ENCODING[c] = true;
+        // control characters
+        ENCODE_PATH_NEEDS_ENCODING[0x7f] = true;
+        for (int i = 0; i < 0x20; i++)
+            ENCODE_PATH_NEEDS_ENCODING[i] = true;
```
sbordet, line 207: I believe SPACE 0x20 should be encoded too.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -197,242 +189,92 @@
...
-                            bytes = path.getBytes(URIUtil.__CHARSET);
-                            buf = new StringBuilder(path.length() * 2);
-                            break loop;
-                        }
-                }
-            }
-            if (buf == null)
-                return null;
+            char c = path.charAt(i);
+            if (c > 0x7F) // 8-bit +
```
sbordet, line 231: 0x7F should be encoded too, so `c >= 0x7F` here.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -197,242 +189,92 @@
...
+                needsEncoding = true;
         }
 
-        int i;
+        if (needsByteEncoding)
+            return encodePathBytes(path);
+        else if (needsEncoding)
+        {
+            return encodePathString(path);
+        }
```
sbordet: Drop the braces.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/TypeUtil.java
```diff
@@ -518,6 +518,21 @@ public static boolean isHex(String str, int offset, int len)
...
+            buf.append((char)((d > 9 ? ('A' - 10) : '0') + d));
+            d = 0xf & c;
+            buf.append((char)((d > 9 ? ('A' - 10) : '0') + d));
+        }
+        catch (IOException e)
+        {
+            throw new RuntimeException(e);
+        }
+    }
+
```
sbordet: No need for this method (see comment in URIUtil).

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -197,242 +189,92 @@
...
-
-                default:
-                    if (c < 0x20 || c >= 0x7f)
-                    {
-                        bytes = path.getBytes(URIUtil.__CHARSET);
-                        break loop;
-                    }
-                    buf.append(c);
+                buf.append('%');
+                TypeUtil.toHex(c, buf);
```
sbordet: You can safely cast `c` to a `byte` here, since this method is called only if all chars are ASCII, therefore bytes.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -197,242 +189,92 @@
...
-                            TypeUtil.toHex(c, buf);
-                        }
-                        else
-                            buf.append((char)c);
-                }
+                buf.append('%');
+                TypeUtil.toHex(b, buf);
             }
+            else
+                buf.append((char)b);
```
sbordet, line 284: Add braces.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -1582,32 +1428,29 @@ public static void appendSchemeHostPort(StringBuilder url, String scheme, String
...
+                    break;
+
+                default:
+                    url.append(':').append(port);
             }
         }
     }
 
-    // Only URIUtil is using this method
+    // Only used by URIUtil
```
sbordet: Make it private then.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -1622,6 +1465,7 @@ static boolean equalsIgnoreEncodings(String uriA, String uriB)
         }
     }
 
+    // Only used by URIUtil
```
sbordet: Make it private then.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/URIUtilTest.java
```diff
@@ -1045,7 +1021,7 @@ public void testEncodeDecodeVisibleOnly()
         {
             assertTrue(c > 0x20 && c < 0x80);
```
sbordet: Should not `c < 0x7F`, since `7F` is encoded and can never be present as a character.
Also, the comment 2 lines above has a typo: `endoded`.

sbordet (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: Already covered in the special chars above

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: already covered in ENCODE_PATH_NEEDS_ENCODING

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: Can't make it private, as it's tested in URIUtilTest.
And protected isn't valid.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: Can't make it private, as it's tested in URIUtilTest.
And protected isn't valid.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/URIUtilTest.java
joakime: fair enough, but the check for control characters is checked 2 lines below this, which would have tripped on the raw 0x7f usage as well.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: added testcase specific for this to show it's protected

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: added testcase specific for this to show it's protected

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
sbordet: Then the comment is wrong, should be `Only used by URIUtil*Test*.`

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
sbordet: Then the comment is wrong, should be `Only used by URIUtil*Test*.`

joakime: It's used by URIUtil as a sub-method, and that sub-method is tested in URIUtilTest, it's not exposed to others.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
sbordet: Doh, you're right.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: Went ahead and cleaned it out

sbordet (approved):

PR #8868, opened 2022-11-07
Jetty 10 - Improve test reliability

joakime:
Some more test reliability fixes

### DIFF
--- a/jetty-http2/http2-client/src/test/java/org/eclipse/jetty/http2/client/ConnectTimeoutTest.java
+++ b/jetty-http2/http2-client/src/test/java/org/eclipse/jetty/http2/client/ConnectTimeoutTest.java
@@ -37,6 +37,7 @@ public class ConnectTimeoutTest extends AbstractTest
     @Tag("external")
     public void testConnectTimeout() throws Exception
     {
+        // Using IANA hosted example.com:81 to reliably produce a Connect Timeout.
         final String host = "example.com";
         final int port = 81;
         int connectTimeout = 1000;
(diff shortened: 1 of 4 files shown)


jetty-http2/http2-client/src/test/java/org/eclipse/jetty/http2/client/ConnectTimeoutTest.java
```diff
@@ -50,6 +51,13 @@ public void testConnectTimeout() throws Exception
             @Override
             public void failed(Throwable x)
             {
+                if (x instanceof NoRouteToHostException)
+                {
+                    // Unable to connect to destination is not a failure.
+                    latch.countDown();
+                    return;
+                }
```
sbordet: @joakime I don't know what to say about this. 

We have already called `assumeConnectTimeout()` which verified that connecting to that address _indeed_ throws a `SocketTimeoutException`.

Few lines later we do it via `client.connect()` and we get a different exception?
We should really have `if (docker.screwsUpNetworking())`?

sbordet (changes requested):

jetty-http2/http2-client/src/test/java/org/eclipse/jetty/http2/client/ConnectTimeoutTest.java
joakime: Let me try a non-routable, local only, IP instead.

joakime: The 10.* network is how docker itself communicates among containers.

joakime: The original test is flaky, even on my local machine (no docker) this test will fail about 15% of the time.
For various reasons: "No Route to Host", "Host unreachable".
I'll use an IANA recommended host instead.

sbordet (approved):

PR #8873, opened 2022-11-08
Fixes #7117 - Timeout with Expect 100 continue when using ProxyServlet.

### LINKED ISSUE #7117: Timeout with Expect 100 continue when using ProxyServlet
**Jetty version**
9.4.43, 9.4.44

**Java version**
OpenJDK 11

**OS type/version**
Windows 10

**Description**
I am using embedded jetty version 9.4.43 and use ProxyServlet to reverse proxy requests to the server.
A servlet behind this proxy which reads all lines causes a timeout when Expect: 100-continue is set.
If servlet reads all bytes binary it works. Whithout reverse proxy the servlet works also.
This [Example](https://github.com/dejpec/dejpec/blob/main/jetty_problems/Expect100Timeout.java) shows the problem.

sbordet:
Now getReader() tests whether it has to send a 100 continue in case getInputStream() is not called.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
@@ -1161,16 +1161,13 @@ public class Request implements HttpServletRequest
 
         if (_reader == null || !encoding.equalsIgnoreCase(_readerEncoding))
         {
-            final ServletInputStream in = getInputStream();
+            ServletInputStream in = getInputStream();
             _readerEncoding = encoding;
-            _reader = new BufferedReader(new InputStreamReader(in, encoding))
-            {
-                @Override
-                public void close() throws IOException
-                {
-                    in.close();
-                }
-            };
+            _reader = new BufferedReader(new InputStreamReader(in, encoding));
+        }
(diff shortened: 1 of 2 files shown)


joakime (dismissed):
Surprised this one lingered in the code for this long

gregw (dismissed):
LGTM

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -1161,17 +1161,23 @@ public BufferedReader getReader() throws IOException
...
-            final ServletInputStream in = getInputStream();
+            ServletInputStream in = getInputStream();
             _readerEncoding = encoding;
             _reader = new BufferedReader(new InputStreamReader(in, encoding))
             {
                 @Override
                 public void close() throws IOException
                 {
+                    // Do not call super to avoid marking this reader as closed,
+                    // but do close the ServletInputStream that can be reopened.
```
joakime, line 1172: Ah.   Now this makes sense!

gregw: Ah good catch! (and good to comment it)

PR #8872, opened 2022-11-08
More reliable `AllowedResourceAliasCheckerTest`

joakime:
Signed-off-by: Joakim Erdfelt <joakim.erdfelt@gmail.com>

PR #8874, opened 2022-11-09
add some fixes to SizeLimitHandler

lachlan-roberts:

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/handler/SizeLimitHandler.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/handler/SizeLimitHandler.java
@@ -126,6 +126,9 @@ public class SizeLimitHandler extends HandlerWrapper
         @Override
         public HttpInput.Content readFrom(HttpInput.Content content)
         {
+            if (content == null)
+                return null;
+
             if (content.hasContent())
             {
                 _read += content.remaining();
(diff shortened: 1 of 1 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/handler/SizeLimitHandler.java
```diff
@@ -126,6 +126,9 @@ public boolean isOptimizedForDirectBuffers()
         @Override
         public HttpInput.Content readFrom(HttpInput.Content content)
         {
+            if (content == null)
+                return null;
```
lachlan-roberts, line 130: maybe this should also be resetting `_read` back to 0.

joakime:
Regarding the AllowedResourceAliasCheckTest failures, I'll need to backport (to `jetty-9.4.x`) the fixes from PR #8872

joakime:
Opened #8875 for issue in `AllowedResourceAliasCheckTest` (it's just a backport of a previous stability fix from `jetty-10.0.x`)

jetty-server/src/main/java/org/eclipse/jetty/server/handler/SizeLimitHandler.java
gregw: I don't think so, as reading null just means there is nothing currently available, but there may be after a demand.

gregw (approved):

PR #8888, opened 2022-11-11
Issue #8886 - support extensible Resource URI schemes

### LINKED ISSUE #8886: Changes to Resource no longer support custom url schemes.
**jetty-12**

In previous versions of jetty, a call like `Resource.newResource("bundleresource://62.fwk1116834242/org/eclipse/jetty/webapp/webdefault.xml")` would return a `URLResource` that could be used to eg open an input stream and read the xml content. 

In jetty-12, we get the following exception instead:

> Found defaults descriptor via WebXmlConfiguration.classloader: bundleresource://62.fwk1116834242/org/eclipse/jetty/ee9/webapp/webdefault-ee9.xml
> 2022-11-11 13:53:43.834:WARN :oejew.WebAppContext:Start Level: Equinox Container: ac1d208f-4617-41a3-b6a8-a7a1b3d9eea7: Failed startup of context o.e.j.e.w.WebAppContext@7a8e2a3c{ROOT,/,jar:file:///home/janb/src/jetty-eclipse/jetty-12.0.x/jetty-ee9/jetty-ee9-osgi/test-jetty-ee9-osgi/target/1668135220776-0/org.eclipse.osgi/89/0/bundleFile!/,false}{jar:file:/home/janb/src/jetty-eclipse/jetty-12.0.x/jetty-ee9/jetty-ee9-osgi/test-jetty-ee9-osgi/target/1668135220776-0/org.eclipse.osgi/89/0/bundleFile!/}
> java.nio.file.FileSystemNotFoundException: Provider "bundleresource" not installed
> 	at java.base/java.nio.file.Path.of(Path.java:212)
> 	at java.base/java.nio.file.Paths.get(Paths.java:98)
> 	at org.eclipse.jetty.util.resource.MountedPathResource.of(MountedPathResource.java:33)
> 	at org.eclipse.jetty.util.resource.Resource.create(Resource.java:91)
> 	at org.eclipse.jetty.util.resource.ResourceFactoryInternals$LifeCycle.newResource(ResourceFactoryInternals.java:113)
> 	at org.eclipse.jetty.ee9.webapp.WebXmlConfiguration.preConfigure(WebXmlConfiguration.java:67)
> 	at org.eclipse.jetty.ee9.webapp.Configurations.preConfigure(Configurations.java:498)
> 	at org.eclipse.jetty.ee9.webapp.WebAppContext.preConfigure(WebAppContext.java:499)
> 	at org.eclipse.jetty.ee9.webapp.WebAppContext.doStart(WebAppContext.java:528)
> 	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:93)
> 	at org.eclipse.jetty.util.component.ContainerLifeCycle.start(ContainerLifeCycle.java:171)
> 	at org.eclipse.jetty.util.component.ContainerLifeCycle.doStart(ContainerLifeCycle.java:114)
> 	at org.eclipse.jetty.server.Handler$Abstract.doStart(Handler.java:391)
> 	at org.eclipse.jetty.server.handler.ContextHandler.lambda$doStart$0(ContextHandler.java:535)
> 	at org.eclipse.jetty.server.handler.ContextHandler$Context.call(ContextHandler.java:1053)
> 	at org.eclipse.jetty.server.handler.ContextHandler.doStart(ContextHandler.java:535)
> 	at org.eclipse.jetty.ee9.nested.ContextHandler$CoreContextHandler.doStart(ContextHandler.java:2391)
> 	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:93)
> 	at org.eclipse.jetty.deploy.bindings.StandardStarter.processBinding(StandardStarter.java:40)
> 	at org.eclipse.jetty.deploy.AppLifeCycle.runBindings(AppLifeCycle.java:183)
> 	at org.eclipse.jetty.deploy.DeploymentManager.requestAppGoal(DeploymentManager.java:519)
> 	at org.eclipse.jetty.deploy.DeploymentManager.addApp(DeploymentManager.java:165)
> 	at org.eclipse.jetty.osgi.BundleWebAppProvider.bundleAdded(BundleWebAppProvider.java:202)
> 	at org.eclipse.jetty.osgi.BundleWebAppProvider$WebAppTracker.addingBundle(BundleWebAppProvider.java:72)
> 	at org.osgi.util.tracker.BundleTracker$Tracked.customizerAdding(BundleTracker.java:477)
> 	at org.osgi.util.tracker.BundleTracker$Tracked.customizerAdding(BundleTracker.java:422)
> 	at org.osgi.util.tracker.AbstractTracked.trackAdding(AbstractTracked.java:258)
> 	at org.osgi.util.tracker.AbstractTracked.track(AbstractTracked.java:231)
> 	at org.osgi.util.tracker.BundleTracker$Tracked.bundleChanged(BundleTracker.java:452)
> 	at org.eclipse.osgi.internal.framework.BundleContextImpl.dispatchEvent(BundleContextImpl.java:949)
> 	at org.eclipse.osgi.framework.eventmgr.EventManager.dispatchEvent(EventManager.java:234)
> 	at org.eclipse.osgi.framework.eventmgr.ListenerQueue.dispatchEventSynchronous(ListenerQueue.java:151)
> 	at org.eclipse.osgi.internal.framework.EquinoxEventPublisher.publishBundleEventPrivileged(EquinoxEventPublisher.java:229)
> 	at org.eclipse.osgi.internal.framework.EquinoxEventPublisher.publishBundleEvent(EquinoxEventPublisher.java:138)
> 	at org.eclipse.osgi.internal.framework.EquinoxEventPublisher.publishBundleEvent(EquinoxEventPublisher.java:130)
> 	at org.eclipse.osgi.internal.framework.EquinoxContainerAdaptor.publishModuleEvent(EquinoxContainerAdaptor.java:217)
> 	at org.eclipse.osgi.container.Module.publishEvent(Module.java:499)
> 	at org.eclipse.osgi.container.Module.start(Module.java:486)
> 	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel$2.run(ModuleContainer.java:1852)
> 	at org.eclipse.osgi.internal.framework.EquinoxContainerAdaptor$1$1.execute(EquinoxContainerAdaptor.java:136)
> 	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:1845)
> 	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:1788)
> 	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.doContainerStartLevel(ModuleContainer.java:1750)
> 	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.dispatchEvent(ModuleContainer.java:1672)
> 	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.dispatchEvent(ModuleContainer.java:1)
> 	at org.eclipse.osgi.framework.eventmgr.EventManager.dispatchEvent(EventManager.java:234)
> 	at org.eclipse.osgi.framework.eventmgr.EventManager$EventThread.run(EventManager.java:345)

joakime:
Initial experiment to see what Resource would look like with extensible URI scheme support.

This is to allow the eventual `BundleResource` implementation in the `jetty-osgi` packages that respond to a uri schemes like `bundleresource://62.fwk1116834242/org/eclipse/jetty/ee9/webapp/webdefault-ee9.xml`


jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/ResourceFactoryTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+public class ResourceFactoryTest
+{
+    @Test
+    public void testCustomUriScheme()
+    {
+        ResourceFactory.addResourceFactory("custom", new CustomResourceFactory());
+        Resource resource = ResourceFactory.root().newResource("custom://foo");
+        assertThat(resource.getURI(), is(URI.create("custom://foo")));
+        assertThat(resource.getName(), is("custom-impl"));
```
joakime, line 55: This shows how the custom uri scheme could be supported.

I see the bootstrap of `jetty-osgi` registering their `ResourceFactory` for the `bundlresource` scheme in it's initialization phase.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactory.java
```diff
@@ -236,6 +236,21 @@ default Resource newJarFileResource(URI uri)
         return newResource(URIUtil.toJarFileUri(uri));
     }
 
+    static void addResourceFactory(String scheme, ResourceFactory resource)
```
lorban: I would prefer this method to be called `register()`, and it should also have its sibling `unregister()`.

I'm also wondering if `ResourceFactory` shouldn't have a `Set<String> getSupportedSchemes()` method instead of passing the scheme as a separate parameter.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
```diff
@@ -30,6 +31,20 @@
...
+        .mutable()
+        .build();
+
+    static
+    {
+        // The default resource factories
+        RESOURCE_FACTORIES.put("jar", new MountedPathResourceFactory());
+        PathResourceFactory pathResourceFactory = new PathResourceFactory();
+        RESOURCE_FACTORIES.put("file", pathResourceFactory);
+        RESOURCE_FACTORIES.put("jrt", pathResourceFactory);
```
lorban, line 54: With this extra metadata, `mountIfNeeded()` could make mounting more generic. Here's some pseudocode to illustrate this:

```java
ResourceFactory rf = RESOURCE_FACTORIES.get(scheme);
if (rf instanceof MountedPathResourceFactory) // or instanceof NewInterfaceThatIsAMountingMarker maybe?
  return FileSystemPool.INSTANCE.mount(uri);
else
  return null;
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -176,7 +168,7 @@ public static boolean isSameName(Path pathA, Path pathB)
     {
         if (!uri.isAbsolute())
             throw new IllegalArgumentException("not an absolute uri: " + uri);
-        if (!bypassAllowedSchemeCheck && !ALLOWED_SCHEMES.contains(uri.getScheme()))
+        if (!bypassAllowedSchemeCheck && !SUPPORTED_SCHEMES.contains(uri.getScheme()))
```
joakime, line 171: I should review the need for `bypassAllowedSchemeCheck`, it probably doesn't have much purpose left

gregw (changes requested):
I like the general approach.    This would allow us to add back `MemoryResource` if the need arose in the future.

However, I'm wondering about the static nature of it?   Should there be a non static API so that individual resource factories can be configured?  Or are they free to use a non static impl anyway already?    Perhaps the static methods need to named to make it clear they are default mappings used by the default factory implementations?

I also think it would be good to have a fallback resource factory, that is tried if the scheme is not known.  By default this would be a factory that just returns null, but could be replaced by a factory that tried a `URL` to see if a handler was installed.

Since we know we have at least one usage of URL, we should probably still add a `UrlResource` impl, but only use it if there is a specific registration or a fallback registration.

joakime:
> Since we know we have at least one usage of URL, we should probably still add a `UrlResource` impl, but only use it if there is a specific registration or a fallback registration.

The OSGi layer's using URL stream handlers are going to have to change with the new JDK releases.
The URL stream handling is changing, significantly enough that the existing OSGI implementations must make changes to satisfy the new JDK reality.
(The OpenJDK folks are addressing the numerous StackOverflowException that the existing URL Stream handling causes, such as changing _when_ a URL scheme is registered, and no longer supporting lazy initialization)

Having a `UrlResource` serves no purpose in the general sense.
For our `jetty-osgi` layer, we would need 2 implementations of this new `ResourceFactory`.
 * `ClassicJdkBundleResourceFactory` - using the old school URL Stream Handling that OSGi provides on older JDKs
 * `ModernBundleResourceFactory` - that uses the OSGi bundle behaviors directly, not via the URL Stream Handlers. (this is one way that the OSGi implementations are researching, the other being the FileSystem implementation, both not using URL Stream Handlers).

The `jetty-osgi` bootstrapping would have to make a call on what `ResourceFactory` to register/use, presumably based on some combination on OSGi support level, and the JDK runtime version level.

gregw:
@joakime I don't understand. You say `UrlResource` "serves no purpose in the general sense", but then go onto say we need a resource factory that supports "old school URL Stream Handling"?  So it serves the purpose to support that needed implementation.

Note also that a `UrlResource` is entirely decoupled from the URL stream handling mechanism.   It may well be that stream handlers change in future JVMs, but the URL class will remain in all it's uglyness for sometime.   A `UrlResource` class will continue to work against the `URL` API even if the handler mechanism underneath is significantly changed.

We have a use-case for it and it is a blocker for releasing jetty-12, so we need to support a `UrlResource` at least as an option.

gregw:
Note also, that because URL handling might be changing in future JVMs, then that is PRECISELY why we should be hiding URL behind the `Resource` abstraction.     Rather than have URL specific code scattered around our code base, we should just use `Resource`.  If/when URL mechanism changes in the JVM, then we can adapt in our resource layer and all the code using it will now be able to deal with the new URL mechanism (or something else if that scheme is now handled by something different - like a modern bundle resource factory).

This is exactly why we have our resource abstraction, so we can hide such details and migrations from our code and deal with it in only one place.

joakime:
"serves no purpose in the general sense"

This means in a general URL support sense.
We have support for schemes `jar`, `jrt`, and `file` currently.

We don't want to support all of the schemes that have registered URL Stream Handler's for Resource, that's just dangerous.
This is opening us up to have a log4j style attack.

Here's a small list, as seen during our jetty-osgi test execution (using Java 11)

protocol | owner
--- | ---
`ftp:` | JVM
`http:` | JVM
`https:` | JVM
`jmod:` | JVM
`jrt:` | JVM
`mailto:` | JVM
`war:` | Jetty
`bundleentry:` | Eclipse OSGi
`bundleresource:` | Eclipse OSGi
`reference:` | Eclipse OSGi
`fake:` | Felix
`classpath:` | Pax
`link:` | Pax
`mvn:` | Pax
`wrap:` | Pax
`activator:` | Pax
`bytes:` | Plexus (now deprecated)
`nested:` | Eclipse Sisu
`bndloader:` | bndlib

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/AbstractUrlResourceFactory.java
```diff
@@ -0,0 +1,234 @@
...
+                return null;
+            }
+        }
+
+        @Override
+        public boolean exists()
+        {
+            try
+            {
+                newConnection();
```
gregw: Existence could probably be cached, i.e have a Boolean that is null if never tested and set by the first call to `newConnection()`... but then we should not worry too much about this class being efficient.

gregw:
> We don't want to support all of the schemes that have registered URL Stream Handler's for Resource, that's just dangerous

Primarily we don't want to use a URL stream handler for any resource for which we have a better implementation.
But I'm OK with not supporting arbitrary by default, so long as it is easy to define new mappings (and easy in XML/module).

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/AbstractUrlResourceFactory.java
```diff
@@ -0,0 +1,234 @@
...
+import java.time.Instant;
+
+import org.eclipse.jetty.util.FileID;
+
+/**
+ * Abstract ResourceFactory for {@link java.net.URL} based resources.
+ */
+public abstract class AbstractUrlResourceFactory implements ResourceFactory
+{
+    private final String supportedProtocol;
```
gregw: This class does not need to be abstract.  It just has the protocol injected via the constructor and it creates a needless amount of work to subclass just to do that.

Actually, it doesn't even need to check the protocol, as the factory is only registered for the protocols that we wish it to handle, thus a second check is duplication.

At the very least, it should have a set of protocols, as we don't want to create a new instance for every protocol if we have multiple, but I would prefer no check and just let the registration map do it's job.

gregw (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/AbstractUrlResourceFactory.java
joakime: I'm leaving the check there.

The protocol will be used ....

* to ensure that an old `ResourceFactory` instance isn't being reused for a different protocol.
* for output on `toString()` to indicate the supported protocol.
* for configuring the URLConnection via protocol based system properties (usecaches, connecttimeout, etc).

There's no harm in having multiple `UrlResourceFactory` implementations, one for each requested protocol, it's a support layer for old school behaviors anyway.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/AbstractUrlResourceFactory.java
joakime: If this class is used for `resourcebundle:` based URL protocols then we don't want existence cached.
If this class is used for `https:` then we also don't want the existence cached.

Besides, any kind of caching on our part would just conflict with the `URLConnection.setUseCaches(boolean)` behaviors.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
```diff
@@ -29,103 +33,169 @@
...
 
         @Override
         public void dump(Appendable out, String indent) throws IOException
         {
-            Dumpable.dumpObjects(out, indent, this, new DumpableCollection("mounts", _mounts));
+            Dumpable.dumpObjects(out, indent, this, new DumpableCollection("mounts", _compositeResourceFactory.getMounts()));
+        }
+    }
+
+    static class CompositeResourceFactory implements ResourceFactory
```
gregw, line 116: I know this is not a public name.... but `CompositeResourceFactory` is not really correct.
How about `DelegatingResourceFactory` as it delegates to a another factory.
Or maybe `BySchemeResourceFactory` as it chooses another factory by scheme.
Or to be absolutely precise `DelegatingBySchemeResourceFactory`.

Anything to avoid the contentious and misunderstood "Composite"!

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/UrlResourceFactory.java
```diff
@@ -0,0 +1,240 @@
...
+/**
+ * {@link ResourceFactory} for {@link java.net.URL} based resources.
+ */
+class UrlResourceFactory implements ResourceFactory
+{
+    private final String supportedProtocol;
+    private int connectTimeout;
+    private boolean useCaches;
+
+    protected UrlResourceFactory(String protocol)
```
gregw: @joakime said:
> I'm leaving the check there.

Sorry, but I still don't agree with having the protocol here.  

> The protocol will be used ....
> 
> * to ensure that an old `ResourceFactory` instance isn't being reused for a different protocol.

But that is controlled by the `Index` of scheme to factory.  We don't need to encode checks that index works.

> * for output on `toString()` to indicate the supported protocol.

That is only meaningful in context.  In context we'd be dumping the Index with it's full mapping of scheme to factory, so such a dump will now contain the scheme twice: once in the key and then again in the value.

> * for configuring the URLConnection via protocol based system properties (usecaches, connecttimeout, etc).

But is is not?

> 
> There's no harm in having multiple `UrlResourceFactory` implementations, one for each requested protocol, it's a support layer for old school behaviors anyway.

It is not "old school" yet.   It is currently the only supported mechanism for several schemes.  Besides, "old school" is not an excuse for not doing it right.

Having the protocol in this class prohibits it ever being used as a catch-all.  I'm not sure we need a catch-all, but it is conceivable. If we wanted one, then with this needless protocol check, we'd have to iterator over all the URL handlers registered and register a factory for each.  Or we could just trust that `Map` works and do away with a needless check.

gregw (changes requested):
Sorry, but whilst this PR is needed and holding up other work... I still do not agree with the needless protocol check.  It adds no value and is just something that will need to be worked around in future if we ever want a catch-all factory.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
joakime: how about `UriResourceFactory`?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/UrlResourceFactory.java
joakime: > > * for configuring the URLConnection via protocol based system properties (usecaches, connecttimeout, etc).
> 
> But is is not?

Don't follow you on this one?

> Having the protocol in this class prohibits it ever being used as a catch-all. I'm not sure we need a catch-all, but it is conceivable. If we wanted one, then with this needless protocol check, we'd have to iterator over all the URL handlers registered and register a factory for each. Or we could just trust that Map works and do away with a needless check.

A catch-all is dangerous, and should never be a solution we implement.
And we have an `Index` controlling use/access to these implementation already.

gregw: > > > * for configuring the URLConnection via protocol based system properties (usecaches, connecttimeout, etc).
> > 
> > 
> > But is is not?
> 
> Don't follow you on this one?

You don't use it for configuring the "URLConnection via protocol bases system properties".  The System properties used are class name based.  Not sure we should be using any system properties for configuration anyway!

> And we have an Index controlling use/access to these implementation already

Exactly, so we don't need a double check!

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/UrlResourceFactory.java
```diff
@@ -0,0 +1,240 @@
...
+    {
+        this.supportedProtocol = protocol;
+        this.connectTimeout = Integer.parseInt(System.getProperty(this.getClass().getName() + ".connectTimeout", "1000"));
+        this.useCaches = Boolean.parseBoolean(System.getProperty(this.getClass().getName() + ".useCaches", "true"));
```
gregw: Let's not continue using System Properties for configuration.  There are setters available.

gregw (changes requested):

gregw (approved):
LGTM.

PR #8899, opened 2022-11-14
Fix #8897 Ignore conditional headers as per RFC7232

### LINKED ISSUE #8897: Update Conditional request handling for RFC7232
**Jetty version(s)**
10, 11, 12

**Enhancement Description**

[RFC7232 3.3](https://www.rfc-editor.org/rfc/rfc7232#section-3.3) says:
> A recipient MUST ignore If-Modified-Since if the request contains an
> If-None-Match header field; the condition in If-None-Match is
> considered to be a more accurate replacement for the condition in
> If-Modified-Since, and the two are only combined for the sake of
> interoperating with older intermediaries that might not implement
> If-None-Match

and [RFC7232 3.44](https://www.rfc-editor.org/rfc/rfc7232#section-3.4) says:
> A recipient MUST ignore If-Unmodified-Since if the request contains
> an If-Match header field; the condition in If-Match is considered to
> be a more accurate replacement for the condition in
> If-Unmodified-Since, and the two are only combined for the sake of
> interoperating with older intermediaries that might not implement
> If-Match.

Currently we fall through to handling the older style header if etags is not known or the previous condition passes.    
We MUST respect rfc7232 in this regard.

gregw:
Ignore date based headers if etag ones are present. Also avoid parsing dates unless necessary.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
@@ -596,10 +596,14 @@ public class ResourceService
                 }
 
                 long ifmsl = DateParser.parseDate(ifms);
-                if (ifmsl != -1 && content.getResource().lastModified() / 1000 <= ifmsl / 1000)
+                if (ifmsl != -1)
                 {
-                    sendStatus(response, HttpServletResponse.SC_NOT_MODIFIED, content::getETagValue);
-                    return false;
+                    long lm = content.getResource().lastModified();
+                    if (lm != -1 && lm / 1000 <= ifmsl / 1000)
+                    {
+                        sendStatus(response, HttpServletResponse.SC_NOT_MODIFIED, content::getETagValue);
+                        return false;
+                    }
                 }
             }
 
@@ -607,10 +611,14 @@ public class ResourceService
             if (ifums != null && ifm == null)
             {
                 long ifumsl = DateParser.parseDate(ifums);
-                if (ifumsl != -1 && content.getResource().lastModified() / 1000 > ifumsl / 1000)
+                if (ifumsl != -1)
                 {
-                    response.sendError(HttpServletResponse.SC_PRECONDITION_FAILED);
-                    return false;
+                    long lm = content.getResource().lastModified();
+                    if (lm != -1 && lm / 1000 > ifumsl / 1000)
+                    {
+                        response.sendError(HttpServletResponse.SC_PRECONDITION_FAILED);
+                        return false;
+                    }
                 }
             }
         }
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
@@ -497,7 +497,7 @@ public class ResourceService
             String ifm = null;
             String ifnm = null;
             String ifms = null;
-            long ifums = -1;
+            String ifums = null;
 
             if (request instanceof Request)
             {
@@ -518,7 +518,7 @@ public class ResourceService
                                 ifms = field.getValue();
                                 break;
                             case IF_UNMODIFIED_SINCE:
-                                ifums = DateParser.parseDate(field.getValue());
+                                ifums = field.getValue();
                                 break;
                             default:
                         }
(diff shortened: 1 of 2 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -604,10 +604,14 @@ protected boolean passConditionalHeaders(HttpServletRequest request, HttpServlet
             }
 
             // Parse the if[un]modified dates and compare to resource
-            if (ifums != -1 && content.getResource().lastModified() / 1000 > ifums / 1000)
+            if (ifums != null && ifm == null)
```
joakime, line 611: Since this is the `ResourceService` (handling only `GET` and `HEAD`), the `If-Unmodified-*` headers are never going to be seen here.

jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -595,7 +595,7 @@ protected boolean passConditionalHeaders(HttpServletRequest request, HttpServlet
                     return false;
                 }
 
-                long ifmsl = request.getDateHeader(HttpHeader.IF_MODIFIED_SINCE.asString());
+                long ifmsl = DateParser.parseDate(ifms);
```
joakime, line 598: Doesn't `request.getDateHeader()` return a long already?  Why return/keep the String then convert it later to a long?

gregw: Calling `getDateHeader` does another iteration over the fields, which we have already done above.  We don't get the long value above because we don't want to parse it unless we know there is no If-None-Match header, but we won't know that until after we finished the iteration.

So we just remember the String and only convert if we need to.  Thus only a single iteration and only a conversion if we need it.

jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
gregw: Is it "never" or just very VERY rarely?
I'm inclined to leave support in for 10/11 and remove support in 12.

joakime: Prior to RFC9110, the `If-Unmodified-Since` header was for unsafe methods (like `POST`) and had no meaning for safe methods (like `GET`).

But now https://www.rfc-editor.org/rfc/rfc9110#name-if-unmodified-since says ..

> A client MAY send an If-Unmodified-Since header field in a [GET](https://www.rfc-editor.org/rfc/rfc9110#GET) request to indicate that it would prefer a [412 (Precondition Failed)](https://www.rfc-editor.org/rfc/rfc9110#status.412) response if the selected representation has been modified. However, this is only useful in range requests ([Section 14](https://www.rfc-editor.org/rfc/rfc9110#range.requests)) for completing a previously received partial representation when there is no desire for a new representation. [If-Range](https://www.rfc-editor.org/rfc/rfc9110#field.if-range) ([Section 13.1.5](https://www.rfc-editor.org/rfc/rfc9110#field.if-range)) is better suited for range requests when the client prefers to receive a new representation.

So I'm actually thinking we should instead add back support in Jetty 12.

jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -595,7 +595,7 @@ protected boolean passConditionalHeaders(HttpServletRequest request, HttpServlet
...
                 if (ifmsl != -1 && content.getResource().lastModified() / 1000 <= ifmsl / 1000)
```
joakime: Per https://www.rfc-editor.org/rfc/rfc9110#section-13.1.3-7

> A recipient MUST ignore the If-Modified-Since header field if the resource does not have a modification date available.

The value of `content.getResource().lastModified()` should be checked to know if the resource has a modification date and skip this check if it has no modification date.

jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
joakime, line 598: Per https://www.rfc-editor.org/rfc/rfc9110#section-13.1.3-6

> A recipient MUST ignore the If-Modified-Since header field if the received field value is not a valid HTTP-date, the field value has more than one member, or if the request method is neither GET nor HEAD.

If `ifmsl` is `-1` indicating a bad Date, we should ignore the rest of the handling of `If-Modified-Since`.

jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
gregw: done

joakime (approved):

PR #8900, opened 2022-11-14
Improved documentation about virtual threads.

### LINKED ISSUE #8892: Document useVirtualThreads
**Jetty version(s)**
10+

**Description**
The newly introduced support for virtual threads in #8007 was not documented, and it should be.

sbordet:
Added programming guide section about Jetty threading model.


documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-threadpool.adoc
```diff
@@ -37,3 +37,11 @@ The time, in milliseconds, after which an idle thread is released from the pool
...
 `jetty.threadPool.maxThreads`::
 The max number of threads pooled by the thread pool -- defaults to 200.
+
+`jetty.threadPool.useVirtualThreads`::
+Whether to enable use of virtual threads, if the runtime supports them.
+Virtual threads are supported as a preview feature since Java 19.
++
+NOTE: To enable preview features, you need to specify the `+--enable-preview+` command line option, see the section about the xref:og-modules-directive-exec[[exec\] directive].
++
+By default, virtual threads are not used even if the runtime support them -- their use should be explicitly enabled.
```
gregw:
```suggestion
By default, virtual threads are not used even if the runtime support them -- their use must be explicitly enabled.
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
```diff
@@ -0,0 +1,200 @@
...
+0 is T1 #lightgreen
+1 is "Run T1" #dodgerblue
+5 is T2 #lightgreen
+6 is "Run T2" #dodgerblue
+8 is T3 #lightgreen
+9 is "Run T3" #dodgerblue
+12 is {hidden}
+----
+
+If the task is a NIO selection event, then this model is the thread-per-selector model which is very CPU core cache efficient, but suffers from the link:http://en.wikipedia.org/wiki/Head-of-line_blocking[head-of-line blocking]: if one of the tasks blocks or runs slowly, then subsequent tasks cannot be produced (and therefore cannot be consumed either) and will pay in latency the cost of running previous, possibly unrelated, tasks.
```
gregw: Maybe we should include hints about the Adaptive Strategy:
```suggestion
If the task is a NIO selection event, then this model is the thread-per-selector model which is very CPU core cache efficient, but suffers from the link:http://en.wikipedia.org/wiki/Head-of-line_blocking[head-of-line blocking]: if one of the tasks blocks or runs slowly, then subsequent tasks cannot be produced (and therefore cannot be consumed either) and will pay in latency the cost of running previous, possibly unrelated, tasks.

This strategy should only be used if the produced task is known not to block or the deployment is able to suffer head of line blocking.
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
```diff
@@ -0,0 +1,200 @@
...
+2 is "Run T2" #dodgerblue
+4 is "Run T4" #dodgerblue
+8 is {hidden}
+
+@WT3
+3 is "Run T3" #dodgerblue
+6 is {hidden}
+----
+
+The `Executor` implementation typically adds the task to a queue, and dequeues the task when there is a worker thread available to run it.
```
gregw, line 102:
```suggestion
The `Executor` implementation typically adds the task to a queue, and dequeues the task when there is a worker thread available to run it. Configurations that support virtual threading are available that avoid such task queues.
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
```diff
@@ -0,0 +1,200 @@
...
+
+This mode may also operate similarly to xref:pg-arch-threads-execution-strategy-pec[`Produce-Execute-Consume`] when the take over production task always finds a free CPU core immediately (for example, in a mostly idle server): thread `Thread 1` will produce a task, yield production to `Thread 2` while `Thread 1` is running the task; `Thread 2` will produce a task, yield production to `Thread 3` while `Thread 2` is running the task, etc.
+
+Differently from `Produce-Execute-Consume`, here production happens on different threads, but the advantage is that the task is run by the same thread that produced it (which is CPU core cache efficient).
+
+[[pg-arch-threads-execution-strategy-adaptive]]
+===== Adaptive Execution Strategy
+The modes of task consumption discussed above are captured by the `org.eclipse.jetty.util.thread.ExecutionStrategy` interface, with an additional implementation that also takes into account the behavior of the task when the task implements `Invocable`.
+
+For example, a task that declares itself as non-blocking could be consumed using the `Produce-Consume` mode, since there is no risk to stop production because the task will not block.
```
gregw:
```suggestion
For example, a task that declares itself as non-blocking can be consumed using the `Produce-Consume` mode, since there is no risk to stop production because the task will not block.
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
```diff
@@ -0,0 +1,200 @@
...
+
+Conversely, a task that declares itself as blocking will stop production, and therefore must be consumed using either the `Produce-Execute-Consume` mode or the `Execute-Produce-Consume` mode.
+Deciding between these two modes depends on whether there is a free thread immediately available to take over production, and this is captured by the `org.eclipse.jetty.util.thread.TryExecutor` interface.
+
+An implementation of `TryExecutor` could be asked whether it can immediately allocate a free thread to run a task, as opposed to a normal `Executor`, that can only queue the task in the hope that there will be a thread in the near future that could run the task.
```
gregw:
```suggestion
An implementation of `TryExecutor` can be asked whether a thread can be immediately allocated to run a task, as opposed to a normal `Executor`, that can only queue the task in the expectation that there will be a thread available in the near future to run the task.
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
```diff
@@ -0,0 +1,200 @@
...
+
```
gregw, line 165:
```suggestion

If a task declares that it can run in either blocking or non blocking mode, then the strategy may choose to use Produce-Consume with the additional invocation flag to force non-blocking mode; or it may choose one of the other strategies used for blocking tasks.
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
```diff
@@ -0,0 +1,200 @@
...
+Virtual threads have been introduced in Java 19 as a preview feature.
+
+NOTE: In Java versions where virtual threads are a preview feature, remember to add `+--enable-preview+` to the command line options to use virtual threads.
+
+`QueuedThreadPool` can be configured to use virtual threads via:
+
+[source,java,indent=0]
+----
+include::{doc_code}/org/eclipse/jetty/docs/programming/QueuedThreadPoolDocs.java[tags=virtual-threads]
+----
```
gregw, line 203:
```suggestion
----
[NOTE]
====
If virtual thread are enabled in `QueuedThreadPool`, then the heuristic for determining the number of reserved threads will return 0, so that a virtual thread will always be used for blocking tasks.   If a deployment wishes to use enable the use of Execute-Produce-Consume strategy for blocking tasks, then a specific number of reserved threads must be configured. 
====
```

gregw (changes requested):

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
```diff
@@ -0,0 +1,200 @@
...
+
+[[pg-arch-threads]]
+=== Jetty Threading Architecture
+
+Writing a performant client or server is difficult, because it should:
+
+* Scale well with the number of processors.
+* Be efficient at using processor caches to avoid link:https://en.wikipedia.org/wiki/Parallel_slowdown[parallel slowdown].
+* Support multiple network protocols that may have very different requirements; for example, multiplexed protocols such as HTTP/2 introduce new challenges that are not present in non-multiplexed protocols such as HTTP/1.1.
+* Support different application threading models; for example, if a Jetty server invokes server-side application code that is allowed to call blocking APIs, then the Jetty server should not be affected by how long the blocking API call is, and should be able to process other connections or other requests in a timely fashion.
```
lorban:
```suggestion
Support different application threading models; for example, if a Jetty server invokes server-side application code that is allowed to call blocking APIs, then the Jetty server should not be affected by how long the blocking API call takes, and should be able to process other connections or other requests in a timely fashion.
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
```diff
@@ -0,0 +1,200 @@
...
+@WT3
+3 is "Run T3" #dodgerblue
+6 is {hidden}
+----
+
+The `Executor` implementation typically adds the task to a queue, and dequeues the task when there is a worker thread available to run it.
+
+This mode solves the head-of-line blocking discussed in the xref:pg-arch-threads-execution-strategy-pc[`Produce-Consume` section], but suffers of other issue:
+
+* It is not CPU core cache efficient, as the data available to the producer thread will need to be accessed by another thread that likely won't have that data in its caches.
```
lorban:
```suggestion
* It is not CPU core cache efficient, as the data available to the producer thread will need to be accessed by another thread that likely is going to run on a CPU core that won't have that data in its cache.
```

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-threadpool.adoc
sbordet: Done

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
sbordet: Slightly modified, but done.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
sbordet: I don't think that is true. A virtual thread is "queued" as a continuation into the scheduler, typically the `ForkJoinPool`, so the generic sentence works for virtual thread too.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
sbordet: Done.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
sbordet: Done.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
sbordet: I left this out on purpose to avoid to go too deep into the implementation details.
For example, I do not mention the EITHER invocation type, the PRODUCE_INVOKE_CONSUME mode, etc.
For now I'd leave it out.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
sbordet: Modified and expanded, but done.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
sbordet: Done.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
sbordet: Done.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
gregw: But... But... Loom is magic! It is not bound by such concerns!

But seriously, i think a mention here that virtual threads are an alternative to thread pooling would be good

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
gregw: I think it is important to add at least a small mention like this. Else it appears that the documentation does not describe the actual implementation. It doesn't need a deep dive, just a mention.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
joakime: Loom is trying to make it so that you don't care about those nuances and details.

That's what it seems like (at least) when you read their own documentation and blog posts about it.

sbordet: This paragraph is not the right section to introduce virtual threads.
I'm just explaining PEC, and how it works.
It is when PEC is used in the context of AdaptiveExecutionStrategy that virtual threads can be used, so it is explained there.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
sbordet: Your suggestion is just too obscure for those that don't know the implementation.
99.999% of usages will be either blocking or non-blocking, "either" is really for super-special cases and if a developer is there, reading the code/javadocs is better than reading the documentation.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-threadpool.adoc
```diff
@@ -16,15 +16,15 @@
...
 The `threadpool` module allows you to configure the server-wide thread pool.
 
-The thread pool creates threads on demand up to `maxThreads`, and idle them out if they are not used.
+The thread pool creates threads on demand up to `maxThreads`, and idles them out if they are not used.
 
 Since Jetty uses the thread pool internally to execute critical tasks, it is not recommended to constrain the thread pool to small values of `maxThreads` with the purpose of limiting HTTP request concurrency, as this could very likely cause a server lockup when Jetty needs to run a critical task but there are no threads available.
 Start with the default value of `maxThreads`, and tune for larger values if needed.
 
-The module file is `$JETTY_HOME/modules/threadpool.mod`:
+The module properties to configure the thread pool are:
```
joakime, line 24: I think you can just drop this line and/or merge it with line 30

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
```diff
@@ -0,0 +1,210 @@
...
+----
+
+`AdaptiveExecutionStrategy` makes use of this setting when it determines that a task should be run with the xref:pg-arch-threads-execution-strategy-pec[`Produce-Execute-Consume` mode]: rather than submitting the task to `QueuedThreadPool` to be run in a platform thread, it runs the task in a new virtual thread.
+
+[NOTE]
+====
+Enabling virtual threads in `QueuedThreadPool` will default the number of reserved threads to zero, unless the number of reserved threads is explicitly configured to a positive value.
+
+Defaulting the number of reserved threads to zero ensures that the xref:pg-arch-threads-execution-strategy-pec[Produce-Execute-Consume mode] is always used, which means that virtual threads will always be used for blocking tasks.
+====
```
joakime, line 217: A common question that comes up is how threading is used during an HTTP exchange, is this the document/place to discuss this?

joakime (commented):
Just some questions

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-threadpool.adoc
sbordet: It's a pattern that I used for all the modules docs, so I'd like to keep it.

joakime: Then line 30 is redundant.

sbordet: Why?

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
sbordet: Not here particularly, as this is a generic section that works for both client and server (so maybe 2 sections, one for the client, one for the server), but yes it would be nice to have.
Not in this PR though.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
gregw: Then leave out the typically explanation of what an executor does.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/arch-threads.adoc
gregw: I still think that EITHER and PIC needs to be mentioned somewhere in the doco, otherwise it looks like it is for a different version of the code.

joakime (dismissed):

gregw (dismissed):

joakime (approved):

PR #8903, opened 2022-11-16
Fixes #8863 - Provide a possibility to name virtual threads

### LINKED ISSUE #8863: Provide a possibility to name virtual threads
**Enhancement Description**

Jetty 11 supports Loom, but we don't have any control over the used executor:

https://github.com/eclipse/jetty.project/blob/1e0ef57bb4656b7e5e29bf38b1c7a2772e2d0a2a/jetty-util/src/main/java/org/eclipse/jetty/util/VirtualThreads.java#L34-L41

We should be able to customize executor using `OfVirtual`:

* https://download.java.net/java/early_access/loom/docs/api/java.base/java/lang/Thread.Builder.OfVirtual.html

To e.g. name virtual threads (currently it looks like `VirtualThread[#349]/runnable@ForkJoinPool-1-worker-5`). Would be nice, if Jetty could provide an abstraction over this builder or just provide a possibility to manually assign customized instance. Currently it's not possible, because this field is final, so some kind of a setter could be an option.

sbordet:
Reworked the VirtualThreads APIs to be based on Executor rather than just boolean.

### DIFF
--- a/jetty-http-spi/src/main/java/org/eclipse/jetty/http/spi/DelegatingThreadPool.java
+++ b/jetty-http-spi/src/main/java/org/eclipse/jetty/http/spi/DelegatingThreadPool.java
@@ -63,16 +63,16 @@ public class DelegatingThreadPool extends ContainerLifeCycle implements ThreadPo
     }
 
     @Override
-    public boolean isUseVirtualThreads()
+    public Executor getVirtualThreadsExecutor()
     {
-        return VirtualThreads.isUseVirtualThreads(_executor);
+        return VirtualThreads.getVirtualThreadsExecutor(_executor);
     }
 
     @Override
-    public void setUseVirtualThreads(boolean useVirtualThreads)
+    public void setVirtualThreadsExecutor(Executor executor)
     {
         if (_executor instanceof VirtualThreads.Configurable)
-            ((VirtualThreads.Configurable)_executor).setUseVirtualThreads(useVirtualThreads);
+            ((VirtualThreads.Configurable)_executor).setVirtualThreadsExecutor(executor);
     }
 
     @Override
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
@@ -136,8 +136,8 @@ public class AdaptiveExecutionStrategy extends ContainerLifeCycle implements Exe
     private final Producer _producer;
     private final Executor _executor;
     private final TryExecutor _tryExecutor;
+    private final Executor _virtualExecutor;
     private final Runnable _runPendingProducer = () -> tryProduce(true);
-    private Executor _virtualThreadsExecutor;
     private State _state = State.IDLE;
     private boolean _pending;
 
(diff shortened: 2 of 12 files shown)


jetty-http-spi/src/main/java/org/eclipse/jetty/http/spi/DelegatingThreadPool.java
```diff
@@ -63,16 +63,16 @@ public boolean tryExecute(Runnable task)
     }
 
     @Override
-    public boolean isUseVirtualThreads()
```
joakime, line 66: Do we have a release with this name?
If so, this can't be removed, only deprecated.

jetty-http-spi/src/main/java/org/eclipse/jetty/http/spi/DelegatingThreadPool.java
```diff
@@ -63,16 +63,16 @@ public boolean tryExecute(Runnable task)
...
+    public Executor getVirtualThreadsExecutor()
     {
-        return VirtualThreads.isUseVirtualThreads(_executor);
+        return VirtualThreads.getVirtualThreadsExecutor(_executor);
     }
 
     @Override
-    public void setUseVirtualThreads(boolean useVirtualThreads)
```
joakime, line 72: Do we have a release with this name?
If so, this can't be removed, only deprecated.

joakime (commented):
Do we have a release with the old method names?
If so, this can't be removed, only deprecated.

jetty-http-spi/src/main/java/org/eclipse/jetty/http/spi/DelegatingThreadPool.java
sbordet: It's not removed, it's `default` in the interface so it's inherited anyway.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/strategy/AdaptiveExecutionStrategy.java
```diff
@@ -471,10 +471,10 @@ private void execute(Runnable task)
...
         try
         {
-            if (isUseVirtualThreads())
-                VirtualThreads.executeOnVirtualThread(task);
-            else
-                _executor.execute(task);
+            Executor executor = _virtualThreadsExecutor;
+            if (executor == null)
+                executor = _executor;
+            executor.execute(task);
```
gregw: It is a little strange that we have both an _executor and a _virtualThreadsExecutor, but use one or the other.

I guess we need the _executor when running with virtual threads and reserved threads, so the reserved threads can be started by the normal executor.

It would be nice if it could more closely resemble the _tryExecutor pattern.

gregw (commented):
I don't love it, as it is similar yet different to the pattern we've applied for TryExecutor.   
Was there a reason we didn't just invent our own interface like:
```java
interface VirtualExecutor
{
    void executeVirtually(Runnable task);
}
```
So we could follow the pattern of TryExecutor ?

sbordet:
@gregw `TryExecutor` is an "internal" abstraction -- we do not allow users to set their own `TryExecutor` on an existing thread pool.

For virtual threads is different, we want to give users a way to pass their own virtual `Executor` which they can build how they want (e.g. with names, with ThreadLocals support, and all the properties that `Thread.Builder` supports).

We want also to support a "simple" way to configure this in Jetty modules.
Right now it's a simple boolean, but feels like we should have instead a `threadpool-virtual.mod` instead where we can configure the virtual `Executor` (with names, etc.).

WRT to `AdaptiveExecutionStrategy`, it's really an `if/else`: if virtual threads are enabled do this, else do that. 
Before it was done with a boolean.

Even if from `AdaptiveExecutionStrategy` we call only `executeVirtually(task)`, the `QueuedThreadPool` would have to do the `if/else` and delegate either to a virtual `Executor`, or to a itself as a non-virtual `Executor`.

I think we need 2 `Executors` because we want to give different names, and it's not only the names: the virtual `Executor` can be configured to _not_ support `ThreadLocal`s, etc.

PR #8909, opened 2022-11-17
Jetty 12 recycle servlet channel

gregw:
Implement TODO's in ee10-servlet regarding Jetty 12 recycling of `ServletChannel`


jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextHandler.java
```diff
@@ -1198,14 +1198,11 @@ protected ServletContextRequest wrap(Request request)
             return null;
 
         // Get a servlet request, possibly from a cached version in the channel attributes.
-        // TODO We should cache this heavy weight object!  Something like:
-        // TODO: ServletChannel is not properly cleared out so I have disabled the caching of this for now.
-        ServletChannel servletChannel = null; // (ServletChannel)request.getComponents().getCache().get("blah.blah.ServletChannel");
+        ServletChannel servletChannel = (ServletChannel)request.getComponents().getCache().get(ServletChannel.class.getName());
```
gregw: This wont work for HTTP2 as there can be concurrent requests for the same connection

gregw: Oh yes it will, cache is scoped to `HttpChannelState`

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -64,52 +64,59 @@ public class ServletChannel implements Runnable
...
-        _expects100Continue = request.getHeaders().contains(HttpHeader.EXPECT, HttpHeaderValue.CONTINUE.asString());
-        _combinedListener = new Listeners(request);
-
-        request.getHttpInput().init();
+        _state.recycle();
+        _httpInput.reopen();
+        _servletContextApi = servletContextRequest.getContext().getServletContext();
+        _servletContextRequest = servletContextRequest;
+        _expects100Continue = servletContextRequest.getHeaders().contains(HttpHeader.EXPECT, HttpHeaderValue.CONTINUE.asString());
+        _callback = null;
```
lachlan-roberts: `_state.recycle()` and `_callback = null` should be done in `recycle()` method not in the init.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -345,10 +352,12 @@ public void continue100(int available) throws IOException
         }
     }
 
-    public void recycle()
+    private void recycle()
     {
         _written = 0;
         _oldIdleTimeout = 0;
+        _httpInput.recycle();
+        _callback = null;
```
lachlan-roberts, line 377: Can we clear out things like `_servletContextRequest` and `_servletContextApi` as well.

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/AsyncServletIOTest.java
```diff
@@ -416,6 +416,7 @@ public void doGet(final HttpServletRequest request, final HttpServletResponse re
                     @Override
                     public void onError(Throwable t)
                     {
+                        t.printStackTrace();
```
lachlan-roberts: Do we need this?

lachlan-roberts (changes requested):

lachlan-roberts:
@gregw Looks like there was some failure with this PR on Jenkins.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
```diff
@@ -47,9 +47,10 @@ class AsyncContentProducer implements ContentProducer
     private long _firstByteNanoTime = Long.MIN_VALUE;
     private long _rawBytesArrived;
 
-    AsyncContentProducer(ServletChannel servletChannel)
+    AsyncContentProducer(ServletChannel servletChannel, AutoLock lock)
```
janbartel, line 54: This constructor should have a comment to explain why locking is now being controlled externally.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/BlockingContentProducer.java
```diff
@@ -28,17 +28,19 @@ class BlockingContentProducer implements ContentProducer
 
     private final AsyncContentProducer _asyncContentProducer;
     private final AsyncContentProducer.LockedSemaphore _semaphore;
+    private final AutoLock _lock;
 
-    BlockingContentProducer(AsyncContentProducer delegate)
+    BlockingContentProducer(AsyncContentProducer delegate, AutoLock lock)
```
janbartel: Same as previous - this should have a comment explaining the locking strategy

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/HttpOutput.java
```diff
@@ -354,7 +354,7 @@ else if (closeContent != null)
         finally
         {
             if (wake)
-                _servletChannel.execute(_servletChannel); // TODO review in jetty-10 if execute is needed
+                _servletChannel.execute(_servletChannel::handle); // TODO review in jetty-10 if execute is needed
```
janbartel: Is this TODO still relevant, and shouldn't it be in jetty-10 code rather than jetty-12 code?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -60,24 +60,39 @@
 /**
  * TODO describe what this class does
```
janbartel: Yes please, describe what this class does!

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -794,19 +810,25 @@ public void onCompleted()
...
         // Callback will either be succeeded here or failed in abort().
+        Callback callback = _callback;
+        // Must recycle before notification to allow for reuse.
+        // Recycle always done here even if an abort is called.
+        recycle();
         if (_state.completeResponse())
-            _callback.succeeded();
-        _combinedListener.onComplete(_request);
+        {
+            _combinedListener.onComplete(_servletContextRequest);
```
janbartel: If line 826 calls `recycle()`, won't the `_servletContextRequest` at line 829 always be null for the call to `onComplete()`?

janbartel (changes requested):

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -58,31 +58,42 @@
...
+    private volatile Callback _callback;
     // Bytes written after interception (e.g. after compression).
-    private long _written;
+    private volatile long _written;
 
     public ServletChannel(ServletContextHandler servletContextHandler, Request request)
     {
         _state = new ServletRequestState(this);
+        _context = servletContextHandler.getContext();
+        _servletContextApi = _context.getServletContext();
```
lachlan-roberts, line 96: Couldn't this be recycled and go through a different `ServletContextHandler` which would have a different `ServletContextApi`? 

If so then this would need to be done in `associate()`.

gregw: No, see the test when we reuse a `ServletChannel` that it is for the same context, else we create a new one.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
gregw: done

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
lachlan-roberts: Wouldn't it be better to have a `ServletChannel` which could be re-used for any context instead of always remaking it if it goes to a different context.

gregw: Yeah maybe, but as @lorban would say, that is optimizing the slow path.  Cross context requests are rare and many deployments only have 1 context per server.
Having these big objects with lots of final fields increases the benefit of re-using them, as all those final fields don't need to be re-enabled.

So I'm happy with it as is for now.   Perhaps if we see cross context slowness as a problem in future we invest in a reusable instance per context?

gregw:
@janbartel @lachlan-roberts re-review please

janbartel (approved):

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
lachlan-roberts: Ok, at least its better than before so I'll approve.

I can't see that there would be much extra cost to making it fully re-usable though.

lachlan-roberts (approved):

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
gregw: Maybe, but I don't really know the cost of creating new ServletContextAPI instances for each request.  I'd much rather reuse them most of the time and then make a new one only if changing contexts.

PR #8915, opened 2022-11-18
Fixes #8913 - Review Jetty XML syntax to allow calling JDK methods

### LINKED ISSUE #8913: Review Jetty XML syntax to allow calling JDK methods
**Jetty version(s)**
10+

**Description**
Jetty XML syntax allow to call methods of any class via reflection.

However, when calling JDK methods, they may return JDK-private implementation classes that are not accessible from Jetty code.
This surfaced in particular while working on #8863, where a new Jetty module for the thread pool that supports virtual threads need to call JDK methods that return JDK-private classes.

For that example, we have:

```xml
<!-- returns JDK private class ThreadBuilders.VirtualThreadBuilder -->
<Call class="java.lang.Thread" name="ofVirtual">
  <!-- Throws IllegalAccessException: class org.eclipse.jetty.xml.XmlConfiguration$JettyXmlConfiguration 
cannot access a member of class java.lang.ThreadBuilders$VirtualThreadBuilder (in module java.base) 
with modifiers "public volatile" -->
  <Call name="name">virtual</Call>
</Call>
```

The JDK-private class `ThreadBuilders.VirtualThreadBuilder` is not accessible, but its JDK-public ancestor `Thread.Builder` is.

The proposal is to change slightly the semantic of the `Call` element to support this:

```xml
<!-- static invocation -->
<Call class="java.lang.Thread" name="ofVirtual">
  <!-- virtual invocation -->
  <Call class="java.lang.Thread$Builder" name="name">virtual</Call>
</Call>
```

The second `Call` element specifies the `class` attribute that should be used to reflect the method named `name` but it performs a virtual call, not a static call (because it is in the context of an enclosing `Call` element).

Ambiguities are easily resolved because it's always possible to ask a `Method` if it is `static` and if so perform the invocation passing `null` as the target of the invocation.

sbordet:
Now <Call>, <Get> and <Set> elements can use the "class" attribute to specify the exact class to perform method lookup.

Improved support for <Property>, <SystemProperty> and <Env> so that attribute "name" is now optional (as specified in the DTD), and a "deprecated" attribute may be present instead.
This is necessary to terminally deprecate properties that have no replacement.

### DIFF
--- a/documentation/jetty-documentation/src/main/asciidoc/operations-guide/xml/xml-syntax.adoc
+++ b/documentation/jetty-documentation/src/main/asciidoc/operations-guide/xml/xml-syntax.adoc
@@ -74,6 +74,8 @@ var mystring = new String();
 
 If an object with the id `mystring` already exists, then it is not created again but rather just referenced.
 
+Within element `<Configure>`, the created object (if any) is in xref:og-xml-syntax-scope[scope] and may be the implicit target of other, nested, elements.
+
 Typically the `<Configure>` element is used to configure a `Server` instance or `ContextHandler` subclasses such as `WebAppContext` that represent web applications.
 
 [[og-xml-syntax-arg]]
@@ -187,12 +189,38 @@ This is equivalent to the following Java code:
 var myhost = InetAddress.getByName("jdk.java.net");
 ----
 
+The `class` attribute can also be used to specify the Java class or interface to use to lookup the non-``static`` method name.
+This is necessary when the object in scope, onto which the `<Call>` would be applied, is an instance of a class that is not visible to Jetty classes, or not accessible because it is not `public`.
+For example:
+
+[source,xml,subs=normal]
+----
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "https://www.eclipse.org/jetty/configure_10_0.dtd">
+
+<Configure>
+  <Call class="java.util.concurrent.Executors" name="newSingleThreadScheduledExecutor">
+    #<Call class="java.util.concurrent.ExecutorService" name="shutdown" />#
+  </Call>
+</Configure>
+----
+
+In the example above, `Executors.newSingleThreadScheduledExecutor()` returns an object whose class is a private JDK implementation class.
+Without an explicit `class` attribute, it is not possible to invoke the method `shutdown()` when it is obtained via reflection from the private JDK implementation class, because while the method is `public`, the private JDK implementation class is not, therefore this exception is thrown:
+
+[source]
+----
+java.lang.IllegalAccessException: class org.eclipse.jetty.xml.XmlConfiguration$JettyXmlConfiguration (in module org.eclipse.jetty.xml) cannot access a member of class java.util.concurrent.Executors$DelegatedExecutorService (in module java.base) with modifiers "public"
+----
+
+The solution is to explicitly use the `class` attribute of the `<Call>` element that is invoking the `shutdown()` method, specifying a publicly accessible class or interface that the object in scope extends or implements (in the example above `java.util.concurrent.ExecutorService`).
+
 [[og-xml-syntax-get]]
 ===== `<Get>`
 
 Element `<Get>` retrieves the value of a JavaBean property specified by the mandatory `name` attribute.
 
-If the JavaBean property is `foo` (or `Foo`), `<Get>` first attempts to invoke _method_ `getFoo()`; failing that, attempts to retrieve the value from _field_ `foo` (or `Foo`).
+If the JavaBean property is `foo` (or `Foo`), `<Get>` first attempts to invoke _method_ `getFoo()` or _method_ `isFoo()`; failing that, attempts to retrieve the value from _field_ `foo` (or `Foo`).
 
 [source,xml]
 ----
@@ -211,6 +239,8 @@ If the JavaBean property is `foo` (or `Foo`), `<Get>` first attempts to invoke _
   </Get>
 </Configure>
 ----
+
+The `class` attribute allows to perform `static` calls, or to lookup the getter method from the specified class, as described in the xref:og-xml-syntax-call[`<Call>` section].
 
 [[og-xml-syntax-set]]
 ===== `<Set>`
@@ -234,6 +264,8 @@ If the JavaBean property is `foo` (or `Foo`), `<Set>` first attempts to invoke _
   </Set>
 </Configure>
 ----
+
+The `class` attribute allows to perform `static` calls, or to lookup the setter method from the specified class, as described in the xref:og-xml-syntax-call[`<Call>` section].
 
 [[og-xml-syntax-map]]
 ===== `<Map>` and `<Entry>`
--- a/documentation/jetty-documentation/src/main/asciidoc/operations-guide/xml/xml-syntax.adoc
+++ b/documentation/jetty-documentation/src/main/asciidoc/operations-guide/xml/xml-syntax.adoc
@@ -189,7 +189,7 @@ This is equivalent to the following Java code:
 var myhost = InetAddress.getByName("jdk.java.net");
 ----
 
-The `class` attribute can also be used to specify the Java class or interface to use to lookup the non-``static`` method name.
+The `class` attribute (or `<Class>` element) can also be used to specify the Java class or interface to use to lookup the non-``static`` method name.
 This is necessary when the object in scope, onto which the `<Call>` would be applied, is an instance of a class that is not visible to Jetty classes, or not accessible because it is not `public`.
 For example:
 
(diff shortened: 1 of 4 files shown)


jetty-xml/src/test/java/org/eclipse/jetty/xml/XmlConfigurationTest.java
```diff
@@ -1765,6 +1767,94 @@ public void testResolvePathInvalidBase()
...
+    public void testVirtualCallWithClassAttribute() throws Exception
+    {
+        // Executors.newSingleThreadScheduledExecutor() returns an
+        // internal class that cannot be accessed, so <Call> must
+        // specify the class to use to lookup the 'shutdown' method.
+        XmlConfiguration configuration = asXmlConfiguration(
+            "<Configure class=\"org.eclipse.jetty.xml.TestConfiguration\">" +
+            "  <Set name=\"Test\">" +
+            "    <Call class=\"java.util.concurrent.Executors\" name=\"newSingleThreadScheduledExecutor\">" +
+            "      <Call class=\"java.util.concurrent.ExecutorService\" name=\"shutdown\" />" +
```
joakime, line 1846: So the result of this `<Call name="newSingleThreadScheduledExecutor">` is a `ScheduledExecutorService` instance.

The next call, to `shutdown`, should be calling `ScheduledExecutorService.shutdown()` right?

jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlConfiguration.java
```diff
@@ -955,10 +937,6 @@ private Object call(Class<?> oClass, String methodName, Object obj, Args args) t
                 Object[] arguments = args.applyTo(method);
                 if (arguments == null)
                     continue;
-                if (Modifier.isStatic(method.getModifiers()) != (obj == null))
-                    continue;
-                if ((obj == null) && method.getDeclaringClass() != oClass)
-                    continue;
```
joakime, line 961: Is this what was preventing `testVirtualCallWithClassAttribute` from working?

jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlConfiguration.java
```diff
@@ -1177,27 +1155,39 @@ private Object propertyObj(XmlParser.Node node) throws Exception
         {
             AttrOrElementNode aoeNode = new AttrOrElementNode(node, "Id", "Name", "Deprecated", "Default");
             String id = aoeNode.getString("Id");
-            String name = aoeNode.getString("Name", true);
+            String name = aoeNode.getString("Name", false);
```
joakime, line 1158: This matches the DTD/XSD now, right?

jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlConfiguration.java
```diff
@@ -1177,27 +1155,39 @@ private Object propertyObj(XmlParser.Node node) throws Exception
...
+                    if (d == null)
+                        continue;
+                    String v = properties.get(d.toString());
                     if (v != null)
                     {
                         if (value == null)
-                            LOG.warn("Property '{}' is deprecated, use '{}' instead", d, name);
+                        {
+                            if (name == null)
+                                LOG.warn("Property '{}' is deprecated, do not use it", d);
```
joakime: s/do not use it/no replacement is available/

jetty-xml/src/main/resources/org/eclipse/jetty/xml/configure_10_0.dtd
```diff
@@ -286,7 +286,7 @@ System Property Element.
 This element allows JVM System properties to be retrieved as
 part of the value of elements such as Set, Put, Arg, etc.
 The name attribute specifies the property name and the optional
-default argument provides a default value.
+default attribute provides a default value.
```
joakime, line 289: leave this as argument, as default can be ...

``` xml
<SystemProperty name="Test">
  <Default>value</Default>
</SystemProperty>
```

jetty-xml/src/main/resources/org/eclipse/jetty/xml/configure_10_0.dtd
```diff
@@ -303,15 +303,14 @@ Environment variable Element.
 This element allows OS Environment variables to be retrieved as
 part of the value of elements such as Set, Put, Arg, etc.
 The name attribute specifies the env variable name and the optional
-default argument provides a default value.
+default attribute provides a default value.
```
joakime, line 306: Ditto, `<Default>value</Default>` is valid too.

jetty-xml/src/main/resources/org/eclipse/jetty/xml/configure_10_0.dtd
```diff
@@ -321,7 +320,7 @@ This is equivalent to:
 Property Element.
 This element allows arbitrary properties to be retrieved by name.
 The name attribute specifies the property name and the optional
-default argument provides a default value.
+default attribute provides a default value.
```
joakime, line 323: Ditto, Argument, not attribute

documentation/jetty-documentation/src/main/asciidoc/operations-guide/xml/xml-syntax.adoc
```diff
@@ -187,12 +189,38 @@ This is equivalent to the following Java code:
 var myhost = InetAddress.getByName("jdk.java.net");
 ----
 
+The `class` attribute can also be used to specify the Java class or interface to use to lookup the non-``static`` method name.
```
joakime:
```
The `class` argument ...
```
(it's not always an attribute)

documentation/jetty-documentation/src/main/asciidoc/operations-guide/xml/xml-syntax.adoc
```diff
@@ -187,12 +189,38 @@ This is equivalent to the following Java code:
...
+
+<Configure>
+  <Call class="java.util.concurrent.Executors" name="newSingleThreadScheduledExecutor">
+    #<Call class="java.util.concurrent.ExecutorService" name="shutdown" />#
+  </Call>
+</Configure>
+----
+
+In the example above, `Executors.newSingleThreadScheduledExecutor()` returns an object whose class is a private JDK implementation class.
+Without an explicit `class` attribute, it is not possible to invoke the method `shutdown()` when it is obtained via reflection from the private JDK implementation class, because while the method is `public`, the private JDK implementation class is not, therefore this exception is thrown:
```
joakime: Again, `attribute` should be `argument` here.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/xml/xml-syntax.adoc
```diff
@@ -187,12 +189,38 @@ This is equivalent to the following Java code:
...
+
+[source]
+----
+java.lang.IllegalAccessException: class org.eclipse.jetty.xml.XmlConfiguration$JettyXmlConfiguration (in module org.eclipse.jetty.xml) cannot access a member of class java.util.concurrent.Executors$DelegatedExecutorService (in module java.base) with modifiers "public"
+----
+
+The solution is to explicitly use the `class` attribute of the `<Call>` element that is invoking the `shutdown()` method, specifying a publicly accessible class or interface that the object in scope extends or implements (in the example above `java.util.concurrent.ExecutorService`).
```
joakime: s/attribute/argument/

documentation/jetty-documentation/src/main/asciidoc/operations-guide/xml/xml-syntax.adoc
```diff
@@ -212,6 +240,8 @@ If the JavaBean property is `foo` (or `Foo`), `<Get>` first attempts to invoke _
 </Configure>
 ----
 
+The `class` attribute allows to perform `static` calls, or to lookup the getter method from the specified class, as described in the xref:og-xml-syntax-call[`<Call>` section].
```
joakime: s/attribute/argument/

joakime (changes requested):

jetty-xml/src/test/java/org/eclipse/jetty/xml/XmlConfigurationTest.java
sbordet: It calls `shutdown()` on the object, so it's going to be a virtual call, not a static call.

It's not calling `ScheduledExecutorService.shutdown()` as that it's an interface method.

It's going to lookup the `shutdown` method in the `ScheduledExecutorService` interface, obtain a `Method` object, and then calling `Method.invoke()` with the object as a target of the virtual call.

jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlConfiguration.java
sbordet: Not only. There were places where if the `class` attribute was present, then `obj` was nulled out, which also I removed.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/xml/xml-syntax.adoc
sbordet: Don't like "argument", added  "(or `<Class>` element)" instead.

jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlConfiguration.java
sbordet: Yes.

jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlConfiguration.java
sbordet: Done.

jetty-xml/src/main/resources/org/eclipse/jetty/xml/configure_10_0.dtd
sbordet: "The name attribute ..." so if `name` is an attribute (but can be an element), then also `default` should use the same naming. 
Either they are both "attribute" (and I think it's ok because the first line below enumerate the possible child elements), or we use "attribute/element" everywhere.

jetty-xml/src/test/java/org/eclipse/jetty/xml/XmlConfigurationTest.java
joakime: I meant the return type of call ...

``` xml
<Call class="java.util.concurrent.Executors" name="newSingleThreadScheduledExecutor">
```

Is an instance of `java.util.concurrent.ScheduledExecutorService`, which has the `shutdown()` method.
Why is `XmlConfiguration` not using the declared return type? but is instead using the virtual / internal class?

joakime (approved):

PR #8919, opened 2022-11-20
Jetty 12 mimetype cleanup

gregw:
Cleanup a bunch of TODOs related to MimeType lookup.
Primarily `MimeTypes` are now available via the `Context`.


jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/Response.java
```diff
@@ -793,12 +793,12 @@ private String getCharacterEncoding(boolean setContentType)
             return _mimeType.getCharsetString();
 
         // Try charset assumed from content type (assumed charsets are not added to content type header).
-        encoding = MimeTypes.getCharsetAssumedFromContentType(_contentType);
+        encoding = _channel.getRequest().getCoreRequest().getContext().getMimeTypes().getCharsetAssumedFromContentType(_contentType);
```
lachlan-roberts: Can we assign `_channel.getRequest().getCoreRequest().getContext().getMimeTypes()` to local variable.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -87,9 +88,15 @@ public static String getServerInfo()
         return "jetty/" + Server.getVersion();
     }
 
-    // TODO should persistent attributes be an Attributes.Layer over server attributes?
-    private final Attributes _persistentAttributes = new Mapped();
+    /*
+     * The context and specifically attributes and mimeTypes are not implemented as a layer over
+     * The server context, attributes and mimeTypes, because ultimately we may support cross context
```
lachlan-roberts: Maybe review this comment, could use some further clarification, and `The` should not be capitalised.

lachlan-roberts (changes requested):

lachlan-roberts (approved):
LGTM

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
```diff
@@ -375,79 +247,188 @@ public static String getDefaultMimeByExtension(String filename)
                     break;
 
                 String ext = StringUtil.asciiToLowerCase(filename.substring(i + 1));
```
joakime: Use `FileID.getExtension(filename);`

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
```diff
@@ -222,145 +206,33 @@ public Type getBaseType()
...
     /**
      * Get the MIME type by filename extension.
-     * Lookup only the static default mime map.
      *
      * @param filename A file name
      * @return MIME type matching the longest dot extension of the
      * file name.
      */
-    public static String getDefaultMimeByExtension(String filename)
+    public String getMimeByExtension(String filename)
```
joakime, line 236: We should have a `getMimeByPath(Path path)` and which can use the fallback JVM mechanism.

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Files.html#probeContentType(java.nio.file.Path)

joakime (changes requested):
Minor changes

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
gregw: I think it is probably a good idea to use the prob somewhere in our resource handling, but I'm not sure that doing file ops in `MimeTypes` is the right way to go.   Maybe, but let's do that when we actually start using it. 

@lachlan-roberts can you think about this in your ResourceService work.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
```diff
@@ -202,7 +187,7 @@ public Type getBaseType()
         }
     }
 
-    public static final Index<Type> CACHE = new Index.Builder<Type>()
+    public static Index<Type> CACHE = new Index.Builder<Type>()
```
janbartel: why is this no longer final?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
```diff
@@ -222,232 +207,215 @@ public Type getBaseType()
...
     }
 
-    /**
-     * Set a mime mapping
-     *
-     * @param extension the extension
-     * @param type the mime type
-     */
-    public void addMimeMapping(String extension, String type)
+    public static MimeTypes DEFAULTS = new MimeTypes(null)
```
janbartel: final?

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/MimeTypesTest.java
```diff
@@ -30,6 +30,8 @@ public static Stream<Arguments> mimeTypesByExtensionCases()
     {
         return Stream.of(
             Arguments.of("test.gz", "application/gzip"),
+            Arguments.of("test.tar.gz", "application/gzip"),
+            Arguments.of("test.tgz", "application/x-gtar"),
```
janbartel, line 34: This is deliberate, yes? You removed type `application/x-gtar` from mime.properties ...

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/FileID.java
```diff
@@ -65,9 +65,23 @@ public static String getFileName(URI uri)
     }
 
     /**
+     * <p>
```
janbartel: Strictly doesn't have anything to do with `MimeTypes`, but seems like a good cleanup.

janbartel (approved):
Apart from a couple of finals etc. seems good.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/FileID.java
gregw: yeah, this is actually a pre-merge of #8924 (pending it going Green and being merged).  So will not be a part of this PR once merged.

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/MimeTypesTest.java
gregw: Correct.  Aligning with common extension handling practice of just looking at the final dot segment.  This is more efficient, same as most other projects and there is only one extension we had with a dot in it.

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/Response.java
```diff
@@ -792,13 +792,16 @@ private String getCharacterEncoding(boolean setContentType)
         if (_mimeType != null && _mimeType.isCharsetAssumed())
             return _mimeType.getCharsetString();
 
+        // Get MimeTypes from context
+        MimeTypes mimeTypes = _channel.getRequest().getCoreRequest().getContext().getMimeTypes();
```
joakime, line 796: That's a long way to go, isn't the Context available already from this line?

joakime (approved):

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/Response.java
gregw: No.  Need to get the context and it is only available from the request.

PR #8922, opened 2022-11-21
Issue #8921 Fix normalization of META-INF resources for quickstart

### LINKED ISSUE #8921: Normalization of resources incorrect for quickstart in jetty-12
**jetty-12**

`AttributeNormalizer` changed in jetty-12 to normalize files/uris/resources relative to the set of base resources for a webapp. Previously, normalization was only performed relative to the single "root" base resource. Resolving against all of the base resources leads to incorrect substitutions.

For example, given a fragment jar in `WEB-INF/lib/demo-web-fragment-11.0.13-SNAPSHOT.jar`, the `quickstart-web.xml` file generated:
```
  <context-param>
    <param-name>org.eclipse.jetty.resources</param-name>
    <param-value><![CDATA[
    "jar:${WAR.uri}/WEB-INF/lib/demo-web-fragment-11.0.13-SNAPSHOT.jar!/META-INF/resources"]]></param-value>
  </context-param>
```

In jetty-12 however we see the following incorrect generation:
```
  <context-param>
    <param-name>org.eclipse.jetty.resources</param-name>
    <param-value><![CDATA[]]></param-value>
  </context-param>
```

This is also seen when using the maven jetty plugin, which uses quickstart for the `forked` goals.
A fragment jar (on disk unassembled) of:
`file:///xyz/src/jetty-12.0.x/jetty-ee9/jetty-ee9-maven-plugin/target/it/jetty-effective-web-xml-it/resources/target/classes/META-INF/resources/`

is generated incorrectly as:
```
  <context-param>
    <param-name>org.eclipse.jetty.resources</param-name>
    <param-value><![CDATA[${WAR.uri}]]></param-value>
  </context-param>
```

janbartel:
Closes #8921


jetty-ee10/jetty-ee10-quickstart/src/main/java/org/eclipse/jetty/ee10/quickstart/QuickStartGeneratorConfiguration.java
```diff
@@ -167,7 +168,11 @@ public void generateQuickStartWebXml(WebAppContext context, OutputStream stream)
         // Set some special context parameters
 
         // The location of the war file on disk
-        AttributeNormalizer normalizer = new AttributeNormalizer(context.getBaseResource());
+        Resource base = (Resource)context.getAttribute(WebInfConfiguration.UNPACKED_RESOURCE_BASE);
+        //Not unpacked, pick the first base
+        if (base == null && context.getBaseResource() != null)
+            base = context.getBaseResource().iterator().next();
+        AttributeNormalizer normalizer = new AttributeNormalizer(base);
```
gregw: On consideration, I've changed my mind and I think that taking the first resource is always the best we can do.  In the case that `UNPACKED_RESOURCE_BASE` is set, is the well known case that the base resource will be well known and taking the first will work.     There are still strange edge cases if somebody sets a strange/complex baseResource, but the attribute will not help with those anyway.

gregw: What happens if there is no base resource?

gregw (changes requested):
I agree that the javadoc of resources needs to be improved.  I'll do that.

jetty-ee10/jetty-ee10-quickstart/src/main/java/org/eclipse/jetty/ee10/quickstart/QuickStartGeneratorConfiguration.java
joakime: In the strange/complex baseResource case, couldn't the user specify/override for a simpler `UNPACKED_RESOURCE_BASE` attribute to be precise on behavior?

janbartel: @joakime we've always allowed the user to specify the tmp directory, so the user has some control over where the webapp will be unpacked. However, it's not always the case that it _will_ be unpacked: it could be already unpacked or it could be a synthesis of many existing directories as in the case of the maven jetty plugin. In the case of a packed webapp, the unpacked location will always be the first base resource (there may be others in the case of fragment jars etc); in the case of the maven jetty plugin, the webapp project is the first base resource (there may be others in the case of overlaid wars etc). So, if we can guarantee that the ordering of `Resource`s is preserved, then selecting the first is reasonable enough.

gregw (approved):
thanks

PR #8927, opened 2022-11-22
Jetty 12 - Fix `GzipHandler` handling of chunked and `Content-Length`

joakime:
The implementation of `GzipResposne` (used by `GzipHandler`) doesn't guarantee `Content-Length` for simple responses the response headers.

This PR fixes `GzipResponse` deflation, and also addresses a few small bugs around `HEAD` responses during HTTP/2 and HTTP/3


jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
```diff
@@ -177,6 +271,133 @@ public void process(org.eclipse.jetty.server.Request request, Response response,
...
+        assertThat(response.getStatus(), is(HttpStatus.OK_200));
+    }
+
+    /**
+     * Test for HEAD behavior on persistent connection (not Connection: close)
+     *
+     * @throws Exception if there is a problem
+     */
+    @Test
+    public void testHEADNoConnectionClose() throws Exception
```
joakime, line 306: @gregw this is one of the HEAD testcases that is causing issues.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
```diff
@@ -177,6 +271,133 @@ public void process(org.eclipse.jetty.server.Request request, Response response,
...
+        }
+    }
+
+    /**
+     * Test for HEAD behavior on persistent connection (not Connection: close)
+     *
+     * @throws Exception if there is a problem
+     */
+    @Test
+    public void testHEADWithConnectionClose() throws Exception
```
joakime, line 344: @gregw this is one of the HEAD testcases that is causing issues.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/gzip/GzipHandlerTest.java
```diff
@@ -290,33 +302,132 @@ public void testBlockingResponse() throws Exception
         assertEquals(CONTENT, testOut.toString(StandardCharsets.UTF_8));
     }
 
+    /**
+     * Test a HEAD request (that is not processed by GZIP) then a GET request (which is compressed)
+     */
+    @Test
+    public void testHEADThenGETPersistent() throws Exception
```
joakime, line 305: @gregw this is one of the HEAD testcases that is causing issues.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/gzip/GzipHandlerTest.java
```diff
@@ -290,33 +302,132 @@ public void testBlockingResponse() throws Exception
...
+        _server.start();
+
+        // generated and parsed test
+        HttpTester.Response response;
+
+        // Request to a handler that writes a single buffer, with a valid Content-Length header, and Content-Encoding header.
+        String rawRequest = """
+            HEAD /ctx/buffer/info HTTP/1.1
+            Host: tester
+            Connection: close
```
joakime, line 366: Test without `Connection: close` as well.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipResponse.java
```diff
@@ -38,13 +39,20 @@
...
+    };
+    // Per RFC-1952, the GZIP trailer is 8 bytes
+    public static final int GZIP_TRAILER_SIZE = 8;
 
     public static final HttpField VARY_ACCEPT_ENCODING = new PreEncodedHttpField(HttpHeader.VARY, HttpHeader.ACCEPT_ENCODING.asString());
 
     private enum GZState
     {
-        MIGHT_COMPRESS, NOT_COMPRESSING, COMMITTING, COMPRESSING, FINISHED
+        MIGHT_COMPRESS, NOT_COMPRESSING, COMMITTING, COMPRESSING, FINISHING, FINISHED
```
gregw: Some javadoc against each state would be helpful (we should have had it before).

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipResponse.java
```diff
@@ -249,136 +271,162 @@ protected void onCompleteFailure(Throwable x)
...
             }
 
             // If we have no buffer
             if (_buffer == null)
             {
-                // allocate a buffer and add the gzip header.
                 _buffer = getRequest().getComponents().getByteBufferPool().acquire(_bufferSize, false);
-                BufferUtil.fill(_buffer, GZIP_HEADER, 0, GZIP_HEADER.length);
+                // Per RFC-1952, GZIP is LITTLE_ENDIAN
+                _buffer.order(ByteOrder.LITTLE_ENDIAN);
```
gregw, line 305: This is true, but do we ever do non-byte operations on the buffer?

The endian state of a buffer from the pool perhaps is something we need to reset before putting back into the pool... except that we don't do non-byte operations.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipResponse.java
```diff
@@ -249,136 +271,162 @@ protected void onCompleteFailure(Throwable x)
...
+                _buffer = null;
+            }
+        }
+
+        private int getFlushMode()
+        {
+            return _syncFlush ? Deflater.SYNC_FLUSH : Deflater.NO_FLUSH;
+        }
+
+        private Action compressing(Deflater deflater, ByteBuffer outputBuffer)
```
gregw, line 350: some javadoc to say something to the effect of "this method is called by `process` when in COMPRESSING state.  It may call `finishing` to complete the compression.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipResponse.java
gregw: Can you javadoc each of these states (should have been done before).

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipResponse.java
```diff
@@ -249,136 +271,162 @@ protected void onCompleteFailure(Throwable x)
...
+
+            // the content held by GzipBufferCB is fully consumed as input to the Deflater instance, we are done
+            if (BufferUtil.isEmpty(_content))
+                return Action.SUCCEEDED;
+
+            // No progress made on deflate, but the _content wasn't consumed, we shouldn't be able to reach this.
+            throw new AssertionError("No progress on deflate made for " + this);
+        }
+
+        private Action finishing(Deflater deflater, ByteBuffer outputBuffer)
```
gregw, line 395: Add javadoc to say something like: "this method is called by `compressing`, once the last chunk is compressed; or directly from `process` if an earlier call to `finishing` was unable to complete.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipResponse.java
```diff
@@ -249,136 +271,162 @@ protected void onCompleteFailure(Throwable x)
...
 
             // If we have no buffer
             if (_buffer == null)
             {
-                // allocate a buffer and add the gzip header.
                 _buffer = getRequest().getComponents().getByteBufferPool().acquire(_bufferSize, false);
-                BufferUtil.fill(_buffer, GZIP_HEADER, 0, GZIP_HEADER.length);
+                // Per RFC-1952, GZIP is LITTLE_ENDIAN
+                _buffer.order(ByteOrder.LITTLE_ENDIAN);
+                BufferUtil.flipToFill(_buffer);
```
gregw, line 306: This is a slightly unconventional way to use BufferUtil, i.e. to flip to fill mode and leave it that way for some time.      Can you make it clear in the javadoc and/or argument names that the `compressing` and `finishing` methods expect to be called with a buffer in fill mode.  Also a comment here about who/when/where the buffer is flipped to flush would be handy.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipResponse.java
gregw, line 305: Whilst this is true, do we ever do any non-byte operations on the buffer (e.g. adding integers or longs)?

If we start using that aspect of buffers, then we will need to reset the endian state of buffers as we recycle them through pools.

Maybe best is we just avoid non-byte operations and put some javadoc on our buffer pool saying that endianness is platform default and that any usage of a buffer that switches the endianess should revert it before putting back in the buffer.... or maybe we say that we don't guarantee the endian state of buffers from the pool?

It's a little can of worms... best just to do byte operations :)

gregw (changes requested):
Looks pretty good.  Only really javadoc niggles and a question about endianness

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipResponse.java
joakime: The buffer pool resets the order to BIG_ENDIAN on release already.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipResponse.java
joakime: I should rename `_buffer` to `_outputBuffer` too.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipResponse.java
joakime: Yes, we do multi-byte put (of ints).
And those _must_ be in LITTLE_ENDIAN mode (per spec).

There's no such thing as "platform default" for endianness in ByteBuffer, it's always BIG_ENDIAN (per [javadoc](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/ByteBuffer.html))

joakime: Each of the ByteBufferPool implementations call `BufferUtil.reset(ByteBuffer)` on release.

Which does ...

https://github.com/eclipse/jetty.project/blob/d0cf73f1c7c7a37ad006abb0a6c52d3b37d6722f/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/BufferUtil.java#L179-L187

gregw: OK, so we need to put these buffers back to `BIG_ENDIAN` before releasing them back to the pool.  We could make the pool do it, but let's not add the expense when it is only needed here.

gregw (approved):

PR #8928, opened 2022-11-23
Cleanup ContextHandler

gregw:
Extracted some of the goodness from #8793:
 + Clear enter/exit scope methods rather than opaque suppliers and Runnables
 + Removed overloading of "Context" class name to avoid accidental usage of wrong type.
 + Less holding onto request/response as fields


gregw:
@sbordet can you review this today.  I think it will make our experiments on alternate handler APIs easier if the ContextHandler is simpler.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextRequest.java
```diff
@@ -14,81 +14,52 @@
...
-            // Let's be less verbose with BadMessageExceptions & QuietExceptions
-            if (!LOG.isDebugEnabled() && (t instanceof BadMessageException || t instanceof QuietException))
-                LOG.warn("context bad message {}", t.getMessage());
-            else
-                LOG.warn("context handle failed {}", this, t);
+            Response.writeError(this, contextResponse, callback, t);
+        }
+        finally
+        {
+            _contextHandler.exitScope(this, request.getContext(), lastLoader);
```
sbordet, line 56: Is this correct, since `super.process()` is async? It may return immediately without doing anything, and then all the processing happens later, but the classloader has been unset, etc.

Point being try/finally assumes blocking code, but `process()` is not, so just asking whether this is the correct pattern.
If not, we should wrap the callback.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -166,10 +173,10 @@ public void setServer(Server server)
         super.setServer(server);
         _mimeTypes.setWrapped(server.getMimeTypes());
     }
-
-    protected Context newContext()
+    
+    protected ScopedContext newContext()
```
sbordet, line 177: Why do we need to return the implementation class type here?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -192,7 +199,7 @@ public void dump(Appendable out, String indent) throws IOException
     }
 
     @ManagedAttribute(value = "Context")
-    public ContextHandler.Context getContext()
+    public ScopedContext getContext()
```
sbordet, line 202: Why do we need to return the implementation class type here?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextRequest.java
```diff
@@ -108,7 +79,7 @@ public boolean addErrorListener(Predicate<Throwable> onError)
     }
 
     @Override
-    public ContextHandler.Context getContext()
+    public ContextHandler.ScopedContext getContext()
```
sbordet, line 84: Why do we need to return the implementation class type here?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextResponse.java
```diff
@@ -22,9 +22,9 @@
 
 public class ContextResponse extends Response.Wrapper
 {
-    private final ContextHandler.Context _context;
+    private final ContextHandler.ScopedContext _context;
 
-    public ContextResponse(ContextHandler.Context context, Request request, Response response)
+    public ContextResponse(ContextHandler.ScopedContext context, Request request, Response response)
```
sbordet, line 27: Why do we need to take the implementation class type here?

sbordet (changes requested):

lachlan-roberts (approved):
Don't like the renames of the `Context` classes much.

For me reading `ContextHandler.Context` is easier to understand than just `ScopedContext` as you know exactly what implementation it is.

sbordet (approved):
Resolved my question about enter/exit scope for the call to `super.process()`.
If we can use `o.e.j.server.Context` in signatures it's better.

gregw:
> Don't like the renames of the `Context` classes much.
> 
> For me reading `ContextHandler.Context` is easier to understand than just `ScopedContext` as you know exactly what implementation it is.

I'm also not a fan of the `ScopedContext` name.   However, it was apparent that by calling all the implementations `Context` when the interface was also called `Context`, that we were inadvertently using the implementation in places that we should just depend on the interface (see @sbordet comment above about not using `ScopedContext` when `Context` will do).

Happy for a suggestion of a different name than `ScopedContext`, but it can't be the interface name.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextResponse.java
gregw: In this case, yes we do as we call the `ScopedContext.apply` method, which is not in the Context signature.   I will check for other places that we should avoid the impl and use the interface.
Maybe also a different PR could move apply to the interface if we think it is generally useful

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextRequest.java
gregw: Not strictly, but it does avoid some instanceof checks elsewhere.
If the caller has just a `Request` reference, then they will only see the return as a `Context` . If they have a `ContextRequest` reference then they see the `ScopedContext`.    I think that is a reasonable access to the impl when needed, i.e if they are already using the impl of `ContextRequest` they get the impl of `ScopedContext` (which should perhaps be called `ContextContext` :)

PR #8931, opened 2022-11-23
Issue #7121 - Add documentation for websocket configuration.

### LINKED ISSUE #7121: Jetty 9 doc issue - WebSocket message size
**Jetty version(s)**
9

**Description**
In WebSocket Introduction I can read:
_A single message in WebSocket can be of **any size** (the underlying framing however does have a single frame limit of 63-bits). There can be an unlimited number of messages sent. Messages are sent sequentially, the base protocol does not support interleaved messages._
Link: https://www.eclipse.org/jetty/documentation/jetty-9/index.html#websocket-intro

But by default it seems to be max 64Ko (and may be modified with Annotation 
`@WebSocket(maxTextMessageSize = 64 * 1024)
`
or with Session:
```
javax.websocket.Session.setMaxBinaryMessageBufferSize(int)
javax.websocket.Session.setMaxTextMessageBufferSize(int)
```

I think it should be written that there is a default 64Ko limit.

I also checked Jetty 10 documentation: exemples are set with a 1Mo limit, but I think there is also a default one.

lachlan-roberts:
## Issue #7121
Add documentation how to configure websocket connections and the default values.


documentation/jetty-documentation/src/main/asciidoc/programming-guide/websocket.adoc
```diff
@@ -28,6 +28,31 @@ Each side of a WebSocket connection, either client or server, is made of two ent
 * A xref:pg-websocket-endpoints[WebSocket _endpoint_], the entity that _receives_ WebSocket events.
 * A xref:pg-websocket-session[WebSocket _session_], the entity that offers an API to _send_ WebSocket data (and to close the WebSocket connection), as well as to configure WebSocket connection parameters.
 
+[[pg-websocket-configuration]]
+==== WebSocket Configuration
+
+There are various different ways to configure WebSocket connections, default configuration can be applied to the `WebSocketUpgradeFilter` and the `WebSocketContainer`, this default configuration is applied before the WebSocket is given to the application in the `onOpen()` notification, this configuration can further be modified in `onOpen()` and any time after using the configuration on the `Session`.
```
sbordet: Sorry, but I cannot parse and understand this sentence.
I don't know what you want to say, but you need to be concrete.
You want to talk about WebSocket `Session` not "connection" since there is no "connection" concept in the API.
The section before this one you added introduces the terms "endpoint" and "session" because they are used later, so stick with those -- "connection" is not defined.
You want to explicitly say that you can configure the `Session` object, and with what methods, otherwise it seems that there is a "configuration" object.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/websocket.adoc
```diff
@@ -28,6 +28,31 @@ Each side of a WebSocket connection, either client or server, is made of two ent
...
+
+Settings which can be configured include:
+
+* MAX_BINARY_MESSAGE_SIZE - The maximum size of a binary message which can be received. (default: 64 * 1024)
```
sbordet: These capital letters things are undefined.
You want to refer to explicit methods on the `Session` object, or otherwise a reader would have no idea _where_ to configure the maxBinaryMessageSize.

Same with the others below.

Also, no need of an empty line between the bullets.

See also the bullet list in the previous section: use:

* "definition", "explanation" 

don't use the single dash, if you want separation use either the comma or the [em dash](https://www.thepunctuationguide.com/em-dash.html).

sbordet (changes requested):

documentation/jetty-documentation/src/main/asciidoc/programming-guide/websocket.adoc
lachlan-roberts: @sbordet is there any way I can link java constants defined in `org.eclipse.jetty.websocket.core.WebSocketConstants` in case the defaults change?

documentation/jetty-documentation/src/main/asciidoc/programming-guide/websocket.adoc
lachlan-roberts: There are many places to configure these things so its hard to be super specific without greatly expanding this section.  There is like 10 different classes used to configure these things between Javax, Jetty, and Core layers. The on the filter, session, container, handler, RemoteEndpoint, through init params and methods.

The idea was to give some overview of what the configuration is available, not the definitive way to configure it. But I will rewrite this to be more clear.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/websocket.adoc
sbordet: @lachlan-roberts constant field values are reported in Javadocs, see for example:
https://www.eclipse.org/jetty/javadoc/jetty-11/constant-values.html#org.eclipse.jetty.websocket.core.WebSocketConstants.DEFAULT_AUTO_FRAGMENT

Again, in the documentation I would refer to the javadocs for further details, unless it's really important or you want to explain what happens if e.g. a message is too large and you need a specific value.

lachlan-roberts:
@sbordet I have added some further clarification for the docs.

But to be specific with the exact configuration methods, I will need to split it out into 3 different sections for each of the websocket APIs.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/WebSocketConstants.java
```diff
@@ -22,14 +22,49 @@
...
     public static final int DEFAULT_MAX_TEXT_MESSAGE_SIZE = 64 * 1024;
+
+    /**
+     * Max Frame Size - The maximum payload size of any WebSocket Frame which can be received.
+     */
     public static final int DEFAULT_MAX_FRAME_SIZE = 64 * 1024;
+
+    /**
+     * Output Buffer Size - The output (write to network layer) buffer size. This is the raw write operation buffer size and has no relationship to the websocket frame.
+     */
```
sbordet, line 42: Wrong javadoc? The constant is called _input_ but the javadoc refers to _output_.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/WebSocketConstants.java
```diff
@@ -22,14 +22,49 @@
...
     public static final int DEFAULT_INPUT_BUFFER_SIZE = 4 * 1024;
+
+    /**
+     * Input Buffer Size - The input (read from network layer) buffer size. This is the raw read operation buffer size, before the parsing of the websocket frames.
+     */
```
sbordet, line 47: Ditto.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/WebSocketConstants.java
```diff
@@ -22,14 +22,49 @@
...
     public static final boolean DEFAULT_AUTO_FRAGMENT = true;
+
+    /**
+     * Idle Timeout - The duration that a websocket may be idle before being closed by the implementation.
+     */
     public static final Duration DEFAULT_IDLE_TIMEOUT = Duration.ofSeconds(30);
+
+    /**
+     * Write Timeout - The maximum time a frame may be waiting to be sent.
```
sbordet, line 66: Add the fact that if <= 0 then it's an infinite timeout.

sbordet (changes requested):
@lachlan-roberts I don't think you should document all 3, here we just want to document the Jetty APIs.
Jakarta we don't document, and Core is not at all documented for now, so I'd leave it out of this PR anyway.

Therefore, I would be explicit on the APIs and add examples.

sbordet (approved):

PR #8933, opened 2022-11-24
Mutable server MimeTypes

gregw:
Make the MimeTypes available from the server mutable. Context MimeTypes are now wrappers over the server MimeTypes, so configuration at the server level is inherited by all contexts.

### DIFF
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
@@ -96,7 +96,7 @@ public class ContextHandler extends Handler.Wrapper implements Attributes, Grace
      */
     private final Context _context;
     private final Attributes _persistentAttributes = new Mapped();
-    private final MimeTypes.Mutable _mimeTypes = new MimeTypes.Mutable();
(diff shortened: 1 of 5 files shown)


jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -160,11 +160,23 @@ public ContextHandler(Handler.Container parent, String contextPath)
             addAliasCheck(new SymlinkAllowedResourceAliasChecker(this));
     }
 
+    @Override
+    public void setServer(Server server)
+    {
+        super.setServer(server);
+        _mimeTypes.setWrapped(server.getMimeTypes());
```
joakime, line 167: Does this mean `_mimeTypes` isn't available to modify until the server is started? 
_(as the `setServer(Server)` call usually only happens in a doStart() call.)_

gregw: No,
all modifications go into the context wrapper, before or after the wrapped types are set.

It just means that the defaults set in the server are not visible until the server is set.   
So context MimeTypes will initially look empty, but if you set some types, you will see exactly only those types... until the server is set, at which time you will also see all the server types... except the context types always override server types.

olamy (approved):

PR #8934, opened 2022-11-24
Jetty-12 Rewrite RuleProcessor

gregw:
This extracts some goodness from the closed  #8793.
This PR decouples the rewrite module from the `WrapperProcessor` class, which is being considered for significant refactoring or removal. Having a module specific version of that class allows better code readability and a more appropriate API that avoids duplication request instances.

### DIFF
--- a/jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/CookiePatternRule.java
+++ b/jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/CookiePatternRule.java
@@ -76,7 +76,7 @@ public class CookiePatternRule extends PatternRule
     }
 
     @Override
-    public Request.WrapperProcessor apply(Request.WrapperProcessor input) throws IOException
+    public RuleProcessor apply(RuleProcessor input) throws IOException
     {
         // TODO: fix once Request.getCookies() is implemented (currently always returns null)
         // Check that cookie is not already set
@@ -90,13 +90,13 @@ public class CookiePatternRule extends PatternRule
             }
         }
 
-        return new Request.WrapperProcessor(input)
+        return new RuleProcessor(input)
         {
             @Override
-            public void process(Request ignored, Response response, Callback callback) throws Exception
+            public void process(Response response, Callback callback) throws Exception
             {
                 Response.addCookie(response, new HttpCookie(_name, _value));
-                super.process(this, response, callback);
+                super.process(response, callback);
             }
         };
     }
--- a/jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/CookiePatternRule.java
+++ b/jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/CookiePatternRule.java
@@ -76,9 +76,8 @@ public class CookiePatternRule extends PatternRule
     }
 
     @Override
-    public RuleProcessor apply(RuleProcessor input) throws IOException
+    public Processor apply(Processor input) throws IOException
     {
-        // TODO: fix once Request.getCookies() is implemented (currently always returns null)
         // Check that cookie is not already set
         List<HttpCookie> cookies = Request.getCookies(input);
         if (cookies != null)
@@ -90,7 +89,7 @@ public class CookiePatternRule extends PatternRule
             }
         }
 
-        return new RuleProcessor(input)
+        return new Processor(input)
         {
             @Override
             public void process(Response response, Callback callback) throws Exception
--- a/jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/HeaderPatternRule.java
+++ b/jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/HeaderPatternRule.java
@@ -73,9 +73,9 @@ public class HeaderPatternRule extends PatternRule
     }
 
     @Override
-    public RuleProcessor apply(RuleProcessor input) throws IOException
+    public Processor apply(Processor input) throws IOException
     {
-        return new RuleProcessor(input)
+        return new Processor(input)
         {
             @Override
             public void process(Response response, Callback callback) throws Exception
--- a/jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/HeaderPatternRule.java
+++ b/jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/HeaderPatternRule.java
@@ -15,7 +15,6 @@ package org.eclipse.jetty.rewrite.handler;
 
 import java.io.IOException;
 
-import org.eclipse.jetty.server.Request;
 import org.eclipse.jetty.server.Response;
 import org.eclipse.jetty.util.Callback;
 import org.eclipse.jetty.util.annotation.Name;
@@ -74,18 +73,18 @@ public class HeaderPatternRule extends PatternRule
     }
 
     @Override
-    public Request.WrapperProcessor apply(Request.WrapperProcessor input) throws IOException
+    public RuleProcessor apply(RuleProcessor input) throws IOException
     {
-        return new Request.WrapperProcessor(input)
+        return new RuleProcessor(input)
         {
             @Override
-            public void process(Request ignored, Response response, Callback callback) throws Exception
+            public void process(Response response, Callback callback) throws Exception
             {
                 if (isAdd())
                     response.getHeaders().add(getHeaderName(), getHeaderValue());
                 else
                     response.getHeaders().put(getHeaderName(), getHeaderValue());
-                super.process(ignored, response, callback);
+                super.process(response, callback);
             }
         };
     }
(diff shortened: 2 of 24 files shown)


jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/Rule.java
```diff
@@ -55,4 +58,116 @@ public String toString()
...
+    /**
+     * A tuple of a {@link Request.Wrapper} and is also a {@link org.eclipse.jetty.server.Request.Processor},
+     * that is used to chain a sequence of {@link Rule}s together.
+     * The tuple is initialized with only the request, then the processor is
+     * then passed to a chain of rules before the ultimate processor is
+     * passed in {@link #wrapProcessor(Processor)}. Finally, the response
+     * and callback are provided in a call to {@link #process(Request, Response, Callback)},
+     * which calls the {@link #process(Response, Callback)}.
+     */
+    public static class RuleProcessor extends Request.Wrapper implements Request.Processor
```
sbordet: I'd just call it `Processor`, as it is enclosed in `Rule` already, so we don't duplicate "Rule" as in `Rule.RuleProcessor`.

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/Rule.java
```diff
@@ -55,4 +58,116 @@ public String toString()
...
+                wrapped = wrapped instanceof Request.Wrapper w ? w.getWrapped() : null;
+            }
+
+            // We need a new instance of the wrapper
+            Request wrapper = wrapRequest(request);
+            process(wrapResponse(wrapper, response), callback);
+        }
+    }
+
+    public static class UriRuleProcessor extends RuleProcessor
```
sbordet: Rename to `HttpURIRuleProcessor` as everywhere else it's `HttpURI`.

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/Rule.java
```diff
@@ -55,4 +58,116 @@ public String toString()
...
     }
+
+    /**
+     * A tuple of a {@link Request.Wrapper} and is also a {@link org.eclipse.jetty.server.Request.Processor},
+     * that is used to chain a sequence of {@link Rule}s together.
+     * The tuple is initialized with only the request, then the processor is
+     * then passed to a chain of rules before the ultimate processor is
+     * passed in {@link #wrapProcessor(Processor)}. Finally, the response
+     * and callback are provided in a call to {@link #process(Request, Response, Callback)},
+     * which calls the {@link #process(Response, Callback)}.
```
sbordet: Javadoc needs `<p>` to wrap paragraphs.

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/Rule.java
```diff
@@ -55,4 +58,116 @@ public String toString()
...
+        }
+
+        @Override
+        public void process(Request request, Response response, Callback callback) throws Exception
+        {
+            assert getWrapped() == request;
+            process(response, callback);
+        }
+
+        /** Process this wrapped request together with the passed response and
```
sbordet: Don't write test in the opening of javadocs, should be:

```java
/**
 * Process
```

Also, javadocs should be written in 3rd person, so "Processes".

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/CookiePatternRule.java
```diff
@@ -76,7 +76,7 @@ public void setValue(String value)
     }
 
     @Override
-    public Request.WrapperProcessor apply(Request.WrapperProcessor input) throws IOException
+    public RuleProcessor apply(RuleProcessor input) throws IOException
     {
         // TODO: fix once Request.getCookies() is implemented (currently always returns null)
         // Check that cookie is not already set
```
sbordet, line 81: This TODO still necessary?

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/Rule.java
```diff
@@ -55,4 +58,116 @@ public String toString()
...
+                    return;
+                }
+                wrapped = wrapped instanceof Request.Wrapper w ? w.getWrapped() : null;
+            }
+
+            // We need a new instance of the wrapper
+            Request wrapper = wrapRequest(request);
+            process(wrapResponse(wrapper, response), callback);
+        }
+    }
```
sbordet, line 114: Ugh, no.
`RuleProcessor` is 99% of the times an on-the-fly object that is never reused.
Only the last one returned from the `RewriteHandler` may be reused.
However, there is currently no reuse, and we have not decided whether reusing `Request.Processor` is even a good idea, so I would not introduce it here.

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/HeaderPatternRule.java
```diff
@@ -74,18 +73,18 @@ public void setAdd(boolean add)
...
             @Override
-            public void process(Request ignored, Response response, Callback callback) throws Exception
+            public void process(Response response, Callback callback) throws Exception
             {
                 if (isAdd())
                     response.getHeaders().add(getHeaderName(), getHeaderValue());
                 else
                     response.getHeaders().put(getHeaderName(), getHeaderValue());
-                super.process(ignored, response, callback);
+                super.process(response, callback);
```
sbordet, line 87: Nice! This seems to me all that's needed 😬

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/Rule.java
```diff
@@ -55,4 +58,116 @@ public String toString()
     {
         return "%s@%x[terminating=%b]".formatted(getClass().getSimpleName(), hashCode(), isTerminating());
     }
+
+    /**
+     * A tuple of a {@link Request.Wrapper} and is also a {@link org.eclipse.jetty.server.Request.Processor},
```
sbordet: It's not really a tuple like a Java `record` is. Maybe just "A Request.Wrapper that is also a Request.Processor".

sbordet (changes requested):
I like the `process(Response, Callback)` method.

All the rest I am not sure.
Not sure whether having 2 abstractions that represent the same thing (a request wrapper that is also a request processor) makes sense.
Definitely not sold about reusability that I would remove for now from this PR until we settle on a decision.

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/Rule.java
gregw: This is not re-wrapping the request for reuse.  This is wrapping the response if the Rule wishes to do so?

Oh I see, I still have the reuse code copied/pasted over.    I'll get rid of that.... or maybe leave as an assert.

gregw:
@sbordet I removed all the residual re-use code (that was unintentional included by refactor).

I don't mind having two abstractions for Request + Processor, as I believe they will diverge as we clean up the issues in the Handler tree.
If they don't then, it will be trivial to change the parent of `Rule.Processor` to the common one.   For now, the separation will allow us to experiment with Handlers without constantly breaking/fixing rewrite.

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/Rule.java
```diff
@@ -55,4 +58,81 @@ public String toString()
...
+
+        public Processor(Request request)
+        {
+            super(request);
+        }
+
+        @Override
+        public void process(Request request, Response response, Callback callback) throws Exception
+        {
+            assert getWrapped() == request;
```
sbordet: Still needed?

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/Rule.java
```diff
@@ -55,4 +58,81 @@ public String toString()
...
+        /**
+         * <p>Wraps the given {@code Processor} within this instance and returns this instance.</p>
+         *
+         * @param processor the {@code Processor} to wrap
+         * @return this instance
+         */
+        public Processor wrapProcessor(Processor processor)
+        {
+            _processor = processor;
+            return processor == null ? null : this::processWithWrappers;
```
sbordet: Can't you just return `this`?

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/Rule.java
```diff
@@ -55,4 +58,81 @@ public String toString()
...
+        }
+
+        protected void processWithWrappers(Request ignored, Response response, Callback callback) throws Exception
```
sbordet: Can you get rid of this method?

sbordet (approved):
LGTM, just clean up `Rule.Processor` as per comments.

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/Rule.java
gregw: No. and it is actually wrong !

PR #8940, opened 2022-11-28
Converted writeTrailers to a static method

gregw:
Fixed TODO regarding making writeTrailers a static method.

### DIFF
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
@@ -69,7 +69,7 @@ public interface Response extends Content.Sink
      * <p>Returns a chunk processor suitable to be passed to the
      * {@link Content#copy(Content.Source, Content.Sink, Content.Chunk.Processor, Callback)}
      * method, that will handle {@link Trailers} chunks
-     * by adding the their fields to the {@link HttpFields} supplied by
+     * by adding their fields to the {@link HttpFields} supplied by
      * {@link Response#getTrailersSupplier()}.</p>
      * <p>This is specifically useful for writing trailer that have been received via
      * the {@link Content.Source#read()} API, for example when echoing a request to a response:</p>
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
@@ -66,12 +66,12 @@ public interface Response extends Content.Sink
     CompletableFuture<Void> writeInterim(int status, HttpFields headers);
 
     /**
-     * <p>Return a representation of the response as a chunk handler suitable for being
-     * passed to the {@link Content#copy(Content.Source, Content.Sink, Content.Chunk.Processor, Callback)}
-     * method.  Specifically, as a chunk handler that will handle {@link Trailers} chunks
-     * by adding the chunk fields to the {@link HttpFields} supplied by
+     * <p>Return chunk processor suitable to be passed to the
+     * {@link Content#copy(Content.Source, Content.Sink, Content.Chunk.Processor, Callback)}
+     * method, that will handles {@link Trailers} chunks
+     * by adding the their fields to the {@link HttpFields} supplied by
      * {@link Response#getTrailersSupplier()}.</p>
-     * <p>This is specifically useful for writing chunks that have been received via
+     * <p>This is specifically useful for writing trailer that have been received via
      * the {@link Content.Source#read()} API, for example when echoing a request to a response:</p>
      * <pre>
      *   Content.copy(request, response, Response.asTrailerChunkHandler(response), callback);
@@ -79,7 +79,7 @@ public interface Response extends Content.Sink
      * @param response The response for which to process trailer chunks.
      *                 If the {@link Response#setTrailersSupplier(Supplier)}
      *                 method has not been called prior to this method, then a noop processor is returned.
-     * @return A chunk handler that will add trailer chunks to the response's trailer supplied fields.
+     * @return A chunk processor that will add trailer chunks to the response's trailer supplied fields.
      * @see Content#copy(Content.Source, Content.Sink, Content.Chunk.Processor, Callback)
      * @see Trailers
      */
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
@@ -76,7 +76,7 @@ public interface Response extends Content.Sink
      * <pre>
      *   Content.copy(request, response, Response.asTrailerChunkHandler(response), callback);
      * </pre>
-     * @param response The response for which to process trailer chunks.
+     * @param response The response for which to process a trailers chunk.
      *                 If the {@link Response#setTrailersSupplier(Supplier)}
      *                 method has not been called prior to this method, then a noop processor is returned.
      * @return A chunk processor that will add trailer chunks to the response's trailer supplied fields.
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
@@ -16,6 +16,7 @@ package org.eclipse.jetty.server;
 import java.nio.ByteBuffer;
 import java.util.ListIterator;
 import java.util.concurrent.CompletableFuture;
+import java.util.function.BiPredicate;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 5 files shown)


jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -65,24 +66,46 @@ public interface Response extends Content.Sink
...
-            if (requestTrailers != null)
+            if (chunk instanceof Trailers trailers)
             {
-                Supplier<HttpFields> supplier = getTrailersSupplier();
-                if (supplier != null)
+                HttpFields requestTrailers = trailers.getTrailers();
+                if (requestTrailers != null)
                 {
                     HttpFields responseTrailers = supplier.get();
                     if (responseTrailers instanceof HttpFields.Mutable mutable)
```
lorban, line 101: You should move this `instanceof` check out of the lambda as when it is false, you can return a `BiPredicate` that always returns false.

lorban (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -65,24 +66,46 @@ public interface Response extends Content.Sink
...
+     * the {@link Content.Source#read()} API, for example when echoing a request to a response:</p>
+     * <pre>
+     *   Content.copy(request, response, Response.asTrailerChunkHandler(response), callback);
+     * </pre>
+     * @param response The response for which to process trailer chunks.
+     * @return A chunk handler that will add trailer chunks to the response's trailer supplied fields.
+     * @see Content#copy(Content.Source, Content.Sink, BiPredicate, Callback)
+     * @see Trailers
+     */
+    static BiPredicate<Content.Chunk, Callback> asTrailerChunkHandler(Response response)
```
sbordet: Bit of a mouthful. Can't we just keep it named `writeTrailers`?
Or perhaps `trailersWriter` to indicate that it returns a writer that will write the trailers?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
sbordet: @lorban it cannot be moved outside, because you want to call `supplier.get()` as late as possible.

sbordet (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
lorban: Ah, right. Since I've been caught by this, it may be worth a comment?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
gregw: I think the new name is good:
 + It does not write trailers, it copies them from one HttpFields to another, to be written at a later time.
 + If it was just writing/setting trailers, then the signature would be a lot different.  It has a very specific signature so that it can be used for a very specific purpose: to be a chunk handler for a copy operation.
 + It could perhaps just be `asChunkHandler`, so that in future other types of chunk might be handled, but I doubt we will ever do that.  Having a name that tells us it is used for trailers will help coders looking for a way to handle `Trailers` chunks

lorban (approved):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
sbordet: `Handler` is so overloaded...

Taking into account your points, how about `Response.trailersCopier(Response)`?

The "as" in front should be dropped as it is not "converting" the response to something else -- I would tolerate "get" as in `getTrailersCopier`, but then it creates a new one every time, so `newTrailersCopier`, but then dropping the prefix make it nicer:

`Content.copy(request, response, Response.trailersCopier(response), callback);`

gregw: But the argument in the copy signature is called `chunkHandler`, so the names have to match (and that is your name choice in Content!).
I could rename the arg to `chunkTransformer`? but then this isn't transforming it?  or `chunkProcessor`, but then `Processor` is overloaded as well.

But regardless,  this PR is just about fixing a TODO in the `Response` class.   If you want to change the name of the `chunkHandler` in the `Content` class, then do that in a different PR?

gregw:
closed as we couldn't agree on a name.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -65,24 +65,49 @@ public interface Response extends Content.Sink
 
     CompletableFuture<Void> writeInterim(int status, HttpFields headers);
 
-    // TODO: make it static, otherwise we must override it in Wrapper.
-    default boolean writeTrailers(Content.Chunk chunk, Callback ignored)
+    /**
+     * <p>Return a representation of the response as a chunk handler suitable for being
```
sbordet:
```suggestion
     * <p>Returns a chunk processor suitable for being
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -65,24 +65,49 @@ public interface Response extends Content.Sink
...
+     * passed to the {@link Content#copy(Content.Source, Content.Sink, Content.Chunk.Processor, Callback)}
+     * method.  Specifically, as a chunk handler that will handle {@link Trailers} chunks
```
sbordet:
```suggestion
     * method, that handles {@link Trailers} chunks
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -65,24 +65,49 @@ public interface Response extends Content.Sink
...
+     * by adding the chunk fields to the {@link HttpFields} supplied by
+     * {@link Response#getTrailersSupplier()}.</p>
+     * <p>This is specifically useful for writing chunks that have been received via
```
sbordet:
```suggestion
     * <p>This is specifically useful for writing trailers that have been received via
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -65,24 +65,49 @@ public interface Response extends Content.Sink
...
+     * the {@link Content.Source#read()} API, for example when echoing a request to a response:</p>
+     * <pre>
+     *   Content.copy(request, response, Response.asTrailerChunkHandler(response), callback);
+     * </pre>
+     * @param response The response for which to process trailer chunks.
+     *                 If the {@link Response#setTrailersSupplier(Supplier)}
+     *                 method has not been called prior to this method, then a noop processor is returned.
+     * @return A chunk handler that will add trailer chunks to the response's trailer supplied fields.
```
sbordet:
```suggestion
     * @return A chunk processor that will copy the chunk trailers to the response's trailers.
```

sbordet (approved):
Looks fiiine 🥲, just update the javadocs as suggested 😄

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -65,24 +65,49 @@ public interface Response extends Content.Sink
 
     CompletableFuture<Void> writeInterim(int status, HttpFields headers);
 
-    // TODO: make it static, otherwise we must override it in Wrapper.
-    default boolean writeTrailers(Content.Chunk chunk, Callback ignored)
+    /**
+     * <p>Return chunk processor suitable to be passed to the
```
sbordet:
```suggestion
     * <p>Returns a chunk processor suitable to be passed to the
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -65,24 +65,49 @@ public interface Response extends Content.Sink
...
+     * {@link Content#copy(Content.Source, Content.Sink, Content.Chunk.Processor, Callback)}
+     * method, that will handles {@link Trailers} chunks
```
sbordet:
```suggestion
     * method, that will handle {@link Trailers} chunks
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -65,24 +65,49 @@ public interface Response extends Content.Sink
...
+     * by adding the their fields to the {@link HttpFields} supplied by
```
sbordet:
```suggestion
     * by adding their fields to the {@link HttpFields} supplied by
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -65,24 +65,49 @@ public interface Response extends Content.Sink
...
+     * {@link Response#getTrailersSupplier()}.</p>
+     * <p>This is specifically useful for writing trailer that have been received via
```
sbordet:
```suggestion
     * <p>This is specifically useful for writing trailers that have been received via
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -65,24 +65,49 @@ public interface Response extends Content.Sink
...
+     * the {@link Content.Source#read()} API, for example when echoing a request to a response:</p>
+     * <pre>
+     *   Content.copy(request, response, Response.asTrailerChunkHandler(response), callback);
+     * </pre>
+     * @param response The response for which to process trailer chunks.
```
sbordet:
```suggestion
     * @param response The response for which to process a trailers chunk.
```

sbordet (approved):
@gregw more javadoc fixes.

PR #8941, opened 2022-11-28
Fixes #8938 - Improve API compatibility between Jetty 11 and Jetty 12…

### LINKED ISSUE #8938: API compatibility between Jetty 11 and 12: org.eclipse.jetty.server.Request/Response#getHeaders()
For Spring Framework 6.0's core Jetty integration in WebFlux, we interact with `org.eclipse.jetty.server.Request#getHttpFields()` and `org.eclipse.jetty.server.Response#getHttpFields()` in Jetty 11 which changed to `#getHeaders()` in Jetty 12.

Trying to avoid reflection for supporting both versions of Jetty, is there any chance you could keep providing `#getHttpFields()` in Jetty 12 (in deprecated form) or alternatively - even preferably? - introduce `#getHeaders()` in a Jetty 11.0.x release as well, allowing integrators like Spring to have a single method to call between Jetty 11.0.x and 12 there?

sbordet:
… for Spring.

Introduced Request.getHeaders() and Response.getHeaders().

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
@@ -242,6 +242,11 @@ public class Request implements HttpServletRequest
 
     public HttpFields getHttpFields()
     {
+        return getHeaders();
+    }
+
+    public HttpFields getHeaders()
+    {
         return _httpFields;
     }
 
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
@@ -1387,6 +1387,11 @@ public class Response implements HttpServletResponse
 
     public HttpFields.Mutable getHttpFields()
     {
+        return getHeaders();
+    }
+
+    public HttpFields.Mutable getHeaders()
+    {
         return _fields;
     }
 


jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -241,6 +241,11 @@ public Request(HttpChannel channel, HttpInput input)
     }
 
     public HttpFields getHttpFields()
+    {
+        return getHeaders();
+    }
+
+    public HttpFields getHeaders()
```
gregw, line 248: Why keep both methods?   If `getHeaders` is a better name, let's just use it.
If we really REALLY need both, then we need comments/javadoc saying why.

gregw (changes requested):
Can we just have one method.... or if we really need two, can we have javadoc and/or comments

jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
sbordet: @gregw this PR is for 10+ so in 10/11 we need to keep both methods, but in 12 we only have `getHeaders()` already.

sbordet:
As discussed in #8938, there is no need for this PR.

Issue #8895, opened 2022-11-14
Generate downloadable version of javadocs documentation in website deploy script


boughtonp:
**Jetty version(s)**
all

**Description**

As per discussion on mailing list, please update the script that generates https://www.eclipse.org/jetty/javadoc/ to also create a downloadable archive for current/supported versions of Jetty.

joakime:
Most IDEs will provide the javadocs natively if you are using the Jetty dependencies from any build tool (of many) or build repository (of many).

The javadoc's are already available on Maven Central for every artifact that is present on Maven Central  - a feature that is mandated by Maven Central btw.

Our aggregate javadoc does not contain everything that exist as produced by Jetty on Maven Central.
The javadoc on https://www.eclipse.org/jetty/javadoc/ is no exception to this.

  **The only true javadoc is the per-artifact javadoc which is based on what features you actually use from Jetty.**

See the list of exclusions in the aggregate javadoc build.
https://github.com/eclipse/jetty.project/blob/jetty-11.0.12/javadoc/pom.xml#L79-L123

Nobody (and I do mean nobody) can use 100% of Jetty, it's an impossible scenario that would cause conflicts at startup time.
There's no such thing as an "all" or "full" or "complete" Jetty release, so as a consequence there's no such thing as a complete javadoc either (the javadoc tooling in Java will actually fail if this is attempted)

joakime:
This will become even more ridiculous with Jetty 12.
I don't think we can have the ENV sub-sections present in the same Javadoc at all (too many conflicting names at different versions).

joakime:
@jmcc0nn3ll its a shame we can't just deploy the `/javadoc/` project to maven central (the maven central rules will reject it as it contains references to classes that are not in the non-javadoc jar)

joakime:
@jmcc0nn3ll we have a [`/org/eclipse/jetty/documentation/` groupId](https://repo1.maven.org/maven2/org/eclipse/jetty/documentation/) already, perhaps we should revive it and have it be a pom project that just builds a zip including the adoc and javadoc aggregate during the build?

boughtonp:
This isn't one of those "I want full Jetty" requests.

I use jetty-home (in home/base setup), and when I'm trying to do something new I first check the regular documentation, but this is frequently incomplete or references the relevant class in the online javadoc, where further details can often by found.

Thus I want to also be able to read those pages when offline, and assumed there would already be an archive to download I just wasn't finding.

[...]

PR #8977, opened 2022-11-29
Issue #8895 - delete javadoc sources

sbordet:
Delete java source files used to generate javadocs, so that IDEs do not report duplicate sources.

### DIFF
--- a/javadoc/pom.xml
+++ b/javadoc/pom.xml
@@ -53,7 +53,7 @@
         <artifactId>maven-antrun-plugin</artifactId>
         <executions>
           <execution>
-            <id>init-jetty-sources-dir</id>
+            <id>create-sources-dir</id>
             <phase>generate-sources</phase>
             <goals>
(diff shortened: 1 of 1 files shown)


javadoc/pom.xml
```diff
@@ -64,6 +64,18 @@
...
           </execution>
+          <execution>
+            <id>delete-sources-dir</id>
+            <phase>verify</phase>
+            <goals>
+              <goal>run</goal>
+            </goals>
+            <configuration>
+              <target>
+                <delete dir="${sources-directory}" />
```
joakime: This kind of fast-delete will cause problems on windows based builds.

Use this ..
``` xml
<delete dir="${sources-directory}" failonerror="false" />
```

as not being able to delete the directory isn't a fatal build issue.

joakime (changes requested):
Minor change for windows builds

joakime (approved):

PR #8978, opened 2022-11-29
Jetty 12 handler as processor #8929

### LINKED ISSUE #8929: Jetty-12 Review Handler/Processor design
**Jetty version(s)**
12

**Background**

This issue is to discuss some alternatives to #8793, which whilst initially inspired by reusable Processors, was ultimately a PR about cleaning up some strange aspects of the current jetty-12 API.

**The Current Design**

The current design of `Handler` and `Request.Processor` is designed to:
 + allow for a blocking phase of calling the `Handler` tree to determine which handler (if any) would handle the request.  Instead of calling `request.setHandled(true)`  as in previous jetty versions, acceptance of request handling is now indicated by returning a `Request.Processor`.
+ only enable the full request API (e.g. read and demand) only once the request has been accepted (by returning a `Request.Processor`)
+ only provide the response and callback once the request is accepted, via an async call to `process(request, response, callback);`

**The Problem(s)**

+ Any `Handler` that needs to wrap the request, must also wrap the `Request.Processor` only so it can capture and re-use the request (wrapped in `handle()`) in `process()`.  The extra allocation can be avoided by having a `WrapperProcessor` that is-a `Request.Wrapper` that also implements `Request.Processor`, but this results in further complication and obscure code. 
 + Because the request instance is passed to both `Handler.handle(Request)` and `Request.Processor(Request, Response, Callback)`, there are often cases where there are two request instances and one must be ignored. We have already seen bugs where the wrong request has been used.
 + To avoid the allocation of a `Request.Processor` wrapper, some handlers use the `WrapperProcessor` the processors they return are one use only and cannot be reused for other requests (even though the signature looks like it can be).   This results in two classes of `Processor`, those that must be able to be reused (like those used as error processors) and those that must not be reused (like some returned from `handle()`).    
 + It is often unclear what handling of the request should be done in `handle()` vs `process()`.  For example, `GzipHandler` does some handling in `handle()`, more in a processor wrapper and more in the wrapped request/response/callback.

**Discussion**

There are at least two styles of Handler: ones that wish to process the request; ones that wish to wrap the request that will be processed by another handler:

```java
    public class ExampleProcessingHandler extends Handler.Abstract implements Request.Processor
    {
        @Override
        public Request.Processor handle(Request ignored) throws Exception
        {
            return this;
        }

        @Override
        public void process(Request request, Response response, Callback callback) throws Exception
        {
            // process the request, produce the response and complete the callback.
            callback.succeeded();
        }
    }

    public class ExampleWrappingHandler extends Handler.Wrapper
    {
        @Override
        public Request.Processor handle(Request request) throws Exception
        {
            Request.Wrapper wrapper = new Request.Wrapper(request);
            Request.Processor processor = super.handle(wrapper);
            if (processor == null)
                return null;
            // need to wrap/capture processor, just to remember the wrapper
            return (ignored, response, callback) ->
            {
                assert wrapper.getWrapped() == ignored;
                processor.process(wrapper, response, callback);
            };
        }
    }
```

Observations:
 + The `ExampleProcessingHandler` is simple enough, although it ignores the request passed to `handle()`.
 + The `ExampleWrappingHandler` must wrap the `Request.Processor` even though it has no processing of its own to do. The processor is wrapped only so that the wrapped request instance can be remembered and substituted for the ignored request passed to `process()`.  We could do the assert, but we never do (and that is a band-aid anyway).
 + We have replaced the simple `setHandled(true)` as a means of accepting the request with the return of a `Request.Processor`, which is a new entity that needs to be wrapped that is in addition to the `Handler`, `Request`, `Response` and `Callback`, which all can be wrapped.  In an attempt to reduce allocations, we make either the `Handler` or the `Request` implement `Processor`, but each results in an extra request object that needs to be ignored.

**Alternative 0**
Keep the design, but improve convenience implementations to hide the duplicate request:

```java
    public class ExampleProcessingHandler extends Handler.Processor
    {
        @Override
        public void process(Request request, Response response, Callback callback) throws Exception
        {
            // process the request, produce the response and complete the callback.
            callback.succeeded();
        }
    }

    public class ExampleWrappingHandler extends Handler.Wrapper
    {
        @Override
        public Request.Processor handle(Request request) throws Exception
        {
            Request.WrapperProcessor wrapper = new Request.WrapperProcessor(request)
            return wrapper.wrapProcessor(super.handle(wrapper));
        }
    }
```
This has already been done to some extent, yet it doesn't need an example to be much more complex than these simple ones before the duplicate re-appears. Perhaps more can be done.

**Alternative 1**

One way to avoid having an ignored request is to get rid of the second request object by simply not passing it in `process`:

```java
    public class ExampleProcessingHandler extends Handler.Abstract implements Request.Processor
    {
        @Override
        public Request.Processor handle(Request request) throws Exception
        {
            // We must wrap/capture the request to use it in process
            return (response, callback) ->
            {
                // process the request, produce the response and complete the callback.
                callback.succeeded();
            };
        }
    }

    public class ExampleWrappingHandler extends Handler.Wrapper
    {
        @Override
        public Request.Processor handle(Request request) throws Exception
        {
            Request.Wrapper wrapper = new Request.Wrapper(request);
            Request.Processor processor = super.handle(request);
            if (processor == null)
                return null;
            // need to wrap/capture processor, just to remember the wrapper
            return (response, callback) ->
            {
                processor.process(wrapper, response, callback);
            };
        }
    }
```

This now avoid the ignored requests entirely but:
 + Now `ExampleProcessingHandler` must always allocate a processor just to capture the request. This is now an expensive way to avoid a `boolean` acceptance.
 + The `ExampleWrappingHandler` is more or less as before. 
 + A `Request.Processor` can now no longer be used for error handling, as no request is provided. Either a new interface with the request is needed for error handling, or we need to go back to an `ErrorHandler` and call both `handle` and `process`
 + We have found elsewhere with methods that only take the response, that it is better to also pass the request as it is often needed.  It will be strange that we pass request where it might sometimes be needed, but don't pass it where it is always needed, simply because there is another source of the request.

**Alternative 2**
Go back to a single method on `Handler` with a mechanism to explicitly accept the request.  We remove the whole concept of `Request.Processor` and instead only have `Handler`

```java
    public class ExampleProcessingHandler extends Handler.Abstract
    {
        @Override
        public void handle(Request request, Response response, Callback callback) throws Exception
        {
            // accept the request
            request.accept();
            // process the request, produce the response and complete the callback.
            callback.succeeded();
        }
    }

    public class ExampleWrappingHandler extends Handler.Wrapper
    {
        @Override
        public void handle(Request request, Response response, Callback callback) throws Exception
        {
            Request.Wrapper wrapper = new Request.Wrapper(request);
            super.handle(wrapper, response, callback);
        }
    }
```

Note that:
 + Aspects of the request API will be inactive until accepted (e.g. reading and demanding).
 + The response and callback can also be made inactive until the request is accepted.
 + acceptance cannot be indicated by a boolean return, else there would be a race between any async processing started and the returned true value being used to activate the request.
 + The only allocation is for the request wrapper because we want to wrap it. There are no allocations just to hold a value not provided by the API.
 + This is similar to our jetty<12 handlers, with `accept()` replacing `setHandled(true)`
 + The handler tree is navigated only once per request.
 + There are no Processors to be reused. Handlers by their very nature are reusable.
 + There is no question of where to put the handling (`handle` vs `process`) as there is only one.

**Alternative N**
You tell me!

**Conclusion**
We have introduced a whole new entity that needs to be instantiated/wrapped/defined/documented simply in order to avoid a call like `request.accept()` and to avoid exposing `response` and `callback` objects before they are active.  I think that made more sense when we tried to have different request APIs for the two calls, but it no longer makes sense now they are the same (to allow only one wrapping).   I think we need to think about this a bit more.

gregw:
Refactor Handler to implement Request.Processor for #8929.

Monster PR sorry.


gregw:
Currently jetty-server is expected to work.

gregw:
This monster PR is ready for review (the 2 build failures are CI flakes I believe).
The issue  #8929 has some detailed discussion as to why this PR exists, but in short the key reasons for it are:
 + to avoid the double dispatch into a context requiring classloader, threadlocal and listeners evoked on handle and then again on process
 + to avoid the need to wrap the Processor just to remember any request wrapper that has been applied.
 + to be more similar to our legacy handlers.

I think the best way to review this is for us all to look at just a few handlers to see how the style works:

- Anything that is a `Handler.Processor` has been trivially changed to extend `doProcess` instead of `process`
- The `ContextHandler` class is key, so perhaps everybody should have a look at that
- @lachlan-roberts Can you look at ee10 ServletHandler, ee9 nested and websocket upgrade handlers
- @janbartel Can you look at all the session handlers, also double check servlet stuff would be good.
- @lorban I think the Stats handler needs some review, plus eitherway I think we need to discuss timings at end of processing or on callback completion.   Also can you benchmark (maybe build without tests).
- @joakime GzipHandler has been substantially rewritten, so it needs both careful review and thought about how your jetty-10 changes will merge forward
- @sbordet I know you'll look at everything, but proxy stuff would be good if you could specifically cover it.

Whilst reviewing, think about 2 different things: firstly review for correctness (have I messed up anything), but also look for handlers that you don't like how they are written in this style or that you think might be inefficient.

I'm happy to hangout to explain any aspects of this.

gregw:
Note this also contains #8998 changes. They should go away once I get that built/reviewed/merged.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/SessionHandler.java
```diff
@@ -714,43 +715,37 @@ public void setSessionTrackingModes(Set<SessionTrackingMode> sessionTrackingMode
     }
 
     @Override
-    public Request.Processor handle(Request request) throws Exception
+    public void process(Request request, Response response, Callback callback) throws Exception
     {
+        Handler next = getHandler();
```
janbartel, line 720: Strictly speaking shouldn't the SessionHandler retrieve the session even if the next handler is null?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/security/ConstraintSecurityHandler.java
```diff
@@ -649,6 +650,7 @@ protected boolean checkUserDataPermissions(String pathInContext, Request request
             if (request.isSecure())
                 return true;
 
+            request.accept();
```
janbartel, line 653: Shouldn't the ConstraintSecurityHandler accept the request in any case at the top of the method? If it was already accepted, another request.accept() shouldn't matter.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/security/authentication/BasicAuthenticator.java
```diff
@@ -92,6 +92,9 @@ public Authentication validateRequest(Request req, Response res, Callback callba
         if (charset != null)
             value += ", charset=\"" + charset.name() + "\"";
         res.getHeaders().put(HttpHeader.WWW_AUTHENTICATE.asString(), value);
+        // We may already be accepted if called from HttpServletRequest.authenticate
+        if (!req.isAccepted())
+            req.accept();
```
janbartel, line 97: Request.accept() should do the test isAccepted() rather than the caller?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/security/authentication/FormAuthenticator.java
```diff
@@ -235,6 +235,7 @@ public Authentication validateRequest(Request req, Response res, Callback callba
         // Handle a request for authentication.
         if (jSecurityCheck)
         {
+            req.accept();
```
janbartel, line 238: Everywhere else you've tested if the request is accepted first before accepting it. So either need to do the same here, or if request.accept ignores subsequent accepts then you can remove all of the tests prior to the call.

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/ContextHandler.java
```diff
@@ -274,7 +274,7 @@ public APIContext getServletContext()
     @ManagedAttribute("Checks if the /context is not redirected to /context/")
     public boolean getAllowNullPathInfo()
     {
-        return _allowNullPathInfo;
+        return _coreContextHandler.getAllowNullPathInContext();
```
janbartel: Seems an unrelated change to the thrust of this PR, but as it's a good one, I'm gonna let it go through to the keeper ;)

jetty-ee9/jetty-ee9-webapp/src/test/java/org/eclipse/jetty/ee9/webapp/WebAppContextTest.java
```diff
@@ -490,7 +490,7 @@ public void testBaseResourceAbsolutePath() throws Exception
         // On Unix / Linux this should have no issue.
         // On Windows with fully qualified paths such as "E:\mybase\webapps\dump.war" the
         // resolution of the Resource can trigger various URI issues with the "E:" portion of the provided String.
-        context.setResourceBase(warPath.toString());
+        context.setBaseResourceAsString(warPath.toString());
```
janbartel: I thought all these were in a separate PR.

janbartel (approved):
I love the simplicity of the new Handler process method.

A few questions and quibbles but in general it looks good.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/SessionHandler.java
gregw: I don't think so. If there is no next handler, then this branch of the handler tree is not handling the request and thus it is not visiting the session and should not bring it into memory or touch the last access date etc.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/security/authentication/BasicAuthenticator.java
gregw: Normally it should be very clear if you need to accept or not, thus I don't want to make accept idempotent else we will get lazy and just call it all the time.
But `Authenticator`s are a bit special, as they can be called by mandatory constraints, and thus they have to accept the request... or they can be called by deferred and/or programmatic authentication, in which case the handler/server will have already accepted.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/security/authentication/FormAuthenticator.java
gregw: Because you can't do deferred FORM authentication, there is never a case where the request is already accepted

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/ContextHandler.java
gregw: It is, and is already in #8998 which is waiting for independent review

jetty-ee9/jetty-ee9-webapp/src/test/java/org/eclipse/jetty/ee9/webapp/WebAppContextTest.java
gregw: There was another PR to add the name, but the deprecated method was left in ee9 and this test had not been updated.  So an unrelated change... sorry just fixed it on the way.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/security/ConstraintSecurityHandler.java
gregw: We don't want duplicate accept calls, as there is a cost grabbing the lock. So duplicate accepts cause ISEs to be thrown.

Thus we accept here if and only if we know we are handling.

lorban:
@gregw Since this PR introduces a breaking API change that impacts the perf test suite's code, perf testing this change and comparing it to the current 12.0.0-SNAPSHOT isn't easy:

I can adapt the perf test suite to this PR's API then run the test, but I need to revert the changes afterwards to be able to run against the 12.0.x branch again.

lorban:
@gregw is this the right way to record latencies with this API?

```java
public class LatencyRecordingHandler extends Handler.Wrapper
{
    private final LatencyRecorder recorder;

    public LatencyRecordingHandler(LatencyRecorder recorder)
    {
        this.recorder = recorder;
    }

    @Override
    public void process(Request request, Response response, Callback callback) throws Exception
    {
        long before = System.nanoTime();
        request.addHttpStreamWrapper(httpStream -> new HttpStream.Wrapper(httpStream)
        {
            @Override
            public void succeeded()
            {
                super.succeeded();
                recorder.recordValue(System.nanoTime() - before);
            }

            @Override
            {
                super.failed(x);
                recorder.recordValue(System.nanoTime() - before);
            }
        });
        super.process(request, response, callback);
    }
}
```

gregw:
@lorban That code does record latency, but ultimately it would probably be best to use the new `getNanoTimeStamp` method on the request to get the earliest the request was seen as the beginning (same value/method can be accessed via the HttpStream).   But if the current code is measuring the beginning as the first time the handler sees the request, then keep it the same (or change it in both).

Just note that the call to `HttpStream.succeeded` is serialized with any other calls to the handlers. So the Callback might be succeeded, but the stream will not be succeeded until a call to process or onDataAvailable returns.  Again, keep it the same for load testing.

Ultimately for the stats handler, we should use the earlier beginning, and then consider if we want to measure the different end latencies or not.

joakime:
@gregw compilation failure ...

```
[ERROR] /home/jenkins/agent/workspace/jetty.project_PR-8978/jetty-core/jetty-http2/jetty-http2-tests/src/test/java/org/eclipse/jetty/http2/tests/ContentLengthTest.java:[131,9] 
  <anonymous org.eclipse.jetty.http2.tests.ContentLengthTest$6> is not abstract and does not override abstract method 
  doProcess(org.eclipse.jetty.server.Request,org.eclipse.jetty.server.Response,org.eclipse.jetty.util.Callback) in org.eclipse.jetty.server.Handler.Processor

[ERROR] /home/jenkins/agent/workspace/jetty.project_PR-8978/jetty-core/jetty-http2/jetty-http2-tests/src/test/java/org/eclipse/jetty/http2/tests/ContentLengthTest.java:[133,25] 
  process(org.eclipse.jetty.server.Request,org.eclipse.jetty.server.Response,org.eclipse.jetty.util.Callback) in <anonymous org.eclipse.jetty.http2.tests.ContentLengthTest$6> 
  cannot override process(org.eclipse.jetty.server.Request,org.eclipse.jetty.server.Response,org.eclipse.jetty.util.Callback) in org.eclipse.jetty.server.Handler.Processor

[ERROR]   overridden method is final
```

gregw:
@joakime yeah, some bad merges.  On it.

sbordet:
@gregw I looked at the `DelayedHandler` and I don't like what I see.

First of all, seems that now it forcibly allocates a `DelayedRequest` (the current code also allocates, but it'll be trivial to remove the allocation), which calls `accept()` on the innermost `Request`.

By doing that, the `Response` and the `Callback` are now "enabled", so it is entirely possible to use them _before_ calling `DelayedRequest.accept()`.
For example, the `Handler` nested within the `DelayedHandler` can completely forget to call `DelayedRequest.accept()` and everything will work.

A `Handler` that wraps `Request.accept()` has now to also wrap the `Response` and the `Callback` (see for example `ThreadLimitHandler`).

Furthermore, `DelayedRequest.run()` is duplicating the error handling if the `DelayedRequest` is not accepted.
This is necessary because the `Handler` that calls the `DelayedHandler` only has a reference to the innermost `Request`, which is accepted, so from its point of view the request has been accepted, but it has really not been accepted.
This is going to be super-confusing for users that implement their own `Handler`s.

The model of having to call `accept()/setHandled()` works fine (say in Jetty 11) when we do not wrap the `Request` and `Request` is mutable.
In a model where we wrap `Request` it does not work so well because it is just confusing since you can have an "almost-accepted" request, some request wrapper is accepted and some is not, and most of all the acceptance does not propagate outward, so the outermost `Handler` does not really know if the request is legitimately accepted or not.

I'll look at other `Handler`s and comment more.

sbordet:
```
[git:jetty-12-handler-as-processor]$ git diff --shortstat  jetty-12.0.x
 235 files changed, 2453 insertions(+), 2326 deletions(-)
```

Slightly more code, which I found surprising -- I would have expected much less code in this PR (due to the coalescing of `handle()` and `process()` into just `process()`).
Not really important though, I just wanted to report it.

sbordet:
One thing that surfaces from `GzipHandler` is that in this PR there may be the need for "cleanup" code that was not necessary before.

```java
// current
GzipHandler.handle(Request) {
  // A million conditions to check, then finally:
  gzipRequest = new GzipRequest(...);
  return gzipRequest.wrapProcessor(super.handle(gzipRequest));
}

// proposed in this PR
GzipHandler.process(Request) {
  // A million conditions to check, then:
  gzipRequest = new GzipRequest(...);
  nextHandler.process(gzipRequest, gzipResponse, gzipCallback);
  
  // Must cleanup if not accepted!
  if (!request.isAccepted() && request instanceof GzipRequest gzipRequest)
    gzipRequest.destroy();
}
```

Not terrible, but again it is something to remember that the current API solves more cleanly (no need to cleanup at all).

gregw:
@sbordet see below

> @gregw I looked at the `DelayedHandler` and I don't like what I see.

It is definitely true that the handle/process model is good for delay, as the process dispatch does give a hook to insert async behavior between accepting and handling.   This PR also does have a hook at the same point (the `accept` method), but as it is a blocking API you cannot insert delays there and have to do them later in processing.  However, I'm not sure this is a killer handler for the architecture, see bottom of comment. 


> First of all, seems that now it forcibly allocates a `DelayedRequest` (the current code also allocates, but it'll be trivial to remove the allocation), which calls `accept()` on the innermost `Request`.

Not trivial to delay process calls without allocating. Something has to remember the processor, request, response and callback so process can be invoked in a runnable or callback (or other async mechanism).


> By doing that, the `Response` and the `Callback` are now "enabled", so it is entirely possible to use them _before_ calling `DelayedRequest.accept()`. For example, the `Handler` nested within the `DelayedHandler` can completely forget to call `DelayedRequest.accept()` and everything will work.

We do check that the wrapped accept is called, so the delayed handler cannot forget to call accept.    We could also wrap the response/callback to enforce their inactivity  if we wanted, at the expense of an extra allocations (I do think we could consider some method renaming so that a single class could implement/wrap request/response/callback, even if not used as that tuple).

> 
> A `Handler` that wraps `Request.accept()` has now to also wrap the `Response` and the `Callback` (see for example `ThreadLimitHandler`).

No a handler that wants to act on write callback needs to wrap response and one that wants to act on completion needs to wrap the Callback.     Note this is fixing a bug in the current ThreadLimitHandler as it did not limit calls to demand/write callbacks.   Hence the extra wrapping and code.

 
> Furthermore, `DelayedRequest.run()` is duplicating the error handling if the `DelayedRequest` is not accepted. This is necessary because the `Handler` that calls the `DelayedHandler` only has a reference to the innermost `Request`, which is accepted, so from its point of view the request has been accepted, but it has really not been accepted. This is going to be super-confusing for users that implement their own `Handler`s.

We've already got such duplicated error handling the current DelayedHandler. You need it because exceptions thrown from a nested process call may not propagate back to the server as they may be called from a different thread/mechanism. 

> The model of having to call `accept()/setHandled()` works fine (say in Jetty 11) when we do not wrap the `Request` and `Request` is mutable. In a model where we wrap `Request` it does not work so well because it is just confusing since you can have an "almost-accepted" request, some request wrapper is accepted and some is not, and most of all the acceptance does not propagate outward, so the outermost `Handler` does not really know if the request is legitimately accepted or not.

I don't think that is a function of mutability.  That is a function of activity.   In Jetty 11 the request & response are always active and unrelated to `setHandled(true)`.  In Jetty-12 we are trying to solve the "problem" of having them active too early.   I don't actually think it has ever been a big problem in jetty<=11 that the request/response are always active.     I'm not opposed to putting in some extra checks around accept to  enforce no early usage, but I just don't see it as a big enough problem to justify the complexity and expense of a double dispatch.

Please ignore the incumbency of the current solution in the 12.0.x thread.  If we had a jetty 11 style architecture now, would we introduce a double dispatch model just to solve the "problem" of request/response being always active and to generalize the delay mechanisms?      I don't think so.

gregw:
> ```
> [git:jetty-12-handler-as-processor]$ git diff --shortstat  jetty-12.0.x
>  235 files changed, 2453 insertions(+), 2326 deletions(-)
> ```
> 
> Slightly more code, which I found surprising -- I would have expected much less code in this PR (due to the coalescing of `handle()` and `process()` into just `process()`). Not really important though, I just wanted to report it.

A few handlers have better implementations, fixing omissions and bugs in the current branch (eg ThreadLimitHandler didn't handle callback).  I've also added more javadoc as many methods were without.

gregw:
> One thing that surfaces from `GzipHandler` is that in this PR there may be the need for "cleanup" code that was not necessary before.

It is true that a request wrapper that allocates resources that needs to be released has to check for that release on: not accepted or callback completion.
But this is not that different from now where a wrapped request may still need to deal with a null processor.   It's not so in this case because the allocation is only done on read, but in general there is still an end-of-life-cycle event from null processors that wrappers need to consider.

gregw:
No longer a serious contender

PR #8994, opened 2022-12-01
Make the `Content.Chunk.isTerminal()` contract stricter

### LINKED ISSUE #8993: `Content.Chunk.isTerminal()` cannot discriminate `EOF` from chunks containing a pooled empty buffer
**Jetty version(s)**
12.0.x

**Description**
`Content.Chunk.isTerminal()` has a contract that allows some ambiguity: it is not possible to distinguish a `Content.Chunk` that contains a pool-acquired zero-byte buffer and is last from `Content.Chunk.EOF`.

Since `Content.Chunk.EOF.retain()` throws `UnsupportedOperationException`, each time a new chunk is read and before it is wrapped, it must be checked to make sure it isn't `EOF` as `retain()` cannot be called on the latter.

We currently use `isTerminal()` to perform that check but since it cannot discriminate `EOF` from an empty-and-last chunk we either leak the chunk's buffer or double-release it.

The `Content.Chunk` creation, `isTerminal()` and `Retainable` contracts should be reworked to lift this ambiguity.

lorban:
Make the `Content.Chunk.isTerminal()` contract stricter by explicitly disallowing the creation of new last chunks with an empty `ByteBuffer`.

Also replace non-last empty chunks with EMPTY where it makes  sense.

Closes #8993

### DIFF
--- a/jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/MultiPartTest.java
+++ b/jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/MultiPartTest.java
@@ -58,7 +58,7 @@ public class MultiPartTest
 
         assertEquals(0, listener.events.size());
 
-        parser.parse(Content.Chunk.from(BufferUtil.EMPTY_BUFFER, true));
+        parser.parse(Content.Chunk.EOF);
 
         assertEquals(1, listener.events.size());
         assertEquals("complete", listener.events.poll());
--- a/jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
+++ b/jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
@@ -444,7 +444,9 @@ public class Content
          */
         static Chunk from(ByteBuffer byteBuffer, boolean last)
         {
-            return new ByteBufferChunk.WithReferenceCount(byteBuffer, last);
+            if (byteBuffer.hasRemaining())
+               return new ByteBufferChunk.WithReferenceCount(byteBuffer, last);
+            return last ? EOF : EMPTY;
         }
 
         /**
@@ -453,11 +455,15 @@ public class Content
          *
          * @param byteBuffer the ByteBuffer with the bytes of this Chunk
          * @param last whether the Chunk is the last one
+         * @param releaser the code to run when this Chunk is released
          * @return a new Chunk
          */
         static Chunk from(ByteBuffer byteBuffer, boolean last, Runnable releaser)
         {
-            return new ByteBufferChunk.ReleasedByRunnable(byteBuffer, last, Objects.requireNonNull(releaser));
+            if (byteBuffer.hasRemaining())
+                return new ByteBufferChunk.ReleasedByRunnable(byteBuffer, last, Objects.requireNonNull(releaser));
+            releaser.run();
+            return last ? EOF : EMPTY;
         }
 
         /**
@@ -471,7 +477,10 @@ public class Content
          */
         static Chunk from(ByteBuffer byteBuffer, boolean last, Consumer<ByteBuffer> releaser)
         {
-            return new ByteBufferChunk.ReleasedByConsumer(byteBuffer, last, Objects.requireNonNull(releaser));
+            if (byteBuffer.hasRemaining())
+                return new ByteBufferChunk.ReleasedByConsumer(byteBuffer, last, Objects.requireNonNull(releaser));
+            releaser.accept(byteBuffer);
+            return last ? EOF : EMPTY;
         }
 
         /**
@@ -486,7 +495,10 @@ public class Content
          */
         static Chunk from(ByteBuffer byteBuffer, boolean last, Retainable retainable)
         {
-            return new ByteBufferChunk.WithRetainable(byteBuffer, last, Objects.requireNonNull(retainable));
+            if (byteBuffer.hasRemaining())
+                return new ByteBufferChunk.WithRetainable(byteBuffer, last, Objects.requireNonNull(retainable));
+            retainable.release();
+            return last ? EOF : EMPTY;
         }
 
         /**
@@ -540,28 +552,6 @@ public class Content
         }
 
         /**
-         * <p>Creates a chunk that is a slice of the given chunk.</p>
-         * <p>
-         *  The chunk slice has a byte buffer that is a slice of the original chunk's byte buffer, the last flag
-         *  copied over and the original chunk used as the retainable of the chunk slice.
-         *  Note that after this method returns, an extra Chunk instance refers to the same byte buffer,
-         *  so the original chunk is retained and is used as the {@link Retainable} for the returned chunk.
-         * </p>
-         * <p>Passing a null chunk returns null.</p>
-         * <p>Passing a terminal chunk returns it.</p>
-         *
-         * @param chunk the chunk to slice.
-         * @return the chunk slice.
-         */
-        static Chunk slice(Chunk chunk)
-        {
-            if (chunk == null || chunk.isTerminal())
-                return chunk;
-            chunk.retain();
-            return Chunk.from(chunk.getByteBuffer().slice(), chunk.isLast(), chunk);
-        }
-
-        /**
          * @return the ByteBuffer of this Chunk
          */
         ByteBuffer getByteBuffer();
@@ -572,8 +562,37 @@ public class Content
         boolean isLast();
 
         /**
+         * <p>Returns a new {@code Chunk} whose {@code ByteBuffer} is a slice of the
+         * {@code ByteBuffer} of the source {@code Chunk} unless the source
+         * {@link #hasRemaining() has no remaining byte} in which case:</p>
+         * <ul>
+         * <li>{@code this} is returned if it is an instance of {@link Error}</li>
+         * <li>{@link #EOF} is returned if {@link #isLast()} is {@code true}</li>
+         * <li>{@link #EMPTY} is returned if {@link #isLast()} is {@code false}</li>
+         * </ul>
+         * <p>If the source has remaining bytes, the returned {@code Chunk} retains
+         * the source {@code Chunk} and it is linked to it via
+         * {@link #from(ByteBuffer, boolean, Retainable)}.</p>
+         *
+         * @return a new {@code Chunk} retained from the source {@code Chunk} with a slice
+         * of the source {@code Chunk}'s {@code ByteBuffer}
+         */
+        default Chunk slice()
+        {
+            if (isTerminal())
+                return this;
+            if (!hasRemaining())
+                return EMPTY;
+            retain();
+            return from(getByteBuffer().slice(), isLast(), this);
+        }
+
+        /**
          * <p>Returns a new {@code Chunk} whose {@code ByteBuffer} is a slice, with the given
-         * position and limit, of the {@code ByteBuffer} of the source {@code Chunk}.</p>
+         * position and limit, of the {@code ByteBuffer} of the source {@code Chunk} unless the
+         * source is {@link #isTerminal() terminal} in which case {@code this} is returned, or
+         * if {@code position == limit} in which case {@link #EOF} or {@link #EMPTY} is
+         * returned depending on the value of {@code last}.</p>
          * <p>The returned {@code Chunk} retains the source {@code Chunk} and it is linked
          * to it via {@link #from(ByteBuffer, boolean, Retainable)}.</p>
          *
@@ -587,6 +606,8 @@ public class Content
         {
             if (isTerminal())
                 return this;
+            if (position == limit)
+                return last ? EOF : EMPTY;
             ByteBuffer sourceBuffer = getByteBuffer();
             int sourceLimit = sourceBuffer.limit();
             sourceBuffer.limit(limit);
@@ -649,20 +670,27 @@ public class Content
 
         /**
          * <p>Returns whether this Chunk is a <em>terminal</em> chunk.</p>
-         * <p>A terminal chunk is either an {@link Error error chunk},
-         * or a Chunk that {@link #isLast()} is true and has no remaining
-         * bytes.</p>
+         * <p>A terminal chunk is a Chunk that {@link #isLast()} is true
+         * and has no remaining bytes.</p>
+         * <p><em>Terminal</em> chunks cannot be lifecycled using the
+         * {@link Retainable} contract like other chunks: they always throw
+         * {@link UnsupportedOperationException} on {@link #retain()} and
+         * always return {@code true} on {@link #release()}. As such, they
+         * cannot contain a recyclable buffer and calling their
+         * {@link #release()} method isn't necessary, although harmless.
+         * </p>
          *
          * @return whether this Chunk is a terminal chunk
          */
         default boolean isTerminal()
         {
-            return this instanceof Error || isLast() && !hasRemaining();
+            return isLast() && !hasRemaining();
         }
 
         /**
          * <p>A chunk that wraps a failure.</p>
-         * <p>Error Chunks are always last and have no bytes to read.</p>
+         * <p>Error Chunks are always last and have no bytes to read,
+         * as such they are <em>terminal</em> Chunks.</p>
          *
          * @see #from(Throwable)
          */
--- a/jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
+++ b/jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
@@ -47,13 +47,61 @@ public class AsyncContent implements Content.Sink, Content.Source, Closeable
     private Runnable demandCallback;
     private long length = UNDETERMINED_LENGTH;
 
+    /**
+     * {@inheritDoc}
+     * <p>The write completes when the {@link Content.Chunk} returned by {@link #read()}
+     * that wraps {@code byteBuffer} is released.</p>
+     */
     @Override
     public void write(boolean last, ByteBuffer byteBuffer, Callback callback)
     {
-        write(Content.Chunk.from(byteBuffer, last, callback::succeeded), callback);
+        Content.Chunk chunk;
+        if (byteBuffer.hasRemaining())
+            chunk = Content.Chunk.from(byteBuffer, last, callback::succeeded);
+        else
+            chunk = last ? Content.Chunk.EOF : Content.Chunk.EMPTY;
+        offer(chunk, callback);
     }
 
+    /**
+     * <p>Writes the given {@link Content.Chunk}, notifying the {@link Callback} when the
+     * write is complete.</p>
+     * <p>The callback completes:</p>
+     * <ul>
+     * <li>immediately with a failure when the written chunk is an instance of {@link Content.Chunk.Error}</li>
+     * <li>successfully when the {@link Content.Chunk} returned by {@link #read()} is released</li>
+     * <li>successfully just before the {@link Content.Chunk} is returned if the latter {@link Content.Chunk#hasRemaining() has no remaining byte}</li>
+     * </ul>
+     *
+     * @param chunk the Content.Chunk to write
+     * @param callback the callback to notify when the write operation is complete
+     */
     public void write(Content.Chunk chunk, Callback callback)
+    {
+        Content.Chunk c;
+        if (chunk.isTerminal())
+        {
+            c = chunk;
+        }
+        else if (!chunk.hasRemaining())
+        {
+            c = Content.Chunk.EMPTY;
+        }
+        else
+        {
+            c = Content.Chunk.from(chunk.getByteBuffer(), chunk.isLast(), () ->
+            {
+                chunk.release();
+                callback.succeeded();
+            });
+        }
+        offer(c, callback);
+    }
+
+    /**
+     * The callback is only ever going to be succeeded if the chunk is terminal.
+     */
+    private void offer(Content.Chunk chunk, Callback callback)
     {
         Throwable failure = null;
         boolean wasEmpty = false;
@@ -109,7 +157,7 @@ public class AsyncContent implements Content.Sink, Content.Source, Closeable
                     if (writeClosed && chunks.size() == 1)
                     {
                         Content.Chunk chunk = chunks.peek().chunk();
-                        if (chunk.isLast() && !chunk.hasRemaining())
+                        if (chunk.isTerminal())
                             return;
                     }
                     l.await();
@@ -166,6 +214,8 @@ public class AsyncContent implements Content.Sink, Content.Source, Closeable
             if (chunks.isEmpty())
                 l.signal();
         }
+        if (!current.chunk().hasRemaining())
+            current.callback().succeeded();
         return current.chunk();
     }
 
--- a/jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ChunksContentSource.java
+++ b/jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ChunksContentSource.java
@@ -73,7 +73,7 @@ public class ChunksContentSource implements Content.Source
             if (last)
                 terminated = Content.Chunk.EOF;
         }
-        return Content.Chunk.from(chunk.getByteBuffer().slice(), chunk.isLast(), chunk);
+        return chunk;
     }
 
     @Override
--- a/jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/AsyncContentSourceTest.java
+++ b/jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/AsyncContentSourceTest.java
@@ -13,9 +13,14 @@
 
 package org.eclipse.jetty.io;
 
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.List;
 import java.util.concurrent.CancellationException;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.eclipse.jetty.io.content.AsyncContent;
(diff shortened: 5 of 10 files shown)


jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
```diff
@@ -50,7 +50,9 @@ public class AsyncContent implements Content.Sink, Content.Source, Closeable
     @Override
     public void write(boolean last, ByteBuffer byteBuffer, Callback callback)
     {
-        write(Content.Chunk.from(byteBuffer, last, callback::succeeded), callback);
+        boolean terminal = !byteBuffer.hasRemaining() && last;
+        Content.Chunk chunk = terminal ? Content.Chunk.EOF : Content.Chunk.from(byteBuffer, last, callback::succeeded);
```
gregw: Doesn't this imply that callback succeeded will get called twice? Once for the chunk release and again when the write succeeds?

gregw: Hmmm it always did... Still a good question?

gregw (approved):
I like the change. But it raises a question about calling succeeded twice?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
lorban: The short answer to your question is no, but you spotted the opposite bug nevertheless: when the chunk is replaced by EOF, the callback is never succeeded.

The reason why is because it is the call to `Chunk.release()` that is meant to succeed the callback, so nowhere in the `AsyncContent` code do we succeed it. Now that I've replaced empty callbacks with EOF, there's no way to detect when EOF.release() is called.

Let me fix that.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ChunksContentSource.java
```diff
@@ -73,7 +73,7 @@ public Content.Chunk read()
             if (last)
                 terminated = Content.Chunk.EOF;
         }
-        return Content.Chunk.from(chunk.getByteBuffer().slice(), chunk.isLast(), chunk);
```
lorban, line 76: Either, this was missing a `retain()` call, or slicing the chunk is the wrong thing to do.

Do we consider the chunks given to this content source to be managed by the reader, or do we consider the reader is getting duplicates and the writer is still supposed to lifecycle the chunks?

The previous impl was a bit in the middle (see the `fail()` impl vs `read()`'s odd creation of a new chunk with a sliced buffer and a common releaser).

I've decided to go for "the reader has to lifecycle the chunks, the writer fire&forgets them" but that should probably be javdoc'ed.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ByteBufferContentSource.java
```diff
@@ -78,6 +78,8 @@ public Content.Chunk read()
             if (last)
                 terminated = Content.Chunk.EOF;
         }
+        if (!buffer.hasRemaining())
```
lorban: We should probably javadoc that the code instantiating this class is responsible for managing the lifecycle of the buffers, unlike `ChunksContentSource` for which the exact opposite is true.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
```diff
@@ -50,7 +50,31 @@ public class AsyncContent implements Content.Sink, Content.Source, Closeable
...
+                public void succeeded()
+                {
+                    // Noop, chunk.release() will trigger the callback's success.
+                }
+            };
+        }
+        write(chunk, cb);
     }
 
     public void write(Content.Chunk chunk, Callback callback)
```
lorban: We should probably javadoc when the callback is called as it could be confused with the chunk's lifecycle w.r.t when the chunk's byte buffer gets released.

lorban:
@gregw I answered your question w.r.t succeeding twice the callback in your code comment. In a nutshell, it's never succeeded twice and my original change introduced a bug which made it never be succeeded in certain cases. This is now fixed with some comments explaining the somewhat baroque logic.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
```diff
@@ -50,7 +50,31 @@ public class AsyncContent implements Content.Sink, Content.Source, Closeable
...
+            cb = new Callback.Nested(callback)
+            {
+                @Override
+                public void succeeded()
+                {
+                    // Noop, chunk.release() will trigger the callback's success.
+                }
+            };
+        }
+        write(chunk, cb);
```
sbordet: I would do the other way around.
Have an `private offer()` method called by the 2 `write()` methods, so that you can release the chunk when the read releases the chunk.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
```diff
@@ -50,7 +50,31 @@ public class AsyncContent implements Content.Sink, Content.Source, Closeable
     @Override
     public void write(boolean last, ByteBuffer byteBuffer, Callback callback)
     {
-        write(Content.Chunk.from(byteBuffer, last, callback::succeeded), callback);
+        // This class' read() impl calls succeeded() on the callback before returning it to the
+        // caller, so when EOF / EMPTY are used this is perfectly right as the chunk can be
+        // considered consumed and there is no buffer to reclaim. But when a proper buffer is
+        // used the passed callback must be wrapped to make its succeeded() method a noop
+        // as it is the release of the chunk that should trigger the success of the callback.
```
sbordet: I'd change this comment to a javadoc to explain the semantic of `write()` which is: "the callback is completed when the chunk is read *and* released".

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ChunksContentSource.java
sbordet: Agreed, I think just `return chunk` should be enough here.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -571,6 +555,23 @@ static Chunk slice(Chunk chunk)
          */
         boolean isLast();
 
+        /**
+         * <p>Returns a new {@code Chunk} whose {@code ByteBuffer} is a slice of the
+         * {@code ByteBuffer} of the source {@code Chunk}.</p>
+         * <p>The returned {@code Chunk} retains the source {@code Chunk} and it is linked
+         * to it via {@link #from(ByteBuffer, boolean, Retainable)}.</p>
```
sbordet: Update javadoc with "unless is EOF or EMPTY" or similar.

sbordet (changes requested):

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
lorban: Done.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
lorban: Done, but handling EOF/EMPTY adds some complexity which I accounted for in the javadoc.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
lorban: Done.

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/MultiPartTest.java
```diff
@@ -58,7 +58,7 @@ public void testParseNoPartsInTwoChunks()
 
         assertEquals(0, listener.events.size());
 
-        parser.parse(Content.Chunk.from(BufferUtil.EMPTY_BUFFER, true));
+        parser.parse(Content.Chunk.EOF);
```
gregw, line 61: My understanding of this PR is that you didn't need to change this, as `from` would have returned EOF anyway.    Just checking that is true?

jetty-core/jetty-http2/jetty-http2-client-transport/src/main/java/org/eclipse/jetty/http2/client/transport/internal/HttpReceiverOverHTTP2.java
```diff
@@ -80,10 +81,21 @@ public Content.Chunk read(boolean fillInterestIfNeeded)
...
+            {
+                responseSuccess(getHttpExchange(), null);
+                return Content.Chunk.EOF;
+            }
+            else
+            {
+                return Content.Chunk.EMPTY;
+            }
+        }
+        return Content.Chunk.from(buffer, false, data);
```
gregw: Doesn't `from` now do all this logic about choosing to return `EMPTY` or `EOF` if possible?  It was much simpler code to just call `from` and let it work out the details, rather than make the caller do this all the time.

jetty-core/jetty-http2/jetty-http2-server/src/main/java/org/eclipse/jetty/http2/server/internal/HttpStreamOverHTTP2.java
```diff
@@ -241,12 +241,19 @@ public Runnable onTrailer(HeadersFrame frame)
...
+        ByteBuffer byteBuffer = frame.getData();
+        if (!byteBuffer.hasRemaining())
+        {
+            // We must NOT retain because we are not passing the
+            // ByteBuffer to the Chunk.
+            if (frame.isEndStream())
+                return Content.Chunk.EOF;
+            else
+                return Content.Chunk.EMPTY;
+        }
```
gregw: again, can't we just let `from` do it's stuff rather than put this logic everywhere?

jetty-core/jetty-http3/jetty-http3-client-transport/src/main/java/org/eclipse/jetty/http3/client/transport/internal/HttpReceiverOverHTTP3.java
```diff
@@ -64,10 +64,20 @@ public Content.Chunk read(boolean fillInterestIfNeeded)
...
+            data.release();
+            if (data.isLast())
+            {
+                responseSuccess(getHttpExchange(), null);
+                return Content.Chunk.EOF;
+            }
+            else
+            {
+                return Content.Chunk.EMPTY;
+            }
```
gregw: ditto

jetty-core/jetty-http3/jetty-http3-server/src/main/java/org/eclipse/jetty/http3/server/internal/HttpStreamOverHTTP3.java
```diff
@@ -233,12 +233,19 @@ public Runnable onTrailer(HeadersFrame frame)
...
+        ByteBuffer byteBuffer = data.getByteBuffer();
+        if (!byteBuffer.hasRemaining())
+        {
+            // As we are NOT passing the ByteBuffer to the Chunk,
+            // we must not retain.
+            if (data.isLast())
+                return Content.Chunk.EOF;
+            else
+                return Content.Chunk.EMPTY;
+        }
```
gregw: ditto

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
```diff
@@ -47,13 +47,55 @@ public class AsyncContent implements Content.Sink, Content.Source, Closeable
...
+     */
     @Override
     public void write(boolean last, ByteBuffer byteBuffer, Callback callback)
     {
-        write(Content.Chunk.from(byteBuffer, last, callback::succeeded), callback);
+        Content.Chunk chunk;
+        if (!byteBuffer.hasRemaining())
+            chunk = last ? Content.Chunk.EOF : Content.Chunk.EMPTY;
+        else
+            chunk = Content.Chunk.from(byteBuffer, last, callback::succeeded);
```
gregw: ditto

gregw (changes requested):
I think we should rely on `Chunk.from` to correctly choose from `EOF`, `EMPTY` or a new chunk, rather than have that same logic scatter around the code base.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
lorban: Done.

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/MultiPartTest.java
lorban: If `Chunk.from` would choose from `EOF` or `EMPTY`, you'd be right except that `from(ByteBuffer, boolean)` doesn't know about the lifecycle of the given buffer so it has no way to release it when it decides to replace it with EOF or EMPTY.

We could change this call to `Content.Chunk.from(BufferUtil.EMPTY_BUFFER, true, () -> {})` but I'm not sure it would be any better.

jetty-core/jetty-http2/jetty-http2-client-transport/src/main/java/org/eclipse/jetty/http2/client/transport/internal/HttpReceiverOverHTTP2.java
lorban: That can be done with the 3 variants of `from` that take a `ByteBuffer`-controlling object as a parameter (when `from` wants to replace the empty byte buffer with EOF/EMPTY, it must release the byte buffer) but not with the 4th one.

I've changed the replacement code scattered everywhere so that this is now done by the 3 `from` variants, I made the 4th one throw an exception and changed the calls to the 4th one to one of the 3 non-throwing one where appropriate.

I find this much cleaner, so I'm in favor of this change.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
lorban: This one  is an exception: the callback must not be succeeded until the write is over, which happens when another thread releases the generated chunk after reading it.

So we cannot let `from` call the release callback if it decides to replace the chunk with EOF or EMPTY otherwise we would succeed the callback too early.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
```diff
@@ -624,7 +624,7 @@ public Content.Chunk read()
                     // TODO: use a ByteBuffer pool and direct ByteBuffers?
                     ByteBuffer byteBuffer = UTF_8.encode(builder.toString());
                     state = State.CONTENT;
-                    yield Content.Chunk.from(byteBuffer, false);
+                    yield Content.Chunk.from(byteBuffer, false, () -> {});
```
sbordet: Why this?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -457,6 +463,11 @@ static Chunk from(ByteBuffer byteBuffer, boolean last)
          */
         static Chunk from(ByteBuffer byteBuffer, boolean last, Runnable releaser)
         {
+            if (!byteBuffer.hasRemaining())
```
sbordet: I think it's more readable if you invert the `if`.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -471,6 +482,11 @@ static Chunk from(ByteBuffer byteBuffer, boolean last, Runnable releaser)
          */
         static Chunk from(ByteBuffer byteBuffer, boolean last, Consumer<ByteBuffer> releaser)
         {
+            if (!byteBuffer.hasRemaining())
```
sbordet: Invert the `if`

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -486,6 +502,11 @@ static Chunk from(ByteBuffer byteBuffer, boolean last, Consumer<ByteBuffer> rele
          */
         static Chunk from(ByteBuffer byteBuffer, boolean last, Retainable retainable)
         {
+            if (!byteBuffer.hasRemaining())
```
sbordet: Invert the `if`.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
```diff
@@ -47,13 +47,55 @@ public class AsyncContent implements Content.Sink, Content.Source, Closeable
...
+        offer(chunk, callback);
     }
 
+    /**
+     * <p>Writes the given {@link ByteBuffer}, notifying the {@link Callback} when the
```
sbordet:
```suggestion
     * <p>Writes the given chunk, notifying the {@link Callback} when the
```

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
```diff
@@ -47,13 +47,55 @@ public class AsyncContent implements Content.Sink, Content.Source, Closeable
...
+     * <li>successfully just before the {@link Content.Chunk} is returned if the latter {@link Content.Chunk#hasRemaining() has no remaining byte}</li>
+     * </ul>
+     *
+     * @param chunk the Content.Chunk to write
+     * @param callback the callback to notify when the write operation is complete
+     */
     public void write(Content.Chunk chunk, Callback callback)
+    {
+        Content.Chunk c;
+        if (chunk.isTerminal())
```
sbordet, line 82: Add braces to this chained `if/else`.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
```diff
@@ -47,13 +47,55 @@ public class AsyncContent implements Content.Sink, Content.Source, Closeable
...
+            chunk = Content.Chunk.from(byteBuffer, last, callback::succeeded);
+        offer(chunk, callback);
     }
 
+    /**
+     * <p>Writes the given {@link ByteBuffer}, notifying the {@link Callback} when the
+     * write is complete.</p>
+     * <p>The callback completes:</p>
+     * <ul>
+     * <li>immediately with a failure when the written chunk is an instance of {@link Content.Chunk.Error}</li>
```
sbordet, line 71: This feel weird.
I think it should be treated like a normal "special" write of, say, `EOF`, so succeeding the callback.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ByteBufferContentSource.java
```diff
@@ -78,7 +78,7 @@ public Content.Chunk read()
             if (last)
                 terminated = Content.Chunk.EOF;
         }
-        return Content.Chunk.from(buffer, last);
+        return Content.Chunk.from(buffer, last, () -> {});
```
sbordet: Unnecessary?

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/AsyncContentSourceTest.java
```diff
@@ -125,4 +129,63 @@ public void testWriteErrorChunk() throws Exception
...
+            async.write(true, ByteBuffer.wrap(new byte[0]), Callback.from(successCounter::incrementAndGet, failureRef::set));
+
+            Content.Chunk chunk = async.read();
+            assertThat(successCounter.get(), is(1));
+            assertThat(chunk.isTerminal(), is(true));
+            assertThat(chunk.release(), is(true));
+            assertThat(successCounter.get(), is(1));
+            assertThat(failureRef.get(), is(nullValue()));
+        }
+    }
```
sbordet, line 193: Add one more test for the `Error` behavior.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/MockHttpStream.java
```diff
@@ -33,7 +33,32 @@
 public class MockHttpStream implements HttpStream
 {
     private static final Throwable SUCCEEDED = new Throwable();
-    private static final Content.Chunk DEMAND = Content.Chunk.from(BufferUtil.EMPTY_BUFFER, false);
+    private static final Content.Chunk DEMAND = new Content.Chunk()
```
sbordet, line 36: Isn't this `EMPTY`? Why do you need it to be different from `EMPTY`?

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/MockHttpStream.java
```diff
@@ -65,7 +90,7 @@ public boolean isDemanding()
 
     public Runnable addContent(ByteBuffer buffer, boolean last)
     {
-        return addContent((last && BufferUtil.isEmpty(buffer)) ? Content.Chunk.EOF : Content.Chunk.from(buffer, last));
+        return addContent(Content.Chunk.from(buffer, last, () -> {}));
```
sbordet: Unnecessary?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletMultiPartFormData.java
```diff
@@ -107,7 +107,7 @@ private Parts parse(ServletContextRequest.ServletApiRequest request) throws IOEx
                 formData.parse(Content.Chunk.EOF);
                 break;
             }
-            formData.parse(Content.Chunk.from(ByteBuffer.wrap(buffer, 0, read), false));
+            formData.parse(Content.Chunk.from(ByteBuffer.wrap(buffer, 0, read), false, () -> {}));
```
sbordet: Unnecessary?

sbordet (changes requested):

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
lorban: Fixed.

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/AsyncContentSourceTest.java
lorban: There already is `testWriteErrorChunk` that checks error chunks are read and the callback failed.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/MockHttpStream.java
lorban: No, it's a different instance on purpose as it is assigned and compared with `==`. The buffer and last are meaningless as they are never read, and this chunk never leaves this implementation.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
lorban: Ok.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
lorban: Ok.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
lorban: Ok.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
lorban: It's because the `byteBuffer` may contain 0 byte, so it should be replaced with `EMPTY` but `Content.Chunk.from(ByteBuffer, boolean)` cannot do that as it doesn't know how to release the buffer.

Calling `Content.Chunk.from(ByteBuffer, boolean, Runnable)` can replace a 0 byte buffer with `EMPTY` as when it does so, it can call the `Runnable` to release to buffer. And since this buffer was allocated and not taken from a pool, the `Runnable` can be a no-op and leave the buffer to the GC.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ByteBufferContentSource.java
lorban: See above.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/MockHttpStream.java
lorban: See above.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletMultiPartFormData.java
lorban: See above.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
lorban: Done.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
lorban: I did not change this contract, it is now in `offer()` but I just documented the existing impl. Do you think it should be changed?

sbordet (approved):

PR #9002, opened 2022-12-05
Jetty 12 - Issue #8999 - Remove FileID.isArchive() from ResourceFactory

### LINKED ISSUE #8999: Overuse of FileID.isArchive() and inability to deal with packed jars without .jar extension
**jetty-12.0.x**

```
    default Resource newJarFileResource(URI uri)
    {
        if (!FileID.isArchive(uri))
            throw new IllegalArgumentException("Path is not a Java Archive: " + uri);
        if (!uri.getScheme().equalsIgnoreCase("file"))
            throw new IllegalArgumentException("Not an allowed path: " + uri);
        return newResource(URIUtil.toJarFileUri(uri));
    }
```

Line 240 of `ResourceFactory` we test if the uri has an appropriate file extension with `FileID.isArchive`.
Then at line 244 we call `URIUtil.toJarFileUri`, which _also_ tests if the uri has an appropriate extension:

```
    public static URI toJarFileUri(URI uri)
    {
        Objects.requireNonNull(uri, "URI");
        String scheme = Objects.requireNonNull(uri.getScheme(), "URI scheme");

        if (!FileID.isArchive(uri))
            return uri;

        boolean hasInternalReference = uri.getRawSchemeSpecificPart().indexOf("!/") > 0;

        if (scheme.equalsIgnoreCase("jar"))
        {
            if (uri.getRawSchemeSpecificPart().startsWith("file:"))
            {
                // Looking good as a jar:file: URI
                if (hasInternalReference)
                    return uri; // is all good, no changes needed.
                else
                    // add the internal reference indicator to the root of the archive
                    return URI.create(uri.toASCIIString() + "!/");
            }
        }
        else if (scheme.equalsIgnoreCase("file"))
        {
            String rawUri = uri.toASCIIString();
            if (hasInternalReference)
                return URI.create("jar:" + rawUri);
            else
                return URI.create("jar:" + rawUri + "!/");
        }

        // shouldn't be possible to reach this point
        throw new IllegalArgumentException("Cannot make %s into `jar:file:` URI".formatted(uri));
    }
```

We don't need to do the test multiple times.

Moreover, the `FileID.isArchive` method does not work with files that are actually packed jar format, but missing the `.jar` extension, for example as seen with `Equinox` in the osgi environment. I tried modifying the method to also try the `Files.probeContentType` method, but it failed to detect a filetype for these types of files. Do we really need to check the file extension? If you make a uri into a `jar:file` uri and its the wrong type of file, then using it will fail. Same thing for the `JarFileResource`.

joakime:
Fixes #8999

Remove `FileID.isArchive()` check from ResourceFactory.
Now the exception will occur from the JVM if the archive is unsupported.

### DIFF
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
@@ -1862,9 +1862,6 @@ public final class URIUtil
         Objects.requireNonNull(uri, "URI");
         String scheme = Objects.requireNonNull(uri.getScheme(), "URI scheme");
 
-        if (!FileID.isArchive(uri))
-            return uri;
-
         boolean hasInternalReference = uri.getRawSchemeSpecificPart().indexOf("!/") > 0;
 
         if (scheme.equalsIgnoreCase("jar"))
@@ -1882,6 +1879,8 @@ public final class URIUtil
         else if (scheme.equalsIgnoreCase("file"))
         {
             String rawUri = uri.toASCIIString();
+            if (rawUri.endsWith("/")) // skip directories
+                return uri;
             if (hasInternalReference)
                 return URI.create("jar:" + rawUri);
             else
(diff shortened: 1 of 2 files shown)


janbartel (changes requested):
Still not going to work, as the newJarFileResource calls URIUtil.toJarFileUri which also calls FileID.isArchive.

joakime:
@janbartel updated the `URIUtil.toJarFileUri` as well.

janbartel (approved):
Works for me

PR #9007, opened 2022-12-05
Improved locking for HttpReceiver.ContentSource.

sbordet:
Improved response failure code path.
Now either responseFailure() must be called, or exchange.responseComplete() followed by HttpReceiver.abort().

Fixed failAndClose() for HTTP/2 and HTTP/3: the connection must not be closed, stream.reset() is sufficient.

Fixed flaky test HttpClientDemandTest.testTwoListenersWithDifferentDemand().

Fixed DistributionTests.testVirtualThreadPool().

### DIFF
--- a/jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/HttpReceiver.java
+++ b/jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/HttpReceiver.java
@@ -31,6 +31,7 @@ import org.eclipse.jetty.io.Content;
 import org.eclipse.jetty.io.content.ContentSourceTransformer;
 import org.eclipse.jetty.util.BufferUtil;
 import org.eclipse.jetty.util.Promise;
+import org.eclipse.jetty.util.thread.AutoLock;
 import org.eclipse.jetty.util.thread.SerializedInvoker;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -266,6 +267,9 @@ public abstract class HttpReceiver
             List<Response.ResponseListener> responseListeners = exchange.getConversation().getResponseListeners();
             notifier.notifyHeaders(responseListeners, response);
 
+            if (exchange.isResponseComplete())
+                return;
+
             if (HttpStatus.isInterim(response.getStatus()))
             {
                 if (LOG.isDebugEnabled())
@@ -338,24 +342,18 @@ public abstract class HttpReceiver
         if (LOG.isDebugEnabled())
             LOG.debug("Invoking responseSuccess on {}", this);
 
-        NotifiableContentSource contentSource = this.contentSource;
-        if (contentSource != null)
-        {
-            this.contentSource = null;
-            contentSource.eof();
-        }
+        // Mark atomically the response as completed, with respect
+        // to concurrency between response success and response failure.
+        if (!exchange.responseComplete(null))
+            return;
 
         invoker.run(() ->
         {
             if (LOG.isDebugEnabled())
                 LOG.debug("Executing responseSuccess on {}", this);
 
-            // Mark atomically the response as completed, with respect
-            // to concurrency between response success and response failure.
-            if (!exchange.responseComplete(null))
-                return;
-
             responseState = ResponseState.IDLE;
+
             reset();
 
             HttpResponse response = exchange.getResponse();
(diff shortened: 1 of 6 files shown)


jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/HttpReceiver.java
```diff
@@ -385,34 +383,26 @@ protected void responseSuccess(HttpExchange exchange, Runnable afterSuccessTask)
     protected void responseFailure(Throwable failure, Promise<Boolean> promise)
     {
         if (LOG.isDebugEnabled())
-            LOG.debug("Invoking responseFailure with {} on {}", failure, this);
-
-        invoker.run(() ->
+            LOG.debug("Failing with {} on {}", failure, this);
+
+        HttpExchange exchange = getHttpExchange();
```
gregw, line 388: Given the comment/test below, it feels like this should be a `takeHttpExchange()` semantic, or at least the test below could be:
```java
if (exchange == null || !disassociate(exchange))
```

At the very least, can the comment below explain how the race of exchange becoming null just after the get is handled?

gregw (changes requested):
Either take semantic is needed or better comment to explain why not.

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/HttpReceiver.java
sbordet: @gregw I don't understand. 
The semantic cannot be "take", as the "take" happens later when also the request is completed.
The call `exchange.responseComplete()` is atomic, so the decision of whether doing something or not is taken there.

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/HttpReceiver.java
```diff
@@ -482,10 +472,12 @@ public void abort(HttpExchange exchange, Throwable failure, Promise<Boolean> pro
         if (LOG.isDebugEnabled())
             LOG.debug("Invoking abort with {} on {}", failure, this);
 
+        assert exchange.isResponseComplete();
```
lorban: Other assertions are done by throwing `IllegalStateException`. Align?

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/HttpReceiver.java
```diff
@@ -654,42 +649,36 @@ private class ContentSource implements NotifiableContentSource
...
-                try
-                {
-                    demandCallback.run();
-                }
-                catch (Throwable x)
+                if (currentChunk != null)
                 {
-                    fail(x);
+                    // There was a concurrent call to fail().
+                    current.release();
```
lorban: `current` can be null here, you're missing a null check.

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/HttpReceiver.java
```diff
@@ -731,12 +711,30 @@ private void processDemand()
...
+            {
+                current = currentChunk;
+            }
+
+            if (current == null)
             {
-                currentChunk = HttpReceiver.this.read(true);
-                if (currentChunk == null)
+                current = HttpReceiver.this.read(true);
+                if (current == null)
```
lorban, line 726: If there was a concurrent call to `fail()` here, you want to return `currentChunk` even if `current` is null.

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/HttpReceiver.java
```diff
@@ -765,20 +763,43 @@ private void invokeDemandCallback(boolean invoke)
...
+            if (failed)
+                HttpReceiver.this.failAndClose(failure);
+            invokeDemandCallback(true);
+        }
+
+        @Override
+        public boolean error(Throwable failure)
         {
             if (LOG.isDebugEnabled())
                 LOG.debug("Failing {}", this);
```
lorban: This log is confusing: the code is error'ing, not failing.

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/HttpReceiver.java
```diff
@@ -765,20 +763,43 @@ private void invokeDemandCallback(boolean invoke)
 
         @Override
         public void fail(Throwable failure)
+        {
+            boolean failed = error(failure);
+            if (failed)
+                HttpReceiver.this.failAndClose(failure);
+            invokeDemandCallback(true);
```
lorban, line 775: Shouldn't `invokeDemandCallback(true)` only be called if `failed` is true?

lorban (changes requested):

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/HttpReceiver.java
sbordet: This method returns `void` so if `current==null` we do not want to overwrite `currentChunk`, and there is nothing to release, so I think the code is correct.

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/HttpReceiver.java
sbordet: I would leave it in this way, as `invokeDemandCallback()` is "safe" in the sense that if there is no callback it's a no-op.

lorban (approved):

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/HttpReceiver.java
gregw: @sbordet It is a general pattern for a race to have code in an async environment that does non atomic test and then act.    It may well be that the exchange is atomically "taken" and nulled elsewhere, but a thread here may still have tested the exchange whilst non null and then act on it whilst it is null.

So either this is a race here, or there is some external circumstance that ensures that the `responseComplete` can never be called between the test and the action.    So we need to either fix the race or document why it is not a race.

PR #9010, opened 2022-12-06
Fixes #9006 - WebSocket MessageInputStream.read() returns signed byte

### LINKED ISSUE #9006: WebSocket Message InputStream read() returns signed byte
**Jetty version(s)**

`10.0.12`

```
openjdk version "11.0.10" 2021-01-19 LTS
OpenJDK Runtime Environment Zulu11.45+28-SA (build 11.0.10+9-LTS)
OpenJDK 64-Bit Server VM Zulu11.45+28-SA (build 11.0.10+9-LTS, mixed mode)
```

**OS type/version**

Tested on `macOS 12.6.1` (though this is likely a OS-agnostic issue)

**Description**

In Jetty 9, reading any unsigned byte from 0 to 255 inclusive using the `InputStream#read()` API against the `InputStream` instance injected by Jetty into the method annotated with `@OnWebSocketMessage` for class annotated with `@WebSocket` worked well.

However, in Jetty 10, attempting to read an unsigned byte in the range 128 to 255 inclusive using the same setup fails since values in this range are returned as signed bytes (-128 to -1, inclusive), which breaks the `InputStream#read()` [contract](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/InputStream.html#read()).

This is causing a problem for us when we use `DataOutputStream` / `DataInputStream` to write / read shorts (for example) using Jetty's WebSocket `OutputStream` / `InputStream`, since the signed byte returned by Jetty is interpreted as `EOF`.

**How to reproduce?**

See: https://github.com/jspechts/IGN-6520-JETTY-SSCCE

sbordet:
Now properly coverting to `int`.
Added test.

Also fixed MultiPartInputStreamParser.Base64InputStream for the same issue.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/MultiPartInputStreamParser.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/MultiPartInputStreamParser.java
@@ -941,7 +941,7 @@ public class MultiPartInputStreamParser
                 _pos = 0;
             }
 
-            return _buffer[_pos++];
+            return _buffer[_pos++] & 0xFF;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


joakime (approved):

gregw:
@sbordet don't forget to merge to 12

PR #9014, opened 2022-12-06
Issue #8973 - Rework KeyStoreScanner handling for symlink related changes

### LINKED ISSUE #8973: Does ssl-reload monitor keystore when it is a symlink?
**jetty-9.4.49.v20220914** and **11.0.12**

**Java 11**

**Does ssl-reload monitor keystore when it is a symlink?**

Have deployed my application which uses jetty in kubernetes pod. We are using cert-manger for certificate management.
Secret is mounted inside a pod as below:

```
drwxr-xr-x    2 root     root         140 Nov 29 09:38 ..2022_11_29_09_38_19.297914115
lrwxrwxrwx    1 root     root          31 Nov 29 09:38 ..data -> ..2022_11_29_09_38_19.297914115
lrwxrwxrwx    1 root     root          13 Nov 29 07:06 ca.crt -> ..data/ca.crt
lrwxrwxrwx    1 root     root          19 Nov 29 07:06 keystore.p12 -> ..data/keystore.p12
lrwxrwxrwx    1 root     root          14 Nov 29 07:06 tls.crt -> ..data/tls.crt
lrwxrwxrwx    1 root     root          14 Nov 29 07:06 tls.key -> ..data/tls.key
lrwxrwxrwx    1 root     root          21 Nov 29 07:06 truststore.p12 -> ..data/truststore.p12

```

When the certificates are renewed automatically by cert-manager , the target file for symlink keystore.p12 is changed , but even after that jetty is not picking up the updated certs. If keystore is not symlink , jetty is picking the updated certs.

joakime:
Fixes #8973

+ Removed changes from #8786 and #8787
+ More test cases

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/ssl/KeyStoreScanner.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/ssl/KeyStoreScanner.java
@@ -15,6 +15,7 @@ package org.eclipse.jetty.util.ssl;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Path;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
@@ -44,11 +45,6 @@ public class KeyStoreScanner extends ContainerLifeCycle implements Scanner.Discr
 
     public KeyStoreScanner(SslContextFactory sslContextFactory)
     {
-        this(sslContextFactory, true);
-    }
-
-    public KeyStoreScanner(SslContextFactory sslContextFactory, boolean followLinks)
-    {
         this.sslContextFactory = sslContextFactory;
         try
         {
@@ -58,12 +54,6 @@ public class KeyStoreScanner extends ContainerLifeCycle implements Scanner.Discr
                 throw new IllegalArgumentException("keystore file does not exist");
             if (monitoredFile.isDirectory())
                 throw new IllegalArgumentException("expected keystore file not directory");
-
-            if (followLinks && keystoreResource.isAlias())
-            {
-                // This resource has an alias, so monitor the target of the alias.
-                monitoredFile = new File(keystoreResource.getAlias());
-            }
 
             keystoreFile = monitoredFile;
             if (LOG.isDebugEnabled())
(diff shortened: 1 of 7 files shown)


lachlan-roberts:
@joakime you need to update the `jetty-ssl-context-reload.xml` and `ssl-reload.mod` files as well.

joakime:
@lachlan-roberts done, please re-review

lachlan-roberts:
@joakime I made a commit to remove some bad documentation about the followLinks setting from `module-ssl-reload.adoc`.

jetty-util/src/main/java/org/eclipse/jetty/util/ssl/KeyStoreScanner.java
```diff
@@ -88,11 +78,23 @@ public KeyStoreScanner(SslContextFactory sslContextFactory, boolean followLinks)
...
+            return keystoreFile.toPath();
+        }
+    }
+
     @Override
     public void fileAdded(String filename)
     {
         if (LOG.isDebugEnabled())
-            LOG.debug("added {}", filename);
+            LOG.debug("fileAdded {} - keystoreFile.toReal {}", filename, getRealKeyStorePath());
```
lachlan-roberts, line 97: I don't think we need to do a `toRealPath()` for debugging when we don't do it for the monitored file or for the scanner.

jetty-util/src/main/java/org/eclipse/jetty/util/ssl/KeyStoreScanner.java
```diff
@@ -88,11 +78,23 @@ public KeyStoreScanner(SslContextFactory sslContextFactory, boolean followLinks)
         addBean(_scanner);
     }
 
+    private Path getRealKeyStorePath()
+    {
+        try
+        {
+            return keystoreFile.toPath().toRealPath();
```
lachlan-roberts, line 85: You could assign `keystoreFile.toPath()` to a variable and prevent doing this twice in the case of a failure.
(not that this matters that much as its only for debugging)

lachlan-roberts (approved):
I can't think of an example where it would be necessary to resolve the symlinks in the `KeyStoreScanner`, so I think removing `followLinks` setting is ok.

jetty-util/src/main/java/org/eclipse/jetty/util/ssl/KeyStoreScanner.java
joakime: it helps, in debugging, to see that it picked up the right file.

PR #9020, opened 2022-12-07
jetty 10.0.x disable concurrent build abort previous

olamy:
- disable concurrent builds by aborting previous one
BUT we will not know which change break the build!

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,6 +7,7 @@ pipeline {
     skipDefaultCheckout()
     durabilityHint('PERFORMANCE_OPTIMIZED')
     buildDiscarder logRotator( numToKeepStr: '60' )
+    disableConcurrentBuilds abortPrevious: true
   }
   stages {
     stage("Parallel Stage") {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -2,11 +2,12 @@
 
 pipeline {
   agent any
-  // save some io during the build
   options {
     skipDefaultCheckout()
+    // save some io during the build
     durabilityHint('PERFORMANCE_OPTIMIZED')
     buildDiscarder logRotator( numToKeepStr: '60' )
+    disableConcurrentBuilds abortPrevious: true
```
joakime, line 10: Wow, this is now built-in to Jenkins?  

In the past, a script was how that was done.

joakime:
> * BUT we will not know which change break the build!

True, this might be important to know for the mainline builds `jetty-10.0.x`, `jetty-11.0.x`, and `jetty-12.0.x`.
But for PRs we don't care about the interim commits, only the last one.

Can this be enabled only for PR builds?

joakime:
Not useful anymore

PR #9026, opened 2022-12-08
Avoid Processor creation when possible.

gregw:
Various cleanups from handler-as-processor branch to reduce processor wrapping. Sessions are now touched before handling.


sbordet (approved):

gregw:
I'm not committing this yet, since if #9035 is merged, then this is not needed

gregw:
No longer necessary as #9035 will be merged

PR #9028, opened 2022-12-09
Jetty 12 Handler.Conditional

gregw:
An alternative to #8978
 + All Handlers are Producers and any call to produce MUST result in the Callback being succeeded or failed
 + Some Handlers are Conditional, and only those get the chance to accept/deny requests.


sbordet:
@gregw I like this, comments below.

I think `Handler.Conditional.handle(Request)` should return a `boolean` (and renamed to `accept(Request)`), and we should specify a very strict contract for it (no calling user code, no loading classes, etc.).

There are very few cases (if none at all) where a `Handler.Conditional` would need to wrap the request, and very few where it would need to delegate to a child to decide whether to accept.
Even assuming the worst case that both a conditional parent and a conditional child need to wrap, we don't want to return the child-wrapped request to the outermost caller, to have it call `process(Request, Response, Callback)`, because the conditional parent would expect to see its own wrapper, not the wrapper of its child.

If that's fine, then I think a `boolean` return is enough.

To save allocations, each wrapping in `accept(Request)` could be saved as a request attribute, and pulled out by any `Handler` that put it there, or they can just re-wrap (re-wrapping would cost one more allocation, but again I think it's an incredibly rare case -- if one at all).

`ContextHandler` would just do the checks on path and virtual host, and return `true` (no wrapping happening now).
`ServletContextHandler` would just call `super` and then look into the Servlet mapping and just return `true/false` based on whether there is a mapping (no Servlet class loading now, no request wrapping yet), etc.
`InetAccessHandler` does not need to wrap the request, so a `boolean` is just fine.
Say I write a `HTTPMethodHandler` that only accepts `GET` requests, and I nest it to `InetAccessHandler`: again no need to wrap the request, so `boolean` is enough.

If the above holds (`boolean` return value and strict contract on `accept(Request)`), then all the work (wrapping, etc.) is done in `process(Request, Response, Callback)` with no need for `request.accept()` or return a `Processor`, etc. which would simplify a lot and keep a nice-to-use and easy-to-document APIs to `Handler` writers.

joakime:
I would have preferred to see the "Condition" be a `Predicate<Request>`

That way we can combine all sorts of things with the JVM built-in logic for `and` / `or` / `not` (that Predicate gives us).

Some example Predicates we have already ..

* InetAcccess
* VirtualHosts
* Named Connector Selection (finally break this out of the virtual hosts logic)
* Connection security (only select if connection is secure)

That way we can attach the conditions as a `Predicate<Request>` to any Handler (a ContextHandler, a Handler.Wrapper, etc)

gregw:
Thanks for the feed back, i agree there is something here.... and a boolean return would be better. But I've put it aside for another similar idea of a more general contract (also returning boolean), but may come back to this if that fails.

It turns out that once we don't have modal APIs we don't have race conditions on boolean returns.

I also have a sneaky plan to well support the tryPaths semantic by having a wrapped response that turns 404s into noops.

I'll have some more details Monday.

gregw:
Not the preferred solution

PR #9037, opened 2022-12-12
response#getWriter should throw UnsupportedEncodingException in case of bad character encoding used

olamy:
TCK issue 
```
response.setCharacterEncoding("allez-les-bleus");
then 
response.getWriter(); 
```
According to TCK and servlet api this should throws `UnsupportedEncodingException` 
but actually EncodingHttpWriter catch it then wrap it in new  RuntimeException 

@gregw I was not sure how to fix that.
I would have preferred to change the constructor (to avoid such `catch` and `instanceof` block 
`public EncodingHttpWriter(HttpOutput out, String encoding)`
to
`public EncodingHttpWriter(HttpOutput out, String encoding) throws IOException`


gregw:
@olamy changing constructor to throw is fine. It is only used in 1 place.

gregw (approved):
LGTM

PR #9040, opened 2022-12-12
Fixes #9038 - Jetty 12 - Review EE10 Http[Input|Output].Interceptor APIs

### LINKED ISSUE #9038: Jetty 12 - Review EE10 Http[Input|Output].Interceptor APIs
**Jetty version(s)**
12+

**Description**
`HttpInput.Interceptor` has the same signature as `ContentSourceTransformer`, and should have a similar contract.
Furthermore, the current implementation has a API contract that is not precise, so it is possible that an `Interceptor` implementation releases the input (raw) chunk (which goes back into the pool), but the `HttpInput` implementation still references it, causing race conditions.
It would be best if the API contract could be aligned with the one in `ContentSourceTransformer`.

Also, `HttpOutput.Interceptor` should be a sub-interface of `Content.Sink` as their `write()` methods have the same signature.

sbordet:
Removed HttpInput.Interceptor and HttpOutput.Interceptor.
Simplified AsyncContentProducer.

### DIFF
--- a/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
+++ b/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
@@ -13,7 +13,6 @@
 
 package org.eclipse.jetty.ee10.servlet;
 
-import java.io.IOException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Condition;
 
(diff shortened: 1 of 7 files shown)


jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
```diff
@@ -303,197 +257,71 @@ boolean isUnready()
...
     }
 
-    private Content.Chunk produceRawChunk()
+    private Content.Chunk readChunk()
     {
         Content.Chunk chunk = _servletChannel.getServletContextRequest().read();
         if (chunk != null)
         {
-            _rawBytesArrived += chunk.remaining();
+            _bytesArrived += chunk.remaining();
```
gregw, line 314: It might be a tiny bit more efficient to use _bytesArrived > 0 in the test below for setting the first byte time.

gregw (approved):
This is a good change. Lots of old code deleted

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
sbordet: @gregw I think there are cases where the chunk may be empty, for example `EOF` or `Trailers`, so I think it's more correct to compare to `MIN_VALUE`.

lorban (approved):
LGTM

PR #9048, opened 2022-12-13
Test for ee10 response commit

gregw:
This is a test and a partial fix demonstration that the response committed state is not correctly handled in ee10.

Currently the `ServletRequestState` class tracks the `OutputState` which includes the `COMMITTED` state.  But currently nothing calls `org.eclipse.jetty.ee10.servlet.ServletRequestState#commitResponse`, so that state is never entered.
The partial fix is to use the committed state from the core request.   If this is sufficient then the `OutputState` needs to be reviewed to see what states it should track.


gregw:
@lachlan-roberts @janbartel do either one of you want to take this one?
@olamy FYI

gregw:
@lachlan-roberts @janbartel I think `ServletRequestState` needs a good review as there are many warnings and a few methods that are never called.

lachlan-roberts:
@gregw were you expecting the test cases to fail here?
They pass for me.

gregw:
@lachlan-roberts my change to use the core isCommitted state fixed the test.... but it leaves kruft in HttpOutput, as the committed state is still there and not used... as is a method for intermediate 1xx responses reopening the printwriter after commit.    So it needs review and cleanup

gregw:
@lachlan-roberts can you also review org.eclipse.jetty.ee10.servlet.ServletRequestState#asyncError which is not called.  Need to check if that is correct and that onError is correctly called when necessary.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
```diff
@@ -292,11 +292,14 @@ private Content.Chunk produceChunk()
...
+                else
+                {
+                    _servletChannel.getState().onContentAdded();
+                }
             }
 
-            assert _chunk != null;
             // Release the chunk immediately, if it is empty.
-            if (!_chunk.hasRemaining() && !_chunk.isTerminal())
+            if (_chunk != null && !_chunk.hasRemaining() && !_chunk.isTerminal())
```
lachlan-roberts, line 302: @sbordet can you review this change.
Looks like if we ever hit line 281 then `_chunk` will be null here.

lachlan-roberts:
@gregw can you review, can't request it through github because you created the PR.
I opened separate issue for https://github.com/eclipse/jetty.project/issues/9060.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
gregw: I think we need to avoid using isTerminal as a condition for releasing.  We should just release any chunk we are given.

gregw:
@lachlan-roberts I can't review as I created the PR.  Other than the question about releasing terminal chunks (which @sbordet is addressing in another PR anyway), I think this is good.   So if you think so too, then you can approve and I will merge.

olamy:
@lachlan-roberts @gregw I have just tested this PR with TCK and I get this error 
```
Exception in thread "TestTimer" java.lang.NullPointerException: Cannot invoke "org.eclipse.jetty.server.handler.ContextHandler$ScopedContext.run(java.lang.Runnable)" because the return value of "org.eclipse.jetty.ee10.servlet.AsyncContextEvent.getContext()" is null
	at org.eclipse.jetty.ee10.servlet.ServletRequestState.runInContext(ServletRequestState.java:1148)
	at org.eclipse.jetty.ee10.servlet.ServletRequestState.complete(ServletRequestState.java:692)
	at org.eclipse.jetty.ee10.servlet.AsyncContextState.complete(AsyncContextState.java:61)
	at com.sun.ts.tests.servlet.spec.async.TestServlet$1.run(TestServlet.java:47)
	at java.base/java.util.TimerThread.mainLoop(Timer.java:566)
	at java.base/java.util.TimerThread.run(Timer.java:516)
```

lachlan-roberts:
@olamy can you please test this again with TCK

olamy:
> @olamy can you please test this again with TCK

no more NPE but this increase the failure number by 7

gregw:
@olamy can you link us to the new failures?

gregw:
@olamy I just synced this branch with head, so that might fix some unrelated failures?

olamy:
@gregw re run TCK with synced branch and definitely get better result!

Screenshot 2022-12-22 at 7 52 15 am

janbartel (approved):

PR #9053, opened 2022-12-14
Jetty 12 - Fix wasted effort on empty `IncludeExcludeSet`

joakime:
Improve (slightly) the performance of `IncludeExcludeSet` when it comes to an empty exclude set.

PR #9055, opened 2022-12-14
PathMappings optimizations

gregw:
Avoid iterations if only ServletPathSpec instances
Avoid tests for empty mappings.
Better reset implementation

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
@@ -17,8 +17,10 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.function.Predicate;
@@ -43,12 +45,13 @@ public class PathMappings<E> implements Iterable<MappedResource<E>>, Dumpable
     private static final Logger LOG = LoggerFactory.getLogger(PathMappings.class);
     private final Set<MappedResource<E>> _mappings = new TreeSet<>(Comparator.comparing(MappedResource::getPathSpec));
 
-    private boolean _nonServletPathSpecs;
+    /**
+     * When _orderIsSignificant is true, the order of the MappedResources is significant and a match needs to be iteratively
+     * tried against each mapping (ordered by group then add order) to find the first that matches.
+     */
+    private boolean _orderIsSignificant;
     private boolean _optimizedExact = true;
-    private final Index.Mutable<MappedResource<E>> _exactMap = new Index.Builder<MappedResource<E>>()
-        .caseSensitive(true)
-        .mutable()
-        .build();
+    private final Map<String, MappedResource<E>> _exactMap = new HashMap<>();
     private boolean _optimizedPrefix = true;
     private final Index.Mutable<MappedResource<E>> _prefixMap = new Index.Builder<MappedResource<E>>()
         .caseSensitive(true)
@@ -94,7 +97,7 @@ public class PathMappings<E> implements Iterable<MappedResource<E>>, Dumpable
         _optimizedExact = true;
         _optimizedPrefix = true;
         _optimizedSuffix = true;
-        _nonServletPathSpecs = false;
+        _orderIsSignificant = false;
         _servletRoot = null;
         _servletDefault = null;
     }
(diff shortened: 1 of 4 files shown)


gregw:
This is a more comprehensive alternative to #9053

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -123,29 +134,97 @@ public List<MappedResource<E>> getMatches(String path)
...
+            MatchedPath matchedPath = candidate.getPathSpec().matched(path);
+            if (matchedPath != null)
+                return new MatchedResource<>(candidate.getResource(), candidate.getPathSpec(), matchedPath);
+        }
+
+        if (!_suffixMap.isEmpty())
+        {
+            int i = 0;
+            // Loop through each suffix mark
+            // Input is "/a.b.c.foo"
```
joakime, line 220: What if the input is `/a.b.c/d.e.foo` ?

We don't want to do `path.indexOf('.', i+1)` within the path portion, only the last segment.
This needs test cases.

joakime (changes requested):
Needs more tests

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
gregw: Good question.    The impl for suffixes has not been changed, so this may be an existing problem.... well it is not really a problem, as I think it will give the right result, but will be inefficient as it will look for suffixes that could never exist.
I'll write another test case

gregw: @joakime more tests added plus a small optimization for the "/a.b/c.d.foo" case.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -123,29 +134,97 @@ public List<MappedResource<E>> getMatches(String path)
     {
         boolean isRootPath = "/".equals(path);
 
-        List<MappedResource<E>> ret = new ArrayList<>();
+        if (_mappings.isEmpty())
```
janbartel, line 138: Could actually move this test to the first line of the method.

janbartel (dismissed):

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
gregw: done

lorban:
It feels like a set of micro benchmarks would help now that this class' perf appeared on the radar.

gregw:
> It feels like a set of micro benchmarks would help now that this class' perf appeared on the radar.

Maybe, but these changes are primarily removing work that was never actually necessary, so it is more of a bug fix than an optimization.    I don't think we should hold up a merge of this whilst over analyzing it.      I'd like to get this merged and only if we then see it on flame graphs should we drill down with micro benchmarks.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -121,31 +135,117 @@ public List<MatchedResource<E>> getMatchedList(String path)
...
+        // Try a suffix match
+        if (!_suffixMap.isEmpty())
+        {
+            int i = Math.max(0, path.lastIndexOf("/"));
+            // Loop through each suffix mark
+            // Input is "/a.b.c.foo"
+            //  Loop 1: "b.c.foo"
+            //  Loop 2: "c.foo"
+            //  Loop 3: "foo"
+            while ((i = path.indexOf('.', i + 1)) > 0)
```
joakime, line 224: There are no test cases (that I can find) that enter this while loop.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -173,19 +273,9 @@ public MatchedResource<E> getMatched(String path)
...
-                                if (candidate == null)
-                                    continue;
-
-                                matchedPath = candidate.getPathSpec().matched(path);
-                                if (matchedPath != null)
-                                {
-                                    return new MatchedResource<>(candidate.getResource(), candidate.getPathSpec(), matchedPath);
-                                }
-                            }
+                            MappedResource<E> exact = _exactMap.get(path);
```
joakime, line 276: There are no test cases that enter this condition.

joakime: After the commit with the `testOptimalExactIterative` this line is still not being hit

joakime (dismissed):
Approved.
Should probably merge PR #9071 first (or revert your IncludeExcludeSet change in this PR)

joakime (approved):

Issue #9051, opened 2022-12-13
Review Jetty-12 DelayedHandler


gregw:
**Jetty version(s)**
12

**Enhancement Description**
Jetty-12 has generalized the previous jetty delayed dispatch mechanism.   Currently there are `Handler`s that can delay dispatch until:
 + some content is available
 + form content has been read and converted to a form
 + multipart content has been asynchronously read and converted to parts.

However, it is possible that a deployment might need multiples of these behaviors and it is undesirable to have 3 such handlers stacked up.   Thus we should consider refactoring so that there is a single `DelayedHandler` that can optionally perform various types of delay depending on content-type.

Furthermore, whilst the downstream handling of forms appears to have been updated to use any form read by this handler, the same has not been done for multiparts.   The same pattern of looking for the processed content in the request attributes should be applied to parts.

gregw:
I have a draft of this, but it is waiting for #9035 to be merged

gregw:
#9056 has improved the implementation and merged the multiple handlers into a single one.
However, the API for extending it is not satisfactory and more thought is needed, so this issue should remain open until a better API is developed.

gregw:
@sbordet Currently the problematic API is:
```java
    protected DelayedProcess newDelayedProcess(boolean contentExpected, String contentType, MimeTypes.Type mimeType, Handler handler, Request request, Response response, Callback callback)
```

So first question: Should this handle be able to delay for non-content related purposes?   We have no real use-cases, but we do have a test case for that, hence the need to pass `contentExpected`.

A good first simplification could be to rename the handler to `DelayedContentHandler` and to remove this parameter, so that if there is no content, then there is never any delay:


```java
    protected DelayedProcess newDelayedProcess(String contentType, MimeTypes.Type mimeType, Handler handler, Request request, Response response, Callback callback)
```

We are passing both `contentType` and `mimeType` so we can switch efficiently on the later to find cases for known types, but we also need the former so we can extract a boundary from it for multipart.
It is a 1 liner to get the mimeType from the contentType, so perhaps we just move that into the method ?:

```java
    protected DelayedProcess newDelayedProcess(String contentType, Handler handler, Request request, Response response, Callback callback)
```

Alternately, we could have a `DelayedProcessFactory` pattern:

```java
    protected DelayedProcessFactory getDelayedProcessFactory(String contentType);
    interface DelayedProcessFactory
    {
[...]

PR #9056, opened 2022-12-15
Alternate DelayedHandler & ThreadLimitHandler implementations #9051

gregw:
An alternate DelayedHandler for #9051


gregw:
@sbordet let's experiment with alternate DelayedHandlers in this branch

gregw:
We need to also review `ThreadLimitHandler`, specifically for callback in scope that already have a permit.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
```diff
@@ -207,6 +207,25 @@ public Type getBaseType()
         })
         .build();
 
+    public static Type getBaseType(String contentType)
+    {
+        if (StringUtil.isEmpty(contentType))
+            return null;
+        Type type = CACHE.getBest(contentType);
```
joakime, line 215: What happens if the mime-type is a substring of another mime-type?

Eg: `application/xml` vs `application/xml-dtd` ?
or `video/h264` and `video/h264-rcdo` and `video/h264-svc`

gregw: Note this is only for our known types in the enum. But also the code below. We check for how it terminates.
I'll add some more explicit tests for that.

gregw:
@simone @lorban This is my refactor of the `DelayedHandler` into a single class for all content types.

Not that current HEAD and boolean handler branch have a bug in `DelayedHandler`.  We can't call process from a demand callback because demand callbacks are serialized.  If the process blocks reading, then it will register a demand callback that can never be delivered as the previous has not returned.   This branch fixes it with an execute... but perhaps that can be conditional on the invocation type of the handler?

This PR also fixed the reentrant issue of ThreadLimitHandler

Also added tests for delayed multipart handling, which has highlighted a problem @lachlan-roberts @janbartel the current implementation of getParts does not appear to be setting the files directory from any configuration from the webapp.   Also if we use the DelayedHandler, then it might not be in the context and thus might not know the temp directory.  Perhaps it would be good enough to move the files into the configured tmp? Or perhaps we need to only use delayed handlers in context... in which case we need to think about how that works with servlet API wrappers.     Either way, we need to improve this PR in how it finds the temp directory, as it should be able to use a server or non-webapp component for that.

gregw:
@sbordet @lorban I've made a lot of changes after our chat last night to both ThreadLimit and Delay handlers:
 + a bit of restructuring has allowed to avoidance of a lot of lambda allocation
 + Checking for isDone in the delay call is used to detect if we are still in scope of process, and thus if we can call process without execution
 + Thread limit now has a Permit interface, which can be used to avoid a completable future if possible, else it has to execute.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -409,6 +414,12 @@ static void write(Sink sink, boolean last, String utf8Content, Callback callback
         /**
          * <p>Writes the given {@link ByteBuffer}, notifying the {@link Callback}
          * when the write is complete.</p>
+         * <p>Implementations guarantee that calls to this method are safely reentrant so that
+         * stack overflows are avoided in the case of mutual recursion between the execution of
+         * the {@code Callback} and a call to this method.  Invocations of the passed
+         * {@code Callback} are serialized and a callback for a completed {@code write} call is
+         * not invoked until any previous {@code write} callback has returned.
+         * Thus the {@code Callback} should not block waiting for a callback of a future write call.</p>
```
sbordet: It's very common to do:

```java
void someMethod(Response response, Callback callback) {
  response.write(false, bb1, Callback.from(() -> response.write(true, bb2, callback), callback::fail);
}
```

I don't think you can specify what you added in the javadocs -- an implementation cannot do anything about the example application code above.
That's why we use `IteratingCallback` for writes, also at the application level.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -183,6 +184,17 @@ static String getPathInContext(Request request)
      */
     HttpFields getHeaders();
 
+    /**
+     * @param demandCallback the demand callback to invoke where there is a content chunk available.
```
sbordet: where -> when

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -183,6 +184,17 @@ static String getPathInContext(Request request)
      */
     HttpFields getHeaders();
 
+    /**
```
sbordet, line 187: Add `{@inheritDoc}` here.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -183,6 +184,17 @@ static String getPathInContext(Request request)
...
+     * @param demandCallback the demand callback to invoke where there is a content chunk available.
+     *                       In addition to the invocation guarantees of {@link Content.Source#demand(Runnable)},
+     *                       this implementation serializes the invocation of the {@code Runnable} with
+     *                       invocations of any {@link Response#write(boolean, ByteBuffer, Callback)}
+     *                       {@code Callback} invocations.
```
sbordet, line 193: This does not feel right at all.
Are we saying that we cannot read a large upload _and_ write a large download concurrently?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -65,6 +65,18 @@ public interface Response extends Content.Sink
...
     CompletableFuture<Void> writeInterim(int status, HttpFields headers);
 
+    /**
+     * @param last whether the ByteBuffer is the last to write
+     * @param byteBuffer the ByteBuffer to write
+     * @param callback the callback to notify when the write operation is complete
+     *                 In addition to the invocation guarantees of {@link Content.Sink#write(boolean, ByteBuffer, Callback)},
+     *                 this implementation serializes the invocation of the {@code Callback} with
+     *                 invocations of any {@link Request#demand(Runnable)} {@code Runnable} invocations.
+     * @see Content.Sink#write(boolean, ByteBuffer, Callback)
```
sbordet, line 79: Ditto.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextRequest.java
```diff
@@ -34,7 +34,7 @@ protected ContextRequest(ContextHandler.ScopedContext context, Request request)
     @Override
     public void demand(Runnable demandCallback)
     {
-        super.demand(() -> _context.run(demandCallback, this));
+        super.demand(new ContextDemand(demandCallback));
```
sbordet: The previous version was clearer as it did not need a lookup elsewhere to read the code.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
```diff
@@ -39,29 +47,79 @@ public boolean process(Request request, Response response, Callback callback) th
...
+                        break loop;
+                    }
+                    break;
+                default:
+                    break;
+            }
+        }
+
+        MimeTypes.Type mimeType = MimeTypes.getBaseType(contentType);
+        DelayedProcess delayed = newDelayedProcess(contentExpected, contentType, mimeType, next, request, response, callback);
```
sbordet, line 84: Feels clunky, as `newDelayProcess()` "filters" whether to delay or not, which is something that is already done above, but also allocates, and takes 7 parameters, which is too many.

Why can we not have a `boolean canDelay(request)` that takes the decision of whether delaying, and a `delay(request, response, callback) that only needs to lookup the mimeType to actually delay?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
```diff
@@ -89,166 +147,194 @@ protected Callback getCallback()
...
         }
 
         @Override
-        protected void delay(DelayedProcess request)
+        protected void delay()
         {
-            request.getRequest().demand(request);
+            Content.Chunk chunk = super.getRequest().read();
+            if (chunk == null)
+                getRequest().demand(this::onContent);
```
sbordet, line 179: Braces around this line.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
```diff
@@ -89,166 +147,194 @@ protected Callback getCallback()
...
+        public void onContent()
         {
-            if (!request.getConnectionMetaData().getHttpConfiguration().isDelayDispatchUntilContent())
-                return null;
-            if (FormFields.getFormEncodedCharset(request) == null)
-                return null;
-
-            return new FormDelayedProcess(next, request, response, callback);
+            // We must execute here, because demand callbacks are serialized and process may block on a demand callback
+            getRequest().getContext().execute(this::process);
```
sbordet, line 197: I am not a fan of this execution, nor the fact that we read before demanding.
And we should check for the `InvocationType` of the child `Handler` to know if it's non-blocking, no?

Feels like we should revert to Jetty 11 behavior where delaying is done in the implementation, not in a Handler.

sbordet (changes requested):
Feels like we had a good solution (the Jetty 11 one), that we want to replace with something that is becoming more and more complicated, and that in order to implement it we need to put big constraints on the implementation with more and more complicated code that was almost a one-liner in Jetty 11.
I fail to see the benefits.
I'm leaning towards removing `DelayedHandler` and replacing it with the Jetty 11 implementation.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/DelayedHandlerTest.java
```diff
@@ -179,6 +177,74 @@ public void testOnContent() throws Exception
             @Override
             public boolean process(Request request, Response response, Callback callback) throws Exception
             {
+                // Check that we are not called via any demand callback
+                ByteArrayOutputStream out = new ByteArrayOutputStream(8192);
+                new Throwable().printStackTrace(new PrintStream(out));
+                String stack = out.toString(StandardCharsets.ISO_8859_1);
+                assertThat(stack, not(containsString("DemandContentCallback.succeeded")));
+                assertThat(stack, not(containsString("UntilContentDelayedProcess.onContent")));
```
lorban: These two assertions are quite weak as renaming the classes would render them meaningless.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/DelayedHandlerTest.java
```diff
@@ -179,6 +177,74 @@ public void testOnContent() throws Exception
...
+        {
+            @Override
+            public boolean process(Request request, Response response, Callback callback) throws Exception
+            {
+                // Check that we are not called via any demand callback
+                ByteArrayOutputStream out = new ByteArrayOutputStream(8192);
+                new Throwable().printStackTrace(new PrintStream(out));
+                String stack = out.toString(StandardCharsets.ISO_8859_1);
+                assertThat(stack, not(containsString("DemandContentCallback.succeeded")));
+                assertThat(stack, not(containsString("UntilContentDelayedProcess.onContent")));
```
lorban: Ditto.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/IO.java
```diff
@@ -559,6 +559,25 @@ public static long write(GatheringByteChannel out, ByteBuffer[] buffers, int off
 
         return total;
     }
+
+    /**
+     * <p>Convert an object to a {@link File} if possible.</p>
+     * @param fileObject A File, String, Path or null to be converted into a File
+     * @return A File representation of the passed argument or null.
+     */
+    public static File asFile(Object fileObject)
```
lorban, line 568: This doesn't look like a routine we want to generalize.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
```diff
@@ -39,29 +47,79 @@ public boolean process(Request request, Response response, Callback callback) th
...
         if (delayed == null)
             return next.process(request, response, callback);
 
-        delay(delayed);
+        delayed.delay();
         return true;
     }
 
-    protected DelayedProcess newDelayedProcess(Handler next, Request request, Response response, Callback callback)
+    protected DelayedProcess newDelayedProcess(boolean contentExpected, String contentType, MimeTypes.Type mimeType, Handler handler, Request request, Response response, Callback callback)
```
lorban, line 92: This looks quite complex to override, this contract should be simplified.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
lorban: This impl is probably acceptable for now, but I agree that trading off a dispatch for the guarantee of being called when content is readable doesn't sound like a great deal.

lorban (approved):
I believe this is an improvement over what we currently have, but I agree that we should revisit this implementation as this started as a simple impl that badly grew in complexity over time.

gregw:
@sbordet & @lorban , whilst I'm open to looking at better ways of doing this, I'm not sure I agree with your concerns about complexity:
 + This PR is doing a lot LOT more.  Previously we had no pre-reading mechanism for forms or multi-part, we simple could delay the first dispatch
 + It includes the ThreadLimitHandler, which is also less/broken in 11, as it does not count threads in async callbacks of onDataAvailable and onWritePossible
 + The Content/request/response javadoc in here is simply describing existing behavior.  If you don't like that behavior, then that is an entirely different PR.
 + The previous delay dispatch code was scattered over multiple different connection types and required special conditional handling in 3 or 4 methods in each connection type. I don't see the UntilContent approach here as that much more complex.

If somebody can come up with a better approach, then great... but for now this is better (at least not broken) that what is in HEAD.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: I'm not advocating different behavior, I'm just documenting what it currently does.
There is nothing preventing the code example that you wrote as it does not block waiting for content in a callback.   A bad example would be:
```java
void someMethod(Response response, Callback callback) {
  response.write(false, bb1, Callback.from(() -> try (Blocker b = new Blocker() { response.write(true, bb2, blocker); blocker.block()} , callback::fail);
}

```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
gregw: No.  We can read and write at the same time, but only a single callback will be called at once.   Note this is the same existing serialization that happens on `onDataAvailable` and `onWritePossible` in the servlet API.
Besides, I thought it was your idea to serialize all the actions.

Anyway, this is just documenting what is.  If you want to change the behavior, then that is a different PR.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
gregw: ditto

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextRequest.java
gregw: It was harder to debug, as the lambda is not clear in the stack traces and flame graphs.  How about if it is called `OnContextDemand` ?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
gregw: Yeah this method is definitely chunky and clunky.  Plenty of scope for a  better mechanism.    I had a version that hard coded some of the contentExpected logic and then just select a delay factory by contentType.   But it was not truly generic and I could not implement some of the existing tests with it.     This approach captures all the known delay decisions we have, so think of it as a specification for a better API.

Note that the `canDelay` and `delay` approach doesn't work well as you have to do double work on content-type and headers as they might be needed in the boolean decision and then redone when making the Delayed instance.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
gregw: `InvocationType` of the Handler does not help us much as there is still a behavior change, as demand callbacks are allowed concurrently with process, but not with another demand.    Making this execute conditional on Invocation type might be OK, but calculating  the type can be expensive as it needs to walk the entire handler tree and combine the types.  Perhaps that could be done in `doStart`?

Note that by reading the chunk initially, we avoid this execution unless there is a delay.  See the unit tests that check this.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/DelayedHandlerTest.java
gregw: Yep.  But how else do you test for the negative?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/IO.java
gregw: It is called in 2 places and I think there are probably more that could use it.   When users set a configuration as a bean or an attribute, it can be unclear if they are specifying a String filename, File, Path etc. (probably should support URI, URL and Resource as well).  So it is useful to be forgiving of such configuration issues.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
gregw: Definitely.   Proposals?

gregw:
I see the main con of this approach is that sometimes we need to execute to work around serialization.    It would be simpler if there was no serialization, but that is a bigger design decision.  We've always had async serialization for async servlets and would need to implement for the servlet API anyway.
I think the until content delay could be hard coded lower down before serialization, but we are not extensible there and doing arbitrary pre-loading of various content types would be difficult.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
gregw: Note also, that if somebody is writing an asynchronous application, then they are unlikely to put this handler out front.  It should only be used in front of blocking applications.

gregw:
For the record, here the complexity of the jetty-11 approach just for HTTP/1 is:
 + 10 references to  _delayedForContent in `HttpChannelOverHttp`
 + `content(Bytebuffer)` is intercepted to see the first content
 + `contentComplete` is intercepted just in case the body is empty
 + `earlyEOF` needs special handling so we generate errors the same way as no delay
 + `headerComplete` makes the decision to delay or not.
 + `onIdleTimeout` is intercepted to handle a timeout during delay.

I think the same basic approach can be implemented in 12, but mostly on `org.eclipse.jetty.server.internal.HttpConnection.RequestHandler` where the parser handler is now.  One complication is that timeout events are no longer delivered there, rather directly to the HttpChannel, which would know an onRequest `Runnable` had been returned, but not yet executed.   To mimic 11 we would setup the EOF exception and then somehow prompt the channel to execute the `Runnable` it was holding.  Doable, but logic split between two classes, one HTTP/1 specific and the other protocol independent.   Also would not do the pre-fetching of form/multipart.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/DelayedHandlerTest.java
gregw: I made a bit less fragile by using reflection (there is a sentence I never thought I'd type :)

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/DelayedHandlerTest.java
gregw: as above

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: @gregw I meant the whole paragraph.
It is wrong because it does not describe the current behavior in general, perhaps just that of the server-side implementation.
Since it's only specific to how the server is implemented, perhaps just move it to the `Response` class.

And BTW I am surprised even by the server side behavior, because pending writes are protected by throwing `WritePendingException`, so there is no need to serialize the write callbacks.
And we always solved the stack overflow with `IteratingCallback`, so again no need to serialize the write callbacks.
What am I missing?

jetty-core/jetty-server/src/main/config/modules/delay-until-content.mod
```diff
@@ -0,0 +1,22 @@
+
+[description]
+Applies DelayedHandler to entire server.
+
+[tags]
+server
+
+[description]
```
sbordet: Two `[description]` sections. Keep it as the first section of the file, but use the more detailed one.

jetty-core/jetty-server/src/main/config/modules/delay-until-content.mod
```diff
@@ -0,0 +1,22 @@
...
+Delay request handling until any body content has arrived, to minimize blocking.
```
sbordet: Delay -> Delays

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/FormFields.java
```diff
@@ -60,21 +59,20 @@ public static Charset getFormEncodedCharset(Request request)
 
     public static CompletableFuture<Fields> from(Request request)
     {
-        Object attr = request.getAttribute(FormFields.class.getName());
-        if (attr instanceof FormFields futureFormFields)
-            return futureFormFields;
+        // TODO make this attributes provided by the ContextRequest wrapper
+        int maxFields = getRequestAttribute(request, FormFields.MAX_FIELDS_ATTRIBUTE);
+        int maxLength = getRequestAttribute(request, FormFields.MAX_LENGTH_ATTRIBUTE);
```
sbordet, line 64: Can you call the 2 parameter overload, so these attributes are retrieved only in 1 place?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextRequest.java
```diff
@@ -66,4 +66,20 @@ public Object getAttribute(String name)
             default -> super.getAttribute(name);
         };
     }
+
+    private class OnContextDemand implements Runnable
```
sbordet, line 71: Please add a comment such as "using class instead of lambda for better stack traces", because screams like "inline this class".

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
sbordet: Given all the inefficiencies that `DelayedHandler` introduces, I would not be worried about double work.
Once we have people overriding a 7 parameter method for their own subclasses we will be in troubles if we need one more parameter or one less, and we won't be able to change the signature.

Currently, none of the overrides of `newDelayedProcess()` make use of _any_ of the parameters passed in.
This means the base implementation can take just the request as the parameter and just return a `boolean` and be renamed `canDelay(Request)`.

As for `DelayedProcess` subclasses, the only extra parameter that seems to be needed is the contentType, only used by form and multipart, which are less common and we can afford one more lookup.
Therefore we can get rid of `DelayedProcess` as well and just have `delay(Request, Response, Callback)` if `canDelay(Request)` returns `true`.

Finally, I liked better the previous approach because it was more flexible. Before, I would delay only multipart by allocating a `DelayedHandler.UntilMultiPartFormData`.
Now, if I want to delay multipart, I will delay also all other cases.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
sbordet: See previous comment.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
```diff
@@ -89,166 +147,200 @@ protected Callback getCallback()
...
-                return null;
-
-            return new FormDelayedProcess(next, request, response, callback);
+            // We must execute here, because demand callbacks are serialized and process may block on a demand callback
+            getRequest().getContext().execute(this::process);
         }
 
-        @Override
-        protected void delay(DelayedProcess delayed)
+        static class RewindRequest extends Request.Wrapper
```
sbordet: Make it `private`.
Rename to `RequestWithChunk`?
Or perhaps to `DelayedRequest`, so from stack traces we know where it's coming from?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
```diff
@@ -89,166 +147,200 @@ protected Callback getCallback()
...
-            if (boundary == null)
-                return null;
+            // if we are done already, then we are still in the scope of the original process call and can
+            // accept directly, otherwise we must execute a call to process as we are within a serialized
+            // demand callback.
+            futureFormFields.whenComplete(futureFormFields.isDone() ? this::accept : this::acceptAndExecute);
+        }
 
-            return new MultiPartDelayedProcess(next, request, response, callback, boundary);
+        public void accept(Fields fields, Throwable x)
```
sbordet: Make `private` and rename to `process` -- it does not really "accept".

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
```diff
@@ -89,166 +147,200 @@ protected Callback getCallback()
...
+        {
+            if (x == null)
+                super.process();
+            else
+                Response.writeError(getRequest(), getResponse(), getCallback(), x);
         }
 
-        @Override
-        protected void delay(DelayedProcess request)
+        public void acceptAndExecute(Fields fields, Throwable x)
```
sbordet: Make it `private` and rename to `executeProcess`

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
```diff
@@ -89,166 +147,200 @@ protected Callback getCallback()
...
+        }
 
-            public MultiPartDelayedProcess(Handler handler, Request wrapped, Response response, Callback callback, String boundary)
+        public void accept(MultiPartFormData.Parts parts, Throwable x)
+        {
+            if (x == null)
             {
-                super(handler, wrapped, response, callback);
-                _formData = new MultiPartFormData(boundary);
                 getRequest().setAttribute(MultiPartFormData.class.getName(), _formData);
```
sbordet: Move this to the constructor so it's done only once.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
```diff
@@ -89,166 +147,200 @@ protected Callback getCallback()
...
+        public UntilMultiPartDelayedProcess(Handler handler, Request wrapped, Response response, Callback callback, String contentType)
         {
-            private final MultiPartFormData _formData;
+            super(handler, wrapped, response, callback);
+            String boundary = MultiPart.extractBoundary(contentType);
+            _formData = boundary == null ? null : new MultiPartFormData(boundary);
+        }
 
-            public MultiPartDelayedProcess(Handler handler, Request wrapped, Response response, Callback callback, String boundary)
+        public void accept(MultiPartFormData.Parts parts, Throwable x)
```
sbordet: `private` and rename to `process`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
```diff
@@ -89,166 +147,200 @@ protected Callback getCallback()
...
             }
+            else
+            {
+                Response.writeError(getRequest(), getResponse(), getCallback(), x);
+            }
+        }
 
-            @Override
-            public void accept(MultiPartFormData.Parts parts, Throwable x)
+        public void acceptAndExecute(MultiPartFormData.Parts parts, Throwable x)
```
sbordet: `private` and rename to `executeProcess`

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
```diff
@@ -89,166 +147,200 @@ protected Callback getCallback()
...
+                // accept directly, otherwise we must execute a call to process as we are within a serialized
+                // demand callback.
+                _formData.whenComplete(_formData.isDone() ? this::accept : this::acceptAndExecute);
+            }
+        }
 
-            public void whenDone()
+        private void readAndParse()
+        {
+            while (!_formData.isDone())
```
sbordet: This can be a `while (true)`, since we read until the last chunk anyway.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ThreadLimitHandler.java
```diff
@@ -204,7 +208,7 @@ private Remote getRemote(Request baseRequest)
         Remote remote = _remotes.get(ip);
         if (remote == null)
         {
-            Remote r = new Remote(ip, limit);
+            Remote r = new Remote(baseRequest.getContext(), ip, limit);
```
sbordet, line 201: [I could only add a comment to modified code sections, so the comment below refers to other parts of code].

Since it's almost 2023, can we make the parameterless constructor default to `true` for the rfc7239 parameter?

Also, method `getAndClose()` is unused and should be removed.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ThreadLimitHandler.java
```diff
@@ -305,10 +310,10 @@ protected Callback getCallback()
 
         protected void process() throws Exception
         {
-            CompletableFuture<Closeable> futurePermit = _remote.acquire();
+            Permit futurePermit = _remote.acquire();
```
sbordet: Just call the variable `permit`?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ThreadLimitHandler.java
```diff
@@ -305,10 +310,10 @@ protected Callback getCallback()
...
 
             // Did we get a permit?
-            if (futurePermit.isDone())
+            if (futurePermit.isAllocated())
             {
                 if (LOG.isDebugEnabled())
                     LOG.debug("Threadpermitted {}", _remote);
```
sbordet: Missing a space between `Thread` and `permitted`. Maybe worth logging also the request and the current thread?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ThreadLimitHandler.java
```diff
@@ -318,13 +323,13 @@ protected void process() throws Exception
             {
                 if (LOG.isDebugEnabled())
                     LOG.debug("Threadlimited {}", _remote);
```
sbordet: Missing a space between `Thread` and `limited`. Maybe worth logging also the request and the current thread?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ThreadLimitHandler.java
```diff
@@ -339,45 +344,38 @@ protected void process(CompletableFuture<Closeable> futurePermit)
...
+            else
+                permit.whenAllocated(this::onPermittedContent);
+        }
 
-            super.demand(permittedDemand);
+        private void onPermittedContent(Permit permit)
+        {
+            try
+            {
+                _onContent.run();
```
sbordet: Should not `_onContent` be nulled out here, so that future calls to `demand()` won't throw?
Are we missing a test case?

lorban (approved):
This is an improvement over what we currently have.

sbordet (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ThreadLimitHandler.java
gregw: Indeed. Bug and missing test.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ThreadLimitHandler.java
gregw: I'll add the request and handler. The logger already includes the thread.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
gregw: I'm slightly worried about setting it before we have successfully parsed it.   But I can't see why, so OK.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/FormFields.java
gregw: Yep.  These parameters are still a bit clumsy. Ultimately would be good to do differently

jetty-core/jetty-server/src/main/config/modules/delay-until-content.mod
gregw: Oooh that is a cut and paste issue from threadlimit.mod... which also has 2 descriptions!!! I just thought title and detail!

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: I'm not exactly sure what you are missing.  This was your design to serialize all the async events?  It's not unusual and is the behavior specified for async servlets.  

I don't see how `WritePendingException` helps here? You are allowed to do a write from a write callback, so we don't throw a WPE in that case.   You are just not allowed to wait for the callback of that nested write from within a previous callback.

I think the protection from stack overflows applies to all implementations, client and server.  Then the part about serialization is a corollary of that , so also true in the client. 

I'm pretty sure you have a queue of events in the client that also serialize callbacks, so if one blocks, then the next will never be called.  In fact the client is tougher as a callback can never block without potentially stopping some other unrelated event, whilst the server can block as it will only stop processing of this request.

I think this is documenting existing behavior, but it is kind of out of scope for this PR, so I'll move the second part to the Response.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/FormFields.java
gregw: Turns out no... at least not easily.  The constructors with charset and those without are different paths.

This needs to be cleaned up, but that can be done when fixing the TODO.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ThreadLimitHandler.java
gregw: I added a test that showed the bug... plus another one.
ThreadLimitHandler looks to be very under tested (and always has been).

Perhaps it needs a separate PR?  I think it is still better (or at least broken differently) than HEAD.

sbordet (changes requested):
@gregw you have not addressed a few of my concerns, are you still working on it?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DelayedHandler.java
gregw: `RewindChunkRequest`?

gregw:
> @gregw you have not addressed a few of my concerns, are you still working on it?

@sbordet I wasn't intending to (at least not in this PR) but I've done a few more renames  and comments if that helps??

> Given all the inefficiencies that DelayedHandler introduces, I would not be worried about double work.

I really don't see "all the inefficiencies", especially when compared to IO delays.  There is an execute in some circumstances, but that's not more than the execution strategy would do if we somehow propagated the mechanism back to it.  There is a bit of wrapping in other circumstances, but we do more wrapping in general in 12 and it has not yet shown up on flame graphs.   There is a volatile read on every `read` call, which itches a bit, but reads are the minority of requests and involve much more expensive operations.   It is also doing a lot more useful stuff than previous mechanism.   As an optional handler that will help applications written in blocking style avoid blocking on reads, I think this handler is still a big winner... but we should probably benchmark it to test that.  Obviously if virtual threads are used, then this handler is not so appropriate.  All the more reason to make it truly optional and not baked into core.

> Once we have people overriding a 7 parameter method ...

Yep. As I say, that method is only a stop gap until we think of something better.  The alternatives that I tried (including the boolean version you suggest) did not work for all the unit tests we have, so perhaps we need to remove/change tests to get a better API.  But I didn't want to do that in this PR as I want to ensure we can meet the tests that are there (this area is under tested and I don't think removing tests is the right way to go).  So sure, it needs a better API, but can that be done in a different PR that is focused on just that.  I've added TODO and deprecated the API to make it clear that it is not the final API.

There are still TODOs in the code and I agree more work is necessary.   But let's not make perfect the enemy of good (or not broken) and progress.   I think that this is already too much of an omnibus PR. Work should probably continue in 3 different strands at least:  1) improving javadoc on core classes & resolving any issues about serialization; 2) ThreadLimitHandler needs a lot more testing; 3) DelayedHandler needs a better API for extensibility/plugability/configuration

sbordet (approved):

PR #9057, opened 2022-12-15
Jetty 12.0.x ee10 convert cookie

janbartel:
Fix a number of ee10 cookie parsing issues, and also ensure  `ServletContextRequest.getCookies()` implements spec of returning `null` for no cookies rather than empty list.

Fixes these tck tests:
 com/sun/ts/tests/servlet/api/jakarta_servlet_http/cookie/URLClient.java.getDomainTest
 com/sun/ts/tests/servlet/api/jakarta_servlet_http/cookie/URLClient.java.getPathTest
 com/sun/ts/tests/servlet/api/jakarta_servlet_http/cookie/URLClient.java.setMaxAgeZeroTest
 com/sun/ts/tests/servlet/api/jakarta_servlet_http/httpservletrequest/URLClient.java.getCookiesNoCookiesTest
 com/sun/ts/tests/servlet/pluggability/api/jakarta_servlet_http/cookie/URLClient.java.getDomainTest
 com/sun/ts/tests/servlet/pluggability/api/jakarta_servlet_http/cookie/URLClient.java.getPathTest
 com/sun/ts/tests/servlet/pluggability/api/jakarta_servlet_http/cookie/URLClient.java.setMaxAgeZeroTest
 com/sun/ts/tests/servlet/pluggability/api/jakarta_servlet_http/httpservletrequest/URLClient.java.getCookiesNoCookiesTest

### DIFF
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
@@ -367,12 +367,8 @@ public class HttpCookie
             else
                 DateGenerator.formatCookieDate(buf, System.currentTimeMillis() + 1000L * _maxAge);
 
-            // for v1 cookies, also send max-age
-            if (version >= 1)
-            {
-                buf.append(";Max-Age=");
-                buf.append(_maxAge);
-            }
+            buf.append(";Max-Age=");
+            buf.append(_maxAge);
         }
 
         // add the other fields
--- a/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextRequest.java
+++ b/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextRequest.java
@@ -58,6 +58,7 @@ import jakarta.servlet.http.PushBuilder;
 import org.eclipse.jetty.ee10.servlet.security.Authentication;
 import org.eclipse.jetty.ee10.servlet.security.UserIdentity;
 import org.eclipse.jetty.http.BadMessageException;
+import org.eclipse.jetty.http.CookieCompliance;
 import org.eclipse.jetty.http.HttpCookie;
 import org.eclipse.jetty.http.HttpField;
 import org.eclipse.jetty.http.HttpFields;
@@ -491,13 +492,12 @@ public class ServletContextRequest extends ContextRequest
         public Cookie convertCookie(HttpCookie cookie)
         {
             Cookie result = new Cookie(cookie.getName(), cookie.getValue());
-            // TODO: inbound (client-to-server) cookies don't have all these parameters.
-//            result.setPath(cookie.getPath());
-//            result.setDomain(cookie.getDomain());
-//            result.setSecure(cookie.isSecure());
-//            result.setHttpOnly(cookie.isHttpOnly());
-//            result.setMaxAge((int)cookie.getMaxAge());
-            // TODO: sameSite?
+            //RFC2965 defines the cookie header as supporting path and domain but RFC6265 permits only name=value
+            if (CookieCompliance.RFC2965.equals(getRequest().getConnectionMetaData().getHttpConfiguration().getRequestCookieCompliance()))
+            {
+                result.setPath(cookie.getPath());
+                result.setDomain(cookie.getDomain());
+            }
             return result;
         }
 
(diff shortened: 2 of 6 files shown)


jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
```diff
@@ -367,12 +367,8 @@ else if (version > 1)
             else
                 DateGenerator.formatCookieDate(buf, System.currentTimeMillis() + 1000L * _maxAge);
 
-            // for v1 cookies, also send max-age
-            if (version >= 1)
-            {
-                buf.append(";Max-Age=");
-                buf.append(_maxAge);
-            }
+            buf.append(";Max-Age=");
```
joakime, line 370: IIRC, `Max-Age` shouldn't be sent along with `Expires` on the same cookie.
And starting with RFC6265 the `Expires` attribute is preferred.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextRequest.java
```diff
@@ -482,22 +483,24 @@ public String getAuthType()
         @Override
         public Cookie[] getCookies()
         {
+            List<HttpCookie> httpCookies = Request.getCookies(getRequest());
+            if (httpCookies.isEmpty())
+                return null;
             // TODO: optimize this.
```
joakime: Drop the comment

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
janbartel: @joakime if you look at the code in eg jetty-11, we _have_ been sending `max-age` along with `expires` on RFC2965 cookies:  https://github.com/eclipse/jetty.project/blob/jetty-11.0.x/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java#L355

As version is no longer meaningful for the `jakarta.servlet.http.Cookie` class I didn't include the test, but I can add it back in if you like?

joakime: In Jetty 11 the `Version` field is only generated when the cookie.version is 1.
It defaults to 0.

And the `;Version=` field only exists for REALLY early Mozilla cookies (think pre-ietf, pre-RFC time frame)
IMO, we should drop `Max-Age` entirely, we have `Expires` which is supported in RFC2965 and RFC6265.

Cookies, and what they say, has gotten far more strict over the years (you can thank Firefox and Google for this).
Examples:

* Both browsers ignore the entire `Set-Cookie` line if it has any `$` (prefixed) attributes.
* Both browsers ignore `Set-Cookie` that attempts to set a cookie from a non-secure connection (if the cookie was previously set from the secure location, and this has nothing to do with any of the secure cookie attributes)
* Both browsers ignore the entire `Set-Cookie` line if it has a mangled quoting or escaping (this includes improperly escaped content within the double quotes, or bad pct-encoding)
* Both browsers ignore the entire `Set-Cookie` line if it has a `Version` attribute (of any value)
* The default CORS behaviors are getting more strict (this is the mode that the browser operates in if no CORS headers are sent)
* Chrome ignores `Set-Cookie` on 302 redirects.
* Both browsers ignore the entire `Set-Cookie` line if `SameSite` is set poorly.
* Chrome ignores the `Set-Cookie` line if `SameSite=None` and the `Secure` attribute is missing. (which automatically means you cannot use `SameSite=None` on a normal HTTP connection)
* Both browsers currently reject the entire `Set-Cookie` line if [cookie prefixes](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-cookie-prefixes-00) are used in conjunction with forbidden common attributes. (eg: `__Host` prefix with `Domain` attribute is forbidden)
* Both browsers currently throw warnings if `Max-Age` and `Expires` are both present on a `Set-Cookie` line (despite what the RFC says) (they are evaluating what to do, both seem to indicate that they will just ignore `Max-Age` in this situation in the future)

janbartel: @joakime bottom line is the servlet tck is testing for `Max-Age` being sent (and on a Version=0 cookie):   https://github.com/jakartaee/platform-tck/blob/master/src/com/sun/ts/tests/servlet/api/jakarta_servlet_http/cookie/URLClient.java#L368

Also, RFC6265 has no issue with having both `Expires` and `Max-Age` in the same `Set-Cookie` header:

**Section 4.1.2.2**:
>    If a cookie has both the Max-Age and the Expires attribute, the Max-
>    Age attribute has precedence and controls the expiration date of the
>    cookie.  If a cookie has neither the Max-Age nor the Expires
>    attribute, the user agent will retain the cookie until "the current
>    session is over" (as defined by the user agent).

joakime (approved):

PR #9058, opened 2022-12-15
Enable all jetty-deploy tests

joakime:
Enabling all jetty-deploy tests.
Cleaning up tests for modern techniques.
Removing test resources that are no longer being used. (most moved to the ee specific tests)


jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ContextProvider.java
```diff
@@ -144,6 +146,31 @@ public Map<String, String> getProperties()
...
+    {
+        Properties props = new Properties();
+        try (InputStream inputStream = Files.newInputStream(path))
+        {
+            props.load(inputStream);
+            props.forEach((key, value) -> _properties.put((String)key, (String)value));
+        }
+    }
+
+    public void loadPropertiesFromString(String path) throws IOException
```
joakime, line 169: These 3 replace the old `org.eclipse.jetty.deploy.PropertiesConfigurationManager` functionality (that was removed in Jetty 12)

You can see how this is used from an XML configuration below in the [`src/test/resources/jetty-deploymgr-contexts.xml`](https://github.com/eclipse/jetty.project/pull/9058/files#diff-8dfa3c951d4199a30180d67945ddb463902062b71293950e7ee217e5eb03c635) file

jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ScanningAppProvider.java
```diff
@@ -76,7 +76,7 @@ public void pathChanged(Path path) throws Exception
         @Override
         public void pathRemoved(Path path) throws Exception
         {
-            ScanningAppProvider.this.pathChanged(path);
+            ScanningAppProvider.this.pathRemoved(path);
```
joakime, line 79: Look at that, a legit bug enabling the tests revealed.

janbartel (approved):

PR #9062, opened 2022-12-16
Serialize onCompleteFailure for #9059

### LINKED ISSUE #9059: IteratingCallback not serializing close() and failed()
**Jetty version(s)**
10+

**Description**
Thread1 is inside `IC.process()`, acting on some custom state.

Thread2 calls `IC.close()` or `IC.failed(failure)`, and this may call `onCompleteFailure()` directly, which may (for example) reset the custom state, or null it out, etc.

Therefore, the contract of IC that it will serialize the processing is not respected, because now 2 threads can access the IC custom state.

We should be more precise at serializing in case of a `close()` call, and possibly introduce a `close(Throwable)` or `fail(Throwable)` different from `failed(Throwable)` if necessary to handle the serialization more precisely.

gregw:
Delay `onCompleteFailure` call until process has returned.

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/IteratingCallback.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/IteratingCallback.java
@@ -122,6 +122,7 @@ public abstract class IteratingCallback implements Callback
 
     private final AutoLock _lock = new AutoLock();
     private State _state;
+    private Throwable _failure;
     private boolean _iterate;
 
     protected IteratingCallback()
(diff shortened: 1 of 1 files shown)


jetty-util/src/main/java/org/eclipse/jetty/util/IteratingCallback.java
```diff
@@ -19,103 +19,111 @@
...
+         * sub-task, where the execution has started but the callback
+         * that signals the completion of the asynchronous sub-task
          * may have not yet been invoked.
          */
         SCHEDULED,
-
         /**
-         * Indicates that {@link #process()} has completed the overall job.
+         * Indicates that {@link #process()} has completed the whole
+         * iteration successfully.
```
joakime, line 126: +1 for the clearer javadoc on `Action`

sbordet (approved):

PR #9071, opened 2022-12-19
Improve `IncludeExcludeSet` testing

joakime:
Improve testing and code coverage of `IncludeExcludeSet`.

Updating `IncludeExcludeSet` for best performing `.test(input)` method.
(jmh testing of different algorithms shows this one to be fastest)

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/IncludeExcludeSet.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/IncludeExcludeSet.java
@@ -65,6 +65,7 @@ public class IncludeExcludeSet<T, P> implements Predicate<P>
      */
     public IncludeExcludeSet()
     {
+        // noinspection unchecked
         this(HashSet.class);
     }
 
@@ -75,8 +76,9 @@ public class IncludeExcludeSet<T, P> implements Predicate<P>
      * one for include patterns and one for exclude patters.  If the class is also a {@link Predicate},
      * then it is also used as the item test for the set, otherwise a {@link SetContainsPredicate} instance
      * is created.
-     * @param <SET> The type of a set to use as the backing store
+     * @param <SET> The type of {@link Set} to use as the backing store
      */
+    @SuppressWarnings({"unchecked", "rawtypes"})
     public <SET extends Set<T>> IncludeExcludeSet(Class<SET> setClass)
     {
         try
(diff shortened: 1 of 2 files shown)


jetty-util/src/main/java/org/eclipse/jetty/util/IncludeExcludeSet.java
```diff
@@ -154,25 +157,47 @@ public void exclude(T... element)
...
+     * <p>
+     *     Empty includes means all inputs are allowed.
+     * </p>
+     *
+     * @param t the input argument
+     * @return true if the input matches an include, and is not excluded.
+     */
     @Override
-    public boolean test(P t)
+    public boolean test(P t) // TestNew
```
gregw: removed `// TestNew` comment

jetty-util/src/main/java/org/eclipse/jetty/util/IncludeExcludeSet.java
```diff
@@ -154,25 +157,47 @@ public void exclude(T... element)
...
     {
-        if (!_includes.isEmpty() && !_includePredicate.test(t))
-            return false;
+        if (!_includes.isEmpty())
+        {
+            // We have includes defined, if input has no match, return false immediately
+            if (!_includePredicate.test(t))
+                return false;
+            // We have an include match, let excludes override if needed.
+        }
```
gregw: I don't find the cascading ifs with hanging comment clearer.  How about:
```suggestion
        // If we have defined includes, but none match then
        if (!_includes.isEmpty() && !_includePredicate.test(t))
        {
            // return false immediately, no need to check excluded
            return false;
        }
        
        // We are included by omission or  match, so we now check excludes
```

gregw (changes requested):

jetty-util/src/main/java/org/eclipse/jetty/util/IncludeExcludeSet.java
joakime: This is actually one of the implementations we tried.

This was the "simplified" algorithm from ...
https://github.com/eclipse/jetty.project/blob/fix/jetty-12.0.x/improve-includeexcludeset/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/IncludeExcludeSet.java#L160-L164

It (surprisingly) is 10% slower in the jmh testing than what we are using now.
I'll be happy to adjust the comments though.

joakime: It _should be_ identical, but it wasn't.

jetty-util/src/main/java/org/eclipse/jetty/util/IncludeExcludeSet.java
joakime: Removed.

jetty-util/src/main/java/org/eclipse/jetty/util/IncludeExcludeSet.java
gregw: I'm very dubious of the test if they are not identical!

gregw (approved):

jetty-util/src/main/java/org/eclipse/jetty/util/IncludeExcludeSet.java
joakime: > I'm very dubious of the test if they are not identical!

You aren't the only one, even @sbordet and @lorban had the same initial reaction, and even with some jmh test tweaking couldn't make it identical.
Even testing both techniques side by side on the same jmh run showed the difference! (consistently the same difference each jmh run)

PR #9073, opened 2022-12-19
Fixes retainability of special Chunks

### LINKED ISSUE #8993: `Content.Chunk.isTerminal()` cannot discriminate `EOF` from chunks containing a pooled empty buffer

sbordet:
Alternative to #9018.

* Restored Jetty 11's AsyncContentProducer and related classes.
* Introduced Retainable.canRetain().
* Removed Chunk.isTerminal() and replaced it with alternative method calls.


jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/ResponseNotifier.java
```diff
@@ -368,6 +369,12 @@ public boolean isLast()
                     throw new UnsupportedOperationException();
                 }
 
+                @Override
+                public boolean canRetain()
+                {
+                    throw new UnsupportedOperationException();
```
gregw: Surely this should return false?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
```diff
@@ -631,20 +631,20 @@ public Content.Chunk read()
...
-                        }
-                        state = State.MIDDLE;
+                        Content.Chunk wrapped = Content.Chunk.from(chunk.getByteBuffer(), false, chunk);
+                        chunk.release();
+                        yield wrapped;
+                    }
+                    else
+                    {
+                        chunk.release();
+                        yield Content.Chunk.EMPTY;
```
gregw:
```suggestion
                    chunk.release();
                    yield Content.Chunk.EMPTY;
```

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
```diff
@@ -631,20 +631,20 @@ public Content.Chunk read()
...
-                            chunk.release();
-                            chunk = Content.Chunk.EMPTY;
-                        }
-                        else
-                        {
-                            chunk = Content.Chunk.from(chunk.getByteBuffer(), false, chunk);
-                        }
-                        state = State.MIDDLE;
+                        Content.Chunk wrapped = Content.Chunk.from(chunk.getByteBuffer(), false, chunk);
+                        chunk.release();
```
gregw: It's kind of annoying that we know `Chunk.from` is going to do a retain, that counter acts this release.    Perhaps would be nice to have a version of `from` that doesn't do the retain, so we don't have to do the release?

jetty-core/jetty-http2/jetty-http2-client-transport/src/main/java/org/eclipse/jetty/http2/client/transport/internal/HttpReceiverOverHTTP2.java
```diff
@@ -77,7 +77,9 @@ public Content.Chunk read(boolean fillInterestIfNeeded)
         boolean last = frame.remaining() == 0 && frame.isEndStream();
         if (last)
             responseSuccess(getHttpExchange(), null);
-        return Content.Chunk.from(frame.getData(), last, data);
+        Content.Chunk chunk = Content.Chunk.from(frame.getData(), last, data);
+        data.release();
```
gregw: same pattern of a `retain` followed by a `release`

jetty-core/jetty-http3/jetty-http3-client-transport/src/main/java/org/eclipse/jetty/http3/client/transport/internal/HttpReceiverOverHTTP3.java
```diff
@@ -62,7 +62,9 @@ public Content.Chunk read(boolean fillInterestIfNeeded)
         boolean last = !byteBuffer.hasRemaining() && data.isLast();
         if (last)
             responseSuccess(getHttpExchange(), null);
-        return Content.Chunk.from(byteBuffer, last, data);
+        Content.Chunk chunk = Content.Chunk.from(byteBuffer, last, data);
+        data.release();
+        return chunk;
```
gregw: Ditto, plus it would remove the need for the local variable:
```suggestion
        return Content.Chunk.fromWithImplicitRetain(byteBuffer, last, data);
```

yeah I know the name is the difficult part!

jetty-core/jetty-http3/jetty-http3-server/src/main/java/org/eclipse/jetty/http3/server/internal/HttpStreamOverHTTP3.java
```diff
@@ -235,9 +235,7 @@ private Content.Chunk createChunk(Stream.Data data)
     {
         if (data == Stream.Data.EOF)
             return Content.Chunk.EOF;
-
-        // As we are passing the ByteBuffer to the Chunk we need to retain.
-        data.retain();
+        // Implicit retain because we are passing the ByteBuffer to the Chunk.
```
gregw: OK, So here we want the retain without release.  Interesting you used the phrase "implicitRetain", but kind of in the opposite sense to my suggestion above.

gregw (commented):
Just some comments on my first look

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/ResponseNotifier.java
lorban: No, `ALREADY_READ_CHUNK` is a sentinel that should only be used for `instanceof` checks. Calling any of its methods is a bug.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Retainable.java
```diff
@@ -26,6 +26,19 @@
...
 {
+    /**
+     * <p>Returns whether this resource can be retained, that is whether {@link #retain()}
+     * can be called safely.</p>
+     * <p>Implementations may decide that special resources are not retainable (for example,
+     * {@code static} constants) so calling {@link #retain()} is not safe because it may throw.</p>
+     *
+     * @return whether it is safe to call {@link #retain()}
+     */
+    default boolean canRetain()
```
lorban: I would not provide a default impl, this is worth explicitly stating in all implementations.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
lorban: I now see that a lot of places the implicit `retain()` that was added has to be compensated by a `release()` and @gregw is proposing a version of `Chunk.from` that doesn't `retain()` to avoid two CAS operations and help readability.

`Chunk.from` used to not retain, why was this changed?

lorban (changes requested):
I'm surprised by the fact that `Chunk.from` now calls `retain()`, and it doesn't help readability and possibly hurts perf.

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/ResponseNotifier.java
gregw: Hmmmm   EOF is also a sentinel instance, but we return true from it's isLast method rather than expecting callers to always do an instanceof check.

I guess if this chunk is never passed externally then OK.

gregw:
@sbordet  ARGH!  why force pushed????  We lose the context of all the conversations so re-review is so much harder!   The PR is going to be squashed anyway, so forced pushes should be avoided!

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -571,61 +574,67 @@ static Chunk next(Chunk chunk)
...
         default Chunk slice()
         {
-            if (isTerminal())
-                return this;
-            if (!hasRemaining())
-                return EMPTY;
-            retain();
-            return from(getByteBuffer().slice(), isLast(), this);
+            if (hasRemaining())
+                return from(getByteBuffer().slice(), isLast(), this);
```
gregw: Shouldn't a retain be done somewhere here? We now have 2 chunks pointing at the same buffer, so 2 releases are now required.   I could not see any retain call in the from method or any of the constructors?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -571,61 +574,67 @@ static Chunk next(Chunk chunk)
...
+            else
+                return isLast() ? this : EMPTY;
```
gregw: why not:
```suggestion
                return isLast() ? EOF : EMPTY;
```

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -571,61 +574,67 @@ static Chunk next(Chunk chunk)
...
+                    return last ? EOF : EMPTY;
+                ByteBuffer sourceBuffer = getByteBuffer();
+                int sourceLimit = sourceBuffer.limit();
+                sourceBuffer.limit(limit);
+                int sourcePosition = sourceBuffer.position();
+                sourceBuffer.position(position);
+                ByteBuffer slice = sourceBuffer.slice();
+                sourceBuffer.limit(sourceLimit);
+                sourceBuffer.position(sourcePosition);
+                return from(slice, last, this);
```
gregw: If I'm  reading this correctly, the source buffer is momentarily having its limit and position changed, then a slice taken, then the pointers reverted.   This is not thread safe as another thread looking at the source buffer will momentarily see the wrong pointers.

Can't we slice (or duplicate) and then adjust the pointers?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -571,61 +574,67 @@ static Chunk next(Chunk chunk)
...
+            }
+            else
+            {
+                return isLast() ? this : EMPTY;
```
gregw:
```suggestion
                return isLast() ? EOF : EMPTY;
```

gregw (commented):
Not a full review... just some questions???

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: No, the retain operation should be (possibly) done by the receiver of the sliced chunk, because it can just consume the sliced chunk synchronously, so there would be no need to retain.
If the consumer of the sliced chunk stores it away for later use, then the consumer is responsible to retain it.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: Because it can also be an instance of `Error`, so returning `this` is correct.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: No we cannot because otherwise we are allowing the code that receives the slice to see bytes past the slice's position and limit.
We should never have 2 threads looking at the same buffer, and I believe the code is already built in a way to guarantee this.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: Nope, `this` can be an `Error` chunk.

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/ResponseNotifier.java
sbordet: Yes it is never passed externally. It is only a marker to indicate that the chunk of one of the copies has been read, so further reads should return `null` until all copies have read the chunk.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: That is a very fragile assumption for a class in util. 

Perhaps we need to duplicate, adjust then slice?

joakime: This is in jetty-io, not jetty-util.  does that change your opinion?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: We don't allow a caller of read to avoid calling release of they consume the chunk synchronously.

I think it is just simpler to say that if you obtain a new chunk then it must be released.

What if the caller releases the original chunk before finishing with the synchronous processing of the new chunk?

Feels fragile.  If there are performance reasons to avoid the retain/release then we probably need a specially named method with very explicit java doc

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: But if this is just an empty normal buffer, then the pointers of the "slice" can be adjusted to see content around it. It will also change the original chunk.

Perhaps Error needs to implement slice

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: Io is a "util" package. Big call to say that it's classes will never be used by multiple threads

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: The change has more a semantic reason, and has performance reasons too.
The places where it is used all consume the slice synchronously, so it felt better to not retain.

The caller should not release the original chunk before consuming the slice, and typically does not because it took a slice of it, meaning that there is something left in the original chunk that needs further processing.

The probability of API usage errors are the same as:
```java
Chunk chunk = read();
chunk.release();
chunk.getByteBuffer().get(0);
```
Yes the API can be used in a way that the chunk is released before being used, but I think it's implicit you should not do that. Same for `slice()`, it's documented that it depends on the original chunk.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: I don't understand? 
We don't return empty Chunks that are not `EMPTY`, which has `capacity==0` so you cannot adjust position/limit.
In all other cases, we slice the `ByteBuffer` from the network buffer, so you can never see around what you're supposed to see.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: > Big call to say that it's classes will never be used by multiple threads

It's not. 
`ByteBuffer`s should not be used by multiple threads and `Chunk` is just a wrapper around a `ByteBuffer`, so inherits the same limitation.
Slices should not allow applications to look around for bytes that possibly don't belong to them, so they must be sliced at the right capacity or we have a security issue. We have been very careful elsewhere in doing that, and here is no different.

gregw: We definitely have examples of ByteBuffers being used by multiple threads.  We have them in the HttpContent cache and multiple threads each slice off a buffer to use.
I don't see why we could not retain a Content.Chunk that we want to send to multiple destinations in multiple threads, each slicing off the chunk they want to send.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: Consider if we have a chunk that we are parsing other chunks out of and slicing them to pass other processing.   That other processing should not be able to adjust the pointers to look at content that was not destined for it.   If the slice contains data, then they cannot do so as the code creates a new chunk from a slice.    But if the original chunk is empty and we slice off it, then we just get the original chunk.   So problem 1 is if we release that, then we are releasing the original. Second problem is that "slice" can have it's pointers adjusted to look at data it should not be able to see, which also modifies the original chunk.

A slice should be a slice and not `this`

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: The sequence I'm concerned about is more like:
```java
    Chunk chunk = read();
    // parse some data
    chunk.position(newPos);
    chunk.limit(newLimit);
    Chunk slice = chunk.slice();
    
    // try to parse more data, but you find none, so ...
    chunk.release();

    // now try to do something with the slice
    slice.get(0); // BOOM! the buffer has already been released because the slice did not retain
```

If slices are consumed synchronously, then why slice at all?
If you really need to do this then have a `sliceWithoutRetain` method, else it is really surprising we allow two chunks to refer to the same buffer, but with a retain counter of only 1!

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: So they are not used by multiple threads. 
They are just sliced by multiple threads, and then each thread uses its own slice.

The reason to not retain inside `slice()` is that `slice()` does not know what use you are doing with the slice, so it cannot just retain() a priori, only to be released 2 lines later -- you complained about this pattern before.

So `slice()` does not retain, and if the caller code wants to send it to multiple consumers, it will retain the slice as many times as it wants. How can `slice()` know how many times you want to retain it because you are sending it to more than one consumer?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: > But if the original chunk is empty and we slice off it, then we just get the original chunk.

No, you get `EMPTY`.

`this` is only returned if it is empty _and_ last, to cover the cases of `Error` and `Trailers` (and it's just that we have to return something, as no code will ever slice an empty last chunk).

A slice of an empty and last chunk is itself, and you cannot move position/limit anyway even in the original chunk, so there is no point in returning anything different from `this`.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: Yours is a wrong sequence, like this one:

```java
Chunk chunk = read();
chunk.release();
chunk.getByteBuffer().get(0); // BOOM!
```

You pointed out in a previous review, when I changed `Chunk.from()` to implicitly retain, that it was wrong because of the silly retain/release cycle, and I agreed it was silly, and it was making `Chunk.from()` a "magic" method because it was, like `read()`, a method that implicitly retains and there was no reason to have 2 "magic" methods.

Here is similar, there is no point to do a silly retain/release cycle and have `slice()` to be a "magic" method.
In this way, there only is one "magic" method: `read()`.

For example, multipart would have to do this in your proposal:

```java
// Implicit retain.
Chunk partContent = chunk.slice(pos, lim, last);

// Give the part content to the application.
onPartContent(partContent);

// Silly release for an implicit retain() 2 lines above.
// The application may have fully consumed 
// the partContent by saving it to disk.
// Plus a performance killer for no reason.
partContent.release();
```

We don't know what the application is doing with the slice, and it's responsibility of the application to retain it if it does not consume it -- not a responsibility of `slice()`.

We document that the slice shares the lifecycle of the original chunk, like we would have to document that it does not and you have to release it if your idea is followed.
So the 2 solutions are equivalent wrt to document their semantic/usage and both rely on documentation and knowing how to use a chunk and therefore write correct code.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: So if it was a slice of an empty last normal chunk then returning this is wrong. Return EOF here and extend Error to return this from slice.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: The point here is not about retain or release. The point is that slice is not thread safe. If each thread calls slice then they will get random results as they may see the pointers passed in by a different thread.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: Maybe don't call it slice, as that implies making an independent chunk. Perhaps call it view, as it makes a dependent view of the original chunk. Then document that a slice is a view + retain

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: > The point is that slice is not thread safe.

It's not supposed to be, never was, and never will.
I don't understand why you are making this argument of thread safety?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: You want to rename `slice()` to `view()` and keep the current semantic?

We do not need the view+retain semantic, so there won't be a "slice" with retain semantic, because it won't be used.

gregw: That will help. The normal slice semantic creates an independent reference to the same buffer. Actions on one cannot affect the other (unless the buffer if written to).

With this current semantic, the method is creating a dependent view to the same buffer. Releasing the original may invalidate that view. So it's not a slice semantic.

If a slice (with retain) is not needed, then ok not to have the method.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: I think this is a bug. But it's a preexisting bug unrelated to this PR. I'll open another issue.

It's definitely plausible that multiple threads may wish to concurrent slice a chunk. It needs to be thread safe.

sbordet: I don't think it is a bug, because `Chunk`s are supposed to be used by 1 thread only.
I don't think we have a use case for Chunks used concurrently, so why do you think it is plausible?

We can make the same argument for `ByteBuffer` or `ArrayList` -- it's plausible to use them concurrently, but we don't make them thread safe because they are not supposed to be used concurrently.

Before the `Chunk` abstraction, in Jetty 11, we are using `ByteBuffer`s (in the client, directly; in the server, wrapped into a `HttpInput.Content`). None of these usages is concurrent, and the design and usage of the `ByteBuffer` never takes thread safety into account (so much so that `ByteBuffer` was used directly), so I'm not sure why now we should?

gregw: We store ByteBuffers in the static content cache and multiple threads can concurrently  slice the same buffer to serve it. The buffer itself is not thread safe, but the slice method must be.

If multiple threads are not being used, then why slice? Just use the original chunk.

It's entirely plausible for a chunk to be retained and used in a similar way, so the slice method needs to be thread safe.

sbordet: I understand that you are proposing two choices:

1. keep a parameter-less `slice()` method similar to `ByteBuffer` and change the name of the `slice(int, int, boolean)` to `view(int, int, boolean)`, so that it's clearer that it is not thread safe; or
2. remove `slice(int, int, boolean)`, do the position/limit modification on the original chunk, and then call the parameter-less `slice()`

Did I get that correctly?
If so, which one you like better?

> If multiple threads are not being used, then why slice? Just use the original chunk.

We cannot because we will expose to applications the whole buffer (possibly with bytes belonging to the next request, or the next frames).

In Jetty 11 we expose the whole buffer up to `HttpInput`, but then applications only have one way of reading from it via `read(byte[])`, so the application could never read before or after the position/limit of the `ByteBuffer`.

In Jetty 12 we have a different API where we may expose the whole buffer directly to applications, so we slice to avoid that, even in the single thread case.

gregw: I've not reviewed in enough detail to say that i have an actual prism that i like. At this stage so i can tell you it's what I don't like.

I don't like a method called slice that produces a dependent chunk that calling release on the original will release the slice, out that releasing the slice will also release the original.

The method you want us not slicing to create a new chunk that can be independently used. From the sound of it, it is temporarily restricting the range of the existing chunk, without creating a new reference, so a synchronous call can be made with that chunk to application coffee that we don't want it peeking at data outside the range.. But that synchronous call somehow knows not to release the chunk it is passed, but also will call retain itself if it is actually asynchronous.

I need to get on a proper computer and look in detail how you are using these methods to really say more (early next week).

I'm short, slice implies creating something independent, which you apparently don't need. You just need something that restricts the view.... but that can be retained if need be.

gregw: I've opened #9141 as this bug is unrelated to this PR.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
```diff
@@ -631,20 +631,13 @@ public Content.Chunk read()
...
-                        state = State.MIDDLE;
-                    }
-                    yield chunk;
+                    if (!chunk.isLast())
+                        yield chunk;
+                    state = State.MIDDLE;
+                    if (chunk.hasRemaining())
+                        yield Content.Chunk.from(chunk.getByteBuffer(), false, chunk);
+                    chunk.release();
+                    yield Content.Chunk.EMPTY;
```
gregw: The from code already does the remaining and EMPTY/EOF checks:
```suggestion
                    yield Content.Chunk.from(chunk.getByteBuffer(), false, chunk);
```

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: I've had a chance to look at the usage more now and I agree that it is OK for the `from` method to not do the retain.   There is nothing about the from method that says it is the original chunk that should be passed as the retainable.    So I'm good with it, but I have some other review comments coming.

gregw (changes requested):

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: as above: return `EOF` here and extend the method in `Error` to return `this`

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: Sorry, I'm unresolving.  It's OK for the from method to not have a retain, but I think the `slice` definitely needs the retain, as we get 2 chunks pointing to the same buffer, but with a reference count of 1.

gregw: Note that the only time this method is called, it does the retain anyway!

gregw: Sorry for the noise.  I will resolve this and make a new comment with less verbosity.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: Also resolving this conversation to be replaced by new review conversation.

gregw: The `from` method already has logic to check for remaining and do EOF/EMPTY return, so we should not duplicate here.   We also should retain here as a slice results in 2 chunks pointing to the same buffer.   The only place this method is currently called from is `ResponseNotifier#onChunk` and it does the retain anyway!

```suggestion
            retain();
            return from(getByteBuffer().slice(), isLast(), this);
```

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/ResponseNotifier.java
```diff
@@ -401,9 +408,17 @@ private void onChunk(Content.Chunk chunk)
                 if (LOG.isDebugEnabled())
                     LOG.debug("Registering content in multiplexed content source #{} that contains {}", index, currentChunk);
                 if (currentChunk == null || currentChunk == ALREADY_READ_CHUNK)
-                    this.chunk = chunk.slice();
+                {
+                    Content.Chunk slice = chunk.slice();
+                    // Retain the slice because it is stored for later reads.
+                    if (slice.canRetain())
+                        slice.retain();
```
gregw: The retain should be done by slice.  If the chunk is not retainable, should it be duplicated here?

gregw (changes requested):
As I look at this PR more, I'm more OK with it.  Specifically I'm OK with the from method not doing a retain.  But I think the slice method should.
I also think we are duplicating the hasRemaining checks too often.

But these are not almost down to niggle status for me. (other than the this return).

gregw:
@sbordet In multipart line 1206 there is existing code that does:
```
                Content.Chunk content = chunk.slice(buffer.position(), sliceLimit, true);
                buffer.position(buffer.position() + boundaryOffset + boundaryFinder.getLength());
                notifyPartContent(content);
                notifyPartEnd();
```

Currently the slice call does not retain, so the content chunk passed to the notifyPartContent method, which call the listener, whose javadoc says it must release the chunk.... but since slice didn't retain, then when the listener releases it will release the original chunk.  

I'm back at my desk tomorrow.  Let's schedule a time to walk through this code together.

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/ResponseNotifier.java
sbordet: The retain should not be done by `slice()`, because `slice()` does not know how the slice is used (whether consumed immediately or stored away for later use).
The retain is done here exactly because the sliced chunk is stored for later use, like the comment says.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: The point of not retaining in `slice()` is that it does not know how the returned slice is used.
Even if the (currently) only call stores the slice away for later use (and so has to retain), in future that may not be true, and we only change the code that actually uses the slice because that is what would have changed.

sbordet:
@gregw you are right that the javadoc was still documenting the old behavior. I fixed the javadoc to document the new behavior.

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/ResponseNotifier.java
gregw: If a chunk is created, then it needs to be released. Methods read, from and slice create a new chunk, so a release should be  needed, so slice needs to retain.   Only if you are passed a chunk as an arg do you not have to release it. Instead you have to retain it if you want to keep it (and then call release).

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -571,61 +574,67 @@ static Chunk next(Chunk chunk)
...
+            }
```
gregw, line 637: This method could just be (from #9141 ):
```suggestion
            {
                ByteBuffer sourceBuffer = getByteBuffer();
                if (position == limit)
                    return last ? EOF : EMPTY;
                return from(sourceBuffer.slice(position, limit - position), last, this);
            }
```

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/ResponseNotifier.java
gregw: @sbordet Even in the case where a slice is used synchronously (i.e Multi-part), I'm cautious about not retaining or even passing a reference to the original retainable.   The chunk is passed to application code, which can be wrong.   We used to require a release, but now we no longer do.  If the application erroneously does release, then they will release the original buffer back into the pool, but the MultiPart parser will keep parsing.      Thus I'm thinking that if we are slicing to get a chunk that we protect from evil code moving pointers where they should not, then we should also protect against evil/bad code going negative with their reference count.  I.e the application should never be able to cause the parser to release the buffer it is currently parsing.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Retainable.java
```diff
@@ -26,6 +26,16 @@
  */
 public interface Retainable
 {
+    /**
+     * <p>Returns whether this resource can be retained, that is whether {@link #retain()}
+     * can be called safely.</p>
+     * <p>Implementations may decide that special resources are not retainable (for example,
+     * {@code static} constants) so calling {@link #retain()} is not safe because it may throw.</p>
```
gregw, line 33: Can we say that calling release on non-retainable Retainable is OK.

gregw (approved):
Other than a niggle and CI failure. OK!

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/transport/internal/HttpReceiverOverHTTP.java
```diff
@@ -450,9 +450,10 @@ public boolean content(ByteBuffer buffer)
         if (chunk != null)
             throw new IllegalStateException("Content generated with unconsumed content left");
 
-        RetainableByteBuffer networkBuffer = this.networkBuffer;
+        // Retain the chunk because it is stored for later use.
```
lorban, line 453:
```diff
--- // Retain the chunk because it is stored for later use.
+++ // Retain the networkBuffer because it is stored as a chunk for later use.
```

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/Response.java
```diff
@@ -190,7 +190,7 @@ default void onContentSource(Response response, Content.Source contentSource)
                 response.abort(error.getCause());
                 return;
             }
-            if (chunk.isTerminal())
+            if (chunk.isLast() && !chunk.hasRemaining())
                 return;
```
lorban, line 194: Missing a `release` before `return`?

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/Response.java
```diff
@@ -190,7 +190,7 @@ default void onContentSource(Response response, Content.Source contentSource)
...
 
             onContent(response, chunk, () -> contentSource.demand(demandCallback));
```
lorban: Missing a `release` after `onContent`?

jetty-core/jetty-http3/jetty-http3-client-transport/src/main/java/org/eclipse/jetty/http3/client/transport/internal/HttpReceiverOverHTTP3.java
```diff
@@ -60,9 +60,10 @@ public Content.Chunk read(boolean fillInterestIfNeeded)
         }
         ByteBuffer byteBuffer = data.getByteBuffer();
         boolean last = !byteBuffer.hasRemaining() && data.isLast();
-        if (last)
-            responseSuccess(getHttpExchange(), null);
-        return Content.Chunk.from(byteBuffer, last, data);
+        if (!last)
+            return Content.Chunk.asChunk(byteBuffer, last, data);
+        responseSuccess(getHttpExchange(), null);
+        return Content.Chunk.EOF;
```
lorban, line 67: Missing `data.release()` before `return`?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
```diff
@@ -53,22 +53,13 @@ public class AsyncContent implements Content.Sink, Content.Source, Closeable
...
-        // possible to create chunks with no remaining byte, when the byte
-        // buffer is empty we need to replace it with EOF / EMPTY and cannot
-        // be notified about the release of the latter two.
-        // This is why read() succeeds the callback if it has no remaining
-        // byte, meaning it is either EOF or EMPTY. The callback is succeeded
-        // once and only once, but that happens either during read() if the
-        // byte buffer is empty or during Chunk.release() if it contains at
-        // least one byte.
         Content.Chunk chunk;
         if (byteBuffer.hasRemaining())
```
lorban, line 64: Isn't `Content.Chunk.from` already doing that `byteBuffer.hasRemaining()` check?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
```diff
@@ -77,53 +68,19 @@ public void write(boolean last, ByteBuffer byteBuffer, Callback callback)
         offer(chunk, callback);
     }
```
lorban, line 69: Technically, the chunk should be released after `offer` returns. We can avoid doing it because `offer` is private and we know it always retains to store the chunk for later and we save a useless retain/release pair.

But that reason probably is worth commenting.

lorban (changes requested):

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
```diff
@@ -233,9 +183,9 @@ public Content.Chunk read()
             }
             readClosed = current.chunk().isLast();
             if (chunks.isEmpty())
-                l.signal();
+                condition.signal();
         }
-        if (!current.chunk().hasRemaining())
+        if (!current.chunk().canRetain())
```
gregw, line 188: This is the only call to `canRetain` that has any substantive behavior associated with it.  All other calls are in the form:
```java
    if (chunk.canRetain())
        chunk.retain();
````
and there is never an else clause or change to the following logic.   So that raises the question: why don't we just make `retain` a noop for chunks that are not cannot be retained?     The answer is that we need to implement this very specific if statement in this and  only this method!!!!

So at the very least, we should put a great big comment here saying how significant this if statement is.

Also the comment at line 73 above still talks about "terminal".   So it needs to be updated and this mechanism better documented.

I think think we need to take one final step back and decide if `canRetain` really is the right API?  Perhaps we just need an `isConstant` that is used only at this location?

gregw (changes requested):

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
gregw: More over, we are only using canRetain to pass information between methods of this class, which already has a queue of records.  We can just put that information (to succeed when read) into the object that we queue.

Standby for a PR to this PR....

gregw: See #9159 that shows t hat `canRetain` is no longer needed

gregw: I've updated #9159 as per our hangout discussion. I'm OK with it being merged here or this being merged first and mine second.

gregw (approved):

lorban (approved):
LGTM

PR #9074, opened 2022-12-19
TCK: Dispatch forward and includes attributes do not meet the spec

janbartel:
Fix several spec non-compliances with `forward` and `include` attribute handling:

* Section 9.3.1 of the spec says the target of a named dispatcher that is an `include` must _not_ have the `jakarta.servlet.include.*` attributes set.
* Section 9.4.2 of the spec says that the target of a named dispatcher `forward` must _not_ have the `jakarta.servlet.forward.*` attributes set.
* Section 9.4.2 of the spec says that a `forward` must report the values for the `jakarta.servlet.forward.*` attributes as the equivalent values from the _original request_  (ie `getRequestURI` `getPathInfo` `getQueryString` etc  etc) that went into the servlet call chain.

Added tests for the above.


janbartel:
@lachlan-roberts need your review on this. Pay particular attention to the changes to solve point 3 (ie reporting the values from the original request) - is there a better solution here?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/Dispatcher.java
```diff
@@ -271,6 +272,18 @@ public String getRequestURI()
...
+
+            //Servlet Spec 9.4.2 must return the values from the original request
+            if (name != null && name.startsWith(__FORWARD_PREFIX))
+            {
+                Object val = super.getAttribute(name);
+                if (val != null)
+                    return val;
+            }
+
             switch (name)
```
lachlan-roberts, line 313: If `name` is null this could throw NPE.
I know this was the case before but now you have the `name != null` check intellij is warning me about it.

Maybe we should have some code like this at the start of these methods
```java
if (name == null)
    return super.getAttribute(name);
```
or 
```java
if (name == null)
    return null;
```

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/Dispatcher.java
```diff
@@ -271,6 +272,18 @@ public String getRequestURI()
...
+            if (_named != null && name != null && name.startsWith(__FORWARD_PREFIX))
+                return null;
+
+            //Servlet Spec 9.4.2 must return the values from the original request
+            if (name != null && name.startsWith(__FORWARD_PREFIX))
+            {
+                Object val = super.getAttribute(name);
+                if (val != null)
+                    return val;
+            }
```
gregw: I'm a bit worried that this approach will get confused if the value is meant to be null (eg a null pathInfo).
So perhaps use the attribute technique to fetch the original request, each `ForwardRequest` asks it's wrapped request for the original request attribute. When it get's null, it knows it is the first forward and returns it's wrapper.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/Dispatcher.java
```diff
@@ -271,6 +272,18 @@ public String getRequestURI()
         @Override
         public Object getAttribute(String name)
         {
+            //Servlet Spec 9.4.2 no forward attributes if a named dispatcher
+            if (_named != null && name != null && name.startsWith(__FORWARD_PREFIX))
+                return null;
+
+            //Servlet Spec 9.4.2 must return the values from the original request
+            if (name != null && name.startsWith(__FORWARD_PREFIX))
```
lachlan-roberts: I don't think this always works.

If we did a forward -> named forward -> forward, then I think this breaks. 

Then the last forward would see the null from the named dispatch wrapper and assume that it is the first forward dispatch.

lachlan-roberts (changes requested):

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/Dispatcher.java
```diff
@@ -55,6 +56,11 @@ public class Dispatcher implements RequestDispatcher
      * Dispatch include attribute names
      */
     public static final String __FORWARD_PREFIX = "jakarta.servlet.forward.";
+    
+    /**
+     * Name of original request attribute
+     */ 
+    public static final String __ORIGINAL_REQUEST = "org.eclipse.jetty.originalRequest";
```
gregw, line 63: maybe put ee10 in the name to protect against future cross context dispatches between ee10, ee11 and then back to ee10.

gregw (approved):

lachlan-roberts (commented):
Where do you set the `__ORIGINAL_REQUEST` attribute?

I would have thought it would be made available by `ServletApiRequest#getAttribute` or something similar.

gregw:
> Where do you set the `__ORIGINAL_REQUEST` attribute?
> 
> I would have thought it would be made available by `ServletApiRequest#getAttribute` or something similar.

@lachlan-roberts 
It is in the code:
```java
                case __ORIGINAL_REQUEST:
                    HttpServletRequest originalRequest = (HttpServletRequest)super.getAttribute(name);
                    return originalRequest == null ? _httpServletRequest : originalRequest;
```

So the inner most forwarded wrapper works out that it is the innermost by getting a null value for the super.getAttribute, thus it returns it's wrapped request as the original.
All outer forward wrappers will see this non null value and pass it up.

lachlan-roberts:
@gregw ok, so its not really the original request but the first forwarded request.

gregw:
> @gregw ok, so its not really the original request but the first forwarded request.

It's the request wrapped by the first forwarded request, so that is either the original request or something that wraps the original request in a non consequential way.

lachlan-roberts (approved):

PR #9075, opened 2022-12-20
Simpler TreeTrie.isEmpty() method

joakime:
Avoid creating (Key)Set to just test for empty.


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/TreeTrie.java
```diff
@@ -276,7 +276,31 @@ private V getBest(Node<V> node, byte[] b, int offset, int len)
...
+                        return true;
+                }
+            }
+            for (int i = t._nextOther.size(); i-- > 0; )
+            {
+                if (hasAnyKey(t._nextOther.get(i)))
+                    return true;
+            }
+        }
+        return false;
```
gregw: probably invert the boolean and rename isEmpty(Node)

gregw (approved):
Other than a rename LGTM

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/TreeTrie.java
joakime: Inverted / Renamed

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/TreeTrie.java
```diff
@@ -276,7 +276,31 @@ private V getBest(Node<V> node, byte[] b, int offset, int len)
...
+    }
+
+    private boolean isEmpty(Node<V> t)
+    {
+        if (t != null)
+        {
+            if (t._key != null)
+                return false;
+
+            for (int i = 0; i < INDEX; i++)
```
lorban, line 289: Is there some test coverage for this logic?

lorban (approved):
This looks alright, but I think we should add some testing to cover the two loops in `isEmpty(Node)`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/TreeTrie.java
joakime: Yup, the first loop is 100% covered by the 2 new test cases.
The second loop executes, but nothing can trigger the internal isEmpty() logic with the test cases as it stands.

gregw:
I just pushed 3d4bb492d0c7ffff350ad627fbf3a33382c999d3 which slightly improves the test coverage

PR #9077, opened 2022-12-21
Jetty 12.0.x h2c perf improvement: implement equals/hashcode in Stream

lorban:
Implementing equals/hashcode in `Stream` instead of relying on `System.identityHashCode` is worth a couple % of performance.


gregw (approved):
Looks fine

sbordet (approved):

PR #9079, opened 2022-12-22
Fix `getBuffer()` on cached HttpContent

### LINKED ISSUE #9078: Header image for Jetty demo page fails on hard refresh
**Jetty version(s)**
_Jetty 12_

**Java version/vendor** 
`openjdk version "19.0.1" 2022-10-18`

**OS type/version**
macOS Monterey 12.6.2

**Description**
When running the `demos` module that ships with Jetty 12, I am seeing some odd behavior upon accessing the demo home page at http://0.0.0.0:8080/.

When the page first loads, everything looks fine. But if I do a hard refresh (Cmd+Shift+R), the header image no longer appears and I get the following stacktrace in my terminal:

```
2022-12-21 15:32:43.501:WARN :oejs.Response:qtp640363654-29: writeError: status=500, message=java.io.IOException: written 0 < 493852 content-length, response=org.eclipse.jetty.server.internal.HttpChannelState$ErrorResponse@7db7a010
java.io.IOException: written 0 < 493852 content-length
	at org.eclipse.jetty.server.internal.HttpChannelState$ChannelResponse.write(HttpChannelState.java:1111)
	at org.eclipse.jetty.ee10.servlet.HttpOutput.channelWrite(HttpOutput.java:213)
	at org.eclipse.jetty.ee10.servlet.HttpOutput.complete(HttpOutput.java:443)
	at org.eclipse.jetty.ee10.servlet.ServletContextResponse.completeOutput(ServletContextResponse.java:152)
	at org.eclipse.jetty.ee10.servlet.ServletChannel.handle(ServletChannel.java:597)
	at org.eclipse.jetty.ee10.servlet.ServletHandler.process(ServletHandler.java:442)
	at org.eclipse.jetty.ee10.servlet.security.SecurityHandler.process(SecurityHandler.java:563)
	at org.eclipse.jetty.ee10.servlet.SessionHandler.process(SessionHandler.java:756)
	at org.eclipse.jetty.server.handler.ContextHandler.process(ContextHandler.java:696)
	at org.eclipse.jetty.server.handler.ContextHandlerCollection.process(ContextHandlerCollection.java:166)
	at org.eclipse.jetty.server.Handler$Collection.process(Handler.java:577)
	at org.eclipse.jetty.server.Handler$Wrapper.process(Handler.java:541)
	at org.eclipse.jetty.rewrite.handler.RewriteHandler.process(RewriteHandler.java:118)
	at org.eclipse.jetty.server.Handler$Wrapper.process(Handler.java:541)
	at org.eclipse.jetty.server.internal.HttpChannelState$HandlerInvoker.run(HttpChannelState.java:594)
	at org.eclipse.jetty.server.internal.HttpConnection.onFillable(HttpConnection.java:445)
	at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:322)
	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:100)
	at org.eclipse.jetty.io.SelectableChannelEndPoint$1.run(SelectableChannelEndPoint.java:53)
	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:934)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1078)
	at java.base/java.lang.Thread.run(Thread.java:1589)
```

The fact that the page loads correctly at first and only breaks upon refresh is what seems particularly weird to me here.

**How to reproduce?**
`
$ java -jar $JETTY_HOME/start.jar --add-modules=demos
$ java -jar $JETTY_HOME/start.jar
`
Then navigate to http://0.0.0.0:8080/. The page loads correctly upon first load, but if I do a hard refresh (Cmd+Shift+R) the image no longer appears.

lorban:
Fixes #9078.

The bug does not happen when the served resource is deemed to small to be mmap'ed, so `HttpContent.getBuffer()` returns null and the slow-path serving is used.


gregw (approved):
This fix itself looks OK, but a bit hard to tell as there is absolutely no javadoc on the whole `HttpContent` contract.

So let's get this fix merged, but please work on completing the resource refactor, including documenting it.  

To me, it is entirely unclear if the content should duplicate or if the caller of `HttpContent.getByteBuffer()` should duplicate.    Perhaps the name should not be a getter if it is always making a new `ByteBuffer`?   But those are issues for another PR... or an entirely different HttpContent approach (@lachlan-roberts :)

gregw:
I've opened #9080 as HttpContent needs to be documented

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
```diff
@@ -376,7 +376,7 @@ public long getContentLengthValue()
         @Override
         public ByteBuffer getByteBuffer()
         {
-            return _buffer;
+            return _buffer == null ? null : _buffer.duplicate();
```
gregw:
```suggestion
            return _buffer == null ? null : _buffer.asReadOnlyBuffer();
```

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappingHttpContentFactory.java
```diff
@@ -86,13 +86,13 @@ public ByteBuffer getByteBuffer()
...
             if (buffer != null)
-                return (buffer == SENTINEL_BUFFER) ? super.getByteBuffer() : buffer;
+                return (buffer == SENTINEL_BUFFER) ? super.getByteBuffer() : buffer.duplicate();
 
             try (AutoLock lock = _lock.lock())
             {
                 if (_buffer == null)
                     _buffer = getMappedByteBuffer();
-                return (_buffer == SENTINEL_BUFFER) ? super.getByteBuffer() : _buffer;
+                return (_buffer == SENTINEL_BUFFER) ? super.getByteBuffer() : _buffer.duplicate();
```
gregw: ditto

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappingHttpContentFactory.java
```diff
@@ -86,13 +86,13 @@ public ByteBuffer getByteBuffer()
         {
             ByteBuffer buffer = _buffer;
             if (buffer != null)
-                return (buffer == SENTINEL_BUFFER) ? super.getByteBuffer() : buffer;
+                return (buffer == SENTINEL_BUFFER) ? super.getByteBuffer() : buffer.duplicate();
```
gregw: ditto

gregw (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
lorban: Done.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappingHttpContentFactory.java
lorban: Done.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/FileMappingHttpContentFactory.java
lorban: Done.

lorban:
@lachlan-roberts we need to talk about this to make sure we're aligned on where we currently stand and where we want to go.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
```diff
@@ -376,7 +376,7 @@ public long getContentLengthValue()
         @Override
         public ByteBuffer getByteBuffer()
         {
-            return _buffer;
+            return _buffer == null ? null : _buffer.asReadOnlyBuffer();
```
lachlan-roberts: I don't think this should be done here, it should be done in the constructor as `_buffer` is final.

lachlan-roberts (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CachingHttpContentFactory.java
lorban: Have a look at the test I added. The bug is that the buffer returned by this method gets consumed by the write(). So someone has to duplicate the buffer: either here, or just before write(), where I added an extra comment.

I decided here arbitrarily to do it in `getBuffer()`, which is why @gregw pointed out that maybe that method shouldn't be called getBuffer but something like newBuffer or createBuffer. And that this bug happened because this contract isn't documented.

gregw: I remain 99% convinced that the whole `HttpContent.getByteBuffer` approach is fundamentally flawed.  The implementations need to know why they are being called before they can know what type of buffer they should return.  If the `HttpContent` is being cached, then the buffer needs to be duplicated, but if it is not being cached, then a duplication is wasted effort on a buffer that is being thrown away anyway.

This bug/PR has only increased that conviction.   This is something we really need to resolve early in the new year and I think we need to reconsider @lachlan-roberts work on this.

But for now, we need a fix so we can serve content more than once.     Either the buffer needs to be duplicated/sliced within getByteBuffer implementations; or the caller needs to duplicate/slice before it does anything to modify the psoition/limit.  Pick one approach, document it, implement it.   Get this working and then we review.

gregw (approved):

gregw:
@lachlan-roberts can you re-review.  Let's get a fix committed as soon as we can and work in #9080 to strive for a better solution.

lachlan-roberts (approved):

PR #9093, opened 2022-12-27
jetty 12.0.x cleanup duplicate osgi pom metadata

olamy:
- fix naming convention, artifactId == directtory, add comment on maven-eclipse-plugin
- cleanup dependencies from jetty-ee9 osgi
- cleanup dependencies from jetty-ee10 osgi

### DIFF
--- a/jetty-ee10/jetty-ee10-osgi/pom.xml
+++ b/jetty-ee10/jetty-ee10-osgi/pom.xml
@@ -67,22 +67,6 @@
         </excludes>
       </resource>
     </resources>
-    <plugins>
-      <!-- is it really used?? this plugiin has been abandonned and no version since 2012... -->
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-eclipse-plugin</artifactId>
-        <version>2.9</version>
-        <configuration>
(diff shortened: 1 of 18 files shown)


jetty-ee10/jetty-ee10-osgi/pom.xml
```diff
@@ -67,9 +68,11 @@
       </resource>
     </resources>
     <plugins>
+      <!-- is it really used?? this plugiin has been abandonned and no version since 2012... -->
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-eclipse-plugin</artifactId>
+        <version>2.9</version>
```
olamy: @janbartel do you really use this plugin?

joakime (approved):

jetty-ee10/jetty-ee10-osgi/pom.xml
olamy: @janbartel ping? maybe we can remove those lines?

janbartel: @olamy nope, I don't use that plugin at all, so go ahead and remove.

janbartel (approved):

PR #9118, opened 2023-01-02
Removing `@Disabled` from jetty-client `Usage`

joakime:


sbordet (changes requested):
@joakime please drop this PR.
The `Usage` class is just meant to show the usage of the APIs, not running real tests.
We can probably drop the entire class since now we have documentation code snippets that replace this.
Besides, we can't start a server on port 8080, it'll clash on CI, etc.

joakime:
@sbordet the existing asserts make it look like a unit test.

I'll close this PR and just remove that Usage class entirely

PR #9123, opened 2023-01-03
Wrong value of `RequestDispatcher.FORWARD_CONTEXT_PATH` attribute on root context

### LINKED ISSUE #9119: Wrong value of javax.servlet.forward.context_path attribute
**Jetty version(s)**
10.0.13

11.0.7

**OS type/version**
Linux/Fedora 37

**Description**
I work on Pax Web project and it unifies 3 embedded web servers (Jetty, Tomcat and Undertow) in OSGi runtime.
I try hard to make the tests consistent and ensure similar behavior between the runtimes.

According to Chapter _9.4.2 Forwarded Request Parameters_ of the Servlets 4 specification:
> The following request attributes must be set:
javax.servlet.forward.mapping
javax.servlet.forward.request_uri
javax.servlet.forward.context_path
javax.servlet.forward.servlet_path
javax.servlet.forward.path_info
javax.servlet.forward.query_string
The values of these attributes must be equal to the return values of the
HttpServletRequest methods getRequestURI, getContextPath, getServletPath,
getPathInfo, getQueryString respectively, invoked on the request object passed to
the first servlet object in the call chain that received the request from the client.

When using Jetty 9.4.50, I see correct (in [org.eclipse.jetty.server.Dispatcher#forward()](https://github.com/eclipse/jetty.project/blob/jetty-9.4.50.v20221201/jetty-server/src/main/java/org/eclipse/jetty/server/Dispatcher.java#L153)):
```java
final String old_context_path = baseRequest.getContextPath();
...
ForwardAttributes attr = new ForwardAttributes(old_attr);
...
attr._contextPath = old_context_path;
```

However in Jetty 10, it's:
```java
if (old_attr.getAttribute(FORWARD_REQUEST_URI) == null)
    baseRequest.setAttributes(new ForwardAttributes(old_attr,
        old_uri.getPath(),
        old_context == null ? null : old_context.getContextHandler().getContextPathEncoded(),
        baseRequest.getPathInContext(),
        source_mapping,
        old_uri.getQuery()));
```

And the forwarded context path is `"/"` instead of `""` (empty string).

**How to reproduce?**

I have a `"/gateway/*"` mapped servlet with:
```java
protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
	switch (what) {
		case "redirect":
			resp.sendRedirect(where);
			return;
		case "forward":
			return;
		case "include":
			return;
		default:
	}
}
```

And I call it with this URI:
```
"/gateway/x?what=forward&where=/"
```

The test is https://github.com/ops4j/org.ops4j.pax.web/blob/web-9.0.4/pax-web-jetty/src/test/java/org/ops4j/pax/web/service/jetty/internal/UnifiedJettyTest.java#L505-L516

joakime:
* Fixes #9119 - uses proper context path that satisfies the root context rules of the servlet spec

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Dispatcher.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Dispatcher.java
@@ -178,7 +178,7 @@ public class Dispatcher implements RequestDispatcher
                 if (old_attr.getAttribute(FORWARD_REQUEST_URI) == null)
                     baseRequest.setAttributes(new ForwardAttributes(old_attr,
                         old_uri.getPath(),
-                        old_context == null ? null : old_context.getContextHandler().getRequestContextPath(),
+                        baseRequest.getContextPath(),
                         baseRequest.getPathInContext(),
                         source_mapping,
                         old_uri.getQuery()));
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Dispatcher.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Dispatcher.java
@@ -178,7 +178,7 @@ public class Dispatcher implements RequestDispatcher
                 if (old_attr.getAttribute(FORWARD_REQUEST_URI) == null)
                     baseRequest.setAttributes(new ForwardAttributes(old_attr,
                         old_uri.getPath(),
-                        old_context == null ? null : old_context.getContextHandler().getContextPathEncoded(),
+                        old_context == null ? null : old_context.getContextHandler().getRequestContextPath(),
                         baseRequest.getPathInContext(),
                         source_mapping,
                         old_uri.getQuery()));
--- a/jetty-servlet/src/test/java/org/eclipse/jetty/servlet/DispatcherTest.java
+++ b/jetty-servlet/src/test/java/org/eclipse/jetty/servlet/DispatcherTest.java
@@ -13,13 +13,21 @@
 
 package org.eclipse.jetty.servlet;
 
+import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.PrintWriter;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.List;
+import java.util.Objects;
+import java.util.Properties;
+import java.util.stream.Collectors;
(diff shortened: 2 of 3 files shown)


jetty-servlet/src/test/java/org/eclipse/jetty/servlet/DispatcherTest.java
```diff
@@ -878,35 +1003,56 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) t
         }
     }
 
-    public static class AssertForwardServlet extends HttpServlet implements Servlet
+    public static class DumpForwardServlet extends HttpServlet
```
joakime, line 1006: I made this generic, so it can be used from different test case setups.
It's a better idea to assert values in the same thread the test is running in anyway.

grgrzybek (dismissed):

sbordet (dismissed):

jetty-server/src/main/java/org/eclipse/jetty/server/Dispatcher.java
```diff
@@ -178,7 +178,7 @@ protected void forward(ServletRequest request, ServletResponse response, Dispatc
                 if (old_attr.getAttribute(FORWARD_REQUEST_URI) == null)
                     baseRequest.setAttributes(new ForwardAttributes(old_attr,
                         old_uri.getPath(),
-                        old_context == null ? null : old_context.getContextHandler().getContextPathEncoded(),
+                        old_context == null ? null : old_context.getContextHandler().getRequestContextPath(),
```
gregw: I think this can just be 
```suggestion
                        baseRequest.getContextPath(),
```

which has some logic in it about cross context include dispatches.

gregw (changes requested):
I'm trying to work out why the code was written that way in the first place?  I can't see why as the request getContextPath also returns the encoded value???

Might be worthwhile to add a test of doing a forward from a cross context include.

@olamy can we run the TCK against this branch before merging?

jetty-server/src/main/java/org/eclipse/jetty/server/Dispatcher.java
joakime: Fixed.

gregw (approved):

gregw:
I've approved changes, but let's keep issue open until TCK tests are double checked.

joakime:
@olamy can you test this branch with the TCK?

olamy:
all good regarding TCK build here https://jenkins.webtide.net/job/tck/job/servlettck-run-jetty-10.0.x/2044/
please note 2 failing tests are due to outdated certificate included in the TCK distribution.

PR #9133, opened 2023-01-07
jetty 12: ee10: DefaultServlet should initialize ResourceFactory unconditionally

kohlschuetter:
Currently, unlike ee9 and ee8, ee10's DefaultServlet initializes _resourceFactory  only when either "baseResource" or "resourceBase" are set as init parameters.

In embedded jetty, the resource base can also be set using ContextHandler#setBaseResource(Resource), which means that DefaultServlet would not initialize correctly and therefore prevent operations like overriding the jetty-dir.css stylesheet by throwing an otherwise unexplicable NullPointerException.

Change ee10 DefaultServlet to always initialize ResourceFactory, like the ee9 and ee8 variants do.


lorban (approved):
LGTM

Issue #9116, opened 2023-01-01
"resource:" URL scheme not supported (needed for GraalVM)


kohlschuetter:
**Jetty version(s)**
12.0.0.alpha3

**Java version/vendor** `(use: java -version)`
openjdk version "19.0.1" 2022-10-18
OpenJDK Runtime Environment GraalVM CE 22.3.0 (build 19.0.1+10-jvmci-22.3-b08)
OpenJDK 64-Bit Server VM GraalVM CE 22.3.0 (build 19.0.1+10-jvmci-22.3-b08, mixed mode, sharing)

**OS type/version**
macOS

**Description**
When using Resources from the resource class path, GraalVM turns them into URLs of the abstract "resource:" scheme (e.g., resource:/com/example/server/webapp/). This resource scheme appears to be currently unsupported by Jetty:

```
java.lang.IllegalArgumentException: URI scheme not supported: resource:/com/example/server/webapp/
	at org.eclipse.jetty.util.resource.ResourceFactoryInternals$CompositeResourceFactory.newResource(ResourceFactoryInternals.java:145)
	at org.eclipse.jetty.util.resource.ResourceFactory.newResource(ResourceFactory.java:229)
```
**How to reproduce?**

Place a file named "test.txt" in the resource class path relative to some class (e.g., `foo.bar.SomeClass` -> `src/main/resources/foo/bar/test.txt`). Build GraalVM native image.

Compare the output of
```
Resource res = ResourceFactory.root().newResource(SomeClass.class.getResource("test.txt"));
System.out.println(res);    
```
between Hotspot and native-image.

With Hotspot, the URL will be something like `jar:file:///path/to/classpath.jar!/foo/bar/test.txt` or `file:///path/to/resource/foo/bar/test.txt`).

With GraalVM native-image, it's `resource:/foo/bar/test.txt`.

Note that this also means that tricks to enumerate/list the contents of a resource directory (by traversing the file hierarchy or jar archive) will fail.

joakime:
The `Resource` layer in Jetty 12 is based on `java.nio.file.Path` and `java.nio.file.FileSystem`.
This was desired as the old school URL Protocol Handling is now deprecated in newer versions of Java.
So are **all** of the constructors on the `URL` class.

We have a fallback `Resource` implementation for URL support, but you have to register your scheme.
Example:

``` java
ResourceFactory.registerResourceFactory("resource", new URLResourceFactory());
```

[...]

PR #9136, opened 2023-01-07
util: Add support for GraalVM Native-Image resource:-URIs and Paths

kohlschuetter:
GraalVM Native-Image makes its classpath resources accessible through an opaque resource: URL scheme.

Add support for this scheme in URIUtil and PathResource.

Automatically create the NativeImageResourceFileSystem when necessary. Also add a workaround where converting such Native-Image Paths back to URI would yield the wrong URI (potentially a bug in GraalVM).

https://github.com/eclipse/jetty.project/issues/9116 https://github.com/oracle/graal/issues/5720


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -54,6 +57,7 @@
         .with("file:")
         .with("jrt:")
         .with("jar:")
+        .with("resource:")
```
joakime: Nope, this should not be part of the default schemes.
This scheme has dangerous side effects outside of GraalVM and should be only be allowed to be configured, not always on.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -48,6 +48,7 @@ public class PathResource extends Resource
         .caseSensitive(false)
         .with("file")
         .with("jrt")
+        .with("resource")
```
joakime: Again, this is not safe for the majority of users.

joakime (changes requested):
Having `resource` scheme be enabled by default is unsafe for the majority of users (esp those outside of GraalVM).

Why can't this exist as it's own ResourceFactory?

joakime:
This PR breaks OSGI users. (which also register `resource` scheme but for different purpose and techniques)
This fact alone makes this PR not useful in it's current state.

joakime:
This default list of schemes is carefully selected.

We support:

```
file:
jar:
jrt:
```

The rest are problematic on different runtimes and must be configured on a case by case basis, not defaulted.

This is the problematic list:

```
ftp:
http:
https:
jmod:
mailto:
war:
bundle:
bundleentry:
bundle-entry:
bundleresource:
bundle-resource:
entry:
resource:
reference:
fake:
classpath:
classloader:
loader:
link:
mvn:
maven:
wrap:
activator:
bytes:
nested:
bndloader:
```

These should be provided by a unique ResourceFactory that is registered via

``` java
ResourceFactory.registerResourceFactory(String scheme, ResourceFactory factory);
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
kohlschuetter: Is there another environment where this scheme is actually available?
What specific dangerous side effects do you have in mind?

I think `resource:` URLs/URIs are exclusive to GraalVM native-image. On a regular hotspot VM, I simply get an error when `resource:` URLs are requested.

Besides that, URIUtil.KNOWN_SCHEMES is actually not used and private.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -141,15 +142,16 @@ public static boolean isSameName(Path pathA, Path pathB)
      * Must be an absolute URI using the <code>file</code> scheme.
      *
      * @param uri the URI to build this PathResource from.
+     * @throws IOException if the path could not be resolved.
      */
-    PathResource(URI uri)
+    PathResource(URI uri) throws IOException
```
joakime: Nope, IOException should not be thrown here.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
```diff
@@ -52,6 +52,7 @@
         PathResourceFactory pathResourceFactory = new PathResourceFactory();
         RESOURCE_FACTORIES.put("file", pathResourceFactory);
         RESOURCE_FACTORIES.put("jrt", pathResourceFactory);
+        RESOURCE_FACTORIES.put("resource", pathResourceFactory); // GraalVM native-image resource
```
joakime: This shall not be automatic or defaulted.
The modifications you have made to `URIUtil` and `PathResource` should be refactored out into a standalone `GraalVMResourceFactory` implementation, not by modifying the default behaviors.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
kohlschuetter: URIUtil.KNOWN_SCHEMES is actually not used, and private. Can we remove the entire declaration?

joakime: if it's unused, i don't see why not.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
kohlschuetter: I've thought about that, but then GraalVM-unaware code that simply tries to serve resources from the classpath will fail.

I've amended my change to transparently support the following real-world use case:

    URL resourceURL = MyClass.class.getResource("some/resource");
    Resource resource = ResourceFactory.root().newResource(resourceURL);

The new change only enables "resource:" URLs when a well-known GraalvM Native Image system property is set.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResource.java
```diff
@@ -44,11 +44,20 @@ public class PathResource extends Resource
 {
     private static final Logger LOG = LoggerFactory.getLogger(PathResource.class);
 
-    public static Index<String> SUPPORTED_SCHEMES = new Index.Builder<String>()
-        .caseSensitive(false)
-        .with("file")
-        .with("jrt")
-        .build();
+    public static Index<String> SUPPORTED_SCHEMES;
```
kohlschuetter: Is there actually any point in exposing SUPPORTED_SCHEMES as a public, mutable member? (This is existing code, not altered by this change)

gregw:
@joakime I'm still not convinced that it is "dangerous" for us to support arbitrary URL schemes? We've done so up until jetty 11 without issue.   I'm not totally opposed to having to register exotic schemes, but then I'm also not opposed to automatic discovery either. Let's schedule a meeting this week where you can outline exactly why you are advocating a new policy of no automatic scheme discovery 

@kohlschuetter it's interesting that this scheme can be implemented with a PathResource. Would i be correct in assuming that in previous releases, the scheme was sorted by a discovered URL handler?  If so, if we only had generic UrlResource handling enabled, would that work for Graal? Is there a performance penalty of going via URL rather than Path?

@joakime i think we also need to discuss schemes that are supported by Path that we have not considered.  Let's chat mid week.

gregw:
Note also, that if we can't support Graal out of the box (preferred outcome), then we should be able to have a module that simply adds/configures support.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResourceFactory.java
```diff
@@ -13,19 +13,63 @@
...
+import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
+import java.util.Collections;
 
 public class PathResourceFactory implements ResourceFactory
 {
+    static final boolean ENABLE_GRAALVM_RESOURCE_SCHEME = (System.getProperty(
+        "org.graalvm.nativeimage.kind") != null);
```
gregw: We really don't like configuration by System property.  Yes I know that we do it elsewhere in the code, but we are trying to remove them.

Either we should always attempt discovery of unknown schemes OR we should provide a simple way to add `ResourceFactory`s if needed.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResourceFactory.java
```diff
@@ -13,19 +13,63 @@
...
+                }
+                catch (FileSystemNotFoundException e)
+                {
+                    FileSystems.newFileSystem(uri, Collections.emptyMap());
+                }
+            }
+            catch (IOException | URISyntaxException | RuntimeException ignore)
+            {
+              // ignore
+            }
```
gregw: I don't like such specific initializer code in a fairly generic class.   This would probably be better put in the constructor of a `ResourceFactory` that is installed if needed.  This would also allow parameters to be set and passed in if needed.

Note that the style of initialization could probably be generalized so that any JVM supported filesystem that needs to be created can be.  i.e. it does not need to have "resource" hard coded.

gregw (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
gregw: We can go one of two ways:
1)  automatic discovery of unknown schemes.  We have always done this for URL handlers, but now it looks like we'd also need to do an attempt to to look for a filesystem for a Path based unknown scheme provided by the JVM.  That makes me less inclined to go automatic.
2) A simple registration mechanism, that can easily be put into a module.  Potentially fairly generic so it can add either URL handlers or FileSystem mounters for an arbitrary resource.   A graalvm module would then do the mapping for "resource" to the FileSystem mounting extension of PathResourceFactory.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResourceFactory.java
joakime: Where does this System Property come from?
Is it something that is always set by a GraalVM?

joakime: Instead of a System Property, is there a class/resource that can be requested that will return a URL with scheme `resource:` that will work on all GraalVM runtimes?

That way startup could look for it, and auto-register a unique `GraalVMResourceFactory`, if the returned URL is scheme `resource`.

kohlschuetter: The system property is set for all GraalVM Native-Image environments (i.e., binary executable or shared library). Since it's a property, it may also be set for testing purposes in a non-native (HotSpot) environment. The code is written in a way that should not break anything in that case (notwithstanding the potential OSGI issue with the resource:/ scheme that Oracle probably has to figure out somehow)

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResourceFactory.java
kohlschuetter: We might be able to move that initializer to all to a wrapper for `Path.of`/`Paths.get` that always tries to initialize the filesystem for the given URI upon `FileSystemNotFoundException`.

One minor issue is that I haven't found a way to convince Github's CodeQL checker to not throw crazy errors at me when using a "user-supplied" parameter to `Path.of`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResourceFactory.java
sbordet: This is quite invasive, as we put another project's system property into common Jetty code. 
I would prefer a Graal-specific ResourceFactory, along with anything else that Graal requires, isolated into a Jetty module as @gregw suggests.

kohlschuetter:
@gregw As I got sucked down the "path" of adding support for resources in GraalVM native-image, I feel adding a custom GraalVM ResourceFactory only makes sense if it's enabled by default in that environment. After all, this eventually will just be "another JVM", not something that the Jetty user has much control over.

While accessing these resources almost always starts with a URL (`URL u = SomeClass.class.getResource("file")`), we need to be able to enumerate/list directory contents, which is not possible with the `java.net.URL` API. I therefore haven't had much success with earlier Jetty versions.

I can try moving most code into a `NativeImagePathResourceFactory` subclass (named after the `NativeImageResourceFileSystem` since it's not really GraalVM at fault here). However `ResourceFactoryInternals` should still automatically enable that `resource` scheme for Native-Image environments.

To add some context, I'm working on a project that builds around an embedded Jetty server. With this patch set, I can run a full http/servlet environment as a single executable binary in less than 100 MB, including content and all. This makes Jetty an ideal candidate to run in a constrained virtual environment like Firecracker using Linux or even OSv.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/PathResourceFactory.java
kohlschuetter: I've added this check upon request because @joakime mentioned that OSGI would use `resource:` for other purposes. I don't know exactly how always registering `resource:` would be a problem though. It's of equivalent standard as `jrt`, which also isn't always available.

Moreover, checking for the availability of classes in Native-Image comes at the cost of adding support for reflection, which needs additional configuration. I'd be happy to remove that check. @gregw mentioned earlier that there could be an interest to support arbitrary schemes / add discovery of new ones.

joakime:
Looking back at Issue #9116, this started as support for `resource:` on a URL (with a GraalVM specific `java.net.URLStreamHandler` support for `resource:` protocol/scheme).

It looks like GraalVM now supports `resource` scheme as a `java.nio.files.Path` (via a custom `java.nio.files.FileSystem` implementation I take it).

I'd like to experiment a bit with this new reality.

Lets use this project -> https://github.com/joakime/graalvm-experiments  (I've added you as a collaborator if you want to muck with it there)

How can I run this demo project with GraalVM that has `resource:/path/to/blah` enabled for `Path`?
I would expect that the URLs returned in that demo application return a path that is scheme `resource:/` based.

joakime:
I managed to get graalvm `native-image` support working on the demo project at https://github.com/joakime/graalvm-experiments

If we look at the [`Demo.java`](https://github.com/joakime/graalvm-experiments/blob/main/src/main/java/graaldemo/Demo.java) and see what happens on a graalvm `native-image` result we see ...

```
$ ./graaldemo.demo
-- Looking for: org/eclipse/jetty/version/build.properties
FOUND: resource:/org/eclipse/jetty/version/build.properties
java.nio.file.FileSystemNotFoundException: Unable to open URI directly, switching to mount version.
Mounting: resource:/org/eclipse/jetty/version/build.properties
-- root of mount
Path: /
Path.toURI: resource:file:///resources!/
Path.getFileSystem.className: (com.oracle.svm.core.jdk.resources.NativeImageResourceFileSystem)
Files.isDirectory(path): true
Files.isRegularFile(path): false
Files.isReadable(path): true
Files.exists(path): true
Files.size(path): 0
Files.getLastModifiedTime(path): 2023-01-09T17:32:01.829Z
-- Specific URI
Path: /org/eclipse/jetty/version/build.properties
Path.toURI: resource:file:///resources!/org/eclipse/jetty/version/build.properties
Path.getFileSystem.className: (com.oracle.svm.core.jdk.resources.NativeImageResourceFileSystem)
Files.isDirectory(path): false
Files.isRegularFile(path): true
Files.isReadable(path): true
Files.exists(path): true
Files.size(path): 163
Files.getLastModifiedTime(path): 2023-01-09T17:32:01.829Z
-- Looking for: PathResource.class
WARNING: Unable to find PathResource.class
```

This means `resource:` is built-in when running a graalvm `native-image`.
It's provided by the `FileSystem` implementation at `com.oracle.svm.core.jdk.resources.NativeImageResourceFileSystem`
However, the resulting `Path` objects have a bug in their `Path.toURI()` results.

Example of the bug ...

``` java
Path path = Path.of(URI.create("resource:/org/eclipse/jetty/version/build.properties"));
path.toURI(); // is "resource:file:///resources!/org/eclipse/jetty/version/build.properties"
```

This is not a symmetrical Path URI input to output.  ugh.

Unfortunately, the resulting output URI is also not valid in itself.
If you feed this back into Java and attempt to open a Path for it, you get ...

```
-- Specific URI: resource:file:///resources!/org/eclipse/jetty/version/build.properties
Path: file:/resources!/org/eclipse/jetty/version/build.properties
Path.toURI: resource:file:///resources!/file:/resources!/org/eclipse/jetty/version/build.properties
Path.getFileSystem.className: (com.oracle.svm.core.jdk.resources.NativeImageResourceFileSystem)
Files.isDirectory(path): false
Files.isRegularFile(path): false
Files.isReadable(path): false
Files.exists(path): false
Files.size(path): java.nio.file.NoSuchFileException: file:/resources!/org/eclipse/jetty/version/build.properties
Files.getLastModifiedTime(path): java.nio.file.NoSuchFileException: file:/resources!/org/eclipse/jetty/version/build.properties
```

Which can be mounted as a java `FileSystem`, but the overall URI is broken and results in `FileNotFoundException`, and we also get a nested URI output that's way mangled.
`Path.toURI: resource:file:///resources!/file:/resources!/org/eclipse/jetty/version/build.properties`

But what this has revealed is that I think we could easily make a lookup of something in Jetty (I used `ClassLoader.getResource("org/eclipse/jetty/version/build.properties")` above)
The resulting URL/URI can tell us what schemes are available as a `FileSystem` and need to be supported (at least).

kohlschuetter:
Yes, that's the situation, and this is why I was dancing around `PathResource` and fixing URIs all along. I've reported the bug upstream to https://github.com/oracle/graal/issues/5720, so expect it to be fixed some time in 2035.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/NativeImagePathResource.java
```diff
@@ -0,0 +1,80 @@
...
+        return "resource".equals(uri.getScheme());
+    }
+
+    /**
+     * Corrects any bad {@code resource} based URIs, such as those starting with {@code resource:file:///resources!}.
+     * 
+     * @param uri
+     *            The URI to correct.
+     * @return the corrected URI, or the original URI.
+     * @see <a href="https://github.com/oracle/graal/issues/5720">Graal issue 5720</a>
```
joakime: Thanks for linking to the bug here.  It will help in the future.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/NativeImagePathResourceFactory.java
```diff
@@ -0,0 +1,71 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Collections;
+
+/**
+ * GraalVM Native-Image {@link PathResourceFactory}.
+ */
+public class NativeImagePathResourceFactory extends PathResourceFactory
+{
+    static final boolean ENABLE_GRAALVM_RESOURCE_SCHEME = (System.getProperty("org.graalvm.nativeimage.kind") != null);
```
joakime: Alternatively we could test a `Classloader.getResource(reference)` for scheme `resource` here too (instead of relying on a System property)

joakime (commented):
This is much better.

I think we are down to just naming nits now.
@gregw WDYT?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/NativeImagePathResource.java
```diff
@@ -0,0 +1,80 @@
...
+package org.eclipse.jetty.util.resource;
+
+import java.net.URI;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * GraalVM Native-Image {@link Path} Resource.
+ */
+public class NativeImagePathResource extends PathResource
```
joakime: Can this include the name `GraalVM` in the beginning?

aka `GraalVMNativeImagePathResource`

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/NativeImagePathResourceFactory.java
```diff
@@ -0,0 +1,71 @@
...
+import java.nio.file.FileSystemNotFoundException;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Collections;
+
+/**
+ * GraalVM Native-Image {@link PathResourceFactory}.
+ */
+public class NativeImagePathResourceFactory extends PathResourceFactory
```
joakime: Can this include the name `GraalVM` in the beginning?

aka `GraalVMNativeImagePathResourceFactory`

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/NativeImagePathResource.java
gregw: I would prefer to avoid GraalVM specific classes.   

However, am I correct is thinking that this class is only needed to work around the Graal bug.  Once that bug is fixed, the factory can just create normal `PathResource` instances?

If so, then maybe even call this `GraalIssue5720PathResource` to make it very clear why it exists.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/NativeImagePathResourceFactory.java
gregw: I'd very much like to avoid GraalVM specific classes.

This class is could be made generic for any scheme that is supported by Path and a FileSystem if the scheme was parameterized.
So perhaps parameterize the scheme and name it something like `FileSystemResourceFactory`?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/NativeImagePathResourceFactory.java
```diff
@@ -0,0 +1,71 @@
...
+        try
+        {
+            URI uri = new URI("resource:/");
+            try
+            {
+                Path.of(uri);
+            }
+            catch (FileSystemNotFoundException e)
+            {
+                FileSystems.newFileSystem(uri, Collections.emptyMap());
```
gregw: If this was a generic class, then we should allow parameters to be configured.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/NativeImagePathResourceFactory.java
```diff
@@ -0,0 +1,71 @@
...
+            }
+        }
+        catch (IOException | URISyntaxException | RuntimeException ignore)
+        {
+            // ignore
```
gregw: Shouldn't we log exceptions as warnings here, as they may be thrown by `newFileSystem` and that feels significant?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
```diff
@@ -52,6 +52,9 @@
         PathResourceFactory pathResourceFactory = new PathResourceFactory();
         RESOURCE_FACTORIES.put("file", pathResourceFactory);
         RESOURCE_FACTORIES.put("jrt", pathResourceFactory);
+
+        if (NativeImagePathResourceFactory.ENABLE_GRAALVM_RESOURCE_SCHEME)
+            RESOURCE_FACTORIES.put("resource", new NativeImagePathResourceFactory());
```
gregw: I'm still cautious about this.  If this was not a VM feature, then I would definitely not like this.

The alternative is to have a module that adds in this mapping "resource" to the resource factory, but then embedded code would still need to do a registration explicitly.

gregw (changes requested):
This is looking pretty good and we'll get it merged one way or another.

I'm on the fence about which way to go to compete this. Either folow @joakime's suggestions of renaming to `GraalVMNativeImageXxx` classes; OR make it generic so that any VM supplied FileSystem type can be supported against an arbitrary scheme.

Perhaps the solution is to make a generic base factory that is parameterized over scheme and FileSystem configuration properties.  Then have a GraalVM specific extension that is instantiated for "resource" and has the bug work around?

Or is that trying to support generic N from an example of 1?  @joakime thoughts?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/NativeImagePathResource.java
joakime: Correct, this class doesn't need to exist once the GraalVM native-image bugs are fixed.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
joakime: We could easily add other Path/FileSystem implementation schemes automatically here.

I would probably access something in Jetty that is known, and use it's protocol/scheme to register other schemes at this point in time.

Eg:

``` java
ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
URL url = classLoader.getResource("org/eclipse/jetty/version/build.properties");
if (url != null)
{
    RESOURCE_FACTORIES.put(url.getProtocol(), pathResourceFactory);
}
```

In the future where the GraalVM bug is fixed, this would be enough to make native-image work.

joakime:
@gregw since this is basically a work around for a bug in the FileSystem implementation in GraalVM native-image.

What about having this as a standalone jar, at https://github.com/jetty-project/graalvm-native-image-bug-resource-factory ?
Then our ResourceFactory defaults could load it via a ServiceLoader?

In the future, when the GraalVM native-image bug is fixed, there only needs to be a small change to `ResourceFactoryInternals` to support any arbitrary Path/FileSystem scheme in use.

See: https://github.com/eclipse/jetty.project/pull/9136/files#r1065133573

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
gregw: Isn't a `newFileSystem` call still needed somewhere?    Doesn't need to be in a `ResourceFactory` but is still needed right?

kohlschuetter: The `newFileSystem` call is now made from within the `NativeImagePathResourceFactory` constructor.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/NativeImagePathResourceFactory.java
kohlschuetter: I don't think so, since there's nothing we can do here.

If any of this fails, subsequent attempts to `Path.of`/`Paths.get` a `resource:` URI will throw an exception from within the calling stack.

We try-catch `RuntimeException` for example to ignore a race condition between us trying to initialize the filesystem and application code that triggers the filesystem initialization independently (we might get a `FileSystemAlreadyExistsException` just at an inopportune time);

kohlschuetter:
I would advise against creating a new jar:

1. If this was an optional dependency, who would be responsible to include it? We would just be creating new obstacles of adopting Jetty with Graal Native-Image.
2. If this dependency is always required, why move it to a separate jar in the first place?

I'm OK with renaming the classes to `GraalIssue5720PathResource`/`-Factory`, in the hope that the original bug will eventually be resolved upstream, and the classes be removed afterwards. I also propose to make these classes package-private to prevent further reliance on this implementation detail.

Regarding @joakime's suggestion to replace the System property check with a call to `Classloader.getResource(reference)`:  
I think that's a reasonable proposal. Since resources are not accessible by default in native-image, we would need to add a configuration file that allows accessing said reference resource. Looking ahead, we should expect such configuration files to exist in Jetty anyways, to further improve compatibility with native-image.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
joakime: @gregw AFAICT, there is only 1 root FileSystem for the `resource:` scheme, representing all of the content in the entire native-image.
So having the `NativeImagePathResourceFactory` control the mount makes sense here.
But a more general approach (for any other FileSystem types) is definitely something to seek out.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/GraalIssue5720PathResourceFactory.java
```diff
@@ -0,0 +1,80 @@
...
+ * @see <a href="https://github.com/oracle/graal/issues/5720">Graal issue 5720</a>
+ */
+final class GraalIssue5720PathResourceFactory extends PathResourceFactory
+{
+    static final boolean ENABLE_NATIVE_IMAGE_RESOURCE_SCHEME;
+
+    static
+    {
+        URL url = GraalIssue5720PathResourceFactory.class.getResource("/org/eclipse/jetty/version/build.properties");
+        ENABLE_NATIVE_IMAGE_RESOURCE_SCHEME = (url != null && "resource".equals(url.getProtocol()));
```
joakime: For this to be true, the bug (eg: `file` or `resource!` is present in `Path.toURI()`) should be detected as well.
Can you do that here? (followup in ResourceFactoryInternals related to this request as well).
We don't want this ResourceFactory to be used once the bug is fixed.

jetty-core/jetty-util/src/main/resources/META-INF/native-image/org.eclipse.jetty/jetty-core/resource-config.json
```diff
@@ -0,0 +1,9 @@
+{
```
joakime, line 1: We would need several dozen more of these scattered around the codebase for this concept.

Can this have a comment indicating it's for GraalVM native-image support?

Frankly, I'm leaning more and more towards a `jetty-graalvm-native-image` support jar, which has all of these things in it, including the bug workaround resource factory.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
```diff
@@ -52,6 +52,9 @@
         PathResourceFactory pathResourceFactory = new PathResourceFactory();
         RESOURCE_FACTORIES.put("file", pathResourceFactory);
         RESOURCE_FACTORIES.put("jrt", pathResourceFactory);
+
+        if (GraalIssue5720PathResourceFactory.ENABLE_NATIVE_IMAGE_RESOURCE_SCHEME)
+            RESOURCE_FACTORIES.put("resource", new GraalIssue5720PathResourceFactory());
```
joakime: This whole block can be replaced with a more general approach, that also supports the bug workaround, along with supporting other FileSystem schemes, and even the future when the GraalVM native-image bug is fixed ...

``` java
        // The default resource factories
        MountedPathResourceFactory mountedPathResourceFactory = new MountedPathResourceFactory();
        RESOURCE_FACTORIES.put("jar", mountedPathResourceFactory);
        PathResourceFactory pathResourceFactory = new PathResourceFactory();
        RESOURCE_FACTORIES.put("file", pathResourceFactory);
        RESOURCE_FACTORIES.put("jrt", pathResourceFactory);

        URL url = ResourceFactoryInternals.class.getResource("/org/eclipse/jetty/version/build.properties");
        if (url != null)
        {
            if (GraalIssue5720PathResourceFactory.hasBug(url))
                RESOURCE_FACTORIES.put(url.getProtocol(), new GraalIssue5720PathResourceFactory());
            else if (!RESOURCE_FACTORIES.contains(url.getProtocol()))
                RESOURCE_FACTORIES.put(url.getProtocol(), mountedPathResourceFactory);
        }
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/GraalIssue5720PathResourceFactory.java
```diff
@@ -0,0 +1,80 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Collections;
+
+/**
+ * GraalVM Native-Image {@link PathResourceFactory}.
+ * 
+ * @see <a href="https://github.com/oracle/graal/issues/5720">Graal issue 5720</a>
+ */
+final class GraalIssue5720PathResourceFactory extends PathResourceFactory
```
joakime, line 26: Could this be extended from the `MountedResourceFactory` and gain any benefits?

joakime (changes requested):
We want to make this work even in the future, once the graalvm native-image bug is fixed.

See comments for suggestions on how to improve this.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
joakime: The `GraalIssue5720PathResourceFactory.hasBug(URL)` would need to check for protocol `resource`, and then perform a path filesystem mount and test of `Path.toURI()` to detect the bug.
Thankfully, this static list is only initialized once at the start of the JVM, so it's not a performance hit.

joakime:
> I would advise against creating a new jar:
> 
> 1. If this was an optional dependency, who would be responsible to include it? We would just be creating new obstacles of adopting Jetty with Graal Native-Image.

This is true of many other 3rd party integrations with Jetty, Graal is not unique or special in this case.

> 2. If this dependency is always required, why move it to a separate jar in the first place?

It's not always required, it's only required if your use case consists of all three: Jetty Server, with GraalVM, with native-image.
If any of those 3 are false, then these things are not required.

Back in the Jetty 9 days we even had a special artifact for users of Oracle JVM running in Compact 3 profile - https://github.com/eclipse/jetty.project/tree/jetty-9.4.x/aggregates/jetty-all-compact3
(It was removed from Jetty 10.x, as Oracle removed compact profiles in newer JVMs)

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/GraalIssue5720PathResourceFactory.java
kohlschuetter: You would still want to enable "resource" as a scheme in `ResourceFactoryInternals`, which is what this boolean controls...

Keep in mind that I started out with deliberately not writing a subclass, and instead use a method that simply fixes the URIs if they have a `resource!` substring, which is a no-op once graal fixes this upstream.

I'm happy to entertain this PR a little longer but I wonder if we're chasing a moving goal.

jetty-core/jetty-util/src/main/resources/META-INF/native-image/org.eclipse.jetty/jetty-core/resource-config.json
kohlschuetter: No, these json files sadly don't accept comments.

I'm also afraid that yes, for native-image you will need a couple more native-image json files in your project. This isn't much of a problem, though, once you get accommodated with the fact that there's this new concept of ahead-of-time-compiled native binaries from Java code.

In my humble opinion, the whole support-jar concept is flawed due to the way (Maven) dependencies are determined. 

As a user of some feature that pulls in a Java dependency (in this case, Jetty), I don't really want to deal with separate artifacts for different JVM implementations.

You see a comparable case with multi-release jars, which are so much better than having per-version jars. It just makes more sense to embrace the capabilities of `META-INF`-based configuration

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
kohlschuetter: I actually don't see a reason for `MountedPathResourceFactory` to exist. Maybe I'm missing a point but you could do this all with a single `PathResourceFactory`. All you'd need to do is to wrap `Path.of` (or `Paths.get`) with a `try-catch` block to find `FileSystemNotFoundException` very much like I do [in this demo code](https://github.com/joakime/graalvm-experiments/blob/ck/src/main/java/graaldemo/Demo.java). You don't have to manually wrangle with exclamation points for jar URLs. Just create a `newFileSystem` for the requested URI. It will work just fine.

Such `PathResourceFactory` would be capable of supporting all `Path`-providing URL schemes.
It would still be the place where would need to handle the "graal issue 5720"-specific manipulation, which is how it was in my original solution.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/GraalIssue5720PathResourceFactory.java
kohlschuetter: see my above comment

joakime:
I went ahead and added a PR to add tests for supporting alternate FileSystem implementations at PR #9149

jetty-core/jetty-util/src/main/resources/META-INF/native-image/org.eclipse.jetty/jetty-core/resource-config.json
joakime: Would we need to expose `META-INF/services` files with this `resource-config.json` too?
Do you know of a maven plugin that can auto-generate these files during the build?

Checking the rest of Jetty 12, I see ...

``` shell
[jetty-12.0.x]$ find . -type f -wholename "*/src/main/resources/*" | grep -v demo | grep -v jetty-home | grep -v "\-test" | grep -v "test-"
./jetty-ee8/jetty-ee8-glassfish-jstl/src/main/resources/readme.txt
./jetty-ee8/jetty-ee8-maven-plugin/src/it/jetty-start-mojo-it/jetty-simple-base/src/main/resources/META-INF/web-fragment.xml
./jetty-ee8/jetty-ee8-maven-plugin/src/it/jetty-effective-web-xml-it/resources/src/main/resources/META-INF/resources/extra.html
./jetty-ee8/jetty-ee8-maven-plugin/src/it/jetty-start-war-distro-mojo-it/jetty-simple-base/src/main/resources/META-INF/web-fragment.xml
./jetty-ee8/jetty-ee8-maven-plugin/src/it/jetty-start-distro-mojo-it/jetty-simple-base/src/main/resources/META-INF/web-fragment.xml
./jetty-ee8/jetty-ee8-maven-plugin/src/it/jetty-start-forked-mojo-it/jetty-simple-base/src/main/resources/META-INF/web-fragment.xml
./jetty-ee8/jetty-ee8-maven-plugin/src/it/jetty-start-war-forked-mojo-it/jetty-simple-base/src/main/resources/META-INF/web-fragment.xml
./jetty-ee8/jetty-ee8-maven-plugin/src/it/jetty-run-mojo-jar-scan-it/MyLibrary/src/main/resources/META-INF/services/javax.servlet.ServletContainerInitializer
./jetty-ee8/jetty-ee8-maven-plugin/src/main/resources/ee8-maven.mod
./jetty-ee8/jetty-ee8-maven-plugin/src/main/resources/maven-ee8.xml
./jetty-ee8/jetty-ee8-maven-plugin/src/main/resources/jetty-ee8-maven.xml
./jetty-ee8/jetty-ee8-webapp/src/main/resources/org/eclipse/jetty/ee8/webapp/catalog-ee8.xml
./tests/jetty-jmh/src/main/resources/jetty-logging.properties
./documentation/jetty-asciidoctor-extensions/src/main/resources/META-INF/services/org.asciidoctor.jruby.extension.spi.ExtensionRegistry
./build/build-resources/src/main/resources/jetty-checkstyle.xml
./jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jetty-server/src/main/resources/META-INF/services/org.eclipse.jetty.ee9.webapp.Configuration
./jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jetty-server/src/main/resources/META-INF/services/jakarta.servlet.ServletContainerInitializer
./jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jakarta-server/src/main/resources/META-INF/services/org.eclipse.jetty.ee9.webapp.Configuration
./jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jakarta-server/src/main/resources/META-INF/services/jakarta.websocket.server.ServerEndpointConfig$Configurator
./jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jakarta-server/src/main/resources/META-INF/services/jakarta.servlet.ServletContainerInitializer
./jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jetty-client/src/main/resources/META-INF/services/org.eclipse.jetty.ee9.webapp.Configuration
./jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jetty-common/src/main/resources/META-INF/services/org.eclipse.jetty.ee9.websocket.api.ExtensionConfig$Parser
./jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jakarta-client/src/main/resources/META-INF/services/jakarta.websocket.ContainerProvider
./jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jakarta-client/src/main/resources/META-INF/services/jakarta.servlet.ServletContainerInitializer
./jetty-ee9/jetty-ee9-plus/src/main/resources/META-INF/services/org.eclipse.jetty.ee9.webapp.Configuration
./jetty-ee9/jetty-ee9-apache-jsp/src/main/resources/META-INF/services/org.apache.juli.logging.Log
./jetty-ee9/jetty-ee9-apache-jsp/src/main/resources/META-INF/services/jakarta.servlet.ServletContainerInitializer
./jetty-ee9/jetty-ee9-osgi/jetty-ee9-osgi-boot/src/main/resources/META-INF/services/org.eclipse.jetty.ee9.webapp.Configuration
./jetty-ee9/jetty-ee9-annotations/src/main/resources/META-INF/services/org.eclipse.jetty.ee9.webapp.Configuration
./jetty-ee9/jetty-ee9-maven-plugin/src/it/jetty-start-mojo-it/jetty-simple-base/src/main/resources/META-INF/web-fragment.xml
./jetty-ee9/jetty-ee9-maven-plugin/src/it/jetty-effective-web-xml-it/resources/src/main/resources/META-INF/resources/extra.html
./jetty-ee9/jetty-ee9-maven-plugin/src/it/jetty-start-war-distro-mojo-it/jetty-simple-base/src/main/resources/META-INF/web-fragment.xml
./jetty-ee9/jetty-ee9-maven-plugin/src/it/jetty-start-distro-mojo-it/jetty-simple-base/src/main/resources/META-INF/web-fragment.xml
./jetty-ee9/jetty-ee9-maven-plugin/src/it/jetty-start-forked-mojo-it/jetty-simple-base/src/main/resources/META-INF/web-fragment.xml
./jetty-ee9/jetty-ee9-maven-plugin/src/it/jetty-start-war-forked-mojo-it/jetty-simple-base/src/main/resources/META-INF/web-fragment.xml
./jetty-ee9/jetty-ee9-maven-plugin/src/it/jetty-run-mojo-jar-scan-it/MyLibrary/src/main/resources/META-INF/services/jakarta.servlet.ServletContainerInitializer
./jetty-ee9/jetty-ee9-maven-plugin/src/main/resources/jetty-ee9-maven.xml
./jetty-ee9/jetty-ee9-maven-plugin/src/main/resources/META-INF/services/org.eclipse.jetty.ee9.webapp.Configuration
./jetty-ee9/jetty-ee9-maven-plugin/src/main/resources/ee9-maven.mod
./jetty-ee9/jetty-ee9-maven-plugin/src/main/resources/maven-ee9.xml
./jetty-ee9/jetty-ee9-quickstart/src/main/resources/META-INF/services/org.eclipse.jetty.ee9.webapp.Configuration
./jetty-ee9/jetty-ee9-jaspi/src/main/resources/META-INF/services/org.eclipse.jetty.ee9.security.Authenticator$Factory
./jetty-ee9/jetty-ee9-openid/src/main/resources/META-INF/services/org.eclipse.jetty.ee9.security.Authenticator$Factory
./jetty-ee9/jetty-ee9-jspc-maven-plugin/src/main/resources/META-INF/m2e/lifecycle-mapping-metadata.xml
./jetty-ee9/jetty-ee9-cdi/src/main/resources/META-INF/services/org.eclipse.jetty.ee9.webapp.Configuration
./jetty-ee9/jetty-ee9-cdi/src/main/resources/META-INF/services/jakarta.servlet.ServletContainerInitializer
./jetty-ee9/jetty-ee9-runner/src/main/resources/META-INF/MANIFEST.MF
./jetty-ee9/jetty-ee9-runner/src/main/resources/jetty-logging.properties
./jetty-ee9/jetty-ee9-nested/src/main/resources/org/eclipse/nested/favicon.ico
./jetty-ee9/jetty-ee9-nested/src/main/resources/jetty-dir.css
./jetty-ee9/jetty-ee9-ant/src/main/resources/tasks.properties
./jetty-ee9/jetty-ee9-ant/src/main/resources/META-INF/services/org.eclipse.jetty.ee9.webapp.Configuration
./jetty-ee9/jetty-ee9-webapp/src/main/resources/META-INF/services/org.eclipse.jetty.ee9.webapp.Configuration
./jetty-ee9/jetty-ee9-webapp/src/main/resources/org/eclipse/jetty/ee9/webapp/catalog-ee9.xml
./jetty-ee9/jetty-ee9-glassfish-jstl/src/main/resources/readme.txt
./jetty-integrations/jetty-infinispan/jetty-infinispan-common/src/main/resources/session.proto
./jetty-ee10/jetty-ee10-maven-plugin/src/it/jetty-start-mojo-it/jetty-simple-base/src/main/resources/META-INF/web-fragment.xml
./jetty-ee10/jetty-ee10-maven-plugin/src/it/jetty-effective-web-xml-it/resources/src/main/resources/META-INF/resources/extra.html
./jetty-ee10/jetty-ee10-maven-plugin/src/it/jetty-start-war-distro-mojo-it/jetty-simple-base/src/main/resources/META-INF/web-fragment.xml
./jetty-ee10/jetty-ee10-maven-plugin/src/it/jetty-start-distro-mojo-it/jetty-simple-base/src/main/resources/META-INF/web-fragment.xml
./jetty-ee10/jetty-ee10-maven-plugin/src/it/jetty-start-forked-mojo-it/jetty-simple-base/src/main/resources/META-INF/web-fragment.xml
./jetty-ee10/jetty-ee10-maven-plugin/src/it/jetty-start-war-forked-mojo-it/jetty-simple-base/src/main/resources/META-INF/web-fragment.xml
./jetty-ee10/jetty-ee10-maven-plugin/src/it/jetty-run-mojo-jar-scan-it/MyLibrary/src/main/resources/META-INF/services/jakarta.servlet.ServletContainerInitializer
./jetty-ee10/jetty-ee10-maven-plugin/src/main/resources/maven-ee10.xml
./jetty-ee10/jetty-ee10-maven-plugin/src/main/resources/META-INF/services/org.eclipse.jetty.ee10.webapp.Configuration
./jetty-ee10/jetty-ee10-maven-plugin/src/main/resources/jetty-ee10-maven.xml
./jetty-ee10/jetty-ee10-maven-plugin/src/main/resources/ee10-maven.mod
./jetty-ee10/jetty-ee10-jspc-maven-plugin/src/main/resources/META-INF/m2e/lifecycle-mapping-metadata.xml
./jetty-ee10/jetty-ee10-openid/src/main/resources/META-INF/services/org.eclipse.jetty.ee10.servlet.security.Authenticator$Factory
./jetty-ee10/jetty-ee10-glassfish-jstl/src/main/resources/readme.txt
./jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jakarta-client/src/main/resources/META-INF/services/jakarta.websocket.ContainerProvider
./jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jakarta-client/src/main/resources/META-INF/services/jakarta.servlet.ServletContainerInitializer
./jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jetty-server/src/main/resources/META-INF/services/org.eclipse.jetty.ee10.webapp.Configuration
./jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jetty-server/src/main/resources/META-INF/services/jakarta.servlet.ServletContainerInitializer
./jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jetty-common/src/main/resources/META-INF/services/org.eclipse.jetty.ee10.websocket.api.ExtensionConfig$Parser
./jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jakarta-server/src/main/resources/META-INF/services/org.eclipse.jetty.ee10.webapp.Configuration
./jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jakarta-server/src/main/resources/META-INF/services/jakarta.websocket.server.ServerEndpointConfig$Configurator
./jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jakarta-server/src/main/resources/META-INF/services/jakarta.servlet.ServletContainerInitializer
./jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jetty-client/src/main/resources/META-INF/services/org.eclipse.jetty.ee10.webapp.Configuration
./jetty-ee10/jetty-ee10-cdi/src/main/resources/META-INF/services/org.eclipse.jetty.ee10.webapp.Configuration
./jetty-ee10/jetty-ee10-cdi/src/main/resources/META-INF/services/jakarta.servlet.ServletContainerInitializer
./jetty-ee10/jetty-ee10-plus/src/main/resources/META-INF/services/org.eclipse.jetty.ee10.webapp.Configuration
./jetty-ee10/jetty-ee10-runner/src/main/resources/META-INF/MANIFEST.MF
./jetty-ee10/jetty-ee10-runner/src/main/resources/jetty-logging.properties
./jetty-ee10/jetty-ee10-annotations/src/main/resources/META-INF/services/org.eclipse.jetty.ee10.webapp.Configuration
./jetty-ee10/jetty-ee10-jaspi/src/main/resources/META-INF/services/org.eclipse.jetty.security.Authenticator$Factory
./jetty-ee10/jetty-ee10-osgi/jetty-ee10-osgi-boot/src/main/resources/META-INF/services/org.eclipse.jetty.ee10.webapp.Configuration
./jetty-ee10/jetty-ee10-ant/src/main/resources/tasks.properties
./jetty-ee10/jetty-ee10-ant/src/main/resources/META-INF/services/org.eclipse.jetty.ee10.webapp.Configuration
./jetty-ee10/jetty-ee10-quickstart/src/main/resources/META-INF/services/org.eclipse.jetty.ee10.webapp.Configuration
./jetty-ee10/jetty-ee10-apache-jsp/src/main/resources/META-INF/services/org.apache.juli.logging.Log
./jetty-ee10/jetty-ee10-apache-jsp/src/main/resources/META-INF/services/jakarta.servlet.ServletContainerInitializer
./jetty-ee10/jetty-ee10-webapp/src/main/resources/META-INF/services/org.eclipse.jetty.ee10.webapp.Configuration
./jetty-ee10/jetty-ee10-webapp/src/main/resources/org/eclipse/jetty/ee10/webapp/catalog-ee10.xml
./jetty-core/jetty-http/src/main/resources/META-INF/services/org.eclipse.jetty.http.HttpFieldPreEncoder
./jetty-core/jetty-http/src/main/resources/org/eclipse/jetty/http/encoding.properties
./jetty-core/jetty-http/src/main/resources/org/eclipse/jetty/http/mime.properties
./jetty-core/jetty-http/src/main/resources/org/eclipse/jetty/http/useragents
./jetty-core/jetty-xml/src/main/resources/org/eclipse/jetty/xml/catalog-configure.xml
./jetty-core/jetty-xml/src/main/resources/org/eclipse/jetty/xml/datatypes.dtd
./jetty-core/jetty-xml/src/main/resources/org/eclipse/jetty/xml/xml.xsd
./jetty-core/jetty-xml/src/main/resources/org/eclipse/jetty/xml/configure_7_6.dtd
./jetty-core/jetty-xml/src/main/resources/org/eclipse/jetty/xml/configure_9_0.dtd
./jetty-core/jetty-xml/src/main/resources/org/eclipse/jetty/xml/catalog-org.w3.xml
./jetty-core/jetty-xml/src/main/resources/org/eclipse/jetty/xml/configure_10_0.dtd
./jetty-core/jetty-xml/src/main/resources/org/eclipse/jetty/xml/configure_6_0.dtd
./jetty-core/jetty-xml/src/main/resources/org/eclipse/jetty/xml/configure_9_3.dtd
./jetty-core/jetty-xml/src/main/resources/org/eclipse/jetty/xml/XMLSchema.dtd
./jetty-core/jetty-http-spi/src/main/resources/META-INF/services/com.sun.net.httpserver.spi.HttpServerProvider
./jetty-core/jetty-alpn/jetty-alpn-conscrypt-client/src/main/resources/META-INF/services/org.eclipse.jetty.io.ssl.ALPNProcessor$Client
./jetty-core/jetty-alpn/jetty-alpn-java-client/src/main/resources/META-INF/services/org.eclipse.jetty.io.ssl.ALPNProcessor$Client
./jetty-core/jetty-alpn/jetty-alpn-java-server/src/main/resources/META-INF/services/org.eclipse.jetty.io.ssl.ALPNProcessor$Server
./jetty-core/jetty-alpn/jetty-alpn-conscrypt-server/src/main/resources/META-INF/services/org.eclipse.jetty.io.ssl.ALPNProcessor$Server
./jetty-core/jetty-start/src/main/resources/org/eclipse/jetty/start/build.properties
./jetty-core/jetty-start/src/main/resources/org/eclipse/jetty/start/usage.txt
./jetty-core/jetty-deploy/src/main/resources/org/eclipse/jetty/deploy/lifecycle-bindings.txt
./jetty-core/jetty-http3/jetty-http3-qpack/src/main/resources/META-INF/services/org.eclipse.jetty.http.HttpFieldPreEncoder
./jetty-core/jetty-http2/jetty-http2-hpack/src/main/resources/META-INF/services/org.eclipse.jetty.http.HttpFieldPreEncoder
./jetty-core/jetty-util/src/main/resources/org/eclipse/jetty/version/build.properties
./jetty-core/jetty-slf4j-impl/src/main/resources/META-INF/services/org.slf4j.spi.SLF4JServiceProvider
./jetty-core/jetty-server/src/main/resources/org/eclipse/jetty/server/favicon.ico
./jetty-core/jetty-server/src/main/resources/org/eclipse/jetty/server/jetty-dir.css
./jetty-core/jetty-jndi/src/main/resources/jndi.properties
./jetty-core/jetty-quic/jetty-quic-quiche/jetty-quic-quiche-jna/src/main/resources/META-INF/services/org.eclipse.jetty.quic.quiche.QuicheBinding
./jetty-core/jetty-quic/jetty-quic-quiche/jetty-quic-quiche-foreign-incubator/src/main/resources/META-INF/services/org.eclipse.jetty.quic.quiche.QuicheBinding
./jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/resources/META-INF/services/org.eclipse.jetty.websocket.core.Extension
```

joakime: The first modules I would worry about are ...

* `jetty-http`
* `jetty-xml`
* `jetty-server`

kohlschuetter: Yes, `META-INF/services` would also need to be included in resource-config.json.

An initial list for native-image configurations can be obtained by running the Native Image [Tracing Agent](https://www.graalvm.org/latest/reference-manual/native-image/metadata/AutomaticMetadataCollection/) using non-native GraalVM on either your final product (perform some regular usage patterns, the files will be saved upon JVM termination) or by running all unit tests.

The agent will also determine reflection accesses (not all of them, unfortunately).

For my [junixsocket](https://github.com/kohlschutter/junixsocket/) project, I've created a [small script](https://github.com/kohlschutter/junixsocket/blob/main/junixsocket-native-graalvm/bin/build-selftest) along with [instructions](https://kohlschutter.github.io/junixsocket/graalvm.html) to determine these files programmatically via a selftest application. The selftest app essentially runs all unit tests on the same platform as your native-image target and so you can increase the chance of catching all necessary configurations (macOS/Linux/Windows).

Getting this started is a bit tedious but once the files are in place, users of your library won't have to worry anymore.

Regular expressions can be used to implement "wildcard" rules, but don't make them too wide so we only include what is provided by your own artifact, not the entire Java universe of unrelated resources.

(Note that all of this is out of scope for this PR)

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/NativeImagePathResourceFactory.java
gregw: I think at least a debug log would be helpful.   If somebody is getting URI exceptions elsewhere because the filesystem didn't mount, it would be good to be able to see the exceptions without needing to modify code

gregw (approved):
I think there are probably refinements that can be made, but as we still have some thinking to do on the whole resource factory stuff, I think this PR is better merged and considered as part of the whole as we clean up resources for a beta release of 12.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
janbartel: If `GraalIssue5720PathResourceFactory` and `GraalIssue5720PathResource` classes only exist because of the graal bug && line 56 will be deleted once it's fixed && and line 57 can be moved elsewhere from `ResourceFactoryInternals` once it's fixed, then I'm ok with it. 

What I would like us to do is to use this example to learn how to handle, in a generic way, any url scheme. So far we've now had 2 examples: `bundleresource:` and `resource:`. Have we used those to extract a meaningful generic mechanism? If so, can we document that please.

kohlschuetter:
As per @gregw 
> I think there are probably refinements that can be made, but as we still have some thinking to do on the whole resource factory stuff, I think this PR is better merged and considered as part of the whole as we clean up resources for a beta release of 12.

@joakime Any blocking changes or can this be merged?

joakime:
@kohlschuetter this PR has conflicts currently, and cannot be merged.

You'll need to fix the conflicts.
Also, please [implement the change from my comment](https://github.com/eclipse/jetty.project/pull/9136/files#r1065892708) as you are resolving the conflicts.

joakime (approved):

PR #9142, opened 2023-01-09
Fixes #9141 - Thread-safe Content.Chunk#slice

### LINKED ISSUE #9141: Thread-safe Content.Chunk#slice
**Jetty version(s)**
12

**Description**

The implementation of [`Content.Chunk#slice(int,int,boolean)`](https://github.com/eclipse/jetty.project/blob/jetty-12.0.0.alpha3/jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java#L605-L621) is not threadsafe.

If multiple threads are slicing off the same chunk, then incorrect results may occur as each thread modifies the pointers of the original chunk.   It should be able to use the `ByteBuffer#slice(int,int)` method to slice without modifying the original buffer.

sbordet:
* Changed Content.Chunk.slice(int, int, boolean) to have the same parameters as ByteBuffer.slice(int, int) for consistency.
* Updated Chunk.slice(int, int, boolean) javadocs.
* Update code that was calling Chunk.slice(int, int, boolean).

### DIFF
--- a/jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
+++ b/jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
@@ -597,7 +597,7 @@ public class Content
 
         /**
          * <p>Returns a new {@code Chunk} whose {@code ByteBuffer} is a slice of the
-         * {@code ByteBuffer} of the this {@code Chunk}.</p>
+         * {@code ByteBuffer} of this {@code Chunk}.</p>
          * <p>The returned {@code Chunk} is:</p>
          * <ul>
          * <li>{@code this}, if this {@code Chunk} is {@link #isTerminal() terminal}</li>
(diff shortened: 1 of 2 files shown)


jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -596,36 +596,34 @@ default Chunk slice()
         }
 
         /**
-         * <p>Returns a new {@code Chunk} whose {@code ByteBuffer} is a slice, with the given
-         * position and limit, of the {@code ByteBuffer} of the source {@code Chunk} unless the
-         * source is {@link #isTerminal() terminal} in which case {@code this} is returned, or
-         * if {@code position == limit} in which case {@link #EOF} or {@link #EMPTY} is
-         * returned depending on the value of {@code last}.</p>
+         * <p>Returns a new {@code Chunk} whose {@code ByteBuffer} is a slice of the
+         * {@code ByteBuffer} of the this {@code Chunk}.</p>
```
lorban: typo: `of the this` -> `of this`

lorban (changes requested):
LGTM besides a typo in the javadoc.

lorban (approved):

gregw (approved):

PR #9143, opened 2023-01-09
Remove `@Disabled` from `jetty-jmx`

joakime:
Removing `@Disabled` annotation from `jetty-core/jetty-jmx` tests

### DIFF
--- a/jetty-core/jetty-jmx/src/main/java/org/eclipse/jetty/jmx/ConnectorServer.java
+++ b/jetty-core/jetty-jmx/src/main/java/org/eclipse/jetty/jmx/ConnectorServer.java
@@ -59,7 +59,7 @@ public class ConnectorServer extends AbstractLifeCycle
 {
     public static final String RMI_REGISTRY_CLIENT_SOCKET_FACTORY_ATTRIBUTE = "com.sun.jndi.rmi.factory.socket";
     private static final Logger LOG = LoggerFactory.getLogger(ConnectorServer.class);
-    public static final int DEFAULT_PORT = 1099;
+    private static final int DEFAULT_REGISTRY_PORT = 1099;
 
     private JMXServiceURL _jmxURL;
     private final Map<String, Object> _environment;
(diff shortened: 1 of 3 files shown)


jetty-core/jetty-jmx/src/main/java/org/eclipse/jetty/jmx/ConnectorServer.java
```diff
@@ -58,6 +59,7 @@ public class ConnectorServer extends AbstractLifeCycle
 {
     public static final String RMI_REGISTRY_CLIENT_SOCKET_FACTORY_ATTRIBUTE = "com.sun.jndi.rmi.factory.socket";
     private static final Logger LOG = LoggerFactory.getLogger(ConnectorServer.class);
+    public static final int DEFAULT_PORT = 1099;
```
sbordet: Make it `private` and rename to `DEFAULT_REGISTRY_PORT`.

jetty-core/jetty-jmx/src/main/java/org/eclipse/jetty/jmx/ConnectorServer.java
```diff
@@ -182,7 +184,20 @@ public void doStart() throws Exception
...
+            if (StringUtil.isBlank(rawHost)) // no host
+            {
+                hostPort = new HostPort(InetAddress.getLocalHost().getHostName(), DEFAULT_PORT);
+            }
+            else if (rawHost.startsWith(":")) // port without host
+            {
+                hostPort = new HostPort(InetAddress.getLocalHost().getHostName() + rawHost);
+            }
+            else
+                hostPort = new HostPort(rawHost);
```
sbordet, line 195: Drop the braces, since they are all one-liners.

Also, `getHostName()` performs a reverse DNS lookup -- use `getHostAddress()` instead?

sbordet (changes requested):

jetty-core/jetty-jmx/src/main/java/org/eclipse/jetty/jmx/ConnectorServer.java
joakime: Done.

jetty-core/jetty-jmx/src/main/java/org/eclipse/jetty/jmx/ConnectorServer.java
joakime: Done.

sbordet (approved):
LGTM. 
However, I'm worried that because this test uses default ports (1099) that might be taken already.
We'll see if CI passes.

joakime:
@sbordet it's green. yay!

Merging.

PR #9149, opened 2023-01-10
Jetty 12 - Add tests in util/resource for alternate FileSystem implementations

joakime:
This adds some jetty-util Resource tests for alternate FileSystem providers, to ensure that we are not doing anything awkward in our implementation that would prevent it.

This uses the https://github.com/google/jimfs implementation of an In-Memory FileSystem implementation to accomplish these tests.
(This filesystem uses scheme `jimfs` and has a URI authority representing the named FileSystem)

This was initiated as a response to a question about our support for alternate FileSystem implementations in PR #9136

### DIFF
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
@@ -16,6 +16,7 @@ package org.eclipse.jetty.util.resource;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URL;
 import java.nio.file.Path;
 import java.nio.file.ProviderNotFoundException;
 import java.util.List;
@@ -48,10 +49,17 @@ class ResourceFactoryInternals
         CURRENT_WORKING_DIR = Path.of(System.getProperty("user.dir"));
 
         // The default resource factories
-        RESOURCE_FACTORIES.put("jar", new MountedPathResourceFactory());
+        MountedPathResourceFactory mountedPathResourceFactory = new MountedPathResourceFactory();
+        RESOURCE_FACTORIES.put("jar", mountedPathResourceFactory);
         PathResourceFactory pathResourceFactory = new PathResourceFactory();
         RESOURCE_FACTORIES.put("file", pathResourceFactory);
         RESOURCE_FACTORIES.put("jrt", pathResourceFactory);
+
+        URL url = ResourceFactoryInternals.class.getResource("/org/eclipse/jetty/version/build.properties");
+        if ((url != null) && !RESOURCE_FACTORIES.contains(url.getProtocol()))
+        {
+            RESOURCE_FACTORIES.put(url.getProtocol(), mountedPathResourceFactory);
+        }
     }
 
     static ResourceFactory ROOT = new CompositeResourceFactory()
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
@@ -55,6 +55,12 @@ class ResourceFactoryInternals
         RESOURCE_FACTORIES.put("file", pathResourceFactory);
         RESOURCE_FACTORIES.put("jrt", pathResourceFactory);
 
+        /* Best effort attempt to detect that an alternate FileSystem type that is in use.
+         * We don't attempt to look up a Class, as not all runtimes and environments have the classes anymore
+         * (eg: they were compiled into native code)
+         * The build.properties is present in the jetty-util jar, so it's reasonably safe to look for that
+         * as a resource
+         */
         URL url = ResourceFactoryInternals.class.getResource("/org/eclipse/jetty/version/build.properties");
         if ((url != null) && !RESOURCE_FACTORIES.contains(url.getProtocol()))
         {
--- a/jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/AlternateFileSystemResourceTest.java
+++ b/jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/AlternateFileSystemResourceTest.java
@@ -1,0 +1,112 @@
+//
+// ========================================================================
+// Copyright (c) 1995-2022 Mort Bay Consulting Pty Ltd and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.util.resource;
+
+import java.io.IOException;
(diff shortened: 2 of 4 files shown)


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
```diff
@@ -48,10 +49,17 @@
...
 
         // The default resource factories
-        RESOURCE_FACTORIES.put("jar", new MountedPathResourceFactory());
+        MountedPathResourceFactory mountedPathResourceFactory = new MountedPathResourceFactory();
+        RESOURCE_FACTORIES.put("jar", mountedPathResourceFactory);
         PathResourceFactory pathResourceFactory = new PathResourceFactory();
         RESOURCE_FACTORIES.put("file", pathResourceFactory);
         RESOURCE_FACTORIES.put("jrt", pathResourceFactory);
+
+        URL url = ResourceFactoryInternals.class.getResource("/org/eclipse/jetty/version/build.properties");
```
lorban, line 64: Is that code meant to discover the scheme of internal resources' URL? It's worth commenting a bit as this logic is mildly cryptic.

Also, are you sure `build.properties` is always going to be available? Can't you load the `ResourceFactoryInternals.class` file's bytes instead as you can be confident this is always available?

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/AlternateFileSystemResourceTest.java
```diff
@@ -0,0 +1,112 @@
...
+{
+    private static final Logger LOG = LoggerFactory.getLogger(AlternateFileSystemResourceTest.class);
+    private FileSystem jimfs;
+    private URI fsBaseURI;
+
+    @BeforeEach
+    public void initInMemoryFileSystem(TestInfo testInfo)
+    {
+        Optional<Method> testMethod = testInfo.getTestMethod();
+        String testMethodName = testMethod.map(Method::getName).orElseGet(AlternateFileSystemResourceTest.class::getSimpleName);
```
lorban: I would call `orElseThrow()` instead of `orElseGet()`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
```diff
@@ -48,10 +49,17 @@
...
+        if ((url != null) && !RESOURCE_FACTORIES.contains(url.getProtocol()))
+        {
+            RESOURCE_FACTORIES.put(url.getProtocol(), mountedPathResourceFactory);
```
lorban, line 67: Why not using a new `MountedPathResourceFactory` instance?

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/AlternateFileSystemResourceTest.java
```diff
@@ -0,0 +1,112 @@
...
+        jimfs = Jimfs.newFileSystem(testMethodName, Configuration.unix());
+        fsBaseURI = jimfs.getPath("/").toUri();
+
+        ResourceFactory.registerResourceFactory(fsBaseURI.getScheme(), new MountedPathResourceFactory());
```
lorban, line 72: jimfs doesn't really need mounting, does it?

lorban (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
joakime: Yeah, I'll add comments.

Oddly, on some environments (like GraalVM or Android) there are no class files to request as a resource.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/AlternateFileSystemResourceTest.java
joakime: Nah, I want a name from the Test framework.
I'll rework with a different approach.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
joakime: No real reason to, the implementation wouldn't change with each new protocol.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/AlternateFileSystemResourceTest.java
joakime: Some FileSystems require it to be mounted using its technique in order for it to be used by Jetty, basically it was mounted ahead of time and then Jetty was started.

This is how GraalVM native-image and Android work.

joakime: The point of this test case is to ensure that we don't put any artificial restrictions on "jar" or "jar:file:" only.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/AlternateFileSystemResourceTest.java
joakime: Changed.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
joakime: More comments added

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
joakime: An argument could be made that there should be a test here, to see if the URL points to a mounted FileSystem or not, and adding the appropriate ResourceFactory based on results.

Or using some variation of the `mountIfNeeded()` logic always here?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
kohlschuetter: This check will not succeed for native-image environments unless you add a corresponding setting to allow access to to /org/eclipse/jetty/version/build.properties (e.g., via resource-config.json)

joakime: Like the documentation says in the current HEAD of this branch, this is a best effort to discover schemes that are in use.

I'm aware that native-image needs more metadata to expose resources, but that's what your PR is for (and your PR is still relevant, this PR does not replace it)

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/ResourceFactoryInternals.java
lorban: If adding some logic to figure out if mounting is needed or not doesn't take too much, let's add it. Otherwise let's get these changes make it in and eventually open an enhancement bug.

lorban (approved):

PR #9151, opened 2023-01-10
Jetty 12.0.x document modules

gregpoulos:
@sbordet I wanted to run a couple of potential changes by you to get your thoughts.

1) What do you think of using the `module-eex-modulename.adoc` naming convention when we create documentation listings for modules that have multiple different versions for different EE platforms? The main advantage here is that `eex` uses a common substitution convention that plays nicely with filenames. One disadvantage to this approach is that there there isn't an obvious "right" way to do alphabetical sorting. For example, `module-eex-webapp.adoc` get automatically sorted under "E", but there's a good argument that it should appear under "W".

2) For the `ee{8,9,10}-webapp` module documentation, I've used an ASCIIdoc `include::` statement to directly transclude the `[description]` text directly from `ee10-webapp.mod`. My goal in doing so is to reduce duplicated content, but if you'd prefer to avoid this approach, I can stop. (For example, maybe we'd prefer to avoid cluttering the `.mod` files with extra markup, or perhaps the `[description]` field serves a sufficiently distinct purpose from the documentation that they're worth maintaining separately.)


sbordet:
> What do you think of using the module-eex-modulename.adoc

Fine by me. Maybe `eeN` is better than `eex`?

> I've used an ASCIIdoc include:: statement to directly transclude the [description] text directly from ee10-webapp.mod.

Good idea too.
I think perhaps you can shorten the tags, from `documentation-description` to just `description`, and `documentation-ini-template` to just `ini-template`.

sbordet:
@gregpoulos I would standardize to `Servlet` with capital `S` everywhere.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-eex-webapp.adoc
```diff
@@ -0,0 +1,25 @@
...
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+[[og-module-eex-webapp]]
```
sbordet: Maybe I like better `eeN`, the difference in case of the `N` is more understandable, wdyt?

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-eex-webapp.adoc
```diff
@@ -0,0 +1,25 @@
...
+===== Module `ee{8,9,10}-webapp`
```
sbordet: Can we parametrize this recurring `ee{8,9,10}` into a variable in `operations-guide/.asciidoctorconfig`?

I think all the variables so far are lower case except `JETTY_HOME`... so perhaps convert to `jetty-home` the variable name, and introduce `ee-all` and `ee-current` variables?

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-eex-webapp.adoc
```diff
@@ -0,0 +1,25 @@
...
+
+include::{JETTY_HOME}/modules/ee10-webapp.mod[tags=documentation-description]
```
sbordet: If the variable thing works, use `{ee-current}` instead of `ee10` here?

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-jmx.adoc
```diff
@@ -0,0 +1,20 @@
...
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+[[og-module-jmx]]
+===== Module `jmx`
+
+Enabling this module allows you to locally access and configure Jetty components using a Java Management Extension (JMX) tool.
```
sbordet: Pull from the `[description]` section of the module?

sbordet (changes requested):

gregpoulos:
All these suggestions look good to me. I'll implement them and post an update with the changes.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-eex-webapp.adoc
gregpoulos: This is done, but I'm going do the conversion of "JETTY_HOME" to lower-case in a separate pull request since it will touch a lot of files.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/modules-standard.adoc
```diff
@@ -17,7 +17,7 @@
 include::module-alpn.adoc[]
 include::module-bytebufferpool.adoc[]
 include::module-console-capture.adoc[]
-include::module-deploy.adoc[]
+include::module-eeN-deploy.adoc[]
```
gregpoulos, line 20: @sbordet: As far as I can tell, the functionality of the old `deploy` module has been mostly rolled into the `eeN-deploy` modules. For this reason, I renamed the documentation file.

However, there still seems to be a `deploy` module (i.e., without the `eeN` prefix) [hanging around](https://github.com/gregpoulos/jetty.project/blob/jetty-12.0.x-document-modules/jetty-core/jetty-deploy/src/main/config/modules/deploy.mod). It's not clear to me what purpose it serves; it seems to be a transitive dependency, and not something that the end-user will need to look at very often. For this reason, I don't think it makes sense to document it for the time being.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/begin/deploy.adoc
```diff
@@ -46,46 +46,46 @@ mywebapp.war
...
 ====
 
-To deploy a standard web application, you need to enable the xref:og-module-deploy[`deploy` module]:
+To deploy a standard web application, you need to enable the xref:og-module-eeN-deploy[`{ee-current}-deploy` module].
+
+NOTE: The following examples assume you're deploying a Jakarta {ee-current-caps} app; for other versions of EE, make sure to activate the corresponding `{ee-all}-deploy` module.
 
 ----
-$ java -jar $JETTY_HOME/start.jar --add-module=deploy
+$ java -jar $JETTY_HOME/start.jar --add-module={ee-current}-deploy
```
gregpoulos, line 55: @sbordet: Based on my understanding of the updates in Jetty 12, I think this tutorial should direct users to add the EE-prefixed `eeN-deploy` module rather than the plain `deploy` module. However, please correct me if this is wrong!

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-eeN-webapp.adoc
```diff
@@ -0,0 +1,26 @@
...
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+[[og-module-eeN-webapp]]
+===== Module `{ee-all}-webapp`
+
+include::{JETTY_HOME}/modules/{ee-current}-webapp.mod[tags=description]
+
+If this module is not enabled, only Jetty-specific handlers may be deployed.
```
gregpoulos: @sbordet: I'm not entirely sure what "only Jetty-specific handlers" means in this line of the documentation. Is there another section of the docs where this use case is explained?

documentation/jetty-documentation/src/main/asciidoc/operations-guide/begin/deploy.adoc
```diff
@@ -46,46 +46,46 @@ mywebapp.war
...
+include::jetty[setupArgs="--add-module=http",args="--add-module={ee-current}-deploy"]
 ----
 
-The `deploy` module creates `$JETTY_BASE/webapps`, which is the directory where Jetty looks for any `+*.war+` files or web application directories to deploy.
+The `{ee-current}-deploy` module creates `$JETTY_BASE/webapps`, which is the directory where Jetty looks for any `+*.war+` files or web application directories to deploy.
 
-[NOTE]
-====
-Activating Jetty's `deploy` module enables web application deployment.
+Activating one of Jetty's `{ee-all}-deploy` modules enables web application deployment.
```
gregpoulos, line 65: @sbordet: I've been using the words "enable", "activate", and "add" pretty much interchangeably when referring to modules. Is one of these terms preferred? If not, it may make sense to standardize the terminology.

Another way to think about it: are any of the following sentences more conceptually accurate? Or are they all about the same?
* When I **add** a module, it **activates** one of Jetty's features.
* When I **activate** a module, Jetty **enables support for** some feature.
* When I **enable** a module, it **adds support for** some feature in Jetty.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/begin/deploy.adoc
sbordet: Is it correct.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/begin/deploy.adoc
```diff
@@ -46,46 +46,47 @@ mywebapp.war
 <5> `WEB-INF/web.xml` is the web application deployment descriptor, which defines the components and the configuration of your web application.
 ====
 
-To deploy a standard web application, you need to enable the xref:og-module-deploy[`deploy` module]:
+To deploy a standard web application, you need to enable the xref:og-module-eeN-deploy[`{ee-current}-deploy` module].
 
+NOTE: The following examples assume you're deploying a Jakarta {ee-current-caps} app; for other versions of EE, make sure to activate the corresponding `{ee-all}-deploy` module.
```
sbordet:
```suggestion
NOTE: The following examples assume you're deploying a Jakarta {ee-current-caps} application; for other versions of Jakarta EE, make sure to activate the corresponding `{ee-all}-deploy` module.
```

sbordet: Let's resist the temptation to abbreviate things such as "app", etc. 😃

documentation/jetty-documentation/src/main/asciidoc/operations-guide/begin/deploy.adoc
sbordet: Good question, I would not use "activate" and stick with "enable" and "add" interchangeably.

"When you **add** a module, it **enables** support for feature X" 

or similarly:

"When you **enable** a module, it **adds** support for feature X".

documentation/jetty-documentation/src/main/asciidoc/operations-guide/begin/deploy.adoc
```diff
@@ -46,46 +46,47 @@ mywebapp.war
...
 
-[NOTE]
-====
-Activating Jetty's `deploy` module enables web application deployment.
+Activating one of Jetty's `{ee-all}-deploy` modules enables web application deployment.
 Whether these web applications are served via clear-text HTTP/1.1, or secure HTTP/1.1, or secure HTTP/2, or HTTP/3 (or even all of these protocols) depends on if the correspondent Jetty modules have been enabled.
-
-Refer to the xref:og-protocols[section about protocols] for further information.
-====
+(Refer to the xref:og-protocols[section about protocols] for further information.)
```
sbordet: Remove parens.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/begin/deploy.adoc
```diff
@@ -96,12 +97,12 @@ $ java -jar $JETTY_HOME/start.jar
...
 ----
-include::jetty[setupArgs="--add-modules=http,deploy,ee10-demo-simple",highlight="WebAppContext"]
+include::jetty[setupArgs="--add-modules=http,deploy,{ee-current}-demo-simple",highlight="WebAppContext"]
 ----
 
-Note the highlighted line that logs the deployment of `ee10-demo-simple.war`.
+Note the highlighted line that logs the deployment of `{ee-current}-demo-simple.war`.
 
-Now you can access the web application by pointing your browser to `+http://localhost:8080/ee10-demo-simple+`.
+Now you can access the web application by pointing your browser to `pass:a[http://localhost:8080/{ee-current}-demo-simple]`.
```
sbordet, line 105: Excellent!
(the use of `pass:a`)

documentation/jetty-documentation/src/main/asciidoc/operations-guide/jaspi/chapter.adoc
```diff
@@ -27,28 +27,29 @@ Only modules conforming to the "Servlet Container Profile" with the ServerAuthMo
...
+include::{JETTY_HOME}/modules/{ee-current}-jaspi.mod[]
 ----
 
 [[og-jaspi-xml]]
 ===== Configure JASPI
 
-Activate the `ee{9,10}-jaspi` module that matches your EE platform version.
+Activate either the `ee9-jaspi` or `{ee-current}-jaspi` module, whichever matches your EE platform version.
 
+[subs=attributes]
```
sbordet, line 38: Do you need `[subs=attributes]`? 
I ask because few lines above and few lines below there are similar code snippets with `include`s, and they do not have `[subs=attributes]`.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-eeN-deploy.adoc
```diff
@@ -0,0 +1,38 @@
...
+// ========================================================================
+//
+
+[[og-module-eeN-deploy]]
+===== Module `{ee-all}-deploy`
+
+include::{JETTY_HOME}/modules/{ee-current}-deploy.mod[tags=description]
+
+Deployment is managed via a `DeploymentManager` component that watches a directory for changes.
+(See xref:og-deploy[how to deploy web applications] for more information.)
```
sbordet: Drop parens.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-eeN-deploy.adoc
```diff
@@ -0,0 +1,38 @@
...
+
+Adding files or directories to this monitored directory will cause the `DeploymentManager` to deploy them as web applications; updating files already existing in this monitored directory will cause the `DeploymentManager` to re-deploy the corresponding web application; removing files in this monitored directory will cause the `DeploymentManager` to "undeploy" the corresponding web application.
+(You can find a more detailed discussion of these rules in the xref:og-deploy-rules[deployment rules] section.)
+
+Multiple versions of this module exist (`{ee-all}-deploy`) to support each EE platform's version of the Java Servlet specification.
```
sbordet: `EE` -> `Jakarta EE`

Let's standardize on "Jakarta EE" everywhere, rather than just "EE".

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-eeN-deploy.adoc
```diff
@@ -0,0 +1,38 @@
...
+Jetty's configuration properties are nearly identical across these versions; the module file for `{ee-current}-deploy` is as follows:
```
sbordet: It's not the module file as a whole, just the ini-template section.
So perhaps: 

```suggestion
Jetty's configuration properties are nearly identical across these versions; the configuration properties for the `{ee-current}-deploy` Jetty module are:
```

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-eeN-webapp.adoc
sbordet: No we need to write that section. For now remove this line, as it duplicates what's in the module description.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-eeN-webapp.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+Multiple versions of this module exist (`{ee-all}-webapp`) to support each EE platform's version of the Java Servlet specification.
```
sbordet: EE -> Jakarta EE

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-eeN-webapp.adoc
```diff
@@ -0,0 +1,26 @@
...
+Jetty's configuration properties are identical across the `{ee-all}-` versions of this module, and are as follows:
```
sbordet: `{ee-all}-` has an extra dash?

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-resources.adoc
```diff
@@ -0,0 +1,22 @@
...
+// ========================================================================
+//
+
+[[og-module-resources]]
+===== Module `resources`
+
+include::{JETTY_HOME}/modules/resources.mod[tags=description]
+
+A common use-case for this module is to provide resources for third-party libraries via the xref:og-start-start-class-path[server classpath].
+For instance, many logging libraries (including https://logging.apache.org/log4j/2.x/[log4j] and https://logback.qos.ch/[Logback]) look for their configuration files on the classpath.
```
sbordet: log4j -> Log4j2

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-resources.adoc
```diff
@@ -0,0 +1,22 @@
...
+
+In fact, Jetty's own logging library -- enabled via the `logging-jetty` module -- stores its configuration in `$JETTY_BASE/resources/jetty-logging.properties`.
```
sbordet:
```suggestion
Jetty provides a logging library implementation -- enabled via the `logging-jetty` module -- whose configuration file is `$JETTY_BASE/resources/jetty-logging.properties`.
```

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/modules-standard.adoc
sbordet: The `deploy` module is an "abstract" base module that the `eeN-deploy` modules depend on.
Agreed that it's "internal" and does not need docs.

jetty-core/jetty-deploy/src/main/config/modules/deploy.mod
```diff
@@ -1,5 +1,5 @@
 [description]
-Enables web application deployment from the $JETTY_BASE/webapps/ directory.
+This module enables webapp deployment from the `$JETTY_BASE/webapps` directory.
```
sbordet: webapp -> web application

sbordet: Please move to this module the section

```
[files]
webapps/
```

and remove it from the 3 `eeN-deploy.mod` files, to reduce duplication.

jetty-ee10/jetty-ee10-webapp/src/main/config/modules/ee10-deploy.mod
```diff
@@ -1,5 +1,7 @@
 [description]
-Enables web application deployment from the $JETTY_BASE/webapps/ directory.
+# tag::description[]
+This module enables webapp deployment from the `$JETTY_BASE/webapps` directory.
+# end::description[]
```
sbordet, line 4: Remove the empty `[lib]` section.

sbordet (changes requested):

documentation/jetty-documentation/src/main/asciidoc/operations-guide/begin/deploy.adoc
gregpoulos: Sounds good. I'll use this convention moving forward, and will fix existing usages as I come across them.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/jaspi/chapter.adoc
gregpoulos: You're right: this is unnecessary. I had added it in an attempt to fix a separate issue, but that issue turned out to be unrelated. I'll revert.

gregpoulos: For reasons that weren't totally clear to me, the `[subs=attributes]` annotation was necessary here but not in the other cases.

After a little research, I think I found the explanation [here](https://docs.asciidoctor.org/asciidoc/latest/directives/include/#include-processing):

> Although the include directive looks like a block macro, it’s not a macro and therefore isn’t processed like one. It’s a preprocessor directive; it’s important to understand the distinction.
> 
> A preprocessor directive is processed when the lines of a document are read, but before the document structure is parsed. Therefore, it’s not aware of the surrounding document structure. A preprocessor directive merely adds lines to the reader or takes lines away. The include directive is a preprocessor directive that always adds lines.

So I think you typically do need `[subs=attributes]` for literal blocks that are offset with `----`, but the `include::` directive is a special case, because it is being processed before the document parser runs. Fortunately, it seems like the preprocessor is smart enough to do the attribute substitution, which is why the `include::` directive still executes correctly.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-eeN-deploy.adoc
gregpoulos: Sounds good to me.

documentation/jetty-documentation/src/main/asciidoc/operations-guide/modules/module-eeN-webapp.adoc
gregpoulos: Reworded this to fix it.

sbordet (approved):
Good job!

PR #9152, opened 2023-01-10
For consistency, convert "JETTY_HOME" attribute to "jetty-home"

gregpoulos:
This is the implementation of [a suggestion from @sbordet](https://github.com/eclipse/jetty.project/pull/9151#discussion_r1066309175).


sbordet (approved):
👍🏼

PR #9153, opened 2023-01-11
Jetty 12 temp directory cleanup

gregw:
+ Server now has setter/getter for a temp directory
 + Context now has a getter for a temp directory
 + Server root Context temp directory is either whatever is set, else a work directory, else java.io.tmpdir
 + WebInfConfiguration will still create a temp directory name, but defers to ContextHandler for creation/persistence of the temp directory
 + temp directory (and BASE) removed from the deployer, as it is now the server temp directory.


jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java
```diff
@@ -132,6 +135,48 @@ public String getServerInfo()
...
+    public void setTempDirectory(String temp)
+    {
+        setTempDirectory(new File(temp));
+    }
+
+    /**
+     * <p>Set the temporary directory returned by {@link Context#getTempDirectory()} for the root
+     * {@link Context} returned {@link #getContext()}. If not set explicitly here, then the root
+     * {@link Context#getTempDirectory()} will return either the directory found at
+     * {@code new File(IO.asFile(System.getProperty("jetty.base")), "work")} if it exists, else
```
janbartel: Duplicate "else" on lines 153 and 154

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -179,6 +183,60 @@ protected ScopedContext newContext()
...
+    public File getTempDirectory()
+    {
+        return _tempDirectory;
+    }
+
+    /**
+     * <p>Set the temporary directory returned by {@link ScopedContext#getTempDirectory()}.  If not set here,
+     * then the {@link Server#getTempDirectory()} is returned by {@link ScopedContext#getTempDirectory()}.</p>
+     * <p>If {@link #isTempDirectoryPersistent()} is true, the directory set here is used directly but may
+     * be created if it does not exist. If {@link #isTempDirectoryPersistent()} is false, then any ile set
```
janbartel:
```suggestion
     * be created if it does not exist. If {@link #isTempDirectoryPersistent()} is false, then any file set
```

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextRequest.java
```diff
@@ -224,14 +224,14 @@ public Charset getQueryEncoding()
...
     {
-        // return hidden attributes for request logging
-        // TODO does this actually work?   Does the request logger have the wrapped request?
         return switch (name)
         {
             case "o.e.j.s.s.ServletScopedRequest.request" -> _httpServletRequest;
             case "o.e.j.s.s.ServletScopedRequest.response" -> _response.getHttpServletResponse();
             case "o.e.j.s.s.ServletScopedRequest.servlet" -> _mappedServlet.getServletPathMapping(getPathInContext()).getServletName();
             case "o.e.j.s.s.ServletScopedRequest.url-pattern" -> _mappedServlet.getServletPathMapping(getPathInContext()).getPattern();
+            case __MULTIPART_CONFIG_ELEMENT -> (_mappedServlet.getServletHolder().getRegistration() instanceof ServletHolder.Registration registration)
```
janbartel: Nothing to do with temp directory??

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -787,7 +884,7 @@ public void setBaseResourceAsPath(Path path)
         if (path == null)
         {
             // allow user to unset variable
-            setBaseResource((Resource)null);
+            setBaseResource(null);
```
janbartel, line 896: Not relevant, but as it's small ok.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletHolder.java
```diff
@@ -709,14 +708,6 @@ protected void prepare(ServletRequest request, ServletResponse response) throws
     {
         // Ensure the servlet is initialized prior to any filters being invoked
         getServlet();
-
```
janbartel: Why is this deleted? Nothing to do with temp directory?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -122,6 +123,9 @@ public static String getServerInfo()
     private boolean _allowNullPathInContext;
     private Index<ProtectedTargetType> _protectedTargets = Index.empty(false);
     private final List<AliasCheck> _aliasChecks = new CopyOnWriteArrayList<>();
+    private File _tempDirectory;
+    private boolean _tempDirectoryPersisted = false;
+    private boolean _tempDirectoryConfigured = false;
```
janbartel: I don't see a getter for this field?  Also, is this field performing the same function as the context attribute that's being checked at line 88 in `WebInfConfiguration`?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java
```diff
@@ -721,6 +766,10 @@ private static class DynamicErrorProcessor extends ErrorProcessor {}
 
     class ServerContext extends Attributes.Wrapper implements Context
     {
+        private final File jettyBase = IO.asFile(System.getProperty("jetty.base"));
+        private final File workDir = jettyBase != null && jettyBase.isDirectory() && jettyBase.canWrite() ? new File(jettyBase, "work") : null;
```
janbartel, line 770: I think this is a behaviour change. If `jetty.base` exists, then the server's tmp directory will be `$jetty.base/work`, and I assume contexts will create their own tmp dirs inside of that, but they won't be persistent unless `context.setPersistTempDirectory(true)` has been called. In previous versions of jetty, any tmp `work` directory has always been automatically persisted.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -179,6 +183,60 @@ protected ScopedContext newContext()
         return new ScopedContext();
     }
 
+    @ManagedAttribute(value = "temporary directory location", readonly = true)
+    public File getTempDirectory()
```
janbartel, line 191: So this can be null? As sometimes methods called getTempDirectory() never return null, it would be good to javadoc when they do and when they don't.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -1076,6 +1173,15 @@ public Resource getBaseResource()
             return _baseResource;
         }
 
+        @Override
+        public File getTempDirectory()
+        {
+            File tempDirectory = ContextHandler.this.getTempDirectory();
```
janbartel, line 1188: There's a difference between tmp directory configured at the server level and one at the context level. At  the server level, it's the _parent_ directory of all context tmp dirs, but configured at the context level, it is the directory where the temporary files are written. With the code the way it is, how would each context get it's own separate directory inside the server's directory? What am I missing?

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebInfConfiguration.java
```diff
@@ -80,14 +84,6 @@ public void configure(WebAppContext context) throws Exception
     @Override
     public void deconfigure(WebAppContext context) throws Exception
     {
```
janbartel, line 86: Seems a little asymmetric that we're causing the directory to be created in `WebInfConfiguration.preconfigure`, but in `WebInfConfiguration.deconfigure` we do nothing and do it in `ContextHandler.doStop` instead. Are there any implications for the timing changes of when `WebIfConfigurion.deconfigure` is called vs when `ContextHandler.doStop` is called?

janbartel (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
gregw: No getter needed.
It is remembering if the temp directory has been setup (i.e. had contents deleted), so it doesn't delete again once WebInfConfiguration has unpacked stuff into it.
Renaming to `tempDirectoryCleaned` to be clearer.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
gregw: I don't think it is correct the say: "At the server level, it's the parent directory of all context tmp dirs".
The `WebInfConfiguration` will make context temporary dirs with the server temp directory as their parent, but that is only a convention/heutistic, otherwise:
 + A context might not have a configured temp directory, in which case the server temp directory can be directly used (if configured else `java.io.tempdir` is used).
 + a context might have a configured temp directory that is entirely unrelated to the temp director configured for the server.
 + a context might have a configured temp directory that is a child or descendant of any temp director configured for the server.

This is consistent with we do other configuration: if somebody explicitly sets some configuration, we put very few limits on what can be set, but if they don't set something then we use good conventions to come up with a reasonable value for them.

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/WebInfConfiguration.java
gregw: It is not asymmetric.  The `ContextHandler` always creates and deletes any temporary directory. All `WebInfConfiguration.preconfigure` does is ensure that it is created before it unpacks the webapp into it.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java
gregw: Well it is a change because servers nor contexts ever had temporary directories before. 
But I don't think it is a change for `WebAppContext`s as the `WebInfConfiguration` also checks for the "work" name in it's parent.

Potentially `ContextHandler` could also look for a "work" parent in any temporary directory set on it and force it to be persistent if found??? However that kind of breaks our pattern of putting the smart/heuristics into a Configuration and the base classes are just pretty dumb.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
janbartel: The `WebInfConfiguration` approach is safest and ensures that multiple contexts don't overwrite each other. There is no such protection with multiple `Contexts`, but I guess as there isn't a predefined structure for on-disk info for `Contexts` like there is with `WebAppContexts` its fine to rely on caveat emptor.

janbartel (approved):

PR #9159, opened 2023-01-12
Jetty 12 simplify Retainable.canRetain usage

gregw:
I believe that `canRetain` is no longer needed. The only place it is used for real logic is in `AsyncContent`, which can use it's own chunk class to pass on how to succeed the callback.

### DIFF
--- a/jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
+++ b/jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
@@ -436,11 +436,50 @@ public class Content
         /**
          * <p>An empty, non-last, chunk.</p>
          */
-        Chunk EMPTY = ByteBufferChunk.EMPTY;
+        Chunk EMPTY = new Chunk()
+        {
+            @Override
+            public ByteBuffer getByteBuffer()
+            {
+                return BufferUtil.EMPTY_BUFFER;
+            }
+
+            @Override
+            public boolean isLast()
+            {
+                return false;
+            }
+
+            @Override
+            public String toString()
+            {
+                return "EMPTY";
+            }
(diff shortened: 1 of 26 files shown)


gregw:
@sbordet @lorban the main change in this PR is in `AsyncContent`, everything else is just a mechanical removal of `canRetain`.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -502,12 +541,9 @@ static Chunk from(ByteBuffer byteBuffer, boolean last, Consumer<ByteBuffer> rele
          * @param retainable the Retainable this Chunk links to
          * @return a new Chunk
          * @throws IllegalArgumentException if the {@code Retainable}
-         * {@link Retainable#canRetain() cannot be retained}
          */
         static Chunk asChunk(ByteBuffer byteBuffer, boolean last, Retainable retainable)
         {
-            if (!retainable.canRetain())
```
lorban: This is the most important spot where `canRetain` is used. We must not allow linking the lifecycle of a Chunk that potentially contains a pooled buffer to the lifecycle of non-retaining retainables like EOF or EMPTY.

While this check is a nice to have extra safeguard and the current code would work without it, it was added because that error is way too easy to make and hard to track down.

lorban (changes requested):
I've added a comment with a quick summary of our discussion around `Content.asChunk` and `canRetain` for future reference.

lorban (approved):
LGTM

gregw:
@sbordet @lorban rebased against HEAD of jetty-12.0.x after merge of #9073

sbordet (approved):

Issue #9145, opened 2023-01-09
Failure when running `add-module` for openid, websocket, and stats modules


gregpoulos:
**Jetty version(s)**
12

**Java version/vendor** `(use: java -version)`
openjdk version "19.0.1" 2022-10-18
OpenJDK Runtime Environment (build 19.0.1+10-21)
OpenJDK 64-Bit Server VM (build 19.0.1+10-21, mixed mode, sharing)

**OS type/version**
macOS Monterey, 12.6.2

**Description**
I get error messages when trying to add the `openid` and `stats` modules:
```
$ java -jar $JETTY_HOME/start.jar --add-modules=openid
ERROR : No module found to provide security for openid{enabled}

Usage: java -jar $JETTY_HOME/start.jar [options] [properties] [configs]
       java -jar $JETTY_HOME/start.jar --help  # for more information
```

```
$ java -jar $JETTY_HOME/start.jar --add-modules=stats
ERROR : No module found to provide servlet for stats{enabled}

Usage: java -jar $JETTY_HOME/start.jar [options] [properties] [configs]
       java -jar $JETTY_HOME/start.jar --help  # for more information
```

This is confusing, because both `openid` and `stats` appear in the list of modules that appears when I run `java -jar $JETTY_HOME/start.jar --list-modules=*`:

```
$ java -jar $JETTY_HOME/start.jar --list-modules=* | grep openid
                           ee10-openid - Adds OpenId Connect authentication to the server.
                            ee9-openid - Adds OpenId Connect authentication to the server.
                                openid - Adds OpenId Connect authentication to the server.
$ java -jar $JETTY_HOME/start.jar --list-modules=* | grep stats
                                 stats - Enables detailed statistics collection for the server.
```

**How to reproduce?**
Run `java -jar $JETTY_HOME/start.jar --add-modules=openid,stats`

lorban:
The `stats` module bug is a bit complex and will need some time to get fixed, but the `openid` one looks fairly trivial.

The `openid` module should have been called `ee8-openid`. You could fix it temporarily by renaming `$JETTY_HOME/modules/openid.mod` to `$JETTY_HOME/modules/ee8-openid.mod` and editing it, you'll see that it contains a `[depends]` section like the following that needs to be changed:

```
[depend]
security
client
```

The `security` module doesn't exist anymore as it's specific to which ee version you're running so it should be changed to:

```
[depend]
ee8-security
client
```

[...]

PR #9160, opened 2023-01-12
Jetty 12.0.x fix openid and websocket modules

lorban:
Partial fix for #9145.

Fixes both `openid` and `websocket` modules, but not the `stats` one.


sbordet (approved):

sbordet:
@lorban LGTM, but can you add a minimal `DistributionTest` that proves the modules are now fine?

sbordet (changes requested):
Can you please add a minimal `DistributionTest` that proves these changes are correct?

joakime (approved):
Seems pretty straight forward.  

Approved.

lachlan-roberts:
@lorban this is now fixed for ee9 and ee10, but for ee8 I can still see duplication of the jars between `lib` and `lib/ee8-websocket/`

also no idea why this is happening but if I do a `--list-modules` i see this the `websocket` module, but looks like you removed it in this PR.

lorban:
@lachlan-roberts I figured why jars were duplicated for ee8, the jetty-home pom was simply lacking ee8-related exclusions in  the resource copying phase.

I'm also puzzled as why a `websocket` module still persists while I removed it.

I also noticed that `DynamicListenerTests` is testing websockets but is currently disabled. I've re-enabled it and dusted it off to fix the incorrect module and package references but it refuses to work because it cannot find the classes that are in the security module. Investigating that too...

lorban:
@olamy I'm facing a build problem when maven creates the jetty-home archive.

In `jetty-home/pom.xml`, `maven-dependency-plugin`'s `unpack-config-deps` execution eventually resolves `jetty-websocket-core-server-12.0.0-SNAPSHOT-config.jar` and unpacks it into the target assembly-directory.

I can't figure out where this websocket-core-server artifact with config classifier is coming from, and it contains a single file `modules/websocket.mod` that I have deleted in this PR but keeps coming back. Could you please help me figuring out what's going on?

lorban:
@lachlan-roberts @olamy I figured both problem, thanks to @joakime.

The `websocket.mod` is going to be included in jetty-home for as long as a build without it hasn't been published in our maven snapshot repo. So instead of deleting it, I've edited it so that it's totally empty then I'll commit its deletion once a build with that empty mod file has been published.

Regarding the `DynamicListener` test, it was a matter of mismatched environments: the class that was not found was part of the ee10 environment while the ini file that referred to it was in the core environment. The solution was simply to get rid of the ini file as this is now part of the ee10-demo module, so simply deploying that one solves the problem.

joakime:
> The `websocket.mod` is going to be included in jetty-home for as long as a build without it hasn't been published in our maven snapshot repo. So instead of deleting it, I've edited it so that it's totally empty then I'll commit its deletion once a build with that empty mod file has been published.

What he means here ...

The `websocket.mod` was previously found in `org.eclipse.jetty.websocket:jetty-websocket-core-server:12.0.0-SNAPSHOT:config:jar`

@lorban deleted the `websocket.mod` file, and then removed the whole `-config.jar` from even being created (there was nothing left to put in it)

But that left the old `-config.jar` file present in the maven local repo.

```
$ ls $HOME/.m2/repository/org/eclipse/jetty/websocket/jetty-websocket-core-server/12.0.0-SNAPSHOT/
total 100
drwxrwxr-x 2 joakim joakim  4096 Jan 26 09:18 .
drwxrwxr-x 3 joakim joakim  4096 Aug 16 12:28 ..
-rw-rw-r-- 1 joakim joakim   612 Jan 24 09:19 jetty-websocket-core-server-12.0.0-SNAPSHOT-config.jar
-rw-rw-r-- 1 joakim joakim 45449 Jan 26 09:19 jetty-websocket-core-server-12.0.0-SNAPSHOT.jar
-rw-rw-r-- 1 joakim joakim   978 Sep 15 10:49 jetty-websocket-core-server-12.0.0-SNAPSHOT.pom
-rw-rw-r-- 1 joakim joakim 29252 Jan 26 09:19 jetty-websocket-core-server-12.0.0-SNAPSHOT-sources.jar
-rw-rw-r-- 1 joakim joakim  1154 Jan 26 09:19 maven-metadata-local.xml
-rw-rw-r-- 1 joakim joakim   355 Jan 26 09:19 _remote.repositories
```

And that's where the `websocket.mod` kept coming back from.

So I advised him to temporarily restore the `-config.jar`, but leave it essentially blank.
Then after his PR has been merged, and the oss.sonatype.org has a copy of this empty `-config.jar`, we can remove that `-config.jar` from being created at all.

olamy (approved):

tests/test-distribution/test-distribution-common/src/test/java/org/eclipse/jetty/tests/distribution/ModulesTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.slf4j.LoggerFactory;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+public class ModulesTest
+{
+    private static class LogDisabler implements Closeable
```
sbordet, line 31: Could you not use `StackessLogging`?

lorban:
#9215 was opened as a reminder that `websocket.mod` still has to be removed.

sbordet (changes requested):

tests/test-distribution/test-distribution-common/src/test/java/org/eclipse/jetty/tests/distribution/ModulesTest.java
lorban: No because the stack trace is generated by a spawned process, and that process' output is just pumped to `LOGGER.info(...) calls`. But I can generalize that class and make it a sibling of `StackessLogging`.

sbordet (approved):

PR #9163, opened 2023-01-12
MultiPart API improvements for the Servlet implementation.

sbordet:
* Renamed Part.getContent() to getContentSource().
* Introduced Part.newContentSource().

Apparently, the Servlet implementation needs Part.getInputStream() to return a new InputStream every time, so we need a way to reproduce the part's content. This is now possible with Part.newContentSource(), which can be implemented easily for the parts that have a reproducible content (all apart one).


sbordet:
@lachlan-roberts let's base your work on top of this PR that improves the core MultiPart APIs so that you can more easily implement the Servlet MultiPart APIs.

gregw:
@sbordet @lachlan-roberts is this PR still live, or does @lachlan-roberts work replace it?

gregw:
Closing since I believe this has been replaced @lachlan-roberts  @sbordet ?.   Re-open if I got this wrong.

PR #9171, opened 2023-01-13
Jetty 12 - Fix for HttpParser quick HTTP Version lookahead

### LINKED ISSUE #9170: Jetty 12 - HttpParser Version optimization never works
**Jetty version(s)**
Jetty 12

**Description**

There exists and optimization in HttpParser > parseLine > status=SPACE2 that never hits due to position of the buffer.

https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java#L824-L829

https://github.com/eclipse/jetty.project/blob/7c7a7f3c39cee9316c8e352974d3b87603bdb575/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java#L824-L829

The token / buffer at this point in time is ...

```
HttpTokens.Token = ALPHA("H")
BufferUtil.toDetailString(buffer) = DirectByteBuffer@1456af41[p=8,l=87,c=8192,r=79]={POST / H<<<TTP/1.1\r\nHost: localhost\r...ent-Length: 10\r\n\r\n12345>>>\x00\x00\x00\x00\x00\x00\x00\x00\x00...\x00\x00\x00\x00\x00\x00\x00}
```

The call to `HttpVersion.CACHE.getBest()` (at line 829) is starting at the `"T"` and never returns a value.

joakime:
Fixes #9170


gregw:
@joakime I further improved this fix and pushed 593cfdb to your branch:
+ Removed the array optimized look ahead
+ Added caches for version follow by eoln and space
+ version cache hits now include following character and skip state transition.
+ improved test harness to check lf and crlf endings

gregw:
Added another optimization to check for HTTP as a single get of an int.

gregw:
@joakime I found a few more issues with the field cache, plus I went a simpler route for the version lookahead using getInt

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/HttpParserTest.java
```diff
@@ -170,10 +168,11 @@ public void testLineParse1()
...
     }
 
-    @Test
-    public void testLineParse2RFC2616()
+    @ParameterizedTest
+    @ValueSource(strings = {"\r\n", "\n"})
+    public void testLineParse2RFC2616(String eoln)
     {
-        ByteBuffer buffer = BufferUtil.toBuffer("POST /222  \r\n");
+        ByteBuffer buffer = BufferUtil.toBuffer("POST /222  " + eoln);
```
joakime: Did you mean to do 2 spaces here?

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/HttpParserTest.java
```diff
@@ -187,10 +186,11 @@ public void testLineParse2RFC2616()
...
     }
 
-    @Test
-    public void testLineParse2()
+    @ParameterizedTest
+    @ValueSource(strings = {"\r\n", "\n"})
+    public void testLineParse2(String eoln)
     {
-        ByteBuffer buffer = BufferUtil.toBuffer("POST /222  \r\n");
+        ByteBuffer buffer = BufferUtil.toBuffer("POST /222  " + eoln);
```
joakime: Did you want 2 spaces here?

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/HttpParserTest.java
```diff
@@ -978,8 +1065,9 @@ public void testChunkParse()
         assertTrue(_messageCompleted);
     }
 
-    @Test
-    public void testBadChunkLength()
+    @ParameterizedTest
+    @ValueSource(strings = {"\r\n", "\n"})
+    public void testBadChunkLength(String eoln)
```
joakime, line 1095: `eoln` not used here

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/HttpParserTest.java
```diff
@@ -1008,8 +1096,9 @@ public void testBadChunkLength()
         assertFalse(_messageCompleted);
     }
 
-    @Test
-    public void testBadTransferEncoding()
+    @ParameterizedTest
+    @ValueSource(strings = {"\r\n", "\n"})
+    public void testBadTransferEncoding(String eoln)
```
joakime, line 1126: `eoln` not used here

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/HttpParserTest.java
```diff
@@ -947,8 +1033,9 @@ public void testSplitHeaderParse()
         }
     }
 
-    @Test
-    public void testChunkParse()
+    @ParameterizedTest
+    @ValueSource(strings = {"\r\n", "\n"})
+    public void testChunkParse(String eoln)
```
joakime, line 1064: `eoln` not used here

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/HttpParserTest.java
```diff
@@ -1169,8 +1262,9 @@ public void testChunkParseNoTrailer()
         assertTrue(_messageCompleted);
     }
 
-    @Test
-    public void testStartEOF()
+    @ParameterizedTest
+    @ValueSource(strings = {"\r\n", "\n"})
+    public void testStartEOF(String eoln)
```
joakime: `eoln` not used here

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/HttpParserTest.java
```diff
@@ -3074,8 +3243,9 @@ public void onComplianceViolation(ComplianceViolation.Mode mode, ComplianceViola
         }
     }
 
-    @Test
-    public void testHttpHeaderValueParseCsv()
+    @ParameterizedTest
+    @ValueSource(strings = {"\r\n", "\n"})
+    public void testHttpHeaderValueParseCsv(String eoln)
```
joakime: `eoln` not used here

joakime (commented):
Not every place uses eoln

joakime:
You introduced a parameterized `eoln`, but it's not used in all test cases.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
```diff
@@ -164,12 +165,17 @@
...
+
     private static final Index.Mutable<HttpField> NO_CACHE = new Index.Builder<HttpField>()
         .caseSensitive(false)
         .mutable()
         .maxCapacity(0)
         .build();
 
+    public static final int HTTP_AS_INT = ('H' & 0xff) << 24 | ('T' & 0xFF) << 16 | ('T' & 0xFF) << 8 | ('P' & 0xFF);
+    public static final int HTTP_1_0_AS_INT = ('/' & 0xff) << 24 | ('1' & 0xFF) << 16 | ('.' & 0xFF) << 8 | ('0' & 0xFF);
+    public static final int HTTP_1_1_AS_INT = ('/' & 0xff) << 24 | ('1' & 0xFF) << 16 | ('.' & 0xFF) << 8 | ('1' & 0xFF);
```
sbordet: Make them `private`.
The chars are all ASCII, so no need to `& 0xFF`. But if you want to retain it, please use capital `0xFF` everywhere.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
```diff
@@ -454,15 +460,21 @@ private HttpTokens.Token next(ByteBuffer buffer)
...
                         _headerBytes++;
-                    return next(buffer);
+                    ch = buffer.get();
+                    t = HttpTokens.TOKENS[0xff & ch];
+                    return switch (t.getType())
+                    {
+                        case CNTL -> throw new IllegalCharacterException(_state, t, buffer);
+                        case LF -> t;
+                        default -> throw new BadMessageException("Bad EOL");
+                    };
```
sbordet, line 479: I don't understand this change?
Seems much better to retain the original recursive call, that seems to me equivalent, and reads better.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
```diff
@@ -489,24 +501,26 @@ private HttpTokens.Token next(ByteBuffer buffer)
...
             }
         }
-        else if (_responseHandler != null)
+        else if (_responseHandler != null && buffer.remaining() >= 9 && buffer.getInt(position) == HTTP_AS_INT)
         {
-            _version = HttpVersion.lookAheadGet(buffer);
-            if (_version != null)
+            int v = buffer.getInt(position + 4);
+            _version = v == HTTP_1_1_AS_INT ? HttpVersion.HTTP_1_1 : v == HTTP_1_0_AS_INT ? HttpVersion.HTTP_1_0 : null;
+            if (_version != null && buffer.get(position + 8) == ' ')
```
sbordet: I find it quite hard to read and reason about the buffer with these "magic" numbers, like `9`, `4`, `8`, etc.
Can it be rewritten using a `int current` to indicate where in the buffer we are looking at?
Something like:

```java
else if (_responseHandler != null && buffer.remaining() >= 9)
{
    int current = position;
    int h = buffer.getInt(current);
    if (h == HTTP_AS_INT)
    {
        current += 4;
        int v = buffer.getInt(current);
        ...
    }
```

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
```diff
@@ -762,9 +780,38 @@ private boolean parseLine(ByteBuffer buffer)
...
+                                    if (next == '\n')
+                                    {
+                                        buffer.position(p);
+                                        _version = version;
+                                        _string.setLength(0);
+                                        checkVersion();
+                                        _fieldCache.prepare();
+                                        setState(State.HEADER);
+                                        _requestHandler.startRequest(_methodString, _uri.toString(), _version);
+                                        continue;
```
sbordet, line 817: Ditto re: hard to read and follow with the magic numbers.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
```diff
@@ -762,9 +780,38 @@ private boolean parseLine(ByteBuffer buffer)
...
+                            {
+                                // try look ahead for request HTTP Version
+                                int v = buffer.getInt(position + 4);
+                                HttpVersion version = v == HTTP_1_1_AS_INT ? HttpVersion.HTTP_1_1 : v == HTTP_1_0_AS_INT ? HttpVersion.HTTP_1_0 : null;
+
+                                if (version != null)
+                                {
+                                    int p = position + 8;
+                                    byte next = buffer.get(p++);
+                                    if (next == '\r' && remaining >= 10)
```
sbordet: Style elsewhere would be `next == HttpTokens.CARRIAGE_RETURN`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
```diff
@@ -762,9 +780,38 @@ private boolean parseLine(ByteBuffer buffer)
...
+                                        next = buffer.get(p++);
+
+                                    if (next == '\n')
```
sbordet: `next == HttpTokens.LINE_FEED`?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
```diff
@@ -1233,9 +1243,9 @@ else if (_endOfContent == EndOfContent.UNKNOWN_CONTENT)
                             if (buffer.hasRemaining())
                             {
                                 // Try a look ahead for the known header name and value.
-                                HttpField cachedField = _fieldCache.getBest(buffer, -1, buffer.remaining());
+                                HttpField cachedField = _fieldCache.getBest(buffer, -1, buffer.remaining() + 1);
                                 if (cachedField == null)
-                                    cachedField = CACHE.getBest(buffer, -1, buffer.remaining());
+                                    cachedField = CACHE.getBest(buffer, -1, buffer.remaining() + 1);
```
sbordet, line 1256: I think it'll be much clearer if we do `buffer.position(buffer.position() - 1)`, and then normally call `cache.getBest(buffer, 0, buffer.remaining());`

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpTokens.java
```diff
@@ -183,5 +183,8 @@ else if (b >= 0x80) // OBS
             }
         }
     }
+
+    public static final Token CR = TOKENS['\r'];
+    public static final Token LF = TOKENS['\n'];
```
sbordet: Are they used at all?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
```diff
@@ -1267,25 +1277,24 @@ else if (_endOfContent == EndOfContent.UNKNOWN_CONTENT)
...
+                                    if (v == null || pos >= buffer.limit())
                                     {
                                         // Header only
                                         setState(FieldState.VALUE);
                                         _string.setLength(0);
                                         _length = 0;
-                                        buffer.position(buffer.position() + n.length() + 1);
+                                        buffer.position(v == null ? pos : pos - v.length());
                                         break;
                                     }
```
sbordet: Does not feel right.
`pos` may be set past the cached value, and we may still enter the `if` block due to `pos >= buffer.limit()`, but then we set the state to be `VALUE`?

Even if it is correct it's just way more difficult to read, so perhaps needs comments and/or examples.

sbordet (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
gregw: The recursive call was doing more options that are not appropriate and hence had to be guarded by setting, then testing and maybe clearing _cr.   This is much simpler, if the next charater is availbel and it is LF, then we have a LF otherwise anything after a CR is an exception of one sort or another.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
gregw: The problem is that if we miss the cache lookup, then we'd need to put the position back as we have already just parsed that byte... which is why we know we can do the -1.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
gregw: I've added some constants. Is that clearer?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
gregw: restructured to `posAfterName` and `posAfterValue`... but doing some more checks.

gregw: more cleanup and testing done. OK now.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
gregw: To be more specific, every single CRLF in a header was being handled by:
 + switch on CR
 + setting _cr true
 + recursing
 + switch on LF
 + clearing _cr.

Now we just:
 + switch on CR
 + switch on LF

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
```diff
@@ -489,24 +505,26 @@ private HttpTokens.Token next(ByteBuffer buffer)
...
     {
+        int position = buffer.position();
         if (_requestHandler != null)
         {
             _method = HttpMethod.lookAheadGet(buffer);
             if (_method != null)
             {
                 _methodString = _method.asString();
-                buffer.position(buffer.position() + _methodString.length() + 1);
+                buffer.position(position + _methodString.length() + 1);
```
sbordet, line 517: This seems unsafe.
If I get a buffer with just `POST` in it, I think here we set the position to `0 + 4 + 1 == 5`, which is beyond `limit()` and throws.
If that is not the case, deserves a comment about why not.

sbordet (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
gregw: OK, will look.... but this has not been changed by this PR, I only added a variable.

sbordet (approved):

PR #9172, opened 2023-01-16
Add latency recording utility handler 

### LINKED ISSUE #8069: Jetty 12 is missing a way to record server latencies
**Jetty version(s)**
12.0.x

**Description**
12.0.x currently does not have anything equivalent to 11's `HttpChannel.Listener` API which is needed at least to record server latencies with our performance tests.

lorban:
Add a small utility handler that standardizes the way latencies should be recorded.

Closes #8069.

### DIFF
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/AbstractLatencyRecordingHandler.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/AbstractLatencyRecordingHandler.java
@@ -1,0 +1,86 @@
+//
+// ========================================================================
+// Copyright (c) 1995-2022 Mort Bay Consulting Pty Ltd and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.server.handler;
+
+import java.util.function.Function;
+
+import org.eclipse.jetty.server.Handler;
+import org.eclipse.jetty.server.HttpStream;
+import org.eclipse.jetty.server.Request;
+import org.eclipse.jetty.server.Response;
+import org.eclipse.jetty.util.Callback;
+import org.eclipse.jetty.util.NanoTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A <code>Handler</code> that allows recording the latencies of the requests executed by the wrapped handler.
+ */
+public abstract class AbstractLatencyRecordingHandler extends Handler.Wrapper
+{
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractLatencyRecordingHandler.class);
+
+    private final Function<HttpStream, HttpStream> recordingWrapper;
+
+    public AbstractLatencyRecordingHandler()
+    {
+        this.recordingWrapper = httpStream -> new HttpStream.Wrapper(httpStream)
+        {
+            @Override
+            public void succeeded()
+            {
+                long begin = httpStream.getNanoTime();
+                super.succeeded();
+                try
+                {
+                    onRequestComplete(NanoTime.since(begin));
+                }
+                catch (Throwable t)
+                {
+                    if (LOG.isDebugEnabled())
+                        LOG.debug("Error thrown by onRequestComplete", t);
+                }
+            }
+
+            @Override
+            public void failed(Throwable x)
+            {
+                long begin = httpStream.getNanoTime();
+                super.failed(x);
+                try
+                {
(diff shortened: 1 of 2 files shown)


joakime:
Got an example of how it would be used?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/AbstractLatencyRecordingHandler.java
```diff
@@ -0,0 +1,86 @@
...
+import org.eclipse.jetty.server.HttpStream;
+import org.eclipse.jetty.server.Request;
+import org.eclipse.jetty.server.Response;
+import org.eclipse.jetty.util.Callback;
+import org.eclipse.jetty.util.NanoTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A <code>Handler</code> that allows recording the latencies of the requests executed by the wrapped handler.
```
gregw: I think the javadoc needs to be more specific about which latency is being measured.  This is the latency from the first notice of the request (obtained from the stream timestamp)  until after all processing has completed and all serialized methods have returned and the stream completion even has been handled (stream completion).

This is different from the handling latency, which is from when the handler is called until the passed callback is completed.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/AbstractLatencyRecordingHandler.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+public abstract class AbstractLatencyRecordingHandler extends Handler.Wrapper
+{
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractLatencyRecordingHandler.class);
+
+    private final Function<HttpStream, HttpStream> recordingWrapper;
```
gregw: why not just have a private method and use a method reference?

gregw (approved):
Just some niggles that you can fix if you like, otherwise good.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/AbstractLatencyRecordingHandler.java
lorban: Because that would mean creating a new `Function<HttpStream, HttpStream>` instance per request to pass it to `addHttpStreamWrapper`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/AbstractLatencyRecordingHandler.java
lorban: I added some javadoc more precisely describing what we mean with 'latency' in this context. Hopefully it's clear enough.

lorban:
@joakime The fact that you made this comment means a test is needed, at least as some internal doc showing how this works. I've now added such test.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/AbstractLatencyRecordingHandler.java
```diff
@@ -0,0 +1,84 @@
...
+            }
+        };
+    }
+
+    @Override
+    public boolean process(Request request, Response response, Callback callback) throws Exception
+    {
+        request.addHttpStreamWrapper(recordingWrapper);
+        return super.process(request, response, callback);
+    }
```
gregw: I don't think this allocates.
```suggestion
    public AbstractLatencyRecordingHandler()
    {
    }
    
    private HttpStream recordingWrapper(HttpStream httpStream)
    {
        return new HttpStream.Wrapper(httpStream)
        {
            @Override
            public void succeeded()
            {
                super.succeeded();
                fireOnRequestComplete();
            }

            @Override
            public void failed(Throwable x)
            {
                super.failed(x);
                fireOnRequestComplete();
            }

            private void fireOnRequestComplete()
            {
                try
                {
                    onRequestComplete(NanoTime.since(httpStream.getNanoTime()));
                }
                catch (Throwable t)
                {
                    if (LOG.isDebugEnabled())
                        LOG.debug("Error thrown by onRequestComplete", t);
                }
            }
        };
    }

    @Override
    public boolean process(Request request, Response response, Callback callback) throws Exception
    {
        request.addHttpStreamWrapper(this::recordingWrapper);
        return super.process(request, response, callback);
    }
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/AbstractLatencyRecordingHandler.java
gregw: I don't think so. See below.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/AbstractLatencyRecordingHandler.java
lorban: You're so right, what was I thinking?

gregw (approved):
that's better!

PR #9174, opened 2023-01-16
Jetty 12 - Introduce `GracefulShutdownHandler` and Test

joakime:
+ started with removing `@Disabled` from `GracefulStopTest.java`


jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
```diff
@@ -0,0 +1,125 @@
...
+            }
+            return true;
+        }
+
+        try
+        {
+            dispatchedStats.increment();
+
+            // TODO: what to do if wrapped handler uses request.demand(Runnable) ??
+            // TODO: should we wrap the incoming Callback to handle decrement logic better??
```
joakime: @sbordet @gregw Do I need to worry about these?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
```diff
@@ -0,0 +1,125 @@
...
+import org.eclipse.jetty.util.statistic.CounterStatistic;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Handler to track active requests and allow them to gracefully complete.
+ */
+public class GracefulShutdownHandler extends Handler.Wrapper implements Graceful
+{
+    private static final Logger LOG = LoggerFactory.getLogger(GracefulShutdownHandler.class);
```
sbordet, line 33: Add newline after to separate from non-static.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
```diff
@@ -0,0 +1,125 @@
...
+                    LOG.debug("Service Unavailable (before wrapped process): {}", request.getHttpURI());
+                response.getHeaders().put(HttpHeader.CONNECTION, HttpHeaderValue.CLOSE);
+                Response.writeError(request, response, callback, HttpStatus.SERVICE_UNAVAILABLE_503);
+            }
+            return true;
+        }
+
+        try
+        {
+            dispatchedStats.increment();
```
sbordet: There is a race between `isShutdown()` and this increment.
A thread may see `isShutdown()==false`, but then another thread can shutdown and the `CompletableFuture` will be completed because `dispatchedStats` has not been incremented yet.
Maybe increment `dispatchedStats` before checking for `isShutdown()` is enough?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
```diff
@@ -0,0 +1,125 @@
...
+    {
+        Handler handler = getHandler();
+        if (handler == null || !isStarted() || isShutdown())
+        {
+            if (!response.isCommitted())
+            {
+                if (LOG.isDebugEnabled())
+                    LOG.debug("Service Unavailable (before wrapped process): {}", request.getHttpURI());
+                response.getHeaders().put(HttpHeader.CONNECTION, HttpHeaderValue.CLOSE);
+                Response.writeError(request, response, callback, HttpStatus.SERVICE_UNAVAILABLE_503);
```
sbordet: This is a request/response that needs to be counted and waited.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
```diff
@@ -0,0 +1,125 @@
...
+
+        try
+        {
+            dispatchedStats.increment();
+
+            // TODO: what to do if wrapped handler uses request.demand(Runnable) ??
+            // TODO: should we wrap the incoming Callback to handle decrement logic better??
+
+            // Count incoming request
+            return super.process(request, response, callback);
```
sbordet: If `super.process()` returns false, decrement immediately, otherwise you should always wrap the callback to decrement.

sbordet (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
joakime: Changed to use Callbacks instead.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
joakime: Alternate approach (using Callbacks) now pushed

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
joakime: Alternate approach pushed.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
joakime: Fixed in alternate approach that was pushed.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
```diff
@@ -0,0 +1,161 @@
...
+import org.eclipse.jetty.util.Callback;
+import org.eclipse.jetty.util.component.Graceful;
+import org.eclipse.jetty.util.statistic.CounterStatistic;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Handler to track active requests and allow them to gracefully complete.
+ */
+public class GracefulShutdownHandler extends Handler.Wrapper implements Graceful
```
gregw, line 31: Ultimately the stats handle might extend this handler ?  Or if not, the stats module could provide graceful, so that this handler is not needed if there is a stats handler.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
```diff
@@ -0,0 +1,161 @@
...
+        {
+            decrement();
+
+            Callback callback = Callback.from(getCallback(), shutdown::check);
+
+            if (isShutdown())
+            {
+                if (LOG.isDebugEnabled())
+                    LOG.debug("Service Unavailable: {}", request.getHttpURI(), x);
+                Response.writeError(request, response, callback, HttpStatus.SERVICE_UNAVAILABLE_503);
```
gregw: I'm questioning this now.  If an in progress request completes with an exception, should we not just write the error response normally to report that problem and decrement the count?   No need to swap in a 503 response... that is for any subsequent requests that arrive after shutdown.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
```diff
@@ -0,0 +1,161 @@
...
+    public boolean process(Request request, Response response, Callback callback) throws Exception
+    {
+        Handler handler = getHandler();
+        ShutdownTrackingCallback shutdownCallbackTracking = new ShutdownTrackingCallback(request, response, callback);
+        if (handler == null || !isStarted() || isShutdown())
+        {
+            // always return 503
+            if (serviceUnavailable(request, response, shutdownCallbackTracking));
+                return true;
+        }
```
gregw: I don't  think we need to increment/decrement if we are already stopped or shutdown:

```suggestion
        if (handler == null || !isStarted() || isShutdown())
        {
            // always return 503
            if (serviceUnavailable(request, response, callback));
                return true;
        }
        ShutdownTrackingCallback shutdownCallbackTracking = new ShutdownTrackingCallback(request, response, callback);

```

gregw (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
joakime: This change will prevent `shutdown.check()` behavior while in `isShutdown() == true`.
Is that acceptable?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
joakime: A few scenarios ...

Note: "When" can be ...

* Immediate : it happens immediately.
* Graceful : during Graceful Shutdown, not yet a Timeout.
* Timeout : when the connection IdleTimeout or Shutdown Timeout occurs.

Scenario | When | Result
---- | ---- | ---
A new request (entering `process()` after shutdown) | Immediate | a rapid "503 Service Unavailable" response.
An active request (`process()` returned true), but uncommitted | Timeout | "503 Service Unavailable" response.
An active request (`process()` not returned), but uncommitted | Timeout | "503 Service Unavailable" response.
An active request (`process()` returned true), committed | Timeout | Close connection
An active request (`process()` not returned), committed | Timeout | Close connection
An active request, using `writeError()` | Graceful | Don't change the error status
An active request, throwing uncaught Exception | Graceful | Use `writeError(Throwable)`
An active request, completing normally | Graceful | Decrement (do nothing extra)

Am I missing anything?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
joakime: Having it in 2 places seems confusing (for users).

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
joakime: Actually ... shouldn't the error response be also part of the graceful shutdown?
In other words, shouldn't we let those finish before we allow the Connector or Endpoint to shutdown?
If so, then we need the counting here too.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
sbordet: I would not like the `StatisticsHandler` to depend on this one, just to reuse 1 field only.
We change the logic in one of the 2 we break the other, so separated is best IMHO.
Also, the logic is different.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
```diff
@@ -0,0 +1,161 @@
...
+ * Handler to track active requests and allow them to gracefully complete.
+ */
+public class GracefulShutdownHandler extends Handler.Wrapper implements Graceful
+{
+    private static final Logger LOG = LoggerFactory.getLogger(GracefulShutdownHandler.class);
+
+    private CounterStatistic dispatchedStats = new CounterStatistic();
+    private Shutdown shutdown;
+
+    class ShutdownTrackingCallback extends Callback.Nested
```
sbordet: Make it `private` and move to end of file, like everywhere else.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
```diff
@@ -0,0 +1,161 @@
...
+
+        @Override
+        public void succeeded()
+        {
+            decrement();
+            super.succeeded();
+            if (isShutdown())
+            {
+                shutdown.check();
+            }
```
sbordet: No braces.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
sbordet: @gregw your code has a race.

Thread1 enters, sees `isShutdown()==false`, but does not increment yet the counter.
Thread2 calls `shutdown()`, and the `CompletableFuture` is immediately completed because the counter is `0`, but that is wrong because Thread1 is now processing the request.

@joakime's version is first incrementing the counter, and then checking for `isShutdown()`, which avoids the race.
However, it deserves a comment about why the increment is always done as the first thing.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
sbordet: @gregw we need to call `Shutdown.check()`, so a possible solution is:

```java
decrement();
super.failed(x);
if (isShutdown())
    shutdown.check();
```

This is symmetric with `succeeded()`.

However, if the error response is asynchronous, would be ok to complete the `CompletableFuture` while the response is still being written?
The problem would be that if the handler's `CompletableFuture` is completed, then we are entitled to close the connections, but the response is still being written.

sbordet (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
```diff
@@ -0,0 +1,143 @@
...
+                if (LOG.isDebugEnabled())
+                    LOG.debug("isShutdownDone: count {}", count);
+                return count == 0;
+            }
+        };
+    }
+
+    /**
+     * Test during Graceful shutdown to see if we are done with all
+     * graceful actions.
```
sbordet: Javadoc is wrong. `isShutdown()` only reports whether the shutdown has been initiated, not if it is finished.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
```diff
@@ -0,0 +1,143 @@
...
+     *
+     * @return true if graceful is active
```
sbordet:
```suggestion
     * @return whether the graceful shutdown has been initiated
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
```diff
@@ -0,0 +1,143 @@
...
+    @Override
+    public boolean isShutdown()
+    {
+        return shutdown.isShutdown();
+    }
+
+    @Override
+    public boolean process(Request request, Response response, Callback callback) throws Exception
+    {
+        // Increment the counter always, before we test for isShutdown to avoid race.
```
sbordet:
```suggestion
        // Increment the counter before the test for isShutdown(), to avoid race conditions.
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
```diff
@@ -0,0 +1,143 @@
...
+            if (isShutdown())
+                shutdown.check();
+        }
+    }
+
+    @Override
+    public CompletableFuture<Void> shutdown()
+    {
+        if (LOG.isInfoEnabled())
+            LOG.info("Shutdown requested");
```
sbordet: This should be DEBUG logging.

sbordet (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
joakime: `server.stop()`, in general, isn't DEBUG, it's INFO.
That's why I chose INFO.

sbordet (approved):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
gregw: Agreed. Shutdown and stats are looking at different latencies anyway.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
```diff
@@ -0,0 +1,143 @@
...
+        ShutdownTrackingCallback shutdownCallback = new ShutdownTrackingCallback(request, response, callback);
+
+        Handler handler = getHandler();
+        if (handler == null || !isStarted() || isShutdown())
+        {
+            if (LOG.isDebugEnabled())
+                LOG.debug("Service Unavailable: {}", request.getHttpURI());
+            Response.writeError(request, response, shutdownCallback, HttpStatus.SERVICE_UNAVAILABLE_503);
+            return true;
+        }
```
gregw: This still does not feel correct.  If we are not started, then we should not be processing the request, even for a 503. 

```suggestion
        Handler handler = getHandler();
        if (handler == null || !isStarted())
            return false;
        
        // Increment the counter before the test for isShutdown(), to avoid race conditions.
        ShutdownTrackingCallback shutdownCallback = new ShutdownTrackingCallback(request, response, callback);

        if (isShutdown())
        {
            if (LOG.isDebugEnabled())
                LOG.debug("Service Unavailable: {}", request.getHttpURI());
            Response.writeError(request, response, shutdownCallback, HttpStatus.SERVICE_UNAVAILABLE_503);
            return true;
        }
```

The `handler==null` case can be before or after.

gregw (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
gregw: The intention was that thread1 will never actually enter the processing of the request.   See better version in latest review.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulShutdownHandler.java
joakime: Changed.

joakime:
Before I dive into the jetty-start module for this ...

I'm wondering if this is the right name for this handler?
It doesn't provide GracefulShutdown, it manages the requests during graceful shutdown.
So how about `GracefulRequestHandler` instead?

sbordet (approved):

PR #9175, opened 2023-01-16
Issue #9173 - Make wrapping of `ServletApiResponse` easier

### LINKED ISSUE #9173: Configuring SameSite on a per-cookie basis in Jetty 12
**Jetty version**

12.0.0.alpha3

**Java version**

17

**Question**

We'd like to support Jetty 12 in Spring Boot 3.1 and I am [investigating the changes](https://github.com/wilkinsona/spring-boot/tree/jetty-12) that we'll need to make. It's going pretty well but I've hit a problem with cookies and `SameSite`.

Spring Boot allows users to implement [a strategy interface](https://docs.spring.io/spring-boot/docs/3.0.x/api/org/springframework/boot/web/servlet/server/CookieSameSiteSupplier.html) that can supply a cookie's `SameSite` value on a per-cookie basis. With Jetty 11 and earlier, we've implemented this functionality using [a `HandlerWrapper` sub-class](https://github.com/spring-projects/spring-boot/blob/3867c2d79cd9545dd71764a07da3409bd62f0d6c/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/JettyServletWebServerFactory.java#L670-L727). Can you please give me some pointers on how this can be achieved using `Handler.Wrapper` in Jetty 12? Thus far, I have been unable to figure out how to intercept [the addition of a cookie](https://github.com/spring-projects/spring-boot/blob/3867c2d79cd9545dd71764a07da3409bd62f0d6c/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/JettyServletWebServerFactory.java#L695-L705) with the new API.

joakime:
This is an effort in ee10 to allow wrapping of `ServletApiResponse` easier to support the spring-boot 6 use case.

* Moved `ServletContextRequest.ServletApiRequest` to top level class.
* Moved `ServletContextResponse.ServletApiResponse` to top level class.
* Added `ServletContextHandler.newServletContextRequest()` method to allow overrides.


jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
```diff
@@ -0,0 +1,1209 @@
...
+import org.eclipse.jetty.session.Session;
+import org.eclipse.jetty.session.SessionManager;
+import org.eclipse.jetty.util.Fields;
+import org.eclipse.jetty.util.HostPort;
+import org.eclipse.jetty.util.StringUtil;
+import org.eclipse.jetty.util.URIUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ServletApiRequest implements HttpServletRequest
```
gregw, line 88: javadoc needed now (always was)

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiResponse.java
```diff
@@ -0,0 +1,660 @@
...
+import org.eclipse.jetty.server.Request;
+import org.eclipse.jetty.server.Response;
+import org.eclipse.jetty.session.Session;
+import org.eclipse.jetty.session.SessionManager;
+import org.eclipse.jetty.util.Blocker;
+import org.eclipse.jetty.util.FutureCallback;
+import org.eclipse.jetty.util.StringUtil;
+import org.eclipse.jetty.util.URIUtil;
+
+public class ServletApiResponse implements HttpServletResponse
```
gregw, line 61: javadoc needed now (always was)

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiResponse.java
```diff
@@ -0,0 +1,660 @@
...
+{
+    private static final int MIN_BUFFER_SIZE = 1;
+    private static final EnumSet<ServletContextResponse.EncodingFrom> LOCALE_OVERRIDE = EnumSet.of(ServletContextResponse.EncodingFrom.NOT_SET, ServletContextResponse.EncodingFrom.DEFAULT, ServletContextResponse.EncodingFrom.INFERRED, ServletContextResponse.EncodingFrom.SET_LOCALE);
+    private final ServletContextResponse _response;
+
+    ServletApiResponse(ServletContextResponse response)
```
gregw: should this now be protected rather than package scope?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
```diff
@@ -0,0 +1,1209 @@
...
+    private Session _coreSession;
+    private String _requestedSessionId;
+    private boolean _requestedSessionIdFromCookie;
+    private Authentication _authentication;
+    private String _method;
+    private ServletMultiPartFormData.Parts _parts;
+    private ServletPathMapping _servletPathMapping;
+    private boolean _asyncSupported = true;
+
+    public ServletApiRequest(ServletContextRequest servletContextRequest)
```
gregw: protected rather than public?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextHandler.java
```diff
@@ -1148,6 +1150,18 @@ protected void doStop() throws Exception
...
+                                                             Request request,
+                                                             Response response,
+                                                             String pathInContext,
+                                                             ServletHandler.MappedServlet mappedServlet,
+                                                             PathSpec pathSpec,
+                                                             MatchedPath matchedPath)
+    {
+        return new ServletContextRequest(_servletContext, servletChannel, request, response, pathInContext,
+            mappedServlet, pathSpec, matchedPath);
+    }
```
gregw: Maybe this is less fat:
```suggestion
    protected ServletContextRequest newServletContextRequest(ServletChannel servletChannel,
                                                             Request request,
                                                             Response response,
                                                             String pathInContext,
                                                             MatchedResource<ServletHandler.MappedServlet> matchedServlet)
    {
        return new ServletContextRequest(_servletContext, servletChannel, request, response, pathInContext,
            matchedServlet.getResource(), matchedServlet.getPathSpec(), matchedServlet.getMatchedPath());
    }
```

maybe even pass into constructor like that?

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/ServletContextHandlerTest.java
```diff
@@ -2368,4 +2373,100 @@ public void init() throws ServletException
...
+        {
+            super(response.getResponse());
+        }
+
+        @Override
+        public void addCookie(Cookie cookie)
+        {
+            // Let's set SameSite to STRICT
+            // But we cannot use Servlet jakarta.servlet.http.Cookie.setComment(String) technique of old
+            // as the Comment field is always null and is slated for removal.
```
gregw, line 2390: Let's remove the comment handling code if it is no longer usable.... or do we still need it for ee8/ee9?

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/ServletContextHandlerTest.java
```diff
@@ -2368,4 +2373,100 @@ public void init() throws ServletException
...
+            Host: localhost\r
+            Connection: close\r
+            \r
+            """;
+
+        String rawResponse = _connector.getResponse(rawRequest);
+        HttpTester.Response response = HttpTester.parseResponse(rawResponse);
+        String setCookieValue = response.get(HttpHeader.SET_COOKIE);
+        assertThat(setCookieValue, containsString("example=bogus; SameSite=Strict"));
+    }
```
gregw, line 2483: no EOLN on last line

gregw (changes requested):
Just a few niggles, but mostly good

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextHandler.java
joakime: `MatchedResource<ServletHandler.MappedServlet>` is not retained and usable for folks that are wrapping this class.

janbartel (commented):
If @gregw's comments are actioned, I'm good with this PR.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextHandler.java
gregw: I don't understand? `MatchedResource<ServletHandler.MappedServlet>` is a public class and contains a: ServletHandler.MappedServlet, PathSpec & MatchedPath, so why pass them separately when they can be passed simply in a single public class?

I think it definitely can be done and folks that wrap the class will get the info they need one way or another.   The only question is do we want to do it or not?   PRO: not so fat signature, CON: moderately verbose generic class.  I'm 60:40 for the thinner API.

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/ServletContextHandlerTest.java
joakime: I'll remove from ee10 only, as comment is only deprecated there.

joakime: Actually, this is turning out to be a big change in itself, I'll do the minimal here in this PR and follow it up with a the larger change in a different PR. (to avoid omnibus PRs)

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
```diff
@@ -127,6 +127,9 @@ public HttpCookie(String name, String value, String domain, String path, long ma
         _version = version;
         _expiration = maxAge < 0 ? -1 : NanoTime.now() + TimeUnit.SECONDS.toNanos(maxAge);
         _attributes = (attributes == null ? Collections.emptyMap() : attributes);
+
+        // remove attributes that conflict with other values here
+        List.of("Domain", "Path", "Max-Age", "HttpOnly", "Secure", "Comment").forEach(_attributes::remove);
```
gregw: make this list final static field rather than recreate in every constructor (Java might do that already... but let's not rely on it)

gregw: Note also that this is mutating the passed in map, which is not exactly best practice.  At least this should be javadoc'd.  Maybe we create new map if we need to mutate it? maybe we always create a new map?

gregw (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
joakime: It needs improvement in other ways as well.

```
2023-01-18 21:41:57.389:WARN :oejes.ServletChannel:qtp1687702287-501: /another.dump
java.lang.UnsupportedOperationException
	at java.base/java.util.Collections$UnmodifiableMap.remove(Collections.java:1508)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at org.eclipse.jetty.http.HttpCookie.<init>(HttpCookie.java:132)
	at org.eclipse.jetty.session.Session.generateSetCookie(Session.java:156)
	at org.eclipse.jetty.ee10.servlet.SessionHandler.getSessionCookie(SessionHandler.java:441)
	at org.eclipse.jetty.ee10.servlet.ServletApiRequest.getSession(ServletApiRequest.java:424)
	at org.eclipse.jetty.ee10.demos.DumpServlet.doGet(DumpServlet.java:44)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:527)
```

joakime: We should remove them, otherwise they get added to the generated cookie twice.

joakime: Pushed a change to this, including the requested javadoc.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
```diff
@@ -126,26 +207,53 @@ public HttpCookie(String name, String value, String domain, String path, long ma
         _comment = comment;
         _version = version;
         _expiration = maxAge < 0 ? -1 : NanoTime.now() + TimeUnit.SECONDS.toNanos(maxAge);
-        _attributes = (attributes == null ? Collections.emptyMap() : attributes);
+        Map<String, String> attrs = null;
+        if (attributes == null)
```
gregw:
```suggestion
        if (attributes == null || attributes.isEmpty())
```

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
```diff
@@ -126,26 +207,53 @@ public HttpCookie(String name, String value, String domain, String path, long ma
...
+     */
     public HttpCookie(String name, String value, int version, Map<String, String> attributes)
     {
         _name = name;
         _value = value;
         _version = version;
-        _attributes = (attributes == null ? Collections.emptyMap() : new TreeMap<>(attributes));
+
+        Map<String, String> attrs = null;
+        if (attributes == null)
```
gregw:
```suggestion
        if (attributes == null || attributes.isEmpty())
```

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
```diff
@@ -126,26 +207,53 @@ public HttpCookie(String name, String value, String domain, String path, long ma
         _comment = comment;
         _version = version;
         _expiration = maxAge < 0 ? -1 : NanoTime.now() + TimeUnit.SECONDS.toNanos(maxAge);
-        _attributes = (attributes == null ? Collections.emptyMap() : attributes);
+        Map<String, String> attrs = null;
+        if (attributes == null)
+            attrs = Collections.emptyMap(); // unmodifiable empty map
+        else
+        {
+            attrs = new HashMap<>(attributes); // create new map, to only capture relevant attributes
```
gregw: The API Cookie class uses a tree map that is case insensitive: `new TreeMap<>(String.CASE_INSENSITIVE_ORDER);`. 

We should probably mimic.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiResponse.java
```diff
@@ -0,0 +1,649 @@
...
+    {
+        if (StringUtil.isBlank(cookie.getName()))
+            throw new IllegalArgumentException("Cookie.name cannot be blank/null");
+
+        // new style cookie, everything is an attribute
+        addCookie(new HttpCookie(
+            cookie.getName(),
+            cookie.getValue(),
+            cookie.getVersion(),
+            cookie.getAttributes()));
```
gregw, line 94: I'm a little cautious about this, as if there are attributes, then we will always create a new map.
But I think most usages will have no attributes.... so maybe OK. But maps are expensive... TreeMaps more so.

gregw: Actually most usages will have attributes, since path & domain are implemented as attributes.

gregw (changes requested):
A couple more niggles is we are to leave it this way.... but always's copying the attribute map is a bit of expense, specially as many cookies will have path and domain attributes.

Would it be so bad to use the attribute map passed in without copying?

I think the real solution is for us to make our HttpCookie an interface and pass in the API Cookie wrapped with our interface rather than copy everything over.   But I think that might be a different PR.

So just the niggles for now.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
```diff
@@ -126,26 +207,53 @@ public HttpCookie(String name, String value, String domain, String path, long ma
...
         _maxAge = StringUtil.isBlank(tmp) ? -1L : Long.valueOf(tmp);
         _expiration = _maxAge < 0 ? -1 : NanoTime.now() + TimeUnit.SECONDS.toNanos(_maxAge);
-        _httpOnly = Boolean.parseBoolean(_attributes.remove("HttpOnly"));
-        _secure = Boolean.parseBoolean(_attributes.remove("Secure"));
-        _comment = _attributes.remove("Comment");
+        _httpOnly = Boolean.parseBoolean(attrs.remove("HttpOnly"));
+        _secure = Boolean.parseBoolean(attrs.remove("Secure"));
+        _comment = attrs.remove("Comment");
+
+        _attributes = Collections.unmodifiableMap(attrs); // don't allow attributes to be modified
```
gregw: Also check here for an empty map.

joakime:
> Would it be so bad to use the attribute map passed in without copying?

We have the same data in 2 places (a field, and the attributes).
And depending on which `HttpCookie` constructor is used, we can even wind up with different values for those pieces of information.
Lets get this PR in with a focus on ServletApiResponse, and then make a new PR for HttpCookie cleanup. (The comment stuff is going to be messy to cleanup anyway)
We don't want an omnibus PR here.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiResponse.java
joakime: We can address this concern in the followup HttpCookie cleanup PR.
I have a few ideas on how to minimize this concern when going from Servlet Cookie API to our HttpCookie API.

gregw (approved):

gregw:
I'm going to have a go at making HttpCookie an interface to get around the difficulties of attributes and multiple servlet APIs.     I'll base it of this branch, but let's not wait for it before merging this one.... I can rebase on main branch once merged.

PR #9184, opened 2023-01-17
Fixes #9183 - ConnectHandler may close the connection instead of send…

### LINKED ISSUE #9183: ConnectHandler may close the connection instead of sending 200 OK
**Jetty version(s)**
10+

**Description**
In case of HTTP CONNECT tunnelling with HTTP/1.1, the upstream server may close the connection immediately.
In such case, there is a race condition trying to write the 200 OK (for a successful tunnel establishment) and closing the connection (because the server closed it immediately).

The client may not see the 200 OK that it is supposed to see, and just see the connection closed.

sbordet:
…ing 200 OK.

Delaying the call to UpstreamConnection.fillInterested() until the 200 OK response has been sent.

### DIFF
--- a/jetty-proxy/src/main/java/org/eclipse/jetty/proxy/ConnectHandler.java
+++ b/jetty-proxy/src/main/java/org/eclipse/jetty/proxy/ConnectHandler.java
@@ -26,6 +26,8 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.Executor;
 import javax.servlet.AsyncContext;
+import javax.servlet.AsyncEvent;
+import javax.servlet.AsyncListener;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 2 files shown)


jetty-proxy/src/main/java/org/eclipse/jetty/proxy/ConnectHandler.java
```diff
@@ -589,6 +592,38 @@ protected void write(EndPoint endPoint, ByteBuffer buffer, Callback callback)
         {
             ConnectHandler.this.write(endPoint, buffer, callback, getContext());
         }
+
+        private class OnCompleteListener implements AsyncListener
+        {
+            private final Runnable onComplete;
+
+            private OnCompleteListener(Runnable onComplete)
```
gregw: This feels needlessly generic?   Why not just make the `UpstreamConnection` implement `AsyncListener`, then it can just call `fillInterest` in `onComplete`?

gregw (dismissed):
OK other than a niggle

electrum:
I tested this with https://github.com/airlift/airlift/pull/1032 (using [2cdd9e26](https://github.com/airlift/airlift/commit/2cdd9e26ada3aef4e3e8b272c78371e65b74831b)) running the test 100 times, all passed. Without this, there were 31 failures.

gregw (approved):

PR #9191, opened 2023-01-18
Jetty-12 Remove usage of HandlerList and reduce usage of Handler.Collection

gregw:
"The best part is no part" - Elon Musk!

The overwhelming usage of `HandlerList` and `Handler.Collection` was for adding the `DefaultHandler` after the main handler.  This PR adds a getter/setter for a `DefaultHandler` on the server, so we no longer need to always create a `Handler.Collection` structure.   This has allowed the deprecated `HandlerList` and `HandlerWrapper` classes to be removed.

In implementing this PR, several problems were found in the calculation of `InvocationType`, not least that it was assumed that an empty `Handler.Collection` was `BLOCKING`.  When this issue was fixed, any dynamic addition of contexts (deployer or SPI server) failed as the `InvocationType` changed.  So this PR also introduces the `isDynamic()` attribute of all `Handler.Container`s.  A dynamic container will always return `BLOCKING` from `getInvocationType()`, as there is always a race with a new handler being added.   A non-dynamic container will return a real `InvocationType`, calculated from its children, but it's mutator methods will ISE if contained handlers are changed.


gregw:
@sbordet @janbartel @joakime This is only a draft PR, but I would appreciate some early review to see if this is a viable approach.

I suggest starting with `jetty-core/jetty-deploy/src/test/resources/jetty.xml` to see the essence of this PR.

Then look at `jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java` to see the dynamic status I added to better deal with the `InvocationType` complications.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java
```diff
@@ -128,6 +132,32 @@ public Server(@Name("threadpool") ThreadPool pool)
         addBean(_threadPool);
         setServer(this);
         addBean(FileSystemPool.INSTANCE, false);
+        setDefaultHandler(new DefaultHandler());
```
gregw: Should we set a `DefaultHandler` by default?  That is a little behavior change, but the clue is in the name!

gregw:
@olamy this PR updates a lot of the integration test XMLs.   I think a lot of them were a bit wrong with strange structures of HandlerList and ContextHandlerCollections.   I think I've fixed most of them up, but would be good if you could double check that I've not thwarted the purpose of any test.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler-use.adoc
```diff
@@ -70,14 +70,14 @@ Server
 
 Server applications may need to deploy to Jetty more than one web application.
 
-Recall from the xref:pg-server-http-handler[introduction] that Jetty offers `HandlerCollection` and `HandlerList` that may contain a sequence of children ``Handler``s.
-However, both of these have no knowledge of the concept of _context_ and just iterate through the sequence of ``Handler``s.
+Recall from the xref:pg-server-http-handler[introduction] that Jetty offers `Handler.Collection` that may contain a collection of children ``Handler``s.
```
janbartel:
```suggestion
Recall from the xref:pg-server-http-handler[introduction] that Jetty offers `Handler.Collection` that contains a sequence of child ``Handler``s.
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler.adoc
```diff
@@ -16,21 +16,14 @@
...
-** `HandlerWrapper` implements this behavior
 
-The `HandlerCollection` behavior (invoking _all_ handlers) is useful when for example the last `Handler` is a logging `Handler` that logs the request (that may have been modified by previous handlers).
-
-The `HandlerList` behavior (invoking handlers up to the first that calls `Request.setHandled(true)`) is useful when each handler processes a different URIs or a different virtual hosts: ``Handler``s are invoked one after the other until one matches the URI or virtual host.
-
-The nested behavior is useful to enrich the request with additional services such as HTTP session support (`SessionHandler`), or with specific behaviors dictated by the Servlet specification (`ServletHandler`).
-
-``Handler``s can be organized in a tree by composing them together:
+A ``Handler``s tree is created by composing ``Handler``s together:
```
janbartel:
```suggestion
A ``Handler`` tree is created by composing ``Handler``s together:
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java
janbartel: You could, but then if someone wants to configure the `DefaultHandler` they have to `new` it again in the xml or code. Instead you could check in `doStart` if there's no `_defaultHandler` configured, then instantiate our `DefaultHandler`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java
```diff
@@ -128,6 +132,32 @@ public Server(@Name("threadpool") ThreadPool pool)
...
+        return _defaultHandler;
+    }
+
+    /**
+     * @param defaultHandler The handler to use if no other handler is set or accepts the request. This handler should
+     *                       always accept the request, even if only to send a 404.
+     */
+    public void setDefaultHandler(Handler defaultHandler)
+    {
+        if (isStarted())
```
janbartel: If `isDynamic()` allow to set a `DefaultHandler` ?

janbartel (approved):
Couple of niggles and comments.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java
gregw: maybe... but that is getting a bit meta.  Let me think on that one.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java
gregw: Note that we don't strictly need a default handler as we will generate a 404 fine without one.   We only really need it for the convenience of getting a context list on a root 404, plus the marketing of serving fav.icon.

Perhaps we just get rid of the whole default handler concept, as it is not really needed?

documentation/jetty-documentation/src/main/asciidoc/old_docs/extras/default-handler.adoc
```diff
@@ -24,24 +24,5 @@
 [[default-handler-usage]]
 ==== Usage
 
-A simple handler that is useful to terminate handler chains with a clean fashion.
-As in the example below, if a resource to be served is not matched within the resource handler the `DefaultHandler` will take care of producing a 404 page.
-This class is a useful template to either extend and embrace or simply provide a similar implementation for customizing to your needs.
+A simple handler that is used by the `Server` to handle request that are not processed by any other `Handler`. If a request is processed by the `DefaultHandler` then it will typically produce a 404 response. However, if the request is for `flav.ico`, then the jetty icon may be served. Also a request to the root URI can be served with a list of available contexts.
```
lorban: Shouldn't that be `favicon.ico` instead of `flav.ico`?

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java
```diff
@@ -505,14 +506,14 @@ public boolean process(Request request, Response response, Callback callback) th
...
         Server server = new Server();
 
-        LoggingHandler loggingHandler = new LoggingHandler();
-        // Link the root Handler with the Server.
-        server.setHandler(loggingHandler);
+        GzipHandler gzipHandler = new GzipHandler();
+        server.setHandler(gzipHandler);
 
         Handler.Collection collection = new Handler.Collection();
+        gzipHandler.setHandler(gzipHandler);
```
lorban: `gzipHandler.setHandler(gzipHandler)`???

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ResourceHandlerTest.java
```diff
@@ -657,6 +657,7 @@ public void before() throws Exception
         FS.ensureEmpty(docRoot);
 
         _server = new Server();
+        _server.setDefaultHandler(null);
```
lorban: This feels really odd. Maybe the test should be adapted to assume the default handler exists by default?

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java
```diff
@@ -999,22 +1006,16 @@ public void defaultHandler() throws Exception
     {
         // tag::defaultHandler[]
         Server server = new Server();
+        server.setDefaultHandler(new DefaultHandler());
```
lorban: Is this explicit set needed?

lorban (changes requested):
I like the idea of treating the `DefaultHandler` specially, like the `ErrorHandler`. This helps making the XML configs simpler.

But this opens the door to oddities which make me wonder if the `DefaultHandler` functionalities shouldn't be provided by the `Server` instead if allowing to configure a special handler?

joakime:
> I like the idea of treating the `DefaultHandler` specially, like the `ErrorHandler`. This helps making the XML configs simpler.
> 
> But this opens the door to oddities which make me wonder if the `DefaultHandler` functionalities shouldn't be provided by the `Server` instead if allowing to configure a special handler?

In Jetty 7 thru 11 ...

The various manipulations of `DefaultHandler` seen in the community are (in this order) ...

1. Remove it from the handler tree entirely (most common, either intentionally, or accidentally)
2. Configuring it (eg: setting `showContexts` to false, which is a straight 404 with no response body)
3. Replacing it with a RewriteHandler that just sends a redirect to a configured location.
4. Replacing it with their own custom implementation.

I'd be curious to know what would happen if the `DefaultHandler` is present in the `Server.setDefaultHandler()` _and_ the handler tree.

documentation/jetty-documentation/src/main/asciidoc/old_docs/extras/default-handler.adoc
sbordet: This is the old documentation, should not be touched.

documentation/jetty-documentation/src/main/asciidoc/old_docs/extras/shutdown-handler.adoc
```diff
@@ -34,10 +34,9 @@ It can also be added as a part of handler chain or collection.
 [source, java, subs="{sub-order}"]
 ----
     Server server = new Server(8080);
-    HandlerList handlers = new HandlerList();
-    handlers.setHandlers(new Handler[]
-    { someOtherHandler, new ShutdownHandler(server,"secret password") });
-    server.setHandler(handlers);
+    ShutdownHandler shutdown = new ShutdownHandler(server,"secret password") });
+    server.setHandler(shutdown);
+    shutdown.setHandler(someOtherHandler);
```
sbordet: Old documentation.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler-use.adoc
```diff
@@ -70,14 +70,14 @@ Server
...
 
-Recall from the xref:pg-server-http-handler[introduction] that Jetty offers `HandlerCollection` and `HandlerList` that may contain a sequence of children ``Handler``s.
-However, both of these have no knowledge of the concept of _context_ and just iterate through the sequence of ``Handler``s.
+Recall from the xref:pg-server-http-handler[introduction] that Jetty offers `Handler.Collection` that contains a sequence of child ``Handler``s.
+However, this has no knowledge of the concept of _context_ and just iterates through the sequence of ``Handler``s.
 
 A better choice for multiple web application is `ContextHandlerCollection`, that matches a _context_ from either its _context path_ or _virtual host_, without iterating through the ``Handler``s.
 
-If `ContextHandlerCollection` does not find a match, it just returns.
+If `ContextHandlerCollection` does not find a match, it just returns false from ``process``.
```
sbordet:
```suggestion
If `ContextHandlerCollection` does not find a match, it just returns `false` from its `process(\...)` method.
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler-use.adoc
```diff
@@ -70,14 +70,14 @@ Server
...
 What happens next depends on the `Handler` tree structure: other ``Handler``s may be invoked after `ContextHandlerCollection`, for example `DefaultHandler` (see xref:pg-server-http-handler-use-util-default-handler[this section]).
-Eventually, if `Request.setHandled(true)` is not called, Jetty returns an HTTP `404` response to the client.
+Eventually, if no ``Handler`` returns `true` from `process`, then Jetty returns an HTTP `404` response to the client.
```
sbordet:
```suggestion
Eventually, if no `Handler` returns `true` from their own `process(\...)` method, then Jetty returns an HTTP `404` response to the client.
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler-use.adoc
```diff
@@ -132,7 +132,7 @@ If you need to serve static resources from multiple directories:
 include::../../{doc_code}/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java[tags=multipleResourcesHandler]
 ----
 
-If the resource is not found, `ResourceHandler` will not call `Request.setHandled(true)` so what happens next depends on the `Handler` tree structure.
+If the resource is not found, `ResourceHandler` will not return `true` from the `process` method, so what happens next depends on the `Handler` tree structure.
```
sbordet:
```suggestion
If the resource is not found, `ResourceHandler` will not return `true` from the `process(\...)` method, so what happens next depends on the `Handler` tree structure.
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler-use.adoc
```diff
@@ -277,7 +277,7 @@ include::../../{doc_code}/org/eclipse/jetty/docs/programming/server/http/HTTPSer
 [[pg-server-http-handler-use-util-default-handler]]
 ====== DefaultHandler
 
-`DefaultHandler` is a terminal `Handler` that always calls `Request.setHandled(true)` and performs the following:
+`DefaultHandler` is a terminal `Handler` that always returns `true` from its `process` method and performs the following:
```
sbordet:
```suggestion
`DefaultHandler` is a terminal `Handler` that always returns `true` from its `process(\...)` method and performs the following:
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler-use.adoc
```diff
@@ -350,7 +349,7 @@ Note also how adding a `Servlet` or a `Filter` returns a _holder_ object that ca
 
 When a request arrives to `ServletContextHandler` the request URI will be matched against the ``Filter``s and ``Servlet`` mappings and only those that match will process the request, as dictated by the Servlet specification.
 
-IMPORTANT: `ServletContextHandler` is a terminal `Handler`, that is it always calls `Request.setHandled(true)` when invoked.
+IMPORTANT: `ServletContextHandler` is a terminal `Handler`, that is it always returns `true` from its `process` method when invoked.
```
sbordet:
```suggestion
IMPORTANT: `ServletContextHandler` is a terminal `Handler`, that is it always returns `true` from its `process(\...)` method when invoked.
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler-use.adoc
```diff
@@ -350,7 +349,7 @@ Note also how adding a `Servlet` or a `Filter` returns a _holder_ object that ca
...
 Server applications must be careful when creating the `Handler` tree to put ``ServletContextHandler``s as last ``Handler``s in a `HandlerList` or as children of `ContextHandlerCollection`.
```
sbordet: Still a reference to `HandlerList` here?

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler.adoc
```diff
@@ -16,21 +16,14 @@
 
 An `org.eclipse.jetty.server.Handler` is the component that processes incoming HTTP requests and eventually produces HTTP responses.
 
-``Handler``s can be organized in different ways:
+``Handler``s can process and request themselves, or they can be ``Handler.Container``s that delegate requests to one or more contained ``Handler``s.  This allows ``Handler``s to be organised as a tree comprised of:
```
sbordet:
```suggestion
``Handler``s can process the HTTP request themselves, or they can be ``Handler.Container``s that delegate HTTP request processing to one or more contained ``Handler``s.
This allows ``Handler``s to be organized as a tree comprised of:
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler.adoc
```diff
@@ -16,21 +16,14 @@
...
+* leaf ``Handler``s that return true from the ``process`` method, generate a response and succeed the ``Callback``.
```
sbordet:
```suggestion
* Leaf ``Handler``s that return `true` from the `process(\...)` method, generate a response and succeed the ``Callback``.
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler.adoc
```diff
@@ -16,21 +16,14 @@
...
+* a ``Handler.Wrapper`` can be used to form a chain of ``Handler``s where request, response or callback objects are wrapped in the ``process`` method before being passed down the chain.
```
sbordet:
```suggestion
* A ``Handler.Wrapper`` can be used to form a chain of ``Handler``s where request, response or callback objects are wrapped in the ``process(\...)`` method before being passed down the chain.
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler.adoc
```diff
@@ -16,21 +16,14 @@
...
+* a ``Handler.Collection`` contains a sequence of ``Handler``s, with each being called in series until one returns ``true`` from its ``process`` method.
```
sbordet:
```suggestion
* A ``Handler.Collection`` that contains a sequence of ``Handler``s, with each `Handler` being called in sequence until one returns `true` from its `process(\..)` method.
```

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler.adoc
```diff
@@ -16,21 +16,14 @@
...
+* a specialized ``Handler.Container`` may use attributes of the request (e.g. `getHttpURI()``) to select from one or more contained ``Handler``s to delegate the `process` method call to.
```
sbordet:
```suggestion
* A specialized ``Handler.Container`` that may use properties of the request (for example, the URI, or one of the headers, etc.) to select from one or more contained ``Handler``s to delegate the HTTP request processing to.
```

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java
sbordet:
```suggestion
        gzipHandler.setHandler(collection);
```

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java
```diff
@@ -582,29 +583,35 @@ public boolean process(Request request, Response response, Callback callback)
...
                     // Rewrite old paths to new paths.
-                    HttpURI uri = jettyRequest.getHttpURI();
+                    HttpURI uri = request.getHttpURI();
                     String newPath = "/new_path/" + path.substring("/old_path/".length());
-                    HttpURI newURI = HttpURI.build(uri).path(newPath);
-                    // Modify the request object.
-                    jettyRequest.setHttpURI(newURI);
+                    HttpURI newURI = HttpURI.build(uri).path(newPath).asImmutable();
+
+                    // Modify the request object by wrapping
```
sbordet:
```suggestion
                    // Modify the request object by wrapping the HttpURI.
```

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java
```diff
@@ -582,29 +583,35 @@ public boolean process(Request request, Response response, Callback callback)
...
+                        @Override
+                        public HttpURI getHttpURI()
+                        {
+                            return newURI;
+                        }
+                    };
                 }
 
                 // This Handler is not handling the request, so
                 // it does not call jettyRequest.setHandled(true).
```
sbordet: Remove these 3 lines.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java
```diff
@@ -618,7 +625,7 @@ public void handle(String target, Request jettyRequest, HttpServletRequest reque
         server.setHandler(filter);
 
         server.start();
-*/
+
```
sbordet: Remove this empty line.

jetty-core/jetty-server/src/main/config/modules/server.mod
```diff
@@ -111,3 +111,13 @@ etc/jetty.xml
 ## The number of server scheduler threads.
 # jetty.scheduler.threads=1
 # end::documentation-scheduler-config[]
+
+## If the handlers of the ContextHandlerCollection can be updated once started
```
sbordet:
```suggestion
## Whether the handlers of the ContextHandlerCollection can be updated once the Server is started.
```

jetty-core/jetty-server/src/main/config/modules/server.mod
```diff
@@ -111,3 +111,13 @@ etc/jetty.xml
...
+## If set to false, then eeX-deploy module jetty.deploy.scanInterval should also be set to 0.
```
sbordet:
```suggestion
## If set to false, then eeN-deploy module jetty.deploy.scanInterval should also be set to 0.
```

jetty-core/jetty-server/src/main/config/modules/server.mod
```diff
@@ -111,3 +111,13 @@ etc/jetty.xml
...
+# jetty.server.contexts.dynamic=true
+
+## Should the DefaultServlet server the jetty favicon.ico
```
sbordet:
```suggestion
## Should the DefaultHandler serve the Jetty favicon.ico.
```

jetty-core/jetty-server/src/main/config/modules/server.mod
```diff
@@ -111,3 +111,13 @@ etc/jetty.xml
...
+# jetty.server.default.serveIcon=true
+
+## Should the DefaultServlet server show a list of known contexts
```
sbordet:
```suggestion
## Should the DefaultHandler show a list of known contexts.
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java
sbordet: Feels weird that we serve `jetty-dir.css` from `ResourceHandler`, but `favicon.ico` from the `DefaultHandler`, given the files are actually in the same place.

If we move out the serving of `favicon.ico`, what remains is the small HTML to present the list of contexts, which is useful, and people may want to replace it, so I think the concept of a `DefaultHandler` should remain.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java
```diff
@@ -128,6 +132,32 @@ public Server(@Name("threadpool") ThreadPool pool)
...
+        setDefaultHandler(new DefaultHandler());
+    }
+
+    public Handler getDefaultHandler()
+    {
+        return _defaultHandler;
+    }
+
+    /**
+     * @param defaultHandler The handler to use if no other handler is set or accepts the request. This handler should
```
sbordet:
```suggestion
     * @param defaultHandler The handler to use if no other handler is set or has processed the request. This handler should
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DefaultHandler.java
```diff
@@ -64,6 +64,14 @@ public class DefaultHandler extends Handler.Abstract
...
     public DefaultHandler()
     {
+        super(InvocationType.NON_BLOCKING);
+    }
+
+    public DefaultHandler(boolean serveIcon, boolean showContexts)
+    {
+        this();
+        _serveIcon = serveIcon;
+        _showContexts = showContexts;
```
sbordet, line 75: Should be the opposite: the constructor with less parameters calling the one with more parameters, so the defaults are clear.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/PartialRFC2616Test.java
```diff
@@ -62,7 +61,7 @@ public void init() throws Exception
         context.setContextPath("/");
         context.setHandler(new DumpHandler());
 
-        server.setHandler(new HandlerList(vcontext, context));
+        server.setHandler(vcontext);
```
sbordet: Not the same as before, because now `context` is not added to the context tree.

sbordet (changes requested):

gregw:
@joakime 

> I'd be curious to know what would happen if the `DefaultHandler` is present in the `Server.setDefaultHandler()` _and_ the handler tree.

Essentially this change makes Server a `Handler.Collection` of 2 `Handler`s.  If a user sets up a structure with a`DefaultHandler` in it, then it will work perfectly fine and the server's default would just never get called.

gregw:
@sbordet 
> Feels weird that we serve jetty-dir.css from ResourceHandler, but favicon.ico from the DefaultHandler, given the files are actually in the same place.

The two are different.  We want `jetty-dir.css` to appear in every directory in the content tree, whilst `favicon.ico` is only in the top level, outside of any context.
Furthermore, `jetty-dir.css` is necessary for a feature we offer: listing directories, `whilst `favicon.ico` is a marketing stunt of dubious value.

We could definitely remove `favicon.ico` and nobody would care/complain and many would be happy to not have to turn it off!

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ResourceHandlerTest.java
gregw: Not necessary. Removed and testing for "Not Found" in the body rather than the repeated URI.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java
gregw: It's not needed, but it is used for the documentation about how to set a `DefaultHandler`.  I'll add some configuration to the example to make it more meaningful.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/PartialRFC2616Test.java
gregw: Ooops.  Note this test is disabled.  I have undisasbled it and updated a few tests for the new DumpHandler.   But 4 tests still fail.  I have disabled those and they need to be looked at in a different PR

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DefaultHandler.java
gregw: OK.   Should we make them final fields that can only be set with constructor injection?  Guess it doesn't matter as it is not on hot path.

joakime:
> We could definitely remove `favicon.ico` and nobody would care/complain and many would be happy to not have to turn it off!

I see a `FavIconHandler` as an option.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java
gregw: Since we set a `DefaultHandler` in our `jetty.xml`, like we always have, I have made the embedded server no have a default handler, unless one is set. That is the same behavior as before and a simple raw 404 will be produced by the server.

> Feels weird that we serve jetty-dir.css from ResourceHandler, but favicon.ico from the DefaultHandler, given the files are actually in the same place.

The two are different. We want jetty-dir.css to appear in every directory in the content tree, whilst favicon.ico is only in the top level, outside of any context.
Furthermore, jetty-dir.css is necessary for a feature we offer: listing directories, whilst favicon.ico` is a marketing stunt of dubious value.
We could definitely remove favicon.ico and nobody would care/complain and many would be happy to not have to turn it off!, but now that it is not set by default (unless using our module which is easily configured), I'm OK keeping it.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java
gregw: ok, added isDynamic check

gregw:
@joakime @sbordet @lorban Thanks for the review. Summary of updates:
 + doco suggestions all accepted
 + Server now does not create a `DefaultServlet` by default, so embedded usage remains unchanged.
 + If users add a `DefaultServlet` with a `Handler.Collection` like before, everything will work normally
 + Better examples of how we want the defaults server configured now.

gregw:
> 
> I see a `FavIconHandler` as an option.

I don't think so.  It's just a bit of cheap marketing, so nobody is ever going to add it deliberately.   We are just using our default server setup with an incomplete user provided configuration (no root context) to: a) be helpful with a list of contexts; b) sneak in a little jetty promo.

Both are entirely optional, with b) being border line always unnecessary.

It is now off by default in an embedded usage and only triggers in jetty-home usage if there is no root, plus it is now easier to turn off with a property.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler.adoc
```diff
@@ -16,21 +16,14 @@
 
 An `org.eclipse.jetty.server.Handler` is the component that processes incoming HTTP requests and eventually produces HTTP responses.
 
-``Handler``s can be organized in different ways:
+``Handler``s can process the HTTP request themselves, or they can be a ``Handler.Container``s that delegate HTTP request processing to one or more contained ``Handler``s.  This allows ``Handler``s to be organised as a tree comprised of:
```
sbordet: All sentences on a new line, so `This allows...` needs to be on a new line.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler.adoc
sbordet: This was not applied?

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler.adoc
sbordet: Not applied?

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler.adoc
sbordet: Only partially applied?

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler.adoc
sbordet: Not applied?

sbordet (changes requested):
A few more nibbles.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/server/http/server-http-handler.adoc
gregw: done.

gregw:
@sbordet I can't work out what your "This was not applied?" comment applies to?  It is listed against outdated code?

sbordet (approved):

gregw:
@lorban Can you give me more details of what you want or dismiss your review?

lorban (approved):
@gregw I quite like how things turned: 
 - not needing to explicitly configure the default handler in every XML config (nice cleanup)
 - having the possibility to configure a default handler on the server (nice addition)
 - no default handler configured on the server by default (non-surprising default)

so it all LGTM.

PR #9193, opened 2023-01-18
Issue #9067 - Fixes to include dispatch for pathInContext and welcome files.

### LINKED ISSUE #9067: TCK DefaultServlet handling of dispatch include incorrect
**jetty-12**

The `DefaultServlet.doGet` method is missing code to handle an `INCLUDE`.  For an `INCLUDE`, the resource to be served is _not_ found in the request (as is the case for `REQUEST` and `FORWARD`) but rather found in the request attributes prefixed with `jakarta.servlet.include.`:

```
jakarta.servlet.include.request_uri
jakarta.servlet.include.context_path
jakarta.servlet.include.servlet_path
jakarta.servlet.include.mapping
jakarta.servlet.include.path_info
jakarta.servlet.include.query_string
```

Note that our `DefaultServletTest.testIncludedDifferentWelcomeBase` unit test has been disabled with comment `"Not working as RequestDispatcher.include() isn't behaving as expected"`.

The relevant TCK test is:
https://github.com/jakartaee/platform-tck/blob/master/src/com/sun/ts/tests/servlet/spec/multifiltermapping/URLClient.java#L249

lachlan-roberts:
## Issue #9067 

Fix logic around include dispatches for DefaultServlet.

### DIFF
--- a/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
+++ b/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
@@ -1088,11 +1088,6 @@ public class DefaultServlet extends HttpServlet
     {
         String servletPath = isPathInfoOnly ? "/" : includedServletPath;
         String pathInfo = (String)request.getAttribute(RequestDispatcher.INCLUDE_PATH_INFO);
-        if (servletPath == null)
-        {
-            servletPath = request.getServletPath();
-            pathInfo = request.getPathInfo();
-        }
         return URIUtil.addPaths(servletPath, pathInfo);
     }
 
--- a/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
+++ b/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
@@ -359,10 +359,11 @@ public class DefaultServlet extends HttpServlet
     @Override
     protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException
     {
-        boolean included = isIncluded(req);
+        String includedServletPath = (String)req.getAttribute(RequestDispatcher.INCLUDE_SERVLET_PATH);
+        boolean included = includedServletPath != null;
         String pathInContext;
         if (included)
-            pathInContext = getIncludedPathInContext(req, isPathInfoOnly());
+            pathInContext = getIncludedPathInContext(req, includedServletPath, isPathInfoOnly());
         else
             pathInContext = URIUtil.addPaths(isPathInfoOnly() ? "/" : req.getServletPath(), req.getPathInfo());
 
@@ -477,10 +478,12 @@ public class DefaultServlet extends HttpServlet
             }
 
             _httpFields = fields.asImmutable();
+            String includedServletPath = (String)request.getAttribute(RequestDispatcher.INCLUDE_SERVLET_PATH);
+            boolean included = includedServletPath != null;
             if (request.getDispatcherType() == DispatcherType.REQUEST)
                 _uri = getWrapped().getHttpURI();
-            else if (isIncluded(request))
-                _uri = Request.newHttpURIFrom(getWrapped(), getIncludedPathInContext(request, false));
+            else if (included)
+                _uri = Request.newHttpURIFrom(getWrapped(), getIncludedPathInContext(request, includedServletPath, false));
             else
                 _uri = Request.newHttpURIFrom(getWrapped(), URIUtil.addPaths(_servletRequest.getServletPath(), _servletRequest.getPathInfo()));
         }
@@ -920,8 +923,10 @@ public class DefaultServlet extends HttpServlet
             HttpServletRequest request = getServletRequest(coreRequest);
             String pathInContext = Request.getPathInContext(coreRequest);
             String requestTarget;
-            if (isIncluded(request))
-                requestTarget = getIncludedPathInContext(request, isPathInfoOnly());
+            String includedServletPath = (String)request.getAttribute(RequestDispatcher.INCLUDE_SERVLET_PATH);
+            boolean included = includedServletPath != null;
+            if (included)
+                requestTarget = getIncludedPathInContext(request, includedServletPath, isPathInfoOnly());
             else
                 requestTarget = isPathInfoOnly() ? request.getPathInfo() : pathInContext;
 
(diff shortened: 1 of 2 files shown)


jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -471,10 +475,14 @@ private static class ServletCoreRequest extends Request.Wrapper
...
                 }
             }
+
             _httpFields = fields.asImmutable();
-            _uri = (request.getDispatcherType() == DispatcherType.REQUEST)
-                ? getWrapped().getHttpURI()
-                : Request.newHttpURIFrom(getWrapped(), URIUtil.addPaths(_servletRequest.getServletPath(), _servletRequest.getPathInfo()));
+            if (request.getDispatcherType() == DispatcherType.REQUEST)
+                _uri = getWrapped().getHttpURI();
+            else if (isIncluded(request))
```
gregw: Can we remember isIncluded result and store in this wrapper as a field?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -1077,6 +1079,23 @@ private HttpServletResponse getServletResponse(Response response)
         }
     }
 
+    private static String getIncludedPathInContext(HttpServletRequest request, boolean isPathInfoOnly)
+    {
+        String servletPath = isPathInfoOnly ? "/" : (String)request.getAttribute(RequestDispatcher.INCLUDE_SERVLET_PATH);
```
gregw: It kind of feels like this attribute could be used to do an isIncluded, thus avoiding two attribute lookups? return null if it is not set.

gregw (changes requested):
Can you try to reduce the number of attribute lookups? other than that it looks OK

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
joakime: What if a user decides to add this attribute in their Servlet code?

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletTest.java
```diff
@@ -114,6 +115,11 @@ public void init() throws Exception
         connector = new LocalConnector(server);
         connector.getConnectionFactory(HttpConfiguration.ConnectionFactory.class).getHttpConfiguration().setSendServerVersion(false);
 
+        ServerConnector serverConnector = new ServerConnector(server);
+        serverConnector.setPort(8080);
```
joakime: Don't use a hardcoded port, esp 8080.

joakime (changes requested):
Test case shouldn't use a hardcoded port. (like 8080)

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
gregw: We ignore it. Our getAttribute always returns our values for these.

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletTest.java
lachlan-roberts: oops yeah I did that for debugging, I will remove.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
joakime: In ee9/ee8 we have a wrapper on Request that forces these attributes from our code before it asks the actual stored attributes, I cannot find something similar in ee10.

But what about someone doing this ...

``` java
    public class ZedFilter implements Filter
    {
        @Override
        public void doFilter(ServletRequest request, ServletResponse response, 
              FilterChain chain) throws IOException, ServletException
        {
            request.setAttribute(RequestDispatcher.INCLUDE_SERVLET_PATH, "/alt-servlet-path");
            chain.doFilter(request, response);
        }
    }
```

And this filter is in the chain for DefaultServlet?

gregw: Yes that will look like an include. to the default servlet.  But it has always been possible for users to do this, as they can wrap request and intercept the `getAttribute` method and we have had wrapper object identity up until 12, so we could not stop them doing that.
I think all the servlet API attributes are kind of vulnerable to being messed about with by bad applications, but I don't think they can do anything that way that they couldn't do otherwise.  Filter can just do a real include if they want.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
lachlan-roberts: I don't think this would be very useful because most of the usages of `isIncluded()` is not on the wrapped request.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
lachlan-roberts: > It kind of feels like this attribute could be used to do an isIncluded, thus avoiding two attribute lookups? return null if it is not set.

Does make the a bit harder to follow but will reduce 1 call to `getAttribute()`. Pushed a commit for this now.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -1077,6 +1084,23 @@ private HttpServletResponse getServletResponse(Response response)
         }
     }
 
+    private static String getIncludedPathInContext(HttpServletRequest request, String includedServletPath, boolean isPathInfoOnly)
+    {
+        String servletPath = isPathInfoOnly ? "/" : includedServletPath;
+        String pathInfo = (String)request.getAttribute(RequestDispatcher.INCLUDE_PATH_INFO);
+        if (servletPath == null)
```
gregw: What is the case when `servletPath` is null?  If `isPathInfoOnly` is true, then the it is "/". If iPIO is false, then it is `includedServletPath`, which is never null because it is tested before calling this method.

gregw (changes requested):
Mostly looks good, just some dead code I think needs to be removed.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
lachlan-roberts: yep I don't think the `if` block is needed anymore

gregw (approved):

PR #9185, opened 2023-01-17
Happy New Year 2023!

joakime:
The standard start of year bump of year in license headers.

PR #9194, opened 2023-01-18
License Header at Creation Year

joakime:
An alternative for #9185


joakime:
The license check is quite slow with this approach. (it takes about 18 minutes on my machine)

I've requested a feature from the license-maven-plugin folks mathieucarbou/license-maven-plugin#461 to make this speedier.

joakime:
This has moved the license check to a profile, and introduced a new profile.

* `-Plicense-check` : performs the license check (CI specific profile?)
* `-Plicense-update` : updates all of the files with the creation year. (not something we should be using often at all)

gregw:
I'm closing this PR as it is just too big for the tools. A PR with >3000 files changed will never be correctly reviewed through a web interface.

Also, I'm working on my own script to do the copyright update, which should only ever need to be done once.   We will then need a PR with just the license check changes from this PR in it.

However, this is not a high priority at the moment, as we have too many PRs open to make a change to every source file.

PR #9195, opened 2023-01-22
Fixes #9166 - Jetty 12: review/remove ByteBufferPool

### LINKED ISSUE #9166: Jetty 12: review/remove ByteBufferPool
**Jetty version(s)**
12

**Description**
The Jetty 12 code has now standardized to use `RetainableByteBufferPool`, so `ByteBufferPool` should be removed.

This would simplify a lot the codebase, and also clean up the Jetty component tree, because now the `ByteBufferPool`s appear in the component tree, but not the `RetainableByteBufferPool`s.

This is a prerequisite for #9148, because it would then allow to inject leak-tracking `Pool` instances.

sbordet:
* Replaced usages of ByteBufferPool with RetainableByteBufferPool.
* Removed ByteBufferPool and related classes.
* Updated all methods such as `getBuffer()` to `getByteBuffer()` for consistency.
* Updated javadocs.


sbordet:
@gregw @lachlan-roberts @lorban these are the major files to review:

```
oej.client.ContentDecoder
oej.fcgi.generator.*
oej.http.GZIPContentDecoder
oej.io.RetainableByteBuffer
oej.io.RetainableByteBufferPool
oej.server bytebufferpool-quadratic.mod & jetty-bytebufferpool-quadratic.xml
oej.server.GzipRequest
oej.server.AbstractConnector
oej.server.Server
oej.websocket.core.Parser
oej.websocket.core.PerMessageDeflateExtension
```

Note in particular that I changed the use of `Server.getBean(...)` for the `Scheduler` and the `RetainableByteBufferPool`.

Also changed the `ArrayRetainableByteBufferPool.[Exponential|Logarithmic]` to `Quadratic`.

One thing open for discussion is whether the Jetty module `bytebufferpool.mod` should be renamed to the mouthful `retainablebytebufferpool.mod`, or perhaps just `bufferpool.mod`.

jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jetty-common/src/test/java/org/eclipse/jetty/ee9/websocket/common/MessageOutputStreamTest.java
```diff
@@ -28,29 +32,49 @@
...
-
+    private final AtomicInteger leaks = new AtomicInteger();
+    private RetainableByteBufferPool bufferPool;
     private OutgoingMessageCapture sessionCapture;
 
     @BeforeEach
-    public void setupTest() throws Exception
+    public void beforeEach()
     {
+        bufferPool = new ArrayRetainableByteBufferPool()
```
lachlan-roberts, line 49: Is this a leak tracking `RetainableByteBufferPool` implementation? why delete `TestableLeakTrackingBufferPool` then.

jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jetty-common/src/test/java/org/eclipse/jetty/ee10/websocket/common/MessageOutputStreamTest.java
```diff
@@ -34,12 +36,14 @@
...
-    public TestableLeakTrackingBufferPool bufferPool = new TestableLeakTrackingBufferPool("Test");
+    // TODO: restore leak tracking.
+    public RetainableByteBufferPool bufferPool = new ArrayRetainableByteBufferPool();
 
     @AfterEach
     public void afterEach()
     {
-        bufferPool.assertNoLeaks();
+        // TODO: restore leak tracking.
+//        bufferPool.assertNoLeaks();
```
lachlan-roberts: Why have these tests been commented out but you have a different change for the same ee9 test?

jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jetty-common/src/test/java/org/eclipse/jetty/ee9/websocket/common/MessageOutputStreamTest.java
sbordet: Because it was using the other pool class.

jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jetty-common/src/test/java/org/eclipse/jetty/ee10/websocket/common/MessageOutputStreamTest.java
sbordet: Good catch, restore here too.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
```diff
@@ -14,61 +14,177 @@
...
+        private final List<ByteBuffer> byteBuffers;
+
+        public Accumulator(RetainableByteBufferPool bufferPool)
+        {
+            this.bufferPool = bufferPool;
+            this.buffers = new ArrayList<>();
+            this.byteBuffers = new ArrayList<>();
+        }
+
+        public RetainableByteBuffer acquire(int capacity, boolean direct)
```
lorban: This method feels mis-placed or superfluous: `Accumulator` acting as a pool doesn't look right. This makes me believe there is something not entirely right about this contract.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
```diff
@@ -14,61 +14,177 @@
...
+            }
+            return length;
+        }
+
+        public int getSize()
+        {
+            return byteBuffers.size();
+        }
+
+        public void release()
```
lorban, line 174: Since this `release` method isn't paired with a `retain` sibling (rightly so in this context), maybe it should be called `close` instead, and then make `Accumulator` implement `Closeable`?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
```diff
@@ -14,97 +14,151 @@
...
+ *     and calling {@link #release()} throws {@link IllegalStateException}</li>
+ *     <li>out of pool but not retained; in this case {@link #isRetained()}
+ *     returns {@code false} and calling {@link #release()} returns {@code true}</li>
+ *     <li>out of pool and retained; in this case {@link #isRetained()}
+ *     returns {@code true} and calling {@link #release()} returns {@code false}</li>
  * </ul>
- * <p>Calling {@link #release()} on a out of pool and retained instance does not re-pool it while that re-pools it on a out of pool but not retained instance.</p>
  */
-public class RetainableByteBuffer extends Retainable.ReferenceCounter
+public interface RetainableByteBuffer extends Retainable
```
lorban, line 37: :+1: to the fact that this is now an interface, that will give us a lot of wiggle room to rework the internal bits over time.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
```diff
@@ -14,97 +14,151 @@
...
+     * <p>Use this method to wrap user-provided {@code ByteBuffer}s, or
+     * {@code ByteBuffer}s that hold constant bytes, to make them look
+     * like {@code RetainableByteBuffer}s.</p>
+     * <p>The returned {@code RetainableByteBuffer} {@link #canRetain()}
+     * method always returns {@code false}.</p>
+     *
+     * @param byteBuffer the {@code ByteBuffer} to wrap
+     * @return a non-retainable {@code RetainableByteBuffer}
+     */
+    public static RetainableByteBuffer asNonRetainable(ByteBuffer byteBuffer)
```
lorban: This method should be called `wrap` IMHO.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBuffer.java
```diff
@@ -14,97 +14,151 @@
...
+    /**
+     * @return whether this instance is retained
+     * @see ReferenceCounter#isRetained()
+     */
+    public boolean isRetained();
+
+    /**
+     * @return the wrapped {@code ByteBuffer}
+     */
+    public ByteBuffer getByteBuffer();
```
lorban, line 69: You should document that this can never return `null`.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
```diff
@@ -14,61 +14,177 @@
...
         {
-            _byteBufferPool.release(retainedBuffer.getBuffer());
+            getWrapped().clear();
         }
+    }
 
+    /**
+     * <p>A {@link RetainableByteBufferPool} that does not pool its
+     * {@link RetainableByteBuffer}s.</p>
+     * <p>The returned {@code RetainableByteBuffer}s are reference
```
lorban, line 96:
```diff
---<p>The returned {@code RetainableByteBuffer}s are reference counted.</p>
+++<p>The returned objects do fully respect the {@code RetainableByteBuffer} contract despite that when they are released, they are not re-used.</p>
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java
```diff
@@ -119,13 +125,22 @@ public Server(@Name("address") InetSocketAddress addr)
...
+        addConnector(connector);
     }
 
     public Server(@Name("threadpool") ThreadPool pool)
     {
-        _threadPool = pool != null ? pool : new QueuedThreadPool();
+        this(pool, null, null);
+    }
+
+    public Server(ThreadPool threadPool, Scheduler scheduler, RetainableByteBufferPool bufferPool)
```
lorban: You should also add `@Name` annotations.

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ByteBufferAccumulatorTest.java
```diff
@@ -282,47 +292,35 @@ private int writeInFlushMode(ByteBuffer from, ByteBuffer to)
         return written;
     }
 
-    public static class CountingBufferPool extends LeakTrackingByteBufferPool
+    private static class CountingBufferPool extends RetainableByteBufferPool.Wrapper
```
lorban, line 295: This pattern has been replicated across quite a few tests. Shouldn't we introduce it as a non-test class and use it everywhere there is a `TODO: track leaks`?

It's not a full blown leak tracker, but it could at least be used to verify that tests aren't leaking.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferAccumulator.java
```diff
@@ -24,26 +24,26 @@
...
  * Accumulates data into a list of ByteBuffers which can then be combined into a single buffer or written to an OutputStream.
  * The buffer list automatically grows as data is written to it, the buffers are taken from the
- * supplied {@link ByteBufferPool} or freshly allocated if one is not supplied.
+ * supplied {@link RetainableByteBufferPool} or freshly allocated if one is not supplied.
  *
  * The method {@link #ensureBuffer(int, int)} is used to write directly to the last buffer stored in the buffer list,
  * if there is less than a certain amount of space available in that buffer then a new one will be allocated and returned instead.
  * @see #ensureBuffer(int, int)
  */
 public class ByteBufferAccumulator implements AutoCloseable
```
lorban, line 34: This class is share a bit of functionality, but more visibly its name with `RetainableByteBufferPool.Accumulator`, which I find confusing. It feels like they should be merged, or named differently.

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ArrayRetainableByteBufferPoolTest.java
```diff
@@ -350,42 +349,31 @@ public void testExponentialPool() throws IOException
...
-        buffer.getBuffer().order(ByteOrder.LITTLE_ENDIAN);
+        assertThat(buffer.getByteBuffer().order(), Matchers.is(ByteOrder.BIG_ENDIAN));
+        buffer.getByteBuffer().order(ByteOrder.LITTLE_ENDIAN);
         assertThat(buffer.release(), is(true));
-        assertThat(buffer.getBuffer().order(), Matchers.is(ByteOrder.BIG_ENDIAN));
+        assertThat(buffer.getByteBuffer().order(), Matchers.is(ByteOrder.BIG_ENDIAN));
     }
 
     @Test
     public void testLogarithmic()
```
lorban: Since the logarithmic pool has been replaced with the quadratic one, isn't this redundant with `testQuadraticPool`?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferOutputStream2.java
```diff
@@ -34,35 +34,30 @@ public ByteBufferOutputStream2()
...
-    public ByteBufferPool getByteBufferPool()
-    {
-        return _accumulator.getByteBufferPool();
+        _accumulator = new ByteBufferAccumulator(bufferPool == null ? new RetainableByteBufferPool.NonPooling() : bufferPool, direct);
     }
 
     /**
      * Take the combined buffer containing all content written to the OutputStream.
-     * The caller is responsible for releasing this {@link ByteBuffer} back into the {@link ByteBufferPool}.
+     * The caller is responsible for releasing this {@link ByteBuffer}.
```
lorban:
```diff
--- this {@link ByteBuffer}.
+++ the returned {@link RetainableByteBuffer}.
```

lorban (changes requested):

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ByteBufferAccumulatorTest.java
sbordet: Sure, but in another PR.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferAccumulator.java
sbordet: Added TODO.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
sbordet: I prefer `release()`.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/RetainableByteBufferPool.java
sbordet: Removed `acquire()` from `Accumulator`.

jetty-core/jetty-http2/jetty-http2-common/src/main/java/org/eclipse/jetty/http2/internal/parser/HeaderBlockFragments.java
```diff
@@ -16,52 +16,50 @@
...
     }
 
     public void storeFragment(ByteBuffer fragment, int length, boolean last)
     {
         if (storage == null)
         {
             int space = last ? length : length * 2;
-            storage = byteBufferPool.acquire(space, fragment.isDirect());
-            storage.clear();
+            storage = bufferPool.acquire(space, fragment.isDirect());
```
gregw, line 41: I think we need to better document all allocations.  In this case, the matching release is not in this class as the storage buffer is passed out in the ·`complete` method... so somebody else does the release.

gregw:
Wouldn't it be cool if code like this:
```java
        RetainableByteBuffer retainableByteBuffer = byteBufferPool.acquire(getBufferSize(), isUseDirectByteBuffers());
        ByteBuffer byteBuffer = retainableByteBuffer.getByteBuffer();

        int read;
        try
        {
            BufferUtil.clearToFill(byteBuffer);
            read = read(channel, byteBuffer);
            BufferUtil.flipToFlush(byteBuffer, 0);
        }
        catch (Throwable x)
        {
            retainableByteBuffer.release();
            return failure(x);
        }
```

could become
```java
        RetainableByteBuffer retainableByteBuffer = byteBufferPool.acquire(getBufferSize(), isUseDirectByteBuffers());

        int read;
        try
        {
            read = retainableByteBuffer.fill(channel);
        }
        catch (Throwable x)
        {
            retainableByteBuffer.release();
            return failure(x);
        }
```

The `RetainableByteBuffer` could then be renamed `Buffer` and have methods like:
 + `Buffer.fill(Channel)` 
 + `Buffer.fill(InputStream)` 
 + `Buffer.fill(EndPoint)` 
 + `Buffer.fill(Buffer)` 
 + `Buffer.fill(ByteBuffer)` 
 + `Buffer.fill(byte[],int,int)` 

But that is a different PR :)

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/transport/internal/HttpReceiverOverHTTP.java
```diff
@@ -66,7 +66,7 @@ public HttpReceiverOverHTTP(HttpChannelOverHTTP channel)
             parser.setHeaderCacheSize(httpTransport.getHeaderCacheSize());
             parser.setHeaderCacheCaseSensitive(httpTransport.isHeaderCacheCaseSensitive());
         }
-        retainableByteBufferPool = httpClient.getByteBufferPool().asRetainableByteBufferPool();
+        retainableByteBufferPool = httpClient.getRetainableByteBufferPool();
```
gregw, line 69: Why is the `networkBuffer` not released in the `failAndClose` method?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/HttpOutput.java
```diff
@@ -1025,15 +1027,16 @@ private void print(String s, boolean eoln) throws IOException
 
         CharBuffer in = CharBuffer.wrap(s);
         CharBuffer crlf = eoln ? CharBuffer.wrap("\r\n") : null;
-        ByteBuffer out = _byteBufferPool.acquire((int)(1 + (s.length() + 2) * encoder.averageBytesPerChar()), false);
-        BufferUtil.flipToFill(out);
+        RetainableByteBuffer out = _bufferPool.acquire((int)(1 + (s.length() + 2) * encoder.averageBytesPerChar()), false);
```
gregw, line 1029: This acquire needs to be released in a try finally block, as the loop below can throw and not release.

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/HttpOutput.java
```diff
@@ -1739,43 +1744,38 @@ protected Action process() throws Exception
                 if (!_closed)
                 {
                     _closed = true;
-                    _channel.getByteBufferPool().release(_buffer);
+                    _buffer.release();
```
gregw: It might be more symmetric to release in `onCompleteSuccess`?

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/HttpOutput.java
```diff
@@ -1114,19 +1115,22 @@ else if (crlf != null && crlf.hasRemaining())
...
+                BufferUtil.flipToFill(bigger.getByteBuffer());
+                bigger.getByteBuffer().put(byteBuffer);
+                out.release();
+                BufferUtil.flipToFill(bigger.getByteBuffer());
                 out = bigger;
+                byteBuffer = bigger.getByteBuffer();
                 continue;
             }
 
             result.throwException();
```
gregw: need to release on this throw

gregw (changes requested):
First pass.    

Generally i think we need to review all our acquires and comment how they are released if it is not obvious.

jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jetty-common/src/test/java/org/eclipse/jetty/ee9/websocket/common/MessageOutputStreamTest.java
lachlan-roberts: But why not just update `TestableLeakTrackingBufferPool` to be a `RetainableByteBufferPool` then other tests can potentially use it.

jetty-core/jetty-fcgi/jetty-fcgi-client/src/main/java/org/eclipse/jetty/fcgi/generator/ServerGenerator.java
```diff
@@ -83,50 +83,48 @@ public Result generateResponseHeaders(int request, int code, String reason, Http
         // End of headers
         length += EOL.length;
 
-        ByteBuffer buffer = acquire(length);
-        BufferUtil.clearToFill(buffer);
+        ByteBuffer byteBuffer = BufferUtil.allocate(length, isUseDirectByteBuffers());
```
lorban, line 86: Why isn't this buffer coming from the pool anymore?

jetty-core/jetty-http2/jetty-http2-common/src/main/java/org/eclipse/jetty/http2/internal/parser/HeaderBlockFragments.java
lorban: @sbordet I think we need to document the retain/release doctrine we decided to go with, as it should answer this kind of question that we're going to ask again and again over time as we forget then remind us how this works. I'm no sure where to write that, though.

@gregw I checked this particular case and it looks correct as the retain/release doctrine is well followed. Its idea is rather simple: if you got the RBBP from a method that returned it to you, you're now responsible for releasing it. But if you got it from a method argument, it's the responsibility of the caller to release it. In the latter case, you know that upon return the caller is going to release the RBBP so if you haven't consumed its bytes b/c you want to store it for consuming it later, you have to call retain on it and make sure a matching release call will happen, still following the same retain/release doctrine.

lorban (changes requested):
I have found one last problem in fcgi, but we should review the places where Greg found that we may leak a buffer.
And we should document the retain/release doctrine somewhere.

jetty-core/jetty-http2/jetty-http2-common/src/main/java/org/eclipse/jetty/http2/internal/parser/HeaderBlockFragments.java
gregw: Doctrine sounds good.

sbordet: I added the doctrine to the `Retainable` javadocs.

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/transport/internal/HttpReceiverOverHTTP.java
sbordet: Because `failAndClose()` is called by parser callbacks. After these callbacks return, we are back into `parseAndFill()` where we either continue or stop, but in both cases the `networkBuffer` is taken care of.

Put it in another way, the only place where the `networkBuffer` is acquired is within `parseAndFill()`, and there it is also where it is released.

jetty-core/jetty-fcgi/jetty-fcgi-client/src/main/java/org/eclipse/jetty/fcgi/generator/ServerGenerator.java
sbordet: Because in FastCGI the headers are sent back to the client as if they were content.
This method later calls `generateContent()` which is also used elsewhere for real content (not headers), so this call needed an extra `boolean` parameter to say whether the "content" must be released or not, and this was the only place where this was true.

Rather than carrying an extra parameter, I decided to just allocate on-the-fly as if it was any other content for simplicity.
Also, being server-side FastCGI, not worried about performance.

lorban (approved):

gregw (approved):

PR #9199, opened 2023-01-23
use service loader with gziphandler

### LINKED ISSUE #8769: Introduce CompressionHandler to support compression from gzip, brotli, and zstandard
**Jetty version(s)**
Jetty 12

**Enhancement Description**
The current `GzipHandler` does a good job at handling `Content-Encoding: gzip` (and `Accept-Encoding: gzip`), but we shouldn't hardcode only gzip support.

Can we refactor the existing `GzipHandler` to be more generic, something like `DynamicCompressionHandler` and have the Gzip support (courtesy of the existing `GzipRequest` object) just be loaded via a ServiceLoader?

This way we can encourage other compression algorithms under the same `Accept-Encoding` / `Content-Encoding` / already compressed rules that the existing codebase handles for us?

Other compression algorithms that have been requested in our issue tracker.

* #2553 - [RFC 7932](https://www.rfc-editor.org/rfc/rfc7932) - Brotli Compressed Data Format 
* #6600 - [RFC 8878](https://www.rfc-editor.org/rfc/rfc8878) - Zstandard Compression and the `application/zstd` Media Type

(FYI, `curl` supports both of these compression types now)

jimexist:
- closes #8769

### DIFF
--- a/documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java
+++ b/documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java
@@ -16,6 +16,7 @@ package org.eclipse.jetty.docs.programming.server.http;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.EnumSet;
+import java.util.ServiceLoader;
 import java.util.TimeZone;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -64,6 +65,7 @@ import org.eclipse.jetty.server.handler.RequestLogHandler;
 import org.eclipse.jetty.server.handler.ResourceHandler;
 import org.eclipse.jetty.server.handler.SecuredRedirectHandler;
 import org.eclipse.jetty.server.handler.StatisticsHandler;
+import org.eclipse.jetty.server.handler.compression.DynamicCompressionHandler;
 import org.eclipse.jetty.server.handler.gzip.GzipHandler;
 import org.eclipse.jetty.servlet.DefaultServlet;
 import org.eclipse.jetty.servlet.FilterHolder;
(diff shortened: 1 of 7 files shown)


joakime:
There's no test cases show how this would be used.

Can you create a ExampleCompression whatever in the src/test/java and src/test/resources/META-INF/services/ to show it working?  The compression doesn't have to do actual compression, heck, make it just ROT13 for purposes of the test. (that way we can see it performing the transformation and back again)

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java
```diff
@@ -876,7 +878,7 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
         shopContext.setHandler(new ShopHandler());
 
         // You want to gzip the shop web application only.
-        GzipHandler shopGzipHandler = new GzipHandler();
+        DynamicCompressionHandler shopGzipHandler = ServiceLoader.load(DynamicCompressionHandler.class).iterator().next();
```
janbartel, line 881: I can't see the benefit of using the `ServiceLoader` this way. A user can just extend `GzipHandler` and then call `ServletContextHandler.setGzipHandler`.

Can you explain the problem you're trying to solve a bit more?

janbartel (changes requested):

gregw:
Sorry we have not acted on this PR for while. Busy month.
Will tune in next month.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/http/HTTPServerDocs.java
jimexist: specific approach inspired by original issue https://github.com/eclipse/jetty.project/issues/8769 but personally i think it's okay to have it either way as long as interface is consistent

PR #9201, opened 2023-01-24
Jetty 12 - Cleanup Shutdown classes

### LINKED ISSUE #9200: Jetty 12 - Clean up shutdown / stop of server
**Jetty version(s)**
Jetty 12

**Description**
In Jetty 12 we have the following classes for triggering shutdown / stop of the server.

**`org.eclipse.jetty.ee8.nested.ShutdownHandler`**
    an ee8 `HandlerWrapper` - responds to HTTP requests like `/shutdown?token=blah`

**`org.eclipse.jetty.ee9.nested.ShutdownHandler`**
    an ee9 `HandlerWrapper` (seemingly implemented, with test case) - responds to HTTP requests like `/shutdown?token=blah`

**`org.eclipse.jetty.server.handler.ShutdownHandler`**
    a jetty-core `Handler.Wrapper` (mostly commented out, and disabled test cases).

**`org.eclipse.jetty.server.handler.GracefulShutdownHandler`**
    a jetty-core `Handler.Wrapper` for graceful handling for requests going thru it, not actually involved in triggering stop or shutdown, but rather cooperates with the `Graceful` behaviors triggered elsewhere.

**`org.eclipse.jetty.server.ShutdownMonitor`**
    a jetty-core simple `ServerSocket` that only responds to simple strings (`stop` / `forcestop` / `stopexit` / `exit` / `status` / `pid`) - this is also the central gathering place for any `LifeCycle` in Jetty that wants to stop with the Server but isn't registered as a bean of the Server

What I see here ...

* We have a whole JVM shutdown concept in `ShutdownMonitor`
* We have a Server instance specific stop/shutdown logic in `ShutdownHandler` (in jetty-core)
* We have extra `ShutdownHandler` instances in the environments that seem to have no purpose (lets get rid of them)
* We have a poorly named `GracefulShutdownHandler` that should probably be renamed to just `GracefulHandler` to avoid it being associated with this triggering of stop / shutdown.
* The registration of arbitrary `LifeCycle` in `ShutdownMonitor` isn't mimicked in `ShutdownHandler`

joakime:
* Fixed `ShutdownHandler` in jetty-core (with test cases)
* Removed ee9 `ShutdownHandler`
* Renamed `GracefulShutdownHandler` to just `GracefulHandler`

Fixes #9200


jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulHandler.java
```diff
@@ -28,14 +28,14 @@
 /**
  * Handler to track active requests and allow them to gracefully complete.
  */
-public class GracefulShutdownHandler extends Handler.Wrapper implements Graceful
```
gregw, line 31: Do we have a module and xml to add this to a jetty-home?  If not, we probably need one, if so, then it needs to have the rename.

gregw (changes requested):
I think we need a graceful module

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulHandler.java
joakime: Not that I can find.

Let me add a module to this PR then.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ShutdownHandler.java
```diff
@@ -13,46 +13,52 @@
...
- * <pre>
+ * <pre>{@code
  * public static void attemptShutdown(int port, String shutdownCookie) {
- * try {
- * URL url = new URL("http://localhost:" + port + "/shutdown?token=" + shutdownCookie);
- * HttpURLConnection connection = (HttpURLConnection)url.openConnection();
+ *   try {
+ *     URI uri = URI.create("http://localhost:%d/shutdown?token=%s".formatted(port, shutdownCookie));
+ *     HttpURLConnection connection = (HttpURLConnection)url.openConnection();
  * connection.setRequestMethod("POST");
```
joakime: I need to update this javadoc too.
And the jetty-documentation adoc too.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulHandler.java
joakime: Done.

gregw (approved):

jetty-core/jetty-server/src/main/config/etc/jetty-graceful.xml
```diff
@@ -0,0 +1,14 @@
...
+<!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "https://www.eclipse.org/jetty/configure_10_0.dtd">
+
+<!-- =============================================================== -->
+<!-- Mixin the Graceful Handler                                      -->
+<!-- =============================================================== -->
+
+<Configure id="Server" class="org.eclipse.jetty.server.Server">
+  <Call name="insertHandler">
+    <Arg>
+      <New id="StatsHandler" class="org.eclipse.jetty.server.handler.GracefulHandler" />
```
lorban: Do we want to keep the id as `StatsHandler` or was that overlooked?

lorban (approved):

jetty-core/jetty-server/src/main/config/etc/jetty-graceful.xml
joakime: Good catch. fixing it.

PR #9202, opened 2023-01-24
Jetty 12 - Add and allow plural form of existing `start.jar` args.

joakime:
+ new `--modules=<name>,<name>`
+ new `--libs=<classpath>`
+ updated usage.txt
+ updated warning output hint


gregw (approved):
LGTM

PR #9205, opened 2023-01-25
Jetty 12 - New HTTP Cookie interface

gregw:
This PR replaces the HttpCookie class with an interface, that allows us to facade the servlet API cookies to our API, rather than copy over the state.
Furthermore, this also embraces the "everything is an attribute" approach taken by the servlet API cookies, so we no longer have the difference in the attribute sets from a standard cookie vs our own cookie.   This greatly reduces (but not eliminates) the copying of attribute maps.
Also, the hacks for extracting attributes from comments have been moved to be ee8/ee9 only

### DIFF
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
@@ -77,7 +77,7 @@ public interface HttpCookie
             return this.attributeValue;
         }
 
-        private static Index<SameSite> CACHE = new Index.Builder<SameSite>()
+        private static final Index<SameSite> CACHE = new Index.Builder<SameSite>()
             .caseSensitive(false)
             .with(NONE.attributeValue, NONE)
             .with(STRICT.attributeValue, STRICT)
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
@@ -18,11 +18,11 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.TreeMap;
-import java.util.concurrent.TimeUnit;
 
 import org.eclipse.jetty.util.Attributes;
-import org.eclipse.jetty.util.NanoTime;
+import org.eclipse.jetty.util.Index;
 import org.eclipse.jetty.util.QuotedStringTokenizer;
 import org.eclipse.jetty.util.StringUtil;
 import org.slf4j.Logger;
@@ -31,31 +31,36 @@ import org.slf4j.LoggerFactory;
 /**
  * Jetty Management of RFC6265 HTTP Cookies (with fallback support for RFC2965)
  */
-public class HttpCookie
+public interface HttpCookie
 {
-    private static final Logger LOG = LoggerFactory.getLogger(HttpCookie.class);
+    Logger LOG = LoggerFactory.getLogger(HttpCookie.class);
     
-    private static final String __COOKIE_DELIM = "\",;\\ \t";
-    private static final String __01Jan1970_COOKIE = DateGenerator.formatCookieDate(0).trim();
+    String __COOKIE_DELIM = "\",;\\ \t";
+    String __01Jan1970_COOKIE = DateGenerator.formatCookieDate(0).trim();
 
-    /**
-     * If this string is found within the comment parsed with {@link #isHttpOnlyInComment(String)} the check will return true
-     **/
-    public static final String HTTP_ONLY_COMMENT = "__HTTP_ONLY__";
-    /**
-     * These strings are used by {@link #getSameSiteFromComment(String)} to check for a SameSite specifier in the comment
-     **/
-    private static final String SAME_SITE_COMMENT = "__SAME_SITE_";
-    public static final String SAME_SITE_NONE_COMMENT = SAME_SITE_COMMENT + "NONE__";
-    public static final String SAME_SITE_LAX_COMMENT = SAME_SITE_COMMENT + "LAX__";
-    public static final String SAME_SITE_STRICT_COMMENT = SAME_SITE_COMMENT + "STRICT__";
+    String COMMENT_ATTRIBUTE = "Comment";
+    String DOMAIN_ATTRIBUTE = "Domain";
+    String HTTP_ONLY_ATTRIBUTE = "HttpOnly";
+    String MAX_AGE_ATTRIBUTE = "Max-Age";
+    String PATH_ATTRIBUTE = "Path";
+    String SAME_SITE_ATTRIBUTE = "SameSite";
+    String SECURE_ATTRIBUTE = "Secure";
+    Index<String> KNOWN_ATTRIBUTES = new Index.Builder<String>().caseSensitive(false)
+        .with(COMMENT_ATTRIBUTE)
+        .with(DOMAIN_ATTRIBUTE)
+        .with(HTTP_ONLY_ATTRIBUTE)
+        .with(MAX_AGE_ATTRIBUTE)
+        .with(PATH_ATTRIBUTE)
+        .with(SAME_SITE_ATTRIBUTE)
+        .with(SECURE_ATTRIBUTE)
+        .build();
 
     /**
      * Name of context attribute with default SameSite cookie value
      */
-    public static final String SAME_SITE_DEFAULT_ATTRIBUTE = "org.eclipse.jetty.cookie.sameSiteDefault";
+    String SAME_SITE_DEFAULT_ATTRIBUTE = "org.eclipse.jetty.cookie.sameSiteDefault";
 
-    public enum SameSite
+    enum SameSite
     {
         NONE("None"),
         STRICT("Strict"),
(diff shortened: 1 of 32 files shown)


jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
```diff
@@ -72,313 +77,287 @@ public String getAttributeValue()
...
-        this(name, value, domain, path, maxAge, httpOnly, secure, null, 0);
+        public static SameSite from(String sameSite)
+        {
+            if (sameSite == null)
+                return null;
+            return switch (StringUtil.asciiToLowerCase(sameSite))
+            {
+                case "lax" -> SameSite.LAX;
+                case "strict" -> SameSite.STRICT;
+                case "none" -> SameSite.NONE;
```
joakime: I can imagine that the SameSite stuff is used very often in production, we should probably use one of our fancy Trie implementations for this step, no?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
```diff
@@ -72,313 +77,287 @@ public String getAttributeValue()
...
+        return new Immutable(name, value, 0, attributes);
     }
 
     /**
      * Create new HttpCookie from specific values and attributes.
      *
      * @param name the name of the cookie
      * @param value the value of the cookie
-     * @param version the version of the cookie (not used in RFC6265 or Servlet 6+, only in RFC2965 mode)
+     * @param version the version of the cookie (only used in RFC2965 mode)
```
joakime, line 126: This javadoc change worries me.
Jetty, running on Servlet 6, can still be configured for RFC2965, but this field (version) won't make it through the Servlet API down to Jetty.
That's why I worded the original Javadoc the way I did.

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/HttpCookieTest.java
```diff
@@ -94,34 +90,34 @@ public void testSetRFC2965Cookie() throws Exception
     {
         HttpCookie httpCookie;
 
-        httpCookie = new HttpCookie("null", null, null, null, -1, false, false, null, -1);
-        assertEquals("null=", httpCookie.getRFC2965SetCookie());
+        httpCookie = HttpCookie.from("null", null, -1, Map.of(HttpCookie.MAX_AGE_ATTRIBUTE, Long.toString(-1), HttpCookie.HTTP_ONLY_ATTRIBUTE, Boolean.toString(false), HttpCookie.SECURE_ATTRIBUTE, Boolean.toString(false)));
```
joakime, line 93: Feels like this should have been a builder pattern.

``` java
httpCookie = HttpCookie.from("null", null)
   .maxAge(-1)
   .httpOnly(false)
   .secure(false)
   .build()
assertEquals("null=", HttpCookie.getRFC2965SetCookie(httpCookie));
```

or at least, this specific example should be using all defaults. (i would expect maxAge, httpOnly, and secure, to be defaulted where they wouldn't be set unless specified)

``` java
httpCookie = HttpCookie.from("null", null)
assertEquals("null=", HttpCookie.getRFC2965SetCookie(httpCookie));
```

jetty-core/jetty-session/src/main/java/org/eclipse/jetty/session/AbstractSessionManager.java
```diff
@@ -244,6 +239,11 @@ public void doStart() throws Exception
         _context = ContextHandler.getCurrentContext();
         _loader = Thread.currentThread().getContextClassLoader();
 
+        // ensure a sessiun path is set for non root contexts
+        String contextPath = _context == null ? "/" : _context.getContextPath();
+        if (!"/".equals(contextPath) && getSessionPath() == null)
+            setSessionPath(contextPath);
```
joakime, line 245: This seems unrelated to this HttpCookie effort

jetty-core/jetty-session/src/main/java/org/eclipse/jetty/session/AbstractSessionManager.java
```diff
@@ -244,6 +239,11 @@ public void doStart() throws Exception
         _context = ContextHandler.getCurrentContext();
         _loader = Thread.currentThread().getContextClassLoader();
 
+        // ensure a sessiun path is set for non root contexts
```
joakime: Typo `s/sessiun/session/`

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiResponse.java
```diff
@@ -646,4 +642,92 @@ public void setTrailerFields(Supplier<Map<String, String>> trailers)
...
+        @Override
+        public int getVersion()
+        {
+            return _cookie.getVersion();
+        }
+
+        @Override
+        public SameSite getSameSite()
+        {
+            return SameSite.from(getAttributes().get("SameSite"));
```
joakime: Not using the constant from HttpCookie.SAME_SITE_ATTRIBUTE here?

joakime (changes requested):
I like the direction this has taken, the HttpCookieFacade makes is clear how powerful this approach is.

The API could use some tweaking (builder please).
The Test case could use some cleanup to avoid setting map entries for those things that are defaulted already.
Other nits around typos, etc.

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/HttpCookieTest.java
gregw: I was originally planning on having a Builder, but then worked out that the only places I would use it would be in tests.   
The Map.of is not pretty, but it is only test code.   So it seamed pointless to add a mechanism only for testing.

jetty-core/jetty-session/src/main/java/org/eclipse/jetty/session/AbstractSessionManager.java
gregw: This was part of cleaning up the session cookie configuration.  This code was moved from every time a session cookie was created to here in doStart

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
gregw: This is core, so we shouldn't really codify any specific servlet version here.  And in fact we don't. We only look at the version in RFC2965 mode.  If Servlet 6 doesn't allow a non zero version to be set, then that is its business... it does not change the fact that we only look at it in one mode.     I felt it was clearer just to say what this class did with it.

sbordet:
I'd like to chime in with a client-side view about this PR.

I like `HttpCookie` becomes an interface, would be easier to have an implementation based on `java.net.HttpCookie`, if necessary.

However, it has become clunky to use, as it is quite common for a client to specify domain, path, expirations for "hardcoded" cookies to send to the server.
For example:

```java
HttpCookie cookie = new HttpCookie("foo", "bar");
cookie.setDomain("domain.com");
cookie.setPath("/");
cookie.setMaxAge(TimeUnit.DAYS.toSeconds(1));

httpCookieStore.add(uri, cookie);
```

IIUC, now the API is just `from(name, value, Map)`, but I would prefer a builder-style API like `HttpURI` as in:

```java
HttpCookie.from(name, value).domain("domain.com").maxAge(-1).path("/path").build();
```

Also, there is a little mess with regards to the client-side `HttpCookieStore`, which is currently in `jetty-util` and not in `jetty-http` where it should be.
However, it is currently an implementation of `java.net.CookieStore`, which again ties us to the JDK implementation as in particular it store/returns `java.net.HttpCookie` instances.

So I'm proposing to ditch that class and write a `oej.http.HttpCookieStore` interface and implementation based on our own `HttpURI` and `HttpCookie` instead of JDK classes, so if we need to change something we can (for example, the JDK implementation does not do path matching like it is supposed to, so the client needs to do a double pass at filtering the cookies to send for every request).

Thoughts?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
```diff
@@ -72,313 +76,288 @@ public String getAttributeValue()
...
-     *
-     * @param name the name of the cookie
-     * @param value the value of the cookie
-     * @param maxAge the {@code Max-Age} attribute value (in seconds) for the cookie
-     */
-    public HttpCookie(String name, String value, long maxAge)
-    {
-        this(name, value, null, null, maxAge, false, false);
-    }
+        private static Index<SameSite> CACHE = new Index.Builder<SameSite>()
```
joakime: `s/CACHE/SAMESITE_CACHE/`

joakime (changes requested):
Lots of good here.

Small change of a name of constant left

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
joakime: I wonder, can this be in the `SameSite` type itself?

gregw: But it is already in the SameSite enum, so it is `HttpCookie.SameSite.CACHE`.   I think that is fine.

... although it should have been final. I  fixed that.

sbordet:
Also, `CookieCache` and `CookieCutter` can probably be united into `CookieParser` (a better name for `CookieCutter`?

gregw:
@sbordet if the client needs a fluent interface, then we can easily add one. But can it be done in a different PR rather than keeping this one unmerged for even longer?

Only issue is the class with the `from` name, but elsewhere we have used `build` so we can easily add:

```java
    HttpCookie cookie = HttpCookie.build("name","value").domain("domain.com").maxAge(-1).path("/path").build()
```

But I'd like to do that in a PR that actually used it, i.e one that does changes to the client.

gregw:
@sbordet again I think changes to CookieCutter should be in a different PR, especially if there is going to be work done on adding a cookie store to the client.

joakime (approved):

PR #9213, opened 2023-01-26
Converted PathMappings to be an AbstractMap

gregw:
This was primarily done to avoid surprise of the previous behavior of puts for a duplicate key being ignored. 
The use of the AbstractMap class now provides javadoc and known/expected behavior for key methods. 
The only significant behavior change is in the return type of some key methods, with the old/previous value replacing a boolean. 
Some stream usage has also been replaced by the more efficient iterator.
Some methods provided by AbstractMap may not be efficient.

### DIFF
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
@@ -14,9 +14,9 @@
 package org.eclipse.jetty.http.pathmap;
 
 import java.io.IOException;
+import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -41,10 +41,12 @@
  * @param <E> the type of mapping endpoint
  */
 @ManagedObject("Path Mappings")
-public class PathMappings<E> implements Iterable<MappedResource<E>>, Dumpable
+public class PathMappings<E> extends AbstractMap<PathSpec, E> implements Iterable<MappedResource<E>>, Dumpable
 {
+    // TODO this class should be an AbstractMap
```
joakime: Isn't this PR doing what this TODO is saying?

gregw: ooops

joakime (approved):

PR #9216, opened 2023-01-27
Refactor jetty-maven-plugin to have a wrapper on the top of eeX plugin and refactor to avoid some classes duplication

olamy:
This PR is creating some new modules with a new one from root called `jetty-maven` with 2 sub modules
- `jetty-maven-core` which contains classes used by all eeX plugin versions
- `jetty-maven-plugin` the top plugin which delegates to eeX depending configuration


jetty-maven/jetty-maven-core/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-jndi</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-server</artifactId>
+      <optional>true</optional>
+    </dependency>
+<!--    <dependency>-->
```
janbartel: delete

jetty-maven/jetty-maven-core/src/main/java/org/eclipse/jetty/maven/MavenWebAppContext.java
```diff
@@ -0,0 +1,236 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.eclipse.jetty.util.resource.CombinedResource;
+
+/**
+ * MavenWebAppContext
+ *
+ * Extends the WebAppContext to specialize for the maven environment. We pass in
```
janbartel: The comments need updating - no longer an extension of `WebAppContext`. 

I think this is going to be a difficult approach - you will now need to copy all contents over into a real `WebAppContext`. Also, if this is what the user is configuring in their pom, then it will constrain them as to what can be configured, rather than allowing them to configure a "real" `WebAppContext`.  Additionally there is behaviour in `MavenWebAppContext` that isn't captured here.

jetty-ee10/jetty-ee10-maven-plugin/src/main/java/org/eclipse/jetty/ee10/maven/plugin/AbstractWebAppMojo.java
```diff
@@ -48,8 +48,12 @@
 import org.apache.maven.project.MavenProject;
 import org.codehaus.plexus.util.StringUtils;
 import org.eclipse.aether.RepositorySystem;
-import org.eclipse.jetty.ee10.maven.plugin.utils.MavenProjectHelper;
 import org.eclipse.jetty.ee10.servlet.security.LoginService;
+import org.eclipse.jetty.maven.MavenServerConnector;
```
janbartel, line 52: This class needs to extend `AbstractWebAppMojo` from `jetty-maven-core` doesn't it?

olamy: correct I haven't finished everything yet and it's more removal coming soon ;)

jetty-maven/jetty-maven-core/src/main/java/org/eclipse/jetty/maven/MavenWebAppContext.java
olamy: the copy is done automatically so it's not the problem.
Reducing the surface (e.g access to all fields/methods of `WebAppContext`) a  is a bit on purpose (`by design`).
The idea is to have now only access to a limited set of fields because between ee9, ee10 and future eeX those fields may change and we want to give access to only important fields (and this give us more freedom to change `WebAppContext` without breaking users).
Bear in mind this Maven plugin is used for development only and if users have complicated setup they will need to put it in a `jetty-web.xml` file for production so they can use and validate the file with `jetty-maven-plugin` rather than duplicate the content in both `pom.xml` and `jetty-web.xml`. (and that's what we recommend in some javadoc)

jetty-maven/jetty-maven-core/pom.xml
olamy: done

jetty-maven/jetty-maven-core/src/main/java/org/eclipse/jetty/maven/MavenWebAppContext.java
janbartel: It's a design choice certainly, but that hasn't been the philosophy to date. We've permitted users to directly configure their `WebAppContext` in the pom without forcing them to create a context xml file so some users will be caught out. Also I think this is a fragile change: we add methods to `WebAppContext` in eg `ee10` that we want exposed to the user to configure but because `MavenWebAppContext` does not inherit from `WebAppContext` then they never get found by the ide refactoring tools. The whole point of the plugin is to configure and run a webapp and we want users to be able to do that as easily as possible.

olamy: should be better now.
yup for sure it looks convenient to configure with the pom. But as I tried to explain previously if the users need to do some extra/sophisticated they will have to do it both in the pom and the jetty-web.xml as users will have to provide a jetty-web.xml when deploying the webapp in a production environment. So having to maintain 2 xml configuration with totally different syntax might not be very convenient.

janbartel: I'm not sure we're talking about the same things here. Currently users can configure their webapp in maven in any combination of the following:

+ the `WebAppContext` in the `pom.xml`
+ the `WebAppContext` via a normal context xml file
+ the `WebAppContext` via a normal `WEB-INF/jetty-web.xml` file

olamy: we do :) 
But as you just described with configuration in `pom.xml` which have a totally different syntax from normal context xml or `WEB-INF/jetty-web.xml`, we just give users the option to configure the exact same thing in 2 different files with a totally different format whereas at the end of day they need only one format when deploying in production (e.g the jetty xml format).
And by the way this give us more work to have to maintain this extra layer `MavenWebAppContext`.
I would prefer having a very minimal set of options (port, resources, etc..) well documented by javadoc in the Maven plugin (because ATM with `MavenWebAppContext` users need to navigate to javadoc of classes) in the mojo and then if users need something more fancy they just use a jetty xml file as anyway they will need it to configure their webapp when deploying in a production environment (because they can not use Jetty Maven plugin in a production environment at least I hope not but who knows :)).
I don't mind if we keep all of those options but I find this a bit confusing for users and cause of possible mistakes because they have so many options. 
But hey it was just a thought I had when working on the code. if you don't want to change that's fine we do not have to change it.

jetty-ee10/jetty-ee10-maven-plugin/src/main/java/org/eclipse/jetty/ee10/maven/plugin/AbstractWebAppMojo.java
olamy: > This class needs to extend `AbstractWebAppMojo` from `jetty-maven-core` doesn't it?

nah we can't because each single `AbstractWebAppMojo` from every eeX is using `org.eclipse.jetty.eeX.security.LoginService;` 
I'm very tempted to removed this field which is marked as deprecated as users can simply user a `jetty-web.xml` for this

olamy:
ok some more cleanup should/could/**must** be done.
but at least now we have the upper layer and user can keep using the classic coordinate of the plugin
```
          <artifactId>jetty-maven-plugin</artifactId>
```
And then it will be easier to re factor/cleanup with asking users to change the coordinate.

jetty-maven/jetty-maven-core/src/main/java/org/eclipse/jetty/maven/AbstractUnassembledWebAppMojo.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.io.File;
+
+import org.apache.maven.plugins.annotations.Parameter;
+
+/**
+ * Base class for all goals that operate on unassembled webapps.
+ *
+ */
+public abstract class AbstractUnassembledWebAppMojo extends AbstractWebAppMojo
```
janbartel: There's something not quite right with the inheritance hierarchy here. There's no point in having this class in jetty-maven-core if the only thing that uses it is in jetty-maven-plugin.

jetty-maven/jetty-maven-core/src/main/java/org/eclipse/jetty/maven/AbstractWebAppMojo.java
```diff
@@ -0,0 +1,461 @@
...
+import org.eclipse.jetty.server.RequestLog;
+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.server.handler.ContextHandler;
+import org.eclipse.jetty.util.resource.Resource;
+import org.twdata.maven.mojoexecutor.MojoExecutor;
+
+/**
+ * AbstractWebAppMojo
+ *
+ * Base class for common behaviour of jetty mojos.
```
janbartel: Similarly to AbstractUnassembledWebAppMojo, the only thing that uses this class is now in the jetty-maven-plugin module so not much point in having it in jetty-maven-core.

jetty-maven/jetty-maven-core/src/main/java/org/eclipse/jetty/maven/AbstractWebAppMojo.java
```diff
@@ -0,0 +1,461 @@
...
+            {
+                //remove environment mojo
+                if (!"mojo".equals(child.getName()) && !"environment".equals(child.getName()))
+                {
+                    newConfiguration.addChild(child);
+                }
+            }
+
+            Thread.currentThread().setContextClassLoader(classRealm);
+            MojoExecutor.executeMojo(
```
janbartel: IIUC what's happening here, we're copying the configuration from the core UnassembledWebAppMojo and AbstractWebAppMojo onto the UnassembledWebAppMojo and AbstractWebAppMojo that are in the specific eeX plugin version?

If so, I don't like having those core classes duplicated in every eeX module - they must be exactly identical in terms of declared parameters otherwise configuration info is missed and can cause jetty to fail in unexpected and hard to detect kind of ways.

olamy: please read my comment here https://github.com/eclipse/jetty.project/pull/9216#discussion_r1091387036 about `LoginService` without this field which is already sort of deprecated in jetty-10.0.x (https://github.com/eclipse/jetty.project/blob/ca19b143a9de1c222b517028afc118482f7c093b/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/AbstractWebAppMojo.java#L242) we could get rid of a lot duplicated classes.

jetty-maven/jetty-maven-core/src/main/java/org/eclipse/jetty/maven/AbstractUnassembledWebAppMojo.java
olamy: https://github.com/eclipse/jetty.project/pull/9216#discussion_r1095253119

jetty-maven/jetty-maven-core/src/main/java/org/eclipse/jetty/maven/AbstractWebAppMojo.java
olamy: https://github.com/eclipse/jetty.project/pull/9216#discussion_r1095253119

jetty-maven/jetty-maven-core/src/main/java/org/eclipse/jetty/maven/AbstractWebAppMojo.java
olamy: And BTW I explained in my comment why we can still have duplicated classes (frankly there are already so many ;) ) https://github.com/eclipse/jetty.project/pull/9216#issuecomment-1414563027
BUT the most important is to have this extra layer on the top and this unique groupId/artifactId which is backward compatible.

olamy: and this PR is already removing some duplication :) almost 600 line of code from duplicated classes 
Screenshot 2023-02-03 at 11 32 59 am

jetty-maven/jetty-maven-core/src/main/java/org/eclipse/jetty/maven/AbstractWebAppMojo.java
olamy: the goal was to have it in jetty-maven-core to reuse it in eeX-maven (when we will be able)
But anyway I can move this class to jetty-maven-plugin module

jetty-maven/jetty-maven-core/src/main/java/org/eclipse/jetty/maven/AbstractUnassembledWebAppMojo.java
olamy: https://github.com/eclipse/jetty.project/pull/9216#discussion_r1095263999

jetty-maven/jetty-maven-core/src/main/java/org/eclipse/jetty/maven/AbstractWebAppMojo.java
olamy: class moved here 4435939a33b8f1b1dc4c5d1d350f13396b520a67

jetty-maven/jetty-maven-core/src/main/java/org/eclipse/jetty/maven/AbstractUnassembledWebAppMojo.java
olamy: class moved here 4435939a33b8f1b1dc4c5d1d350f13396b520a67

jetty-maven/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/AbstractWebAppMojo.java
```diff
@@ -0,0 +1,463 @@
...
+ * AbstractWebAppMojo
+ *
+ * Base class for common behaviour of jetty mojos.
+ */
+public abstract class AbstractWebAppMojo extends AbstractMojo
+{
+
+    @Inject
+    protected BuildPluginManager buildPluginManager;
+
```
janbartel, line 68: All of the parameters that are declared in this class are duplicated in each of the `ee8/9/10` classes also called `AbstractWebAppMojo` - that seems undesirable because: a) duplication is bad b) you can't add any new parameter to the `ee` specific versions of this class because it's unconfigurable. 

I still think there's something wrong with the inheritance hierarchy because ISTM more natural that the `ee` specific classes would inherit off these common base classes rather than have nothing to do with them.

jetty-maven/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/AbstractWebAppMojo.java
```diff
@@ -0,0 +1,463 @@
...
+    /**
+     * An instance of org.eclipse.jetty.eeXX.webapp.WebAppContext that represents the webapp.
+     * Use any of its setters to configure the webapp. This is the preferred and most
+     * flexible method of configuration, rather than using the (deprecated) individual
+     * parameters like "tmpDirectory", "contextPath" etc.
+     * 
+     */
+    @Parameter(name = "webApp")
+    protected PlexusConfiguration webApp;
```
janbartel, line 124: The comment is wrong here, it's not a `WebAppContext`.

Is the idea that you can use this `PlexusConfiguration` to hold any old `name,value` pairs, and that during the invocation of the target plugin, any `name,value` pairs that match those of the setters on the target `WebAppContext` will be transferred?

 In that case my reservation is that there's no type checking on what the user is configuring in the pom.xml. If we change eg `ee10` `WebAppContext` then the user won't get an error.

janbartel (commented):
I'm all in favour of refactoring to reduce duplicate code.  So moving some of the classes into `jetty-maven-core` is good, and then extending those classes in the `ee` impls would be good.

However, trying to go the extra step of having only a single plugin exposed to the user, irrespective of which `ee` environment they're targetting, is problematic because it involves _additional_ duplication of code between `core` and `ee` in order to expose necessary configuration to the user.

jetty-maven/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/AbstractWebAppMojo.java
olamy: > All of the parameters that are declared in this class are duplicated in each of the `ee8/9/10` classes also called `AbstractWebAppMojo` - that seems undesirable because: a) duplication is bad b) you can't add any new parameter to the `ee` specific versions of this class because it's unconfigurable.
> 

I explained in this comment: https://github.com/eclipse/jetty.project/pull/9216#discussion_r1091387036

> I still think there's something wrong with the inheritance hierarchy because ISTM more natural that the `ee` specific classes would inherit off these common base classes rather than have nothing to do with them.

what could be so specific for a user? 
We are talking here about a development maven plugin users just neeed to run a webapp module in development not something used in production. 
In case a user really need something specific, he will write a some `jetty-env.xml` or `jetty-webapp.xml` because he will need it anyway for deploying in a jetty distribution.
I would prefer we provide a minimum of usable options for users to configure in the pom (we can't even configure the port in the pom and we are talking about specific eeX configuration maybe used by few users)

jetty-maven/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/AbstractWebAppMojo.java
olamy: > The comment is wrong here, it's not a `WebAppContext`.
> 
> Is the idea that you can use this `PlexusConfiguration` to hold any old `name,value` pairs, and that during the invocation of the target plugin, any `name,value` pairs that match those of the setters on the target `WebAppContext` will be transferred?
> 
> In that case my reservation is that there's no type checking on what the user is configuring in the pom.xml. If we change eg `ee10` `WebAppContext` then the user won't get an error.

maven will still check when running the target mojo.

but yes the problem is this inheritance of `org.eclipse.jetty.ee9.maven.plugin.MavenWebAppContext extends WebAppContext` (same for ee10)
because classes which are mojo fields should not inherit but should be only bean carrying some configuration which would be injected when running the target eeX. 
So this would simplify a lot of code which is already duplicated (not only here) and simplify the code.. 
With exposing those classes we create some problems/maintenance burden to ourself as every changes in the parent class `WebAppContext` potentially create issues for users. 
We should hide that and simplify the configuration (again users wants some very specific they can still use a `jetty-**.xml` file)

gregw:
@olamy I know nothing much about maven, so this may not be that helpful.

However, the general approach in jetty has always been to avoid configuration objects/interfaces/classes/mechanisms.      We did have some in the early days, but they were problematic and we have spent a lot of time getting rid of them.      Jetty is a server that is assembled from a bunch of POJOs that need to be injected with configuration.   You can do that with code, with jetty XML, with spring beans, any other dependency injection framework or, in the case of the maven plugin, with maven pom.xml.

If we create a configuration interface/class common to EE8, EE9, EE10, then there is no guarantee that there will be meaningful configuration interpretations that work for EE11, EE12, EE13 etc.     We will either be restricted in what configurations we can add, or we will need to come up with backwards compatibility mappings for future configurations. 
I'm also very much against requiring configuration. For simple webapps, we should just be able to list the jetty-maven-plugin in the pom.xml and do a "mvn jetty:run".  

On the flip side, I'm very much in favour of finding ways to reduce/avoid code duplication.  But the idea of injecting some MOJO so that we then have to copy over all the configuration to specific EE classes doesn't sound like it is avoiding duplication. It is just changing the code that we duplicate.   I'm also not keen to change our users experience just for the sake of us avoiding code duplication.    If it is trivial duplication, then we can automate; if it is needless duplication then we can factor out common classes.

Let's schedule a discussion on this after the next release.

joakime:
Since this is out of date, and we are in week 4, i'm pushing this to the 12.0.2 release.

olamy:
I think it's better to close it for now...

There will be still this warning. 
```
[INFO] --- plugin:3.9.0:helpmojo (exec-plugin-doc) @ jetty-ee10-maven-plugin ---
[WARNING] 

Goal prefix is specified as: 'jetty'. Maven currently expects it to be 'jetty-ee10'.

```

PR #9220, opened 2023-01-27
Fix/jetty 12 reorganize session tests

sbordet:


sbordet:
@janbartel this PR reorganizes the session tests, by putting them under the `jetty-eeN-tests` parent, and renaming the directories/modules according to the convention we have everywhere else.

However, for ee8 the source files were actually there, rather than being generated from ee9 like all others.
Also in ee8, the `jetty-ee8-tests` directory is completely missing, so we don't run any of the tests under that directory that we run for ee9.
Given this, I have removed entirely the ee8 session tests given that the run them already for ee9 (like we do for all the others).

Happy to change to what you think it's better.

jetty-ee9/pom.xml
```diff
@@ -25,7 +25,6 @@
     <jakarta.annotation.api.version>2.0.0</jakarta.annotation.api.version>
     <jakarta.authentication.api.version>2.0.0</jakarta.authentication.api.version>
     <jakarta.el.api.version>4.0.0</jakarta.el.api.version>
-    <jakarta.enterprise.cdi.api.version>3.0.1</jakarta.enterprise.cdi.api.version>
```
janbartel: Extraneous change.

janbartel (approved):
Just remove the extraneous change with the cdi property and it's good.

Just leaving this comment here, nothing for you to action in this PR: i think in our renaming, prefixing everything with `jetty` has been overkill - of course it's going to be a jetty artifact, what else would it be :) ?

olamy:
might be good to have this as well https://github.com/eclipse/jetty.project/pull/9221.
maybe another PR? (grhh a lot merge for me 🤣)
because currently as junit/surefire only reports package name it's impossible in CI to know if failure of `org.eclipse.jetty.hazelcast.session.ClusteredOrphanedSessionTest` is in ee8 ee9 ee10 
currently only `gcloud` has a good naming.

PR #9223, opened 2023-01-30
Jetty 12 - Added a core Session abstraction

gregw:
Sessions were already a core mechanism, but there was no API for them. 

There is now a new Session interface that is available via the Request API.  It is intended only for users of sessions. 

The previous concrete class Session has been renamed to ManagedSession and it is used by classes that extend AbstractSessionManager.


jetty-core/jetty-session/src/main/java/org/eclipse/jetty/session/ManagedSession.java
```diff
@@ -339,6 +335,7 @@ public boolean isInvalid()
     {
         try (AutoLock l = _lock.lock())
         {
+            // TODO this is not the inverse of isValue???? what about CHANGING
```
janbartel: `VALID` is self evident. `INVALID` also. `CHANGING` means in the process of changing the session id and `INVALIDATING` means in the process of becoming invalid. The spec says that when session callbacks happen during invalidation, the session must still be readable, but is mute about what should happen when the session is changing its id. See `checkValidForRead` and `checkValidForWrite` which try to implement what the spec wants wrt to session validity. We could probably re-evaluate how `isValid` and `isInvalid` are used and see if it can be simplified - after this PR is done with.

jetty-core/jetty-session/src/main/java/org/eclipse/jetty/session/SessionManager.java
```diff
@@ -75,33 +76,33 @@ public interface SessionManager extends LifeCycle, SessionConfig
      */
     String __MaxAgeProperty = "org.eclipse.jetty.servlet.MaxAge";
 
-    Session getSession(String id) throws Exception;
+    ManagedSession getManagedSession(String id) throws Exception;
```
janbartel, line 79: Should these be `Session` and not `ManagedSession`?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextRequest.java
```diff
@@ -292,4 +303,81 @@ static boolean isNoParams(Fields fields)
...
+        if (httpSession == null)
+            throw new IllegalStateException("No session");
+
+        ManagedSession session = _managedSession;
+        session.renewId(this, _response);
+
+        return httpSession.getId();
+    }
+
+    public ManagedSession getManagedSession()
```
janbartel, line 306: Maybe the signature should be `Session`?

jetty-core/jetty-session/src/main/java/org/eclipse/jetty/session/ManagedSession.java
```diff
@@ -111,25 +109,22 @@ public Session(SessionManager manager, SessionData data)
...
-    }
-
-    public static Session getSession(Object session)
-    {
-        if (session instanceof APISession wrapper)
-            return wrapper.getCoreSession();
-        return null;
+        _api = _manager.newSessionAPIWrapper(this);
+        if (_api != null && _api.getSession() != this)
+            throw new IllegalStateException("APISession must wrap this session");
```
janbartel:
```suggestion
            throw new IllegalStateException("Session.API must wrap this session");
```

jetty-core/jetty-session/src/main/java/org/eclipse/jetty/session/ManagedSession.java
```diff
@@ -111,25 +109,22 @@ public Session(SessionManager manager, SessionData data)
...
+     */
+    @Override
     @SuppressWarnings("unchecked")
-    public <T> T getAPISession()
-    {
-        return (T)_apiSession;
-    }
-    
-    public void setAPISession(Object o)
+    public <T extends API> T getAPISession()
```
janbartel:
```suggestion
    public <T extends API> T getAPI()
```

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
```diff
@@ -100,28 +100,16 @@ public class ServletApiRequest implements HttpServletRequest
...
-    private boolean _requestedSessionIdFromCookie;
     private Authentication _authentication;
     private String _method;
     private ServletMultiPartFormData.Parts _parts;
-    private ServletPathMapping _servletPathMapping;
     private boolean _asyncSupported = true;
 
     protected ServletApiRequest(ServletContextRequest servletContextRequest)
     {
         this._request = servletContextRequest;
```
janbartel:
```suggestion
       _request = servletContextRequest;
```

janbartel (approved):
Couple of questions/suggestions.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Session.java
```diff
@@ -0,0 +1,74 @@
...
+
+    void setAttribute(String name, Object value);
+
+    void removeAttribute(String name);
+
+    void renewId(Request request, Response response);
+
+    void invalidate();
+
+    boolean isNew() throws IllegalStateException;
```
lachlan-roberts, line 109: Can we get javadoc for the methods in this interface.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Session.java
```diff
@@ -0,0 +1,74 @@
...
+        if (session instanceof API wrapper)
+            return wrapper.getSession();
+        return null;
+    }
+
+    /**
+     * <p>API wrappers of the {@link Session} must implement this interface and return
+     * the {@link Session} that they wrap.</p>
+     */
+    interface API
```
lachlan-roberts, line 46: Don't really like the name `API` for this, can it be something more specific.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -955,6 +956,12 @@ public void addHttpStreamWrapper(Function<HttpStream, HttpStream> wrapper)
             getHttpChannel().addHttpStreamWrapper(wrapper);
         }
 
+        @Override
+        public Session getSession(boolean create)
+        {
+            return null;
```
lachlan-roberts, line 962: Why is this not `getRequest().getSession(create)`?

jetty-core/jetty-session/src/main/java/org/eclipse/jetty/session/AbstractSessionCache.java
```diff
@@ -107,7 +108,7 @@ public abstract class AbstractSessionCache extends ContainerLifeCycle implements
      * @return a new Session object
      */
     @Override
-    public abstract Session newSession(SessionData data);
+    public abstract ManagedSession newSession(SessionData data);
```
lachlan-roberts, line 111: Why do these all need to use the `ManagedSession` class and not the `Session` interface?

jetty-core/jetty-session/src/main/java/org/eclipse/jetty/session/ManagedSession.java
gregw: I have removed a few usages of it and then renamed to `isInvalidOrInvalidating`

jetty-core/jetty-session/src/main/java/org/eclipse/jetty/session/SessionManager.java
gregw: No.  Users and/or extenders of `SessionManager` are using the concrete `ManagedSession`. For example, soon after calling `getManagedSession(id)`, the `access` method may be called on the returned session. This is not available in the public `Session` API as an application should not call `access`.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
```diff
@@ -442,48 +385,40 @@ public String changeSessionId()
         if (httpSession == null)
             throw new IllegalStateException("No session");
 
-        Session session = SessionHandler.ServletAPISession.getSession(httpSession);
+        ManagedSession session = SessionHandler.ServletAPISession.getSession(httpSession);
```
lachlan-roberts: I don't understand whats going on here, why doesn't the `SessionHandler` override the `getSession()` method on request to provide its own wrapped `Session`?

Why do we need the downcasting to ManagedSession?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextRequest.java
gregw: No.  The callers of this method are doing `ManagedSession` stuff like:
 + resetting cookies on `response.reset()` (and cookies are not a known part of the application `Session` API. Cookies are implementation details as we can track sessions with other mechanisms.
 + committing the response so they need to flush the session to any datastore. This is an implementation detail.
 + completing the response... ditto.

lachlan-roberts (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Session.java
gregw: It's not `API` it is `Session.API` :)

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
gregw: This is the base Request impl, that does not support sessions unless it is wrapped to do so.  There is no request below/under/inside this request to get.

jetty-core/jetty-session/src/main/java/org/eclipse/jetty/session/AbstractSessionCache.java
gregw: Applications use `Session`.
Implementations of `Session` use `SessionManager`s and `SessionHandler`s and `SessionCache`s and `SessionData` to implement a `ManagedSession`, thus they communicate within themselves with `ManagedSession`.

If you are just a handler/servlet that wants to use a session, you just see/use `Session` interface. But if you are implementing/extending our style of concrete session you see/use `ManagedSession`

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
gregw: not needed

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
```diff
@@ -400,33 +364,12 @@ public String getServletPath()
...
-            throw new IllegalStateException("Create session failed");
-
-        var cookie = _sessionManager.getSessionCookie(_coreSession, isSecure());
-
-        if (cookie != null)
-            Response.replaceCookie(_request.getResponse(), cookie);
-
-        return _coreSession.getAPISession();
+        if (session.isNew() && getAuthentication() instanceof Authentication.User)
+            session.setAttribute(ManagedSession.SESSION_CREATED_SECURE, Boolean.TRUE);
```
lachlan-roberts, line 371: You said you would move this attribute onto `Session` interface instead of `ManagedSession`?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiResponse.java
```diff
@@ -559,11 +560,12 @@ public void reset()
 
         _response.reset();
 
+
         ServletApiRequest servletApiRequest = _response.getServletContextRequest().getServletApiRequest();
-        Session session = servletApiRequest.getCoreSession();
+        ManagedSession session = servletApiRequest.getServletContextRequest().getManagedSession();
```
lachlan-roberts, line 565: I don't like that the servlet layer is dependent on a particular implementation of the `Session` interface.

Is it possible to generalize this so that it can all be done through the jetty-server `Session` interface so it would work if you used a different implementation of `Session`. 

Maybe `SessionManager` should be a jetty-server interface as well?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
gregw: It's a security feature. When you move ee10 security to core then I'll move this... or better yet do away with it. For now it has no good home and it's a bit of leaked implementation, so ManagedSession is probably the right place.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiResponse.java
gregw: It is probably possible. But currently ManagedSession contains servlet specific features like passivation and listeners. It would need to be split to move those boys into some servlet common module.... That we don't yet have.  So a good thing to look at, but that's a different PR i think.

lachlan-roberts (approved):
I still don't like the design that jetty-server has the `Session` interface, but any implementation you can use must be based off the `ManagedSession` from jetty-session module.

Id rather a more complete set of interfaces in jetty-server, or have jetty-session use an attribute to store session instead of having a direct getter on the jetty-server request.

But hopefully this can be improved in further PRs.

gregw:
> I still don't like the design that jetty-server has the `Session` interface, but any implementation you can use must be based off the `ManagedSession` from jetty-session module.
> 

Again this is because all our current usages are based on servlet API, which has a lot of it's implementation in ManagedSession.

When we (you) move security to core, it should just use the session interface and nothing else.

We can also eventually split the servlet stuff out of ManagedSession and put in a servlet common package.

> Id rather a more complete set of interfaces in jetty-server, or have jetty-session use an attribute to store session instead of having a direct getter on the jetty-server request.

I don't see how using an attribute would help?

jetty-core/jetty-session/src/main/java/org/eclipse/jetty/session/ManagedSession.java
```diff
@@ -526,27 +521,33 @@ public void setAttribute(String name, Object value)
...
      */
-    public void renewId(Request request)
+    @Override
+    public void renewId(Request request, Response response)
     {
         if (_manager == null)
             throw new IllegalStateException("No session manager for session " + _sessionData.getId());
 
-        String id = null;
+        if (response != null && response.isCommitted())
```
dgodet-distech, line 547: Why is it not possible to renew a session when the response is committed? Specifically, how does deferred authentication handle session renewal, given that a deferred response is always committed?

gregw: @dgodet-distech 
If the response is committed, then the response headers are already gone and we cannot do a Set-Cookie to send the new session ID to the client.

gabrielandrade2: @gregw, what is the proper way to handle the session renew when the authentication is performed on a deferred state, as the response received by the `Authenticator` is a `DeferredResponse`?

gregw: If you do a getUserPrincipal or similar call, then the deferred auth will be resolved (if possible) to a real auth.

gabrielandrade2: Thank you for the answer, @gregw.

Right, I do understand that `getUserPrincipal` will potentially resolve the authentication.

Let me explain a little better the situation we faced. 

We have a session that is created when the user is redirected to the login page (before it is authenticated). It is used to contain an attribute that we must persist at this state.
However, when we try to resolve the deferred authentication, the call to `LoginAuthentication.login()`  successfully retrieves the `UserIdentity` and, because this session exists, it will try to renew its ID (as the session was created in an unauthenticated state).
This renewal fails, as the `Response` that was passed to LoginAuthentication.login()`  is a `AuthenticationState.Deferred.DeferredResponse`, which is "committed" and cannot be written to.
In this case, how to reneaw the session id?

I am not knowledgeable enough about the authentication process to know if this is a bug or if we are doing something wrong.

PR #9225, opened 2023-01-30
Jetty 12.0.x silence tests' stacktraces

lorban:
Prevent expected stack traces to pollute the output during builds.

### DIFF
--- a/jetty-core/jetty-http2/jetty-http2-tests/src/test/java/org/eclipse/jetty/http2/tests/IdleTimeoutTest.java
+++ b/jetty-core/jetty-http2/jetty-http2-tests/src/test/java/org/eclipse/jetty/http2/tests/IdleTimeoutTest.java
@@ -17,6 +17,7 @@ import java.nio.ByteBuffer;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.eclipse.jetty.http.HttpFields;
@@ -46,6 +47,7 @@ import org.eclipse.jetty.util.thread.QueuedThreadPool;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Test;
 
+import static org.awaitility.Awaitility.await;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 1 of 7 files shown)


jetty-core/jetty-http2/jetty-http2-tests/src/test/java/org/eclipse/jetty/http2/tests/IdleTimeoutTest.java
```diff
@@ -686,51 +695,57 @@ public boolean process(Request request, Response response, Callback callback)
...
 
-        assertTrue(resetLatch.await(2 * idleTimeout, TimeUnit.MILLISECONDS));
+            assertThat(((HTTP2Session)client).updateSendWindow(0), Matchers.greaterThan(0));
 
-        // Wait for WINDOW_UPDATEs to be processed by the client.
-        sleep(1000);
-
-        assertThat(((HTTP2Session)client).updateSendWindow(0), Matchers.greaterThan(0));
+            // The server must be stopped before StacklessLogging is closed.
+            LifeCycle.stop(server);
```
sbordet: This feels really wrong.

jetty-core/jetty-http2/jetty-http2-tests/src/test/java/org/eclipse/jetty/http2/tests/IdleTimeoutTest.java
```diff
@@ -676,8 +678,15 @@ public boolean process(Request request, Response response, Callback callback)
...
-                callback.succeeded();
+                try
+                {
+                    Thread.sleep(2 * idleTimeout);
+                    callback.succeeded();
+                }
+                catch (InterruptedException e)
+                {
+                    callback.failed(e);
+                }
```
sbordet: It should not happen that this thread is interrupted... how it is the case?

sbordet (changes requested):

jetty-core/jetty-http2/jetty-http2-tests/src/test/java/org/eclipse/jetty/http2/tests/IdleTimeoutTest.java
lorban: The test ends and reaches `@AfterEach void dispose()` which stops the server before the handler's `Thread.sleep()` call returns.

jetty-core/jetty-http2/jetty-http2-tests/src/test/java/org/eclipse/jetty/http2/tests/IdleTimeoutTest.java
lorban: It's for the same reason as for your other comment: since `@AfterEach` can execute before the handler returns, `StacklessLogging` may have re-enabled stack trace logging.

jetty-core/jetty-http2/jetty-http2-tests/src/test/java/org/eclipse/jetty/http2/tests/IdleTimeoutTest.java
sbordet: Then maybe wait for the response is better?

sbordet (approved):

joakime (approved):

gregw (approved):
awesome!!!

gregw:
I'll merge so I can look for some more....

PR #9228, opened 2023-01-30
Jetty 12 - Dependency updates (with focus on isolated ee deps)

joakime:
Cleanup top level pom.xml and jetty-ee#/pom.xml to isolate ee deps properly.

Also upgrade ee10 deps that need updating.

### DIFF
--- a/javadoc/pom.xml
+++ b/javadoc/pom.xml
@@ -333,11 +333,6 @@
       <scope>provided</scope>
     </dependency>
     <!--<dependency>
-      <groupId>org.eclipse.jetty</groupId>
-      <artifactId>jetty-unixsocket-client</artifactId>
-      <scope>provided</scope>
-    </dependency>-->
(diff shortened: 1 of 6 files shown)


javadoc/pom.xml
```diff
@@ -332,11 +332,6 @@
       <artifactId>jetty-start</artifactId>
       <scope>provided</scope>
     </dependency>
-    <!--<dependency>
-      <groupId>org.eclipse.jetty</groupId>
-      <artifactId>jetty-unixsocket-client</artifactId>
-      <scope>provided</scope>
-    </dependency>-->
     <!--<dependency>
```
sbordet, line 335: I'd get rid of these other commented out dependency, they look leftovers from the 11 module names.

sbordet (approved):
A small nit, otherwise LGTM if it builds 😃

PR #9231, opened 2023-01-31
Jetty 12 - Rename sessions tests with ee# version in package name

olamy:
Rename sessions tests with ee# version in package name as it.
This way we can identify which ee# test is failing within the junit test report.


jetty-core/jetty-session/pom.xml
```diff
@@ -69,5 +69,10 @@
       <artifactId>jetty-slf4j-impl</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
janbartel, line 72: If you've moved this up to core then the dependency in the jetty-eeX-test-sessions-jdbc/pom.xml can be deleted.

janbartel (changes requested):

jetty-core/jetty-session/pom.xml
olamy: nope this will not work as `test` scope is not transitive

jetty-ee10/jetty-ee10-tests/jetty-ee10-test-sessions/jetty-ee10-test-sessions-infinispan/src/test/java/org/eclipse/jetty/ee10/session/hazelcast/remote/RemoteClusteredInvalidationSessionTest.java
```diff
@@ -11,10 +11,10 @@
 // ========================================================================
 //
 
-package org.eclipse.jetty.session.remote;
+package org.eclipse.jetty.ee10.session.hazelcast.remote;
 
 import org.eclipse.jetty.ee10.session.AbstractClusteredInvalidationSessionTest;
-import org.eclipse.jetty.session.LoggingUtil;
+import org.eclipse.jetty.ee10.session.infinispan.LoggingUtil;
```
sbordet: Seems an error that hazelcast depends on infinispan?

sbordet (changes requested):

jetty-ee10/jetty-ee10-tests/jetty-ee10-test-sessions/jetty-ee10-test-sessions-infinispan/src/test/java/org/eclipse/jetty/ee10/session/hazelcast/remote/RemoteClusteredInvalidationSessionTest.java
olamy: ditto good catch!

PR #9235, opened 2023-01-31
Issue #9233 - cleanup of websocket-core-common util package

### LINKED ISSUE #9233: Cleanup of websocket in Jetty-12

lachlan-roberts:
## Issue #9233

removing unused code and performing other cleanups

This is not intended to change any of the logic here, just performing code cleanups.

`ReflectUtils` is doing checks on `jakarta` and `javax` classes which is questionable to have in core, but this PR is not intended to address that.


joakime:
You have 470 test failures to address.

sbordet (approved):

PR #9236, opened 2023-01-31
Jetty 12 - Removing unused / old deps from ee10

joakime:
+ Removing `javax.activation` we are using `jakarta.activation` now
+ Removing osgi bootDelegationPackage for `javax.activation.*` as that package is no longer part of the JVM
+ Removing orbit (glassfish) mail we are using `com.sun.mail` (impl) now
+ Removing `javax.security.auth` we are using `jakarta.security.auth` now

### DIFF
--- a/jetty-ee10/pom.xml
+++ b/jetty-ee10/pom.xml
@@ -13,12 +13,6 @@
   <packaging>pom</packaging>
 
   <properties>
-    <com.sun.mail.version>2.0.1</com.sun.mail.version>
-    <!-- TODO: Remove these javax.* entries? -->
-    <javax.activation.impl.version>1.1.0.v201105071233</javax.activation.impl.version>
-    <javax.security.auth.message.version>1.0.0.v201108011116</javax.security.auth.message.version>
-    <javax.mail.glassfish.version>1.4.1.v201005082020</javax.mail.glassfish.version>
-
     <jakarta.activation.api.version>2.1.1</jakarta.activation.api.version>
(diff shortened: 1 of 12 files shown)


jetty-ee10/pom.xml
```diff
@@ -13,12 +13,6 @@
   <packaging>pom</packaging>
 
   <properties>
-    <com.sun.mail.version>2.0.1</com.sun.mail.version>
-    <!-- TODO: Remove these javax.* entries? -->
```
janbartel, line 17: The last time I tried removing some of these "old" dependencies it turned out that the testing framework - paxweb - needed them. Let's see how it builds on CI.

joakime: The OSGI layers and all of the `jetty-ee10/jetty-ee10-osgi/` tree is passing with green so far.

There's 1 failure remaining, a 502 response from the `DistributionTests#testWebAppWithProxyAndJPMS`.
It doesn't seem related to the change in this PR, but rather a CI flaky network issue ATM. 
I just added DEBUG logging to that test case in the last commit (b29d589ebd05123e808f41f3fa3af4d22e73aad9) to see what's going on exactly.

olamy (approved):
good cleanup! 
only need the debug to be removed before merging :)

joakime:
I'm leaving the added DEBUG in `DistributionTests.testWebAppWithProxyAndJPMS` in place.
The 502 hasn't happened again, so I don't know what's going on.

This DEBUG only adds about 7 extra lines of output, but will be immensely useful should that test fail again.

Issue #9275, opened 2023-02-01
Jetty-12 H3 tests passing after 30s timeout


gregw:
**Jetty version(s)**
12
**Java version/vendor** `(use: java -version)`
java 17

**OS type/version**

linux

**Description**

Many of the H3 tests in org.eclipse.jetty.test.client.transport.HttpClientStreamTest are taking over 30s to complete.

![Screenshot from 2023-02-01 14-54-21](https://user-images.githubusercontent.com/70829/215941878-90ed1074-1457-4ea2-bb79-3d1f49dbef6b.png)
![Screenshot from 2023-02-01 14-54-04](https://user-images.githubusercontent.com/70829/215941907-5aba8b7d-58e0-4148-9df3-e74e89246ebd.png)
![Screenshot from 2023-02-01 14-53-43](https://user-images.githubusercontent.com/70829/215941919-d3a3eea4-3a46-4ae2-9b40-cbb953c02e86.png)

**How to reproduce?**

Run test from IDE

gregw:
There are also several non suppressed stacks from that test as well

PR #9281, opened 2023-02-01
Get rid of unnecessary stack traces printed by H3

lorban:
As reported in #9275.

A connection should not throw when it is asked for its stream IDs after it was disposed of. This gets rid of printed stack traces at WARN level that unnecessarily pollute the output.

### DIFF
--- a/jetty-core/jetty-quic/jetty-quic-common/src/main/java/org/eclipse/jetty/quic/common/ProtocolSession.java
+++ b/jetty-core/jetty-quic/jetty-quic-common/src/main/java/org/eclipse/jetty/quic/common/ProtocolSession.java
@@ -215,25 +215,32 @@ public abstract class ProtocolSession extends ContainerLifeCycle
             if (task != null)
                 return task;
 
-            while (true)
+            try
             {
-                processWritableStreams();
-                boolean loop = processReadableStreams();
+                while (true)
+                {
+                    processWritableStreams();
+                    boolean loop = processReadableStreams();
 
-                task = poll();
-                if (LOG.isDebugEnabled())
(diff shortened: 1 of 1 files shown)


jetty-core/jetty-quic/jetty-quic-quiche/jetty-quic-quiche-foreign-incubator/src/main/java/org/eclipse/jetty/quic/quiche/foreign/incubator/ForeignIncubatorQuicheConnection.java
```diff
@@ -489,7 +489,7 @@ protected List<Long> iterableStreamIds(boolean write)
         try (AutoLock ignore = lock.lock())
         {
             if (quicheConn == null)
-                throw new IllegalStateException("connection was released");
+                return List.of();
```
sbordet: I'd prefer to catch from `ProtocolSession.StreamProducer.produce()`, log at debug, and return `null`.

sbordet (changes requested):

jetty-core/jetty-quic/jetty-quic-quiche/jetty-quic-quiche-foreign-incubator/src/main/java/org/eclipse/jetty/quic/quiche/foreign/incubator/ForeignIncubatorQuicheConnection.java
lorban: Ok, done.

sbordet (approved):

gregw (approved):

PR #9283, opened 2023-02-01
Configurable Unsafe Host Header Behaviors

joakime:
* Adding `HttpCompliance.DUPLICATE_HOST_HEADERS`
  + Optional compliance that allowance duplicate host headers.
* Adding `HttpCompliance.UNSAFE_HOST_HEADER`
  + Optional compliance that allows unsafe host headers.
* Adding warning logging for bad Host / authority situations


jetty-util/src/main/java/org/eclipse/jetty/util/HostPort.java
```diff
@@ -24,9 +26,97 @@
  */
 public class HostPort
 {
+    private static final Logger LOG = LoggerFactory.getLogger(HostPort.class);
     private final String _host;
     private final int _port;
 
+    public static HostPort unsafe(String authority)
```
gregw, line 46: needs javadoc to say what is unsafe about it.

jetty-util/src/main/java/org/eclipse/jetty/util/HostPort.java
```diff
@@ -24,9 +26,97 @@
...
+            }
+            else if (authority.charAt(0) == '[')
+            {
+                // ipv6reference
+                int close = authority.lastIndexOf(']');
+                String host = authority.substring(0, close + 1);
+                if (authority.length() > close + 1)
+                {
+                    // ipv6 with port
+                    String strPort = authority.substring(close + 2);
```
gregw: Shouldn't we still check for a ':' in this case?

jetty-util/src/test/java/org/eclipse/jetty/util/HostPortTest.java
```diff
@@ -77,7 +79,21 @@ public static Stream<Arguments> invalidAuthorityProvider()
             "host:-80", // host + invalid port
             "127.0.0.1:-80", // ipv4 + invalid port
             "[0::0::0::0::1]:-80", // ipv6 + invalid port
```
gregw: Can you also test `[0::0::0::0::1].80` and make sure it does not parse 80 as the port.

gregw (changes requested):
looking good, but a couple of niggles

jetty-util/src/main/java/org/eclipse/jetty/util/HostPort.java
joakime: I added a HostPortTest.testValidAuthorityViaUnsafe() as a way to ensure this logic doesn't change the behavior of valid authorities (if being used).

jetty-util/src/test/java/org/eclipse/jetty/util/HostPortTest.java
joakime: Added.

jetty-util/src/main/java/org/eclipse/jetty/util/HostPort.java
joakime: Added

jetty-util/src/test/java/org/eclipse/jetty/util/HostPortTest.java
gregw: Yeah but I think with the current code it will be parsed as host `[0::0::0::0::1]` and port 80.    I think in this case it should be host `[0::0::0::0::1].80` and port -1

jetty-util/src/test/java/org/eclipse/jetty/util/HostPortTest.java
```diff
@@ -112,4 +151,12 @@ public void testInvalidAuthority(String authority)
             new HostPort(authority);
         });
     }
+
+    @ParameterizedTest
+    @MethodSource("invalidAuthorityProvider")
+    public void testUnsafeAuthority(String authority)
+    {
+        HostPort hostPort = HostPort.unsafe(authority);
+        assertNotNull(hostPort);
```
gregw: We need to test more here.  We need to test if the port was correctly extracted or not.  So I think we need to pass in an expected port

gregw (changes requested):
sorry still a niggle

jetty-util/src/test/java/org/eclipse/jetty/util/HostPortTest.java
joakime: Many will be port 0, this is a test of `unsafe(String authority)` against totally broken authorities that the normal (safe) flows will reject and not allow.

We have a separate test for valid authorities that use `unsafe(String authority)` which does test for valid splits (host + port)

joakime: In fact, in `unsafe(String authority)` we don't even test to make sure that the port is valid (within range 1 thru 65535).

It could be parsed as -80, or 333444555 as its unsafe.

joakime: Basically everything about using `unsafe(String authority)` is unsafe, it can easily allow conditions later that will never work (you cannot create a URL or URI from them, or a valid `Location` header, etc)

gregw: I'm ok with it being totally unsafe, but we need to consistently and predictably in our unsafe handling. Currently if it looks like an IPv6 address we don't check for a : before posting the port. If it doesn't look like IPv6 then we do check for the colon. I think we should check in both cases. No colon then no port.

joakime: Reworked the test cases and the implementation in HostPort.

gregw (approved):

PR #9289, opened 2023-02-02
Fixes #9288 - Jetty 12 - Use oej.http.HttpCookie in jetty-client.

### LINKED ISSUE #9288: Jetty 12 - Use oej.http.HttpCookie in jetty-client
**Jetty version(s)**
12+

**Description**
Module `jetty-client` relies on `java.net.HttpCookie` and related classes, but they are a little outdated (for example, no support for `SameSite`) and with glitches (for example, cookie names are treated case-insensitively, but the RFC implies they are case-sensitive).

Furthermore, they are awkward to use, since they have to go through `java.net.CookieManager` and the default implementation of `java.net.CookieStore` is not particularly efficient.

At least the API usage in `jetty-client` should use `org.eclipse.jetty.http.HttpCookie`, and a Jetty cookie store implementation.

sbordet:
* Replaced usages of java.net.HttpCookie with oej.http.HttpCookie.
* Moved server-side only methods from HttpCookie to HttpCookieUtils.
* Introduced and implemented oej.http.HttpCookieStore.
* Removed now obsolete oej.util.HttpCookieStore.


jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookieStore.java
```diff
@@ -0,0 +1,388 @@
...
+     * @param uri the {@code URI} associated with the cookie
+     * @param cookie the cookie to add
+     * @return whether the cookie has been added
+     */
+    public boolean add(URI uri, HttpCookie cookie);
+
+    /**
+     * @return all the cookies
+     */
+    public List<HttpCookie> all();
```
gregw, line 56: I prefer `getCookies()`

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookieStore.java
```diff
@@ -0,0 +1,388 @@
...
+
+    /**
+     * <p>Returns the cookies that match the given {@code URI}.</p>
+     *
+     * @param uri the {@code URI} to match against
+     * @return a list of cookies that match the given {@code URI}
+     */
+    public List<HttpCookie> match(URI uri);
```
gregw, line 64: Would the following API be more efficient:
```suggestion
    public Iterator<HttpCookie> matches(URI uri);
```

As it would not need to build a list, rather just return a wrapper around the standard iterator.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookieStore.java
sbordet: But the other methods do not have "cookie" is their name, e.g. `add()` (not `addCookie()`), `match()`, `remove()`, so `all()` seemed better than `getCookies()`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookieStore.java
sbordet: I'm not sure I understand.

I prefer `match()` because it is a command (so not `matches()` that looks like a query that should return a boolean).

The algorithm has to do a bunch of passes through the internal data structure to collect the matching cookies, so it is natural to return a list rather than an iterator (because it is where cookies are collected during the various passes).

The only use of this case is for the client to send the cookies on requests, so the list will be fully iterated (so a lazy iterator that may not be fully consumed is not useful).

gregw:
@sbordet please don't force push.  There is no need for that as we squash when we merge.
If you force push you screw up all the tracking of commits.  I can't see what has changed since my last review (there is a reason the options are called -fu )

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookieStore.java
gregw: If multiple passes are done, then a simple wrapper of the internal iterator will not work and a collection is still needed.  So the current signature is good.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookieStore.java
```diff
@@ -0,0 +1,388 @@
...
+
+    /**
+     * <p>Removes the cookie associated with the given {@code URI}.</p>
+     *
+     * @param uri the {@code URI} associated with the cookie to remove
+     * @param cookie the cookie to remove
+     * @return whether the cookie has been removed
+     */
+    public boolean remove(URI uri, HttpCookie cookie);
```
gregw, line 73: Is this another setCookie?  What matching criteria are applied?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookieStore.java
```diff
@@ -0,0 +1,388 @@
...
+     * match that of the URI, etc.</p>
+     * <p>The cookie is associated with the given {@code URI}, so that
+     * a call to {@link #match(URI)} returns the cookie only if the
+     * URIs match.</p>
+     *
+     * @param uri the {@code URI} associated with the cookie
+     * @param cookie the cookie to add
+     * @return whether the cookie has been added
+     */
+    public boolean add(URI uri, HttpCookie cookie);
```
gregw, line 51: I think `setCookie` is a clearer name for the arg
```suggestion
    public boolean add(URI uri, HttpCookie setCookie);
```

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookieStore.java
```diff
@@ -0,0 +1,388 @@
...
+        }
+    }
+
+    /**
+     * <p>A default implementation of {@link HttpCookieStore}.</p>
+     */
+    public static class Default implements HttpCookieStore
+    {
+        private final AutoLock lock = new AutoLock();
+        private final Map<Key, List<HttpCookie>> cookies = new HashMap<>();
```
gregw, line 124: Would this be better as a 
```suggestion
        private final Map<Key, PathMapper<HttpCookie>> cookies = new HashMap<>();
```

I guess that is an optimisation that can come later.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookieStore.java
```diff
@@ -0,0 +1,388 @@
...
+            }
+
+            @Override
+            public boolean isExpired()
+            {
+                long maxAge = getMaxAge();
+                if (maxAge >= 0 && NanoTime.secondsSince(creationNanoTime) > maxAge)
+                    return true;
+                Instant expires = getExpires();
+                return expires != null && Instant.now().isAfter(expires);
```
gregw, line 384: Isn't this just 
```suggestion
                return getWrapped().isExpired();
```

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookieStore.java
```diff
@@ -0,0 +1,388 @@
...
+        private record Key(String scheme, String domain)
+        {
+            private Key(String scheme, String domain)
+            {
+                this.scheme = scheme;
+                this.domain = domain.toLowerCase(Locale.ENGLISH);
+            }
+        }
+
+        private static class Cookie extends HttpCookie.Wrapper
```
gregw, line 368: perhaps:
```suggestion
        private static class SetCookie extends HttpCookie.Wrapper
```

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/transport/HttpDestination.java
```diff
@@ -84,7 +85,9 @@ public HttpDestination(HttpClient client, Origin origin, boolean intrinsicallySe
         this.requestTimeouts = new RequestTimeouts(client.getScheduler());
 
         String host = HostPort.normalizeHost(getHost());
-        if (!client.isDefaultPort(getScheme(), getPort()))
+        String scheme = getScheme();
+        int port = getPort();
```
gregw, line 88: you don't use the port variable below and instead call getPort again

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/transport/HttpDestination.java
```diff
@@ -199,7 +202,8 @@ private ClientConnectionFactory newSslClientConnectionFactory(SslContextFactory.
     @Override
     public boolean isSecure()
     {
-        return HttpClient.isSchemeSecure(getScheme());
+        String scheme = getScheme();
```
gregw: why the variable?

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/transport/HttpDestination.java
```diff
@@ -84,7 +85,9 @@ public HttpDestination(HttpClient client, Origin origin, boolean intrinsicallySe
         this.requestTimeouts = new RequestTimeouts(client.getScheduler());
 
         String host = HostPort.normalizeHost(getHost());
-        if (!client.isDefaultPort(getScheme(), getPort()))
+        String scheme = getScheme();
```
gregw: why the variable?

gregw (approved):
A few niggles, but otherwise I can't see any problems.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookieStore.java
joakime: Our `HttpCookieStore` appears to be modeled after `java.net.CookieStore` in methods and params.
Since this can be used by a user, to preload a CookieStore, and is not solely from a `Set-Cookie` response header, I like `HttpCookie cookie` as the param name.

* https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/net/CookieStore.html

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookieStore.java
sbordet: The cookie may not come from a `Set-Cookie` header, but just pre-loaded by the client application.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookieStore.java
sbordet: There is no `PathMapper` class? Besides, the match is done on domains and paths, so not sure we have an existing solution for that?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookieStore.java
sbordet: No, because it will re-do the check on `maxAge` that we have already done against `creationNanoTime`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookieStore.java
sbordet: Not `SetCookie`, perhaps `StoredCookie`, but it's a private internal class, so I thought shorter was better.

Issue #9072, opened 2022-12-19
The great Jetty-12 renaming omnibus


gregw:
**Jetty version(s)**
12

**Enhancement Description**

As jetty-12 alpha dev comes to a close, there are few classes and methods that could be renamed.   To make our lives simpler, it will be best to do these all at once with minimal other PRs outstanding.   So this issue is here to suggest/discuss renamings that we want  to do, and when we are all agreed we can do them all at once and quickly merge.  Please list suggested renamings in the comments.

gregw:
`Handler.Collection` to `Handler.List` or something else?   The problem with `List` is that it will clash with  `java.util.List` in any `Handler`, be they a `List` or not.   Thus `Handler.Sequence` is probably better.

gregw:
`Handler.process` to `Handler.handle`.  This would mean that `Handler` would not longer implement `Request.Processor`, but I do not think there are any examples of a Handler being used as just a `Request.Processor`.

joakime:
other issues with this 

* #8975 
* #8991
* #4731

gregw:
`SessionConfig` has a TODO for renaming. @janbartel do you know what they are?

sbordet:
> `Handler.process` to `Handler.handle`. This would mean that `Handler` would not longer implement `Request.Processor`, but I do not think there are any examples of a Handler being used as just a `Request.Processor`.

Sure there are! `ErrorProcessor`s! 
I remember there was a problem considering them `Handler`s so I won't mind a common super-interface (maybe separated from `Request`).

gregw:
@sbordet  So an `ErrorProcessor` is currently just a `Processor`.  It is not a `Handler`, so would not be affected by the renaming.
Now we might want to go back to an `ErrorHandler` if we want a life cycle and a server reference in the error thingy, but that would be a different PR as it is more than just a renaming.

I do not think we have any code that is effectively:
```java
    Request.Processor p = myHandler;
```

and even if we did, that could just become:
```java
    Request.Processor p = myHandler::handle;
```

lachlan-roberts:
Adding the websocket isDemanding rename to the omnibus https://github.com/eclipse/jetty.project/issues/8991

gregw:
The jetty-core/jetty-ee module is not really ee specific.  Currently it just contains the Deployable interface, which can apply to non ee contexts.

It could be renamed to `jetty-context-utils` or `jetty-common` and used to hold code that is common between eeN and other types of context.

[...]

Issue #8759, opened 2022-10-24
Jetty 12 - Review Handler.Collection.setHandlers use of "invocationType"


joakime:
**Jetty version(s)**
Jetty 12

**Description**
Review method variable `invocationType`, as it is not apparently used for anything in the method.

PR #9305, opened 2023-02-03
Rename Handler Nested & Collection

### LINKED ISSUE #8765: Review Handler.Container so that it only represents an immutable interface
**Jetty version(s)**
Jetty 12

**Description**
The Handler.Container interface should be an immutable interface (without `addHandler()` methods).

gregw:
Fixes #8765  #8759 . Part of #9072

There is now a Handler interface hierarchy:
 + Container is a Handler that has 1 or more contained Handlers.
 + Wrapper is a Container with only 1 handler and a setHandler method.
 + Collection is a Container with n handlers and a addHandler method

class are now:
 + Abstract implements Handler
 + AbstractContainer extends Abstract implements Container
 + BaseWrapper extends AbstractContainer implements Wrapper
 + Sequence extends AbstractContainer implements Collection

 Lots of other associated cleanups

### DIFF
--- a/jetty-ee9/jetty-ee9-maven-plugin/src/main/java/org/eclipse/jetty/ee9/maven/plugin/ServerSupport.java
+++ b/jetty-ee9/jetty-ee9-maven-plugin/src/main/java/org/eclipse/jetty/ee9/maven/plugin/ServerSupport.java
@@ -23,12 +23,10 @@ import org.eclipse.jetty.ee9.security.LoginService;
 import org.eclipse.jetty.ee9.webapp.Configurations;
 import org.eclipse.jetty.ee9.webapp.WebAppContext;
 import org.eclipse.jetty.server.Connector;
-import org.eclipse.jetty.server.Handler;
 import org.eclipse.jetty.server.RequestLog;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.handler.ContextHandler;
 import org.eclipse.jetty.server.handler.ContextHandlerCollection;
-import org.eclipse.jetty.server.handler.DefaultHandler;
 import org.eclipse.jetty.util.resource.ResourceFactory;
 import org.eclipse.jetty.xml.XmlConfiguration;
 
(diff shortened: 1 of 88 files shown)


gregw:
@sbordet @joakime @janbartel Can I get a review on this ASAP as I really don't want this to go stale and have to do it again.

joakime (approved):
I'm good with this.

sbordet:
@gregw sorry I don't like some parts of this PR.

In every other Jetty class, `Wrapper` is a class and it is used to "wrap" an instance of an interface, so we should stick with this semantic.

`BaseWrapper` is not a good name.

I would rather have `Wrapper` back as a class, and if we really need it use `Single` as the interface opposed to `Collection`.

"Single" is also used in Java's `Collections.singletonList()` to return a list of 1 element ("Singleton" cannot be used as it remembers the well-known pattern which has a different meaning than what we want here).

gregw:
@sbordet how about Delegate rather than Single?

gregw:
Actually, I don't mind `Singleton` as it is now just as much associated with `Collections.singletonXxx()` than the pattern.   Thus a `Wrapper` is a concrete implementation of a `Singleton` `Container`.

jetty-ee9/jetty-ee9-maven-plugin/src/main/java/org/eclipse/jetty/ee9/maven/plugin/ServerSupport.java
```diff
@@ -68,11 +66,7 @@ public static void configureHandlers(Server server, List<ContextHandler> context
         if (contexts == null)
         {
             contexts = new ContextHandlerCollection();
-            Handler.Collection handlers = server.getDescendant(Handler.Collection.class);
-            if (handlers == null)
-                server.setHandler(new Handler.Collection(contexts, new DefaultHandler()));
-            else
-                handlers.addHandler(contexts);
```
basil, line 75: Causes #12207, also decreasing the accuracy of the Javadoc a few lines above, which (now incorrectly) states:

> Also put in a `DefaultHandler` so we get a nicer page than a 404 if we hit the root and the webapp's context isn't at root.

PR #9308, opened 2023-02-03
Issue #7650 - Fix race condition when stopping QueuedThreadPool

### LINKED ISSUE #7650: QueuedThreadPool: Stopped without executing or closing null
**Jetty version(s)**
9.4.44.v20210927

Java 17
we use `gcr.io/distroless/java17-debian11:nonroot`

**OS type/version**
Debian 11

**Description**
We use Spring Boot 2.6.2 with Jetty. 

In our logs the following log appears:

`WARN [SpringApplicationShutdownHook] QueuedThreadPool: Stopped without executing or closing null`

This seems to be the code that writes that warning log:
https://github.com/eclipse/jetty.project/blob/jetty-9.4.44.v20210927/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java#L299

My understand is that either the `BlockingQueue<Runnable> _jobs` contains objects that are null, or there is a race condition so that `!_jobs.isEmpty()` is `true` but then `_jobs.poll()` returns `null` because it is now empty.

I'm not sure if this issue could be related to Spring so I'd like to mention that we implemented a shutdown hook waiting for an ExecutorService:

```java
@Configuration
public class MyConfig {
    ....
    @Bean(destroyMethod = "onShutdown")
    MyWorker createMyWorker() {
        ExecutorService executorService = createExecutorService(); // returns a new ThreadPoolExecutor
        return new MyBean(executorService);
    }
}
```

```java
public class MyWorker {
    ...
    public boolean onShutdown() throws InterruptedException {
        this.executorService.shutdown();
        boolean terminatedGracefully = executorService.awaitTermination(180, TimeUnit.SECONDS);
        return terminatedGracefully;
    }
}
```

**How to reproduce?**
Unfortunately, I cannot reproduce it.

dzoech:
Fixes #7650 that causes the warning message `Stopped without executing or closing null`

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
@@ -280,9 +280,13 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
         }
 
         // Close any un-executed jobs
-        while (!_jobs.isEmpty())
+        while (true)
         {
             Runnable job = _jobs.poll();
+            if (job == null)
+            {
+                break;
+            }
             if (job instanceof Closeable)
             {
                 try


sbordet:
@dzoech can you please redo this PR based on `jetty-10.0.x` branch?
We will do the merge forward, but unlikely to fix this for Jetty 9.4.x due to #7958.

Issue #9311, opened 2023-02-03
Performance of `ArrayRetainableByteBufferPool.acquire()` can degenerate pathologically as the buckets grow in size


lorban:
**Jetty version(s)**
12

**Description**
Since `ArrayRetainableByteBufferPool.acquire()` is backed by `ConcurrentPool.acquire()`, each execution requires iterating over a number of entries proportional to the usage level of the bucket, i.e.: a bucket that is 90% used requires iterating over 90% of the entries on average.

When the number of buffers in a single bucket is quite small (in the 100s), those iterations are so cheap they have a negligible cost as the array is small enough to fit into the CPU cache. But when that number of buffers reaches the 10s of thousands, these iterations become a pathological bottleneck.

lorban:
This was found in [MultiPartServletTest.testLargePart()](https://github.com/eclipse/jetty.project/blob/jetty-12.0.x-multipartCleanups/jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/MultiPartServletTest.java#L134-L174) when the default buffer pool is used as this test needs 262144 buffers (2GB split in 8KB) to complete.

With the `ArrayRetainableByteBufferPool.acquire()` implementation being:
1) ConcurrentPool.acquire()
2) if null, reserve and allocate

this means it would require 262144! (yes; factorial of 262144, which approaches infinity) `compareAndSet` calls for the test to eventually finish.

@lachlan-roberts could you please add a TODO referring to this issue to your test that uses a non-pooling pool as a workaround?

lorban:
The root of this issue seems to be that the algorithm backing `ConcurrentPool` does not scale well to large sizes.

There are two major users of the `ConcurrentPool`:

1) connection pooling
This seems to be less of a problem for this use-case as a connection pool sized at 1000 entries is quite substantial and very uncommon, and acquiring from it may not happen millions of times per second while it mostly contains busy connections sounds like a very niche use-case.
It is nevertheless a possibility, even if a remote one.

2) buffer pooling
Having 10's of thousands of buffers of a specific size (especially the small ones < 16KB) doesn't sound unrealistic nor very uncommon so solving this issue seems more pressing.
Since `ConcurrentPool` is fully abstracted by `ArrayRetainableByteBuffer` we could add some mitigation strategies to handle this problem, like for instance:
   - hard capping the # of buffer per bucket to a few hundreds
[...]

PR #9319, opened 2023-02-06
ByteBufferPool with Queue

gregw:
For #9311 
This adds an optional Queue to the ByteBufferPool. More work is needed to ensure the Pool size is not too large and to better configure the queue size.


gregw:
Closed in favour of approach in #9320

PR #9320, opened 2023-02-07
Pool Wrapper with Queue

### LINKED ISSUE #9311: Performance of `ArrayRetainableByteBufferPool.acquire()` can degenerate pathologically as the buckets grow in size

gregw:
For #9311
This adds a QueuedPool that can wrap any other pool and provided a queue of Entries as a kind of cache.
This is an alternative to #9319

### DIFF
--- a/jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
+++ b/jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
@@ -21,6 +21,8 @@ import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Consumer;
 import java.util.function.IntUnaryOperator;
 
+import org.eclipse.jetty.io.internal.CompoundPool;
+import org.eclipse.jetty.io.internal.QueuedPool;
 import org.eclipse.jetty.util.BufferUtil;
 import org.eclipse.jetty.util.ConcurrentPool;
 import org.eclipse.jetty.util.NanoTime;
@@ -444,9 +446,15 @@ public class ArrayByteBufferPool implements ByteBufferPool, Dumpable
         private final Pool<RetainableByteBuffer> _pool;
         private final int _capacity;
 
-        private RetainedBucket(int capacity, int size)
+        private RetainedBucket(int capacity, int poolSize)
         {
-            _pool = new ConcurrentPool<>(ConcurrentPool.StrategyType.THREAD_ID, size, true);
+            if (poolSize <= ConcurrentPool.OPTIMAL_MAX_SIZE)
+                _pool = new ConcurrentPool<>(ConcurrentPool.StrategyType.THREAD_ID, poolSize, true);
+            else
+                _pool = new CompoundPool<>(
+                    new ConcurrentPool<>(ConcurrentPool.StrategyType.THREAD_ID, ConcurrentPool.OPTIMAL_MAX_SIZE, true),
+                    new QueuedPool<>(poolSize - ConcurrentPool.OPTIMAL_MAX_SIZE)
+                );
             _capacity = capacity;
         }
 
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/ConcurrentPool.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/ConcurrentPool.java
@@ -20,7 +20,6 @@ import java.util.Objects;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.Function;
 import java.util.function.ToIntFunction;
 import java.util.stream.Stream;
 
@@ -46,10 +45,18 @@ import org.slf4j.LoggerFactory;
 @ManagedObject
 public class ConcurrentPool<P> implements Pool<P>, Dumpable
 {
+    /**
+     * {@link ConcurrentPool} internally needs to linearly scan a list to perform an acquisition.
+     * This list needs to be reasonably short otherwise there is a risk that scanning the list
+     * becomes a bottleneck. Instances created with a size at most this value should be immune
+     * to this problem.
+     */
+    public static final int OPTIMAL_MAX_SIZE = 256;
+
     private static final Logger LOG = LoggerFactory.getLogger(ConcurrentPool.class);
 
     private final List<Entry<P>> entries = new CopyOnWriteArrayList<>();
-    private final int maxEntries;
+    private final int maxSize;
     private final StrategyType strategyType;
     /*
      * The cache is used to avoid hammering on the first index of the entry list.
@@ -59,8 +66,8 @@ public class ConcurrentPool<P> implements Pool<P>, Dumpable
      * When an entry can't be found in the cache, the global list is iterated
      * with the configured strategy so the cache has no visible effect besides performance.
      */
-    private final AutoLock lock = new AutoLock();
     private final ThreadLocal<Entry<P>> cache;
+    private final AutoLock lock = new AutoLock();
     private final AtomicInteger nextIndex;
     private final ToIntFunction<P> maxMultiplex;
     private volatile boolean terminated;
@@ -69,23 +76,23 @@ public class ConcurrentPool<P> implements Pool<P>, Dumpable
      * <p>Creates an instance with the specified strategy and no {@link ThreadLocal} cache.</p>
      *
      * @param strategyType the strategy to used to lookup entries
-     * @param maxEntries the maximum number of pooled entries
+     * @param maxSize the maximum number of pooled entries
      */
-    public ConcurrentPool(StrategyType strategyType, int maxEntries)
+    public ConcurrentPool(StrategyType strategyType, int maxSize)
     {
-        this(strategyType, maxEntries, false);
+        this(strategyType, maxSize, false);
     }
 
     /**
      * <p>Creates an instance with the specified strategy and an optional {@link ThreadLocal} cache.</p>
      *
      * @param strategyType the strategy to used to lookup entries
-     * @param maxEntries the maximum number of pooled entries
+     * @param maxSize the maximum number of pooled entries
      * @param cache whether a {@link ThreadLocal} cache should be used for the most recently released entry
      */
-    public ConcurrentPool(StrategyType strategyType, int maxEntries, boolean cache)
+    public ConcurrentPool(StrategyType strategyType, int maxSize, boolean cache)
     {
-        this(strategyType, maxEntries, cache, pooled -> 1);
+        this(strategyType, maxSize, cache, pooled -> 1);
     }
 
     /**
@@ -93,13 +100,15 @@ public class ConcurrentPool<P> implements Pool<P>, Dumpable
      * and a function that returns the max multiplex count for a given pooled object.</p>
      *
      * @param strategyType the strategy to used to lookup entries
-     * @param maxEntries the maximum number of pooled entries
+     * @param maxSize the maximum number of pooled entries
      * @param cache whether a {@link ThreadLocal} cache should be used for the most recently released entry
      * @param maxMultiplex a function that given the pooled object returns the max multiplex count
      */
-    public ConcurrentPool(StrategyType strategyType, int maxEntries, boolean cache, ToIntFunction<P> maxMultiplex)
+    public ConcurrentPool(StrategyType strategyType, int maxSize, boolean cache, ToIntFunction<P> maxMultiplex)
     {
-        this.maxEntries = maxEntries;
+        if (maxSize > OPTIMAL_MAX_SIZE && LOG.isDebugEnabled())
+            LOG.debug("{} configured with max size {} which is above the recommended value {}", getClass().getSimpleName(), maxSize, OPTIMAL_MAX_SIZE);
+        this.maxSize = maxSize;
         this.strategyType = Objects.requireNonNull(strategyType);
         this.cache = cache ? new ThreadLocal<>() : null;
         this.nextIndex = strategyType == StrategyType.ROUND_ROBIN ? new AtomicInteger() : null;
@@ -138,10 +147,10 @@ public class ConcurrentPool<P> implements Pool<P>, Dumpable
 
             // If we have no space
             int entriesSize = entries.size();
-            if (maxEntries > 0 && entriesSize >= maxEntries)
+            if (maxSize > 0 && entriesSize >= maxSize)
             {
                 if (LOG.isDebugEnabled())
-                    LOG.debug("no space: {} >= {}, cannot reserve entry for {}", entriesSize, maxEntries, this);
+                    LOG.debug("no space: {} >= {}, cannot reserve entry for {}", entriesSize, maxSize, this);
                 return null;
             }
 
@@ -214,37 +223,6 @@ public class ConcurrentPool<P> implements Pool<P>, Dumpable
             case ROUND_ROBIN -> nextIndex.getAndUpdate(c -> Math.max(0, c + 1)) % size;
             case THREAD_ID -> (int)(Thread.currentThread().getId() % size);
         };
-    }
-
-    @Override
-    public Entry<P> acquire(Function<Entry<P>, P> creator)
-    {
-        Entry<P> entry = acquire();
-        if (entry != null)
-            return entry;
-
-        entry = reserve();
-        if (entry == null)
-            return null;
-
-        P value;
-        try
-        {
-            value = creator.apply(entry);
-        }
-        catch (Throwable th)
-        {
-            entry.remove();
-            throw th;
-        }
-
-        if (value == null)
-        {
-            entry.remove();
-            return null;
-        }
-
-        return entry.enable(value, true) ? entry : null;
     }
 
     private boolean release(Entry<P> entry)
@@ -325,7 +303,7 @@ public class ConcurrentPool<P> implements Pool<P>, Dumpable
     @Override
     public int getMaxSize()
     {
-        return maxEntries;
+        return maxSize;
     }
 
     @Override
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
@@ -40,7 +40,7 @@ public interface Pool<P>
      * @return a disabled entry that is contained in the pool, or {@code null}
      * if the pool is terminated or if the pool cannot reserve an entry
      */
-    public Entry<P> reserve();
+    Entry<P> reserve();
 
     /**
      * <p>Acquires an entry from the pool.</p>
@@ -50,7 +50,7 @@ public interface Pool<P>
      *
      * @return an entry from the pool or null if none is available.
      */
-    public Entry<P> acquire();
+    Entry<P> acquire();
 
     /**
      * <p>Acquires an entry from the pool,
@@ -59,13 +59,41 @@ public interface Pool<P>
      * @param creator a function to create the pooled value for a reserved entry.
      * @return an entry from the pool or null if none is available.
      */
-    public Entry<P> acquire(Function<Entry<P>, P> creator);
+    default Entry<P> acquire(Function<Entry<P>, P> creator)
+    {
+        Entry<P> entry = acquire();
+        if (entry != null)
+            return entry;
+
+        entry = reserve();
+        if (entry == null)
+            return null;
+
+        P value;
+        try
+        {
+            value = creator.apply(entry);
+        }
+        catch (Throwable th)
+        {
+            entry.remove();
+            throw th;
+        }
+
+        if (value == null)
+        {
+            entry.remove();
+            return null;
+        }
+
+        return entry.enable(value, true) ? entry : null;
+    }
 
     /**
      * @return whether this {@code Pool} has been terminated
      * @see #terminate()
      */
-    public boolean isTerminated();
+    boolean isTerminated();
 
     /**
      * <p>Terminates this {@code Pool}.</p>
@@ -77,30 +105,30 @@ public interface Pool<P>
      *
      * @return a list of all entries
      */
-    public Collection<Entry<P>> terminate();
+    Collection<Entry<P>> terminate();
 
     /**
      * @return the current number of entries in this {@code Pool}
      */
     @ManagedAttribute("The number of entries")
-    public int size();
+    int size();
 
     /**
      * @return the maximum number of entries in this {@code Pool}
      */
     @ManagedAttribute("The maximum number of entries")
-    public int getMaxSize();
+    int getMaxSize();
 
     /**
      * @return a {@link Stream} over the entries
      */
-    public Stream<Entry<P>> stream();
+    Stream<Entry<P>> stream();
 
     /**
      * @return the number of reserved entries
      */
     @ManagedAttribute("The number of reserved entries")
-    public default int getReservedCount()
+    default int getReservedCount()
     {
         return (int)stream().filter(Entry::isReserved).count();
     }
@@ -109,7 +137,7 @@ public interface Pool<P>
      * @return the number of idle entries
      */
     @ManagedAttribute("The number of idle entries")
-    public default int getIdleCount()
+    default int getIdleCount()
     {
         return (int)stream().filter(Entry::isIdle).count();
     }
@@ -118,7 +146,7 @@ public interface Pool<P>
      * @return the number of in-use entries
      */
     @ManagedAttribute("The number of in-use entries")
-    public default int getInUseCount()
+    default int getInUseCount()
     {
         return (int)stream().filter(Entry::isInUse).count();
     }
@@ -127,7 +155,7 @@ public interface Pool<P>
      * @return the number of terminated entries
      */
     @ManagedAttribute("The number of terminated entries")
-    public default int getTerminatedCount()
+    default int getTerminatedCount()
     {
         return (int)stream().filter(Entry::isTerminated).count();
     }
@@ -137,7 +165,7 @@ public interface Pool<P>
      *
      * @param <W> the type of the pooled objects
      */
-    public static class Wrapper<W> implements Pool<W>
+    class Wrapper<W> implements Pool<W>
     {
         private final Pool<W> wrapped;
 
@@ -229,7 +257,7 @@ public interface Pool<P>
      *
      * @param <E> the type of the pooled objects
      */
-    public interface Entry<E>
+    interface Entry<E>
     {
         /**
          * <p>Enables this, previously {@link #reserve() reserved}, {@code Entry}.</p>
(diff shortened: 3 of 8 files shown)


gregw:
@lorban do you want me to polish this whilst you work on benchmarks, or do you wish to take over the branch?
@lorban @sbordet thoughts on this approach vs #9319 ? Should we just close the other one and only resurrect it if this approach doesn't work for some reason?

lorban:
@gregw Benchmarks should be working again by tomorrow, in which case taking over this branch is my next step. If I'm late restoring benchmarks, I'll let you know.

I find this approach superior to https://github.com/eclipse/jetty.project/pull/9319 so IMHO the latter can be closed unless we find a showstopper on the way.

lorban:
I've introduced two `Pool` implementations:
 - `QueuedPool` which is based on a `ConcurrentLinkedQueue`, takes some liberties with the `Pool` contract and cannot multiplex.
 - CompoundPool which wraps `ConcurrentPool` and `QueuedPool`, using a heuristic to figure out the `ConcurrentPool` size. It also cannot multiplex.

`ArrayByteBufferPool` simply uses `CompoundPool` instead of `ConcurrentPool`, and `MultiPartServletTest.testLargePart()` now "only" is ~30% slower with `ArrayByteBufferPool` vs `ByteBufferPool.NonPooling` which IMHO is reasonable.

I still believe tackling this problem at the `Pool` level is the right way, but I'm much less convinced about the details (3 `Pool` impl with one wrapping the other 2, naming, the fact that `QueuedPool` doesn't strictly follow the `Pool` contract...).

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/CompoundPool.java
```diff
@@ -0,0 +1,147 @@
...
+ * This implementation does not support multiplexing.
+ *
+ * @param <P> the type of the pooled objects
+ */
+public class CompoundPool<P> implements Pool<P>
+{
+    private final ConcurrentPool<P> concurrentPool;
+    private final Pool<P> secondBestPool;
+
+    public CompoundPool(ConcurrentPool.StrategyType strategyType, int maxSize, boolean cache)
```
gregw: I'd rather a constructor that just takes 2 other pools as that will allow the full configuration options to be available.  Perhaps have some static builder methods that can make some common combinations, but we shouldn't bake only one combo into the class.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QueuedPool.java
```diff
@@ -0,0 +1,245 @@
...
+        this.maxSize = maxSize;
+    }
+
+    @Override
+    public Entry<P> reserve()
+    {
+        rdLock.lock();
+        try
+        {
+            if (terminated || queue.size() == maxSize)
```
gregw: There should be the option to have an unbounded queue if maxSize is < 0.  Hmm I guess we could just pass in MAX_VALUE

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QueuedPool.java
```diff
@@ -0,0 +1,245 @@
...
+
+/**
+ * This implementation does not support multiplexing.
+ *
+ * @param <P> the type of the pooled objects
+ */
+public class QueuedPool<P> implements Pool<P>
+{
+    private final int maxSize;
+    private final Queue<Pool.Entry<P>> queue = new ConcurrentLinkedQueue<>();
```
gregw: Is this always the best queue?   For known bounds isn't some kind of Array based one likely to be better?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/ConcurrentPool.java
```diff
@@ -46,6 +45,7 @@
 @ManagedObject
 public class ConcurrentPool<P> implements Pool<P>, Dumpable
 {
+    static final int MAX_RECOMMENDED_SIZE = 256;
```
gregw: We need a comment saying why this exists and why this value.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/CompoundPool.java
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+
+    @Override
+    public Entry<P> reserve()
+    {
+        Entry<P> reserved = concurrentPool.reserve();
+        if (reserved != null)
+            return reserved;
+        return secondBestPool.reserve();
```
gregw: Stylewise, and naming-wise, I have a liking for:
```suggestion
        Entry<P> primary = primaryPool.reserve();
        return primary != null ? primary : secondaryPool.reserver();
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/CompoundPool.java
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * This implementation does not support multiplexing.
+ *
+ * @param <P> the type of the pooled objects
+ */
+public class CompoundPool<P> implements Pool<P>
+{
+    private final ConcurrentPool<P> concurrentPool;
+    private final Pool<P> secondBestPool;
```
gregw: I prefer the names:
```suggestion
    private final Pool<P> primaryPool;
    private final Pool<P> secondaryPool;
```
and the primary pool should not be tied to being a ConcurrentPool

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/CompoundPool.java
```diff
@@ -0,0 +1,147 @@
...
+        if (maxSize <= ConcurrentPool.MAX_RECOMMENDED_SIZE)
+        {
+            this.concurrentPool = new ConcurrentPool<>(strategyType, maxSize, cache);
+            this.secondBestPool = new NullPool<>();
+        }
+        else
+        {
+            this.concurrentPool = new ConcurrentPool<>(strategyType, ConcurrentPool.MAX_RECOMMENDED_SIZE, cache);
+            this.secondBestPool = new QueuedPool<>(maxSize - ConcurrentPool.MAX_RECOMMENDED_SIZE);
+        }
```
gregw: We need someway of checking that both pools have compatible multiplexing capabilities.  Do we need to be able to ask a Pool what it's max multiple is?   Or perhaps just ask it if it has a non unary max multiple function in a `isMultiplexing` method.  If one pool is multiplexing, then the other one should be as well.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
```diff
@@ -444,9 +445,9 @@ private Pool.Entry<RetainableByteBuffer> findOldestEntry(long now, Pool<Retainab
         private final Pool<RetainableByteBuffer> _pool;
         private final int _capacity;
 
-        private RetainedBucket(int capacity, int size)
+        private RetainedBucket(int capacity, int poolSize)
         {
-            _pool = new ConcurrentPool<>(ConcurrentPool.StrategyType.THREAD_ID, size, true);
+            _pool = new CompoundPool<>(ConcurrentPool.StrategyType.THREAD_ID, poolSize, true);
```
gregw: Does this need to be pluggable or at least extensible?

gregw (commented):
I think the basic approach shows promise and I don't think we should hold up a merge too much considering how pathological the current pool can be.   

It would be nice to see how we could configure some alternative compound pools in XML.

gregw:
Would it be possible to have a multiplexing `TreeSetPool`?    Pool Entries would be stored in a TreeSet that sorted on the remaining multiplex capability of each entry.  Thus an acquire would take the most allocated entry (until it had no capacity left and was taken out of the set).  Perhaps a TreeSet is not the right data structure as it is hard to remove and then re-add, but some sorted queue could work.
We would need this only if, for example, in the client we have need of more that 256 multiplexed connections to the same destination.   That is plausible!

gregw:
Perhaps a `PriorityQueue` could be used. Entries would be taken from the queue on every acquire, have their multiplex count incremented and then returned to the queue to be re-sorted.  On release, we need to be able also remove and re-add the entry, so the remove(Object) operation on the queue needs to be good... which it might be as there is a heap underneath it?     Besides, for the multiple use case of connection pools, it does not need to by hyper efficient.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QueuedPool.java
lorban: Besides `ConcurrentLinkedQueue`,  `ArrayBlockingQueue` and `LinkedBlockingQueue` are our only thread-safe options, and both use a relatively simple locked-based algorithm for that purpose. 

Concurrency would suffer if we chose one of those.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/ConcurrentPool.java
lorban: I feared that question, as the answer is based on a back-of-the-envelope calculation backed by a quick & dirty test.

I'll document that, since this constant isn't public in any way, it might be good enough.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/CompoundPool.java
lorban: That's a can of worms. I considered adding a `isMultiplexing` method to `Pool` but then I realized that multiplexing is based on a constructor-provided `ToIntFunction` which makes it impossible to be 100% certain both pools are going to use the same, or compatible multiplexing functions.

The best I can think of is to have such `isMultiplexing` method in `Pool` and make `ConcurrentPool` return `false` only when the constructor that doesn't take a `ToIntFunction` has been used.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
lorban: I would say no. We've already noticed that `ArrayByteBufferPool` exposes too many details of the old implementation as constructor arguments and I don't want to corner us in case we want to change the implementation yet again in the future.

There's always the option of using a different `ByteBufferPool` implementation in the worst case.

lorban:
@gregw finding a sorted _and_ concurrent structure with adequate perf to hold the multiplexed entries sounds challenging. I do not think we currently have a good use-case for it, as only H2 connections in the client currently are pooled and using the multiplexing support. But maybe we just found a good reason to not warn when a too large `ConcurrentPool` is configured?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QueuedPool.java
lorban: Yes, and that's what we've been doing with `ConcurrentPool` so far.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/CompoundPool.java
lorban: I considered such constructor, but how to make sure both underlying pools are compatible, especially w.r.t multiplexing?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/CompoundPool.java
lorban: Done.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/CompoundPool.java
lorban: Done.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/CompoundPool.java
gregw: I still think this is the right thing to do, even if we cannot have `isMultiplexing`.     
IT would be better even if the constructor was private an only a static factory method with this signature was available

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/CompoundPool.java
gregw: that works for me.

gregw:
It should be fairly simple to find a `O(log(n))` sorted queue that will outperform an `O(N)` search for a connection with remaining capacity.   For these kind of use-cases, we are not looking for stunning performance, we just want to avoid the lack of graceful degradation we currently see.  Not sucking will be good enough for now.

gregw:
@sbordet, as this was originally my PR, I can't review.    I'm OK with it in it's current form (other than my niggle about a better constructor).

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/ConcurrentPool.java
```diff
@@ -46,6 +45,14 @@
...
 public class ConcurrentPool<P> implements Pool<P>, Dumpable
 {
+    /**
+     * Each {@link ConcurrentEntry) takes about 56 bytes when factoring in the AtomicBiInteger's size,
+     * assuming uncompressed pointers and 128 bit object headers. 256 times this size makes 14KB,
+     * which is roughly half the size of the L1 cache of modern CPUs, which leaves some headroom
+     * free for other data. A quick test has also shown that ConcurrentPool's perf starts dipping
+     * noticeably when more than 512 entries are used.
+     */
+    static final int MAX_RECOMMENDED_SIZE = 256;
```
sbordet: Not a biggie, but I'd rename this to `OPTIMAL_MAX_SIZE`, or `RECOMMENDED_MAX_SIZE` as "max" refers to the size so they should be closer.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/ConcurrentPool.java
```diff
@@ -99,6 +106,9 @@ public ConcurrentPool(StrategyType strategyType, int maxEntries, boolean cache)
      */
     public ConcurrentPool(StrategyType strategyType, int maxEntries, boolean cache, ToIntFunction<P> maxMultiplex)
     {
+        if (maxEntries > MAX_RECOMMENDED_SIZE)
+            LOG.warn("{} configured with max size of {} which is above the recommended value {} which could" +
+                " induce serious performance issues.", getClass().getSimpleName(), maxEntries, MAX_RECOMMENDED_SIZE);
```
sbordet: I would keep this at DEBUG level, and shorten the text (e.g. just log something like "created pool max={}", reason being that we don't want this message to pop up for every bucket (and there can be many).

Also, field `maxEntries` should be renamed to `maxSize` and same with parameters.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/CompoundPool.java
sbordet: But does not work, as I can pass a ToIntFunction that returns `1`.
I think this class is better as a private inner class of `ArrayByteBufferPool`, as it's not really reusable that much.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/ConcurrentPool.java
sbordet: WIth such as low value, I am now worried about `ConnectionPool` having a bad performance, in particular for load testing.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
```diff
@@ -163,12 +191,6 @@ public Entry<W> acquire()
             return getWrapped().acquire();
         }
 
-        @Override
-        public Entry<W> acquire(Function<Entry<W>, W> creator)
-        {
-            return getWrapped().acquire(creator);
-        }
-
```
sbordet: This needs to be restored, as `default` methods should be wrapped too.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QueuedPool.java
```diff
@@ -0,0 +1,245 @@
...
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.stream.Stream;
+
+/**
+ * This implementation does not support multiplexing.
+ *
+ * @param <P> the type of the pooled objects
+ */
+public class QueuedPool<P> implements Pool<P>
```
sbordet: I don't like this class being public in jetty-util, but have the big restriction of not being multiplexed.
For example I would not be able to use it for `ConnectionPool`.

I think this implementation should be private to `ArrayByteBufferPool`, but we do need an implementation that supports multiplexing for `ConnectionPool`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QueuedPool.java
```diff
@@ -0,0 +1,245 @@
...
+{
+    private final int maxSize;
+    private final Queue<Pool.Entry<P>> queue = new ConcurrentLinkedQueue<>();
+
+    // This lock protects the 'terminated' field.
+    private final ReadWriteLock rwLock = new ReentrantReadWriteLock();
+    private final Lock rdLock = rwLock.readLock();
+    private final Lock wrLock = rwLock.writeLock();
```
sbordet: Remove these fields, the pattern in the JDK is always `rwLock.readLock().lock()`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QueuedPool.java
```diff
@@ -0,0 +1,245 @@
...
+            return pool.requeue(this);
+        }
+
+        @Override
+        public boolean remove()
+        {
+            if (inUse == null)
+                return false;
+            inUse = null;
+            return true;
```
sbordet: IIUC, this is not necessary because a queued entry is "on-the-fly" out of the pool and no one sees it in this implementation.
Since this is the only place `inUse` is written, seems that it can be just a primitive `boolean`.

sbordet (changes requested):
I'm not convinced that `CompoundPool` and `QueuedPool` should be public in `jetty-util`, as they are ok for ABBP, but that's basically it.

Also, I am worried about `ConnectionPool` performance too now, so I would like to see a more generic solution that supports multiplexing.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/CompoundPool.java
gregw: If we introduce another pool that can handle queued multiplexed pools, then it will get more usage.   It is a util class and I see no problem exposing it.  All util classes can be used badly.   If we provide some static factory methods that easily build combinations that work well together, then that will encourage good usage.        I am OK with a protected constructor an only public static factory methods.  Then if somebody really wanted to do something strange, they could extend.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/ConcurrentPool.java
gregw: Then let's add a priority queue implementation that will sort entries by multiplexed availability.  It should not be too hard to write and may even be OK as the primary pool for the connections, as the concurrent pool has to use a poor strategy to work with multiplexing anyway.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QueuedPool.java
gregw: I don't think we need to make it private.  It is a perfectly good pool implementation that just happens not to take a multiplex int function, so it can't do multiplexing.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QueuedPool.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    private static class QueuedEntry<P> implements Entry<P>
+    {
+        private final QueuedPool<P> pool;
+        //   null  -> terminated
+        //   false -> idle
+        //   true  -> in use
+        private volatile Boolean inUse = Boolean.FALSE;
+        private volatile P pooled;
```
gregw: Two volatile fields just screams race to me!   I preferred my version is an AtomicMarkableReference

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QueuedPool.java
gregw: Go back to my version with the AtomicMarkableReference.  Two volatile fields shouts race to me!  Even if it is not a race, better to use a known good pattern rather than make up our own

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/ConcurrentPool.java
lorban: Agreed.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Pool.java
lorban: Done.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QueuedPool.java
lorban: Done.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QueuedPool.java
lorban: Changed to use `AtomicMarkableReference`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QueuedPool.java
lorban: It's not racy, but I agree it's ugly. We need four states: reserved, enabled/available, enabled/in-use, terminated but that can be encoded in a `AtomicMarkableReference` so I'm going to use that instead as you recommend.

lorban:
I've renamed the `CompoundPool` to `ConcurrentOverflowingToQueuedPool` and moved it closer to `ArrayByteBufferPool` as it specifically designed to solve that issue and it's fairly trivial.

`QueuedPool` can stay in jetty-util IMHO as it's usable the way it is, even if it does not support multiplexing. We can add an extra `Pool` implementation that is both queue-based and supports multiplexing.

gregw:
@lorban I'm sorry but ConcurrentOverflowingToQueuedPool is a silly name. The only thing about that class which that name applies to is the constructor. Everything else about that class is just CompoundPool. 

We should name the class for what it does, not for hours we use it.

lorban:
Okay, I've renamed the class back to `CompoundPool`, explained the overflow in the javadoc instead of the name but made it a private inner class of `ArrayByteBufferPool` to make it more apparent that it's specific to this use-case only.

I've tried to stick to the "best is the enemy of good" mindset when working on this because:
 - we have a proven perf problem with pooled buffers
 - we have not (yet) a proven perf problem with pooled connections
 - buffers don't need multiplexing
 - `QueuedPool` is generic, but writing a multiplexing version would require non-trivial extra work
 - Compounding `ConcurrentPool` with a `QueuedPool` overflow is very specific to the problem at hand, but solves it well

I'm not against working on a multiplexed `QueuedPool` nor genericizing the above, it it's IMHO outside the scope of this PR, and I'm not even convinced there's a real need for a generic overflowing pool.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
```diff
@@ -534,4 +543,128 @@ public Quadratic(int minCapacity, int maxCapacity, int maxBucketSize, long maxHe
...
+        {
+            if (maxSize <= ConcurrentPool.OPTIMAL_MAX_SIZE)
+            {
+                this.primaryPool = new ConcurrentPool<>(strategyType, maxSize, cache);
+                this.secondaryPool = new NullPool<>();
+            }
+            else
+            {
+                this.primaryPool = new ConcurrentPool<>(strategyType, ConcurrentPool.OPTIMAL_MAX_SIZE, cache);
+                this.secondaryPool = new QueuedPool<>(maxSize - ConcurrentPool.OPTIMAL_MAX_SIZE);
```
gregw: This logic has nothing to do with `CompoundPool` and everything to do with `ArrayByteBufferPool`.  It should be in the `ArrayByteBufferPool` constructor and this constructor should just take 2 pools.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
```diff
@@ -444,9 +450,12 @@ private Pool.Entry<RetainableByteBuffer> findOldestEntry(long now, Pool<Retainab
...
         private final int _capacity;
 
-        private RetainedBucket(int capacity, int size)
+        private RetainedBucket(int capacity, int poolSize)
         {
-            _pool = new ConcurrentPool<>(ConcurrentPool.StrategyType.THREAD_ID, size, true);
+            if (poolSize <= ConcurrentPool.OPTIMAL_MAX_SIZE)
+                _pool = new ConcurrentPool<>(ConcurrentPool.StrategyType.THREAD_ID, poolSize, true);
+            else
+                _pool = new CompoundPool<>(ConcurrentPool.StrategyType.THREAD_ID, poolSize, true);
```
gregw: See comment in `CompoundPool` constructor

gregw (commented):
I agree with the approach of making this PR about fixing the known problems in the ABBP.  We can investigate later if we need to generalise more and come up with solutions for multiplexed pools.

I still don't like the constructor of `CompoundPool`, but since  it is private, that is a niggle rather than a show stopper

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
```diff
@@ -88,11 +94,11 @@ public ArrayByteBufferPool(int minCapacity, int factor, int maxCapacity)
      * @param minCapacity the minimum ByteBuffer capacity
      * @param factor the capacity factor
      * @param maxCapacity the maximum ByteBuffer capacity
-     * @param maxBucketSize the maximum number of ByteBuffers for each bucket
+     * @param maxBucketPoolSize the maximum number of ByteBuffers for each bucket
```
sbordet: Please rename back to `maxBucketSize`.

sbordet (changes requested):

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
lorban: Done.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
lorban: Done.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/ArrayByteBufferPool.java
lorban: Done.

gregw (commented):
LGTM

PR #9331, opened 2023-02-08
Jetty 10 Upgrade to Hazelcast 5 and totally disable auto join multicast etc.. (fix build on CI)

olamy:
- upgrade to hazelcast 5.2.1
- disable discovery/autodetection
- simplify disable autodetection
- pass sys property at start


jetty-hazelcast/src/main/java/org/eclipse/jetty/hazelcast/session/HazelcastSessionDataStoreFactory.java
```diff
@@ -63,6 +63,8 @@
 
     private Config serverConfig;
 
+    private boolean autoDetectionConfigDisabled;
+
```
janbartel:
```suggestion
private boolean autoDetectionConfigEnabled = false;
```

jetty-hazelcast/src/main/java/org/eclipse/jetty/hazelcast/session/HazelcastSessionDataStoreFactory.java
```diff
@@ -171,6 +173,7 @@ public SessionDataStore getSessionDataStore(SessionHandler handler)
                             LOG.warn("Hazelcast xml config is missing org.eclipse.jetty.hazelcast.session.SessionDataSerializer - sessions may not serialize correctly");
                     }
                     config.setInstanceName(hazelcastInstanceName);
+                    if (autoDetectionConfigDisabled) config.getNetworkConfig().getJoin().getAutoDetectionConfig().setEnabled(false);
```
janbartel:
```suggestion
config.getNetworkConfig().getJoin().getAutoDetectionConfig().setEnabled(autoDetectionConfigEnabled);
```

jetty-hazelcast/src/main/java/org/eclipse/jetty/hazelcast/session/HazelcastSessionDataStoreFactory.java
```diff
@@ -296,4 +299,17 @@ public void setServerConfig(Config serverConfig)
     {
         this.serverConfig = serverConfig;
     }
+
+    public boolean isAutoDetectionConfigDisabled()
```
janbartel:
```suggestion
    public boolean isAutoDetectionEnabled()
```

jetty-hazelcast/src/main/java/org/eclipse/jetty/hazelcast/session/HazelcastSessionDataStoreFactory.java
```diff
@@ -296,4 +299,17 @@ public void setServerConfig(Config serverConfig)
...
+    {
+        return autoDetectionConfigDisabled;
```
janbartel:
```suggestion
        return autoDetectionConfigEnabled;
```

jetty-hazelcast/src/main/java/org/eclipse/jetty/hazelcast/session/HazelcastSessionDataStoreFactory.java
```diff
@@ -296,4 +299,17 @@ public void setServerConfig(Config serverConfig)
...
+    }
+
+    /**
+     * @param autoDetectionConfigDisabled will disable Hazelcast auto discovery strategy
```
janbartel:
```suggestion
     * @param autoDetectionConfigEnabled will disable Hazelcast auto discovery strategy 
```

jetty-hazelcast/src/main/java/org/eclipse/jetty/hazelcast/session/HazelcastSessionDataStoreFactory.java
```diff
@@ -296,4 +299,17 @@ public void setServerConfig(Config serverConfig)
...
+
+    public boolean isAutoDetectionConfigDisabled()
+    {
+        return autoDetectionConfigDisabled;
+    }
+
+    /**
+     * @param autoDetectionConfigDisabled will disable Hazelcast auto discovery strategy 
+     */
+    public void setAutoDetectionConfigDisabled(boolean autoDetectionConfigDisabled)
```
janbartel:
```suggestion
    public void setAutoDetectionConfigEnabled(boolean autoDetectionConfigEnabled)
```

jetty-hazelcast/src/main/java/org/eclipse/jetty/hazelcast/session/HazelcastSessionDataStoreFactory.java
```diff
@@ -296,4 +299,17 @@ public void setServerConfig(Config serverConfig)
...
+    {
+        this.autoDetectionConfigDisabled = autoDetectionConfigDisabled;
```
janbartel:
```suggestion
        this.autoDetectionConfigEnabled = autoDetectionConfigEnabled;
```

jetty-hazelcast/src/main/java/org/eclipse/jetty/hazelcast/session/HazelcastSessionDataStoreFactory.java
janbartel: I don't really know what the ramification is of defaulting this to `disabled`? I assume we've historically had this `enabled`, and this is just a point release, so might be a surprise to users? Or is this a new parameter in version 5?

jetty-hazelcast/src/test/java/org/eclipse/jetty/hazelcast/session/TestHazelcastSessions.java
```diff
@@ -108,6 +108,7 @@ public void initialize()
         SessionContext sessionContext = new SessionContext("foo", null);
 
         hazelcastSessionDataStoreFactory = new HazelcastSessionDataStoreFactory();
+        hazelcastSessionDataStoreFactory.setAutoDetectionConfigDisabled(true);
```
janbartel:
```suggestion
        hazelcastSessionDataStoreFactory.setAutoDetectionConfigEnabled(false);
```

tests/test-sessions/test-hazelcast-sessions/pom.xml
```diff
@@ -85,7 +92,7 @@
     </dependency>
     <dependency>
        <groupId>com.hazelcast</groupId>
-       <artifactId>hazelcast-all</artifactId>
+       <artifactId>hazelcast</artifactId>
```
janbartel, line 89: Do we need this dependency here if the `jetty-hazelcast` module already has this dependency?

lachlan-roberts (dismissed):
LGTM

Can't verify it works because the failures were always on Jenkins for me, so if this passes the build I think its good.

janbartel (changes requested):
Generally looks good ,but I think it's a little less confusing to call the new parameter `autoDetectionConfigEnabled`, so it resembles more the naming of `getAutoDetectionConfig().setEnabled()`.  I'm not sure if we're now disabling something that was enabled before, or is this a new parameter?

olamy:
> LGTM
> 
> Can't verify it works because the failures were always on Jenkins for me, so if this passes the build I think its good.

the fix is already in 12.0.x branch

jetty-hazelcast/src/main/java/org/eclipse/jetty/hazelcast/session/HazelcastSessionDataStoreFactory.java
olamy: I choose the name to leave the option enable per default.
I would definitely prefer leave it `enabled` per default as it is the default strategy to discovering other hazelcast servers running in the same network  and so creating automatically some clusters. 
But is causing an issue in the kubernetes test environment.

tests/test-sessions/test-hazelcast-sessions/pom.xml
olamy: yes. because `jetty-hazelcast` is in `test` scope. And `test` scope is not transitive.

jetty-hazelcast/src/main/java/org/eclipse/jetty/hazelcast/session/HazelcastSessionDataStoreFactory.java
olamy: the issue is here for long time :) (since 4.1)
just happening more often recently.
https://hazelcast.com/blog/hazelcast-discovery-auto-detection/

olamy: basically this one of the key feature here. when the instance starts, it auto connect to other hazelcast instance in the same network to share data. having it per default is better but not for testing

janbartel: So it can still be called `autoDetectionConfigEnabled` but be `true` by default, and have the CI config set it to `false`?

tests/test-sessions/test-hazelcast-sessions/pom.xml
```diff
@@ -76,16 +83,10 @@
...
-        <exclusion>
-          <groupId>com.hazelcast</groupId>
-          <artifactId>hazelcast</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
     <dependency>
        <groupId>com.hazelcast</groupId>
-       <artifactId>hazelcast-all</artifactId>
+       <artifactId>hazelcast</artifactId>
```
janbartel, line 89: Heh. Well, the whole module is test module, so `scope` test is kinda funny anyway.

jetty-hazelcast/src/main/java/org/eclipse/jetty/hazelcast/session/HazelcastSessionDataStoreFactory.java
olamy: I usually prefer having boolean with default `false` as jvm default...
as this parameter was really for test purpose for one single test and could potentially create some confusions for our users  I removed and use a single configuration file for the test.

tests/test-sessions/test-hazelcast-sessions/pom.xml
olamy: don't know maybe it's wrong but all test-sessions are using such pattern.

janbartel (approved):

PR #9338, opened 2023-02-09
LowResourceMonitor.getReasons should include detailed reason instead …

### LINKED ISSUE #9337: LowResourceMonitor.getReasons should include detailed reason instead of hard-coded message
**Jetty version(s)**
10.0.13

**Description**
`LowResourceMonitor.getReasons` returns a hard-coded string, namely the string representation of the `LowResourceCheck` that is low on resources. It should contain the detailed low-resource reason instead, which would be more useful.

For example, it should return `Server low on threads: x, idleThreads: y` vs the current `Check if the ThreadPool from monitored connectors are lowOnThreads and if all connections number is higher than the allowed maxConnection`.

Proposed diffs:

```
diff --git a/jetty-server/src/main/java/org/eclipse/jetty/server/LowResourceMonitor.java b/jetty-server/src/main/java/org/eclipse/jetty/server/LowResourceMonitor.java
index 07c7ab427d..7529404b17 100644
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/LowResourceMonitor.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/LowResourceMonitor.java
@@ -280,7 +280,7 @@ public class LowResourceMonitor extends ContainerLifeCycle
         {
             if (lowResourceCheck.isLowOnResources())
             {
-                reasons = lowResourceCheck.toString();
+                reasons = lowResourceCheck.getReason();
                 break;
             }
         }
```
If needed, the LowResourceCheck string representation could be appended to the reason.

jluehe:
…of hard-coded message (#9337)

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/LowResourceMonitor.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/LowResourceMonitor.java
@@ -280,7 +280,7 @@ public class LowResourceMonitor extends ContainerLifeCycle
         {
             if (lowResourceCheck.isLowOnResources())
             {
-                reasons = lowResourceCheck.toString();
+                reasons = lowResourceCheck.getReason();
                 break;
             }
         }


olamy:
@jluehe This sounds a good change to me. Thanks!
it looks there is not eclipse/eca recorded under this email at Eclipse. 
Maybe you can change to a recorded one?

jluehe:
Thank you, @olamy! I just re-executed the ECA ...

olamy (dismissed):

jluehe:
@olamy, could you please approve one more time. I force-pushed my updated author info to match what's on file for the ECA, which invalidated your previous approval. Thank you!

olamy (approved):

gregw:
@olamy feel free to merge through to 10, 11 & 12

gregw (approved):

PR #9352, opened 2023-02-14
Fix/jetty 9.4.x cookie cutter legacy

gregw:
Less lenient RFC6265 parsing, but with a RFC6265_LEGACY mode for those that want it.

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/CookieCutter.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/CookieCutter.java
@@ -155,7 +155,7 @@ public class CookieCutter
                 if (inQuoted)
                 {
                     boolean eol = c == 0 && i == hdr.length();
-                    if (!eol && _compliance == CookieCompliance.RFC6265 && isRFC6265RejectedCharacter(inQuoted, c))
+                    if (!eol && _compliance != CookieCompliance.RFC2965 && isRFC6265RejectedCharacter(inQuoted, c))
                     {
                         reject = true;
                         continue;
@@ -199,7 +199,10 @@ public class CookieCutter
                             continue;
 
                         case ';':
-                            reject = true;
+                            if (_compliance == CookieCompliance.RFC6265)
+                                reject = true;
+                            else
+                                unquoted.append(c);
                             continue;
 
                         default:
@@ -343,6 +346,9 @@ public class CookieCutter
                                     continue;
                                 }
 
+                                if (_compliance == CookieCompliance.RFC6265_LEGACY && isRFC6265RejectedCharacter(inQuoted, c))
+                                    reject = true;
+
                                 if (tokenstart < 0)
                                     tokenstart = i;
                                 tokenend = i;
@@ -388,13 +394,8 @@ public class CookieCutter
                                     continue;
                                 }
 
-                                if (_compliance == CookieCompliance.RFC6265)
-                                {
-                                    if (isRFC6265RejectedCharacter(inQuoted, c))
-                                    {
-                                        reject = true;
-                                    }
-                                }
+                                if (_compliance != CookieCompliance.RFC2965 && isRFC6265RejectedCharacter(inQuoted, c))
+                                    reject = true;
 
                                 if (tokenstart < 0)
                                     tokenstart = i;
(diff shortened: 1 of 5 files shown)


jetty-server/src/main/java/org/eclipse/jetty/server/CookieCutter.java
```diff
@@ -390,28 +413,43 @@ else if (tokenstart >= 0)
...
-                c == ',' || // comma
-                c == ';'; // semicolon
+            else
+            {
+                /* From RFC6265 - Section 4.1.1 - Syntax
+                 *  cookie-octet  = %x21 / %x23-2B / %x2D-3A / %x3C-5B / %x5D-7E
+                 *                  ; US-ASCII characters excluding CTLs,
+                 *                  ; whitespace DQUOTE, comma, semicolon,
+                 *                  ; and backslash
+                 */
```
sbordet: I'd remove the comment because it's not what the code does (the code does not check for ` `, `"` and `\`.
Alternatively, can this block be merged with the block below which is identical and has a clarifying comment?

sbordet (dismissed):
Please add the suggested fix, otherwise LGTM.

jetty-server/src/test/java/org/eclipse/jetty/server/CookieCutterTest.java
```diff
@@ -266,12 +267,93 @@ public void testExcessiveSemicolons()
...
+        return Arrays.asList(
+            new Param("A=1; B=2; C=3", "A=1", "B=2", "C=3"),
+            new Param("A=\"1\"; B=2; C=3", "A=1", "B=2", "C=3"),
+            new Param("A=\"1\"; B=\"2\"; C=\"3\"", "A=1", "B=2", "C=3"),
+            new Param("A=1; B=2; C=\"3", "A=1", "B=2"),
+            new Param("A=1 ; B=2; C=3", "A=1", "B=2", "C=3"),
+            new Param("A= 1; B=2; C=3", "A=1", "B=2", "C=3"),
+            new Param("A=\"1; B=2\"; C=3", "C=3"),
+            new Param("A=\"1; B=2; C=3"),
+            new Param("A=\"1 B=2\"; C=3", "A=1 B=2", "C=3"), // Why should A be rejected? Shouldn't it be A=<1 B=2>?
```
lorban: This comment either is outdated and should go away, or space characters should be rejected and there's a bug left.

jetty-server/src/test/java/org/eclipse/jetty/server/CookieCutterTest.java
```diff
@@ -266,12 +267,93 @@ public void testExcessiveSemicolons()
...
+            new Param("A=\"\"1; B=2; C=3", "B=2", "C=3"),
+            new Param("A=\"\" ; B=2; C=3", "A=", "B=2", "C=3"),
+            new Param("A=\"\"; B=2; C=3", "A=", "B=2", "C=3"),
+            new Param("A=1\"\"; B=2; C=3", "B=2", "C=3"),
+            new Param("A=1\"; B=2; C=3", "B=2", "C=3"),
+            new Param("A=1\"1; B=2; C=3", "B=2", "C=3"),
+            /* TODO Use Legacy mode for these
+            new Param("A=\" 1\"; B=2; C=3", "A=1", "B=2", "C=3"), // Why should the whitespaces be trimmed? They were not in the prev impl.
```
lorban: These assertions should be adapted. In these four cases, `A` should be ignored but `B` and `C` are still valid, aren't they?

lorban (dismissed):
I would clean up the commented-out bits of the test, otherwise LGTM.

jetty-server/src/main/java/org/eclipse/jetty/server/CookieCutter.java
gregw: I'll remove the comment.
It is not identical to the code below.  This is the original code and the code below is updated by @lorban.

jetty-server/src/test/java/org/eclipse/jetty/server/CookieCutterTest.java
gregw: We want to keep us much legacy behaviour as possible for these lenient cases, so I don't want to change the code. I'll remove the comment.

lorban (approved):

sbordet (approved):

atennapel:
This did not update the `getSetCookie` in `HttpCookie`. It will cause a `IllegalStateException` if `RFC6265_LEGACY` is used. Unless I am mistaken... Created issue here: https://github.com/eclipse/jetty.project/issues/9717

Issue #9327, opened 2023-02-07
Jetty 12 - Review POMs to remove OSGi directives for internal packages


sbordet:
**Jetty version(s)**
12+

**Description**
There used to be special directives for OSGi in the POMs to include `internal` packages.
This was necessary because `module-info.java` had `export ... to ...` directives, but these are now gone with the work of #9293.

What's left in the POMs for OSGi, if anything, should be removed.

janbartel:
There are some packages that are still called `internal` that are being used outside of the jar in which it is declared. For example:

> Trying to start the bundle org.eclipse.jetty.websocket.core.server that was supposed to be active or resolved.
> org.eclipse.jetty.websocket.core.server failed to start
> org.osgi.framework.BundleException: Could not resolve module: org.eclipse.jetty.websocket.core.server [69]
>   Unresolved requirement: Import-Package: org.eclipse.jetty.websocket.core.internal
>

lachlan-roberts:
As part of #9233 I will remove any export of internal packages from websocket. At that point the  OSGi directives can be removed.

gregw:
This should be fixed by #9356

lachlan-roberts:
I looked at all the `<Export-Package>` tags in the poms and there are no other ones that export internal packages. So I think this can be closed.

PR #9356, opened 2023-02-15
cleanup of websocket to fix JPMS warnings

lachlan-roberts:
remove the "export to" lines from websocket internal packages
move some classes out of the internal packages so they are now visible


jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/CoreSession.java
```diff
@@ -185,6 +196,47 @@ public interface CoreSession extends OutgoingFrames, Configuration
...
+     * @param cause the cause
+     * @param callback the callback on completion of error handling
+     */
+    void processHandlerError(Throwable cause, Callback callback);
+
+    /**
+     * Set the class loader to be used by this {@link CoreSession}.
+     * @param classLoader the ClassLoader.
+     */
+    void setClassLoader(ClassLoader classLoader);
```
sbordet: This is only used once, but `CoreSession.from()` is called twice (as well as `setWebSocketConnection()`) so feels like there is a missing call to `setClassLoader()`.

Also, given that the classloader is available at the time of the call to `from()` can it not be passed as a parameter to `from()`?

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/CoreSession.java
```diff
@@ -185,6 +196,47 @@ public interface CoreSession extends OutgoingFrames, Configuration
...
+
+    /**
+     * Used to set the WebSocketConnection on this {@link CoreSession}.
+     * @param connection the websocket connection.
+     */
+    void setWebSocketConnection(WebSocketConnection connection);
+
     class Empty extends ConfigurationCustomizer implements CoreSession
```
sbordet: Can this be removed in favor of defaults on `CoreSession`? 
Otherwise leave it as is.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/WebSocketUtil.java
```diff
@@ -11,15 +11,15 @@
...
+package org.eclipse.jetty.websocket.core.util;
 
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.util.Base64;
 
 import org.eclipse.jetty.websocket.core.WebSocketConstants;
 
-public final class WebSocketCore
+public final class WebSocketUtil
```
sbordet: All other classes in this package are named with the plural, so rename to `WebSocketUtils`.

jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jakarta-server/src/main/java/org/eclipse/jetty/ee10/websocket/jakarta/server/config/JakartaWebSocketConfiguration.java
```diff
@@ -35,6 +35,6 @@ public JakartaWebSocketConfiguration()
         protectAndExpose("org.eclipse.jetty.ee10.websocket.servlet."); // For WebSocketUpgradeFilter
         protectAndExpose("org.eclipse.jetty.ee10.websocket.jakarta.server.config.");
         protectAndExpose("org.eclipse.jetty.ee10.websocket.jakarta.client.JakartaWebSocketClientContainerProvider");
-        protectAndExpose("org.eclipse.jetty.ee10.websocket.jakarta.client.JakartaWebSocketShutdownContainer");
+        protectAndExpose("org.eclipse.jetty.ee10.websocket.jakarta.client.internal.JakartaWebSocketShutdownContainer");
```
sbordet, line 38: Feels wrong that we protect and expose an internal class, and not sure it really works in JPMS.
I think it's best to make it non-internal like the other one.

jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jakarta-server/src/main/java/org/eclipse/jetty/ee9/websocket/jakarta/server/config/JakartaWebSocketConfiguration.java
```diff
@@ -35,6 +35,6 @@ public JakartaWebSocketConfiguration()
         protectAndExpose("org.eclipse.jetty.ee9.websocket.servlet."); // For WebSocketUpgradeFilter
         protectAndExpose("org.eclipse.jetty.ee9.websocket.jakarta.server.config.");
         protectAndExpose("org.eclipse.jetty.ee9.websocket.jakarta.client.JakartaWebSocketClientContainerProvider");
-        protectAndExpose("org.eclipse.jetty.ee9.websocket.jakarta.client.JakartaWebSocketShutdownContainer");
+        protectAndExpose("org.eclipse.jetty.ee9.websocket.jakarta.client.internal.JakartaWebSocketShutdownContainer");
```
sbordet, line 38: Feels wrong that we protect and expose an internal class, and not sure it really works in JPMS.
I think it's best to make it non-internal like the other one.

sbordet (changes requested):

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/CoreSession.java
lachlan-roberts: In the server case we use context::run so it also calls the notifyEnterScope / notifyExitScope methods.

But I could do the client case with the `Consumer<Runnable>` constructor instead of setting the classloader.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/CoreSession.java
lachlan-roberts: no because ConfigurationCustomizer is a class.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/WebSocketUtil.java
lachlan-roberts: We use this naming convention in other places like `BufferUtil`, `StringUtil`, `ArrayUtil`

jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jakarta-server/src/main/java/org/eclipse/jetty/ee10/websocket/jakarta/server/config/JakartaWebSocketConfiguration.java
lachlan-roberts: Yeah the problem is if its non-internal we have a transitive dependency for the servlet api for the client module.
As the shutdown container extends `ServletContainerInitializer` and `ServletContextListener`.

gregw: It is a strange case that it needs to be exposed so it can be discovered by the service loader in the webapps classloader when running SCIs.
I wonder if now we have an environment classloader, if perhaps we can look for SCIs at that level that are not themselves exposed to the webapp?

jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jakarta-server/src/main/java/org/eclipse/jetty/ee9/websocket/jakarta/server/config/JakartaWebSocketConfiguration.java
gregw: Perhaps we can look at this in another PR, but I think it should be possible to look for SCIs in the environment classloader, that are not exposed to the webapp classloader.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/WebSocketUtil.java
sbordet: @lachlan-roberts singular is wrong as it is supposed to be a collection of utility methods, and we use the plural in many other cases.

Look at that package where `WebSocketUtil` is and you'll find: `InvokerUtils` (plural), `ReflectUtils` (plural), `TextUtils` (plural).
At least in the same package let's use the same naming.

jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jakarta-server/src/main/java/org/eclipse/jetty/ee10/websocket/jakarta/server/config/JakartaWebSocketConfiguration.java
sbordet: Ugh, why do we need a `ServletContainerInitializer` in the client?!?

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/CoreSession.java
sbordet: Not sure I follow.
I'm suggesting to change `from()` to take an additional `ClassLoader` as parameter, and if not necessary we pass `null`, saving a setter on an interface that is only ever used once, with the idea of completely removing `setClassLoader()` from the interface.

sbordet:
@lachlan-roberts also, `jetty-websocket-core-client` has this warning:

```
[INFO] --- maven-bundle-plugin:5.1.8:manifest (default) @ jetty-websocket-core-client ---
[INFO] No MANIFEST.MF file found, generating manifest.
[WARNING] Manifest org.eclipse.jetty.websocket:jetty-websocket-core-client:jar:12.0.0-SNAPSHOT : Unused Export-Package instructions: [org.eclipse.jetty.websocket.core.client.internal.*] 
[INFO] Writing manifest: /home/simon/opensource/jetty/12.0/jetty-core/jetty-websocket/jetty-websocket-core-client/target/classes/META-INF/MANIFEST.MF
```

Something related to #9327?

jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jakarta-server/src/main/java/org/eclipse/jetty/ee10/websocket/jakarta/server/config/JakartaWebSocketConfiguration.java
gregw: So if a client is started by the application inside a web app, we can clean it up when the context stops, else it leaks and holds the context in memory.

sbordet: I thought we have a special jar classifier for that, like we have for the HTTP client?

gregw:
this PR should also remove the work around from the OSGI manifest.

janbartel:
OSGi workaround lines to remove:

./jetty-websocket-core-common/pom.xml:            <Export-Package>*,org.eclipse.jetty.websocket.core.common.internal.*</Export-Package>
./jetty-websocket-core-client/pom.xml:            <Export-Package>*,org.eclipse.jetty.websocket.core.client.internal.*</Export-Package>

jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jakarta-server/src/main/java/org/eclipse/jetty/ee10/websocket/jakarta/server/config/JakartaWebSocketConfiguration.java
sbordet: Okay, after discussion this can remain as it, provided we look at it from the point of view of the environment classloader, rather than the webapp classloader.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/WebSocketUtil.java
sbordet: @lachlan-roberts please rename this as plural.

PR #9383, opened 2023-02-16
Reinstate ee9 jetty runner.

janbartel:
Make ee9 runner part of build, same as ee10 runner is.


gregw (approved):

olamy:
do we need jetty ee8 runner part?

janbartel:
> do we need jetty ee8 runner part?

@gregw ?

gregw:
If it is trivial to add, then sure add it.  But don't spend more than 15 minutes on it

olamy (approved):

PR #9399, opened 2023-02-20
Improve CookieCompliance testing

gregw:
+ Improved handling of CookieCompliance.from method
+ Added tests for request and response cookie handling 
+ Use from in jetty.xml

### DIFF
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CookieParser.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CookieParser.java
@@ -41,7 +41,7 @@ public interface CookieParser
         return new RFC6265CookieParser(handler, compliance, complianceListener);
     }
 
-    void parseField(String field);
+    void parseField(String field) throws InvalidCookieException;
 
     default void parseFields(List<String> rawFields)
     {
@@ -56,5 +56,28 @@ public interface CookieParser
     interface Handler
     {
         void addCookie(String name, String value, int version, String domain, String path, String comment);
+    }
+
+    class InvalidCookieException extends IllegalArgumentException
+    {
+        public InvalidCookieException()
+        {
+            super();
+        }
+
+        public InvalidCookieException(String s)
+        {
+            super(s);
+        }
+
+        public InvalidCookieException(String message, Throwable cause)
+        {
+            super(message, cause);
+        }
+
+        public InvalidCookieException(Throwable cause)
+        {
+            super(cause);
+        }
     }
 }
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CookieParser.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CookieParser.java
@@ -43,7 +43,7 @@ public interface CookieParser
 
     void parseField(String field) throws InvalidCookieException;
 
-    default void parseFields(List<String> rawFields)
+    default void parseFields(List<String> rawFields) throws InvalidCookieException
     {
         // For each cookie field
         for (String field : rawFields)
@@ -58,6 +58,9 @@ public interface CookieParser
         void addCookie(String name, String value, int version, String domain, String path, String comment);
     }
 
+    /**
+     * <p>The exception thrown when a cookie cannot be parsed and {@link CookieCompliance.Violation#INVALID_COOKIES} is not allowed.</p>
+     */
     class InvalidCookieException extends IllegalArgumentException
     {
         public InvalidCookieException()
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCompliance.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCompliance.java
@@ -37,7 +37,7 @@ import static java.util.EnumSet.of;
  */
 public final class HttpCompliance implements ComplianceViolation.Mode
 {
-    protected static final Logger LOG = LoggerFactory.getLogger(HttpCompliance.class);
+    private static final Logger LOG = LoggerFactory.getLogger(HttpCompliance.class);
 
     // These are compliance violations, which may optionally be allowed by the compliance mode, which mean that
     // the relevant section of the RFC is not strictly adhered to.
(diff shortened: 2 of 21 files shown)


gregw:
@olamy these fixes may help with the TCK cookie failure. At the very least the `ServerHttpCookieTest` can be used to recreate the TCK failure to investigate deeper.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCompliance.java
```diff
@@ -222,7 +222,7 @@ public static HttpCompliance valueOf(String name)
      * <dl>
      * <dt>0</dt><dd>No {@link Violation}s</dd>
      * <dt>*</dt><dd>All {@link Violation}s</dd>
-     * <dt>RFC2616</dt><dd>The set of {@link Violation}s application to https://tools.ietf.org/html/rfc2616,
+     * <dt>RFC2616</dt><dd>The set of {@link Violation}s application to <a href="https://tools.ietf.org/html/rfc2616">...</a>,
```
lorban: I think this javadoc should be reverted. Or the link should be given a text other than `...` and other links should be edited too.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -443,7 +444,14 @@ static List<HttpCookie> getCookies(Request request)
             request.getComponents().getCache().setAttribute(CACHE_ATTRIBUTE, cookieCache);
         }
 
-        cookies = cookieCache.getCookies(request.getHeaders());
+        try
+        {
+            cookies = cookieCache.getCookies(request.getHeaders());
+        }
+        catch (IllegalArgumentException iae)
```
lorban: I think it would be less confusing and safer to introduce a `InvalidCookieException` that extends from `RuntimeException`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/RFC6265CookieParser.java
```diff
@@ -390,7 +390,7 @@ else if (_complianceMode.allows(INVALID_COOKIES))
                     }
 
                     if (state == State.END)
-                        throw new IllegalStateException("Invalid cookie");
+                        throw new IllegalArgumentException("Invalid cookie");
```
lorban: Throw a new `InvalidCookieException` that extends from `RuntimeException`?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/RFC6265CookieParser.java
```diff
@@ -323,7 +323,7 @@ else if (_complianceMode.allows(INVALID_COOKIES))
                         }
                         else
                         {
-                            throw new IllegalStateException("Comma cookie separator");
+                            throw new IllegalArgumentException("Comma cookie separator");
```
lorban: Throw a new `InvalidCookieException` that extends from `RuntimeException`?

lorban (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
gregw: I'm generally not a huge fan of lots of exception types.... I normally ask "do you catch them and handle them differently".  In this case I am.

So I think either I change it so that any exception thrown from getCookies is caught and converted to a BadMessage(400, cause); OR I introduce an `InvalidCookieException` and catch only that.  Any other issues in the cookie parser will result in a 500 rather than a 400.

lorban (approved):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CookieParser.java
```diff
@@ -57,4 +57,27 @@ default void parseFields(List<String> rawFields)
     {
         void addCookie(String name, String value, int version, String domain, String path, String comment);
     }
+
+    class InvalidCookieException extends IllegalArgumentException
```
joakime: Why is this a nested private class?
It's exposed via API already.
Can this be a top level class in jetty-http please?

joakime (changes requested):
Just a niggle that the InvalidCookieException is a declared throwable in some APIs, and it a useful exception type to work with, it shouldn't be a nested private class, it should be it's own type.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/CookieParser.java
gregw: It's not private.  It is public in the CookieParser interface.   We scope many things under interfaces in the API without making them top level classes.

gregw:
@olamy any chance of testing this against the TCK before the merge?

joakime (approved):
Not a fan of the hidden inner exception that isn't at least static, but lets get this merged.

olamy:
> @olamy any chance of testing this against the TCK before the merge?

I did but this didn't look to fix the TCK test.

gregw:
@joakime it is not hidden.  It's static by default because it is in an interface. If I add the static keyword, intellij tells me that it is not needed.   It is not different to `Handler.Wrapper`, `Handler.Collection`, `Request.Wrapper`, `Content.Sink`, `Content.Source`. etc.  We now have lots of classes and interfaces scoped inside an interface.

gregw:
I'm putting this PR on hold an updating #9402 to include the bulk of these changes, and then merge them back through from 10.

gregw:
@sbordet @lorban This PR was mostly done in #9402 in jetty-10 and has now been merged forward to jetty-12.  What is left are just a few cleanups to use modern switches and to align the other compliance mode impls.

Can you review and if OK please merge.

joakime:
> @joakime it is not hidden. It's static by default because it is in an interface. If I add the static keyword, intellij tells me that it is not needed. It is not different to `Handler.Wrapper`, `Handler.Collection`, `Request.Wrapper`, `Content.Sink`, `Content.Source`. etc. We now have lots of classes and interfaces scoped inside an interface.

Exceptions will propagate through lots of complex logging scenarios.
The non-static inner, means the outer instance is now tied to the instances of that inner.
This could potentially lead to scenarios which GC is impacted.

The other examples don't propagate through other layers like logging.

gregw:
@joakime it is static.  All classes within an interface are static:
Screenshot from 2023-02-22 08-00-40

PR #9402, opened 2023-02-21
Fix #9334 Cookie Compliance

### LINKED ISSUE #9334: Better support for Cookie RFC 2965 compliance
**Jetty version(s)**
10+

**Description**
`CookieCompliance` should support better violations related to RFC 2965.

gregw:
Fix incorrect change to RFC6265 to not support dollars in cookie names.


gregw:
@sbordet @lorban We were too tough on RFC2965 with our new RFC6265 parser.  An attribute from an old cookie is a valid cookie name in RFC6265.  So for example: `name=value; $Domain=1` is two cookies, one called "name" and the other called "$Domain".    We don't see these cookies in Servlet 5.0, only because the `Cookie` class cannot carry them.  In Servlet 6.0 it can.   
So in this PR, the low level tests all look for multiple cookies in that situation, but in the high level test (ServerHttpCookieTest), the dollar cookies are ignored (but test will need to be updated for jetty-11 & 12).

Best way to see how this now works is to look at the test case: `RFC6265CookieParserTest#testRFC2965Single` and look at how each compliance mode is tested.

sbordet (approved):

PR #9411, opened 2023-02-23
Fix `HugeResourceTest`

### LINKED ISSUE #9408: HugeResourceTest failing
**Jetty version(s)**
12

**Description**
The `HugeResourceTest` was failing for me by never completing on the 1G size.  I tried making a 100M size, but whilst that passed it actually failed.
However, the CI is green, so either this test is not run there or it is working.

For now, I have disabled the test as it is failing for me in 12 HEAD and delaying other work.

lorban:
Fixes #9408.

Fixed double-release bug in `MultiPartFormData` and used `ByteBufferPool.NonPooling` pool to work around #9311.

### DIFF
--- a/jetty-ee10/jetty-ee10-webapp/src/test/java/org/eclipse/jetty/ee10/webapp/HugeResourceTest.java
+++ b/jetty-ee10/jetty-ee10-webapp/src/test/java/org/eclipse/jetty/ee10/webapp/HugeResourceTest.java
@@ -52,6 +52,7 @@ import org.eclipse.jetty.http.HttpFields;
 import org.eclipse.jetty.http.HttpHeader;
 import org.eclipse.jetty.http.HttpMethod;
 import org.eclipse.jetty.http.MultiPart;
+import org.eclipse.jetty.io.ByteBufferPool;
 import org.eclipse.jetty.io.ClientConnector;
 import org.eclipse.jetty.io.Content;
 import org.eclipse.jetty.server.HttpConfiguration;
@@ -70,7 +71,6 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
@@ -83,7 +83,6 @@ import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 @Tag("large-disk-resource")
-@Disabled // TODO investigate
 public class HugeResourceTest
 {
     private static final long KB = 1024;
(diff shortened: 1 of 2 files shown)


jetty-ee10/jetty-ee10-webapp/src/test/java/org/eclipse/jetty/ee10/webapp/HugeResourceTest.java
```diff
@@ -114,11 +113,10 @@ public static void prepareStaticFiles() throws IOException
...
 
-        makeStaticFile(staticBase.resolve("test-1M.dat"), MB);
-        makeStaticFile(staticBase.resolve("test-100M.dat"), 100 * MB);
-        makeStaticFile(staticBase.resolve("test-1G.dat"), GB);
-        // makeStaticFile(staticBase.resolve("test-4G.dat"), 4 * GB);
-        // makeStaticFile(staticBase.resolve("test-10G.dat"), 10 * GB);
+        makeStaticFile(staticBase.resolve("test-1m.dat"), MB);
+        makeStaticFile(staticBase.resolve("test-1g.dat"), GB);
+        // makeStaticFile(staticBase.resolve("test-4g.dat"), 4 * GB);
+        // makeStaticFile(staticBase.resolve("test-10g.dat"), 10 * GB);
```
joakime: The reason for testing 4GB and 10GB were because of various filesystem handling bugs we had in our code (the 2GB threshold and the 8GB threshold in various FileSystem APIs).

jetty-ee10/jetty-ee10-webapp/src/test/java/org/eclipse/jetty/ee10/webapp/HugeResourceTest.java
```diff
@@ -603,8 +603,9 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws S
                     IO.copy(inputStream, byteCounting);
                     out.printf("part[%s].inputStream.length=%d%n", part.getName(), byteCounting.getCount());
                 }
-                catch (IOException e)
+                catch (Throwable e)
                 {
+                    resp.setStatus(500);
```
joakime: Can't set 500 in the majority of cases here, the response is already committed.
Probably better to let the exception just flow out.

jetty-ee10/jetty-ee10-webapp/src/test/java/org/eclipse/jetty/ee10/webapp/HugeResourceTest.java
lorban: I've just restored the `HugeResourceTest.java` file to exactly what it was before Greg disabled it. But I think your comment is worth adding to the code.

lorban: Okay, rethrowing as `AssertionError` then.

jetty-ee10/jetty-ee10-webapp/src/test/java/org/eclipse/jetty/ee10/webapp/HugeResourceTest.java
lorban: Okay, rethrowing as `AssertionError` then.

gregw (approved):

jetty-ee10/jetty-ee10-webapp/src/test/java/org/eclipse/jetty/ee10/webapp/HugeResourceTest.java
sbordet: But 4g and 10g are still commented out?

jetty-ee10/jetty-ee10-webapp/src/test/java/org/eclipse/jetty/ee10/webapp/HugeResourceTest.java
```diff
@@ -131,11 +131,10 @@ public static Stream<Arguments> staticFiles()
...
 
-        ret.add(Arguments.of("test-1M.dat", MB));
-        ret.add(Arguments.of("test-100M.dat", 100 * MB));
-        ret.add(Arguments.of("test-1G.dat", GB));
-        // ret.add(Arguments.of("test-4G.dat", 4 * GB));
-        // ret.add(Arguments.of("test-1Gg.dat", 10 * GB));
+        ret.add(Arguments.of("test-1m.dat", MB));
+        ret.add(Arguments.of("test-1g.dat", GB));
+        // ret.add(Arguments.of("test-4g.dat", 4 * GB));
+        // ret.add(Arguments.of("test-10g.dat", 10 * GB));
```
sbordet: Ditto.

sbordet (changes requested):
I'm confused about the tests for 4g and 10g... They're supposed to be run?

joakime:
> I'm confused about the tests for 4g and 10g... They're supposed to be run?

IMO, the 4g test should (at a minimum) be enabled.
That exists due to a regression we had with FileChannel behaviors across the 2GB limit (per the JVM API).

Both were disabled less than a day ago by @gregw via commit 791c6a7991841a92678bb04952925c24647c8a99

lorban:
I've re-enabled 4G and 10G tests and I increased the timeout to 60s as 10G was passing right at the limit for `testUploadMultipart` and `testUploadMultipartDelayed`.

jetty-ee10/jetty-ee10-webapp/src/test/java/org/eclipse/jetty/ee10/webapp/HugeResourceTest.java
lorban: Restored them.

sbordet (approved):

Issue #9400, opened 2023-02-20
Jetty logs warning with stacktrace when annotation parser encounters module-info.class file inside elasticsearch-x-content jar


d2a-pnagel:
**Jetty version(s)**
11.0.13

**Java version/vendor** `(use: java -version)`
java version "18.0.1.1" 2022-04-22
Java(TM) SE Runtime Environment (build 18.0.1.1+2-6)
Java HotSpot(TM) 64-Bit Server VM (build 18.0.1.1+2-6, mixed mode, sharing)

**OS type/version**
Windows 11

**Description**
When Jetty loads a `.war` file which includes the elasticsearch, or more specifically, elasticsearch-x-content jar of version 8.2.0 and above, (4) warnings with stack traces are logged to the console by the annotation parser. Example:

```
[WARNING] Failed to handle org.eclipse.jetty.annotations.AnnotationParser$ClassInfo@1972bfd8
java.lang.NullPointerException: Cannot invoke "Object.hashCode()" because "key" is null
    at java.util.concurrent.ConcurrentHashMap.get (ConcurrentHashMap.java:936)
    at org.eclipse.jetty.annotations.ClassInheritanceHandler.addToInheritanceMap (ClassInheritanceHandler.java:70)
    at org.eclipse.jetty.annotations.ClassInheritanceHandler.handle (ClassInheritanceHandler.java:57)
    at org.eclipse.jetty.annotations.AnnotationParser$MyClassVisitor.visit (AnnotationParser.java:482)
    at org.objectweb.asm.ClassReader.accept (ClassReader.java:569)
    at org.objectweb.asm.ClassReader.accept (ClassReader.java:424)
    at org.eclipse.jetty.annotations.AnnotationParser.scanClass (AnnotationParser.java:897)
    at org.eclipse.jetty.annotations.AnnotationParser.parseJarEntry (AnnotationParser.java:881)
    at org.eclipse.jetty.annotations.AnnotationParser.lambda$parseJar$2 (AnnotationParser.java:840)
    at java.util.TreeMap$ValueSpliterator.forEachRemaining (TreeMap.java:3215)
    at java.util.stream.ReferencePipeline$Head.forEach (ReferencePipeline.java:762)
    at org.eclipse.jetty.annotations.AnnotationParser.parseJar (AnnotationParser.java:836)
    at org.eclipse.jetty.annotations.AnnotationParser.parse (AnnotationParser.java:737)
[...]

PR #9433, opened 2023-02-24
Issue #9400 Handle scan of weird jar format

janbartel:


janbartel:
The `elasticsearch-x-content` jar has a very odd format, whereby it seems to contain the contents of many other jars, eg:

>   created: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/
>   created: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/META-INF/
>   created: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/META-INF/versions/
>   created: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/META-INF/versions/9/
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/META-INF/versions/9/module-info.class
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/META-INF/MANIFEST.MF
>   created: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/META-INF/services/
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/META-INF/services/com.fasterxml.jackson.core.JsonFactory
>   created: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/META-INF/maven/
>   created: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/META-INF/maven/com.fasterxml.jackson.dataformat/
>   created: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/META-INF/maven/com.fasterxml.jackson.dataformat/jackson-dataformat-smile/
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/META-INF/maven/com.fasterxml.jackson.dataformat/jackson-dataformat-smile/pom.properties
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/META-INF/maven/com.fasterxml.jackson.dataformat/jackson-dataformat-smile/pom.xml
>   created: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/
>   created: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/
>   created: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/
>   created: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/
>   created: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/SmileBufferRecycler.class
>   created: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/databind/
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/databind/SmileMapper$Builder.class
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/databind/SmileMapper.class
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/SmileParserBootstrapper.class
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/SmileParser.class
>   created: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/async/
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/async/NonBlockingParserBase.class
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/async/NonBlockingByteArrayParser.class
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/SmileConstants.class
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/SmileWriteContext.class
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/SmileGenerator$Feature.class
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/SmileGenerator.class
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/SmileParserBase$1.class
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/SmileFactoryBuilder.class
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/SmileFactory.class
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/SmileUtil.class
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/PackageVersion.class
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/SmileGenerator$SharedStringNode.class
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/SmileParser$Feature.class
>  inflated: IMPL-JARS/x-content/jackson-dataformat-smile-2.13.2.jar/com/fasterxml/jackson/dataformat/smile/SmileParserBase.class
>  inflated: IMPL-JARS/x-content/LISTING.TXT

Our class scanning was expecting to find `module-info.class` at the root of the jar, not buried down at some sublevel. Changed the `AnnotationParser` to look for `module-info.class` as the last name element.

joakime (changes requested):
Another thing we can do.

If a directory contains a `.` (period char), it should be ignored (and all sub-content).
(We could expand this to other illegal characters in the future)

That's an invalid directory for a JAR in a ClassLoader and would never be usable.

We should also ensure that the asm produced package name for the scanned class fits in the jar file.

Example: a file in the jar `a/directory/deep/org/eclipse/jetty/Server.class`, which is scanned by ASM shows the package as `org.eclipse.jetty`.
That is a bad class, that cannot be accessed by the ClassLoader.

sbordet:
@janbartel the format:

`/META-INF/versions/9/module-info.class`

is legal and the widely accepted solution to deploy Java 8 compliant jars (that won't look into `/META-INF/versions/`) -- where all the classes are compiled with target=8 -- but have the same jar also work in Java 9+ runtimes (which will look into `/META-INF/versions/`), so we should definitely support this style.

This also means that we should not look into _any_ sublevel, but only under `/META-INF/versions/<version>/module-info.class`, where `<version>` may be any number >= 9.

janbartel:
> @janbartel the format:
> 
> `/META-INF/versions/9/module-info.class`
> 
> is legal and the widely accepted solution to deploy Java 8 compliant jars (that won't look into `/META-INF/versions/`) -- where all the classes are compiled with target=8 -- but have the same jar also work in Java 9+ runtimes (which will look into `/META-INF/versions/`), so we should definitely support this style.
> 
> This also means that we should not look into _any_ sublevel, but only under `/META-INF/versions/<version>/module-info.class`, where `<version>` may be any number >= 9.

@sbordet the point is that we actually _don't_ want to look at `module-info.class` at all, because it is not a useful class to look at for servlet annotations nor the class inheritance hierarchy. The problem was that before this PR, we couldn't correctly _ignore_ a `module-info.class` file buried at some deep level inside a jar.

janbartel:
> Another thing we can do.
> 
> If a directory contains a `.` (period char), it should be ignored (and all sub-content). (We could expand this to other illegal characters in the future)
> 
> That's an invalid directory for a JAR in a ClassLoader and would never be usable.
> 
> We should also ensure that the asm produced package name for the scanned class fits in the jar file.
> 
> Example: a file in the jar `a/directory/deep/org/eclipse/jetty/Server.class`, which is scanned by ASM shows the package as `org.eclipse.jetty`. That is a bad class, that cannot be accessed by the ClassLoader.

@joakime we already do ignore `.` directories.  For the weird `a/directory/deep/org/eclipse/jetty/Server.class` I don't see where that would be retained as `org.eclipse.jetty.Server.class` other than if the class actually reports that as its package.

janbartel:
> @sbordet the point is that we actually don't want to look at module-info.class at all, because it is not a useful class to look at for servlet annotations nor the class inheritance hierarchy. The problem was that before this PR, we couldn't correctly ignore a module-info.class file buried at some deep level inside a jar.

@sbordet to be clear, this PR has nothing to do with correctly parsing `META-INF/versions` which we already do. This is purely about some odd jar that seems to be a zip file containing jars of other jars - see that everything is contained in some sub-directory called `IMPL-JARS/x-content/` - tripping up the parser and triggering an NPE.

gregw:
I would go as far as saying this is not a jar file.  A jar file is a zip file that in the top level has classes in a package hierarchy and a META-INF directory with some meta data.
This is a zip file that has class files and META-INF directories in some other layout that I have never seen before.

I think it is good that we correctly skip `module-info.class` files no matter where they are found.... but should we actually be scanning any of the classes in this zip file, as none of them are in the correct location.   Should we be rejecting any class file whose location in the zip does not match the package the class files advertises (modulo multi-version support).

janbartel:
Before we expend too much more energy on this issue, let's remember that users can always exclude jars from being parsed.

gregw:
Can you at least put in a TODO in the code somewhere saying that we should check that a discovered class file is in it's correct location within the zip file before scanning it.

Can we also push back to the OP and/or the source of that jar and ask WTF?

janbartel:
@gregw we could check the package matches the path of the class, but it would be a lot easier to do in jetty-12 than in jetty-10. I propose that we commit this change as-is, then open a separate enhancement to check package with class location in jetty-12.

joakime:
> Can you at least put in a TODO in the code somewhere saying that we should check that a discovered class file is in it's correct location within the zip file before scanning it.

I think this would be the other way around, scan the class, take the reported fully qualified class, and then check that it's in a valid location.

gregw (approved):
Approving on the basis that an issue will be opened to add in 12 the checking of a correct package during the scan.  Please open issue before merge.

joakime (approved):

janbartel:
@gregw I raised the following for the jetty-12 enhancement: https://github.com/eclipse/jetty.project/issues/9511

PR #9441, opened 2023-02-26
Fixes #9440 - Jetty 12: HttpCookieStore should return cookies for "ws…

### LINKED ISSUE #9440: Jetty 12: HttpCookieStore should return cookies for "ws" schemes
**Jetty version(s)**
12+

**Description**
`HttpClient` is used for sending the WebSocket upgrade request.
For such cases, the request URI has typically the `ws` or `wss` scheme.

The current implementation of `HttpCookieStore.Default` uses the scheme as part of the key used to store cookies, but that is not really necessary, as the scheme is subsequently never used.
The only information that must be retained from the scheme is whether it is a secure scheme or not, which would allow other schemes such as `ws` to work seamlessly, rather than having to perform expensive URI/string mangling to replace `ws` schemes to `http` when accessing the `HttpCookieStore`.

sbordet:
…" schemes

Now only using whether the scheme is secure in the key used to store cookies, not the scheme string.


gregw (approved):
I'm a bit worried that in future we might want the scheme in the key, but let's not generalise before we have an example.

sbordet:
@gregw it's an internal implementation detail, we can restore the scheme if we need to.

PR #9460, opened 2023-03-01
Fix multipart locking and buffer leaks

### LINKED ISSUE #9410: Jetty 12: review locking in `MultiPartFormData` and `MultiPartByteRanges`
**Jetty version(s)**
12

**Description**
Both classes call `Multipart.AbstractPartsListener.fail(Throwable cause)` when `CompletableFuture.completeExceptionally()` is called. This can happen at any time by any thread, so this code has to be thread-safe and currently isn't.

lorban:
Fixes #9410

### DIFF
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
@@ -123,12 +123,13 @@ public class MultiPart
     {
         static final Throwable CLOSE_EXCEPTION = new StaticException("Closed");
 
+        private final AutoLock lock = new AutoLock();
         private final String name;
         private final String fileName;
         private final HttpFields fields;
-        private Content.Source contentSource;
         private Path path;
-        private boolean temporary = true;
+        private Content.Source contentSource;
+        private boolean temporary;
 
         public Part(String name, String fileName, HttpFields fields)
         {
@@ -141,11 +142,15 @@ public class MultiPart
             this.fileName = fileName;
             this.fields = fields != null ? fields : HttpFields.EMPTY;
             this.path = path;
+            this.temporary = true;
         }
 
         private Path getPath()
         {
-            return path;
+            try (AutoLock ignored = lock.lock())
+            {
+                return path;
+            }
         }
 
         /**
@@ -189,9 +194,12 @@ public class MultiPart
          */
         public Content.Source getContentSource()
         {
-            if (contentSource == null)
-                contentSource = newContentSource();
-            return contentSource;
+            try (AutoLock ignored = lock.lock())
+            {
+                if (contentSource == null)
+                    contentSource = newContentSource();
+                return contentSource;
+            }
         }
 
         /**
@@ -260,26 +268,39 @@ public class MultiPart
          */
         public void writeTo(Path path) throws IOException
         {
-            if (this.path == null)
+            Path currentPath = getPath();
+            Path newPath;
+            if (currentPath == null)
             {
                 try (OutputStream out = Files.newOutputStream(path))
                 {
                     IO.copy(Content.Source.asInputStream(newContentSource()), out);
                 }
-                this.path = path;
-                this.temporary = false;
+                newPath = path;
             }
             else
             {
-                this.path = Files.move(this.path, path, StandardCopyOption.REPLACE_EXISTING);
+                newPath = Files.move(currentPath, path, StandardCopyOption.REPLACE_EXISTING);
+            }
+
+            try (AutoLock ignored = lock.lock())
+            {
                 this.temporary = false;
+                this.path = newPath;
             }
         }
 
         public void delete() throws IOException
         {
-            if (this.path != null)
-                Files.delete(this.path);
+            Path path = getPath();
+            if (path != null)
+            {
+                Files.delete(path);
+                try (AutoLock ignored = lock.lock())
+                {
+                    this.path = null;
+                }
+            }
         }
 
         @Override
@@ -293,8 +314,17 @@ public class MultiPart
             try
             {
                 getContentSource().fail(t);
-                if (temporary)
-                    delete();
+                Path path = null;
+                try (AutoLock ignored = lock.lock())
+                {
+                    if (temporary)
+                    {
+                        path = this.path;
+                        this.path = null;
+                    }
+                }
+                if (path != null)
+                    Files.delete(path);
             }
             catch (Throwable x)
             {
@@ -347,15 +377,15 @@ public class MultiPart
      */
     public static class ChunksPart extends Part
     {
-        private final List<Content.Chunk> content;
-        private final List<Content.Source> contentSources = new ArrayList<>();
         private final AutoLock lock = new AutoLock();
+        private final List<Content.Chunk> content = new ArrayList<>();
+        private final List<Content.Source> contentSources = new ArrayList<>();
         private boolean closed = false;
 
         public ChunksPart(String name, String fileName, HttpFields fields, List<Content.Chunk> content)
         {
             super(name, fileName, fields);
-            this.content = Objects.requireNonNull(content);
+            this.content.addAll(content);
             content.forEach(Content.Chunk::retain);
         }
 
@@ -366,32 +396,33 @@ public class MultiPart
             {
                 if (closed)
                     return null;
-                ChunksContentSource newContentSource = new ChunksContentSource(content.stream()
+                // Slice the chunks as new instances with separate reference counters,
+                // because the content sources may not be read, or their chunks could be
+                // further retained, so those chunks must not be linked to the original ones.
+                List<Content.Chunk> chunks = content.stream()
                     .map(chunk -> Content.Chunk.from(chunk.getByteBuffer().slice(), chunk.isLast()))
-                    .toList());
+                    .toList();
+                ChunksContentSource newContentSource = new ChunksContentSource(chunks);
+                chunks.forEach(Content.Chunk::release);
                 contentSources.add(newContentSource);
                 return newContentSource;
             }
         }
 
         @Override
-        public void close()
+        public void fail(Throwable t)
         {
-            List<Content.Source> contentSourcesToFail = null;
+            List<Content.Source> contentSourcesToFail;
             try (AutoLock l = lock.lock())
             {
                 closed = true;
-                if (!contentSources.isEmpty())
-                {
-                    contentSourcesToFail = new ArrayList<>(contentSources);
-                    contentSources.clear();
-                }
+                content.forEach(Content.Chunk::release);
+                content.clear();
+                contentSourcesToFail = List.copyOf(contentSources);
+                contentSources.clear();
             }
-
-            super.close();
-            content.forEach(Content.Chunk::release);
-            if (contentSourcesToFail != null)
-                contentSourcesToFail.forEach(cs -> cs.fail(CLOSE_EXCEPTION));
+            contentSourcesToFail.forEach(c -> c.fail(t));
+            super.fail(t);
         }
 
         @Override
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
@@ -28,7 +28,6 @@ import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 
 import org.eclipse.jetty.io.Content;
-import org.eclipse.jetty.io.Retainable;
 import org.eclipse.jetty.util.IO;
 import org.eclipse.jetty.util.QuotedStringTokenizer;
 import org.eclipse.jetty.util.thread.AutoLock;
@@ -164,12 +163,7 @@ public class MultiPartFormData extends CompletableFuture<MultiPartFormData.Parts
      */
     public Charset getDefaultCharset()
     {
-        return getParts().stream()
-            .filter(part -> "_charset_".equals(part.getName()))
-            .map(part -> part.getContentAsString(US_ASCII))
-            .map(Charset::forName)
-            .findFirst()
-            .orElse(null);
+        return listener.getDefaultCharset();
     }
 
     /**
@@ -299,9 +293,9 @@ public class MultiPartFormData extends CompletableFuture<MultiPartFormData.Parts
     }
 
     // Only used for testing.
-    List<MultiPart.Part> getParts()
+    int getPartsSize()
     {
-        return listener.getParts();
+        return listener.getPartsSize();
     }
 
     /**
@@ -454,6 +448,11 @@ public class MultiPartFormData extends CompletableFuture<MultiPartFormData.Parts
                             if (ensureFileChannel())
                             {
                                 // Write existing memory chunks.
+                                List<Content.Chunk> partChunks;
+                                try (AutoLock ignored = lock.lock())
+                                {
+                                    partChunks = List.copyOf(this.partChunks);
+                                }
                                 for (Content.Chunk c : partChunks)
                                 {
                                     write(c.getByteBuffer());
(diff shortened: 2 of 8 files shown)


jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
```diff
@@ -126,9 +126,10 @@ public abstract static class Part implements Closeable
         private final String name;
         private final String fileName;
         private final HttpFields fields;
-        private Content.Source contentSource;
         private Path path;
-        private boolean temporary = true;
+        private final AutoLock lock = new AutoLock();
```
sbordet: Move this up close to the other `final` fields. I like initialized fields to be even before `final` fields that are initialized in the constructor.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
```diff
@@ -267,18 +275,21 @@ public void writeTo(Path path) throws IOException
...
+            try (AutoLock ignored = lock.lock())
+            {
                 this.temporary = false;
             }
         }
 
         public void delete() throws IOException
         {
-            if (this.path != null)
+            if (this.path != null && Files.exists(this.path))
```
sbordet: `Files.exist()` seems redundant.
If the implementation sets a value for `path` we know it exists; if it is set by the application, it deserves `Files.delete()` to throw.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
```diff
@@ -267,18 +275,21 @@ public void writeTo(Path path) throws IOException
...
                 Files.delete(this.path);
```
sbordet: Should we not null-out `this.path` after `Files.delete()`?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
```diff
@@ -347,15 +361,18 @@ public String toString()
...
         private final AutoLock lock = new AutoLock();
+        private final List<Content.Chunk> content = new ArrayList<>();
+        private final List<Content.Source> contentSources = new ArrayList<>();
         private boolean closed = false;
 
         public ChunksPart(String name, String fileName, HttpFields fields, List<Content.Chunk> content)
         {
             super(name, fileName, fields);
-            this.content = Objects.requireNonNull(content);
+            try (AutoLock l = lock.lock())
```
sbordet: Don't think it's necessary to lock here (because the constructor publishes)?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
```diff
@@ -366,30 +383,53 @@ public Content.Source newContentSource()
...
         @Override
         public void close()
         {
             List<Content.Source> contentSourcesToFail = null;
             try (AutoLock l = lock.lock())
             {
+                if (closed)
+                    return;
                 closed = true;
                 if (!contentSources.isEmpty())
```
sbordet: Remove the if to make the code more readable?
It's just the `close()` operation, if `contentSources.isEmpty()` (unlikely) we just to very little more work.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
```diff
@@ -435,7 +429,13 @@ public void onPartContent(Content.Chunk chunk)
...
             {
                 long maxFileSize = getMaxFileSize();
-                fileSize += buffer.remaining();
+                long fileSize;
+                try (AutoLock ignored = lock.lock())
+                {
+                    fileSize = this.fileSize;
+                    fileSize += buffer.remaining();
+                    this.fileSize = fileSize;
+                }
```
sbordet: I don't think `this.fileSize` needs locking. It is only ever accessed by the parser in the callbacks.
Same for `this.memoryFileSize`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
```diff
@@ -445,7 +445,13 @@ public void onPartContent(Content.Chunk chunk)
                 long maxMemoryFileSize = getMaxMemoryFileSize();
                 if (maxMemoryFileSize >= 0)
                 {
-                    memoryFileSize += buffer.remaining();
+                    long memoryFileSize;
+                    try (AutoLock ignored = lock.lock())
```
sbordet: No locking necessary here?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
```diff
@@ -468,32 +479,41 @@ public void onPartContent(Content.Chunk chunk)
...
             int remaining = buffer.remaining();
             while (remaining > 0)
             {
-                int written = fileChannel.write(buffer);
-                if (written == 0)
-                    throw new NonWritableChannelException();
-                remaining -= written;
+                try (AutoLock ignored = lock.lock())
+                {
+                    int written = fileChannel.write(buffer);
```
sbordet: Would not do the write with the lock held.
The only reason for the lock is to access `fileChannel`?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
```diff
@@ -468,32 +479,41 @@ public void onPartContent(Content.Chunk chunk)
...
+                        throw new NonWritableChannelException();
+                    remaining -= written;
+                }
             }
         }
 
         private void close()
         {
-            try
+            try (AutoLock ignored = lock.lock())
```
sbordet: Would not do the close with the lock held.
Can we just have a `fileChannel()` method that locks and returns the field?

sbordet (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
lorban: True, a barrier is emitted at the end of the constructor because there are final fields. Let me remove this lock.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
lorban: I thought this `if` made things clearer, but it can be removed. Let me do it.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
lorban: You're right, but those are the only two fields that don't require locking, and they are read and modified under lock in `onPart`. For consistency and ease of mind, I think it's better to always access those two under lock.

Are you sure these two locking bits bother you that much?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
lorban: Same as above.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
lorban: Yes, it's for reading the `fileChannel` field. I could move the `write` out of the lock, but since buffers are unlikely to be big I thought this wouldn't matter.

Let me rework this.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
lorban: Same as above.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
lorban: Done.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
lorban: I've done the change to set `this.path` to null after deletion, but this has one side-effect: calling `getPath()` after `delete()` now returns null while it previously returned the deleted path.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
lorban: This is a slight behavior change compared to what we used to do. I like nulling out the path after deletion better, though.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
```diff
@@ -435,7 +429,13 @@ public void onPartContent(Content.Chunk chunk)
             if (fileName != null || isUseFilesForPartsWithoutFileName())
             {
                 long maxFileSize = getMaxFileSize();
-                fileSize += buffer.remaining();
+                long fileSize;
+                try (AutoLock ignored = lock.lock())
+                {
+                    fileSize = this.fileSize;
+                    fileSize += buffer.remaining();
+                    this.fileSize = fileSize;
```
gregw: I find this way of doing it very confusing.  Isn't this just:
```suggestion
                    fileSize = this.fileSize += buffer.remaining();
```

or maybe:

```suggestion
                    this.fileSize += buffer.remaining();
                    fileSize = this.fileSize;
```

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
```diff
@@ -445,7 +445,13 @@ public void onPartContent(Content.Chunk chunk)
                 long maxMemoryFileSize = getMaxMemoryFileSize();
                 if (maxMemoryFileSize >= 0)
                 {
-                    memoryFileSize += buffer.remaining();
+                    long memoryFileSize;
+                    try (AutoLock ignored = lock.lock())
+                    {
+                        memoryFileSize = this.memoryFileSize;
+                        memoryFileSize += buffer.remaining();
+                        this.memoryFileSize = memoryFileSize;
```
gregw: same as above

gregw (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
lorban: Done.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
lorban: Done.

gregw (approved):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
sbordet: Not done?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
sbordet: Locking identifies the fields that possibly are subject to concurrency, so helps understanding what are the concurrency entry points.
Having other stuff locked that does not need to seems more confusing to me.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
```diff
@@ -293,8 +314,11 @@ public void fail(Throwable t)
             try
             {
                 getContentSource().fail(t);
-                if (temporary)
-                    delete();
+                try (AutoLock ignored = lock.lock())
+                {
+                    if (temporary)
+                        delete();
```
sbordet: As there are only a handful filesystem operations, I'm leaning towards doing them out of locking.
Can we have a `private void delete(Path)` called outside locks that is called after `this.path = null`?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
```diff
@@ -468,23 +475,29 @@ public void onPartContent(Content.Chunk chunk)
...
+            }
         }
 
         private void write(ByteBuffer buffer) throws Exception
         {
             int remaining = buffer.remaining();
             while (remaining > 0)
             {
-                int written = fileChannel.write(buffer);
+                int written = fileChannel().write(buffer);
```
sbordet: `fileChannel()` may return `null`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
```diff
@@ -552,28 +583,34 @@ public void onFailure(Throwable failure)
...
+        }
+
         private void delete()
         {
-            try
+            try (AutoLock ignored = lock.lock())
             {
-                if (fileChannel != null)
+                if (filePath != null)
                     Files.delete(filePath);
```
sbordet: Can we do filesystem I/O outside locking?

sbordet (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
lorban: Done.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
lorban: Done.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
lorban: Okay, but then `fileSize` and `memoryFileSize` are to be reset outside lock scope in `onPart` for consistency.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
lorban: Sure, but apart from throwing a different type of exception than NPE, nothing different should be done.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
sbordet: @lorban this is not resolved.
Also, removing the `if` will make the code in `fail()` and `close()` nearly identical, so perhaps extract it in a common method?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
```diff
@@ -383,13 +433,14 @@ public void close()
...
                 {
-                    contentSourcesToFail = new ArrayList<>(contentSources);
+                    contentSourcesToFail = List.copyOf(contentSources);
                     contentSources.clear();
                 }
+                content.forEach(Content.Chunk::release);
+                content.clear();
             }
 
             super.close();
```
sbordet: Calling `super` calls `fail()`, which will do nothing because the data structures have been cleared.
I find this confusing (as you think "am I failing twice?"), so perhaps just don't call `super` at all, as we are doing all the work anyway in this overridden version.

sbordet (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
lorban: Actually, `fail()` and `close()` are almost doing the same thing, and since `super.close()` already calls `fail()`, `close()` can just go away.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPart.java
lorban: Now done.

sbordet (approved):

PR #9461, opened 2023-03-02
Failure to instantiate a websocket endpoint is now properly reported

### LINKED ISSUE #9412: Jetty 12: WebSocket hangs when ServerEndpointConfig.Configurator.getEndpointInstance() throws
**Jetty version(s)**
12

**Description**
Must verify that a proper upgrade request is replied timely (i.e. no blocking or such) when `ServerEndpointConfig.Configurator.getEndpointInstance()` throws `InstantiationException`.

Currently, the exception is caught by `JakartaWebSocketCreator` and `null` is returned, which eventually blocks into `WebSocketUpgradeFilter` because the callback is not completed.

lorban:
Closes #9412

### DIFF
--- a/jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jakarta-server/src/main/java/org/eclipse/jetty/ee10/websocket/jakarta/server/internal/JakartaWebSocketCreator.java
+++ b/jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jakarta-server/src/main/java/org/eclipse/jetty/ee10/websocket/jakarta/server/internal/JakartaWebSocketCreator.java
@@ -166,9 +166,10 @@ public class JakartaWebSocketCreator implements WebSocketCreator
             Object endpoint = config.getConfigurator().getEndpointInstance(endpointClass);
             return new ConfiguredEndpoint(endpoint, config);
         }
-        catch (InstantiationException e)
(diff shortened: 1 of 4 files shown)


joakime:
The Jakarta WebSocket Spec has a `jakarta.websocket.DeploymentException` that we have to be careful to implement properly.

lorban:
@joakime The [EE10 DeploymentException javadoc](https://jakarta.ee/specifications/webprofile/10/apidocs/jakarta/websocket/deploymentexception) says _Checked exception indicating some kind of failure either to publish an endpoint on its server, or a failure to connect a client to its server._ EE 9 and 8 versions of this class have identical javadoc.

If my understanding of what you're saying is correct:
 - On the server, the `CustomUpgradeServlet` used in `ProgrammaticWebSocketUpgradeTest` should get a `DeploymentException` in response to its `ServerContainer.upgradeHttpToWebSocket()` call.
 - On the client, `WebSocketContainer.connectToServer()` should throw `DeploymentException`.

I can make the necessary changes, but it looks like Jetty 11 also does not respect this contract.

lorban:
@joakime I think there is indeed an issue with `DeploymentException` so I've opened #9466 to track it, and I'm going to happily ignore it in the context of this PR as it's out of this scope IMHO.

sbordet (approved):

PR #9462, opened 2023-03-03
Improved `URIUtil.safeDecodePath(String)`

### LINKED ISSUE #9444: Unexpected encoding in request.getPathInfo() with Jetty 12 beta 0
**Jetty version(s)**
12.0.0.beta0

**Java version/vendor**

**OS type/version**
mac os

**Description**
When requesting a URL like

http://localhost:8888/view/VAADIN/dynamic/resource/5/4a2dc788-7f04-43e4-8471-70d713d6d001/file%20name

with a servlet mapping of `/view/*` then `request.getPathInfo()` in 

Jetty < 12 returns `"/VAADIN/dynamic/resource/5/4a2dc788-7f04-43e4-8471-70d713d6d001/file name"` while 
Jetty 12 returns `"/VAADIN/dynamic/resource/5/4a2dc788-7f04-43e4-8471-70d713d6d001/file%20name"`

Is this an intentional change in some spec or accidental?

This breaks handling of resources like "file name" in Vaadin when run in 12 beta0 as Vaadin assumes the path info is decoded and then encodes it to match with a stored, encoded value.

joakime:
More testing of Issue #9444 revealed other cases where we should be safer with our decoding.

This includes new tests for `URIUtil.safeDecodePath(String)`, along with the ability to configure `Utf8Appendable` instances to use a replacement character instead of always throwing an exception in the case of bad UTF-8 sequences.

### DIFF
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
@@ -480,7 +480,7 @@ public final class URIUtil
      */
     public static String decodePath(String path, int offset, int length)
     {
-        return decodePath(path, offset, length, Utf8StringBuilder::append);
+        return decodePath(path, offset, length, Utf8StringBuilder::append, true);
     }
 
     /**
@@ -504,11 +504,19 @@ public final class URIUtil
      */
     public static String safeDecodePath(String path, int offset, int length)
     {
-        return decodePath(path, offset, length, URIUtil::safePathAppend);
+        return decodePath(path, offset, length, URIUtil::safePathAppend, false);
     }
 
     private static void safePathAppend(Utf8StringBuilder builder, byte b)
     {
+        // control characters
+        if (((b >= 0x00) && (b <= 0x1F)) || (b == 0x7F))
+        {
+            // TODO: perhaps we should just ensure these are encoded instead?
+            builder.append("�"); // unicode replacement character
+            return;
+        }
+
         switch (b)
         {
             case '/' -> builder.append("%2F");
@@ -523,7 +531,7 @@ public final class URIUtil
      * @see #canonicalPath(String)
      * @see #normalizePath(String)
      */
-    public static String decodePath(String path, int offset, int length, BiConsumer<Utf8StringBuilder, Byte> decoder)
+    private static String decodePath(String path, int offset, int length, BiConsumer<Utf8StringBuilder, Byte> decoder, boolean fallbackToIso88591)
     {
         try
         {
@@ -538,6 +546,7 @@ public final class URIUtil
                         if (builder == null)
                         {
                             builder = new Utf8StringBuilder(length);
+                            builder.setThrowOnInvalid(fallbackToIso88591);
                             builder.append(path, offset, i - offset);
                         }
                         if ((i + 2) < end)
@@ -571,6 +580,7 @@ public final class URIUtil
                         if (builder == null)
                         {
                             builder = new Utf8StringBuilder(path.length());
+                            builder.setThrowOnInvalid(fallbackToIso88591);
                             builder.append(path, offset, i - offset);
                         }
 
@@ -602,6 +612,8 @@ public final class URIUtil
         {
             if (LOG.isDebugEnabled())
                 LOG.debug("{} {}", path.substring(offset, offset + length), e.toString());
+            if (!fallbackToIso88591)
+                throw new IllegalArgumentException("Not UTF-8, not decoding in ISO-8859-1", e);
             return decodeISO88591Path(path, offset, length);
         }
         catch (IllegalArgumentException e)
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
@@ -509,14 +509,6 @@ public final class URIUtil
 
     private static void safePathAppend(Utf8StringBuilder builder, byte b)
     {
-        // control characters
-        if (((b >= 0x00) && (b <= 0x1F)) || (b == 0x7F))
-        {
-            // TODO: perhaps we should just ensure these are encoded instead?
-            builder.append("�"); // unicode replacement character
-            return;
-        }
-
         switch (b)
         {
             case '/' -> builder.append("%2F");
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
@@ -81,6 +81,7 @@ public abstract class Utf8Appendable implements CharsetStringBuilder
     };
 
     private int _codep;
+    private boolean _throwOnInvalid = true;
 
     public Utf8Appendable(Appendable appendable)
     {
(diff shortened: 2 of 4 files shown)


jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/Utf8AppendableTest.java
```diff
@@ -139,6 +139,30 @@ public void testGermanUmlauts(Class<Utf8Appendable> impl) throws Exception
...
+        bytes[2] = (byte)0xD8;
+        bytes[3] = (byte)0xA9;
+
+        Utf8Appendable buffer = impl.getDeclaredConstructor().newInstance();
+        for (int i = 0; i < bytes.length; i++)
+        {
+            buffer.append(bytes[i]);
+        }
+
+        assertEquals("ةة", buffer.toString());
```
joakime, line 163: This test case seems out of place for this PR.
But I thought I discovered a bug in Utf8Appendable with bidi unicode glyphs.
Turns out it works as expected.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -504,11 +504,19 @@ public static String safeDecodePath(String path)
...
     public static String safeDecodePath(String path, int offset, int length)
     {
-        return decodePath(path, offset, length, URIUtil::safePathAppend);
+        return decodePath(path, offset, length, URIUtil::safePathAppend, false);
     }
 
     private static void safePathAppend(Utf8StringBuilder builder, byte b)
     {
+        // control characters
+        if (((b >= 0x00) && (b <= 0x1F)) || (b == 0x7F))
```
gregw: There is no need to do this.   If a URI encodes control characters, then it is fine for the resulting String to include them.  They are not ambiguous and if needed we can re-encode them to get a URI.

We only need to protect reserved characters:
-  `/` as that is a path segment delimiter
- `?` as that delimits to query string... actually we might not need to protect this as we have already stripped out the query.
- `%` so we can know if character is left encoded or if it was an encoded %

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -602,6 +612,8 @@ public static String decodePath(String path, int offset, int length, BiConsumer<
         {
             if (LOG.isDebugEnabled())
                 LOG.debug("{} {}", path.substring(offset, offset + length), e.toString());
+            if (!fallbackToIso88591)
+                throw new IllegalArgumentException("Not UTF-8, not decoding in ISO-8859-1", e);
```
gregw, line 608: I also don't think we need this... unless an illegal UTF8 character can be decoded as `/` in iso8859-1, which I don't believe is possible... but even if it were, then we'd leave the `/` encoded and there would be nothing ambigous.

gregw: if anything, I'm dubious that we should ever fall back to iso8859-1

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/Utf8AppendableTest.java
gregw: I like this test.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/URIUtilTest.java
```diff
@@ -221,6 +215,83 @@ public void testBadDecodePath(String encodedPath)
...
+            Arguments.of("/foo/.../bar", "/foo/.../bar"),
+            Arguments.of("/foo/%2e/bar", "/foo/./bar"), // Not by the RFC, but safer
+            Arguments.of("/foo/%2e%2e/bar", "/foo/../bar"), // Not by the RFC, but safer
+            Arguments.of("/foo/%2e%2e%2e/bar", "/foo/.../bar")
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("safeDecodePathSource")
+    public void testSafeDecodePath(String encodedPath, String decodedPath)
```
gregw, line 290: A good test to have, but I think you are making too many things safe.

gregw (changes requested):
I don't think we need to protect control characters.
I'm moderately sure we don't need extra protection of bad UTF-8
I'm pondering if we should ever fallback to ISO8859-1 and would like to work out why we do that or if anybody else does.  Maybe time to stop it.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
joakime: I think you might be right (again).

From the testing of this PR, this protection has broken things ...

```
org.eclipse.jetty.ee10.proxy.BalancerServletTest.testRewrittenBalancerWithEncodedURI()

Expected: a string containing "pathInfo='/test/\n'"
     but: was "requestURI='/context/mapping/test/%0A'
servletPath='/mapping'
pathInfo='/test/?'
"
```

joakime: Removed this block.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/URIUtilTest.java
```diff
@@ -221,6 +215,84 @@ public void testBadDecodePath(String encodedPath)
...
+            Arguments.of("/f\u20ac\u20ac/bar", "/f€€/bar"),
+
+            // Encoded UTF-8 unicode (euro)
+            Arguments.of("/f%e2%82%ac%E2%82%AC/bar", "/f€€/bar"),
+
+            // Encoded delimiters
+            Arguments.of("/foo%2fbar", "/foo%2Fbar"),
+            Arguments.of("/foo%252fbar", "/foo%252fbar"),
+            Arguments.of("/foo%3bbar", "/foo;bar"),
+            Arguments.of("/foo%3fbar", "/foo%3Fbar"),
```
sbordet, line 238: Why were the `f` in `%2f` capitalized?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
```diff
@@ -81,6 +81,7 @@ public abstract class Utf8Appendable implements CharsetStringBuilder
     };
 
     private int _codep;
+    private boolean _throwOnInvalid = true;
```
sbordet, line 84: Do we really need this?
If we keep it, it cannot be mutable, should be set as `final` in the constructor and never modified.

sbordet (commented):
I'll defer to @gregw.

I think falling back to ISO 8859-1 is a thing of the past, as it seems to me that every new spec, everybody and everything is on UTF-8, so I won't do it.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/URIUtilTest.java
gregw: That's because of our canonical path mechanism. We produce an encoded path that can be string compared. I.e. we take out all variations, leave only characters that need encoding encoded and uniform capitalization

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
gregw: I think not. See my other PR that is taking another approach to solving this problem.

joakime:
Closing as PR #9479 was the final version that was merged.

PR #9496 has some of the bits around removing the ISO-8859-1 fallback.

PR #9455, opened 2023-03-01
Decode safe Path for #9444

gregw:
Added URIUtil.decodeSafePath for EE10, to allow for %2F and %25 to remain encoded in the servlet API.

PR #9465, opened 2023-03-05
Fully decode #9444

### LINKED ISSUE #9444: Unexpected encoding in request.getPathInfo() with Jetty 12 beta 0

gregw:
This is an alternative to #9462 as a fix for #9444, replacing changes made by #9455 

getServletPath and getPathInfo will never return an encoded path segment. Instead they will throw an IllegalArgumentException if they are called when there is a URI with violations.


gregw:
@joakime this is a WIP, with many tests broken. But I think the solution is to change the tests.  Let's hangout early next week to discuss tests

gregw:
So to be explicit with the intent of this:
- The methods `getServletPath` and `getPathInfo` will never EVER return any encoded characters
- The method `getContextPath` will continue to return the canonically encoded context path (as always)
- We will match to context using the canonically encoded path
- We will match to filters and servlets using the fully decoded path
- By default we allow no violations to the URI RFC
- If we are configured with any ambiguous violations, then they will play out as they do via matching above... but if anything calls  `getServletPath` and `getPathInfo` then a `BadMessage$IllegalArgumentException` is thrown.

I should add this to the javadoc somwhere.

joakime:
Did you mean to merge PR #9474 into this PR?

gregw:
> Did you mean to merge PR #9474 into this PR?

Yes. It was split out from this PR, as it is needed here

gregw:
close in favour of #9479

PR #9469, opened 2023-03-06
Fixes #9391 - port/move Jetty WebSocket APIs, client and server to je…

### LINKED ISSUE #9391: Jetty 12: port/move Jetty WebSocket APIs, client and server to jetty-core
**Jetty version(s)**
12

**Description**
Would be great if it would be possible to use Jetty WebSocket APIs in applications without the need of Jakarta classes (both Servlet and WebSocket).

Probably the Jetty APIs can be moved, same for the client, but for the server there would be a bit more work to do to make them independent from Jakarta classes.

For example have `WebSocketHandler` instead of `WebSocketFilter`/`WebSocketServlet` to bootstrap WebSocket, and higher level APIs to deal with messages (and not frames).

sbordet:
…tty-core.

* Moved -api, -common, -client to jetty-core/jetty-websocket.
* Implemented jetty-core/jetty-websocket/jetty-websocket-jetty-server using only Jetty core APIs, not Servlet.
* Fixed Graceful shutdown order.
* Fixed mistakes in HttpFieldsWrapper, wrongly calling HttpHeader.name() instead of asString().
* Updated tests to pass cleanly.
* Fixed BOMs and POM dependencies.
* Introduced websocket-jetty.mod and websocket-jetty-client.mod, now used by ee10's Jetty WebSocket.
* Fixed OSGi references to old artifactIds.


jetty-core/jetty-websocket/jetty-websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/internal/ServerUpgradeResponseImpl.java
```diff
@@ -55,6 +56,8 @@ public String getAcceptedSubProtocol()
     public void setAcceptedSubProtocol(String protocol)
     {
         negotiation.setSubprotocol(protocol);
+        // Use the nested headers to avoid recursion.
+        response.getHeaders().put(HttpHeader.SEC_WEBSOCKET_SUBPROTOCOL, protocol);
```
gregw, line 60: It is not clear why this was moved from `WebScoketNegotiation.setProtocol`?
The comment is appreciated, but would be good to say who checks it and thus avoids recursing?

jetty-core/jetty-websocket/jetty-websocket-jetty-server/src/main/java/org/eclipse/jetty/websocket/server/WebSocketUpgradeHandler.java
```diff
@@ -0,0 +1,60 @@
...
+import org.eclipse.jetty.websocket.core.server.WebSocketServerComponents;
+
+public class WebSocketUpgradeHandler extends Handler.Wrapper
+{
+    public static WebSocketUpgradeHandler from(Server server, ContextHandler context)
+    {
+        WebSocketUpgradeHandler wsHandler = new WebSocketUpgradeHandler(WebSocketServerComponents.ensureWebSocketComponents(server, context));
+        ServerWebSocketContainer container = wsHandler.container;
+        context.getContext().setAttribute(WebSocketContainer.class.getName(), container);
+        context.addManaged(container);
```
gregw: why is the context directly managing the container?  Surely it is managed by the WSUHandler, which itself will be managed by the context?

It is kind of strange that this method does not insert the WSUHandler into the context?

jetty-core/jetty-websocket/jetty-websocket-jetty-server/src/main/java/org/eclipse/jetty/websocket/server/WebSocketUpgradeHandler.java
```diff
@@ -0,0 +1,60 @@
...
+        return wsHandler;
+    }
+
+    private final ServerWebSocketContainer container;
+
+    private WebSocketUpgradeHandler(WebSocketComponents components)
+    {
+        this.container = new ServerWebSocketContainer(new WebSocketMappings(components));
```
gregw, line 95: should addBean the container here

gregw (changes requested):
A couple of initial comments.... will look more later today

jetty-core/jetty-websocket/jetty-websocket-jetty-server/src/main/java/org/eclipse/jetty/websocket/server/ServerWebSocketContainer.java
```diff
@@ -0,0 +1,278 @@
...
+{
+    private static final Logger LOG = LoggerFactory.getLogger(ServerWebSocketContainer.class);
+
+    private final List<WebSocketSessionListener> listeners = new ArrayList<>();
+    private final SessionTracker sessionTracker = new SessionTracker();
+    private final Configuration configuration = new Configuration();
+    private final WebSocketMappings mappings;
+    private final FrameHandlerFactory factory;
+
+    ServerWebSocketContainer(WebSocketMappings mappings)
```
lachlan-roberts: Any reason for this to be private. Seems the only way for this to be created is from the `WebSocketUpgradeHandler`.

There is a use case for the `JettyWebSocketServerContainer` where no `WebSocketUpgradeFilter` or `WebSocketServlet` is ever created or any mappings added, and the upgrades can still be done manually through the container with the programmatic upgrade method.

Maybe this is something we want to support here as well?

lachlan-roberts (approved):

lachlan-roberts:
@sbordet approved, but can you respond to the review comments from me and Greg.

jetty-core/jetty-websocket/jetty-websocket-core-server/src/main/java/org/eclipse/jetty/websocket/core/server/internal/ServerUpgradeResponseImpl.java
sbordet: The comment was expanded to specify that the header was set on the _nested_ headers, not the wrapped ones.

jetty-core/jetty-websocket/jetty-websocket-jetty-server/src/main/java/org/eclipse/jetty/websocket/server/WebSocketUpgradeHandler.java
sbordet: Done.

jetty-core/jetty-websocket/jetty-websocket-jetty-server/src/main/java/org/eclipse/jetty/websocket/server/WebSocketUpgradeHandler.java
sbordet: Now the container is managed by the WSUH.

Application code may want to insert another `Handler` between the `ContextHandler` and the WSUH, so it is not done here.

jetty-core/jetty-websocket/jetty-websocket-jetty-server/src/main/java/org/eclipse/jetty/websocket/server/ServerWebSocketContainer.java
sbordet: Agreed, done.

paulrutter:
@sbordet i'm in the process of moving from Jetty 9 (through Apache Felix Http) to Jetty 12, by using the latest Jetty 12 Felix Http bundle (https://github.com/apache/felix-dev/tree/master/http/jetty12).

This is going well, except for the websockets part. I used Jetty Websockets in Jetty 9, which were part of the Jetty classes bundles in the Felix Http bundle, but these are no longer there in 12.

https://github.com/apache/felix-dev/blob/master/http/jetty12/pom.xml doesn't seem to include the websocket bundles. What would be the best way to approach this? Do i need a patched Felix bundle with additional dependencies on the Jetty websocket bundles? Or is there another way?

Thanks in advance.

@cziegeler for visibility.

sbordet:
@paulrutter you need to add the Jetty WebSocket dependencies, I think.

This should help:
https://eclipse.dev/jetty/documentation/jetty-12/programming-guide/index.html#pg-migration-11-to-12

In particular:
https://eclipse.dev/jetty/documentation/jetty-12/programming-guide/index.html#pg-websocket

Jetty jars should be OSGi bundles already. If not, report back to us (in a different issue).

You can now avoid the dependency on Jakarta EE APIs, if you want, and just use the Jetty core WebSocket APIs, see https://eclipse.dev/jetty/documentation/jetty-12/programming-guide/index.html#pg-server-websocket-jetty.

paulrutter:
I opened https://github.com/apache/felix-dev/pull/298 to get the websocket bundles added to felix http jetty. I got it to work this way.

PR #9471, opened 2023-03-07
Fix #9468 Space in Cookie name

### LINKED ISSUE #9468: Jetty 11.0.14 is less tolerant of non-compliant cookies than 11.0.13
**Jetty version(s)**
11.0.14

**OS type/version**
Linux 5.4.0-120-generic

**Description**
Jetty 11.0.14 is less tolerant of non-compliant cookies than 11.0.13.

Specifically, when Jetty 11.0.14 is presented with a cookie containing a space in the value, `org.eclipse.jetty.server.Request.getCookies()` throws a `BadMessageException`, whereas Jetty 11.0.13 parses the cookie successfully.

This is a particular problem because the cookies an application receives might be outside its control: The browser will send `myapp.example.com` invalid cookies set by `badapp.example.com` on the `.example.com` domain.

Reviewing the code, I think the intention was that 11.0.14 would behave in the same way as 11.0.13, but a failure case has slipped through.

**How to reproduce?**

Use this class:
```java
package com.example;

import jakarta.servlet.http.HttpServlet;
import jakarta.servlet.http.HttpServletRequest;
import jakarta.servlet.http.HttpServletResponse;
import org.eclipse.jetty.server.handler.gzip.GzipHandler;
import org.eclipse.jetty.servlet.ServletContextHandler;
import org.eclipse.jetty.servlet.ServletHolder;

import java.io.IOException;
import java.io.Writer;
import java.net.URI;
import java.net.http.HttpClient;
import java.net.http.HttpRequest;
import java.net.http.HttpResponse;
import java.net.http.HttpResponse.BodyHandlers;
import java.util.UUID;

public class CookieServer {
        final Server server = new Server(8080);
        final ServletContextHandler servletContextHandler = new ServletContextHandler();
        servletContextHandler.addServlet(new ServletHolder(new FooServlet()), "/foo");
        server.setHandler(servletContextHandler);
        server.start();
        try {
            final HttpResponse<Void> httpResponse = HttpClient.newHttpClient().send(
                    HttpRequest.newBuilder(new URI("http://localhost:8080/foo"))
                            .setHeader("cookie", "PS_DEVICEFEATURES=width:1920 height:1080")
                            .build(),
                    BodyHandlers.discarding()
            );
            System.out.println("httpResponse.statusCode() = " + httpResponse.statusCode());
        } finally {
            server.stop();
        }
    }

    private static final class FooServlet extends HttpServlet {
        @Override
        protected void doGet(HttpServletRequest req, HttpServletResponse resp) {
            try {
                req.getCookies();
            } catch (Exception e) {
                e.printStackTrace();
                throw e;
            }
        }
    }
}
```
Outputs:
```
org.eclipse.jetty.http.BadMessageException: 400: Invalid cookie
	at org.eclipse.jetty.server.Cookies.getCookies(Cookies.java:106)
	at org.eclipse.jetty.server.Request.getCookies(Request.java:870)
	at com.example.CookieServer$FooServlet.doGet(CookieServer.java:42)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:500)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:587)
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:764)
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:529)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:221)
	at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1382)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:176)
	at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:484)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:174)
	at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1304)
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:129)
	at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:122)
	at org.eclipse.jetty.server.Server.handle(Server.java:563)
	at org.eclipse.jetty.server.HttpChannel.lambda$handle$0(HttpChannel.java:505)
	at org.eclipse.jetty.server.HttpChannel.dispatch(HttpChannel.java:762)
	at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:497)
	at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:282)
	at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:314)
	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:100)
	at org.eclipse.jetty.io.SelectableChannelEndPoint$1.run(SelectableChannelEndPoint.java:53)
	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:936)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1080)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: org.eclipse.jetty.http.CookieParser$InvalidCookieException: Invalid cookie

Caused by: org.eclipse.jetty.http.CookieParser$InvalidCookieException: Invalid cookie
	at org.eclipse.jetty.http.RFC6265CookieParser.parseField(RFC6265CookieParser.java:381)
	at org.eclipse.jetty.http.CookieParser.parseFields(CookieParser.java:50)
	at org.eclipse.jetty.server.Cookies.getCookies(Cookies.java:102)
	... 25 more
```

gregw:
Added a violation to allow unquoted spaces in cookie values to fix #9468


jetty-http/src/test/java/org/eclipse/jetty/http/RFC6265CookieParserLenientTest.java
```diff
@@ -74,6 +74,8 @@ public static Stream<Arguments> data()
             Arguments.of("abc= \"x\" ", "abc", "x"),
             Arguments.of("abc= \"x\" ;", "abc", "x"),
             Arguments.of("abc= \"x\" ; ", "abc", "x"),
+            Arguments.of("abc = x y z ", "abc", "x y z"),
+            Arguments.of("abc = x  y  z ", "abc", "x  y  z"),
```
joakime: Can you add a test case that also uses a value with `token` separators (in this case a `:` colon)?

``` java
Arguments.of("abc=a:x b:y c:z", "abc", "a:x b:y c:z")
```

joakime (changes requested):
One more test case please.

jetty-http/src/test/java/org/eclipse/jetty/http/RFC6265CookieParserLenientTest.java
joakime: What would happen if the input was `abc=a;x b;y c;z` as well? (semi-colon)

jetty-http/src/main/java/org/eclipse/jetty/http/CookieCompliance.java
```diff
@@ -94,7 +94,12 @@ public enum Violation implements ComplianceViolation
         /**
          * Whitespace was found around the cookie name and/or around the cookie value.
          */
-        OPTIONAL_WHITE_SPACE("https://www.rfc-editor.org/rfc/rfc6265#section-5.2", "White space around name/value");
+        OPTIONAL_WHITE_SPACE("https://www.rfc-editor.org/rfc/rfc6265#section-5.2", "White space around name/value"),
+
+        /**
+         * Allow spaces within values without quotes.
+         */
+        SPACE_IN_VALUES("https://www.rfc-editor.org/rfc/rfc6265#section-5.2", "Space in value");
```
sbordet, line 102: Rename to singular, `SPACE_IN_VALUE`.

jetty-http/src/main/java/org/eclipse/jetty/http/RFC6265CookieParser.java
```diff
@@ -220,7 +223,13 @@ else if (_complianceMode.allows(INVALID_COOKIES))
                     break;
 
                 case IN_VALUE:
-                    if (c == ';' || c == ',' || c == ' ' || c == '\t')
+                    if (c == ' ' && _complianceMode.allows(SPACE_IN_VALUES))
```
sbordet, line 226: Do we want to cover also `\t`?
If so, rename the Violation to be `WHITESPACE_IN_VALUE`.
Maybe not such a good idea, it would be almost impossible to restore the correct sequence of spaces and tabs if they are interleaved.

jetty-http/src/main/java/org/eclipse/jetty/http/RFC6265CookieParser.java
```diff
@@ -241,6 +250,33 @@ else if (_complianceMode.allows(INVALID_COOKIES))
                     }
                     break;
 
+                case SPACE_IN_VALUE:
+                    if (c == ' ')
+                    {
+                        spaces++;
+                    }
+                    else if (c == ';' || c == ',' || c == '\t')
```
sbordet, line 258: Regardless of the comment above, I don't think we can treat `\t` as a separator between cookies, can we?!?

jetty-http/src/test/java/org/eclipse/jetty/http/RFC6265CookieParserLenientTest.java
```diff
@@ -49,75 +49,77 @@ public static Stream<Arguments> data()
...
+            Arguments.of("abc = ; ", new String[]{"abc", ""}),
+            Arguments.of("abc = x ", new String[]{"abc", "x"}),
+            Arguments.of("abc=\"\"", new String[]{"abc", ""}),
+            Arguments.of("abc= \"\" ", new String[]{"abc", ""}),
+            Arguments.of("abc= \"x\" ", new String[]{"abc", "x"}),
+            Arguments.of("abc= \"x\" ;", new String[]{"abc", "x"}),
+            Arguments.of("abc= \"x\" ; ", new String[]{"abc", "x"}),
+            Arguments.of("abc = x y z ", new String[]{"abc", "x y z"}),
+            Arguments.of("abc = x  y  z ", new String[]{"abc", "x  y  z"}),
+            Arguments.of("abc=a:x b:y c:z", new String[]{"abc", "a:x b:y c:z"}),
```
sbordet, line 80: Add also:

`abc=x y;def=w z`
`abc="x y";def=w z`
`abc=x y;def="w z"`

sbordet (changes requested):

jetty-http/src/main/java/org/eclipse/jetty/http/CookieCompliance.java
gregw: @sbordet Most of the violations are in the plural.  It is more natural to say you allow the plural than to allow the singular.

jetty-http/src/main/java/org/eclipse/jetty/http/RFC6265CookieParser.java
gregw: I'm torn.  Easy to add if somebody asks, so let's wait until somebody asks.

jetty-http/src/main/java/org/eclipse/jetty/http/RFC6265CookieParser.java
gregw: We don't.  This just changes state to END, which can accept a tab as optional white space around a value, but not as a separator.

jetty-http/src/main/java/org/eclipse/jetty/http/RFC6265CookieParser.java
gregw: I've decided not to add `\t` support unless somebody asks for it.

jetty-http/src/main/java/org/eclipse/jetty/http/CookieCompliance.java
gregw: maybe should be `SPACES_IN_VALUES`?

joakime (approved):

jetty-http/src/main/java/org/eclipse/jetty/http/RFC6265CookieParser.java
sbordet: Not sure I follow. If we encounter `\t` the value ends, so it is a terminator char for the value, hence a separator between cookies, like the other chars in the `if` statement.

`\t` is a control char, so it's forbidden in every RFC, isn't it? 
Why do we treat it specially? Historical reasons?

sbordet (approved):
Just a glitch on how `\t` is treated specially (which is orthogonal to this issue, but it has highlighted this special treatment), otherwise LGTM.

gregw:
@sbordet tab is treated specially as we have a violation that allows optional white space before/after a value.  RFC6265 defines OWS in terms of WSP, which in RFC5234 is space or tab.    Hence when we extended where OWS could be in the violation, we also included tab.

In this PR, we are allowing a space to be in the value, which is different to OWS, hence the decision to not include tab.

PR #9474, opened 2023-03-07
BadMessage interface

gregw:
Convert class BadMessageException to BadMessage interface that allows different types of BadMessage exceptions. This follows the pattern of the QuietException interface.

### DIFF
--- a/jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/transport/internal/HttpReceiverOverHTTP.java
+++ b/jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/transport/internal/HttpReceiverOverHTTP.java
@@ -24,7 +24,7 @@ import org.eclipse.jetty.client.transport.HttpClientTransportOverHTTP;
 import org.eclipse.jetty.client.transport.HttpExchange;
 import org.eclipse.jetty.client.transport.HttpReceiver;
 import org.eclipse.jetty.client.transport.HttpResponse;
-import org.eclipse.jetty.http.BadMessageException;
+import org.eclipse.jetty.http.HttpException;
 import org.eclipse.jetty.http.HttpField;
 import org.eclipse.jetty.http.HttpMethod;
 import org.eclipse.jetty.http.HttpParser;
@@ -539,7 +539,7 @@ public class HttpReceiverOverHTTP extends HttpReceiver implements HttpParser.Res
     }
 
     @Override
-    public void badMessage(BadMessageException failure)
+    public void badMessage(HttpException failure)
     {
         HttpExchange exchange = getHttpExchange();
         if (exchange == null || unsolicited)
@@ -550,7 +550,7 @@ public class HttpReceiverOverHTTP extends HttpReceiver implements HttpParser.Res
         {
             HttpResponse response = exchange.getResponse();
             response.status(failure.getCode()).reason(failure.getReason());
-            failAndClose(new HttpResponseException("HTTP protocol violation: bad response on " + getHttpConnection(), response, failure));
+            failAndClose(new HttpResponseException("HTTP protocol violation: bad response on " + getHttpConnection(), response, HttpException.asThrowable(failure)));
         }
     }
 
--- a/jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/transport/internal/HttpReceiverOverHTTP.java
+++ b/jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/transport/internal/HttpReceiverOverHTTP.java
@@ -550,7 +550,7 @@ public class HttpReceiverOverHTTP extends HttpReceiver implements HttpParser.Res
         {
             HttpResponse response = exchange.getResponse();
             response.status(failure.getCode()).reason(failure.getReason());
-            failAndClose(new HttpResponseException("HTTP protocol violation: bad response on " + getHttpConnection(), response, HttpException.asThrowable(failure)));
+            failAndClose(new HttpResponseException("HTTP protocol violation: bad response on " + getHttpConnection(), response, (Throwable)failure));
         }
     }
 
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/BadMessageException.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/BadMessageException.java
@@ -16,17 +16,24 @@ package org.eclipse.jetty.http;
 /**
  * <p>Exception thrown to indicate a Bad HTTP Message has either been received
  * or attempted to be generated.  Typically these are handled with either 400
- * or 500 responses.</p>
+ * responses.</p>
  */
 @SuppressWarnings("serial")
-public class BadMessageException extends RuntimeException
+public class BadMessageException extends HttpException.RuntimeException
 {
-    final int _code;
-    final String _reason;
-
     public BadMessageException()
     {
-        this(400, null, null);
+        this(null, null);
+    }
+
+    public BadMessageException(String reason)
+    {
+        this(reason, null);
+    }
+
+    public BadMessageException(String reason, Throwable cause)
+    {
+        this(HttpStatus.BAD_REQUEST_400, reason, cause);
     }
 
     public BadMessageException(int code)
@@ -34,38 +41,24 @@ public class BadMessageException extends RuntimeException
         this(code, null, null);
     }
 
-    public BadMessageException(String reason)
-    {
-        this(400, reason);
-    }
-
-    public BadMessageException(String reason, Throwable cause)
-    {
-        this(400, reason, cause);
-    }
-
     public BadMessageException(int code, String reason)
     {
         this(code, reason, null);
     }
 
-    public BadMessageException(int code, Throwable cause)
-    {
-        this(code, null, cause);
-    }
-
     public BadMessageException(int code, String reason, Throwable cause)
     {
-        super(code + ": " + reason, cause);
-        _code = code;
-        _reason = reason;
+        super(code, reason, cause);
+        assert code >= 400 && code < 500;
     }
 
+    @Override
     public int getCode()
     {
         return _code;
     }
 
+    @Override
     public String getReason()
     {
         return _reason;
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/BadMessageException.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/BadMessageException.java
@@ -51,16 +51,4 @@ public class BadMessageException extends HttpException.RuntimeException
         super(code, reason, cause);
         assert code >= 400 && code < 500;
     }
-
-    @Override
-    public int getCode()
-    {
-        return _code;
-    }
-
-    @Override
-    public String getReason()
-    {
-        return _reason;
-    }
 }
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpException.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpException.java
@@ -28,15 +28,10 @@ public interface HttpException extends QuietException
 
     String getReason();
 
-    static Throwable asThrowable(HttpException httpException)
-    {
-        return httpException instanceof Throwable throwable ? throwable : null;
-    }
-
-    static void rethrowRuntime(HttpException httpException)
+    static void throwAsUnchecked(HttpException httpException)
     {
         if (httpException instanceof Throwable throwable)
-            ExceptionUtil.ifExceptionThrowRuntime(throwable);
+            ExceptionUtil.ifExceptionThrowUnchecked(throwable);
     }
 
     /**
@@ -47,8 +42,8 @@ public interface HttpException extends QuietException
     @SuppressWarnings("serial")
     class RuntimeException extends java.lang.RuntimeException implements HttpException
     {
-        final int _code;
-        final String _reason;
+        private final int _code;
+        private final String _reason;
 
         public RuntimeException(int code)
         {
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpException.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpException.java
@@ -88,8 +88,8 @@ public interface HttpException extends QuietException
     @SuppressWarnings("serial")
     class IllegalArgumentException extends java.lang.IllegalArgumentException implements HttpException
     {
-        final int _code;
-        final String _reason;
+        private final int _code;
+        private final String _reason;
 
         public IllegalArgumentException(int code)
         {
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpException.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpException.java
@@ -1,0 +1,128 @@
+//
+// ========================================================================
+// Copyright (c) 1995 Mort Bay Consulting Pty Ltd and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.http;
+
+import org.eclipse.jetty.io.QuietException;
+import org.eclipse.jetty.util.ExceptionUtil;
+
+/**
+ * <p>A tagging interface for Exceptions that carry a HTTP response code and reason.</p>
+ * <p>Exception sub-classes that implement this interface will be caught by the container
+ * and the {@link #getCode()} used to send a response.</p>
+ */
+@SuppressWarnings("serial")
+public interface HttpException extends QuietException
+{
+    int getCode();
+
+    String getReason();
+
+    static Throwable asThrowable(HttpException httpException)
+    {
+        return httpException instanceof Throwable throwable ? throwable : null;
+    }
+
+    static void rethrowRuntime(HttpException httpException)
+    {
+        if (httpException instanceof Throwable throwable)
+            ExceptionUtil.ifExceptionThrowRuntime(throwable);
+    }
+
+    /**
+     * <p>Exception thrown to indicate a Bad HTTP Message has either been received
+     * or attempted to be generated.  Typically these are handled with either 400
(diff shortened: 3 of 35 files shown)


jetty-core/jetty-client/src/test/java/org/eclipse/jetty/client/http/HttpReceiverOverHTTPTest.java
```diff
@@ -206,7 +206,7 @@ public void testReceiveBadResponse(HttpCompliance compliance) throws Exception
 
         ExecutionException e = assertThrows(ExecutionException.class, () -> listener.get(5, TimeUnit.SECONDS));
         assertThat(e.getCause(), instanceOf(HttpResponseException.class));
-        assertThat(e.getCause().getCause(), instanceOf(BadMessageException.class));
+        assertThat(e.getCause().getCause(), instanceOf(BadMessage.RuntimeException.class));
```
joakime: Can't this just be `RuntimeBadMessageException` ?

gregw: It could be just `BadMessage.class`, but it is a test, so no harm testing for a specific type.

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/ErrorPageTest.java
```diff
@@ -390,11 +390,11 @@ public void testBadMessage() throws Exception
...
             String response = _connector.getResponse("GET /app?baa=%88%A4 HTTP/1.0\r\n\r\n");
             assertThat(response, Matchers.containsString("HTTP/1.1 400 Bad Request"));
-            assertThat(response, Matchers.containsString("ERROR_PAGE: /BadMessageException"));
+            assertThat(response, Matchers.containsString("ERROR_PAGE: /BadMessage"));
             assertThat(response, Matchers.containsString("ERROR_MESSAGE: Unable to parse URI query"));
             assertThat(response, Matchers.containsString("ERROR_CODE: 400"));
-            assertThat(response, Matchers.containsString("ERROR_EXCEPTION: org.eclipse.jetty.http.BadMessageException: 400: Unable to parse URI query"));
-            assertThat(response, Matchers.containsString("ERROR_EXCEPTION_TYPE: class org.eclipse.jetty.http.BadMessageException"));
+            assertThat(response, Matchers.containsString("ERROR_EXCEPTION: org.eclipse.jetty.http.BadMessage$RuntimeException: 400: Unable to parse URI query"));
+            assertThat(response, Matchers.containsString("ERROR_EXCEPTION_TYPE: class org.eclipse.jetty.http.BadMessage$RuntimeException"));
```
joakime: This bothers me, as inner classes for working with servlet error pages is confusing enough.
Can this please not be an inner class and be it's own top level?

gregw: The Jetty-12 style is heavily using inner classes of interfaces.   It could be a top level class, but that is against the style we have been using.

gregw: @joakime see QuietException, HttpFields, HttpCookie, HttpURI, Handler, Content, etc. as examples of this style.

sbordet (changes requested):
I don't understand this change.

`BadMessage` is-an exception so should end with `Exception` in the name to follow the convention.

The 2 inner classes are never caught separately to act differently depending on the type, so I don't understand why we need 2 of them?

`BadMessage.IllegalArgumentException` is never thrown, only caught in 1 place where also `BadMessage.RuntimeException` is caught, so these changes seems to change nothing from having just one class?

gregw:
I redid this is a less intrusive way (forced push sorry).

I've kept `BadMessageException`, but it is only used for 4xx responses. It extends the new `HttpException.RuntimeException` class.

All the handling of `BadMessageException` has been converted to checking for `HttpException`

gregw:
@sbordet just to record our hangout....
We can't have a tagging interface extend `Throwable` or `Exception` as they are not themselves interfaces.
So I can call it `HttpException` (even though `HttpThrowable` would be less inaccurate) like `QuietException` but is is-not-a `Exception` other than by convention.  I've added convenience static methods to do any casting back to a `Throwable` as needed.

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/ErrorPageTest.java
joakime: Great, but none of those examples are common for users to capture in a Servlet error page.

gregw: @joakime  it is highly unlikely that webapps will map BadMessageException to an error page.   If they want an error page for 400 responses, they just map code 400 rather than a non standard exception.

But see latest changes anyway, as BadMEssageException is back as a top level class.

jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/transport/internal/HttpReceiverOverHTTP.java
```diff
@@ -550,7 +550,7 @@ public void badMessage(BadMessageException failure)
         {
             HttpResponse response = exchange.getResponse();
             response.status(failure.getCode()).reason(failure.getReason());
-            failAndClose(new HttpResponseException("HTTP protocol violation: bad response on " + getHttpConnection(), response, failure));
+            failAndClose(new HttpResponseException("HTTP protocol violation: bad response on " + getHttpConnection(), response, HttpException.asThrowable(failure)));
```
sbordet: Honestly, I would just cast: `_failure = (Throwable)failure;`.
I'd prefer to fail the cast, rather than having `_failure==null` in the unlikely case we mess up and make one non-exception class implement `HttpException`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/BadMessageException.java
```diff
@@ -16,56 +16,49 @@
...
+    @Override
     public int getCode()
     {
         return _code;
     }
 
+    @Override
     public String getReason()
     {
         return _reason;
```
sbordet: No need for these, they are in the base class.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpException.java
```diff
@@ -0,0 +1,128 @@
...
+public interface HttpException extends QuietException
+{
+    int getCode();
+
+    String getReason();
+
+    static Throwable asThrowable(HttpException httpException)
+    {
+        return httpException instanceof Throwable throwable ? throwable : null;
+    }
```
sbordet: I would remove this and replace its usages with a plain cast to `Throwable`, which fails if we mess up.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpException.java
```diff
@@ -0,0 +1,128 @@
...
+
+    static void rethrowRuntime(HttpException httpException)
```
sbordet: I would prefer this to be `static RuntimeException asRuntimeException(HttpException httpException)` and in the places where it is used have `throw HttpException.asRuntimeException(httpException);`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpException.java
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * <p>Exception thrown to indicate a Bad HTTP Message has either been received
+     * or attempted to be generated.  Typically these are handled with either 400
+     * or 500 responses.</p>
+     */
+    @SuppressWarnings("serial")
+    class RuntimeException extends java.lang.RuntimeException implements HttpException
+    {
+        final int _code;
+        final String _reason;
```
sbordet: Make them `private`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpException.java
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * <p>Exception thrown to indicate a Bad HTTP Message has either been received
+     * or attempted to be generated.  Typically these are handled with either 400
+     * or 500 responses.</p>
+     */
+    @SuppressWarnings("serial")
+    class IllegalArgumentException extends java.lang.IllegalArgumentException implements HttpException
+    {
+        final int _code;
+        final String _reason;
```
sbordet: Make them `private`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
```diff
@@ -429,18 +429,17 @@ else if (status == HttpStatus.NO_CONTENT_204 || status == HttpStatus.NOT_MODIFIE
...
                 catch (BufferOverflowException e)
                 {
                     LOG.trace("IGNORED", e);
                     return Result.HEADER_OVERFLOW;
                 }
                 catch (Exception e)
                 {
-                    throw new BadMessageException(INTERNAL_SERVER_ERROR_500, e.getMessage(), e);
+                    if (e instanceof HttpException httpException)
+                        HttpException.rethrowRuntime(httpException);
```
sbordet: Like done elsewhere, you don't need this, you can just `throw e`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpTester.java
```diff
@@ -428,9 +428,9 @@ public boolean content(ByteBuffer ref)
         }
 
         @Override
-        public void badMessage(BadMessageException failure)
+        public void badMessage(HttpException failure)
         {
-            throw failure;
+            HttpException.rethrowRuntime(failure);
```
sbordet: See previous comment, I would prefer here: `throw HttpException.asRuntimeException(failure);`.
Or perhaps just the cast: `throw (RuntimeException)failure;` since all `HttpException`s are `RuntimeException`s.

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/Dispatcher.java
```diff
@@ -196,17 +196,17 @@ protected void forward(ServletRequest request, ServletResponse response, Dispatc
                     {
                         baseRequest.mergeQueryParameters(old_uri.getQuery(), _uri.getQuery());
                     }
-                    catch (BadMessageException e)
+                    catch (Throwable e)
                     {
-                        // Only throw BME if not in Error Dispatch Mode
+                        // Only throw HttpException if not in Error Dispatch Mode
                         // This allows application ErrorPageErrorHandler to handle BME messages
```
sbordet:
```suggestion
                        // This allows application ErrorPageErrorHandler to handle the HttpException reason.
```

jetty-ee9/jetty-ee9-nested/src/test/java/org/eclipse/jetty/ee9/nested/RequestTest.java
```diff
@@ -220,7 +221,7 @@ public void testParamExtraction() throws Exception
                 request.getParameterMap();
                 return false;
             }
-            catch (BadMessageException e)
+            catch (Exception e)
```
sbordet:
```suggestion
            catch (Throwable e)
```

jetty-ee9/jetty-ee9-nested/src/test/java/org/eclipse/jetty/ee9/nested/RequestTest.java
```diff
@@ -575,7 +576,7 @@ public void testBadUtf8ParamExtraction() throws Exception
                 request.getParameter("param");
                 return false;
             }
-            catch (BadMessageException e)
+            catch (Exception e)
```
sbordet:
```suggestion
            catch (Throwable e)
```

jetty-ee9/jetty-ee9-tests/jetty-ee9-test-client-transports/src/test/java/org/eclipse/jetty/ee9/test/client/transport/ServerTimeoutsTest.java
```diff
@@ -325,9 +325,10 @@ protected void service(HttpServletRequest request, HttpServletResponse response)
                             break;
                     }
                 }
-                catch (BadMessageException x)
+                catch (Exception x)
```
sbordet, line 328:
```suggestion
                catch (Throwable x)
```

sbordet (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpException.java
gregw: Can't do that as we'd need a multi-variant return of Error or RuntimeException.
I'll rename to `throwAsUnchecked(HttpException)`

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpTester.java
gregw: There is no reason why HttpExceptions must be RuntimeExceptions, they could be Errors in future.

sbordet: No, that would be wrong, as `Error`s are only thrown by the JVM runtime, not applications.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpException.java
sbordet: Not sure I follow. An `HttpException` cannot be an `Error`, and it is always a `RuntimeException`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpException.java
```diff
@@ -0,0 +1,123 @@
...
+
+import org.eclipse.jetty.io.QuietException;
+import org.eclipse.jetty.util.ExceptionUtil;
+
+/**
+ * <p>A tagging interface for Exceptions that carry a HTTP response code and reason.</p>
+ * <p>Exception sub-classes that implement this interface will be caught by the container
+ * and the {@link #getCode()} used to send a response.</p>
+ */
+@SuppressWarnings("serial")
```
sbordet: What's with this suppressions? Just remove them, these exceptions are not serializable anyway and we don't want them to be.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpGenerator.java
```diff
@@ -429,18 +429,17 @@ else if (status == HttpStatus.NO_CONTENT_204 || status == HttpStatus.NOT_MODIFIE
...
                 {
                     LOG.trace("IGNORED", e);
                     return Result.HEADER_OVERFLOW;
                 }
                 catch (Exception e)
                 {
-                    throw new BadMessageException(INTERNAL_SERVER_ERROR_500, e.getMessage(), e);
+                    if (e instanceof HttpException)
+                        throw e;
+
```
sbordet: Remove newline.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpException.java
```diff
@@ -0,0 +1,123 @@
...
+public interface HttpException extends QuietException
+{
+    int getCode();
+
+    String getReason();
+
+    static void throwAsUnchecked(HttpException httpException)
+    {
+        if (httpException instanceof Throwable throwable)
+            ExceptionUtil.ifExceptionThrowUnchecked(throwable);
```
sbordet: I'd rather fail a cast than silently ignore an implementation mistake here.
So perhaps just `ExceptionUtil.ifExceptionThrowUnchecked((Throwable)httpException);`, although I would prefer this method to return the exception, and the caller to throw (in this way the method can be reused also to store the returned exception in a field, etc.).

sbordet (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpException.java
gregw: @sbordet a HttpException is just a marker interface. Currently we apply it to `RuntimeException` and `IllegalArgumentException`, but in future we may tag an Error as a HttpException.  Unlikely, but having just made a whole PR about decoupling `HttpException` from the specific type of exception, it seams silly to bind it here.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpException.java
gregw: No idea. They were in the code to start with.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpException.java
gregw: ExceptionUtil definitely can't return the exception, as it may be an Error.  Reusing that method also makes this best as a throwing method.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpTester.java
gregw: Which we might wish to catch and rethrow as a HttpException variant.   The whole point of this PR is decouple `HttpException` from any specific exception class.  A `throwAsUncheck(failure)` method is a reasonable way to keep this decouple, even if we never actually use that for `Error`s

gregw:
I've no idea why this branch suffers from so many flaky tests.  But I can see nothing in the CI that is related to this change.

olamy:
` testDownloadWithCloseBeforeContent(Transport)[1] – org.eclipse.jetty.test.client.transport.HttpClientStreamTest` is already flaky in jetty-12.0.x for many builds https://jenkins.webtide.net/job/jetty.project/job/jetty-12.0.x/1092/

olamy:
issue created https://github.com/eclipse/jetty.project/issues/9485 
best to mark it as flaky using `@Tag("flaky")`

sbordet (approved):

PR #9477, opened 2023-03-08
Issue #9459 empty session path in cookie

### LINKED ISSUE #9459: Path is missing from JSESSIONID cookie in 12 beta 0
**Jetty version(s)**
12.0.0.beta0 / ee10

**Java version/vendor**

**OS type/version**
mac os

**Description**
One of the tests in Vaadin first requests

http://localhost:8888/web-component/themed-component.js

and then later on

http://localhost:8888/

Both of the requests use the session.

Now with Jetty 12 the response to the first request contains
```
Set-Cookie: JSESSIONID=node01esb8me0d0pxv1ecc7ivw1d4uh1.node0
```
causing Chrome to store the cookie with the path `/web-component`

It is thus not passed with the second request to http://localhost:8888/.

Now one of two things can happen, depending on some external circumstances like what is cached and network loading order
1. The request to / has no cookie, Vaadin expects it to have and an error is printed
2. The request to / creates a new session and sets a new session id cookie

If a new session cookie is set, then the following request to / includes two session cookies and Jetty directly returns "bad request"

With Jetty 11 and earlier the first request response has
```
Set-Cookie: JSESSIONID=node0j20fd8h8crx91052t9avzbax11.node0; Path=/
```
and thus no further session id cookies are set and no more sessions created

janbartel:
Closes #9459 

Commit  https://github.com/eclipse/jetty.project/commit/50a88187fa26bf558c092bd59fce67dcf28acd7e only set the session cookie path iff the context path was not the root context.

### DIFF
--- a/jetty-core/jetty-session/src/main/java/org/eclipse/jetty/session/AbstractSessionManager.java
+++ b/jetty-core/jetty-session/src/main/java/org/eclipse/jetty/session/AbstractSessionManager.java
@@ -239,9 +239,9 @@ public abstract class AbstractSessionManager extends ContainerLifeCycle implemen
         _context = ContextHandler.getCurrentContext();
         _loader = Thread.currentThread().getContextClassLoader();
 
-        // ensure a session path is set for non root contexts
+        // ensure a session path is set
         String contextPath = _context == null ? "/" : _context.getContextPath();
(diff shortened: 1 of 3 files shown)


gregw (approved):
ooops that was my bad! thanks

joakime (approved):

PR #9479, opened 2023-03-08
Jetty 12.0.x 9444 servlet paths fully decoded

gregw:
Alternative to #9465 
This is an alternative to #9462 & #9465 as a fix for #9444, replacing changes made by #9455


gregw:
@joakime do you think there are any tests from #9462 that should be kept even without safeDecode in this PR?

gregw:
@sbordet @janbartel @joakime this PR is now building and so is ready for some review, however I think there is still some more work that could be done:
 - [x] Need to merge the BadMessage/HttpException PR so we can correctly throw an IAE rather than a RTE
 - [x] We could potentially make throwing the IAE configurable, so an app could both allow the violation and allow the ambiguous decoding.  The question is, where should that configuration be?
 - [ ] probably more renaming could be done to distinguish between `encodedPathInContext` and `decodedPathInContext`.  Any ideas for better names for those?
 - [x] there are some related unit tests that are disabled. They should be re-enabled and debugged (edit: turned out they were cross context dispatch)

gregw:
How about we use `canonicalPathInContext` instead of `encodedPathInContext` and then just used `pathInContext` for the `decodedPathInContext`?

joakime:
> How about we use `canonicalPathInContext` instead of `encodedPathInContext` and then just used `pathInContext` for the `decodedPathInContext`?

I'd prefer that we keep the "encoded" and "decoded" portions in the method names.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
```diff
@@ -124,21 +127,15 @@ public String getDescription()
...
      * including allowing all additional ambiguous URI Violations.
      */
-    public static final UriCompliance RFC3986 = new UriCompliance("RFC3986", allOf(Violation.class));
-
-    /**
-     * Compliance mode that follows <a href="https://tools.ietf.org/html/rfc3986">RFC3986</a>
-     * plus it does not allow any ambiguous URI {@link Violation}s.
-     */
-    public static final UriCompliance RFC3986_UNAMBIGUOUS = new UriCompliance("RFC3986_UNAMBIGUOUS", noneOf(Violation.class));
+    public static final UriCompliance RFC3986 = new UriCompliance("RFC3986", noneOf(Violation.class));
```
sbordet: This seems a break in backwards compatibility behavior?
At least the javadoc needs to be changed?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Context.java
```diff
@@ -98,14 +98,30 @@ public interface Context extends Attributes, Decorator, Executor
...
+    {
+        return getPathInContext(getContextPath(), canonicallyEncodedPath);
+    }
 
     /**
      * @return a non-{@code null} temporary directory, configured either for the context, the server or the JVM
      */
     File getTempDirectory();
+
+    static String getPathInContext(String encodedContextPath, String canonicallyEncodedPath)
```
sbordet: Javadocs, especially what "canonically" means.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
```diff
@@ -280,6 +282,7 @@ public int getIntHeader(String name)
     @Override
     public String getPathInfo()
     {
+        checkForUriComplianceViolations();
```
sbordet: We don't want to do this check every time this method is called.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
```diff
@@ -364,9 +367,25 @@ public StringBuffer getRequestURL()
     @Override
     public String getServletPath()
     {
+        checkForUriComplianceViolations();
```
sbordet: Ditto.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletHandler.java
```diff
@@ -124,6 +125,21 @@ public ServletHandler()
     {
     }
 
+    @ManagedAttribute(value = "True if URIs with violations are decoded")
+    public boolean getDecodeAmbiguousUris()
```
sbordet: Should be renamed to `isDecodeAmbiguousURIs()`.

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/AsyncContextTest.java
```diff
@@ -71,7 +71,7 @@ public void setUp() throws Exception
         _contextHandler.addServlet(new ServletHolder(new TestServlet2()), "/servletPath2");
 
         ServletHolder testHolder = new ServletHolder(new TestServlet());
-        testHolder.setInitParameter("dispatchPath", "/test2/something%2felse");
+        testHolder.setInitParameter("dispatchPath", "/test2/something%25else");
```
sbordet: Ok to move from `%2F` to `%25`, but do we have `%2F` tests elsewhere now that these have been changed?
Now we throw if we find `%2F` (right?), but do we test that we actually throw?

sbordet (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
gregw: Yeah, I was a bit torn by this one.  Ultimately it seams strange to have something "RFC3986" contain all the violations to RFC3986.   I'll fix the javadoc and think about the names some more.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
gregw: What is the alternative?  The check is cheap if there are no violations, just an isEmpty call on an enumSet.

I guess we could check for the violations when wrapping and if they exist AND the configuration is the throw, then wrap with a different version of the API wrapper.   I'll try that.

joakime:
We'll get rid of the ISO-8859-1 fallback during decode in a different PR?

* #9489

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/AsyncContextTest.java
gregw: I've added another test for %2F, which currently dispatches with a request that returns the fully decoded pathInfo rather than throwing. I figure because it is coming from the app itself rather than over the wire, then we don't need to protect from ambiguous URIs because there is no security anyway.

joakime:
@gregw I added some more URIUtilTest cases, and a block of tests that are commented out for a future PR.
I'm onboard with this PR once the concerns from @sbordet are addressed.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
gregw: fixed javadoc

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
gregw: I now wrap differently, so the check is only done once.   But it is always done, rather than only if the methods are used.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Context.java
```diff
@@ -94,18 +94,47 @@ public interface Context extends Attributes, Decorator, Executor
...
-     *
-     * @param fullPath a full URI path
+     * <p>Returns the URI path scoped to this Context.</p>
+     * @see #getPathInContext(String, String)
+     * @param canonicallyEncodedPath a full URI path that should be canonically encoded as
+     *        per {@link org.eclipse.jetty.util.URIUtil#canonicalPath(String)}
      * @return the URI path scoped to this Context, or {@code null} if the full path does not match this Context.
-     * The empty string is returned if the full path is exactly the context path.
+     *         The empty string is returned if the full path is exactly the context path.
+     *
```
sbordet: Remove empty line.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
```diff
@@ -1246,4 +1247,24 @@ public Map<String, String> getTrailerFields()
         }
         return trailersMap;
     }
+
+    public static class AmbiguousURI extends ServletApiRequest
```
sbordet: Make it `private`.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextRequest.java
```diff
@@ -78,7 +79,7 @@ public static ServletContextRequest getServletContextRequest(ServletRequest requ
     private final ServletContextResponse _response;
     final ServletHandler.MappedServlet _mappedServlet;
     private final HttpInput _httpInput;
-    private final String _pathInContext;
+    private final String _decodedPathInContext;
```
sbordet, line 79: Not for this line, but the `LOG` instance on line 49 is never used, remove it?

sbordet (approved):
Just nits, LGTM.

joakime (approved):

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
gregw: needs to be package protected at least

PR #9481, opened 2023-03-08
Jetty 12 - Dependency Updates and Report Script

joakime:
Introduce a `build/scripts/dependency-update-reports.sh` to produce a `reports/dependency-update-reports/` output of listed dependencies and if they have a version update.

* jetty-core
* jetty-ee10
* jetty-ee9
* jetty-ee8
* root

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,6 @@ bin/
 # test generated content
 */src/test/*/WEB-INF/lib/test*.jar
 .flattened-pom.xml
+
+# reports
+reports/
--- a/build/scripts/dependency-update-reports.sh
+++ b/build/scripts/dependency-update-reports.sh
@@ -19,16 +19,24 @@ mvn -N -Pdependency-updates-reports validate
 cp -Rv target/site/* $REPORT_OUTPUT_DIR
 mv $REPORT_OUTPUT_DIR/dependency-updates-aggregate-report.html $REPORT_OUTPUT_DIR/dependency-updates-report-root.html
 
+pushd jetty-core
+mvn -Pdependency-updates-reports validate
+cp target/site/dependency-updates-aggregate-report.html $REPORT_OUTPUT_DIR/dependency-updates-report-core.html
+popd
+
 pushd jetty-ee10
 mvn -Pdependency-updates-reports validate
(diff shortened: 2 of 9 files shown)


olamy:
we could eventually use https://github.com/jenkinsci/htmlpublisher-plugin. to publish the html reports as part of the build.

gregw:
Couldn't we just use this instead of dependabot?

olamy:
@gregw this is just reports and do not create automatically PRs with update.

gregw:
@olamy I totally understand that. I see very little value in the automatically created PRs, which frequently break things and put silly stress on the CI environment.
We could run this report once a month, select the updates we want to do and create a single PR per branch with the updates we want.    I think it would be far less work than processing many PRs.... but then I'm not the one doing that work.

joakime:
> I think it would be far less work than processing many PRs.... but then I'm not the one doing that work.

Not really.  The technique in this PR will consume about 4x more time than the dependabot technique.
The time (on this PR) that it took to update root and core and fix a upper bounds dep in ee9 (which is all that's updated so far) is about 2 solid hours of dedicated and focused time.
I expect the rest (ee10 + ee9 + ee8) to consume the rest of my day. (and that's me not upgrading infinispan or weld, which is a totally different level of effort)
2 hours of dependabot time would have handled all of jetty-9 + jetty-10 + jetty-11 already.
Keep in mind that time spent with dependabot is mostly in the little gaps between other tasks, not dedicated and focused like this technique.

joakime:
Dependabot remembers decisions too.  

Eg: If we don't want to upgrade infinispan from 11.x to 14.x, we tell dependabot to ignore the major version.
We'll wind up dependabot updates within 11.x instead.

Something that with this technique becomes knowledge kept by an individual, and possible custom rules/configuration in a pom to establish.

joakime:
@olamy I can move these `<ruleSet>` configurations to `build/build-resource/` and use them via classpath.
WDYT?  Is it worth doing?

joakime:
I have not upgraded infinispan or weld, I feel those will need to be separate PRs.

* #9487 
* #9488

olamy (approved):

olamy:
we might need to improve the rendering. the default skin is so 90's 🤣

PR #9492, opened 2023-03-13
Jetty 12.0.x improve multiversion annotations test

janbartel:
The tests for multiversion jar annotation parsing really weren't doing much testing, so this PR ensures some testing is actually done and verified. Also minor fix in AnnotationParser to use the normalized class name, not the internal class name.


janbartel:
The failing test on this pr is nothing to do with this pr - it's some other flakey test.

joakime (approved):

PR #9494, opened 2023-03-13
Improved HttpClient TLS documentation about server host name verifica…

sbordet:
…tion.


boekhold (approved):
This documentation update makes it crystal-clear what the effect is of disabling the Endpoint Identification Algorithm. The update removes the previous ambiguity on whether certificate expiration date and CA certificate checking are affected as well.

PR #9496, opened 2023-03-13
Fixes #94890 - replace ISO-8859-1 fallback with UTF-8 Replacement Character

### LINKED ISSUE #9489: Jetty 12 - Remove ISO-8859-1 fallback during URIUtil.decodePath(String)
**Jetty version(s)**
Jetty 12

Java 17

**OS type/version**
All

**Description**
The URIUtil.decodePath(String) will start out decoding using UTF-8, then fallback to decoding in ISO-8859-1 if there is problem with decoding the UTF-8 sequences.

Bad UTF-8 sequences should be replaced with the Unicode Replacement Character `�`.

joakime:
Fixes #9489 

+ Removes ISO-8859-1 fallback behavior from URIUtil.decodePath()
+ Introduces Replacement Character behaviors to Utf8Appendable
+ New constructors for Utf8StringBuilder and Utf8StringBuffer to allow specifying malformed behavior

### DIFF
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
@@ -71,6 +71,8 @@ public final class UriCompliance implements ComplianceViolation.Mode
          */
         UTF16_ENCODINGS("https://www.w3.org/International/iri-edit/draft-duerst-iri.html#anchor29", "UTF16 encoding");
 
+        // TODO: BAD_UTF8_ENCODINGS
+
         private final String _url;
         private final String _description;
 
--- a/jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/InvalidURIRule.java
+++ b/jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/InvalidURIRule.java
@@ -35,6 +35,7 @@ import org.slf4j.LoggerFactory;
 public class InvalidURIRule extends Rule
 {
     private static final Logger LOG = LoggerFactory.getLogger(InvalidURIRule.class);
+    private static final int REPLACEMENT_CHAR_CODEPOINT = 65533; // UTF-8 Replacement Char as codepoint
 
     private int _code = HttpStatus.BAD_REQUEST_400;
     private String _message = "Illegal URI";
@@ -118,6 +119,9 @@ public class InvalidURIRule extends Rule
 
         if (LOG.isDebugEnabled())
             LOG.debug("{} {} {} {}", Character.charCount(codepoint), codepoint, block, Character.isISOControl(codepoint));
+
+        if (codepoint == REPLACEMENT_CHAR_CODEPOINT)
+            return false;
 
         return (!Character.isISOControl(codepoint)) && block != null && !Character.UnicodeBlock.SPECIALS.equals(block);
     }
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
@@ -30,7 +30,6 @@ import java.util.Objects;
 import java.util.StringTokenizer;
 import java.util.stream.Stream;
 
-import org.eclipse.jetty.util.Utf8Appendable.NotUtf8Exception;
 import org.eclipse.jetty.util.resource.ResourceFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -552,12 +551,6 @@ public final class URIUtil
                 return path;
             return path.substring(offset, end);
         }
-        catch (NotUtf8Exception e)
-        {
-            if (LOG.isDebugEnabled())
-                LOG.debug("{} {}", path.substring(offset, offset + length), e.toString());
-            throw e;
-        }
         catch (IllegalArgumentException e)
         {
             throw e;
@@ -772,16 +765,6 @@ public final class URIUtil
 
             String canonical = (builder != null) ? builder.getString() : encodedPath;
             return normal ? canonical : normalizePath(canonical);
-        }
-        catch (NotUtf8Exception e)
-        {
-            if (LOG.isDebugEnabled())
-                LOG.debug("{} {}", encodedPath, e.toString());
-            throw e;
-        }
-        catch (IllegalArgumentException e)
-        {
-            throw e;
         }
         catch (Exception e)
         {
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
@@ -547,7 +547,7 @@ public final class URIUtil
             }
 
             if (builder != null)
-                return builder.takeFinishedString();
+                return builder.getString();
             if (offset == 0 && length == path.length())
                 return path;
             return path.substring(offset, end);
@@ -770,7 +770,7 @@ public final class URIUtil
                 slash = c == '/';
             }
 
-            String canonical = (builder != null) ? builder.takeFinishedString() : encodedPath;
+            String canonical = (builder != null) ? builder.getString() : encodedPath;
             return normal ? canonical : normalizePath(canonical);
         }
         catch (NotUtf8Exception e)
@@ -1624,7 +1624,7 @@ public final class URIUtil
             }
 
             if (builder != null)
-                return builder.takeFinishedString();
+                return builder.getString();
             return path;
         }
         catch (IllegalArgumentException e)
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
@@ -18,7 +18,6 @@ import java.io.IOException;
 import java.net.URI;
 import java.net.URL;
 import java.net.URLClassLoader;
-import java.nio.charset.CodingErrorAction;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -413,7 +412,7 @@ public final class URIUtil
                 ret.append(c);
             }
         }
-        return ret.toString();
+        return ret.getString();
     }
 
     /**
@@ -492,7 +491,7 @@ public final class URIUtil
                     case '%':
                         if (builder == null)
                         {
-                            builder = new Utf8StringBuilder(length, CodingErrorAction.REPLACE);
+                            builder = new Utf8StringBuilder(length);
                             builder.append(path, offset, i - offset);
                         }
                         if ((i + 2) < end)
@@ -525,7 +524,7 @@ public final class URIUtil
                     case ';':
                         if (builder == null)
                         {
-                            builder = new Utf8StringBuilder(path.length(), CodingErrorAction.REPLACE);
+                            builder = new Utf8StringBuilder(path.length());
                             builder.append(path, offset, i - offset);
                         }
 
@@ -548,7 +547,7 @@ public final class URIUtil
             }
 
             if (builder != null)
-                return builder.toString();
+                return builder.takeFinishedString();
             if (offset == 0 && length == path.length())
                 return path;
             return path.substring(offset, end);
@@ -686,7 +685,7 @@ public final class URIUtil
                     case '%':
                         if (builder == null)
                         {
-                            builder = new Utf8StringBuilder(encodedPath.length(), CodingErrorAction.REPLACE);
+                            builder = new Utf8StringBuilder(encodedPath.length());
                             builder.append(encodedPath, 0, i);
                         }
                         if ((i + 2) < end)
@@ -730,7 +729,7 @@ public final class URIUtil
                     case ';':
                         if (builder == null)
                         {
-                            builder = new Utf8StringBuilder(encodedPath.length(), CodingErrorAction.REPLACE);
+                            builder = new Utf8StringBuilder(encodedPath.length());
                             builder.append(encodedPath, 0, i);
                         }
 
@@ -759,7 +758,7 @@ public final class URIUtil
                     default:
                         if (builder == null && !isSafe(c))
                         {
-                            builder = new Utf8StringBuilder(encodedPath.length(), CodingErrorAction.REPLACE);
+                            builder = new Utf8StringBuilder(encodedPath.length());
                             builder.append(encodedPath, 0, i);
                         }
 
@@ -771,7 +770,7 @@ public final class URIUtil
                 slash = c == '/';
             }
 
-            String canonical = (builder != null) ? builder.toString() : encodedPath;
+            String canonical = (builder != null) ? builder.takeFinishedString() : encodedPath;
             return normal ? canonical : normalizePath(canonical);
         }
         catch (NotUtf8Exception e)
@@ -1625,7 +1624,7 @@ public final class URIUtil
             }
 
             if (builder != null)
-                return builder.toString();
+                return builder.takeFinishedString();
             return path;
         }
         catch (IllegalArgumentException e)
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
@@ -18,6 +18,7 @@ import java.io.IOException;
 import java.net.URI;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.nio.charset.CodingErrorAction;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -491,7 +492,7 @@ public final class URIUtil
                     case '%':
                         if (builder == null)
                         {
-                            builder = new Utf8StringBuilder(length);
+                            builder = new Utf8StringBuilder(length, CodingErrorAction.REPLACE);
                             builder.append(path, offset, i - offset);
                         }
                         if ((i + 2) < end)
@@ -524,7 +525,7 @@ public final class URIUtil
                     case ';':
                         if (builder == null)
                         {
-                            builder = new Utf8StringBuilder(path.length());
+                            builder = new Utf8StringBuilder(path.length(), CodingErrorAction.REPLACE);
                             builder.append(path, offset, i - offset);
                         }
 
@@ -556,7 +557,7 @@ public final class URIUtil
         {
             if (LOG.isDebugEnabled())
                 LOG.debug("{} {}", path.substring(offset, offset + length), e.toString());
-            return decodeISO88591Path(path, offset, length);
+            throw e;
         }
         catch (IllegalArgumentException e)
         {
@@ -685,7 +686,7 @@ public final class URIUtil
                     case '%':
                         if (builder == null)
                         {
-                            builder = new Utf8StringBuilder(encodedPath.length());
+                            builder = new Utf8StringBuilder(encodedPath.length(), CodingErrorAction.REPLACE);
                             builder.append(encodedPath, 0, i);
                         }
                         if ((i + 2) < end)
@@ -729,7 +730,7 @@ public final class URIUtil
                     case ';':
                         if (builder == null)
                         {
-                            builder = new Utf8StringBuilder(encodedPath.length());
+                            builder = new Utf8StringBuilder(encodedPath.length(), CodingErrorAction.REPLACE);
                             builder.append(encodedPath, 0, i);
                         }
 
@@ -758,7 +759,7 @@ public final class URIUtil
                     default:
                         if (builder == null && !isSafe(c))
                         {
-                            builder = new Utf8StringBuilder(encodedPath.length());
+                            builder = new Utf8StringBuilder(encodedPath.length(), CodingErrorAction.REPLACE);
                             builder.append(encodedPath, 0, i);
                         }
 
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
@@ -115,6 +115,7 @@ public abstract class Utf8Appendable implements CharsetStringBuilder
 
     protected void reset()
     {
+        _codep = 0;
         _state = UTF8_ACCEPT;
     }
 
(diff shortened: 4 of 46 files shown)


joakime:
@gregw I added a final commit (d838d3302b1f5c51658fa9d7f296d53213d79b83) to also make `UrlEncoded` use REPLACE behaviors, but I'm on the fence on if this is a good idea (or not).

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
```diff
@@ -80,28 +96,42 @@ public abstract class Utf8Appendable implements CharsetStringBuilder
         12, 36, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12
     };
 
+    private final CodingErrorAction _coderErrorAction;
     private int _codep;
+    private boolean _hasReplacements = false;
```
joakime, line 101: Still need to key a new `UriCompliance` mode off this.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
```diff
@@ -272,21 +337,29 @@ public NotUtf8Exception(String reason)
...
+                }
+                catch (IOException e)
+                {
+                    throw new RuntimeException(e);
+                }
+                _hasReplacements = true;
             }
-            throw new NotUtf8Exception("incomplete UTF8 sequence");
+            _codep = 0;
+            _state = UTF8_ACCEPT;
```
gregw: Put these back before the append, just in case it fails for some reason it will leave appender in better state

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
```diff
@@ -272,21 +337,29 @@ public NotUtf8Exception(String reason)
...
+     * Check the state of the Appendable to know if the current UTF-8 sequence is complete (or not)
+     */
     public void checkState()
     {
         if (!isUtf8SequenceComplete())
         {
-            _codep = 0;
-            _state = UTF8_ACCEPT;
-            try
+            if (_coderErrorAction == CodingErrorAction.REPLACE)
```
gregw: Maybe a switch statement rather than multiple ifs

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -491,7 +492,7 @@ public static String decodePath(String path, int offset, int length)
                     case '%':
                         if (builder == null)
                         {
-                            builder = new Utf8StringBuilder(length);
+                            builder = new Utf8StringBuilder(length, CodingErrorAction.REPLACE);
```
gregw: So my concern with this PR is the hard coding or REPLACE in many places.
I know that previously we had an implicit "hard coding" of REPORT (aka throw), so we are just as hard coded... but we are changing the behaviour.  I need to ponder this for while longer.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -556,7 +557,7 @@ public static String decodePath(String path, int offset, int length)
         {
             if (LOG.isDebugEnabled())
                 LOG.debug("{} {}", path.substring(offset, offset + length), e.toString());
-            return decodeISO88591Path(path, offset, length);
+            throw e;
```
gregw: The NotUtf8Exception should never be thrown now because we hard code REPLACE?

gregw: We could just remove the ISO8891 fallback here and not change the Utf8Builder..... ah but whilst if feels like we were previously hard coded to throw, this catch made us not really throw, but instead REPLACE_WITH_8859_1, which is not standard and silly anyway.   So this is not such a big change after all.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/URIUtil.java
```diff
@@ -758,7 +759,7 @@ public static String canonicalPath(String encodedPath)
                     default:
                         if (builder == null && !isSafe(c))
                         {
-                            builder = new Utf8StringBuilder(encodedPath.length());
+                            builder = new Utf8StringBuilder(encodedPath.length(), CodingErrorAction.REPLACE);
```
gregw: Interesting that we were not falling back to 8859-1 in this method, so we are somewhat inconsistent.    As we pretty much always call canonicalPath, then we pretty much always threw.   Now we will replace. Hmm back to being cautious about a big behaviour change.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
gregw, line 1: I think an alternate/better way to change this class would be to:
 + not pass in CodingErrorAction to the constructor.
 + not check on every byte addition if we should ignore/replace/report.   Instead always add replacement (as we did before) and set boolean that we have replacement.
 + move the checking to the methods that take the string (`toSting`, `toPartialString`, `takeString`, etc.).    These methods could continue to have their default no args version which would implement whatever default we want (probably report to keep behaviour the same).
 + add new `String toString(CodingErrorAction cea)`, `String toPartialString(CodingErrorAction cea)`, etc. methods that implement the action.  If no replacement characters, they just return string. If there are replacements, they either throw, ignore or strip them depending on the action.

I think this has less performance impact on the tight loops of parsing characters and moves the choice of action to one spot where we take the string, which I think will be easier to include into any future compliance modes.

gregw (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
joakime: Can't switch on CodingErrorAction.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
```diff
@@ -327,40 +331,159 @@ public boolean isUtf8SequenceComplete()
...
+     *     This will get the String as it exists currently, without including any trailing incomplete UTF-8 sequences.
+     * </p>
+     * <p>
+     *     Use {@link #finish()} to complete any incomplete UTF-8 sequences before calling this method.
+     * </p>
+     *
+     * @return the String as it exists currently, without throwing an Exception
+     * @see #finish()
+     */
+    public String getString()
```
gregw, line 427: I prefer to keep the methods as `toString()`, `toString(boolean)`, `takeString()` and `takeString(boolean)`.

I understand that many other classes have an informative `toString()` and use something like `asString()` to get the value.   However, this class is meant to be a drop in replacement to `StringBuilder` and `StringBuffer` and they use `toString()` to access the built value. 

You can see by the number of changed files in this PR how changing to `getString()` makes this a much larger change.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
```diff
@@ -327,40 +331,159 @@ public boolean isUtf8SequenceComplete()
...
+        return str;
+    }
+
+    /**
+     * Take the complete UTF-8 string and reset the internal buffer.
+     *
+     * @param throwOnReplacement throw {@link NotUtf8Exception} if input had characters replaced.
+     * @return the UTF-8 so far, including any final incomplete UTF-8 sequences.
+     */
+    public String takeFinishedString(boolean throwOnReplacement)
```
gregw: I can't see the difference between `takeString()` (which calls `finish()`) and `takeFinishedString()`???

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
```diff
@@ -327,40 +331,159 @@ public boolean isUtf8SequenceComplete()
...
+     *
+     * <p>
+     * Calls to this method will {@link #reset()} this {@link Utf8Appendable}.
+     * </p>
+     *
+     * @return the String buffer from the appendable
+     * @throws CharacterCodingException if unable to encode the input bytes (eg: bad UTF-8 sequences)
+     */
+    @Override
+    public String takeString() throws CharacterCodingException
```
gregw, line 445: Do we need to introduce CharacterCodingException?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
```diff
@@ -327,40 +331,159 @@ public boolean isUtf8SequenceComplete()
...
     /**
-     * Not a valid sequences of UTF-8 bytes
+     * Finish the buffer state.
+     * <p>
+     *     This will address any incomplete utf-8 sequences in the buffer, and set Utf8Appendable in a state
+     *     where it will be ready to accept new utf-8 sequences.
+     * </p>
      */
-    public static class NotUtf8Exception extends IllegalArgumentException
+    public void finish()
```
gregw, line 353: this can probably be private now (I like the name better)

gregw (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
joakime: `takeString()` is part of the `CharsetStringBuilder` interface that `Utf8Appendable` implements (and is used in many places).
It must throw a checked `CharacterCodingException` (which is the opposite of what we want).

The original `toString()` was changing the state of the internals of the class (to satisfy unfinished sequences), which broke debugging (it was impossible to debug a `Utf8Appendable` implementation while the debugger itself calls `.toString()` to show things, while the state internally changes), and if you enabled logging you broke the behavior of an active `Utf8Appendable` too for the same reason.   
That nuance, the changing of state during `toString()` was the cause of several other bugs as well.

There's 4 read access type demands on an `Utf8Appendable` implementation.

1. The entire String, with unfinished codepoints/sequences resolved. (optional boolean for throwing)
2. The entire String, without codepoints/sequences resolved. (optional boolean for throwing)
3. The String as it sits now, and then empty/clear the buffer, but not the unfinished codepoints/sequences. (optional boolean for throwing)
4. The `CharsetStringBuilder` in UTF-8 mode (always throws `CharacterCodingException`)

The most common use case is 1, which mimics the StringBuilder and StringBuffer behavior. (approx 75% of cases)
Use case 2 and 3 are very similar IMO.
Use case 2 is for when you want to access the content before it's finished. (few uses, and I think we can drop this as a use case in favor of just using other methods)
Use case 3 is the predominant WebSocket use case.
Use case 4 is used by server FormFields and IO ContentSource implementations.

For Use case 1, the behavior is like this ...
``` java
Utf8StringBuilder buffer = new Utf8StringBuilder();
buffer.append(bytes);
buffer.finish();
String result = buffer.toString();
```

For Use case 2, the behavior is like this ...
``` java
Utf8StringBuilder buffer = new Utf8StringBuilder();
buffer.append(bytes);
String partialResults = buffer.toString();
buffer.append(bytes);
buffer.finish();
String results = buffer.toString();
```

For Use case 3, the behavior is like this ...
``` java
Utf8StringBuilder buffer = new Utf8StringBuilder();
buffer.append(bytes);
String result = buffer.toString();
buffer.clear();
```

For Use case 4, the behavior is ...
``` java
CharsetStringBuilder text = CharsetStringBuilder.forCharset(StandardCharsets.UTF_8); // uses Utf8StringBuilder
text.append(bytes);
String result = text.takeString();
```

Whatever solution we settle on we should not have `toString()` cause internal state changes, and still support the `CharsetStringBuilder` concepts somehow (maybe not all the way down in `Utf8Appendable`).

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
joakime: This is the signature of the `CharsetStringBuilder` interface, that Utf8Appendable always implemented.

gregw: Hmmm i wonder if we can change our NotUtf8Exception to extend CharacterEncodingException and thus avoid the catch and rethrow... or is one checked?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
gregw: Good point about debugging. Hmmm no easy solutions. Ok let's go with getString and takeString.

But why takeFinishedString?

gregw: Is that instead of takePartialString?

I prefer having partial string methods and the getString ones will always be finished strings

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
gregw: Dang it is checked. ignore.

gregw:
@joakime Just to summarise my various comments and your responses and my updated comments...

I think we should have methods:
 + `getString()` and `getString(boolean)` that return the finished string (but the `finish()` method should not be private or protected)
 + `takeString()` and `takeString(boolean)` that are equivalent to a get then clear, but can be more efficient.
 + `getPartialString()` and `getPartialString(boolean)` that return the unfinished string. Note the later can still throw due to earlier utf8 issues.
 + `takePartialString()` and `takePartialString(boolean)` that are equivalent to a getPartial then clear, but can be more efficient.
 
The take methods should throw `CharacterCodingException` to match the signature of CharsetStringBuilder
The get methods should not throw a checked exception, but declare that they can throw our `NotUtf8Exception`

joakime:
How about this ...

``` java
// Get the String
// @param finishCodePoints true to finish any active codepoints/sequences, false to leave codepoints/sequences alone
// @param throwOnReplacement true to throw NotUf8Exception if creation of string caused replacement characters
String .getString(boolean finishCodePoints, boolean throwOnReplacement); 

String .getString(boolean throwOnReplacement) {
    return getString(true, throwOnReplacement);
}

String .getString() {
  return getString(true, false);
}
```

For use case 1, whole string, with replacement chars, no throw, it's just 
``` java
Utf8StringBuilder builder = new Utf8StringBuilder();
builder.append(bytes);
String result = builder.getString();
```

For use case 2, partial, then final it's ...
``` java
Utf8StringBuilder builder = new Utf8StringBuilder();
builder.append(bytes);
String partial = builder.getString(false, false);
builder.append(bytes);
String complete = builder.getString();
```

For use case 3, partial, with clear of buffer, keep codepoints, throw on replacement, it's ...
``` java
Utf8StringBuilder builder = new Utf8StringBuilder();
builder.append(bytes);
String partial = builder.getString(false, true);
builder.clear();
builder.append(bytes);
String final = builder.getString(true, true);
```

joakime:
Heck, could even simplify this by borrowing a feature from java.util.Optional ...

``` java
// Get the String
// @param finishCodePoints true to finish any active codepoints/sequences, false to leave codepoints/sequences alone
// @param exception if non-null, the supplier will be used generate a throwable if there were characters replaced during the creation of this Utf8Appendable
String .getString(boolean finishCodePoints, Supplier<? extends Throwable> exceptionSupplier); 

String .getString() {
  return getString(true, null);
}
```

For use case 1, whole string, with replacement chars, no throw, it's just 
``` java
Utf8StringBuilder builder = new Utf8StringBuilder();
builder.append(bytes);
String result = builder.getString();
```

For use case 2, partial, then final it's ...
``` java
Utf8StringBuilder builder = new Utf8StringBuilder();
builder.append(bytes);
String partial = builder.getString(false, null);
builder.append(bytes);
String complete = builder.getString();
```

For use case 3, partial, with clear of buffer, keep codepoints, throw on replacement, it's ...
``` java
Provider<? extends Throwable> badWebSocketFrameSupplier = () -> new BadWebSocketTextFrame("Invalid UTF-8");
Utf8StringBuilder builder = new Utf8StringBuilder();
builder.append(bytes);
String partial = builder.getString(false, badWebSocketFrameSupplier);
builder.clear();
builder.append(bytes);
String final = builder.getString(true, badWebSocketFrameSupplier);
```

We should also then be able to make the existing `takeString()` look something like this ...
``` java
String takeString() throws CharacterCodingException
{
    try {
      return getString(true, () -> new CharacterCodingException());
    } finally {
      reset();
    }
}
```

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
```diff
@@ -71,6 +71,8 @@ public enum Violation implements ComplianceViolation
          */
         UTF16_ENCODINGS("https://www.w3.org/International/iri-edit/draft-duerst-iri.html#anchor29", "UTF16 encoding");
 
+        // TODO: BAD_UTF8_ENCODINGS
```
joakime, line 74: This still needs to happen

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/InvalidURIRule.java
```diff
@@ -119,6 +120,9 @@ protected boolean isValidChar(int codepoint)
         if (LOG.isDebugEnabled())
             LOG.debug("{} {} {} {}", Character.charCount(codepoint), codepoint, block, Character.isISOControl(codepoint));
 
+        if (codepoint == REPLACEMENT_CHAR_CODEPOINT)
```
joakime, line 123: If the client sends a replacement char, then what?

Eg:

```
GET /foo/%EF%BF%BD HTTP/1.1
Host: local
Connection: close
```

gregw: Then the decoded path will include the replacement character, but the boolean _hasReplacement will be false.

gregw:
hhhmmmmmmm maybe:

```java
String getString(boolean replaceBadCoding, boolean finishCodePoints); 

String getString(boolean replaceBadCoding) 
{
    return getString(replaceBadCoding, true);
}

String .getString() 
{
  return getString(true, true);
}
```

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
joakime: I gave this a whirl.

The websocket ValidationExtension still needs this. (to avoid a String copy).
Once we have a solution for that, then we can go `protected` for this field, as the Utf8StringBuilder and Utf8StringBuffer still use it.

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/InvalidURIRule.java
joakime: We don't have access to the Utf8StringBuilder or Utf8Appendable at this point to ask the .hasReplacement() method.
We have the post-decoded URI.

joakime:
> hhhmmmmmmm maybe:
> 
> ```java
> String getString(boolean replaceBadCoding, boolean finishCodePoints); 
> 
> String getString(boolean replaceBadCoding) 
> {
>     return getString(replaceBadCoding, true);
> }
> 
> String .getString() 
> {
>   return getString(true, true);
> }
> ```

The double boolean technique isn't being used anymore.
See current codebase on this PR.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
gregw: resolving so I can see current state clearly.

jetty-core/jetty-rewrite/src/main/java/org/eclipse/jetty/rewrite/handler/InvalidURIRule.java
gregw: Oh, sorry, I did not see that we are in a Rule.
What is this Rule for? Why is replacement character considered invalid?   We should consider a URI invalid if when we parsed it it had bad UTF8 code points, not because it has replacement characters.

Unless a replacement character is considered invalid regardless of where it came from

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
```diff
@@ -243,94 +302,187 @@ public void appendByte(byte b) throws IOException
...
+     * Get the String.
+     *
+     * @param completeCodePoints true to complete any active codepoints/sequences before returning the String
+     * @param throwableOnReplacementsSupplier if a replacement of a character occurred during append,
+     * and if this supplier is non-null, then use this supplier to throw an error
+     * @return the String
+     * @throws X if replacement of bad characters occurred, and {@code throwableOnReplacementsSupplier} was provided
      */
-    public abstract String getPartialString();
+    public <X extends Throwable> String getString(boolean completeCodePoints, Supplier<? extends X> throwableOnReplacementsSupplier) throws X
```
gregw, line 395: I don't love this API.  We already have some methods throwing `CharacterCodingException` and we have a convenience supplier for `NotUtf8Exception`.  Do we really want to open this class up to throw arbitrary other exceptions from this method?

If feels way too flexible for no good reason, without actually being flexible enough... e.g.  why a `Supplier`?  Maybe it should be a `Function` taking the bad code point and returning an `Exception`?

I'm back to worrying that we are making too big a change to this core class.   Let me ponder on this for a little while....

joakime: This is the same signature as used in Optional.
We now have things throwing `BadPayloadException` in websocket, where it used to throw + catch + wrap + throw.
And we have `CharacterCodingException` flows as well.
This also eliminated a few other throw NotUtf8Exception + catch + wrap + throw different Exception scenarios.

joakime: We don't know the bad code point at the time of `.getString()`, so a Function has no meaning.
If we start to track them (which I don't like as an idea), then we would also possible have more than 1 bad code point, then what?

joakime:
Closing in favor of #9507

Issue #980, opened 2016-10-05
HttpClient should report HTTP/2 INADEQUATE_SECURITY_ERROR better


joakime:
Currently, when using HttpClient with HTTP/2 and attempting to connect to an HTTP/2 server which sends a GO_AWAY / INADEQUATE_SECURITY_ERROR the HttpClient fails with a simple stacktrace showing ...

```
Caused by: java.util.concurrent.ExecutionException: java.nio.channels.ClosedChannelException
    at org.eclipse.jetty.client.util.FutureResponseListener.getResult(FutureResponseListener.java:118)
    at org.eclipse.jetty.client.util.FutureResponseListener.get(FutureResponseListener.java:101)
    at org.eclipse.jetty.client.HttpRequest.send(HttpRequest.java:652)
    at my code
    ... 34 more
Caused by: java.nio.channels.ClosedChannelException
    at org.eclipse.jetty.io.WriteFlusher.onClose(WriteFlusher.java:498)
    at org.eclipse.jetty.io.ssl.SslConnection$DecryptedEndPoint.onIncompleteFlush(SslConnection.java:409)
    at org.eclipse.jetty.io.AbstractEndPoint$2.onIncompleteFlush(AbstractEndPoint.java:54)
    at org.eclipse.jetty.io.WriteFlusher.write(WriteFlusher.java:322)
    at org.eclipse.jetty.io.AbstractEndPoint.write(AbstractEndPoint.java:140)
    at org.eclipse.jetty.http2.HTTP2Flusher.process(HTTP2Flusher.java:243)
    at org.eclipse.jetty.util.IteratingCallback.processing(IteratingCallback.java:241)
    at org.eclipse.jetty.util.IteratingCallback.succeeded(IteratingCallback.java:365)
    at org.eclipse.jetty.http2.HTTP2Flusher.succeeded(HTTP2Flusher.java:258)
    at org.eclipse.jetty.io.WriteFlusher$PendingState.complete(WriteFlusher.java:269)
    at org.eclipse.jetty.io.WriteFlusher.completeWrite(WriteFlusher.java:394)
    at org.eclipse.jetty.io.ssl.SslConnection$1.run(SslConnection.java:101)
    at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:671)
    at org.eclipse.jetty.util.thread.QueuedThreadPool$2.run(QueuedThreadPool.java:589)
    ... 1 more
```

[...]

PR #9498, opened 2023-03-14
Fixes #9237 - Decouple QTP idleTimeout from pool shrink rate.

### LINKED ISSUE #9237: Decouple QTP `idleTimeout` from pool shrink rate

sbordet:
Introduced `QueuedThreadPool.idleTimeoutMaxShrinkCount` to be the number of idle threads that are exited in one idle timeout.

When set to 1 (the default), the old behavior is reproduced: expiring 1 thread every idle timeout. When set to larger values, allows to keep around the threads for the idle timeout (in case of further load spikes), but allows to quickly recover OS memory when they are truly idle.

For example, with 2000 threads, 30 seconds idle timeout and idleTimeoutMaxShrinkCount=1, it will take 995 minutes (about 16.5 hrs) to shrink the pool back to 10 threads. By setting idleTimeoutMaxShrinkCount=100, the thread pool can be shrunk to 10 threads in about 10 minutes.

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
@@ -864,7 +864,7 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
                 LOG.debug("Starting {}", thread);
             _threads.add(thread);
             // Update the shrink threshold to prevent thrashing
-            _shrinkThreshold.set(NanoTime.now());
+            _shrinkThreshold.set(NanoTime.now() + TimeUnit.MILLISECONDS.toNanos(_idleTimeout));
             thread.start();
             started = true;
         }
@@ -1092,7 +1092,7 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
         int idle = Math.max(0, AtomicBiInteger.getLo(count));
         int queue = getQueueSize();
 
-        return String.format("%s[%s]@%x{%s,%d<=%d<=%d,i=%d,r=%d,q=%d}[%s]",
+        return String.format("%s[%s]@%x{%s,%d<=%d<=%d,i=%d,r=%d,t=%d,q=%d}[%s]",
             getClass().getSimpleName(),
             _name,
             hashCode(),
@@ -1102,6 +1102,7 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
             getMaxThreads(),
             idle,
             getReservedThreads(),
+            NanoTime.millisUntil(_shrinkThreshold.get()),
             queue,
             _tryExecutor);
     }
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
@@ -1060,7 +1060,7 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
                 exit: while (true)
                 {
                     // If we had a job,
-                    if (job != null)
+                    if (!idle)
                     {
                         // signal that we are idle again,
                         idle = true;
@@ -1077,59 +1077,66 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
                     {
                         long idleTimeoutNanos = TimeUnit.MILLISECONDS.toNanos(getIdleTimeout());
                         job = idleJobPoll(idleTimeoutNanos);
-                        if (job == null)
+                        while (job != null)
                         {
-                            // No job available
-                            int threads = getThreads();
-                            if (idleTimeoutNanos > 0 && threads > getMinThreads())
+                            idle = false;
+                            // Run the jobs.
+                            if (LOG.isDebugEnabled())
+                                LOG.debug("run {} in {}", job, QueuedThreadPool.this);
+                            runJob(job);
+                            if (LOG.isDebugEnabled())
+                                LOG.debug("ran {} in {}", job, QueuedThreadPool.this);
+
+                            // signal that we are idle again,
+                            idle = true;
+                            if (!addCounts(0, 1))
+                                break;
+
+                            // Look for another job
+                            job = _jobs.poll();
+                        }
+
+                        // No jobs available, should we shrink?
+                        int threads = getThreads();
+                        if (threads > getMinThreads())
+                        {
+                            // We have an idle timeout and excess threads, so check if we should shrink?
+                            long now = NanoTime.now();
+                            long shrinkPeriod = idleTimeoutNanos / getIdleTimeoutMaxShrinkCount();
+                            if (LOG.isDebugEnabled())
+                                LOG.debug("Shrink check, {} > {} period={}ms {}", threads, getMinThreads(), TimeUnit.NANOSECONDS.toMillis(shrinkPeriod), QueuedThreadPool.this);
+                            while (true)
                             {
-                                // We have an idle timeout and excess threads, so check if we should shrink?
-                                long now = NanoTime.now();
-                                long shrinkPeriod = idleTimeoutNanos / getIdleTimeoutMaxShrinkCount();
-                                if (LOG.isDebugEnabled())
-                                    LOG.debug("Shrink check, {} > {} period={}ms {}", threads, getMinThreads(), TimeUnit.NANOSECONDS.toMillis(shrinkPeriod), QueuedThreadPool.this);
-                                while (true)
+                                long shrinkThreshold = _shrinkThreshold.get();
+                                long threshold = shrinkThreshold;
+
+                                // If the threshold is too far in the past,
+                                // advance it to be one idle timeout before now plus a bit of shrink period.
+                                if (NanoTime.elapsed(threshold, now) > idleTimeoutNanos)
+                                    threshold = now - idleTimeoutNanos;
+
+                                // advance the threshold by one shrink period
+                                threshold += shrinkPeriod;
+
+                                // Is the new threshold in the future?
+                                if (NanoTime.isBefore(now, threshold))
                                 {
-                                    long shrinkThreshold = _shrinkThreshold.get();
-                                    long threshold = shrinkThreshold;
+                                    // Yes - we cannot shrink yet, so break and continue looking for jobs
+                                    if (LOG.isDebugEnabled())
+                                        LOG.debug("Shrink skipped, threshold={}ms in future {}", NanoTime.millisElapsed(now, threshold), QueuedThreadPool.this);
+                                    break;
+                                }
 
-                                    // If the threshold is too far in the past,
-                                    // advance it to be one idle timeout before now plus a bit of shrink period.
-                                    if (NanoTime.elapsed(threshold, now) > idleTimeoutNanos)
-                                        threshold = now - idleTimeoutNanos;
-
-                                    // advance the threshold by one shrink period
-                                    threshold += shrinkPeriod;
-
-                                    // Is the new threshold in the future?
-                                    if (NanoTime.isBefore(now, threshold))
-                                    {
-                                        // Yes - we cannot shrink yet, so break and continue looking for jobs
-                                        if (LOG.isDebugEnabled())
-                                            LOG.debug("Shrink skipped, threshold={}ms in future {}", NanoTime.millisElapsed(now, threshold), QueuedThreadPool.this);
-                                        break;
-                                    }
-
-                                    // We can shrink if we can update the atomic shrink threshold?
-                                    if (_shrinkThreshold.compareAndSet(shrinkThreshold, threshold))
-                                    {
-                                        // Yes - we have shrunk, so exit.
-                                        if (LOG.isDebugEnabled())
-                                            LOG.debug("SHRUNK, threshold={}ms in past {}", NanoTime.millisElapsed(threshold, now), QueuedThreadPool.this);
-                                        break exit;
-                                    }
+                                // We can shrink if we can update the atomic shrink threshold?
+                                if (_shrinkThreshold.compareAndSet(shrinkThreshold, threshold))
+                                {
+                                    // Yes - we have shrunk, so exit.
+                                    if (LOG.isDebugEnabled())
+                                        LOG.debug("SHRUNK, threshold={}ms in past {}", NanoTime.millisElapsed(threshold, now), QueuedThreadPool.this);
+                                    break exit;
                                 }
                             }
-                            continue;
                         }
-                        idle = false;
-
-                        // Run the job.
-                        if (LOG.isDebugEnabled())
-                            LOG.debug("run {} in {}", job, QueuedThreadPool.this);
-                        runJob(job);
-                        if (LOG.isDebugEnabled())
-                            LOG.debug("ran {} in {}", job, QueuedThreadPool.this);
                     }
                     catch (InterruptedException e)
                     {
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
@@ -963,7 +963,19 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
      */
     protected void runJob(Runnable job)
     {
-        job.run();
+        try
+        {
+            job.run();
+        }
+        catch (Throwable e)
+        {
+            LOG.warn("Job failed", e);
+        }
+        finally
+        {
+            // Clear any thread interrupted status.
+            Thread.interrupted();
+        }
     }
 
     /**
@@ -1057,22 +1069,8 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
             try
             {
                 Runnable job = null;
-                exit: while (true)
+                exit: while (_counts.getHi() > Integer.MIN_VALUE)
                 {
-                    // If we had a job,
-                    if (!idle)
-                    {
-                        // signal that we are idle again,
-                        idle = true;
-                        if (!addCounts(0, 1))
-                            break;
-                    }
-                    // else check we are still running.
-                    else if (_counts.getHi() == Integer.MIN_VALUE)
-                    {
-                        break;
-                    }
-
                     try
                     {
                         long idleTimeoutNanos = TimeUnit.MILLISECONDS.toNanos(getIdleTimeout());
@@ -1088,9 +1086,9 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
                                 LOG.debug("ran {} in {}", job, QueuedThreadPool.this);
 
                             // signal that we are idle again,
-                            idle = true;
                             if (!addCounts(0, 1))
                                 break;
+                            idle = true;
 
                             // Look for another job
                             job = _jobs.poll();
@@ -1140,18 +1138,10 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
                     }
                     catch (InterruptedException e)
                     {
+                        boolean interrupted = Thread.interrupted();
                         if (LOG.isDebugEnabled())
-                            LOG.debug("interrupted {} in {}", job, QueuedThreadPool.this);
+                            LOG.debug("interrupted {} {} in {}", interrupted, job, QueuedThreadPool.this);
                         LOG.trace("IGNORED", e);
-                    }
-                    catch (Throwable e)
-                    {
-                        LOG.warn("Job failed", e);
-                    }
-                    finally
-                    {
-                        // Clear any thread interrupted status.
-                        Thread.interrupted();
                     }
                 }
             }
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
@@ -968,6 +968,65 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
         job.run();
     }
 
+    /**
+     * <p>Attempts to shrink the thread pool by one thread if {@link #getThreads()} is greater than {@link #getMinThreads()}.</p>
+     * @return true if shrunk, false otherwise.
+     */
+    protected boolean shrinkIfNeeded()
+    {
+        long idleTimeoutNanos = TimeUnit.MILLISECONDS.toNanos(getIdleTimeout());
+        while (true)
+        {
+            // No jobs available, should we shrink?
+            int threads = getThreads();
+            int minThreads = getMinThreads();
+            if (threads > minThreads)
+            {
+                // We have an idle timeout and excess threads, so check if we should shrink?
+                long now = NanoTime.now();
+                long shrinkPeriod = idleTimeoutNanos / getIdleTimeoutMaxShrinkCount();
+                if (LOG.isDebugEnabled())
+                    LOG.debug("Shrink check, {} > {} period={}ms {}", threads, minThreads, TimeUnit.NANOSECONDS.toMillis(shrinkPeriod), QueuedThreadPool.this);
+
+                long shrinkThreshold = _shrinkThreshold.get();
+                long threshold = shrinkThreshold;
+
+                // If the threshold is too far in the past,
+                // advance it to be one idle timeout before now plus a bit of shrink period.
+                if (NanoTime.elapsed(threshold, now) > idleTimeoutNanos)
+                    threshold = now - idleTimeoutNanos;
+
+                // advance the threshold by one shrink period
+                threshold += shrinkPeriod;
+
+                // Is the new threshold in the future?
+                if (NanoTime.isBefore(now, threshold))
+                {
+                    // Yes - we cannot shrink yet, so continue looking for jobs
+                    if (LOG.isDebugEnabled())
+                        LOG.debug("Shrink skipped, threshold={}ms in future {}", NanoTime.millisElapsed(now, threshold), QueuedThreadPool.this);
+                    return false;
+                }
+
+                // We can shrink if we can update the atomic shrink threshold?
+                if (_shrinkThreshold.compareAndSet(shrinkThreshold, threshold))
+                {
+                    // Yes - we have shrunk, so exit.
+                    if (LOG.isDebugEnabled())
+                        LOG.debug("SHRUNK, threshold={}ms in past {}", NanoTime.millisElapsed(threshold, now), QueuedThreadPool.this);
+                    return true;
+                }
+            }
+            else
+            {
+                // We reached min threads, continue looking for jobs
+                if (LOG.isDebugEnabled())
+                    LOG.debug("At min threads, {} > {} {}", threads, minThreads, QueuedThreadPool.this);
+                return false;
+            }
+        }
+    }
+
     private void doRunJob(Runnable job)
     {
         try
@@ -1103,7 +1162,7 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
                             job = _jobs.poll();
                         }
 
-                        if (shrinkIfNeeded(idleTimeoutNanos))
+                        if (shrinkIfNeeded())
                             break;
                     }
                     catch (InterruptedException e)
@@ -1125,63 +1184,6 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
                 // There is a chance that we shrunk just as a job was queued,
                 // so check again if we have sufficient threads to meet demand.
                 ensureThreads();
-            }
-        }
-
-        /**
-         * @return true if shrunk, false otherwise.
-         */
-        private boolean shrinkIfNeeded(long idleTimeoutNanos)
-        {
-            while (true)
-            {
-                // No jobs available, should we shrink?
-                int threads = getThreads();
-                int minThreads = getMinThreads();
-                if (threads > minThreads)
-                {
-                    // We have an idle timeout and excess threads, so check if we should shrink?
-                    long now = NanoTime.now();
-                    long shrinkPeriod = idleTimeoutNanos / getIdleTimeoutMaxShrinkCount();
-                    if (LOG.isDebugEnabled())
-                        LOG.debug("Shrink check, {} > {} period={}ms {}", threads, minThreads, TimeUnit.NANOSECONDS.toMillis(shrinkPeriod), QueuedThreadPool.this);
-
-                    long shrinkThreshold = _shrinkThreshold.get();
-                    long threshold = shrinkThreshold;
-
-                    // If the threshold is too far in the past,
-                    // advance it to be one idle timeout before now plus a bit of shrink period.
-                    if (NanoTime.elapsed(threshold, now) > idleTimeoutNanos)
-                        threshold = now - idleTimeoutNanos;
-
-                    // advance the threshold by one shrink period
-                    threshold += shrinkPeriod;
-
-                    // Is the new threshold in the future?
-                    if (NanoTime.isBefore(now, threshold))
-                    {
-                        // Yes - we cannot shrink yet, so continue looking for jobs
-                        if (LOG.isDebugEnabled())
-                            LOG.debug("Shrink skipped, threshold={}ms in future {}", NanoTime.millisElapsed(now, threshold), QueuedThreadPool.this);
-                        return false;
-                    }
-
-                    // We can shrink if we can update the atomic shrink threshold?
-                    if (_shrinkThreshold.compareAndSet(shrinkThreshold, threshold))
-                    {
-                        // Yes - we have shrunk, so exit.
-                        if (LOG.isDebugEnabled())
-                            LOG.debug("SHRUNK, threshold={}ms in past {}", NanoTime.millisElapsed(threshold, now), QueuedThreadPool.this);
-                        return true;
-                    }
-                }
-                else
-                {
-                    // We reached min threads, continue looking for jobs
-                    if (LOG.isDebugEnabled())
-                        LOG.debug("At min threads, {} > {} {}", threads, minThreads, QueuedThreadPool.this);
-                    return false;
-                }
             }
         }
     }
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
@@ -980,11 +980,13 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
         while (true)
         {
             // No jobs available, should we shrink?
-
-            long encoded = _counts.get();
-            int threads = AtomicBiInteger.getHi(encoded);
-            int idle = AtomicBiInteger.getLo(encoded);
+            int threads = getThreads();
             int minThreads = getMinThreads();
+
+            // This minThreads comparison isn't atomic, so there's a chance that multiple threads might
+            // enter this branch concurrently and shrink the pool below minThreads; this is okay
+            // as when that happens, the finally block in Runner.run() compensates b/c it calls
+            // ensureThreads() - so we can go below minThreads but that's a very transient state.
             if (threads > minThreads)
             {
                 // We have an idle timeout and excess threads, so check if we should shrink?
@@ -1016,14 +1018,6 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
                 // We can shrink if we can update the atomic shrink threshold?
                 if (_shrinkThreshold.compareAndSet(shrinkThreshold, threshold))
                 {
-                    // can we decrement the thread count?
-                    if (!_counts.compareAndSet(encoded, threads - 1, idle - 1))
-                    {
-                        // Rewind the threshold
-                        _shrinkThreshold.addAndGet(-shrinkPeriod);
-                        continue;
-                    }
-
                     // Yes - we have shrunk, so exit.
                     if (LOG.isDebugEnabled())
                         LOG.debug("SHRUNK, threshold={}ms in past {}", NanoTime.millisElapsed(threshold, now), QueuedThreadPool.this);
@@ -1127,7 +1121,6 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
             if (LOG.isDebugEnabled())
                 LOG.debug("Runner started for {}", QueuedThreadPool.this);
 
-            boolean shrunk = false;
             boolean idle = true;
             try
             {
@@ -1159,8 +1152,7 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
                             job = _jobs.poll();
                         }
 
-                        shrunk = shrink();
-                        if (shrunk)
+                        if (shrink())
                             break;
                     }
                     catch (InterruptedException e)
@@ -1175,12 +1167,12 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
                 removeThread(thread);
 
                 // Decrement the total thread count and the idle count if we had no job.
-                if (!shrunk)
-                    addCounts(-1, idle ? -1 : 0);
+                addCounts(-1, idle ? -1 : 0);
                 if (LOG.isDebugEnabled())
                     LOG.debug("{} exited for {}", thread, QueuedThreadPool.this);
 
                 // There is a chance that we shrunk just as a job was queued,
+                // or multiple concurrent threads ran out of jobs,
                 // so check again if we have sufficient threads to meet demand.
                 ensureThreads();
             }
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
@@ -114,6 +114,7 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
     private long _stopTimeout;
     private Executor _virtualThreadsExecutor;
     private int _shrinkCount = 1;
+    private boolean _aggressiveShrinking;
 
     public QueuedThreadPool()
     {
@@ -562,9 +563,24 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
     /**
      * @return the maximum number of idle threads to exit in one idle timeout period
      */
+    @ManagedAttribute("maximum number of idle threads to exit in one idle timeout period")
     public int getIdleTimeoutMaxShrinkCount()
     {
         return _shrinkCount;
+    }
+
+    public void setAggressiveShrinking(boolean aggressiveShrinking)
+    {
+        _aggressiveShrinking = aggressiveShrinking;
+    }
+
+    /**
+     * @return true if the idle threads should exit before the idle timeout period, false otherwise
+     */
+    @ManagedAttribute("should the idle threads exit before the idle timeout period")
+    public boolean isAggressiveShrinking()
+    {
+        return _aggressiveShrinking;
     }
 
     /**
@@ -1072,12 +1088,20 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
             boolean idle = true;
             try
             {
-                exit: while (_counts.getHi() > Integer.MIN_VALUE)
+                while (_counts.getHi() > Integer.MIN_VALUE)
                 {
                     try
                     {
+                        boolean aggressiveShrinking = isAggressiveShrinking();
                         long idleTimeoutNanos = TimeUnit.MILLISECONDS.toNanos(getIdleTimeout());
                         Runnable job = idleJobPoll(idleTimeoutNanos);
+
+                        if (job == null && !aggressiveShrinking)
+                        {
+                            if (shrinkIfNeeded(idleTimeoutNanos))
+                                break;
+                        }
+
                         while (job != null)
                         {
                             idle = false;
@@ -1099,47 +1123,8 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
                             job = _jobs.poll();
                         }
 
-                        // No jobs available, should we shrink?
-                        int threads = getThreads();
-                        if (threads > getMinThreads())
-                        {
-                            // We have an idle timeout and excess threads, so check if we should shrink?
-                            long now = NanoTime.now();
-                            long shrinkPeriod = idleTimeoutNanos / getIdleTimeoutMaxShrinkCount();
-                            if (LOG.isDebugEnabled())
-                                LOG.debug("Shrink check, {} > {} period={}ms {}", threads, getMinThreads(), TimeUnit.NANOSECONDS.toMillis(shrinkPeriod), QueuedThreadPool.this);
-                            while (true)
-                            {
-                                long shrinkThreshold = _shrinkThreshold.get();
-                                long threshold = shrinkThreshold;
-
-                                // If the threshold is too far in the past,
-                                // advance it to be one idle timeout before now plus a bit of shrink period.
-                                if (NanoTime.elapsed(threshold, now) > idleTimeoutNanos)
-                                    threshold = now - idleTimeoutNanos;
-
-                                // advance the threshold by one shrink period
-                                threshold += shrinkPeriod;
-
-                                // Is the new threshold in the future?
-                                if (NanoTime.isBefore(now, threshold))
-                                {
-                                    // Yes - we cannot shrink yet, so break and continue looking for jobs
-                                    if (LOG.isDebugEnabled())
-                                        LOG.debug("Shrink skipped, threshold={}ms in future {}", NanoTime.millisElapsed(now, threshold), QueuedThreadPool.this);
-                                    break;
-                                }
-
-                                // We can shrink if we can update the atomic shrink threshold?
-                                if (_shrinkThreshold.compareAndSet(shrinkThreshold, threshold))
-                                {
-                                    // Yes - we have shrunk, so exit.
-                                    if (LOG.isDebugEnabled())
-                                        LOG.debug("SHRUNK, threshold={}ms in past {}", NanoTime.millisElapsed(threshold, now), QueuedThreadPool.this);
-                                    break exit;
-                                }
-                            }
-                        }
+                        if (aggressiveShrinking && shrinkIfNeeded(idleTimeoutNanos))
+                            break;
                     }
                     catch (InterruptedException e)
                     {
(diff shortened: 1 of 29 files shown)


sbordet:
Alternative to #9354.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -1027,54 +1057,68 @@ public void run()
...
+                                    long lastShrink = _lastShrink.get();
+                                    long prevShrink = lastShrink;
+                                    // If the shrink window is too far in the past,
+                                    // advance it to be one idle timeout before now.
+                                    if (NanoTime.elapsed(lastShrink, now) > idleTimeoutNanos)
+                                        prevShrink = now - idleTimeoutNanos;
+
+                                    // Add shrink periods until the window is full.
+                                    long nextShrink = prevShrink + shrinkPeriod;
+                                    if (NanoTime.isBefore(now, nextShrink))
```
gregw: I think it reads better as:
```suggestion
                                    if (NanoTime.isAfter(nextShrink, now))
```

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -1027,54 +1057,68 @@ public void run()
...
+                                    LOG.debug("Shrink check, period={}ms {}", TimeUnit.NANOSECONDS.toMillis(shrinkPeriod), QueuedThreadPool.this);
+                                while (true)
                                 {
-                                    if (LOG.isDebugEnabled())
-                                        LOG.debug("shrinking {}", QueuedThreadPool.this);
-                                    break;
+                                    long lastShrink = _lastShrink.get();
+                                    long prevShrink = lastShrink;
+                                    // If the shrink window is too far in the past,
+                                    // advance it to be one idle timeout before now.
```
gregw: I don't like using the word window, as we don't fully embrace that abstraction. how about:
```suggestion
                                    // If the shrink threshold is too far in the past,
                                    // set it to just one idle timeout before now.
```

sbordet:
@magibney this is an alternative to your proposal.

It works by maintaining a window of 1 idle timeout, and moving the window every time a thread is eligible for shrinking.

The code reads `_lastShrink` as the start of the window, and if it is too far back in the past, it updates the start of the window to be `now - idleTimeout`.

Then, for every thread eligible for shrinking, the start of window is updated by 1 shrink period, until it is less than `now`. This guarantees that only `idleTimeoutMaxShrinkCount` threads are expired within an idle timeout period.

Say the idle timeout is 10s, and `idleTimeoutMaxShrinkCount=5`.
At time `t` an idle thread is eligible to shrink, so the window start is set to `t-10`.
Since `t-10<t`, the thread is exited and the window moved to `t-8`.
At `t+1` another idle thread is eligible to shrink, and the window start is at `t-8`, the thread is exited and the window moved at `t-6`.
At `t+5` another idle thread is eligible to shrink; the window is at `t-6`, but it's larger than one idle timeout, so it is set at `t-5`, the thread is exited and the window moved at `t-3`.
At `t+6` many threads become idle and eligible to shrink; the window is at `t-3`; one thread is exited and moves the window at `t-1`; another exits and moves to `t+1`; another to `t+3`; another to `t+5`. The next one _would_ move the window at `t+7`, but that is past the current "now" `t+6`, so it is not exited and waits another idle timeout.
And so on.

Graphically the example above is:

```
time t, window start at t-10
event            v
time   -9-------1t------
window ^

time t, window updated to t-8
event            v
time   --8------1t------
window   ^

time t+1, window updated to t-6
event            vv
time   ----6-----t1-----
window     ^

time t+5, window updated to t-3
event            vv   v
time   -------3--t----5-
window        ^

time t+6, 3 window updates to t+5
                       3
event            vv   vv
time   ----------t----56
window                ^
```

When many threads idle at the same time, one will set the window start at `t-10` and the first update at t-8, other 4 will do `+2` updates up to `t`, and the others cannot move the window past `t` and will wait another idle timeout.

@magibney can you try this algorithm for your use case?
We would also appreciate a review of this PR.
Thanks!

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -1027,54 +1057,68 @@ public void run()
...
+                                long shrinkPeriod = idleTimeoutNanos / getIdleTimeoutMaxShrinkCount();
+                                if (LOG.isDebugEnabled())
+                                    LOG.debug("Shrink check, period={}ms {}", TimeUnit.NANOSECONDS.toMillis(shrinkPeriod), QueuedThreadPool.this);
+                                while (true)
                                 {
-                                    if (LOG.isDebugEnabled())
-                                        LOG.debug("shrinking {}", QueuedThreadPool.this);
-                                    break;
+                                    long lastShrink = _lastShrink.get();
+                                    long prevShrink = lastShrink;
```
gregw:
```suggestion
                                    long shrinkThreshold = lastShrink;
```

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -1027,54 +1057,68 @@ public void run()
...
+                                    // If the shrink window is too far in the past,
+                                    // advance it to be one idle timeout before now.
+                                    if (NanoTime.elapsed(lastShrink, now) > idleTimeoutNanos)
+                                        prevShrink = now - idleTimeoutNanos;
```
gregw: Could we just do:
```java
    long shrinkThreshhold = Math.max(lastShrink, now - idleTimeoutNanos);
```
or does nano wrapping make this not work?  It is the semantic we want.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -1027,54 +1057,68 @@ public void run()
...
+
+                                    // Add shrink periods until the window is full.
```
gregw: I don't like "until" in this comment, implies looping until it is full.
```suggestion
                                    // The next shrink can happen one shrink period after the threshold. 
```

gregw (changes requested):
just some naming niggles

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
sbordet: There is no `NanoTime.isAfter()`, and it was explicitly not added to keep the `NanoTime` API minimal.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
sbordet: Yes, it could wrap, so we can't use `Math.max()`.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
sbordet: Uhm, it is a window between some time in the past and the current time.
And it is a credit-based scheme like the HTTP/2 flow control window, every expired thread reduces the window until zero, and the passing of time gives it more credit.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
sbordet: But it does not "happen" at that time... `nextShrink` is perhaps badly named (but can't find better) but it is unrelated with _when_ the shrink happens (which is now, not at `nextShrink`).

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
gregw: doh!

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
gregw: I think it really improves readability, so it is a small price to pay.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
gregw: Maybe call it `nextShrinkAfter`.

Eitherway, the "until" text doesn't help an it is better to just say when the next shrink can happen.

```suggestion
                                    // Another shrink is not allowed before one shrink period after the threshhold.
```

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
gregw: I understand what a window system is and how it is an analogy to the algorithm here.
I just do not think that in this case it adds any clarity by describing it as a window.

Best just to say when the next shrink is allowed to happen.

gregw (changes requested):

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
sbordet: > I just do not think that in this case it adds any clarity by describing it as a window.

But it's not a threshold either, and calling it so it's even more confusing.

> Best just to say when the next shrink is allowed to happen.

The value is completely unrelated to "when the next shrink is allowed to happen", so I cannot say that.

I don't understand why you're so against the term "window", it is perfectly representing what it is: a value that has a fixed maximum, can decrement until it closes down to zero, and when it closes, no more air (err shrink), but can reopen 😃 

We can call it `shrinkCredit` if you want, but seems to me unnecessary wording when "window" is precisely describing it.
"Allowance"? "Quota"?

If we change the math to use a counter instead of a nanoTime, how would you call the counter? 
There would be N "shrinkCredits" and each time a thread is eligible to shrink you would decrement by 1 credit.
As time passes you gain 1 credit for every shrink period that has passed.

gregw: A window has two ends that adjust and can't be allowed to get too far apart.  This just doesn't have that abstraction.  It is kind of similar, but not close enough that the term window introduces confusion rather than clarity.
It only works for you because you have implemented a window with a single counter (window size), but the origin of the term really depends on having two pointers and a window between them that is limited in size.  

It is not a quota, nor allowance, nor credit.   It is a time that is related to now, that moves forward in time and acts as a threshold to determine if an expiry is allowed or skipped.

magibney:
Thanks for this, it definitely looks promising!

IIUC some of the logic in this PR is quite similar to the earlier iteration of #9354 ([diff here](https://github.com/eclipse/jetty.project/compare/beee5ffff0..d365f55fec7951adb946cae15d517c07731df714)): updating a backdated timestamp upon shrink. The main difference between the approach of this PR and the current state of #9354 (aside from the fact that the latter refactors out the shrink bookeeping into a separate class) is that the latter tracks pool-level idle state as a "first-class citizen" (more discussion below).

It was pretty easy to port my tests over. Some of them pass, but the most realistic one appears to have some issues, namely `testShrinkToSustainedBaseline`:
```
2023-03-15 14:42:08.613:INFO :oejut.QueuedThreadPoolShrinkTest:main: CONFIG: baselineTargetThreadCount=178, expectStableThreadCount=179, idleTimeoutDecay=326, busy=false, shift=3, seed=2a0b1a2dcd4f11ba
2023-03-15 14:42:11.309:INFO :oejut.QueuedThreadPoolShrinkTest:main: OK threadCount expect=835, actual=999, variance=-1.0
2023-03-15 14:42:11.815:INFO :oejut.QueuedThreadPoolShrinkTest:main: OK threadCount expect=670, actual=999, variance=-1.0
2023-03-15 14:42:12.320:INFO :oejut.QueuedThreadPoolShrinkTest:main: OK threadCount expect=505, actual=999, variance=-1.0
2023-03-15 14:42:12.825:INFO :oejut.QueuedThreadPoolShrinkTest:main: OK threadCount expect=341, actual=999, variance=-1.0
2023-03-15 14:42:13.330:INFO :oejut.QueuedThreadPoolShrinkTest:main: OK threadCount expect=176, actual=999, variance=-1.0
2023-03-15 14:42:13.835:INFO :oejut.QueuedThreadPoolShrinkTest:main: OK threadCount expect=11, actual=999, variance=-1.0
too many consecutive variance faults; expect=-153, actual=999, variance=-1.0 (baselineTargetThreadCount=178, expectStableThreadCount=179, idleTimeoutDecay=326, busy=false, shift=3, seed=2a0b1a2dcd4f11ba)
2023-03-15 14:42:14.342:INFO :oejut.QueuedThreadPoolShrinkTest:main: OK threadCount expect=-153, actual=999, variance=-1.0
```
This test spins up 1k threads, then pings the pool periodically with _n_ concurrent threads (_n_ here being `expectStableThreadCount` -- 179 in this case) and waits for the superfluous threads to die, which they don't appear to do.

I think this is happening because of random assignment of threads; if you have sufficient load to randomly "ping" a thread alive within its `idleJobPoll` window, threads might never check eligibility to expire, even if load could be serviced by many fewer threads. I don't think that's different now than it was before, so maybe not a blocker here. And I think the only way around it is to track pool-level idle capacity as a "first-class citizen", e.g. as done in the `ShrinkManager` timestamp deque in  #9354. The only practical downside of _not_ explicitly tracking pool-level idle capacity is that it's possible for a handful of clients, each with a sustained load of short jobs, to artificially inflate the pool size. I'll take a look at verifying that this behavior already existed (i.e., was not introduced by this PR).

That said, even if the behavior hasn't (strictly speaking) changed, the user expectations _will_ have changed, in a way that may make more apparent the disparity between how quickly the user thinks the pool _should_ shrink (based on shrink rate configuration) and how quickly it _actually_ shrinks.

`testShrinkToSustainedBaseline` is arguably the most realistic test among the suite of tests I introduced in #9354, in the sense that it tests idleTimeout behavior on a pool that's continuing to get active traffic. I added it to ensure that the new approach would work as expected in a production system, and it's designed to apply more stress than one might reasonably expect in a real production deployment. So the question of whether this alternative PR will meet our needs in practice is harder to answer: a passing `testShrinkToSustainedBaseline` strongly suggests that a solution _will_ suffice, but a failing `testShrinkToSustainedBaseline` doesn't necessarily mean it _won't_ suffice ...

magibney:
>I'll take a look at verifying that this behavior already existed (i.e., was not introduced by this PR).

I followed up on this, and it turns out that the `testShrinkToSustainedBaseline` test, when adjusted to work with the default shrink rate of 1 (`CONFIG: baselineTargetThreadCount=954, expectStableThreadCount=955, idleTimeoutDecay=1, busy=false, shift=1, seed=9efe9d4bb15bfdc3`) works fine against the `main` branch (shrinks as expected to the `baselineTargetThreadCount` --  works on  #9354 too fwiw), but doesn't shrink at all on this PR branch. So there's definitely something that needs to be addressed there ...

gregw (dismissed):

gregw:
Can you make the source of your test available, so we can work on the exact same use-case.

gregw:
I've pushed a slightly different approach.

I added a test that spikes 1000 threads then keeps 200 busy. Old approach didn't shrink. This new approach does shrink down to 200.

However, there are some problems with idle thread accounting, as some other tests fail.

I'm on vacation from tomorrow, so pushing so @sbordet can find my mistakes.

gregw:
@sbordet I think I have fixed the idle accounting.
However, I had to adjust a test as we kind of have a race when a job finishes and the thread might go idle or might now exit.   Green build for me, but needs careful review.

magibney:
Making sure you've seen #9512, opened against this PR. The most recent changes don't actually seem to respect the "peak hold" aspect of `idleTimeout` -- i.e., I think threads start expiring immediately.

Feel free to merge that test suite in here if you like. I also took a bit of extra effort to make it easy to run against `main`, to validate parity of default behavior (should be only a one-line change necessary when porting to `main` branch).

gregw:
@magibney Yes I have seen that. Thanks.

Yeah, this PR is not going to wait an idleTimeout before starting to shrink.  It will shrink up to the shrink count threads every idle period, starting immediately

magibney:
> Yeah, this PR is not going to wait an idleTimeout before starting to shrink. It will shrink up to the shrink count threads every idle period, starting immediately

Perhaps I'm missing something, but then how does this differ from the behavior of the current `main` branch if you divide `idleTimeout / idleTimeoutMaxShrinkCount`?

gregw:
The difference with this latest branch is when we do the shrink check. With previous versions we did the check only when we idle timed out... Which if we had enough small busy tasks, never happened, so 200 concurrent tasks could keep 1000 threads not idle 

We now do the shrink check whenever the queue is empty of jobs. So with 200 busy threads the pool still shrinks from 1000 to 200.

However the shrinking starts immediately after a spike in the threads.  

The only thing we could possibly add is a minimum time to live. So if a thread has been alive for less than an idle timeout, it will not shrink.  This would make the peaks fatter. @sbordet thoughts?

magibney:
Oh! I see. Then iiuc the current incarnation of this PR isn't so much a replacement for #9354 as it is a fix for a latent issue surfaced by tests that were added in #9354. In that case though, for the current state of this PR I don't see the purpose of adding the `idleTimeoutMaxShrinkCount` param: if you start to shrink immediately then setting `idleTimeout=X, idleTimeoutMaxShrinkCount=Y` would be functionally equivalent to setting `idleTimeout=X/Y`, no?

gregw:
@magibney yeah the idle timeout and count now make less sense... unless we add in a min time to live equal to the idle timeout.  We need to think about this some more, but I like the direction we are going.   However, we need to do some performance testing on this to see if we have changed anything one way or the other.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -1027,89 +1069,94 @@ public void run()
...
-                        // run job
-                        if (LOG.isDebugEnabled())
-                            LOG.debug("run {} in {}", job, QueuedThreadPool.this);
-                        runJob(job);
-                        if (LOG.isDebugEnabled())
-                            LOG.debug("ran {} in {}", job, QueuedThreadPool.this);
                     }
                     catch (InterruptedException e)
                     {
+                        boolean interrupted = Thread.interrupted();
```
lorban: Won't this virtually always be false as `InterruptedException` is supposed to be thrown after a `Thread.interrupted()` returns true?

lorban (commented):
Performance wise, I don't think these changes make any difference: a loaded pool isn't going to shrink and spends most of its time in `BlockingArrayQueue`'s `offer` and `poll` methods plus on the `_counts` atomic. The few accesses to the `_shrinkThreshold` atomic only happen when the pool has no queued job, so the extra needed cycles shouldn't matter.

magibney:
Thanks for confirming. Yes, the diff for the current iteration could be a lot more straightforward -- basically an internal refactoring of the loop around `idleJobPoll()` (and any related tests), with no documentation/API updates necessary.

Separately considering the "minimum TTL" thing: I'm increasingly thinking that, at a high level, an approach that extracts the shrink logic out into an interface (similar to what's done with `ShrinkManager` in #9354) would be beneficial.

The implementation in #9354 supports "minimum TTL" and then linear decay; but given the interface, one could write a trivial implementation that recreates exactly the current behavior, supporting immediate linear decay via updates to a single `AtomicLong`. It also occurs to me that the `ShrinkManager` impl in #9354 could be straightforwardly tweaked to support exponential decay (asymptotic to `minThreads`)!

I'm not suggesting that any more complex shrink algorithms be implemented in jetty; but simply that the lifecycle hooks necessary to support pluggable decay are quite clear:
```java
/** called to register a thread as idle (not actively performing work) */
void onIdle();

/** called to register a thread as about to perform work */
void onBusy();

/** called to check whether the pool should shrink by at least on thread */
boolean pollIdleShrink();

/** fallback to ensure consistency upon expected thread death */
void prune();
```
Or however one might prefer to rename the methods?

I'm fairly confident btw that introducing the concept of "minimum TTL" will inevitably run into the same problem that motivated the creation of `ShrinkManager` on #9354: we're either tracking TTL baseline per thread (in which case behavior varies arbitrarily according to thread scheduling), or we're tracking via a single pool-level variable (similar to what we do now) in which case it's not clear what TTL applies to. Or, as in the approach taken by `ShrinkManager`, we must track the TTL baseline of all threads individually but interchangeably at the pool level, and act on them in a consistent way.

magibney:
> Performance wise, I don't think these changes make any difference

Hmm this makes sense looking at the code for current `main` branch. And as I [mentioned above](#issuecomment-1470833629), the "sustained baseline" test indeed passed on current `main` branch; it only failed on the earlier iteration of this branch. So I suspect that the current iteration of this PR incidentally removed the functionality that motivated this PR while fixing a problem that was introduced in the earlier iteration of this PR.

gregw:
We have a release coming in the next week or two. We could merge this as is, or work on it a bit more for the next release.

I'm not opposed to making the algorithm extensible.

I'm also cautious about adding configuration that we are not 100% on, as once in it is forever... hence i lean towards skipping this release cycle.

@sbordet thoughts?

magibney:
At a high level, I think this PR in its current state, for config
```properties
idleTimeout=X
idleTimeoutMaxShrinkCount=Y
aggressiveShrinking=true
```
is functionally nearly identical to current `main` branch for config
```properties
idleTimeout=X/Y
```
IIUC both cases should shrink 1 thread per `X/Y` interval.

At this point I'm honestly not sure what problem this PR is trying solve.

I'm pretty sure existing behavior on main branch is equivalent to `aggressiveShrinking=true`, and that setting `aggressiveShrinking=false` would put you at substantial risk of having a pool that never shrinks (by making it possible for a thread to acquire all its jobs via `idleJobPoll()`, without ever checking eligibility to shrink).

I'm still confident in #9354 and happy to answer any questions if you're open to taking another look at it. Even if just to use that as a base for pluggable shrinking (again, extracting the shrink logic, as in the commit I just pushed to that branch, seems like a win regardless of whether the purpose is to make it pluggable, or just to gain some logical separation).

lorban:
@magibney The `aggressiveShrinking` config param I added was an experiment I made to compare two slightly different design choices. I've now removed it and made what was the aggressive mode the only implementation.

I'm sorry that I haven't been very open about the work I've been doing, but your pull request shed some light on a very interesting problem, and I needed to fully grasp all the details before taking any decision as this is a crucial piece of code. But I'm now done and you can consider the current state of this pull request as the most likely final version of the code, so let me give you some details.

This new shrink implementation is different from the previous one in two ways:
 - The previous implementation requires the thread pool to be virtually idle before it starts shrinking. If it handles a peak that makes the thread count grow to max and after that you always apply a moderate load that keeps just a few threads constantly busy, it will never shrink. The new one configured with `aggressiveShrinking=false` reproduced this behavior, and I added this parameter for the sole purpose of comparing both to better understand them. This new implementation will always try to shrink the pool as soon as one thread runs out of jobs, as we decided it was an improvement worth doing that didn't have any noticeable drawback, especially on the performance side of things.
- You are right that the old implementation with `idleTimeout=X/Y` would shrink at the same pace (assuming your pool becomes idle of course), except for one detail: this one is a lot more performant as it can make Y threads die for each `idleTimeout` period, which basically means the threads needs to be awoken Y times less often by the scheduler before reaching a new stable point.

Thanks for putting us on our toes, we revisited and improved this core component of Jetty in a way we never thought about. Unfortunately, we decided against integrating your suggested shrink manager in the standard code base as its abstraction looks too specific to your implementation, it suffers from a few bugs (the `long[]` is read and updated concurrently without any form of barrier for instance) and fundamentally relies on an algorithm that creates contention, which makes it noticeably slower than our own. You can check the results of the `QTPBenchmark` to get an idea of the perf difference on a loaded pool, it's quite substantial.

But `runJob(Runnable job)` is `protected`, as well as `shrinkIfNeeded()` now, so it seems that you could re-implement you algorithm by subclassing `QueuedThreadPool` and overriding those two methods that provide you the hooks you need should you want to run with it.

lorban:
@sbordet do you have anything more to say about this subject? WDYT of this version?

magibney:
Ah, this makes sense. Thanks for the detailed explanation, and apologies for prematurely calling out the `aggressiveShrinking`!

In the course of investigating whether this change would include the requisite hooks for the functionality I'm looking for, I realized that the new logic in this PR makes it much cleaner/easier, and I generally really like this change! -- though at the moment I suspect the hooks aren't sufficient to extend via simple subclassing.

I've opened #9532 against this PR branch. It addresses the performance concerns (the QTPBenchmark shows parity, regardless of configuration of `idleTimeout` or `idleTimeoutMaxShrinkCount`), and all the tests added for #9354 pass as well.

> it suffers from a few bugs (the long[] is read and updated concurrently without any form of barrier for instance)

Regarding this, I really think it's a benign race condition. As mentioned [here](https://github.com/eclipse/jetty.project/pull/9354#issuecomment-1468247201):
> It's good to call this out; I did this knowingly and I think it should be ok. For practical purposes, we can lean on the fact that the stored values are effectively relevant within the tolerance of idleTimeout. Writes only happen at the head, and we don't care about reading those values at all, and the exact values that might overwrite each other are a benign race condition because they'll all be well within the idleTimeout tolerance, so we actually don't care about the exact value that gets written. As far as reading values, we only care about tail, and we only start to pay attention to those tail values once their idleTimout is passed. The race condition at tail does happen with failed CaS; but it's benign here too, because we don't care who is checking a value, so long as someone is. The top priority at the tail is to ensure that the array entries remain in order (within the acceptable tolerance of race conditions), so note that we don't actually ever write to the array at tail, we simply update the tail pointer.
>
>Again, the tests bear all this out, and are designed to stress the system pretty hard, asserting very tight tolerances wrt expected shrink behavior.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -90,11 +90,11 @@ public class QueuedThreadPool extends ContainerLifeCycle implements ThreadFactor
      * <dt>Hi</dt><dd>Total thread count or Integer.MIN_VALUE if the pool is stopping</dd>
      * <dt>Lo</dt><dd>Net idle threads == idle threads - job queue size.  Essentially if positive,
      * this represents the effective number of idle threads, and if negative it represents the
-     * demand for more threads</dd>
+     * demand for more threads, aka the job queue's size.</dd>
```
sbordet:
```suggestion
     * demand for more threads, which is equivalent to the job queue's size.</dd>
```

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -936,6 +968,82 @@ protected void runJob(Runnable job)
...
+            {
+                // We reached min threads, continue looking for jobs
+                if (LOG.isDebugEnabled())
+                    LOG.debug("At min threads, {} > {} {}", threads, minThreads, QueuedThreadPool.this);
+                return false;
+            }
+        }
+    }
+
+    private void doRunJob(Runnable job)
```
sbordet: Should not this method be in the `Runner` class?

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -936,6 +968,82 @@ protected void runJob(Runnable job)
         job.run();
     }
 
+    /**
+     * <p>Attempts to shrink the thread pool by one thread if {@link #getThreads()} is greater than {@link #getMinThreads()}.</p>
+     * @return true if shrunk, false otherwise.
+     */
+    protected boolean shrinkIfNeeded()
```
sbordet: Perhaps rename to just `shrink()`? The `boolean` return type implies that may or may not shrink the current thread.
Or perhaps `evict()` because it refers to the current thread? The shrinking of the pool is a consequence of evicting the current thread.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -1026,90 +1134,55 @@ public void run()
             boolean idle = true;
             try
             {
-                Runnable job = null;
-                while (true)
+                while (_counts.getHi() > Integer.MIN_VALUE)
```
sbordet: I'd prefer `!=` instead of `>`.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -536,6 +537,37 @@ public void setVirtualThreadsExecutor(Executor executor)
...
+        if (shrinkCount < 1)
+            throw new IllegalArgumentException("Invalid shrink count " + shrinkCount);
+        _shrinkCount = shrinkCount;
+    }
+
+    /**
+     * @return the maximum number of idle threads to exit in one idle timeout period
+     */
+    @ManagedAttribute("maximum number of idle threads to exit in one idle timeout period")
+    public int getIdleTimeoutMaxShrinkCount()
```
sbordet: @gregw suggested to rename this just `maxShrinkCount` but maybe `maxEvictCount` is better?

tests/jetty-jmh/src/main/java/org/eclipse/jetty/util/thread/jmh/QTPBenchmark.java
```diff
@@ -0,0 +1,93 @@
...
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+@State(Scope.Benchmark)
+@Warmup(iterations = 10, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
+@Measurement(iterations = 10, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
+public class QTPBenchmark
```
sbordet: Rename to `QueuedThreadPoolBenchmark`.

sbordet (commented):
@lorban reviewed -- I request changes but I cannot do it through web UI since I'm the author of the PR.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
lorban: Done.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
lorban: Done.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
lorban: `evict()` sounds better, going with that.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
lorban: Done.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
lorban: `maxEvictCount` sounds better, going with that.

tests/jetty-jmh/src/main/java/org/eclipse/jetty/util/thread/jmh/QTPBenchmark.java
lorban: Done.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -936,6 +968,65 @@ protected void runJob(Runnable job)
...
+     * <p>Attempts to evict the current thread from the pool if {@link #getThreads()} is greater than {@link #getMinThreads()}.</p>
+     * @return true if the current thread was evicted, false otherwise.
+     */
+    protected boolean evict()
+    {
+        long idleTimeoutNanos = TimeUnit.MILLISECONDS.toNanos(getIdleTimeout());
+        while (true)
+        {
+            // No jobs available, should we shrink?
+            int threads = getThreads();
```
hiteshk25: It may be minor but looks like there is possibility that thread may read stale `getThreads` value as thread which about to exit may not have decrease this value. That can cause number of threads to alive less than min-threads?

gregw: I think there could be a problem.

It used to work when we could only shrink a single thread per period as if the thread count was stale, then the last shrink time would have been pushed out.

If we are shrinking 10 threads per period, then 10 threads could all wake up, see threads=minthreads+1 and all decide to shrink.  They can all advance the shrink threshold.  If they collide whilst doing so, the they will loop and get the threads count again.  But there is a chance they might not collide on updating the shrink threshhold, so we might let too many threads exit.

thinking....

gregw: @lorban @sbordet this is a difficult one to solve, as we have 2 Atomic counters, and any attempt to keep them "atomic" is bound to be racey.

Currently, we make no CaS operation on the number of threads when a worker thread exits.  Instead we just decrement the thread count and optionally the idle count.

If we are shrinking, perhaps we should CaS the thread count decrement with the number of threads used to make the decision to shrink.  If we fail that CaS, then we have to restore the shrink interval (or more precisely deduct a shrink period from whatever value it currently is) and then loop and try again.  If we success the CaS, we have to make sure the runner does not further decrement the thread count when the thread exits.

I can give this ago... but most times when we think we are being clever with two atomics, we are wrong!

gregw: @lorban @sbordet I've implemented this.  It is not too horrid, but hard to convince myself it is not racy.  But I'm thinking the worst case is we don't shrink when we could.

@magibney I'm struggling to see how this can easily be generalised as the shrink() method needs access to two atomics to CaS on them.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
gregw: Oh sorry. Didn't see this until after I renamed to `shrink`, which I prefer... but I can live with `evict`.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -1023,95 +1125,80 @@ public void run()
...
 
-                        idle = false;
-
-                        // run job
-                        if (LOG.isDebugEnabled())
-                            LOG.debug("run {} in {}", job, QueuedThreadPool.this);
-                        runJob(job);
-                        if (LOG.isDebugEnabled())
-                            LOG.debug("ran {} in {}", job, QueuedThreadPool.this);
+                        shrunk = shrink();
```
hiteshk25: If thread reaches here after doing `one job immediately`, then it might shirk that thread. And that may cause further thread-creation and thread-exit pattern (thrashing)

I think bigger question here is do we want to make sure thread remain idle for `idletimeout` (or keep-alive time) ?

gregw:
@olamy what is going on here: https://jenkins.webtide.net/blue/organizations/jenkins/jetty.project/detail/PR-9498/11/pipeline

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
gregw: I think a minimum time to live of one idle timeout will stop any thrashing.

gregw: Actually, I have just re-added the behaviour that when we start a new thread we advance the shrink threshold to now, so no threads will be able to shrink immediately after one has started.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
gregw: Oops, I renamed evict to shrink before I saw this.  I've also fixed the xml and ini files that did not have this name change done in them.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
gregw: I've also fixed the xml and ini file that didn't have the rename done in them.  If you change back to `evict` then don't forget to change them also.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
lorban: @gregw The race condition between the two atomics that may make us shrink below `minThreads` does exist, but it's quite rate if the threads execute any code at all. You can convince yourself by instrumenting the pool with a `LongAdder` counting the CAS failures & running `QueuedThreadPoolBenchmark`. It shows these counters aren't very contended, if at all.

Sure, it can still happen. But it's going to be very uncommon and the worst case scenario is that the pool can shrink a few threads below `minThreads`. Do we really need to enforce an invariant at the cost of extra complexity, even if it's moderate?

magibney: Some extra context -- I wrote a test that's intended to trigger this, worst-case scenario, pool shrinking to `minThreads=100` from `maxThreads=1000` and `maxThreads=10`, ~~haven't been able to reproduce this at all on my system, even running hundreds of iterations~~ does reproduce the issue, but only transiently -- `ensureThreads()` brings restores the `minThreads` count. ~~I get that it's possible in theory, and might be more likely to happen on a system with a massive number of cores?~~ In practice I was only ever able to read a thread count 2 threads under `minThreads`. But even at that, any real system would of course also not be jumping through hoops to intentionally try to trigger this issue, so ... ; anyway, here's the test:

```java
    @Test
    public void testMinThreads() throws Exception
    {
        QueuedThreadPool tp = new QueuedThreadPool();
        int minThreads = 100;
        tp.setMinThreads(minThreads);
        int maxThreads = 1000;
        tp.setMaxThreads(maxThreads);
        int idleTimeout = 1000;
        tp.setIdleTimeout(idleTimeout);
        tp.setMaxEvictCount(Integer.MAX_VALUE); // very aggressive
        tp.start();
        CountDownLatch cdl1 = new CountDownLatch(maxThreads);
        CountDownLatch cdl2 = new CountDownLatch(maxThreads + 1);
        for (int i = 0; i < maxThreads; i++) {
            tp.execute(() -> {
                try {
                    Thread.sleep(idleTimeout); // to ensure all are immediately eligible to shrink
                    cdl1.countDown();
                    cdl2.countDown();
                    cdl2.await();
                } catch (InterruptedException e) {
                    Thread.currentThread().interrupt();
                }
            });
        }
        cdl1.await();
        assertEquals(maxThreads, tp.getThreads());
        cdl2.countDown();
        while (tp.getThreads() > minThreads) {
            // busy-poll, since we expect this to be really fast.
        }
        Thread.sleep(200); // give it time to drop too low, if it's going to do so
        assertEquals(minThreads, tp.getThreads());
    }
```
I have a clear idea of how this would be generalizable for my use case, but for now it's probably easier to consider this PR in isolation, at least once there's a decision on method naming (looks like converging on `shrink()` etc.?) and whether to strictly enforce minThreads ...

magibney: ahhh so this test may have just been picking up the fact that `ensureThreads()` corrected for overshooting `minThreads` floor -- so maybe the test was triggering the situation after all!

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -936,6 +970,70 @@ protected void runJob(Runnable job)
...
+     * <p>Attempts to shrink the current thread from the pool if {@link #getThreads()} is greater than {@link #getMinThreads()}.</p>
+     * @return true if the current thread was shrunk, false otherwise.
+     */
+    protected boolean shrink()
+    {
+        long idleTimeoutNanos = TimeUnit.MILLISECONDS.toNanos(getIdleTimeout());
+        while (true)
+        {
+            // No jobs available, should we shrink?
+            int threads = getThreads();
```
hiteshk25: How about retrying (threads-getMinThreads()) times only?

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -1026,92 +1124,75 @@ public void run()
...
 
-                        idle = false;
-
-                        // run job
-                        if (LOG.isDebugEnabled())
-                            LOG.debug("run {} in {}", job, QueuedThreadPool.this);
-                        runJob(job);
-                        if (LOG.isDebugEnabled())
-                            LOG.debug("ran {} in {}", job, QueuedThreadPool.this);
+                        if (shrink())
```
hiteshk25: Just wanted to raise this, if anyone has opinion on this or something already happening.

https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadPoolExecutor.html
```
Keep-alive times
If the pool currently has more than corePoolSize threads, excess threads will be terminated if they have been idle for more than the
keepAliveTime (see getKeepAliveTime(java.util.concurrent.TimeUnit)). This provides a means of reducing resource consumption when the pool
is not being actively used. If the pool becomes more active later, new threads will be constructed. This parameter can also be changed
dynamically using method setKeepAliveTime(long, java.util.concurrent.TimeUnit). Using a value of Long.MAX_VALUE TimeUnit.NANOSECONDS
effectively disables idle threads from ever terminating prior to shut down. By default, the keep-alive policy applies only when there are
more than corePoolSizeThreads. But method allowCoreThreadTimeOut(boolean) can be used to apply this time-out policy to core threads as
well, so long as the keepAliveTime value is non-zero.
```

Basically, shrink the thread if that is idle for `idletimeout`.  That means, if `job = idleJobPoll(idleTimeout);` returns null, then that thread is eligible for shrink. 

On line `1134`, we have while (job != null) loop, which can return job, job, job, null. After this sequence thread becomes eligible for shrink immediately. That I think we want to avoid it. We should wait for `idletimeout` before we shrink that thread.

Please let me know your opinion on that.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
lorban: This would just trade off a hard CPU consumption limit with a possibility of not shrinking the pool by the configured amount of threads. IMHO the current option is preferable in this case.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
lorban: `shrink()` checks the `shrinkThreshold` timestamp and returns `false` when it is in the future. Every call to `shrink()` moves the `shrinkThreshold` forward a bit, and every call to `startThread()` sets it to one `idleTimeout` period in the future.

This basically means that if the pool decides to grow, then it cannot shrink for a duration at least equal to `idleTimeout`.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
hiteshk25: On line #980,  we have a while loop with true.  Instead of true, we can have this condition. That way we can ensure total thread count doesn't go less than min thread count.

sbordet: The `while` loop is not really a loop, unless the CAS fails.
A thread would enter, and try to move the threshold. If the threshold is already in the future it exits; if it was able to move the threshold, it exits.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -536,6 +537,37 @@ public void setVirtualThreadsExecutor(Executor executor)
...
+     * {@code idleTimeout=5000} and {@code maxShrinkCount=3}.
+     * Let's assume all 20 threads are executing a task, and they all finish their own tasks
+     * at the same time and no more tasks are submitted; then, all 20 will wait for an idle
+     * timeout, after which 3 threads will be exited, while the other 17 will wait another
+     * idle timeout; then another 3 threads will be exited, and so on until {@code minThreads=2}
+     * will be reached.</p>
+     *
+     * @param shrinkCount the maximum number of idle threads to exit in one idle timeout period
+     */
+    public void setMaxShrinkCount(int shrinkCount)
```
sbordet: I think I like better "evict", because we evict threads and as a consequence the pool shrinks, so `maxEvictCount` here and in other places.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -830,7 +863,8 @@ protected void startThread()
             if (LOG.isDebugEnabled())
                 LOG.debug("Starting {}", thread);
             _threads.add(thread);
-            _lastShrink.set(NanoTime.now());
+            // Update the shrink threshold to prevent thrashing
```
sbordet:
```suggestion
            // Update the shrink threshold to prevent thrashing of just started threads.
```

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -936,6 +970,70 @@ protected void runJob(Runnable job)
         job.run();
     }
 
+    /**
+     * <p>Attempts to shrink the current thread from the pool if {@link #getThreads()} is greater than {@link #getMinThreads()}.</p>
+     * @return true if the current thread was shrunk, false otherwise.
+     */
+    protected boolean shrink()
```
sbordet: Rename to `evict()`?

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -936,6 +970,70 @@ protected void runJob(Runnable job)
...
+                        LOG.debug("Shrink skipped, threshold={}ms in future {}", NanoTime.millisElapsed(now, threshold), QueuedThreadPool.this);
+                    return false;
+                }
+
+                // We can shrink if we can update the atomic shrink threshold?
+                if (_shrinkThreshold.compareAndSet(shrinkThreshold, threshold))
+                {
+                    // Yes - we have shrunk, so exit.
+                    if (LOG.isDebugEnabled())
+                        LOG.debug("SHRUNK, threshold={}ms in past {}", NanoTime.millisElapsed(threshold, now), QueuedThreadPool.this);
```
sbordet: Remove the upper case.

```suggestion
                        LOG.debug("Shrunk, threshold={}ms in past {}", NanoTime.millisElapsed(threshold, now), QueuedThreadPool.this);
```

However, if we agree on "evict", these log statements should be changed.

sbordet (commented):
Doing these myself, as I realize I'm the author of this PR.

sbordet:
@gregw please see whether you like the rename shrink -> evict (I think it's more correct), otherwise I think we are good with the algorithm and ready to merge.

lorban (dismissed):

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
hiteshk25: `The while loop is not really a loop, unless the CAS fails.`
This I wanted to cover that when thread retries it should never go below min-threads. 

Hypothetical example:

1. min-thread = 5
2. We can shrink 10 threads in ideltimeout
3. app creates 3 more threads, total threads = 8 now
4.  All 8 threads reach to that loop at same time, and thread count may go below 5. Thus if every thread tries 8-5 = 3 times only, then only 3 thread will succeed.

lorban: The `ensureThreads()` call in the thread's `finally` block would immediately restart any necessary thread to reach `minThread` again.

sbordet: @hiteshk25 I see what you mean now, let me try to code your suggestion.

sbordet: @hiteshk25 I have pushed the changes to implement your suggestion, please re-review.

gregw:
> @gregw please see whether you like the rename shrink -> evict (I think it's more correct), otherwise I think we are good with the algorithm and ready to merge.

I prefer shrink as it is less imperative, i.e we are shrinking to a minimum size. Evict implies that we are kicking out a thread no matter what.

But i don't mind evict enough to hold this up

gregw (approved):

lorban (approved):

hiteshk25 (approved):
LGTM (Thanks Everyone!)

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
sbordet: Note that it is still possible that all 8 threads will be evicted, if none of them fails the CAS.
However, we have `ensureThreads()` in the `finally` block to bring the number of threads back to `minThreads` if that happens.

gregw: More over, if that happens, then we will not shrink the pool for an idleTimeout period, so will will not allow threads to die and then be recreated in a quick cycle.

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -1011,23 +1015,28 @@ protected boolean evict()
...
                 {
                     if (LOG.isDebugEnabled())
-                        LOG.debug("Evicted, threshold={}ms in past {}", NanoTime.millisElapsed(threshold, now), QueuedThreadPool.this);
+                        LOG.debug("Evicted, threshold={}ms in the past {}", NanoTime.millisElapsed(threshold, now), this);
                     return true;
                 }
+                else
+                {
+                    // Some other thread was evicted.
+                    --threadsToEvict;
```
gregw, line 1032: absolutely trivial, but I much prefer `threadsToEvict--`

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
```diff
@@ -977,23 +977,27 @@ protected void runJob(Runnable job)
...
+        // For example when minThreads=3, threads=8, maxEvictCount=10 we want
+        // to evict at most 5 threads (8-3), not 10.
+        // When a thread fails the CAS, it may assume that another thread has
+        // been evicted, so the CAS should be attempted only a number of times
+        // equal to the most threads we want to evict (5 in the example above).
+        int threads = getThreads();
+        int minThreads = getMinThreads();
+        int threadsToEvict = threads - minThreads;
+
         while (true)
```
gregw, line 992: why not just:
```
while(threadsToEvict-- > 0)
```
??

gregw:
@sbordet oh this just got merged as I was making some trivial extra review comments.  No problem... if I care about it enough, I'll make another PR for those trivial clean ups (I probably don't)

gregw:
Thanks @magibney and @hiteshk25 for your suggestions, help, review and patience while we moved slowly on this critical class.

magibney:
Wonderful, thanks again for all your work on this!

jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java
sbordet: Because we want to log when there are no more excess threads, and if we put the condition on the `while` becomes more difficult.

sbordet:
> Thanks @magibney and @hiteshk25 for your suggestions, help, review and patience while we moved slowly on this critical class.

Totally second that! Thank you!

lorban:
> > Thanks @magibney and @hiteshk25 for your suggestions, help, review and patience while we moved slowly on this critical class.
> 
> Totally second that! Thank you!

I agree, you kept us on our toes and it was well worth it. Thanks again!

PR #9503, opened 2023-03-15
Issue #9497 allow jetty:effective-web-xml for jar projects

### LINKED ISSUE #9497: Maven plugin add support for jar projects in `:effective-web-xml`
I'm building an app as a simple jar project, with embedded Jetty included as a dependency. I've found out that Jetty supports the quickstart mode, thus vastly speeding up boot time (and possibly even enabling native mode with GraalVM).

Unfortunately, `mvn jetty:effective-web-xml` fails with `Packaging type [jar] is unsupported` which leads me to believe that only the WAR-type projects are supported at the moment.

However, the jar-type project has a classpath very similar to the war-type project, and so this ticket could perhaps be fixed simply by removing the limitation in the Jetty Maven Plugin.

Thanks again for your great work :)

I'm using Jetty 11.0.14.

janbartel:
Closes #9497 

Fix NPE to allow `jetty:effective-web-xml` to work with non-war projects.

### DIFF
--- a/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyEffectiveWebXml.java
+++ b/jetty-maven-plugin/src/main/java/org/eclipse/jetty/maven/plugin/JettyEffectiveWebXml.java
@@ -63,12 +63,11 @@ public class JettyEffectiveWebXml extends AbstractUnassembledWebAppMojo
             }
         }
 
-        Path start = path.getName(0);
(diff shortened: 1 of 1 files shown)


mvysny:
Thank you so much @janbartel , I've applied the patch, built Jetty from sources and I confirm that the patch works, and the Maven plugin is able to generate `effective-web.xml` properly now.

joakime (approved):

PR #9507, opened 2023-03-16
Jetty 12 - Remove ISO-8859-1 fallback decoding during UTF-8 decoding

### LINKED ISSUE #9489: Jetty 12 - Remove ISO-8859-1 fallback during URIUtil.decodePath(String)

gregw:
This is a partial replacement for #9496, focused on simplification of the Utf8Appendable contract and removal of inconsistencies in it's use.

The `Utf8Appendable` class now does not throw unless in the `takeString` methods, which explicitly throw `CharacterCodingException` or use a passed in `Supplier<X extends Throwable>.

The `toString` method remains like the `StringBuilder` and `StringBuffer` `toString`, in that it just returns the string.  However, it now does not change state and so can be used for debug.  Equivalent to the previous `toPartialString()`

There is now a `complete()` method to handled incomplete sequences and a `toCompleteString()` convenience method. We should probably rename `takeString` to `takeCompleteString` to be consistent.


gregw:
@joakime if this approach proves workable, we need to extract/combine the goodness out of #9496, as I've just been looking at Utf8Appendable here, but have also removed the 8859-1 fallback, but without any rigour.

@lachlan-roberts there are websocket issues as I believe it was relying on fast fail behaviour of append throwing.   I have added some checks for `hasCodingErrors()` to achieve the same result, but obviously more work is required.

gregw:
@joakime what is the JVM decoding mechanism you alluded to last night?  We do have support for a `CharsetDecoder` in `DecoderStringBuilder`, but the API is very inefficient, requiring allocation of little byte arrays and little char sequences on every append.    Is there a better alternative?

lachlan-roberts:
@gregw I fixed all failing tests in websocket-core-tests, and there seemed to be no failures in websocket-jetty-tests ee10.
So I will wait for the CI build and see if there are any other failures.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/UrlEncoded.java
```diff
@@ -329,8 +329,7 @@ private static void decodeUtf8To(String query, int offset, int length, BiConsume
             switch (c)
             {
                 case '&':
-                    value = buffer.toReplacedString();
-                    buffer.reset();
+                    value = buffer.takeString(IllegalArgumentException::new);
```
joakime: Message-less Exceptions are horrible

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/UrlEncoded.java
```diff
@@ -348,8 +347,7 @@ else if (value != null && value.length() > 0)
                         buffer.append(c);
                         break;
                     }
-                    key = buffer.toReplacedString();
-                    buffer.reset();
+                    key = buffer.takeString(IllegalArgumentException::new);
```
joakime: Ditto

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
```diff
@@ -89,19 +89,47 @@ public Utf8Appendable(Appendable appendable)
...
+        return _codingErrors;
+    }
+
+    /**
+     * Reset the appendable, clearing the buffer, resetting decoding state and clearing any errors.
+     */
+    public void reset()
     {
         _state = UTF8_ACCEPT;
+        _codingErrors = false;
```
joakime: active codepoint needs reset too

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/CloseStatus.java
```diff
@@ -168,21 +168,14 @@ public CloseStatus(ByteBuffer payload)
...
-                    return;
-                }
-                catch (Utf8Appendable.NotUtf8Exception e)
-                {
-                    throw new BadPayloadException("Invalid CLOSE Reason", e);
-                }
+                Utf8StringBuilder utf = new Utf8StringBuilder();
+                // if this throws, we know we have bad UTF8
+                utf.append(reasonBytes, 0, reasonBytes.length);
+                String reason = utf.takeString(() -> new BadPayloadException("Invalid UTF8 in CLOSE Reason", new CharacterCodingException()));
```
joakime: Huh?  why add the CharacterCodingException at all?

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/WebSocketCoreSession.java
```diff
@@ -319,7 +319,7 @@ public void processConnectionError(Throwable cause, Callback callback)
         int code;
         if (cause instanceof CloseException)
             code = ((CloseException)cause).getStatusCode();
-        else if (cause instanceof Utf8Appendable.NotUtf8Exception)
+        else if (cause instanceof CharacterCodingException)
```
joakime: Just get rid of this if case entirely.
The new code throwing `BadPayloadException` is enough, as that class extends from `CloseException` already.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/WebSocketCoreSession.java
```diff
@@ -355,7 +355,7 @@ public void processHandlerError(Throwable cause, Callback callback)
             code = ((CloseException)cause).getStatusCode();
         else if (cause instanceof ClosedChannelException)
             code = CloseStatus.NO_CLOSE;
-        else if (cause instanceof Utf8Appendable.NotUtf8Exception)
+        else if (cause instanceof CharacterCodingException)
```
joakime: Just get rid of this if case entirely.
The new code throwing `BadPayloadException` is enough, as that class extends from `CloseException` already.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/MessageHandler.java
```diff
@@ -200,29 +200,29 @@ protected void onTextFrame(Frame frame, Callback callback)
...
                 if (currentSize > maxSize)
                     throw new MessageTooLargeException("Message larger than " + maxSize + " bytes");
-
                 textBuffer.append(frame.getPayload());
             }
 
             if (frame.isFin())
             {
-                onText(textBuffer.toString(), callback);
+                onText(textBuffer.takeString(CharacterCodingException::new), callback);
```
joakime: This should throw `BadPayloadException`, with a message.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/MessageHandler.java
```diff
@@ -200,29 +200,29 @@ protected void onTextFrame(Frame frame, Callback callback)
...
             }
             else
             {
+                if (textBuffer.hasCodingErrors())
+                    throw new CharacterCodingException();
                 callback.succeeded();
             }
         }
-        catch (Utf8Appendable.NotUtf8Exception e)
+        catch (CharacterCodingException e)
```
joakime: Remove this entire catch block (once you change to throwing a `BadPayloadException`)

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/PartialStringMessageSink.java
```diff
@@ -42,12 +43,17 @@ public void accept(Frame frame, Callback callback)
             out.append(frame.getPayload());
             if (frame.isFin())
             {
-                methodHandle.invoke(out.toString(), true);
+                String complete = out.takeString(CharacterCodingException::new);
```
joakime: Throw `BadPayloadException` here.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/PartialStringMessageSink.java
```diff
@@ -42,12 +43,17 @@ public void accept(Frame frame, Callback callback)
...
+                methodHandle.invoke(complete, true);
                 out = null;
             }
             else
             {
-                methodHandle.invoke(out.takePartialString(), false);
+                String partial = out.toString();
+                if (out.hasCodingErrors())
+                    throw new CharacterCodingException();
```
joakime: Throw `BadPayloadException` here.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/StringMessageSink.java
```diff
@@ -50,8 +51,9 @@ public void accept(Frame frame, Callback callback)
 
             out.append(frame.getPayload());
             if (frame.isFin())
-                methodHandle.invoke(out.toString());
-
+            {
+                methodHandle.invoke(out.takeString(CharacterCodingException::new));
```
joakime: Throw `BadPayloadException` here.

jetty-core/jetty-websocket/jetty-websocket-core-tests/src/test/java/org/eclipse/jetty/websocket/core/util/PartialStringMessageSinkTest.java
```diff
@@ -94,8 +95,8 @@ public void testInvalidSingleFrameUtf8() throws Exception
         messageSink.accept(new Frame(OpCode.TEXT, invalidUtf8Payload).setFin(true), callback);
 
         // Callback should fail and we don't receive the message in the sink.
-        RuntimeException error = assertThrows(RuntimeException.class, () -> callback.block(5, TimeUnit.SECONDS));
-        assertThat(error.getCause(), instanceOf(Utf8Appendable.NotUtf8Exception.class));
+        IOException error = assertThrows(IOException.class, () -> callback.block(5, TimeUnit.SECONDS));
+        assertThat(error.getCause(), instanceOf(CharacterCodingException.class));
```
joakime: `instanceOf(BadPayloadException.class)` here.

jetty-core/jetty-websocket/jetty-websocket-core-tests/src/test/java/org/eclipse/jetty/websocket/core/util/PartialStringMessageSinkTest.java
```diff
@@ -114,8 +115,8 @@ public void testInvalidMultiFrameUtf8() throws Exception
         messageSink.accept(new Frame(OpCode.TEXT, continuationUtf8Payload).setFin(true), continuationCallback);
 
         // Callback should fail and we only received the first frame which had no full character.
-        RuntimeException error = assertThrows(RuntimeException.class, () -> continuationCallback.block(5, TimeUnit.SECONDS));
-        assertThat(error.getCause(), instanceOf(Utf8Appendable.NotUtf8Exception.class));
+        IOException error = assertThrows(IOException.class, () -> continuationCallback.block(5, TimeUnit.SECONDS));
+        assertThat(error.getCause(), instanceOf(CharacterCodingException.class));
```
joakime: `instanceOf(BadPayloadException.class)` here.

jetty-core/jetty-websocket/jetty-websocket-core-tests/src/test/java/org/eclipse/jetty/websocket/core/util/StringMessageSinkTest.java
```diff
@@ -98,8 +99,8 @@ public void testInvalidSingleFrameUtf8() throws Exception
         messageSink.accept(new Frame(OpCode.TEXT, invalidUtf8Payload).setFin(true), callback);
 
         // Callback should fail and we don't receive the message in the sink.
-        RuntimeException error = assertThrows(RuntimeException.class, () -> callback.block(5, TimeUnit.SECONDS));
-        assertThat(error.getCause(), instanceOf(Utf8Appendable.NotUtf8Exception.class));
+        IOException error = assertThrows(IOException.class, () -> callback.block(5, TimeUnit.SECONDS));
+        assertThat(error.getCause(), instanceOf(CharacterCodingException.class));
```
joakime: `instanceOf(BadPayloadException.class)` here.

jetty-core/jetty-websocket/jetty-websocket-core-tests/src/test/java/org/eclipse/jetty/websocket/core/util/StringMessageSinkTest.java
```diff
@@ -118,8 +119,8 @@ public void testInvalidMultiFrameUtf8() throws Exception
         messageSink.accept(new Frame(OpCode.TEXT, continuationUtf8Payload).setFin(true), continuationCallback);
 
         // Callback should fail and we don't receive the message in the sink.
-        RuntimeException error = assertThrows(RuntimeException.class, () -> continuationCallback.block(5, TimeUnit.SECONDS));
-        assertThat(error.getCause(), instanceOf(Utf8Appendable.NotUtf8Exception.class));
+        IOException error = assertThrows(IOException.class, () -> continuationCallback.block(5, TimeUnit.SECONDS));
+        assertThat(error.getCause(), instanceOf(CharacterCodingException.class));
```
joakime: `instanceOf(BadPayloadException.class)` here.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
```diff
@@ -89,19 +89,47 @@ public Utf8Appendable(Appendable appendable)
...
 
     private void checkCharAppend() throws IOException
     {
         if (_state != UTF8_ACCEPT)
         {
             _appendable.append(REPLACEMENT);
-            int state = _state;
             _state = UTF8_ACCEPT;
-            throw new NotUtf8Exception("char appended in state " + state);
+            _codingErrors = true;
```
joakime: This method needs work.
This will now add a replacement character, then the caller will add it's own (bad) character.

joakime (changes requested):
First pass

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/UrlEncoded.java
gregw: Agree. Specially for a generic exception type like IAE. CharacterCodingException doesn't take a message, but the name kind of says it.  Might need to ressurect NotUtf8Exception

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/CloseStatus.java
gregw: It was for websocket tests that were checking cause. Happy to change tests and not have double exception, but wanted to start with minimal test changes

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/WebSocketCoreSession.java
gregw: Probably yes.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
gregw: Isn't the replacement character for the incomplete sequence, then the caller adds it's own character, which can't be bad as it is already a character.

joakime: The concern is that it could mean if you repeatedly pass the output back into Utf8Appendable the string grows, which is bad.
Imagine back/forth requests/responses like this.

gregw: But we want it to grow. Somebody had called a method like append(char). We want to spend that char, but if we are incomplete then we append a replacement first.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
joakime: Done.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
joakime: I think you misunderstand what I mean.

Take this example.

``` java
Utf8StringBuilder builder = new Utf8StringBuilder();
builder.append(stringFromRequest);
builder.complete();
String stringUsedInResponse = builder.toString();
```

Imagine that the client takes `stringUsedInResponse` and sends it back as `stringFromRequest` (don't get hung up on where in the request/response).
If that round trip makes the string grow every time then we have a bug.
The replacement character behavior should result in no growth of the string length when it is passed through the algorithm again.

gregw: if you don't take or reset the builder, then it continues to grow.  Nothing new here???
So I think I do misunderstand what you mean.

gregw:
@joakime thanks for the update.  However, I did revert a few minor changes, primarily to reduce needless differences with the original.  I believe I have kept all your behaviours.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
joakime: Imagine a string, supplied by an HTTP client, initially passed into Jetty as bad, then "sanitized" (with replacement characters).

We give that string back to the client (for whatever reason), in the "sanitized" form.

That same client gives us back that "sanitized" string in a different request.

That's the loop, a new Utf8StringBuilder each time, but the string on output is used again on a new Utf8StringBuilder (in append + toString)
That loop shouldn't result in a different String, or a longer String, as the replacement character should have sanitized it.
If each iteration through Utf8StringBuilder results in a different string, we have a bug.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
```diff
@@ -305,20 +299,20 @@ public void appendByte(byte b) throws IOException
                         }
                     }
                     _codep = 0;
-                    _state = current;
+                    _state = next;
```
joakime: BTW, this code is from http://bjoern.hoehrmann.de/utf-8/decoder/dfa/ (see javadoc)

The name "current" is the name used in the original code.
As is the decode() method.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
```diff
@@ -69,7 +70,11 @@ public enum Violation implements ComplianceViolation
         /**
          * Allow UTF-16 encoding eg <code>/foo%u2192bar</code>.
          */
-        UTF16_ENCODINGS("https://www.w3.org/International/iri-edit/draft-duerst-iri.html#anchor29", "UTF16 encoding");
+        UTF16_ENCODINGS("https://www.w3.org/International/iri-edit/draft-duerst-iri.html#anchor29", "UTF16 encoding"),
+        /**
+         * Allow Bad UTF-8 encodings to be substituted by the replacement character.</code>.
```
joakime: This trailing `</code>` is breaking Javadoc

joakime (dismissed):
Javadoc portion of build is broken

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Utf8Appendable.java
gregw: If we produce a decoded string, then validly encoder it and give it to the client, who then send it back to us, then this method will not add an extra replacement character because a) we don't add characters; b) it will now be well formed so a character will not be added when there is an incomplete code, this a replacement character will not be added.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/PartialStringMessageSink.java
```diff
@@ -42,12 +43,17 @@ public void accept(Frame frame, Callback callback)
...
                 out = null;
             }
             else
             {
-                methodHandle.invoke(out.takePartialString(), false);
+                String partial = out.toString();
+                if (out.hasCodingErrors())
+                    throw new BadPayloadException("Invalid UTF-8");
+                out.partialReset();
+                methodHandle.invoke(partial, false);
```
lachlan-roberts, line 53: This API seems weird that we don't have the same style for taking full and partial strings. 

Can't we add a method `takePartialString` in same style as `takeString` and avoid having these external `hasCodingError()` and `partialReset()` calls every time it is used.

```
out.takePartialString(() -> new BadPayloadException("Invalid UTF-8"));
```

gregw: @lachlan-roberts it is a little complicated with the naming.
`takeString` should really be called `takeCompleteString`, but while that name makes sense for some character codes, it is not so much for things like iso-8859-1.  But then there are other multi-character codes served by `CharSequenceBuilder` interface, so perhaps doing that rename makes sense.

We could then re-introduce a `takeString` API that takes the partial string and matches to `toString` semantic.

@joakime @lachlan-roberts  thoughts?

lachlan-roberts: I think its quite clear to have
- `takeString(Supplier<X>)`
- `takePartialString(Supplier<X>)`
- `toString()`

I'm okay with the rename of `takeString` -> `takeCompleteString`, but think it is unnecessary.
I think renaming `takePartialString` -> `takeString` is going to be very confusing.

joakime: `takeString()` is a different API that `Utf8Appendable` implements via the `CharsetStringBuilder` interface, that name is already defined for other purposes.
It has to be a completed String.

gregw:
If we reintroduce partial string in the naming, then it strange to also have toString be partial.

But toString can't be complete because it is a state change.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/PartialStringMessageSink.java
joakime: The fact that `toString()` is a partial String currently in this PR is actually tricky.  The overwhelming majority of uses of `toString()` are now bad and should be changed to `toCompleteString()`.

See commit f8b4884cfec8891d75ccd9aa642216cc1ab2ef3c here that fixes this bug.
(Submitted as a single commit to make it easier to revert)

joakime:
> If we reintroduce partial string in the naming, then it strange to also have toString be partial.
> 
> But toString can't be complete because it is a state change.

Full circle back to what I had in the previous PR?

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/PartialStringMessageSink.java
gregw: How many of those `toCompleteString` calls should really be `takeString`, except that they don't want a checked exception.... which can be achieved with the version that takes a supplier.

So leaning back towards a toString that gives debugging info.

Let me ponder some more on my next train.

gregw:
Whilst playing with the API i realized that we no longer need Utf8StringBuffer.  So i got rid of it and collapsed Utf8Appendable into Utf8StringBuilder, which allowed me to get rid of all the needless IOException catches.

However, i note have a problem with the Null version, so I've got a bit of work to do. Might take me until tomorrow.

joakime:
> Whilst playing with the API i realized that we no longer need Utf8StringBuffer. So i got rid of it and collapsed Utf8Appendable into Utf8StringBuilder, which allowed me to get rid of all the needless IOException catches.

Yay!

joakime:
> However, i note have a problem with the Null version, so I've got a bit of work to do. Might take me until tomorrow.

The `NullAppender` is used by the `ValidationExtension`.

We we need need either of these anymore?
We do `takeString(() -> new BadPayloadException())` (and similar new Utf8Appendable APIs) in the websocket codebase later now.

gregw:
@lachlan-roberts what @joakime said?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/CharsetStringBuilder.java
```diff
@@ -128,6 +176,42 @@ public void append(byte b)
...
+            try
+            {
+                if (_stringBuilder == null)
+                    _stringBuilder = new StringBuilder();
+                _stringBuilder.append(_decoder.decode(_buffer.flip()));
+                _buffer.clear();
+            }
+            catch (CharacterCodingException e)
+            {
+                // Should do replacement and remember issue.
```
joakime: This is the responsibility of the `CharsetDecoder` and how it's configured, we should honor its `onMalformedInput` configuration.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/CharsetStringBuilder.java
```diff
@@ -128,6 +176,42 @@ public void append(byte b)
...
```
joakime: This is the responsibility of the `CharsetDecoder` and how it's configured, we should honor its `onMalformedInput` configuration.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/CharsetStringBuilder.java
gregw: Yep, didn't have time to work it out exactly. But yes, the decoder should do the replacement, not us.

gregw: The behaviour was correct, so I have just updated the comment.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/CharsetStringBuilder.java
gregw: The behaviour was correct, so I have just updated the comment.

gregw:
@joakime I have updated this PR so that when using `CharsetStringBuilder`, all the various implementations act the same (i.e. throw `CharacterCodingException` in the `takeString` method in a non-fast fail style).  I've also improved the javadoc to explain why the class exists.

The `Utf8StringBuilder` class should be used directly when we know it is UTF-8 and we want either fast fail or replacement behaviour.

gregw:
@olamy This PR is the same - it is like CI is running the new test against the old code?

olamy:
> @olamy This PR is the same - it is like CI is running the new test against the old code?

not sure to understand the problem neither sure to understand which build you are referring to?
Looking at history of the build and commits before and after your comment

before your comment the build here https://jenkins.webtide.net/blue/organizations/jenkins/jetty.project/detail/PR-9507/16/tests is corresponding to this commit (https://github.com/eclipse/jetty.project/pull/9507/commits/850b65761885bfee8c86737ee923e41eb15ccb3d) which have test failures

before your comment the build here https://jenkins.webtide.net/blue/organizations/jenkins/jetty.project/detail/PR-9507/17/tests  is corresponding to this commit (https://github.com/eclipse/jetty.project/pull/9507/commits/bc9382122fe30c129bc12c9eaf625a64d9ce0b9b) which only change some javadoc

after your comment the build here https://jenkins.webtide.net/blue/organizations/jenkins/jetty.project/detail/PR-9507/18/pipeline which turn green is corresponding to this commit https://github.com/eclipse/jetty.project/pull/9507/commits/b307e6223bf493df378c3200ea69bfe64cd95bbb
last changes seems to fix some code and fix the build?

gregw:
@olamy yeah wrong PR sorry.  It is the session listener one that is causing me grief.

gregw:
@joakime @lachlan-roberts OK green build finally.  Can I get reviews please.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
```diff
@@ -1434,12 +1434,19 @@ else if (_path != null)
...
                 // The RFC requires this to be canonical before decoding, but this can leave dot segments and dot dot segments
                 // which are not canonicalized and could be used in an attempt to bypass security checks.
-                _canonicalPath = URIUtil.canonicalPath(_path);
+                _canonicalPath = URIUtil.canonicalPath(_path, this::onBadUtf8);
                 if (_canonicalPath == null)
                     throw new IllegalArgumentException("Bad URI");
             }
         }
 
+        private RuntimeException onBadUtf8()
```
joakime, line 1443: Why is this returning anything at all?
It always returns null below.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
```diff
@@ -69,7 +70,11 @@ public enum Violation implements ComplianceViolation
         /**
          * Allow UTF-16 encoding eg <code>/foo%u2192bar</code>.
          */
-        UTF16_ENCODINGS("https://www.w3.org/International/iri-edit/draft-duerst-iri.html#anchor29", "UTF16 encoding");
+        UTF16_ENCODINGS("https://www.w3.org/International/iri-edit/draft-duerst-iri.html#anchor29", "UTF16 encoding"),
```
joakime: The hyphen matters.

s/UTF16/UTF-16/

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java
```diff
@@ -69,7 +70,11 @@ public enum Violation implements ComplianceViolation
...
+        /**
+         * Allow Bad UTF-8 encodings to be substituted by the replacement character.
+         */
+        BAD_UTF8_ENCODING("https://datatracker.ietf.org/doc/html/rfc5987#section-3.2.1", "Bad UTF8 encoding");
```
joakime: The hyphen matters.

s/UTF8/UTF-8/

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/UrlEncoded.java
```diff
@@ -873,7 +872,14 @@ else if (buffer != null)
...
 
-            return buffer.toString();
+            try
+            {
+                return buffer.takeString();
+            }
+            catch (CharacterCodingException e)
+            {
+                throw new RuntimeException(e);
+            }
```
joakime:
```suggestion
            return buffer.takeString(() -> new RuntimeException("Bad UTF-8"));
```

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/FileSystemResourceTest.java
```diff
@@ -183,6 +184,7 @@ public void testNonAbsoluteURI(WorkDir workDir) throws Exception
     }
 
     @Test
+    @Tag("flaky")
     public void testNotFileURI()
```
joakime, line 188: Is this flakiness filed as an issue?

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/thread/QueuedThreadPoolTest.java
```diff
@@ -317,6 +318,7 @@ public void testThreadPool() throws Exception
     }
 
     @Test
+    @Tag("flaky")
     public void testThreadPoolFailingJobs() throws Exception
```
joakime, line 322: Is this flakiness filed as an issue somewhere?

joakime (commented):
Overall, I like it.
Just some nits and questions.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java
gregw: Because it is being passed in as a `Supplier<X extends Throwable> onBadUtf8` to `URIUtil.canonicalPath`.

In this case we don't want to throw, so we supply null, but we remember that there was a violation

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/thread/QueuedThreadPoolTest.java
gregw: No.  I assume that we will periodically search for flaky tags and work to fix them all.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/FileSystemResourceTest.java
gregw: Ditto

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/UrlEncoded.java
gregw: Currently the API taking a `Supplier<X extends Throwable>` is only on the `Utf8StringBuilder` interface and not on  `CharsetStringBuilder`.  We don't have it there because we have some implementations (i.e. the JVM Decoder ones) that are hard wired to throw `CharacterCodingException`, so we cannot implement an arbitrary supplier.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/CharsetStringBuilder.java
```diff
@@ -40,13 +53,30 @@ default void append(byte[] b, int offset, int length)
...
     default void append(ByteBuffer buf)
     {
         int end = buf.position() + buf.remaining();
         while (buf.position() < end)
             append(buf.get());
     }
 
+    /**
+     * @return The decoded built string.
+     * @throws RuntimeException if there is a {@link CharacterCodingException}.
```
lachlan-roberts: Whats the point of this, why not just throw `CharacterCodingException` if there is a `CharacterCodingException`? And why doesn't `takeString` do the same?

Why is this called `toCompleteString` and the one below is not called `takeCompleteString`?

gregw: This method is actually no longer needed in this interface.  so I have removed it.

There is a slight naming clash between the `CharsetStringBuilding` and `Utf8StringBuilder`.    The later already has a `takeCompleteString()` method that does not throw..... hmm let me see if I can resolve that.....

gregw: OK, I renamed `takeString()` to `build()` as this is a builder after all.

joakime (approved):

PR #9510, opened 2023-03-16
Issue #9509 Preserve query string on dispatch()

### LINKED ISSUE #9509: jetty-12 ee10 com.sun.ts.tests.servlet.api.jakarta_servlet.asynccontext.URLClient.forwardTest1 tck failure
Tck test does:

> Client send a request to AsyncTestServlet at "/AsyncTestServlet?testname=forwardTest1"; getRequestDispatcher("/AsyncTestServlet?testname=forwardDummy1").forward( request, response);
>  In forwardDummy1: AsyncContext ac = request.startAsync(request, response); ac.dispatch(); verifies that it dispatches to "/AsyncTestServlet?testname=forwardDummy1

We do not preserve the query string on the `dispatch`, which means that the wrong test case executes.

janbartel:
Closes #9509


jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/Dispatcher.java
```diff
@@ -270,7 +270,7 @@ public String getQueryString()
         @Override
         public String getRequestURI()
         {
-            return _uri == null ? super.getRequestURI() : _uri.getPath();
+            return _uri == null ? super.getRequestURI() : _uri.getPathQuery();
```
gregw: This is the wrong fix (sorry I know I suggested it).  `getRequestURI` should not include the query string.

gregw (dismissed):

gregw:
I've pushed an alternative fix with a rough test.   Needs more thought about caching result and probably doing the same for include, so I have converted to a draft.

gregw:
@janbartel we need to chat about this one.

gregw:
@lachlan-roberts can you take a close look at this one?

gregw (approved):
I approve, but I've written much of the code.  @lachlan-roberts @janbartel Can I get a second opinion

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/Dispatcher.java
```diff
@@ -569,6 +541,12 @@ public String getQueryString()
             return _httpServletRequest.getQueryString();
         }
 
+        @Override
+        public String[] getParameterValues(String name)
+        {
+            return super.getParameterValues(name);
```
lachlan-roberts: Maybe get rid of this one, does nothing except call super.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/Dispatcher.java
```diff
@@ -153,38 +148,29 @@ public void async(ServletRequest request, ServletResponse response) throws Servl
...
 
         private MultiMap<String> getParams()
         {
-            if (decodedParams)
-                return _params;
-            decodedParams = true;
-
-            Fields contentParams = _request.getServletApiRequest().getContentParams();
-            for (Fields.Field field : contentParams)
+            if (_params == null)
```
lachlan-roberts, line 160: Since this is done lazily, do we need to consider if two threads call getParameter at same time? or maybe thats not an issue.

Why should this not be final and done in the constructor?

lachlan-roberts (approved):
The fix looks good,  the `ServletContextRequest.getServletContextRequest(request)` was bypassing any wrappings on the `HttpServletRequest` and now we used the wrapped request.

So approved, but please address/respond to the comments above.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/Dispatcher.java
gregw: Because if the target of the dispatch never calls any parameter methods, then there is no need to build the merged params.  In fact there is an optimisation available if the target has no query parameters... but leaving that for now.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/Dispatcher.java
```diff
@@ -569,6 +541,12 @@ public String getQueryString()
             return _httpServletRequest.getQueryString();
         }
 
+        @Override
+        public String[] getParameterValues(String name)
+        {
+            return super.getParameterValues(name);
+        }
+
```
gregw:
```suggestion
```

PR #9516, opened 2023-03-17
Jetty 12 - Remove CGI servlet

joakime:
Removing old CGI servlet support.

We have FastCGI servlet that is better maintained.


sbordet (approved):

sbordet (changes requested):
I think you missed a reference to `CGI` in `jetty-ee10-demo-jetty-webapp`'s `web.xml`.
Same for ee9.

sbordet (approved):

PR #9518, opened 2023-03-17
Fixing #9517 - bad url-pattern prefix match behavior

### LINKED ISSUE #9517: Jetty 10.0.14 uses wrong pathSpec for request
**Jetty version(s)**
10.0.14

17

**OS type/version**
Doesn't matter

**Description**

Related to #9055

We have two servlets:
1) "/*"
2) "/foo/*"

Prior to Jetty 10.0.14, requests to "/foobar/something" were handled by servlet 1. After the upgrade, requests are unable to reach it. 

After debugging I found that problem was related to org.eclipse.jetty.http.pathmap.PathMappings#getMatched method:

```
        // path = "/foobar/something"

        MappedResource<E> prefix = _prefixMap.getBest(path); // ServletPathSpec@1c0d1eaa{/foo/*}

        if (prefix != null)
        {
            MatchedPath matchedPath = prefix.getPathSpec().matched(path);     // null, since found prefix doesn't match path
            if (matchedPath != null)
                return new MatchedResource<>(prefix.getResource(), prefix.getPathSpec(), matchedPath);
        }
```

joakime:
New testcase to ensure we don't have a regression on this behavior.
Restore a prefix matching loop that we previously had in Jetty 10.0.13

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
@@ -204,19 +204,14 @@ public class PathMappings<E> implements Iterable<MappedResource<E>>, Dumpable
             return exact.getPreMatched();
 
         // Try a prefix match
-        if (!_prefixMap.isEmpty())
+        MappedResource<E> prefix = _prefixMap.getBest(path);
+        while (prefix != null)
         {
-            int i = path.length();
-            while (i >= 0)
-            {
-                MappedResource<E> candidate = _prefixMap.getBest(path, 0, i--);
-                if (candidate == null)
-                    continue;
-
-                MatchedPath matchedPath = candidate.getPathSpec().matched(path);
-                if (matchedPath != null)
-                    return new MatchedResource<>(candidate.getResource(), candidate.getPathSpec(), matchedPath);
-            }
+            MatchedPath matchedPath = prefix.getPathSpec().matched(path);
+            if (matchedPath != null)
+                return new MatchedResource<>(prefix.getResource(), prefix.getPathSpec(), matchedPath);
+            int specLength = prefix.getPathSpec().getSpecLength();
+            prefix = specLength > 3 ? _prefixMap.getBest(path, 0, specLength - 3) : null;
         }
 
         // Try a suffix match
@@ -293,12 +288,14 @@ public class PathMappings<E> implements Iterable<MappedResource<E>>, Dumpable
                     {
                         if (_optimizedPrefix)
                         {
-                            MappedResource<E> candidate = _prefixMap.getBest(path);
-                            if (candidate != null)
+                            MappedResource<E> prefix = _prefixMap.getBest(path);
+                            while (prefix != null)
                             {
-                                matchedPath = candidate.getPathSpec().matched(path);
+                                matchedPath = prefix.getPathSpec().matched(path);
                                 if (matchedPath != null)
-                                    return new MatchedResource<>(candidate.getResource(), candidate.getPathSpec(), matchedPath);
+                                    return new MatchedResource<>(prefix.getResource(), prefix.getPathSpec(), matchedPath);
+                                int specLength = prefix.getPathSpec().getSpecLength();
+                                prefix = specLength > 3 ? _prefixMap.getBest(path, 0, specLength - 3) : null;
                             }
 
                             // If we reached here, there's NO optimized PREFIX Match possible, skip simple match below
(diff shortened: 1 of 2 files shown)


jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -223,13 +230,13 @@ public MatchedResource<E> getMatched(String path)
             //  Loop 3: "foo"
             while ((i = path.indexOf('.', i + 1)) > 0)
             {
-                prefix = _suffixMap.get(path, i + 1, path.length() - i - 1);
-                if (prefix == null)
+                MappedResource<E> suffix = _suffixMap.get(path, i + 1, path.length() - i - 1);
```
joakime, line 231: Just renamed "prefix" here to "suffix" to be more clear about usage.

gregw:
@joakime I pushed an update to the loop that uses the length of the last match to improve the search for a shorter match.  Much better than searching on every string length.

However, I don't like the magic number 3 (which is the "/*" minus 1 more), but I think it works.  Could perhaps have more testing of 1 char patterns?

Also there was another optimized prefix match that needs same fix.

gregw (dismissed):

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -205,11 +205,13 @@ public MatchedResource<E> getMatched(String path)
 
         // Try a prefix match
         MappedResource<E> prefix = _prefixMap.getBest(path);
-        if (prefix != null)
+        while (prefix != null)
         {
             MatchedPath matchedPath = prefix.getPathSpec().matched(path);
             if (matchedPath != null)
                 return new MatchedResource<>(prefix.getResource(), prefix.getPathSpec(), matchedPath);
+            int specLength = prefix.getPathSpec().getSpecLength();
```
sbordet: Put `prefix.getPathSpec()` in a local variable.

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -205,11 +205,13 @@ public MatchedResource<E> getMatched(String path)
...
+            prefix = specLength > 3 ? _prefixMap.getBest(path, 0, specLength - 3) : null;
```
sbordet: What's the magic `3`?

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -286,12 +288,14 @@ private MatchedResource<E> getMatchedIteratively(String path)
...
+                            MappedResource<E> prefix = _prefixMap.getBest(path);
+                            while (prefix != null)
                             {
-                                matchedPath = candidate.getPathSpec().matched(path);
+                                matchedPath = prefix.getPathSpec().matched(path);
                                 if (matchedPath != null)
-                                    return new MatchedResource<>(candidate.getResource(), candidate.getPathSpec(), matchedPath);
+                                    return new MatchedResource<>(prefix.getResource(), prefix.getPathSpec(), matchedPath);
+                                int specLength = prefix.getPathSpec().getSpecLength();
+                                prefix = specLength > 3 ? _prefixMap.getBest(path, 0, specLength - 3) : null;
```
sbordet: Magic `3`?

jetty-http/src/main/java/org/eclipse/jetty/http/pathmap/PathMappings.java
```diff
@@ -286,12 +288,14 @@ private MatchedResource<E> getMatchedIteratively(String path)
...
-                            if (candidate != null)
+                            MappedResource<E> prefix = _prefixMap.getBest(path);
+                            while (prefix != null)
                             {
-                                matchedPath = candidate.getPathSpec().matched(path);
+                                matchedPath = prefix.getPathSpec().matched(path);
                                 if (matchedPath != null)
-                                    return new MatchedResource<>(candidate.getResource(), candidate.getPathSpec(), matchedPath);
+                                    return new MatchedResource<>(prefix.getResource(), prefix.getPathSpec(), matchedPath);
+                                int specLength = prefix.getPathSpec().getSpecLength();
```
sbordet: Put `prefix.getPathSpec()` in a local variable.

jetty-http/src/test/java/org/eclipse/jetty/http/pathmap/PathMappingsTest.java
```diff
@@ -92,6 +92,42 @@ public void testServletMatchDefault()
...
+        p.put(new ServletPathSpec("/*"), "any");
+        p.put(new ServletPathSpec("/foo/*"), "foo");
+        p.put(new ServletPathSpec("/a/*"), "a");
+        p.put(new ServletPathSpec("/a/b/*"), "ab");
+
+        assertMatch(p, "/abs/path", "any");
+        assertMatch(p, "/abs/foo/bar", "any");
+        assertMatch(p, "/foo/bar", "foo");
+        assertMatch(p, "/", "any");
+        assertMatch(p, "/foobar", "any");
```
sbordet, line 115: Can you add here `/fo` and `/foob`?

jetty-http/src/test/java/org/eclipse/jetty/http/pathmap/PathMappingsTest.java
```diff
@@ -92,6 +92,42 @@ public void testServletMatchDefault()
...
+        assertMatch(p, "/a/", "a");
+        assertMatch(p, "/a", "a");
+
+        // Try now with order important
+        p.put(new RegexPathSpec("/other.*/"), "other");
+        assertMatch(p, "/abs/path", "any");
+        assertMatch(p, "/abs/foo/bar", "any");
+        assertMatch(p, "/foo/bar", "foo");
+        assertMatch(p, "/", "any");
+        assertMatch(p, "/foobar", "any");
```
sbordet, line 133: Ditto

sbordet (changes requested):

sbordet (approved):

Issue #9464, opened 2023-03-04
Respect expiry time of ID token


andrewmcguinness:
**Jetty version(s)**
_[Jetty 9.x is now at End of Community Support](https://github.com/eclipse/jetty.project/issues/7958)_
10.0.14
**Enhancement Description**

The OpenIdAuthenticator goes through the Authorization Code Flow and obtains the ID Token, which it keeps in the session and in the Authentication object. It validates the expiry time of the ID Token when it receives it.

From that point on, the session is authenticated, and remains so for as long as it asks. By default, the only mechanisms for ending the session are an explicit logout, the idle timeout of the session handler, and the cookie max age.

Since there isn't an Access Token being passed around, one reason for having a limited lifetime is not really applicable.  However, if a user has permissions removed or withdrawn in the ID Provider, Jetty will not be aware of it for the lifetime of the session. Note that cookie max age can be ignored by the Browser (and doing so is a feature of at least one browser).

I think there should be an option to have the OpenId authentication remain valid only for the valid lifetime of the ID Token. This would be implemented in OpenIdLoginService.validate().  That should mean that after expiry of the ID Token, the OpenIdAuthenticator.validateRequest() would fall through and repeat the Authorization Code Flow, which is correct. Normally this would be transparent to the user, if they still have a valid session with the OpenId Provider, but if their rights were changed or their OpenId Provider session invalidated in the meantime, they would have to log in again.

Possibly the check should be implemented not in OpenIdLoginService, but in a separate LoginService provided to the OpenIdLoginService in its constructor.

It could also be argued that the check belongs in 'whatever is accessing Session.getAttribute("org.eclipse.jetty.security.openid.claims")' since that includes the "exp". But that would make it awkward to seamlessly reauthenticate the session.

There's also a possibility of using the Refresh Token supplied in the initial response from the OpenId provider, to fetch a new token. On reflection, I don't think that's the right approach since it's for "offline_access", whereas Jetty is really extending an active online session, and is linked to the Authorization Token, which Jetty isn't using. But it is an option.

lachlan-roberts:
There is now a setting in `openid.mod` for this called `jetty.openid.logoutWhenIdTokenIsExpired`.
This is also available to be set with `OpenIdConfiguration.setLogoutWhenIdTokenIsExpired()`.

PR #9405, opened 2023-02-22
Jetty 12.0.x core security

gregw:
This is a core security mechanism, which will ultimately be the base of each of the EEn security handlers.

PR #9528, opened 2023-03-23
Issue #9464 - Add optional configuration to log user out after OpenID idToken expires. (Jetty-10)

lachlan-roberts:
## Issue #9464

Add new configuration parameter called `respectIdTokenExpiry` which if set to true will log the user out after the idToken has expired and attempt to redirect back to the OpenID provider to be re-authenticated.

### DIFF
--- a/jetty-openid/src/main/config/etc/jetty-openid.xml
+++ b/jetty-openid/src/main/config/etc/jetty-openid.xml
@@ -38,6 +38,9 @@
         <Set name="authenticateNewUsers">
           <Property name="jetty.openid.authenticateNewUsers" default="false"/>
         </Set>
+        <Set name="respectIdTokenExpiry">
+          <Property name="jetty.openid.respectIdTokenExpiry" default="false"/>
+        </Set>
         <Call name="addScopes">
           <Arg>
             <Call class="org.eclipse.jetty.util.StringUtil" name="csvSplit">
--- a/jetty-openid/src/main/config/etc/jetty-openid.xml
+++ b/jetty-openid/src/main/config/etc/jetty-openid.xml
@@ -38,8 +38,8 @@
         <Set name="authenticateNewUsers">
           <Property name="jetty.openid.authenticateNewUsers" default="false"/>
         </Set>
-        <Set name="respectIdTokenExpiry">
-          <Property name="jetty.openid.respectIdTokenExpiry" default="false"/>
+        <Set name="logoutWhenIdTokenIsExpired">
+          <Property name="jetty.openid.logoutWhenIdTokenIsExpired" default="false"/>
         </Set>
         <Call name="addScopes">
           <Arg>
--- a/jetty-openid/src/main/config/modules/openid.mod
+++ b/jetty-openid/src/main/config/modules/openid.mod
@@ -45,3 +45,6 @@ etc/jetty-openid.xml
 
 ## What authentication method to use with the Token Endpoint (client_secret_post, client_secret_basic).
 # jetty.openid.authMethod=client_secret_post
+
+## Whether the user should be logged out after the idToken expiry.
+# jetty.openid.respectIdTokenExpiry=false
--- a/jetty-openid/src/main/config/modules/openid.mod
+++ b/jetty-openid/src/main/config/modules/openid.mod
@@ -46,5 +46,5 @@ etc/jetty-openid.xml
 ## What authentication method to use with the Token Endpoint (client_secret_post, client_secret_basic).
 # jetty.openid.authMethod=client_secret_post
 
-## Whether the user should be logged out after the idToken expiry.
+## Whether the user should be logged out after the idToken expires.
 # jetty.openid.logoutWhenIdTokenIsExpired=false
--- a/jetty-openid/src/main/config/modules/openid.mod
+++ b/jetty-openid/src/main/config/modules/openid.mod
@@ -47,4 +47,4 @@ etc/jetty-openid.xml
 # jetty.openid.authMethod=client_secret_post
 
 ## Whether the user should be logged out after the idToken expiry.
-# jetty.openid.respectIdTokenExpiry=false
+# jetty.openid.logoutWhenIdTokenIsExpired=false
--- a/jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
+++ b/jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
@@ -377,11 +377,7 @@ public class OpenIdAuthenticator extends LoginAuthenticator
         {
             Map<String, Object> claims = (Map)session.getAttribute(CLAIMS);
             if (claims != null)
-            {
-                long expiry = (Long)claims.get("exp");
-                long currentTimeSeconds = System.currentTimeMillis() / 1000;
-                return (currentTimeSeconds > expiry);
-            }
+                return OpenIdCredentials.checkExpiry(claims);
         }
         return false;
     }
@@ -404,16 +400,17 @@ public class OpenIdAuthenticator extends LoginAuthenticator
         HttpSession session = request.getSession(false);
         if (_openIdConfiguration.isRespectIdTokenExpiry() && hasExpiredIdToken(session))
         {
+            // After logout, fall through to the code below and send another login challenge.
             logoutWithoutRedirect(request);
-        }
-        else
-        {
+
             // If we expired a valid authentication we do not want to defer authentication,
             // we want to try re-authenticate the user.
-            mandatory |= isJSecurityCheck(uri);
-            if (!mandatory)
-                return new DeferredAuthentication(this);
+            mandatory = true;
         }
+
+        mandatory |= isJSecurityCheck(uri);
+        if (!mandatory)
+            return new DeferredAuthentication(this);
 
         if (isErrorPage(baseRequest.getPathInContext()) && !DeferredAuthentication.isDeferred(response))
             return new DeferredAuthentication(this);
--- a/jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
+++ b/jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
@@ -464,10 +464,8 @@ public class OpenIdAuthenticator extends LoginAuthenticator
                 {
                     if (LOG.isDebugEnabled())
                         LOG.debug("auth revoked {}", authentication);
-                    synchronized (session)
-                    {
-                        session.removeAttribute(SessionAuthentication.__J_AUTHENTICATED);
-                    }
+                    logout(req);
+                    return Authentication.SEND_CONTINUE;
                 }
                 else
                 {
@@ -499,10 +497,10 @@ public class OpenIdAuthenticator extends LoginAuthenticator
                             }
                         }
                     }
+                    if (LOG.isDebugEnabled())
+                        LOG.debug("auth {}", authentication);
+                    return authentication;
                 }
-                if (LOG.isDebugEnabled())
-                    LOG.debug("auth {}", authentication);
-                return authentication;
             }
 
             // If we can't send challenge.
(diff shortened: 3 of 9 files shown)


jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
```diff
@@ -381,17 +401,29 @@ public Authentication validateRequest(ServletRequest req, ServletResponse res, b
         if (uri == null)
             uri = URIUtil.SLASH;
 
-        mandatory |= isJSecurityCheck(uri);
-        if (!mandatory)
-            return new DeferredAuthentication(this);
+        HttpSession session = request.getSession(false);
+        if (_openIdConfiguration.isRespectIdTokenExpiry() && hasExpiredIdToken(session))
```
gregw: this reads better:
```suggestion
        if (hasExpiredIdToken(session) && _openIdConfiguration.isRespectIdTokenExpiry())
```

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
```diff
@@ -381,17 +401,29 @@ public Authentication validateRequest(ServletRequest req, ServletResponse res, b
...
+        {
+            logoutWithoutRedirect(request);
```
gregw, line 404: add comment to say what will happen... i.e they will be logged out and the fall through to code below that will offer a challenge.  Hence they will be redirected to the login service rather than the configured logout page. right???

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
```diff
@@ -381,17 +401,29 @@ public Authentication validateRequest(ServletRequest req, ServletResponse res, b
...
+        }
+        else
+        {
+            // If we expired a valid authentication we do not want to defer authentication,
+            // we want to try re-authenticate the user.
```
gregw, line 407: this is kind of confusing as it is in the branch that will be called even if there is not an expired token.
We get here if we don't have an expired token or we are not "respecting" an expired token - i.e ignoring it.
So I'd expect the comment to be about normal processing

gregw (commented):
I've not reviewed deeply, but I think the core logic change could be better described in comments and perhaps a slight restructure of the if/else

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdCredentials.java
```diff
@@ -137,12 +137,21 @@ private void validateClaims(OpenIdConfiguration configuration) throws Exception
...
 
+    public boolean isExpired()
+    {
+        if (claims == null)
+            return true;
+
+        // Check that the ID token has not expired by checking the exp claim.
+        long expiry = (Long)claims.get("exp");
+        long currentTimeSeconds = System.currentTimeMillis() / 1000;
+        return (currentTimeSeconds > expiry);
```
sbordet: This logic is duplicated in `OpenIdAuthenticator.hasExpiredIdToken()`.
Can it be factored out in a common place?

jetty-openid/src/main/config/etc/jetty-openid.xml
```diff
@@ -38,6 +38,9 @@
         <Set name="authenticateNewUsers">
           <Property name="jetty.openid.authenticateNewUsers" default="false"/>
         </Set>
+        <Set name="respectIdTokenExpiry">
```
sbordet: The word "respect" does not really convey much.
Can this be renamed to `logoutWhenIdTokenIsExpired`?

sbordet (changes requested):
@lachlan-roberts I have concerns about this change. Let's discuss them when possible.

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
lachlan-roberts: Maybe, but its much cheaper to check the boolean than getting the claims from the session and verifying them.
So I think the current order is better.

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdCredentials.java
lachlan-roberts: I could put it in a static method, don't think there is a better way to do it.

In `OpenIdAuthenticator` we don't have the credentials at the point we check expiry.

jetty-openid/src/main/config/etc/jetty-openid.xml
lachlan-roberts: sure

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
lachlan-roberts: I refactored this section to make it clearer. Take a look and let me know if you think this is better.

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
```diff
@@ -366,6 +371,17 @@ public void prepareRequest(ServletRequest request)
         baseRequest.setMethod(method);
     }
 
+    private boolean hasExpiredIdToken(HttpSession session)
+    {
+        if (session != null)
+        {
+            Map<String, Object> claims = (Map)session.getAttribute(CLAIMS);
```
travisspencer, line 378: This is guarenteed to be safe? Should you check in the condition of the if statement on line 376 that the attribute  is a map? e.g.
`if (session != null && session.getAttribute(CLAIMS) instanceof Map)`

travisspencer (dismissed):
Looks good. Didn't review the test closely, but the run-time code looks good IMO.

jetty-openid/src/test/java/org/eclipse/jetty/security/openid/OpenIdProvider.java
```diff
@@ -103,6 +103,16 @@ public OpenIdProvider(String clientId, String clientSecret)
...
 
+    public void setIdTokenExpiry(long expiry)
+    {
+        _idTokenExpiry = expiry;
+    }
+
+    public long getIdTokenExpiry()
+    {
+        return _idTokenExpiry;
+    }
```
sbordet, line 115: Not an expiry, but a duration.
Rename to `idTokenDuration` or `idTokenValidity` or `idTokenValidFor`?
Also, I'd use `Duration` rather than `long` as field/parameter.

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdCredentials.java
```diff
@@ -137,12 +137,26 @@ private void validateClaims(OpenIdConfiguration configuration) throws Exception
...
+
+    public static boolean checkExpiry(Map<String, Object> claims)
+    {
+        if (claims == null)
+            return true;
+
+        // Check that the ID token has not expired by checking the exp claim.
+        long expiry = (Long)claims.get("exp");
+        long currentTimeSeconds = System.currentTimeMillis() / 1000;
+        return (currentTimeSeconds > expiry);
```
sbordet: Let's try to avoid using `System.currentTimeMilllis()`, use:

```suggestion
        Instant.ofEpochSeconds((Long)claims.get("exp")).isBefore(Instant.now())
```

Please also fix `OpenIdCredentialsTest` by using `Instant.now().plusSeconds(5).getEpochSecond()`.

jetty-openid/src/test/java/org/eclipse/jetty/security/openid/OpenIdProvider.java
```diff
@@ -289,7 +299,7 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws S
             long expiry = System.currentTimeMillis() + Duration.ofMinutes(10).toMillis();
```
sbordet: Avoid `System.currentTimeMillis()`, so:

```suggestion
            Instant expiry = Instant.now().plusSeconds(10);
```

jetty-openid/src/test/java/org/eclipse/jetty/security/openid/OpenIdProvider.java
```diff
@@ -374,10 +384,11 @@ public String getSubject()
             return subject;
         }
 
-        public String getIdToken(String provider, String clientId)
+        public String getIdToken(String provider, String clientId, long expiry)
         {
-            long expiry = System.currentTimeMillis() + Duration.ofMinutes(1).toMillis();
-            return JwtEncoder.createIdToken(provider, clientId, subject, name, expiry);
+            long currentTimeMillis = System.currentTimeMillis();
+            long expiryTime = (currentTimeMillis + expiry) / 1000;
```
sbordet: Avoid `System.currentTimeMillis()`.
Also, `expiry` passed here is already a datetime (not a duration), so you must not add the current datetime.

jetty-openid/src/test/java/org/eclipse/jetty/security/openid/OpenIdProvider.java
```diff
@@ -374,10 +384,11 @@ public String getSubject()
             return subject;
         }
 
-        public String getIdToken(String provider, String clientId)
+        public String getIdToken(String provider, String clientId, long expiry)
```
sbordet: As this method creates a new IdToken, rename to `newIdToken()`?

jetty-openid/src/test/java/org/eclipse/jetty/security/openid/OpenIdProvider.java
```diff
@@ -374,10 +384,11 @@ public String getSubject()
...
         {
-            long expiry = System.currentTimeMillis() + Duration.ofMinutes(1).toMillis();
-            return JwtEncoder.createIdToken(provider, clientId, subject, name, expiry);
+            long currentTimeMillis = System.currentTimeMillis();
+            long expiryTime = (currentTimeMillis + expiry) / 1000;
+            return JwtEncoder.createIdToken(provider, clientId, subject, name, expiryTime);
         }
 
         @Override
```
sbordet, line 394: If `equals()` is overridden, also `hashCode()` must be overridden.

jetty-openid/src/main/config/modules/openid.mod
```diff
@@ -45,3 +45,6 @@ etc/jetty-openid.xml
 
 ## What authentication method to use with the Token Endpoint (client_secret_post, client_secret_basic).
 # jetty.openid.authMethod=client_secret_post
+
+## Whether the user should be logged out after the idToken expiry.
```
sbordet:
```suggestion
## Whether the user should be logged out after the idToken expires.
```

sbordet (changes requested):

jetty-openid/src/main/java/org/eclipse/jetty/security/openid/OpenIdAuthenticator.java
lachlan-roberts: We do not store anything else except the map of claims at this attribute so I think this is fine. There is a possibility someone messes with the attribute value, but if they do this and change it to something which is not a map they are asking for trouble.

sbordet (approved):

tests/test-distribution/src/test/java/org/eclipse/jetty/tests/distribution/openid/OpenIdProvider.java
```diff
@@ -278,20 +298,53 @@
...
+            if (logoutRedirect == null)
+            {
+                resp.setStatus(HttpServletResponse.SC_OK);
+                resp.getWriter().println("logout success on end_session_endpoint");
+                return;
+            }
+
+            loggedInUsers.decrement();
+            resp.setContentType("text/plain");
+            resp.sendRedirect(logoutRedirect);
```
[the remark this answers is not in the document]
joakime: This was dismissed, as this was a test case

lachlan-roberts:
Will wait until the jetty-security changes in jetty-12 are done before I merge this, as the OpenID code will be moved to jetty-core. (see https://github.com/eclipse/jetty.project/pull/9405)

PR #9531, opened 2023-03-24
parameterize MultiPartServletTest for complianceModes and add extra test

lachlan-roberts:
No behaviour changes just improving the test.

- add test to check that maximum number of parts is allowed.
- parameterise tests over RFC7578 and LEGACY compliance modes.

### DIFF
--- a/jetty-servlet/src/test/java/org/eclipse/jetty/servlet/MultiPartServletTest.java
+++ b/jetty-servlet/src/test/java/org/eclipse/jetty/servlet/MultiPartServletTest.java
@@ -75,6 +75,7 @@ public class MultiPartServletTest
     private ServerConnector connector;
     private HttpClient client;
     private Path tmpDir;
+    private ServletContextHandler contextHandler;
 
     private static final int MAX_FILE_SIZE = 512 * 1024;
     private static final int MAX_REQUEST_SIZE = 1024 * 1024 * 8;
@@ -155,7 +156,7 @@ public class MultiPartServletTest
         MultipartConfigElement defaultConfig = new MultipartConfigElement(tmpDir.toAbsolutePath().toString(),
             -1, -1, 1);
 
-        ServletContextHandler contextHandler = new ServletContextHandler(ServletContextHandler.SESSIONS);
+        contextHandler = new ServletContextHandler(ServletContextHandler.SESSIONS);
         contextHandler.setContextPath("/");
         ServletHolder servletHolder = contextHandler.addServlet(MultiPartServlet.class, "/");
         servletHolder.getRegistration().setMultipartConfig(config);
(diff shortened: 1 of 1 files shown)


jetty-servlet/src/test/java/org/eclipse/jetty/servlet/MultiPartServletTest.java
```diff
@@ -200,21 +217,57 @@ public void testLargePart() throws Exception
...
+    public void testManyParts(MultiPartFormDataCompliance compliance) throws Exception
+    {
+        connector.getConnectionFactory(HttpConnectionFactory.class).getHttpConfiguration()
+            .setMultiPartFormDataCompliance(compliance);
+
+        byte[] byteArray = new byte[10];
+        Arrays.fill(byteArray, (byte)1);
+
+        MultiPartRequestContent multiPart = new MultiPartRequestContent();
+        for (int i = 0; i < 1000; i++)
```
sbordet: IIUC, this is the default value of `maxFormKeys`. If so, make the `ServletContextHandler` in `start()` a field, and here call `contextHandler.getMaxFormKeys()` so this is not a "magic" number.

jetty-servlet/src/test/java/org/eclipse/jetty/servlet/MultiPartServletTest.java
```diff
@@ -200,21 +217,57 @@ public void testLargePart() throws Exception
...
-        byte[] byteArray = new byte[1024];
+        connector.getConnectionFactory(HttpConnectionFactory.class).getHttpConfiguration()
+            .setMultiPartFormDataCompliance(compliance);
+
+        byte[] byteArray = new byte[5];
         Arrays.fill(byteArray, (byte)1);
 
         MultiPartRequestContent multiPart = new MultiPartRequestContent();
-        for (int i = 0; i < 1024 * 1024; i++)
+        for (int i = 0; i < 1001; i++)
```
sbordet: Ditto.
Number `1001` seems a typo, so use `contextHandler.getMaxFormKeys() + 1` instead.

sbordet (changes requested):

jetty-servlet/src/test/java/org/eclipse/jetty/servlet/MultiPartServletTest.java
lachlan-roberts: Yes, but I also wanted to test that the default value was reasonable as the typical usage of Jetty doesn't have someone explicitly setting `maxFormKeys`.

sbordet: Sure, you can assert that separately, something like `context.getMaxFormKeys() > 500 && < 1500`.

sbordet (approved):

joakime (approved):

joakime:
@lachlan-roberts you can merge this one now

PR #9532, opened 2023-03-24
introduce ShrinkManager

magibney:
Leverages new organization of logic in #9498 for performant min TTL idleTimeout and separately configured linear shrink rate. Basically rebases #9354 onto #9498, but ends up much cleaner/more performant.

### DIFF
--- a/jetty-util/src/main/java/org/eclipse/jetty/util/thread/LinearShrinkManager.java
+++ b/jetty-util/src/main/java/org/eclipse/jetty/util/thread/LinearShrinkManager.java
@@ -1,0 +1,151 @@
+//
+// ========================================================================
+// Copyright (c) 1995 Mort Bay Consulting Pty Ltd and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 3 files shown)


gregw (commented):
I'm open to making shrink logic extensible, but I'm not sure it is worth a whole new interface based API.   I'd lean more towards extensible protected methods in QTP.

However, I'm still not 100% on the correctness of any algorithm yet, so let's get one working first and then generalise.

magibney:
There's still meaningful differences here -- I was planning to just update the base branch to `jetty-10.0.x` for this PR now that #9498 is merged? Would you prefer that I open.a new PR, or would you mind re-opening this one?

sbordet:
No idea what GH did here... why was this PR closed when I merged #9498?
Because the branch does not exist anymore?
I cannot reopen this PR, the "Reopen" button is grayed out.

However, I think we addressed your original points with #9498.
If not, what did we miss?

magibney:
ha, that's wild that GH did this automatically and attributed the closing to you! Let me gather my thoughts and circle back to this in the next few days. Thanks again for the work on #9498!

PR #9513, opened 2023-03-17
fix missing request attributes when using auth-method CLIENT-CERT

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>

PR #9533, opened 2023-03-24
Fixes retrieval of SSL Servlet request attributes.

sbordet:
Alternative to #9513.
Removed requestlog attributes that were never used.
Code cleanups to remove usage of deprecated methods.


jetty-ee10/jetty-ee10-proxy/src/test/java/org/eclipse/jetty/ee10/proxy/ClientAuthProxyTest.java
```diff
@@ -117,7 +118,7 @@ private void startServer(Handler handler) throws Exception
         serverTLS.setSniRequired(false);
         serverTLS.setNeedClientAuth(true);
         // The KeyStore is also a TrustStore.
-        serverTLS.setKeyStorePath(MavenTestingUtils.getTestResourceFile("client_auth/server_keystore.p12").getAbsolutePath());
+        serverTLS.setKeyStorePath(MavenTestingUtils.getTestResourcePath("client_auth/server_keystore.p12").toAbsolutePath().toString());
```
joakime: The `.toAbsolutePath()` part is irrelevant. (both before and now)
The call to `MavenTestingUtils.getTestResourcePath(String)` always returns an absolute path.

joakime (approved):
I can see a small change for the ee9 code to make it more stable for the ee9 -> ee8 translation process, but otherwise I'm happy with it.

joakime (approved):
Looking good.

olamy (approved):
and TCK pass now

Issue #9552, opened 2023-03-30
Jetty 12 - Rewrite of the Jetty WebSocket APIs


sbordet:
**Jetty version(s)**
12+

**Enhancement Description**
The Jetty WebSocket APIs were designed following a similar design of the Jakarta WebSocket APIs.

However, that design is old and it clashes with other Jetty core APIs that we have.
In particular:

* No reason to have `Session.getRemote()`
* Duplicated blocking and non-blocking APIs
* The event handling side (`WebSocket[Connection|Frame|Partial|PingPong]Listener`) has unclear semantic wrt threading. In particular, can an event handler method perform blocking calls?
Not to mention the number of interfaces for the various events that may not be optimal when it comes to `instanceof` checks (see https://bugs.openjdk.org/browse/JDK-8180450).
* Implicit demand for new frames/messages when returning from a method, with `Session.suspend()` to try to cope with it.
This clashes with all the other Jetty 12 APIs that have explicit demand APIs.
* Weird handling of methods that take `Reader` or `InputStream` that the implementation treats specially wrt threading: since they offer blocking APIs, there is no demand because the event handler method is not exited, so the implementation has to demand internally in order to feed the `Reader` or `InputStream`.

This would be the right time to re-design the APIs to align them with the rest of the Jetty 12 APIs.

sbordet:
See also #9550 for an explanation of why the undefined semantic of the APIs, coupled with blocking APIs, causes certain scenarios to lockup.

Would be great if `WebSocketUpgradeHandler` could have back `invocationType==NON_BLOCKING` because now the APIs have a defined semantic wrt threading (and no possibility to invoke WebSocket blocking APIs).

sbordet:
@gregw @lachlan-roberts @lorban @joakime your thoughts?

monish-byte:
hey @sbordet I would like to work on this issue. As I am a beginner I would appreciate your guidance. thank you.

lachlan-roberts:
@monish-byte this issue is not something you could help with.

If you want to contribute look at open issues with the "Help Wanted" tag.
https://github.com/eclipse/jetty.project/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22

lachlan-roberts:
@sbordet I think all of those points are just because the Jetty API is designed to mirror what the Jakarta WebSocket API is doing. 

But I think if we decide to move away from that style now would be a good time.

[...]

PR #9553, opened 2023-03-30
Fixes #9550 - Jetty 12 - Flaky test WebSocketProxyTest.testEcho

### LINKED ISSUE #9550: Jetty 12 - Flaky test `org.eclipse.jetty.websocket.tests.proxy.WebSocketProxyTest.testEcho`
**Jetty version(s)**
Jetty 12

17 & 19

**OS type/version**
Linux

**Description**
New flaky test showing up ...

```
Expected: is "hello world!"
     but: was null
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at org.eclipse.jetty.websocket.tests.proxy.WebSocketProxyTest.testEcho(WebSocketProxyTest.java:117)

Running org.eclipse.jetty.websocket.tests.proxy.WebSocketProxyTest.testEcho()
2023-03-29 15:58:27.823:INFO :oejs.Server:main: jetty-12.0.0-SNAPSHOT; built: 2023-03-29T15:28:30.456Z; git: a818a5369903bba82eaebc7e5c646e47999cd3fb; jvm 17.0.6+10
2023-03-29 15:58:27.825:INFO :oejsh.ContextHandler:main: Started oejsh.ContextHandler@3af7d855{ROOT,/,b=null,a=AVAILABLE,h=oejws.WebSocketUpgradeHandler@77049094{STARTED}}
2023-03-29 15:58:27.825:INFO :oejs.AbstractConnector:main: Started ServerConnector@f88bfbe{HTTP/1.1, (http/1.1)}{0.0.0.0:42983}
2023-03-29 15:58:27.825:INFO :oejs.Server:main: Started oejs.Server@1b7a52dd{STARTING}[12.0.0-SNAPSHOT,sto=0] @29760ms
2023-03-29 15:58:32.838:INFO :oejwc.WebSocketClient:main: Shutdown WebSocketClient@6ab9df83[coreClient=oejwcc.WebSocketCoreClient@440e3ce6{STARTED},openSessions.size=1]
2023-03-29 15:58:32.842:INFO :oejs.Server:main: Stopped oejs.Server@1b7a52dd{STOPPING}[12.0.0-SNAPSHOT,sto=0]
2023-03-29 15:58:32.843:INFO :oejs.AbstractConnector:main: Stopped ServerConnector@f88bfbe{HTTP/1.1, (http/1.1)}{0.0.0.0:0}
```

sbordet:
Unfortunately the fix is to make WebSocketUpgradeHandler.invocationType=BLOCKING, because the current semantic of the Jetty WebSocket APIs is that it is possible to call blocking APIs from within WebSocket event handler methods.

### DIFF
--- a/jetty-core/jetty-websocket/jetty-websocket-jetty-server/src/main/java/org/eclipse/jetty/websocket/server/WebSocketUpgradeHandler.java
+++ b/jetty-core/jetty-websocket/jetty-websocket-jetty-server/src/main/java/org/eclipse/jetty/websocket/server/WebSocketUpgradeHandler.java
@@ -21,6 +21,7 @@ import org.eclipse.jetty.server.Response;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.handler.ContextHandler;
 import org.eclipse.jetty.util.Callback;
+import org.eclipse.jetty.util.thread.Invocable;
 import org.eclipse.jetty.websocket.api.WebSocketContainer;
 import org.eclipse.jetty.websocket.core.WebSocketComponents;
 import org.eclipse.jetty.websocket.core.server.WebSocketMappings;
@@ -120,9 +121,10 @@ public class WebSocketUpgradeHandler extends Handler.Wrapper
     @Override
     public InvocationType getInvocationType()
     {
-        // Must be BLOCKING because EndPoint.upgrade() ends up invoking
-        // application code in the WebSocket "connect" event handler,
-        // e.g. a method annotated with @OnWebSocketConnect, that may block.
-        return InvocationType.BLOCKING;
+        if (isDynamic())
+            return InvocationType.BLOCKING;
+        Handler handler = getHandler();
+        InvocationType result = handler == null ? InvocationType.NON_BLOCKING : handler.getInvocationType();
+        return Invocable.combine(result, container.getInvocationType());
     }
 }
--- a/jetty-core/jetty-websocket/jetty-websocket-jetty-server/src/main/java/org/eclipse/jetty/websocket/server/WebSocketUpgradeHandler.java
+++ b/jetty-core/jetty-websocket/jetty-websocket-jetty-server/src/main/java/org/eclipse/jetty/websocket/server/WebSocketUpgradeHandler.java
@@ -116,4 +116,13 @@ public class WebSocketUpgradeHandler extends Handler.Wrapper
             return true;
         return super.handle(request, response, callback);
     }
+
+    @Override
+    public InvocationType getInvocationType()
+    {
+        // Must be BLOCKING because EndPoint.upgrade() ends up invoking
+        // application code in the WebSocket "connect" event handler,
+        // e.g. a method annotated with @OnWebSocketConnect, that may block.
+        return InvocationType.NON_BLOCKING;
+    }
 }
--- a/jetty-core/jetty-websocket/jetty-websocket-jetty-server/src/main/java/org/eclipse/jetty/websocket/server/WebSocketUpgradeHandler.java
+++ b/jetty-core/jetty-websocket/jetty-websocket-jetty-server/src/main/java/org/eclipse/jetty/websocket/server/WebSocketUpgradeHandler.java
@@ -123,6 +123,6 @@ public class WebSocketUpgradeHandler extends Handler.Wrapper
         // Must be BLOCKING because EndPoint.upgrade() ends up invoking
         // application code in the WebSocket "connect" event handler,
         // e.g. a method annotated with @OnWebSocketConnect, that may block.
-        return InvocationType.NON_BLOCKING;
+        return InvocationType.BLOCKING;
     }
 }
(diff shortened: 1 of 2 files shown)


jetty-core/jetty-websocket/jetty-websocket-jetty-server/src/main/java/org/eclipse/jetty/websocket/server/WebSocketUpgradeHandler.java
```diff
@@ -116,4 +116,13 @@ public boolean handle(Request request, Response response, Callback callback) thr
...
         return super.handle(request, response, callback);
     }
+
+    @Override
+    public InvocationType getInvocationType()
+    {
+        // Must be BLOCKING because EndPoint.upgrade() ends up invoking
+        // application code in the WebSocket "connect" event handler,
+        // e.g. a method annotated with @OnWebSocketConnect, that may block.
+        return InvocationType.NON_BLOCKING;
```
gregw: comment doesn't match the return code?

gregw (changes requested):

jetty-core/jetty-websocket/jetty-websocket-jetty-server/src/main/java/org/eclipse/jetty/websocket/server/WebSocketUpgradeHandler.java
gregw: Can we execute the application code or the upgrade?
Otherwise if we ever use websocket, then we can never be non-blocking

joakime: This should be BLOCKING (like you showed me)

joakime (changes requested):

jetty-core/jetty-websocket/jetty-websocket-jetty-server/src/main/java/org/eclipse/jetty/websocket/server/WebSocketUpgradeHandler.java
sbordet: Doh!

sbordet: We do not want to execute the WebSocket listener (or annotated) methods, otherwise we're going to sink the performance.

I have captured the non-blocking problem in #9552 -- I don't think there is a way around with the current APIs.
If we mark `WebSocketUpgradeHandler` as `NON_BLOCKING`, then we block the selector thread.
If we impose that WebSocket listener (or annotated) methods _must_ be non-blocking, then we either force application to be careful and _not_ use the blocking APIs that we offer, and/or we force applications to always use `suspend()/resume()` -- hence #9552, where we only offer non-blocking APIs, an explicit demand API, we impose non-blocking behavior and if the application code blocks on other APIs, then we block the selector thread, but at least it's clear and documented.

Note that right now, for a plain HTTP setup with just a `ContextHandler`, we block the selector thread if the application `Handler` (child of `ContextHandler`) blocks on other APIs.
At least we would have a consistent behavior.

gregw: Then we need to at least make this configurable, so an application that uses websocket can promise not to block and thus configure this handler to be non-blocking.  Otherwise it is a bit sad that we are saying we can NEVER have a completely asynchronous server+application that uses websocket!

sbordet: `WebSocketUpgradeHandler` inherits `setDynamic(boolean)` which is typically used to mark the `Handler` as `BLOCKING`.

The name `setDynamic()` does not really hint at the `InvocationType`, but would do the job.
Perhaps rename to `setBlocking()` (as we don't really want to expose `EITHER`)?

joakime (approved):

jetty-core/jetty-websocket/jetty-websocket-jetty-server/src/main/java/org/eclipse/jetty/websocket/server/WebSocketUpgradeHandler.java
```diff
@@ -120,9 +121,10 @@ public boolean handle(Request request, Response response, Callback callback) thr
...
     {
-        // Must be BLOCKING because EndPoint.upgrade() ends up invoking
-        // application code in the WebSocket "connect" event handler,
-        // e.g. a method annotated with @OnWebSocketConnect, that may block.
-        return InvocationType.BLOCKING;
+        if (isDynamic())
+            return InvocationType.BLOCKING;
+        Handler handler = getHandler();
+        InvocationType result = handler == null ? InvocationType.NON_BLOCKING : handler.getInvocationType();
+        return Invocable.combine(result, container.getInvocationType());
```
gregw: Tiny quibble with the name. Maybe better to have:
```suggestion
        InvocationType handlerInvocationType = handler == null ? InvocationType.NON_BLOCKING : handler.getInvocationType();
        return Invocable.combine(handlerInvocationType, container.getInvocationType());
```

gregw (approved):
LGTM other than a tiny name quibble

PR #9616, opened 2023-04-03
Fixes #6184 - JEP-411 will deprecate/remove the SecurityManager from …

### LINKED ISSUE #6184: Remove usages of classes associated with JEP-411 that deprecate/remove the SecurityManager from the JVM

sbordet:
…the JVM.

Removed usages of `SecurityManager` and `AccessControlller.doPrivileged()`. In places where they are still necessary, now using reflection.

### DIFF
--- a/jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/DefaultAuthConfigFactory.java
+++ b/jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/DefaultAuthConfigFactory.java
@@ -23,6 +23,7 @@ import javax.security.auth.message.config.AuthConfigFactory;
 import javax.security.auth.message.config.AuthConfigProvider;
 import javax.security.auth.message.config.RegistrationListener;
 
+import org.eclipse.jetty.util.security.SecurityUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -144,34 +145,7 @@ public class DefaultAuthConfigFactory extends AuthConfigFactory
 
     private static void checkPermission(String permission)
     {
-        try
-        {
-            Object securityManager = getSecurityManager();
-            if (securityManager == null)
-                return;
-            securityManager.getClass().getMethod("checkPermission")
-                .invoke(securityManager, new AuthPermission(permission));
-        }
-        catch (SecurityException x)
-        {
-            throw x;
-        }
-        catch (Throwable ignored)
-        {
-        }
-    }
-
-    private static Object getSecurityManager()
-    {
-        try
-        {
-            // Use reflection to work with Java versions that have and don't have SecurityManager.
-            return System.class.getMethod("getSecurityManager").invoke(null);
-        }
-        catch (Throwable ignored)
-        {
-            return null;
-        }
+        SecurityUtils.checkPermission(new AuthPermission(permission));
     }
 
     private static String getKey(String layer, String appContext)
--- a/jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/DefaultAuthConfigFactory.java
+++ b/jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/DefaultAuthConfigFactory.java
@@ -60,9 +60,7 @@ public class DefaultAuthConfigFactory extends AuthConfigFactory
     @Override
     public String registerConfigProvider(String className, Map properties, String layer, String appContext, String description)
     {
-        SecurityManager sm = System.getSecurityManager();
-        if (sm != null)
-            sm.checkPermission(new AuthPermission("registerAuthConfigProvider"));
+        checkPermission("registerAuthConfigProvider");
 
         String key = getKey(layer, appContext);
         AuthConfigProvider configProvider = createConfigProvider(className, properties);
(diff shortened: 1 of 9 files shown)


jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/DefaultAuthConfigFactory.java
```diff
@@ -145,13 +137,43 @@ public RegistrationContext getRegistrationContext(String registrationID)
...
     {
-        SecurityManager sm = System.getSecurityManager();
-        if (sm != null)
-            sm.checkPermission(new AuthPermission("refreshAuth"));
+        checkPermission("refreshAuth");
 
         // TODO: maybe we should re-construct providers created from classname.
     }
 
+    private static void checkPermission(String permission)
```
janbartel, line 146: Can we have a util class or method that does this as this is used in a few places.

jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlConfiguration.java
```diff
@@ -1815,86 +1813,81 @@ public static void main(final String... args) throws Exception
     {
         try
         {
-            AccessController.doPrivileged((PrivilegedExceptionAction<Void>)() ->
-            {
-                Properties properties = new Properties();
-                properties.putAll(System.getProperties());
+            Properties properties = new Properties();
```
janbartel, line 1816: If we're running in < jdk18, why don't we continue to do the privileged action?

jetty-util/src/main/java/org/eclipse/jetty/util/MemoryUtils.java
```diff
@@ -25,18 +22,11 @@
...
         {
-            value = Integer.parseInt(AccessController.doPrivileged(new PrivilegedAction<String>()
-            {
-                @Override
-                public String run()
-                {
-                    return System.getProperty("org.eclipse.jetty.util.cacheLineBytes", String.valueOf(defaultValue));
-                }
-            }));
+            value = Integer.parseInt(System.getProperty("org.eclipse.jetty.util.cacheLineBytes", String.valueOf(defaultValue)));
```
janbartel, line 29: If running in < jdk 18 don't we still want to do the privileged action?

janbartel (changes requested):

jetty-xml/src/main/java/org/eclipse/jetty/xml/XmlConfiguration.java
sbordet: Because the usage of privileged actions was done in very few places, leaving without many other places that would have required it.
Given that the few ones we had were definitely not enough, I think it's better to have none.

jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/DefaultAuthConfigFactory.java
sbordet: I thought about this and there are only 2 places, which are quite unrelated: here and `ContextHandler`.
Given that JASPI is not the most common specification, I thought that the less dependencies the better, so a utility class was not worth the effort.
Note also that all JASPI permissions are `AuthPermission`s, while a generic utility class would need a more generic signature, making this class a little more verbose.

gregw (dismissed):
I kind of would like to see a util class for at least `getSecurityManager()`... but then I'm ok with it as is as well. 60:40

jetty-jaspi/src/main/java/org/eclipse/jetty/security/jaspi/DefaultAuthConfigFactory.java
janbartel: I'm pretty sure this is no more complex than changing the method signature to `checkPermission(Permission)`.  So my preference is still for a utility class/method.

janbartel (dismissed):
Still prefer a utility/class method as the cleanest solution. But I guess we can always refactor that after you commit this.

joakime:
> I kind of would like to see a util class for at least `getSecurityManager()`... but then I'm ok with it as is as well. 60:40

What would this getter return?
The `java.lang.SecurityManager` class is being removed from the JVM.

joakime (dismissed):

joakime (approved):

gregw (approved):
Much better

PR #9618, opened 2023-04-03
Fixes #7608 - Jetty-12 MetaData cleanup needed

### LINKED ISSUE #7608: Jetty-12 MetaData cleanup needed
**Target Jetty version(s)**
Jetty-12

**Enhancement Description**

The MetaData class is badly in need of a clean up:

 + Too many constructors, with multiple paths through to super constructor.
 + The content-length handling is done poorly.  It was meant to be an optimization to avoid HttpFields lookups when the content length is already known, but it now is more likely to cause a HttpFields lookup in every construction.
 + The content-length is sometimes set to -1 and sometimes set tio Integer.MIN_VALUE.  I think the later is probably meant to indicate that a HttpFields lookup should be done.  Eitherway, this needs to be clarified and perhaps some named constants used to communicate the intent.
 + Consideration could be given to using java-17 records

sbordet:
* Removed unnecessary constructors from MetaData, MetaData.Request and MetaData.Response.
* Removed MetaData.Request.getURIString() (available as getHttpURI().toString()).
* Renamed MetaData.getFields() -> getHttpFields(), as they can be headers or trailers.
* Renamed MetaData.Request.getURI() -> getHttpURI().
* Normalized handling of contentLength, now always -1 (rather than Long.MIN_VALUE) if unknown.
* Permutated MetaData.Response constructor parameters to be consistent with MetaData.Request.
* MetaData.Request's method and httpURI must be non-null.

### DIFF
--- a/documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/client/http2/HTTP2ClientDocs.java
+++ b/documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/client/http2/HTTP2ClientDocs.java
@@ -236,7 +236,7 @@ public class HTTP2ClientDocs
                 }
                 else
                 {
-                    System.getLogger("http2").log(INFO, "Received trailers {0}", metaData.getFields());
+                    System.getLogger("http2").log(INFO, "Received trailers {0}", metaData.getHttpFields());
                 }
             }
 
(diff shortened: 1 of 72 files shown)


joakime (approved):

gregw (approved):
Good stuff

PR #9626, opened 2023-04-04
Fixes #9438 - Jetty 12: Review JakartaWebSocketClientContainer use of…

### LINKED ISSUE #9438: Jetty 12: Review JakartaWebSocketClientContainer use of reflection
**Jetty version(s)**
12+

**Description**
https://github.com/eclipse/jetty.project/blob/jetty-12.0.x/jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jakarta-client/src/main/java/org/eclipse/jetty/ee9/websocket/jakarta/client/JakartaWebSocketClientContainer.java#L356

Methods `addToContextHandler()` and `removeFromContextHandler()` use reflection, but they are currently broken as the methods do not exist anymore.

This is very fragile, as it obviously did not stand a method renaming/refactoring already, so should be converted to something more stable.

sbordet:
… reflection

* Removed the use of reflection as it was unnecessary.
* Improved `setShutdownContainer()` to take into account multiple web applications, so it is now a `Map<ClassLoader, ContainerLifeCycle>` so that each web application has its own container to register `JakartaWebSocketClientContainer` instances.

### DIFF
--- a/jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jakarta-client/src/main/java/org/eclipse/jetty/ee10/websocket/jakarta/client/JakartaWebSocketClientContainer.java
+++ b/jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jakarta-client/src/main/java/org/eclipse/jetty/ee10/websocket/jakarta/client/JakartaWebSocketClientContainer.java
@@ -15,8 +15,10 @@ package org.eclipse.jetty.ee10.websocket.jakarta.client;
 
 import java.io.IOException;
 import java.net.URI;
+import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Future;
@@ -44,7 +46,6 @@ import org.eclipse.jetty.ee10.websocket.jakarta.common.JakartaWebSocketFrameHand
 import org.eclipse.jetty.ee10.websocket.jakarta.common.JakartaWebSocketFrameHandlerFactory;
 import org.eclipse.jetty.util.annotation.ManagedObject;
 import org.eclipse.jetty.util.component.ContainerLifeCycle;
-import org.eclipse.jetty.util.component.LifeCycle;
 import org.eclipse.jetty.util.thread.ShutdownThread;
 import org.eclipse.jetty.websocket.core.WebSocketComponents;
 import org.eclipse.jetty.websocket.core.client.WebSocketCoreClient;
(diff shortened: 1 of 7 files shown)


jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jakarta-client/src/main/java/org/eclipse/jetty/ee10/websocket/jakarta/client/JakartaWebSocketClientContainer.java
```diff
@@ -63,13 +64,15 @@
 public class JakartaWebSocketClientContainer extends JakartaWebSocketContainer implements jakarta.websocket.WebSocketContainer
 {
     private static final Logger LOG = LoggerFactory.getLogger(JakartaWebSocketClientContainer.class);
-    private static final AtomicReference<ContainerLifeCycle> SHUTDOWN_CONTAINER = new AtomicReference<>();
+    private static final AtomicReference<Map<ClassLoader, ContainerLifeCycle>> SHUTDOWN_MAP = new AtomicReference<>(new ConcurrentHashMap<>());
```
lachlan-roberts: Do we need this `AtomicReference`? Doesn't look the value is ever modified.

Can't we just have the 
```java
private static final Map<ClassLoader, ContainerLifeCycle> SHUTDOWN_MAP = new ConcurrentHashMap<>();
```

lachlan-roberts (changes requested):

lachlan-roberts:
@sbordet there is now failing tests that are expecting the container to be a bean contained in the contextHandler. So this expectation will need to be changed.

I think the reflection was just an optimisation to be able to be used in embedded Jetty without having to manually add the Shutdown SCI.

sbordet:
> I think the reflection was just an optimisation to be able to be used in embedded Jetty without having to manually add the Shutdown SCI.

Can you expand on this?
IIUC, even when used in embedded Jetty, in a server scenario, the SCI will run because it is in the jar that you must put in the class-path, so the shutdown SCI will always be there and run.
So I don't see the "optimization"?

lachlan-roberts:
> Can you expand on this?
> IIUC, even when used in embedded Jetty, in a server scenario, the SCI will run because it is in the jar that you must put in the class-path, so the shutdown SCI will always be there and run.
> So I don't see the "optimization"?

The SCI is not automatically discovered and run in embedded Jetty using `ServletContextHandler`. You need to manually add it with `contextHandler.addServletContainerInitializer(shutdownContainer)`. To discover the SCIs you need to properly configure a `WebAppContext`.

And so in this case, the container is added to a bean on the context handler and nothing is done with the shutdown container. The websocket client container is then shutdown when the contextHandler stops and no SCI is required.

gregw (commented):
Looks reasonable to me. But will defer to @lachlan-roberts to give the green tick.

lachlan-roberts:
@sbordet @gregw I found a case which worked previously but is now broken after this PR.

If you have two servers using `ServletContextHandler` then stopping one server will stop the ClientContainer from the other server. Previously this did not happen because we were adding to a bean on the ContextHandler and not using the static map on shutdown container.

So this PR actually breaks the use case of using embedded Jetty without classloader isolation. And this was the reason that we introduced this reflection in the first case.

I have a test case for this which passes on Jetty 12.0.x branch and fails on this branch.

gregw:
@lachlan-roberts So is your use-case not using different classloaders for each context, so that the key used in the map is just the system classloader?

At the time the ServletContextListener is called to tell us to context is stopping, we know the context and we know that we might be using jakarta websocket clients.  Surely we can iterate through the map (should it be a list) and just find the containers that are within a specific context to shutdown?

I guess the problem is on the other end... when we start a container, how does it know which context it is in... without using reflection?

sbordet:
@lachlan-roberts @gregw I restored (but simplified) the use of reflection.
Also put in the code the comments about the 3 mechanisms.

Also kept the `Map` keyed to a `ClassLoader` to isolate different web applications.
This is necessary because in the test case the 3 SCIs were run before starting the 3 `JakartaWebSocketClientContainer`.
So only the last SCI container was kept in the field, and all the 3 `JakartaWebSocketClientContainer` were added to the last container.
In case a web application is started after the others, it would have overwritten the SCI container with the existing `JakartaWebSocketClientContainer`s.

gregw (approved):
Good comments!

lachlan-roberts (approved):
LGTM

PR #9628, opened 2023-04-05
Issue #9554 - move common hpack/qpack code to jetty-http

### LINKED ISSUE #9554: Move (qpack/hpack) HuffmanDecoder / HuffmanEncoder / NBitInteger* to common location
**Jetty version(s)**
Jetty 10 / 11 / 12

All

**OS type/version**
All

**Description**
The code between http2 (hpack) and http3 (qpack) are very similar.
Consider moving them to a common location (jetty-util?) to have 1 place to manage issues/bugs around them.

lachlan-roberts:
## Issue #9554
move common hpack/qpack code to jetty-http


gregw (changes requested):
So if I'm reading this right, h2 had a stateless `Huffman` class with encode and decode methods,  but h3 separated those into a stateless `HuffmanEncoder` and a stateful `HuffmanDecoder` class, that uses the table in 'HuffmanEncoder`.   Is that correct?

If so, I'd kind of prefer if we kept the stateless `Huffman` with the tables and static methods.  Then introduce just a `HuffmanDecoder` class that does the stateful decoding.  Perhaps use a Record to hold the state in the base class so it can be passed in by the stateful class

lachlan-roberts:
@gregw I moved out the static huffman codes into a `Huffman` class.

gregw (approved):

PR #9629, opened 2023-04-05
Fix MultiPartServletTest race.

gregw:
The writing of the large content may see an EofException before the 400 response has been processed and delivered to the listener.


gregw:
@lorban @sbordet I think this flake is just a race in the client, which is writing large content that will ultimately cause the server to send a 400 response and close the connection.   Sometimes the 400 is seen by the client, other times the EofException is seen first.

I'm not sure how to solve this, other than for the client to wait after sending every chunk of content to see if a 400 comes back?

So for now, I have modified the test harness to accept either result.

gregw:
note that I have verified that even when the client sees Eof, the server has sent a 400.

sbordet (approved):

Issue #7091, opened 2021-11-09
Add SOCKS5 support


ghost:
**Target Jetty version(s)**
We are currently using the jetty websocket 10 using SOCKS4 proxy.
**Enhancement Description**
We need to support in the feature websocket via SOCKS5 proxy. 
Would it be possible to have the SOCKS5 version in the next release of jetty?

joakime:
The spec for SOCKS5 can be found at https://datatracker.ietf.org/doc/html/rfc1928

This would be a low priority for us at the moment, as we are focused on HTTP/3 and Jetty 12 with Servlet 6.

We'll happily take a PR to add SOCKS5 support to Jetty 10.

PR #9633, opened 2023-04-11
#7091 support socks5

huisongma:
support socks5 proxy

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/Socks5.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/Socks5.java
@@ -1,0 +1,151 @@
+//
+// ========================================================================
+// Copyright (c) 1995 Mort Bay Consulting Pty Ltd and others.
+//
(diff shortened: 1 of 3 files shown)


sbordet:
Thanks for this contribution!

Since SOCKS5 is way more complicated than SOCKS4, I would like to avoid hardcoding the authentication mechanism.
The idea is to have a `Socks5` class, with inner classes that implement different authentication mechanisms, for now only "no authentication" (say `Socks5.NoAuthentication`) and "username/password" (say `Socks5.UsernamePasswordAuthentication`).

Shared code can be put in the `Socks5` class.

Also, test cases are missing.
I would like to see at least test cases for abrupt connection close from the server,  no answer from the server (timeout on client), garbage bytes from the server, response failure from server and response succeeded from server.

Please try to stick with Java/Jetty conventions. For example, `enum` constants are all upper case, the `log` instance is upper case, etc.

huisongma:
continue on #9653

PR #9634, opened 2023-04-11
Issue #9554 - move common hpack/qpack code to jetty-http (jetty-10)

lachlan-roberts:
## Issue #9554
bring back changes from PR https://github.com/eclipse/jetty.project/pull/9628 to Jetty-10


jetty-http/src/main/java/org/eclipse/jetty/http/MetaData.java
```diff
@@ -100,6 +101,28 @@ public Iterator<HttpField> iterator()
...
+
+        MetaData other = (MetaData)obj;
+        if (!Objects.equals(_httpVersion, other._httpVersion))
+            return false;
+        if (!Objects.equals(_contentLength, other._contentLength))
+            return false;
+        if (!Objects.equals(_fields, other._fields))
+            return false;
+        return _trailerSupplier == null && other._trailerSupplier == null;
+    }
```
gregw: A bit strange these changes are in this PR.... but good additions.

jetty-http/src/main/java/org/eclipse/jetty/http/compression/Huffman.java
```diff
@@ -11,14 +11,13 @@
...
-package org.eclipse.jetty.http2.hpack;
-
-import java.nio.ByteBuffer;
-
-import org.eclipse.jetty.util.Utf8StringBuilder;
+package org.eclipse.jetty.http.compression;
 
 public class Huffman
 {
+    private Huffman()
```
gregw, line 21: Javadoc for class needed

jetty-http/src/main/java/org/eclipse/jetty/http/compression/Huffman.java
```diff
@@ -348,199 +347,46 @@
...
-            if (c >= 128 || c < ' ')
-                return -1;
-            needed += table[c][1];
-        }
-
-        return (needed + 7) / 8;
-    }
-
-    private static int octetsNeeded(final int[][] table, byte[] b)
+    public static boolean isIllegalCharacter(char c)
```
gregw: Illegal for what?   Also why character?  Huffman can encode all bytes, so this doesn't really make sense.
It should probably be on a HttpToken class with better name and javadoc.

jetty-http/src/main/java/org/eclipse/jetty/http/compression/Huffman.java
```diff
@@ -348,199 +347,46 @@
...
+        return (c >= 256 || c < ' ');
     }
 
-    /**
-     * @param table The table to encode by
-     * @param buffer The buffer to encode to
-     * @param s The string to encode
-     */
-    private static void encode(final int[][] table, ByteBuffer buffer, String s)
+    public static char getReplacementCharacter(char c)
```
gregw: protected?

also it needs a bit of javadoc and links to why replacement character is different.

plus it needs to be renamed, as it is sanitising not replacing.  Probably also should be on HttpToken

jetty-http/src/main/java/org/eclipse/jetty/http/compression/Huffman.java
```diff
@@ -348,199 +347,46 @@
...
-            while (n >= 8)
-            {
-                n -= 8;
-                buffer.put((byte)(current >> n));
-            }
+            // A recipient of CR, LF, or NUL within a field value MUST either reject the message
+            // or replace each of those characters with SP before further processing
+            case '\r':
+            case '\n':
+            case 0x00:
```
gregw: this assumes a ASCII compatible charset, so why not just cast to a char and call the other method?

jetty-http/src/main/java/org/eclipse/jetty/http/compression/HuffmanDecoder.java
```diff
@@ -11,20 +11,30 @@
...
+import static org.eclipse.jetty.http.compression.Huffman.rowbits;
+import static org.eclipse.jetty.http.compression.Huffman.rowsym;
 
 public class HuffmanDecoder
 {
-    static final char EOS = HuffmanEncoder.EOS;
-    static final char[] tree = HuffmanEncoder.tree;
-    static final char[] rowsym = HuffmanEncoder.rowsym;
-    static final byte[] rowbits = HuffmanEncoder.rowbits;
+    public static String decode(ByteBuffer buffer, int length) throws EncodingException
```
gregw: Didn't we agree to get rid of this method, as it encourages inefficient usage of temporary decoder?

jetty-http/src/main/java/org/eclipse/jetty/http/compression/HuffmanDecoder.java
```diff
@@ -51,17 +61,19 @@ public String decode(ByteBuffer buffer) throws EncodingException
...
+                    if (rowsym[_node] == Huffman.EOS)
                     {
                         reset();
                         throw new EncodingException("eos_in_content");
                     }
 
                     // terminal node
-                    _utf8.append((byte)(0xFF & rowsym[_node]));
+                    int i = 0xFF & rowsym[_node];
+                    i = Huffman.getReplacementCharacter(i);
```
gregw: I don't understand why this is here? Why replace here?  Don't we only replace when converting characters to bytes?

Also the method is poorly named, as it only replaces bad characters.  It is more a sanitise.    But even then, that is applying a HTTP sanity check that perhaps should not be in the base Huffman code?

jetty-http/src/main/java/org/eclipse/jetty/http/compression/HuffmanDecoder.java
```diff
@@ -94,7 +106,9 @@ public String decode(ByteBuffer buffer) throws EncodingException
                 break;
             }
 
-            _utf8.append((byte)(0xFF & rowsym[_node]));
+            int i = 0xFF & rowsym[_node];
+            i = Huffman.getReplacementCharacter(i);
```
gregw: ditto

jetty-http/src/main/java/org/eclipse/jetty/http/compression/HuffmanEncoder.java
```diff
@@ -0,0 +1,144 @@
...
+
+        if (n > 0)
+        {
+            current <<= (8 - n);
+            current |= (0xFF >>> n);
+            buffer.put((byte)(current));
+        }
+    }
+
+    private static void encode(final int[][] table, ByteBuffer buffer, byte[] b)
```
gregw: If our Huffman encode/decode are going to apply RFC9110 rules for what characters are acceptable, then do we need to encode bytes (or decode strings)?      Is it possible to have encode just do String->ByteBuffer and decode just do ByteBuffer->String ?

jetty-http/src/main/java/org/eclipse/jetty/http/compression/NBitIntegerParser.java
```diff
@@ -11,12 +11,27 @@
...
 
 import java.nio.ByteBuffer;
 
 public class NBitIntegerParser
 {
+    public static int decode(ByteBuffer buffer, int prefix) throws EncodingException
+    {
+        // TODO: This is a fix for HPACK as it already takes the first byte of the encoded integer.
+        if (prefix != 8)
+            buffer.position(buffer.position() - 1);
```
gregw: Could this fix be put in a utility method on hpack/qpack?   Does it also apply to qpack?

jetty-http/src/main/java/org/eclipse/jetty/http/compression/NBitStringParser.java
```diff
@@ -11,15 +11,17 @@
...
 //
 
-package org.eclipse.jetty.http3.qpack.internal.util;
+package org.eclipse.jetty.http.compression;
 
 import java.nio.ByteBuffer;
 
+import org.eclipse.jetty.util.CharsetStringBuilder;
+
 public class NBitStringParser
```
gregw: Can you take the opportunity to add a bit of javadoc to these classes

gregw (changes requested):
A few nits

jetty-http/src/main/java/org/eclipse/jetty/http/MetaData.java
lachlan-roberts: This was for a test case I wrote comparing metadata but it didn't work out and I got rid of the test case and thought I might as well keep this.

jetty-http/src/main/java/org/eclipse/jetty/http/compression/HuffmanDecoder.java
lachlan-roberts: Yep, I removed the usage but didn't remove this.
Done.

jetty-http/src/main/java/org/eclipse/jetty/http/compression/HuffmanDecoder.java
lachlan-roberts: I have it now so we won't encode the illegal vchars, but when we decode we replace any bad vchars.

So here someone could send us an encoded huffman string with control chars etc, and we don't want to leave these in the decoded field value.

jetty-http/src/main/java/org/eclipse/jetty/http/compression/HuffmanEncoder.java
lachlan-roberts: Not sure what you mean here.

jetty-http/src/main/java/org/eclipse/jetty/http/compression/NBitIntegerParser.java
lachlan-roberts: No this is specific to HPACK, it does something weird where it peeks at the first byte and then expects this to put it back.

Looks like I can also remove this static and do this in HPACK only.

jetty-http/src/main/java/org/eclipse/jetty/http/compression/NBitStringParser.java
lachlan-roberts: done

jetty-http/src/main/java/org/eclipse/jetty/http/compression/HuffmanEncoder.java
gregw: I see two ways to make these APIs clean and the class behaviours regular.  Either:

a) The Huffman classes are just about encoding/decoding bytes and have no character rules.  The character rules are done at a higher level (either in hpack/qpack or a class in between).
OR b) The Huffman classes are specific to encoding/decoding HTTP headers (they are in http package after all) and thus they know about encoding/decoding characters specifically for HTTP headers.

We are option b) and I think we want to be as we can do more efficient char to byte conversions with the knowledge that we are ISO-8859-1 because we are only HTTP headers.  So given that these classes are only for http headers, then do we need methods to encode from bytes or decode to bytes?    

For example, is the method `public static void encode(ByteBuffer buffer, byte[] b)` ever actually used?  If so then why? by who?

If this class has rules about characters that can be encoded then it should only have the API: `public static void encode(ByteBuffer buffer, String s)` and not the byte version.

jetty-http/src/main/java/org/eclipse/jetty/http/compression/HuffmanDecoder.java
```diff
@@ -51,17 +64,19 @@ public String decode(ByteBuffer buffer) throws EncodingException
...
                     {
                         reset();
                         throw new EncodingException("eos_in_content");
                     }
 
                     // terminal node
-                    _utf8.append((byte)(0xFF & rowsym[_node]));
+                    int i = 0xFF & rowsym[_node];
+                    i = HttpTokens.sanitizeFieldVchar(i);
+                    _builder.append((byte)i);
```
gregw: So this is kind of double sanitizing now.  The 0xFF is already cutting off big values.   The `rowsym` table is already characters, so why not leave it as a char then sanitize using the char method?

jetty-http/src/main/java/org/eclipse/jetty/http/HttpTokens.java
```diff
@@ -231,5 +231,60 @@ else if (b >= 0x80) // OBS
...
         }
     }
+
+    /**
+     * This is used when decoding to not decode illegal characters based on RFC9110.
+     * CR, LF, or NUL are replaced with ' ', all other control and multibyte characters
+     * are replaced with '?'. If this is given a legal character the same value will be returned.
+     *
+     * @param c the character to test.
+     * @return the original character or the replacement character ' ' or '?'.
```
gregw: Need to say more about the char return.  Specifically that it is known to be ISO-8859-1 character (i.e can have a simple coerce to byte conversion applied to it).

jetty-http/src/main/java/org/eclipse/jetty/http/HttpTokens.java
```diff
@@ -231,5 +231,60 @@ else if (b >= 0x80) // OBS
             }
         }
     }
+
+    /**
+     * This is used when decoding to not decode illegal characters based on RFC9110.
+     * CR, LF, or NUL are replaced with ' ', all other control and multibyte characters
+     * are replaced with '?'. If this is given a legal character the same value will be returned.
+     *
```
gregw: Maybe add the ABNF:
```abnf
field-vchar = VCHAR / obs-text
obs-text = %x80-FF
VCHAR          =  %x21-7E
```

gregw (dismissed):
a few more.

jetty-http/src/main/java/org/eclipse/jetty/http/HttpTokens.java
```diff
@@ -236,9 +236,14 @@ else if (b >= 0x80) // OBS
      * This is used when decoding to not decode illegal characters based on RFC9110.
      * CR, LF, or NUL are replaced with ' ', all other control and multibyte characters
      * are replaced with '?'. If this is given a legal character the same value will be returned.
-     *
+     * <pre>
+     * field-vchar = VCHAR / obs-text
+     * obs-text = %x80-FF
+     * VCHAR          =  %x21-7E
```
joakime: Please align

lachlan-roberts: done

jetty-http/src/main/java/org/eclipse/jetty/http/HttpTokens.java
```diff
@@ -231,5 +231,50 @@ else if (b >= 0x80) // OBS
...
+            case '\r':
+            case '\n':
+            case 0x00:
+                return ' ';
+
+            default:
+                if (c >= 256 || c < ' ')
+                    return '?';
+        }
+
```
sbordet: Remove empty line.

jetty-http/src/main/java/org/eclipse/jetty/http/HttpTokens.java
```diff
@@ -231,5 +231,50 @@ else if (b >= 0x80) // OBS
...
+        {
+            // A recipient of CR, LF, or NUL within a field value MUST either reject the message
+            // or replace each of those characters with SP before further processing
+            case '\r':
+            case '\n':
+            case 0x00:
+                return ' ';
+
+            default:
+                if (c >= 256 || c < ' ')
```
sbordet: Replace with the call to `isIllegalVchar()`.

jetty-http/src/main/java/org/eclipse/jetty/http/MetaData.java
```diff
@@ -100,6 +101,28 @@ public Iterator<HttpField> iterator()
...
+            return false;
+
+        MetaData other = (MetaData)obj;
+        if (!Objects.equals(_httpVersion, other._httpVersion))
+            return false;
+        if (!Objects.equals(_contentLength, other._contentLength))
+            return false;
+        if (!Objects.equals(_fields, other._fields))
+            return false;
+        return _trailerSupplier == null && other._trailerSupplier == null;
```
sbordet: You can't compare `_trailerSupplier` in this way.
If you need the equality for testing, just use a static method in the test class.
Remove both equals() and hashCode().

jetty-http/src/main/java/org/eclipse/jetty/http/compression/NBitIntegerParser.java
```diff
@@ -11,29 +11,53 @@
...
 
-package org.eclipse.jetty.http3.qpack.internal.util;
+package org.eclipse.jetty.http.compression;
 
 import java.nio.ByteBuffer;
 
+/**
+ * Used to decode integers as described in RFC7541.
+ */
 public class NBitIntegerParser
```
sbordet: Can it be renamed to `NBitIntegerDecoder` like others are?

jetty-http2/http2-hpack/src/main/java/org/eclipse/jetty/http2/hpack/HpackDecoder.java
```diff
@@ -267,14 +274,61 @@ public MetaData decode(ByteBuffer buffer) throws HpackException.SessionException
...
+            _huffmanDecoder.reset();
+        }
+    }
+
+    public static String toISO8859String(ByteBuffer buffer, int length)
+    {
+        CharsetStringBuilder.Iso8859StringBuilder builder = new CharsetStringBuilder.Iso8859StringBuilder();
         for (int i = 0; i < length; ++i)
         {
             builder.append((char)(0x7F & buffer.get()));
```
sbordet: Change it to `buffer.get() & 0x7F`.
Also feels a bit weird... are you sure this is the right logic? If so, can you comment why, maybe reference the spec if it comes from some RFC?

jetty-http2/http2-hpack/src/main/java/org/eclipse/jetty/http2/hpack/HpackDecoder.java
```diff
@@ -267,14 +274,61 @@ public MetaData decode(ByteBuffer buffer) throws HpackException.SessionException
...
+            compressionException.initCause(e);
+            throw compressionException;
+        }
+        finally
+        {
+            _huffmanDecoder.reset();
+        }
+    }
+
+    public static String toISO8859String(ByteBuffer buffer, int length)
```
sbordet: Rename to `toISO88591String()` -- note the additional `1`.

jetty-http2/http2-hpack/src/main/java/org/eclipse/jetty/http2/hpack/HpackFieldPreEncoder.java
```diff
@@ -67,12 +69,12 @@ else if (header == HttpHeader.CONTENT_LENGTH && value.length() > 1)
...
         int nameIdx = HpackContext.staticIndex(header);
         if (nameIdx > 0)
-            NBitInteger.encode(buffer, bits, nameIdx);
+            NBitIntegerEncoder.encode(buffer, bits, nameIdx);
         else
         {
             buffer.put((byte)0x80);
-            NBitInteger.encode(buffer, 7, Huffman.octetsNeededLC(name));
-            Huffman.encodeLC(buffer, name);
+            NBitIntegerEncoder.encode(buffer, 7, HuffmanEncoder.octetsNeededLowercase(name));
```
sbordet: Rename to `octetsNeededLowerCase()` -- capital `C`.

jetty-http2/http2-hpack/src/main/java/org/eclipse/jetty/http2/hpack/HpackFieldPreEncoder.java
```diff
@@ -67,12 +69,12 @@ else if (header == HttpHeader.CONTENT_LENGTH && value.length() > 1)
...
+            HuffmanEncoder.encodeLowercase(buffer, name);
```
sbordet: Ditto.

jetty-util/src/main/java/org/eclipse/jetty/util/CharsetStringBuilder.java
```diff
@@ -0,0 +1,279 @@
...
+        if (charset == StandardCharsets.ISO_8859_1)
+            return new Iso8859StringBuilder();
+        if (charset == StandardCharsets.US_ASCII)
+            return new UsAsciiStringBuilder();
+
+        // Use a CharsetDecoder that defaults to CodingErrorAction#REPORT
+        return new DecoderStringBuilder(charset.newDecoder());
+    }
+
+    class Iso8859StringBuilder implements CharsetStringBuilder
```
sbordet: Rename to `Iso88591StringBuilder` -- note the additional `1`.

jetty-util/src/main/java/org/eclipse/jetty/util/CharsetStringBuilder.java
```diff
@@ -0,0 +1,279 @@
...
+
+        @Override
+        public void reset()
+        {
+            _stringBuilder.setLength(0);
+        }
+    }
+}
+
+
```
sbordet: Remove extra empty line.

sbordet (changes requested):

jetty-http2/http2-hpack/src/main/java/org/eclipse/jetty/http2/hpack/HpackDecoder.java
joakime: This drops the high-bit, that seems like a bug.

lachlan-roberts: Yeah probably because this was originally for ASCII, I will remove the 0x7F and use the `HttpTokens.sanitizeFieldVchar` instead.

sbordet (approved):

PR #9641, opened 2023-04-12
Renamed "Core" environment to "JVM" in `StartArgs`

gregw:
Maintained a "Core" environment for use by deployer for apps that only use core classes plus whatever else is configured for the core environment.

### DIFF
--- a/jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ContextProvider.java
+++ b/jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ContextProvider.java
@@ -360,7 +360,7 @@ public class ContextProvider extends ScanningAppProvider
                 xmlc.setJettyStandardIdsAndProperties(getDeploymentManager().getServer(), path);
 
                 // If it is a core context environment, then look for a classloader
-                ClassLoader coreContextClassLoader = Environment.SERVER.equals(environment) ? findCoreContextClassLoader(path) : null;
+                ClassLoader coreContextClassLoader = Environment.CORE.equals(environment) ? findCoreContextClassLoader(path) : null;
                 if (coreContextClassLoader != null)
                     Thread.currentThread().setContextClassLoader(coreContextClassLoader);
 
@@ -455,7 +455,7 @@ public class ContextProvider extends ScanningAppProvider
 
         if (urls.isEmpty())
             return null;
-        return new URLClassLoader(urls.toArray(new URL[0]), Environment.SERVER.getClassLoader());
+        return new URLClassLoader(urls.toArray(new URL[0]), Environment.CORE.getClassLoader());
     }
 
     protected void initializeContextPath(ContextHandler context, Path path)
--- a/jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ContextProvider.java
+++ b/jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ContextProvider.java
@@ -360,7 +360,7 @@ public class ContextProvider extends ScanningAppProvider
                 xmlc.setJettyStandardIdsAndProperties(getDeploymentManager().getServer(), path);
 
                 // If it is a core context environment, then look for a classloader
-                ClassLoader coreContextClassLoader = Environment.CORE.equals(environment) ? findCoreContextClassLoader(path) : null;
+                ClassLoader coreContextClassLoader = Environment.SERVER.equals(environment) ? findCoreContextClassLoader(path) : null;
                 if (coreContextClassLoader != null)
                     Thread.currentThread().setContextClassLoader(coreContextClassLoader);
 
@@ -455,7 +455,7 @@ public class ContextProvider extends ScanningAppProvider
 
         if (urls.isEmpty())
             return null;
-        return new URLClassLoader(urls.toArray(new URL[0]), Environment.CORE.getClassLoader());
+        return new URLClassLoader(urls.toArray(new URL[0]), Environment.SERVER.getClassLoader());
     }
 
     protected void initializeContextPath(ContextHandler context, Path path)
--- a/jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ScanningAppProvider.java
+++ b/jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ScanningAppProvider.java
@@ -161,13 +161,13 @@ public abstract class ScanningAppProvider extends ContainerLifeCycle implements 
                         Files.exists(path.getParent().resolve(basename + ".WAR")) ||
                         Files.exists(path.getParent().resolve(basename + "/WEB-INF")));
             boolean coreProvider = _deploymentManager.getAppProviders().stream()
-                .map(AppProvider::getEnvironmentName).anyMatch(Environment.CORE.getName()::equals);
+                .map(AppProvider::getEnvironmentName).anyMatch(Environment.SERVER.getName()::equals);
 
             // TODO review these heuristics... or even if we should have them at all
             if (isWebapp || (Files.isDirectory(path) && _deploymentManager.getDefaultEnvironmentName() != null))
                 environmentName = _deploymentManager.getDefaultEnvironmentName();
             else if (coreProvider)
-                environmentName = Environment.CORE.getName();
+                environmentName = Environment.SERVER.getName();
 
             if (StringUtil.isNotBlank(environmentName))
             {
@@ -213,10 +213,10 @@ public abstract class ScanningAppProvider extends ContainerLifeCycle implements 
             throw new IllegalStateException("No configuration dir specified");
         if (_environmentName == null)
         {
-            List<Environment> nonCore = Environment.getAll().stream().filter(environment -> !environment.equals(Environment.CORE)).toList();
-            if (nonCore.size() != 1)
+            List<Environment> nonServer = Environment.getAll().stream().filter(environment -> !environment.equals(Environment.SERVER)).toList();
+            if (nonServer.size() != 1)
                 throw new IllegalStateException("No environment configured");
-            _environmentName = nonCore.get(0).getName();
+            _environmentName = nonServer.get(0).getName();
         }
 
         Environment environment = Environment.get(_environmentName);
--- a/jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ScanningAppProvider.java
+++ b/jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ScanningAppProvider.java
@@ -161,8 +161,7 @@ public abstract class ScanningAppProvider extends ContainerLifeCycle implements 
                     Files.exists(path.getParent().resolve(basename + ".war")) ||
                         Files.exists(path.getParent().resolve(basename + ".WAR")) ||
                         Files.exists(path.getParent().resolve(basename + "/WEB-INF")));
-            boolean coreProvider = _deploymentManager.getAppProviders().stream()
-                .map(AppProvider::getEnvironmentName).anyMatch(Environment.CORE.getName()::equalsIgnoreCase);
+            boolean coreProvider = _deploymentManager.hasAppProviderFor(Environment.CORE.getName());
 
             // TODO review these heuristics... or even if we should have them at all
             if (isWebapp || (Files.isDirectory(path) && defaultEnvironmentName != null))
@@ -192,9 +191,7 @@ public abstract class ScanningAppProvider extends ContainerLifeCycle implements 
             if (defaultEnvironmentName != null && defaultEnvironmentName.equalsIgnoreCase(getEnvironmentName()))
             {
                 // if the app specified an environment name, then produce warning if there is no provider for it.
-                boolean appProvider4env = _deploymentManager.getAppProviders().stream()
-                    .map(AppProvider::getEnvironmentName).anyMatch(environmentName::equalsIgnoreCase);
-                if (!appProvider4env)
+                if (!_deploymentManager.hasAppProviderFor(environmentName))
                     LOG.warn("No AppProvider with environment {} for {}", environmentName, app);
                 return null;
             }
--- a/jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ScanningAppProvider.java
+++ b/jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ScanningAppProvider.java
@@ -150,6 +150,7 @@ public abstract class ScanningAppProvider extends ContainerLifeCycle implements 
 
         String environmentName = app.getEnvironmentName();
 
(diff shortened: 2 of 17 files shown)


jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ContextProvider.java
```diff
@@ -360,7 +360,7 @@ public void initializeDefaults(Object context)
                 xmlc.setJettyStandardIdsAndProperties(getDeploymentManager().getServer(), path);
 
                 // If it is a core context environment, then look for a classloader
-                ClassLoader coreContextClassLoader = Environment.CORE.equals(environment) ? findCoreContextClassLoader(path) : null;
+                ClassLoader coreContextClassLoader = Environment.SERVER.equals(environment) ? findCoreContextClassLoader(path) : null;
```
sbordet: Should not the method and variables be renamed to `findServerContextClassLoader(path)`?

sbordet (changes requested):

gregw:
@sbordet There is no "Server" environment in the runtime.  Only a "Core" used for deployment.   The "Server" environment only exists in the startArgs as a place to accumulate the config for the server.  Hmmm maybe it should be called "JVM" ?

jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ContextProvider.java
gregw: reverted to CORE

jetty-core/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
```diff
@@ -278,16 +281,16 @@ public void dumpJavaEnvironment(PrintStream out)
...
         out.println();
-        out.println("Jetty Core Environment:");
-        out.println("-----------------------");
-        Environment coreEnvironment = getCoreEnvironment();
-        coreEnvironment.dumpProperty(out, JETTY_VERSION_KEY);
-        coreEnvironment.dumpProperty(out, JETTY_TAG_NAME_KEY);
-        coreEnvironment.dumpProperty(out, JETTY_BUILDNUM_KEY);
-        coreEnvironment.dumpProperty(out, "jetty.home");
-        coreEnvironment.dumpProperty(out, "jetty.base");
+        out.println("JVM Environment:");
```
joakime: "Jetty JVM Environment" perhaps?

Something that indicates that this is both "Jetty" and this is the "JVM" named environment for Jetty.

jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ScanningAppProvider.java
```diff
@@ -161,11 +162,11 @@ protected App createApp(Path path)
                         Files.exists(path.getParent().resolve(basename + ".WAR")) ||
                         Files.exists(path.getParent().resolve(basename + "/WEB-INF")));
             boolean coreProvider = _deploymentManager.getAppProviders().stream()
-                .map(AppProvider::getEnvironmentName).anyMatch(Environment.CORE.getName()::equals);
+                .map(AppProvider::getEnvironmentName).anyMatch(Environment.CORE.getName()::equalsIgnoreCase);
```
sbordet: This code is repeated a few times -> introduce `_deploymentManager.hasEnvironment(Environment.CORE)`?

jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ScanningAppProvider.java
```diff
@@ -180,19 +181,19 @@ else if (coreProvider)
...
+            if (environmentName.equalsIgnoreCase(getEnvironmentName()))
             {
                 if (LOG.isDebugEnabled())
                     LOG.debug("{} created {}", this, app);
                 return app;
             }
 
             // If we are the default provider then we may warn
-            if (Objects.equals(getEnvironmentName(), _deploymentManager.getDefaultEnvironmentName()))
+            if (defaultEnvironmentName != null && defaultEnvironmentName.equalsIgnoreCase(getEnvironmentName()))
```
sbordet: Assuming for this provider the env name is not blank, then:

```suggestion
            if (getEnvironmentName().equalsIgnoreCase(defaultEnvironmentName))
```

jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ScanningAppProvider.java
```diff
@@ -180,19 +181,19 @@ else if (coreProvider)
...
             {
                 // if the app specified an environment name, then produce warning if there is no provider for it.
                 boolean appProvider4env = _deploymentManager.getAppProviders().stream()
-                    .map(AppProvider::getEnvironmentName).anyMatch(environmentName::equals);
+                    .map(AppProvider::getEnvironmentName).anyMatch(environmentName::equalsIgnoreCase);
```
sbordet: Use `hasEnvironment()` as commented above.

jetty-core/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
```diff
@@ -224,9 +227,9 @@ public void expandEnvironments(List<Module> activeModules) throws IOException
         }
     }
 
-    public Environment getCoreEnvironment()
+    public Environment getJvmEnvironment()
```
sbordet: I prefer `getJVMEnvironment()`.

But isn't this better defined as `startEnvironment` or `configurationEnvironment` or `bootEnvironment` or `rootEnvironment`?

jetty-core/jetty-start/src/main/java/org/eclipse/jetty/start/Environment.java
```diff
@@ -79,8 +84,8 @@ public void addPropertyFileRef(String arg)
 
     public void addUniquePropertyFile(String propertyFileRef, Path propertyFile) throws IOException
     {
-        if (!"core".equalsIgnoreCase(getName()))
-            throw new IllegalStateException("Property files not supported in non core environments");
+        if (!"server".equalsIgnoreCase(getName()))
+            throw new IllegalStateException("Property files not supported in non Server environments");
```
sbordet: Java docs says it's a "start" environment, but here is called "server", but I think elsewhere is called "jvm"?

jetty-core/jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
```diff
@@ -239,14 +239,14 @@ public void listConfig(PrintStream out, StartArgs args)
...
-        // Dump Core Classpath
-        dumpClasspathWithVersions(coreEnvironment.getName(), out, coreEnvironment.getClasspath());
-        // Dump Core Resolved XMLs
-        coreEnvironment.dumpActiveXmls(out);
+        // Dump Server Properties
+        serverEnvironment.dumpProperties(out);
+        // Dump Server Classpath
+        dumpClasspathWithVersions(serverEnvironment.getName(), out, serverEnvironment.getClasspath());
+        // Dump Server Resolved XMLs
+        serverEnvironment.dumpActiveXmls(out);
```
sbordet: "server" variables should not be renamed to "jvm"?

jetty-core/jetty-start/src/main/java/org/eclipse/jetty/start/Environment.java
```diff
@@ -23,6 +23,11 @@
 import java.util.List;
 import java.util.Properties;
 
+/**
+ * A start environment that contains the configurations that will be used to
+ * build a runtime {@code org.eclipse.jetty.util.component.Environment} via
+ * {@code --env} arguments passed to {@code org.eclipse.jetty.xml.XmlConfiguration#main(java.lang.String...)}
+ */
 public class Environment
```
sbordet: Rename to something different from `Environment` to not confuse it with the `jetty-util` Environment?
Perhaps `Settings`?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/component/Environment.java
```diff
@@ -17,14 +17,18 @@
...
 import org.eclipse.jetty.util.TypeUtil;
 
+/**
+ * A named runtime environment containing a {@link ClassLoader} and {@link Attributes}.
+ */
 public interface Environment extends Attributes
 {
-    Environment CORE = ensure("core");
+    // Ensure there is a core environment for possible later deployments to it
+    Environment CORE = ensure("Core");
```
sbordet: All others are lower case and so they are in module files, so I'll keep it lower case.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/component/Environment.java
```diff
@@ -46,10 +50,22 @@ static Environment set(Environment environment)
         return Named.__environments.put(environment.getName(), environment);
     }
 
+    /**
+     * @return The case insensitvie name of the environment.
+     */
     String getName();
```
sbordet, line 56: Perhaps add, like other classes, an `is(String name)` method that does the lowercase equals?

sbordet (changes requested):

jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ScanningAppProvider.java
gregw: can't assume it is not null.

sbordet: Why not? What would this provider deploy then, if not one among core or eeN?

gregw: Because I have to read several other methods to work out that at the time this method is called that the originally null field will probably not be null now. But it won't be null so I've switched it.

jetty-core/jetty-start/src/main/java/org/eclipse/jetty/start/Environment.java
joakime: Ya know, we take specific classes from jetty-util and shade (merge) them into the eventual `start.jar`, is this an opportunity to combine these 2 classes into 1?

jetty-core/jetty-start/src/main/java/org/eclipse/jetty/start/Main.java
```diff
@@ -401,7 +400,7 @@ public void start(StartArgs args) throws IOException, InterruptedException
         // Show the version information and return
         if (args.isListClasspath())
         {
-            dumpClasspathWithVersions("Core", System.out, classpath);
+            dumpClasspathWithVersions("Server", System.out, classpath);
```
sbordet: "Server" -> "Jetty"?

jetty-core/jetty-start/src/main/java/org/eclipse/jetty/start/Environment.java
gregw: @joakime no because they are at different points in their lifecycle.  In start and env is properties, xmls and libs, whilst in the runtime an env is a classloader and attributes.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/component/Environment.java
```diff
@@ -46,10 +50,22 @@ static Environment set(Environment environment)
         return Named.__environments.put(environment.getName(), environment);
     }
 
+    /**
+     * @return The case insensitvie name of the environment.
```
sbordet: Typo: `insensitvie`.

jetty-core/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
```diff
@@ -247,25 +252,25 @@ private void addFile(Module module, String uriLocation)
...
         return environment;
     }
 
     public void dumpJavaEnvironment(PrintStream out)
     {
         // Java Details
         out.println();
-        out.println("Java Environment:");
-        out.println("-----------------");
+        out.println("Key System Properties:");
```
sbordet: What does "Key" mean here? Perhaps "JVM version & properties" so that below we can do "Jetty version & properties"?

jetty-core/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
```diff
@@ -278,16 +283,16 @@ public void dumpJavaEnvironment(PrintStream out)
...
         out.println();
-        out.println("Jetty Core Environment:");
-        out.println("-----------------------");
-        Environment coreEnvironment = getCoreEnvironment();
-        coreEnvironment.dumpProperty(out, JETTY_VERSION_KEY);
-        coreEnvironment.dumpProperty(out, JETTY_TAG_NAME_KEY);
-        coreEnvironment.dumpProperty(out, JETTY_BUILDNUM_KEY);
-        coreEnvironment.dumpProperty(out, "jetty.home");
-        coreEnvironment.dumpProperty(out, "jetty.base");
+        out.println("Jetty Version:");
```
sbordet: "Jetty version & properties" instead.

jetty-core/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
```diff
@@ -559,15 +565,15 @@ public CommandLineBuilder getMainArgs(Set<String> parts) throws IOException
         {
             if (isJPMS())
             {
-                Map<Boolean, List<Path>> dirsAndFiles = StreamSupport.stream(coreEnvironment.getClasspath().spliterator(), false)
+                Map<Boolean, List<Path>> dirsAndFiles = StreamSupport.stream(jettyEnvironment.getClasspath().spliterator(), false)
                     .collect(Collectors.groupingBy(Files::isDirectory));
                 Set<Path> files = new HashSet<>(dirsAndFiles.get(false));
 
                 // FIXMW I'm not sure it's a good idea especially with multiple environment..
```
sbordet: Typo: "FIXMW" -> "FIXME"

jetty-core/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
```diff
@@ -1003,8 +1009,8 @@ public void parse(ConfigSources sources)
         ListIterator<ConfigSource> iter = sources.reverseListIterator();
         while (iter.hasPrevious())
         {
-            // Each source starts with core environment.
-            Environment environment = getCoreEnvironment();
+            // Each source starts with server environment.
```
sbordet: "server" -> "Jetty"

sbordet (changes requested):

sbordet (approved):

PR #9644, opened 2023-04-12
Jetty 12.0.x 9643 forward request url

### LINKED ISSUE #9643: jetty-12 tck test com/sun/ts/tests/servlet/spec/requestdispatcher/URLClient.java.getRequestURLForwardTest
This test verifies the implementation of the `HttpServletRequest.getRequestURL` api as concerns `forward` behaviour.

The javadoc states:

```
     * If this request has been forwarded using {@link jakarta.servlet.RequestDispatcher#forward}, the server path in the
     * reconstructed URL must reflect the path used to obtain the RequestDispatcher, and not the server path specified by
     * the client.
```

Our implementation of `ForwardRequest` does not override `getReqeustURL` and thus returns the url of the path specified by the client.

janbartel:
Closes #9643

### DIFF
--- a/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/Dispatcher.java
+++ b/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/Dispatcher.java
@@ -260,6 +260,12 @@ public class Dispatcher implements RequestDispatcher
         }
 
         @Override
+        public StringBuffer getRequestURL()
+        {
+            return _uri == null ? super.getRequestURL() :  new StringBuffer(HttpURI.build(_uri).query(null).scheme(super.getScheme()).host(super.getServerName()).port(super.getServerPort()).asString());
+        }
+
+        @Override
         public Object getAttribute(String name)
         {
             if (name == null)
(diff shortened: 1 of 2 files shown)


gregw (approved):

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/Dispatcher.java
```diff
@@ -259,6 +259,12 @@ public String getRequestURI()
             return _uri == null ? super.getRequestURI() : _uri.getPath();
         }
 
+        @Override
+        public StringBuffer getRequestURL()
+        {
+            return _uri == null ? super.getRequestURL() :  new StringBuffer(HttpURI.build(_uri).query(null).scheme(super.getScheme()).host(super.getServerName()).port(super.getServerPort()).asString());
```
lachlan-roberts, line 265: Should this ever  be returning `super.getRequestURL()`?

From reading https://github.com/jakartaee/servlet/issues/309 and getRequestURL javadoc.

> If this request has been forwarded using RequestDispatcher.forward, the server path in the reconstructed URL must reflect the path used to obtain the RequestDispatcher, and not the server path specified by the client.

It seems `getRequestURL` and `getRequestURI` should never be returning values for the source value but only for the target of the current dispatch.

gregw: @lachlan-roberts named dispatcher don't have paths

PR #9652, opened 2023-04-14
Fixes #9552 - Jetty 12 - Rewrite of the Jetty WebSocket APIs.

### LINKED ISSUE #9552: Jetty 12 - Rewrite of the Jetty WebSocket APIs

sbordet:
* Removed unnecessary classes, among which BatchMode, CloseStatus, etc.
* Coalesced all listener interfaces into Session.Listener.
* Moved RemoteEndpoint functionality to Session.
* Renamed WebSocketPolicy to Configurable.
* Renamed WriteCallback to just Callback, as it is now also used for some listener methods.


jetty-core/jetty-websocket/jetty-websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/Session.java
```diff
@@ -193,5 +170,157 @@ default WebSocketPolicy getPolicy()
      *
      * @return the suspend token suitable for resuming the reading of data on the connection.
      */
+    // TODO: remove and replace with demand()
```
gregw: definitely remove.  This API is always a race!

jetty-core/jetty-websocket/jetty-websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/Session.java
```diff
@@ -14,184 +14,311 @@
...
-     * <p>
-     * If this is called from inside the scope of the message handler the suspend takes effect immediately.
-     * If suspend is called outside the scope of the message handler then the call may take effect
-     * after 1 more frame is delivered.
-     * </p>
-     *
-     * @return the suspend token suitable for resuming the reading of data on the connection.
-     */
-    SuspendToken suspend();
+    interface Listener
```
joakime, line 211: `Session.Listener` is a weird name for receiving websocket messages.

Combining all of the listeners here is also super weird, the old listeners prevented combinations that were problematic.

Examples:
* You couldn't have a frame listener AND a partial OR whole listener at the same time
* You couldn't have a partial AND a whole listener at the same time

jetty-core/jetty-websocket/jetty-websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/Configurable.java
```diff
@@ -64,6 +83,16 @@
...
 
+    /**
+     * The maximum size of a binary message during parsing/generating.
+     * <p>
+     * Binary messages over this maximum will result in a close code 1009 {@link StatusCode#MESSAGE_TOO_LARGE}
+     * </p>
+     *
+     * @param size the maximum allowed size of a binary message.
+     */
+    void setMaxBinaryMessageSize(long size);
```
joakime, line 94: This only applies to whole messages.

Where's the one for partial messages?

jetty-core/jetty-websocket/jetty-websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/Configurable.java
```diff
@@ -80,73 +109,64 @@
...
 
     /**
-     * The input (read from network layer) buffer size.
+     * The maximum payload size of any WebSocket Frame which can be received.
      *
-     * @param size the size in bytes
+     * @return the maximum size of a WebSocket Frame.
      */
-    void setInputBufferSize(int size);
+    long getMaxFrameSize();
```
joakime, line 125: This would be impacted by auto-fragment behavior.
The javadoc here should mention that.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/websocket.adoc
```diff
@@ -248,18 +221,11 @@ While non-blocking APIs are more difficult to use, they don't block the sender t
...
-----
-include::{doc_code}/org/eclipse/jetty/docs/programming/WebSocketDocs.java[tags=streamSendBlocking]
-----
+The APIs offer `sendPartial*(\...)` methods that allow you to send a chunk of the whole message at a time, therefore reducing the memory usage since it is not necessary to have the whole message `String` or `ByteBuffer` in memory to send it.
 
-Streaming sends using non-blocking APIs is more complicated, as you should wait (without blocking!) for the callbacks to complete.
+The APIs for streaming the message content are non-blocking and therefore slightly more complicated to use, as you should wait (without blocking!) for the callbacks to complete.
 
-Fortunately, Jetty provides you with the `IteratingCallback` utility class (described in more details xref:pg-arch-io-echo[in this section]) which greatly simplify the use of non-blocking APIs:
+Fortunately, Jetty provides the `IteratingCallback` utility class (described in more details xref:pg-arch-io-echo[in this section]) which greatly simplify the use of non-blocking APIs:
```
joakime, line 224: Ugh!
Don't promote/push `IteratingCallback` to end users.
Nobody will use it right.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/CoreSession.java
```diff
@@ -152,16 +152,30 @@ public interface CoreSession extends OutgoingFrames, IncomingFrames, Configurati
...
-     * if {@link FrameHandler#isAutoDemanding()} returns true.
+     * <p>Manages flow control by indicating demand for WebSocket frames.</p>
+     * <p>A call to {@link FrameHandler#onFrame(Frame, Callback)} will only
+     * be made if there is demand.</p>
+     * <p>It is an error to call this method if {@link #isAutoDemanding()}
+     * returns {@code true}.</p>
      *
-     * @param n The number of frames that can be handled (in sequential calls to
-     * {@link FrameHandler#onFrame(Frame, Callback)}).  May not be negative.
+     * @param n the number of frames that can be handled in sequential calls to
```
joakime, line 159: And just how will a user use this if they only care about whole message listeners?

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/ByteArrayMessageSink.java
```diff
@@ -57,18 +55,18 @@ public void accept(Frame frame, Callback callback)
...
                 if (frame.hasPayload())
                 {
                     byte[] buf = BufferUtil.toArray(frame.getPayload());
                     methodHandle.invoke(buf, 0, buf.length);
                 }
                 else
+                {
                     methodHandle.invoke(EMPTY_BUFFER, 0, 0);
-
+                }
```
joakime: So much noise on this PR.
Here you add braces to a single line statement, and above you remove them from a single line statement.

jetty-core/jetty-websocket/jetty-websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/annotations/WebSocket.java
```diff
@@ -15,50 +15,31 @@
...
-     * The time in ms (milliseconds) that a websocket may be idle before closing.
-     */
-    int idleTimeout() default -1;
-
-    /**
-     * The maximum size of a text message during parsing/generating.
-     * <p>
-     * Text messages over this maximum will result in a close code 1009 {@link StatusCode#MESSAGE_TOO_LARGE}
-     */
-    int maxTextMessageSize() default -1;
```
joakime, line 56: This ability to configure in the annotation is quite popular.
Why remove it?

jetty-core/jetty-websocket/jetty-websocket-jetty-common/src/test/java/org/eclipse/jetty/websocket/common/EndPoints.java
joakime, line 1: This file should probably be renamed if we are only implementing listeners now

joakime: Each of the internal example classes should drop the name `*Socket` and be perhaps end with `*Listener` now

jetty-core/jetty-websocket/jetty-websocket-jetty-common/src/main/java/org/eclipse/jetty/websocket/common/JettyWebSocketFrameHandlerMetadata.java
```diff
@@ -99,15 +93,15 @@ public MethodHandle getFrameHandle()
         return frameHandle;
     }
 
-    public void setOpenHandler(MethodHandle open, Object origin)
+    public void setConnectHandle(MethodHandle connectHandle, Object origin)
```
joakime: This is a bad name.

Per https://datatracker.ietf.org/doc/html/rfc6455  the named states are ....

* CONNECTING - for when the upgrade has started, and the opening handshake is underway (no frames yet)
* OPEN - for when the upgrade handshake is complete and the two endpoints can now send/receive frames.
* CLOSING - for when the close handshake has begun (limited number of frames can still be sent/received here)
* CLOSED - for when the websocket is fully closed and no more frames can be sent/received.

joakime (changes requested):
Partial review (I'm only 56 files in)

sbordet:
@gregw @lachlan-roberts @joakime here's my review suggestions:

* `FrameHandler`: lost its `isAutodemanding()` because now itself is responsible to perform demand, so there is no need for others to ask it. If it does not do demand itself, it delegates to `MessageSink`s.
* `MessageSink` and subclasses, that all now take a `boolean autoDemand` to decide whether to perform demand automatically or not.
* `WebSocketCoreSession`: the implementation lost its calls to demand, that now are done by the `FrameHandler` or `MessageSink`.
* `core.WebSocketProxy`: while being a test class, it shows how things work differently now that the demand is responsibility of the `FrameHandler`.
* `api.Callback`: was previously called `WriteCallback` but now it is also used for listener methods, so had to drop the `Write`. It is a bit of the same name and functionality of `util.Callback`, but the duplication is necessary as we want the WebSocket `api` to depend only on the JDK classes (for classloading reasons). Open to better names, but `WebSocketCallback` is a mouthful.
* `api.Session`: introduced a `demand()` method, as well as `Listener` to replace the 5 listener interfaces of before, removed `suspend()`. Binary methods in `Session.Listener` now take an additional `Callback` parameter.
* `@OnWebSocketConnect`: historically this annotation was called in this way (and so the correspondent listener method). However, in jetty-websocket-core the same event is called "open" which I think it's more correct.
* `@OnWebSocketXYZ`: should we drop `WebSocket` and just call them `@OnOpen`, `@OnMessage`, etc.? Same for listener methods.
* `@WebSocket` now has an `autoDemand` attribute, correspondent to `Session.Listener.AutoDemanding`.
* `WebSocketSession` now has additional method that were before in `Remote` which is now removed.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/websocket.adoc
gregw: The only way to avoid users needing to do something like `IteratingCallback` is to have a pull API like our `Content.Source` is.   That way, the user can iterate calling something like `readFrame()` until they get a null return and then they can do a `demand` to get an `onFrameAvailable()`.     With such an API, the infinite recursion problem of `demand` calling `onFrame` calling `demand` calling `onFrame` etc. is removed.

Hhmmmm but we could also use the `SerializedInvoker` to ensure that if `onFrame` calls `demand`, that we do not invoke `onFrame` again until the previous one returns? Do we do that?  If so, then we don't need to promote the `IteratingCallback`.  Let me go read what the contract actually is.

gregw: javadoc on FrameHandler says 
>     * <p>This method will never be called concurrently for the
>     * same session

So users don't have infinite recursion problems.   Probably don't need to push `IteratingCallback` to them.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/FrameHandler.java
```diff
@@ -16,102 +16,112 @@
...
-     * Close frames may be responded to by the handler, but if an appropriate close response is not
-     * sent once the callback is succeeded, then a response close will be generated and sent.
+     * <p>Invoked when a WebSocket frame is received.</p>
+     * <p>This method will never be called concurrently for the
+     * same session; will be called sequentially to satisfy the
+     * outstanding demand signaled by calls to
+     * {@link CoreSession#demand(long)}.</p>
+     * <p>Both control and data frames are passed to this method.</p>
+     * <p>CLOSE frames may be responded from this method, but if
+     * they are not responded, then the implementation will respond.
```
gregw: How does the implementation know that the CLOSE will not be responded to?
Should the signature be like our handle method:

```java
    boolean onFrame(Frame frame, Callback callback);
```

If true is returned then the application will invoke the callback, if false is returned then the application will handle the frame (pong for ping, close for close etc.).

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/MessageSink.java
```diff
@@ -14,19 +14,37 @@
...
 public interface MessageSink
 {
     /**
-     * Consume the frame payload to the message.
+     * <p>Consumes the WebSocket frame, possibly asynchronously
+     * when this method has returned.</p>
+     * <p>The callback argument must be completed when the frame
+     * payload is consumed.</p>
+     * <p>The demand for more frames must be explicitly invoked
+     * via {@link CoreSession#demand(long)}.</p>
```
gregw: demand must be invoked by whom?  I assume by the MessageSink implementation?

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/DispatchedMessageSink.java
```diff
@@ -94,74 +94,70 @@
...
 public abstract class DispatchedMessageSink extends AbstractMessageSink
 {
-    private CompletableFuture<Void> dispatchComplete;
-    private MessageSink typeSink;
     private final Executor executor;
+    private volatile CompletableFuture<Void> dispatchComplete;
+    private MessageSink typeSink;
 
-    public DispatchedMessageSink(CoreSession session, MethodHandle methodHandle)
+    public DispatchedMessageSink(CoreSession session, MethodHandle methodHandle, boolean autoDemand)
```
gregw, line 52: If `autoDemand` is false, then what demand(1) mean?  1 frame or 1 message?

At the very least javadoc is needed.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/InputStreamMessageSink.java
```diff
@@ -16,18 +16,17 @@
 import java.lang.invoke.MethodHandle;
 
 import org.eclipse.jetty.websocket.core.CoreSession;
-import org.eclipse.jetty.websocket.core.Frame;
 
 public class InputStreamMessageSink extends DispatchedMessageSink
```
gregw, line 20: We should have a `ContentSourceMessageSink` class that supports our `Content.Source` API.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/PartialStringMessageSink.java
```diff
@@ -24,41 +23,46 @@
 
 public class PartialStringMessageSink extends AbstractMessageSink
 {
-    private Utf8StringBuilder out;
+    private Utf8StringBuilder accumulator;
 
-    public PartialStringMessageSink(CoreSession session, MethodHandle methodHandle)
+    public PartialStringMessageSink(CoreSession session, MethodHandle methodHandle, boolean autoDemand)
```
gregw, line 40: javadoc on methodHandler and class needed

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/StringMessageSink.java
```diff
@@ -27,9 +27,9 @@ public class StringMessageSink extends AbstractMessageSink
     private Utf8StringBuilder out;
     private int size;
 
-    public StringMessageSink(CoreSession session, MethodHandle methodHandle)
+    public StringMessageSink(CoreSession session, MethodHandle methodHandle, boolean autoDemand)
     {
-        super(session, methodHandle);
+        super(session, methodHandle, autoDemand);
```
gregw, line 44: ditto  on javadoc and meaning of autoDemand==false.

If autoDemand is false, should there be an demand API on this the sink that demands n messages rather than n frames?

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java
```diff
@@ -20,12 +20,35 @@
...
 public abstract class AbstractMessageSink implements MessageSink
 {
-    protected final CoreSession session;
-    protected final MethodHandle methodHandle;
+    private final CoreSession session;
+    private final MethodHandle methodHandle;
+    private final boolean autoDemand;
 
-    public AbstractMessageSink(CoreSession session, MethodHandle methodHandle)
+    public AbstractMessageSink(CoreSession session, MethodHandle methodHandle, boolean autoDemand)
```
gregw, line 56: javadoc is needed.

I really question the need for autoDemand=false, as if it is false, how does the user know how many frames to demand? how do they know to demand more when  non-fin frame has arrived and they are waiting for the full message?

Perhaps there should be a demand method on this class, that demand(1) means 1 message and the impl will do frame demands until the fin frame is received?

gregw (changes requested):
In general lots more javadoc is needed.

I think that  `FrameHandler.onFrame` should return a boolean in the same way that `Handler.handle` does. 

Also would be good to have a bridge to our `Content.Source` and `Content.Sink` classes.

Mostly I'm confused as to what `autoDemand=false` means for `MessageSink`s - i.e. same question as @joakime .

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/FrameHandler.java
gregw: As per our call, the close logic is done when the callback is completed, not when the method returns.  That is OK.... but needs javadoc.

jetty-core/jetty-websocket/jetty-websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/Configurable.java
sbordet: It's `setMaxFrameSize(long)` for the size of each frame.

documentation/jetty-documentation/src/main/asciidoc/programming-guide/websocket.adoc
sbordet: @joakime we have a whole section in the docs about `IteratingCallback` so it is documented to be used properly.

The problem is not really reading asynchronously, it's writing asynchronously.
You read one chunk of data, you need to write it passing it a callback.
Then you have to "wait" until the callback is succeeded, so you cannot loop, etc., you really need `IteratingCallback`, or some other APIs (that we don't have for WebSocket) that has an underlying `IteratingCallback`.

I'm all for new APIs, but can be done in a different PR if necessary.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/CoreSession.java
sbordet: This is WebSocket core, so only deals with Frames.

jetty-core/jetty-websocket/jetty-websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/annotations/WebSocket.java
sbordet: Because the annotation was reporting only 4 configuration parameters out of the 8 that can actually be configured.

Furthermore, I see annotation attributes more as something configured once at the class level (like the annotation is specifying, similar to implementing an interface) -- for example you cannot change decoders on-the-fly, or auto-demand, etc., while the setters can be invoked at any time on the session, so it makes less sense to specify them at the class level.

jetty-core/jetty-websocket/jetty-websocket-jetty-common/src/main/java/org/eclipse/jetty/websocket/common/JettyWebSocketFrameHandlerMetadata.java
sbordet: Renamed to "open".

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/DispatchedMessageSink.java
sbordet: This is WebSocket core, so it only deals with frames.

I have expanded and fixed the javadocs for `AbstractMessageSink` and `DispatchedMessageSink`.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java
sbordet: Depends on where the `MessageSink` is used.
Currently, it is not meant to be used in WebSocket core, but if it is, the semantic is that of `CoreSession.demand(long)`, so asking for frames.
Typically `demand(1)` from the application is enough, because then the `MessageSink` will take over the demand when it is not invoking the application.
So, depending on the `MessageSink` implementation, `demand(1)` may mean "next frame" or "next message".

For higher levels, either there is no demand API (Jakarta), so the `MessageSink` is always auto-demanding, or there is `Session.demand()` (with no parameters), so again the "demand" meaning depends on the `MessageSink` implementation.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/InputStreamMessageSink.java
sbordet: Unfortunately this is not possible, as the WebSocket endpoint can only depend on the JDK and the Jetty WebSocket APIs (which in turn only depend on the JDK).

In a webapp I cannot have:

```java
@WebSocket
class MyWSEndpoint {
  @WebSocketMessage
  public void onMessage(Content.Source source) { ... }
}
```

because `Content.Source` would need `jetty-io` in `WEB-INF/lib`, causing classloader woes.

jetty-core/jetty-websocket/jetty-websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/Session.java
sbordet: The problematic combinations are still forbidden.

Both before and now you could/can have a frame and a data (either partial or whole) listener.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/InputStreamMessageSink.java
gregw: Hmmm it is a pity that we can't use our own good IO abstraction.  Surely there must be a way? Even if only for embedded usage without a webapp?

jetty-core/jetty-websocket/jetty-websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/Session.java
```diff
@@ -14,184 +14,349 @@
...
-     * @return the protocol version
+     * @param statusCode the status code
+     * @param reason the (optional) reason
+     * @param callback callback to notify when the send operation is complete
+     * @see StatusCode
+     * @see #close()
+     * @see #disconnect()
      */
-    String getProtocolVersion();
+    void close(int statusCode, String reason, Callback callback);
```
lachlan-roberts, line 156: Maybe this could use a little more javadoc.

Could be output is still open and we have received a close frame and this decides how to respond. (If no response is sent by the time the callback is succeeded the implementation will attempt to send a close frame for you).

Or if output is closed you may have just received a response close frame.

Or if there was a connection error this will be called with the status `1006` to indicate that no close handshake occurred.

jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jakarta-common/src/main/java/org/eclipse/jetty/ee9/websocket/jakarta/common/JakartaWebSocketFrameHandler.java
```diff
@@ -600,6 +594,10 @@ public void onPing(Frame frame, Callback callback)
         {
             callback.succeeded();
             coreSession.demand(1);
+        }, x ->
+        {
+            // Ignore failures, as we might be OSHUT but receive a PING.
+            callback.succeeded();
```
lachlan-roberts, line 600: Wouldn't we need to demand in this case as well?

jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jakarta-common/src/main/java/org/eclipse/jetty/ee10/websocket/jakarta/common/JakartaWebSocketFrameHandler.java
```diff
@@ -600,6 +594,10 @@ public void onPing(Frame frame, Callback callback)
         {
             callback.succeeded();
             coreSession.demand(1);
+        }, x ->
+        {
+            // Ignore failures, as we might be OSHUT but receive a PING.
+            callback.succeeded();
```
lachlan-roberts, line 600: Do we need to demand here if OSHUT?

lachlan-roberts (changes requested):
What is the reason for removing the option for the `BatchMode` in the new Jetty WebSocket API?

jetty-core/jetty-websocket/jetty-websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/Session.java
sbordet: I'm not sure what you're suggesting here?
This is the `close(int, String, Callback)` method, so this method is not called, it's the application that calls it to close.

jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jakarta-common/src/main/java/org/eclipse/jetty/ee10/websocket/jakarta/common/JakartaWebSocketFrameHandler.java
sbordet: Yes, as we might receive another frame, good catch.

jetty-core/jetty-websocket/jetty-websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/Session.java
lachlan-roberts: Meant to leave this comment on 
```
default void onWebSocketClose(int statusCode, String reason)
```

accidentally put it in the wrong place.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java
gregw: Then the javadoc here needs to explain that `autoDemand==false` does not mean the `MessageSink` will never issue demand, just that it will never issue unprovoked demand.   With `autoDemand==false` a sink may still issue demand to complete a partial message or to continue a stream.  But it will not demand if idle.

Perhaps we should call it `idleDemand` here rather than `autoDemand`?

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/PartialStringMessageSink.java
gregw: more needed on autoDemand.  See comments elsewhere

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/InputStreamMessageSink.java
gregw: MessageSink is part of jetty-websocket-core-common, so it already depends on jetty-io and can see `Content.Source`.

I understand that it can't be in jetty-api, but then no MessageSink is.   I don't see any reason that we can't include a `ContentSourceMessageSink`

sbordet:
> What is the reason for removing the option for the BatchMode in the new Jetty WebSocket API?

`BatchMode` was never used as a tri-state, but only as a `boolean`, see Jetty 11's `JettyWebSocketRemoteEndpoint.isBatch()`.

In turn that was passed to WebSocket core (which only exposes a `boolean` API for batch), and eventually arrives to `FrameFlusher`.
In `FrameFlusher`, if `batch==false`, the frame is always written to the network; if `batch==true` it may or may not be written to the network, depending on internal heuristics.

So for simplicity I removed it, hardcoding the calls to WebSocket core with `batch==false`; because if an application sets it to `true` expecting the frame to not be written to the network, it is not what will happen in all cases.
It would also be very difficult to document.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/InputStreamMessageSink.java
sbordet: > I don't see any reason that we can't include a ContentSourceMessageSink

Because we cannot expose `Content.Source` to the web application (see above).

All the other `MessageSink` expose JDK classes such as `ByteBuffer`, `String`, `InputStream`, etc.

We may want to expose `java.util.concurrent.Flow.Publisher`, but that's different from `Content.Source`.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java
sbordet: I have updated the javadocs, and kept `autoDemand` -- don't think `idleDemand` is right.

jetty-core/jetty-websocket/jetty-websocket-jetty-api/src/main/java/org/eclipse/jetty/websocket/api/Session.java
sbordet: I think we need to discuss what to document here, as this method is not clearly defined.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/InputStreamMessageSink.java
gregw: MessageSink interface and implementations are in jetty-websocket-core-common, which already depends on jetty-io, so whatever uses message sinks already can see `Content.Source`.

So embedded usage will be able to use it.     

I'm OK with not exposing it in the API... although exposing Flow in the api is a good idea and that can use the ContentSourceMessageSink behind the scenes in the implementation.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java
gregw: It is better, but a bit wordy and not exactly clear.  Also no documentation on the actual constructor.   How about:
```java
/**
 * @param autoDemand If true, the user of the MessageSink need not (and must not) use the 
 *                CoreSession.demand method. The MessageSink implementation will manage 
 *                demand so that a single message will be delivered at once and frames for the next
 *                message will only be demanded after completion of the handling of the previous
 *                message.
 *                If false, the user of the MessageSink must use the CoreSession.demand 
 *                method to initiate handling of a new message. The MessageSink implementation
 *                must manage demand itself only when non final frames are received and 
 *                partial messages are not being delivered to the user of the MessageSink.
 */

Comment (sbordet) [reply jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java:56]: There is no "user of the MessageSink", so it's still not clear.
And there are too many assumptions about what the `MessageSink` does (not all of them accumulate into messages).

What do you find not clear in the current version?

Comment (gregw) [reply jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java:47]: If there is no "user" then who calls demand when not in auto demand mode?  If the class is not used then delete it!

Comment (gregw) [reply jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java:56]: @sbordet what I find unclear about the current is:
* the javadoc is on the class and not on the parameter in the constructor.
*  it is over 4 paragraphs with references to "the former case" and "the later case" indicating structural complexity
* what does it mean for a methodHandle to explicitly demand? A method handle just invokes another method, which itself might demand or delegate the demand to another method or another process.   Demand is not managed by the method handle, it is managed by something using the MessageSink.

There are two parties to this relationship: the application as represented by the methodHandler; the implementation of the MessageSink.  Let's just call them "application" (what I called user) and ("messageSink").

There are two modes: autoDemand==true/false.    So we just need to be crystal clear about what the two parties can/can't/should do in each mode

autoDemand==true : The application need not and must not call demand.  The messageSink will manage all demand to a single stream of messages.

autoDemand==false:  The application need to call demand to commence reading of a message.  However the messageSink will then manage demand, if need be, so that sufficient frames can be received to invoke the application.  If  the messageSink is delivering complete message to the application, then it will manage demand when non final frames are received.  Once a message (partial or complete) has been delivered to the application, then the application must use demand in order to receive the next message (partial or complete)

Comment (sbordet) [reply jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java:56]: > it is over 4 paragraphs with references to "the former case" and "the later case" indicating structural complexity

It's a boolean parameter, so I need to enumerate 2 cases. What structural complexity are you talking about?

> what does it mean for a methodHandle to explicitly demand?

Nowhere in the javadoc is written that "MethodHandle to explicitly demand".

Application writers will never look at these javadocs, they are not even aware `MessageSink` exists.
These javadocs are for implementers of `MessageSink`, so IMHO it is ok to explain in the class what `autoDemand` is rather than having many paragraphs of explanations in the constructor parameter, to be repeated in the getter and in `autoDemand()`.

Part of what you explain above may be added to documentation for application writers, but it is irrelevant and incomplete for `MessageSink` implementers.

The current javadocs are explicitly tailored for `MessageSink` implementers about how to write a `MessageSink` implementation correctly not about what applications can or cannot do.

Comment (gregw) [reply jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java:56]: The text is structurally complex.  When reading one paragraph and it says "the former case", you have to look back at the previous paragraph and see what that is.    Just enumerate the two cases rather than spread it out over multiple paragraphs.

Your current text says: "The application function passed to the constructor as a MethodHandle can either explicitly demand for more WebSocket frames ...",  so yeah technically is says the application function is demanding, but it's hard to read and overly complex structure.

At the end, I'm just saying that I think I well understand how demand is meant to work for message sinks in both modes, but that I find the javadoc does not well express that, nor is it in the right place.   I'm constantly re-reading it, getting lost/confused, re-reading the bit I just read before.  It does not express what I understand the two modes to be.

If you don't like my suggestions, then at the very least can you:
 + at least have some javadoc to the parameter in the constructor that has a proper link to a section that clearly describes autoDemand
 + explain it in two cases: what true means and what false means, without the need to refer back and forth between the paragraphs.
 + your example is good, but it is unclear which mode (or both) it relates to. Perhaps have examples for both the true and false case.

Comment (sbordet) [reply jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java:56]: @gregw updated javadocs.

File: jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java [hunk 20]
```diff
@@ -18,14 +18,82 @@
...
+ * implementation -- and therefore {@code autoDemand==false}, or explicitly demands
+ * for more WebSocket frames (via APIs that eventually call
+ * {@link CoreSession#demand(long)}) -- and therefore {@code autoDemand==true}.</p>
+ * <p>{@link MessageSink} implementations must handle the demand for WebSocket
+ * frames in this way:</p>
+ * <ul>
+ * <li>When {@code autoDemand==false}, the {@link MessageSink} manages the
+ * demand until the conditions to invoke the application function are met;
+ * when the {@link MessageSink} invokes the application function, then the
+ * application function is responsible to demand for more WebSocket frames.</li>
```
gregw:
```suggestion
 * <li>When {@code autoDemand==false}, the application must initiate demand
 * for a message frame and then the {@link MessageSink}, manages the
 * demand until the conditions to invoke the application function are met;
 * at which point the {@link MessageSink} invokes the application function; and then
 * the application is again responsible for initiate any demand for subsequent message frames.</li>
```

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java
```diff
@@ -18,14 +18,82 @@
 
 import org.eclipse.jetty.websocket.core.CoreSession;
 
+/**
+ * <p>Partial implementation of {@link MessageSink}.</p>
```
gregw: Partial suggests partial message
```suggestion
 * <p>Abstract implementation of {@link MessageSink}.</p>
```

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java
```diff
@@ -18,14 +18,82 @@
...
+ * <p>The application function passed to the constructor as a {@link MethodHandle}
+ * either delegates the demand for more WebSocket frames to the {@link MessageSink}
+ * implementation -- and therefore {@code autoDemand==false}, or explicitly demands
+ * for more WebSocket frames (via APIs that eventually call
+ * {@link CoreSession#demand(long)}) -- and therefore {@code autoDemand==true}.</p>
+ * <p>{@link MessageSink} implementations must handle the demand for WebSocket
+ * frames in this way:</p>
```
gregw: It is not the "application function" that manages demand, but the "application".  The "function" might not even be invoked before the initial demand is required to be issued by the application.

```suggestion
 * <p>Management of demand for message frames may either be entirely managed
 * by the {@link MessageSink} implementation ({@code autoDemand==true}; or
 * it may be managed collaboratively between the application and the 
 * {@link MessageSink} implementation ({@code autoDemand==true}.
 * <p>{@link MessageSink} implementations must handle the demand for WebSocket
 * frames in this way:</p>
```

gregw (changes requested):
Much better... but still needs work.  Sorry to be pedantic, but I think this one is important to get very clear.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java
sbordet: Nope it's wrong.

What you suggest is completely irrelevant from the point of view of the `MessageSink`.
Does not matter to document here how the `MessageSink` is invoked the first time, and less so to document here how "application" code must behave.
For these javadocs needs to be documented how the `MessageSink` manages the demand after it is invoked the first time.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java
sbordet: > It is not the "application function" that manages demand, but the "application". 

It is that particular application function that manages demand, not some undefined "application".
Also, the demand is not for "message frames" (they don't exist), but for generic WebSocket frames (including CLOSE).

I have reworded into a mix of the two.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java
```diff
@@ -18,14 +18,82 @@
 
 import org.eclipse.jetty.websocket.core.CoreSession;
 
+/**
+ * <p>Abstract implementation of {@link MessageSink}.</p>
+ * <p>Management of demand for WebSocket frames may either be entirely managed
+ * by the {@link MessageSink} implementation ({@code autoDemand==true}); or
+ * it may be managed collaboratively between the application function passed
```
gregw: OK I'm completely stuck on this.  I find the  "passed application function" entirely confusing. Specifically:
 + how is demand initiated in the first place if that function has not yet been invoked.
 + must demand only be managed by that function? i.e. can demand be called sometime after that function returns?

It makes no sense to me?    If the usage of the messageSink is not using autoDemand, then it is likely that demand management is complex and specific to all kind of details within the application, not just the invocation of that one function.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java
gregw: why do you think the initial demand is irrelevant?   To me it is a big difference between the modes: one does the initial demand and the other doesn't.

What is wrong with my version?  I think it says much the same as yours (but better) and includes the initial demand.  

At the very least, change your version to not use "when" in two different senses.  "when the messageSink invokes..." reads like and alternative to "when autodemand==false".      Use "then" or some other language that implies a sequence of events.

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/messages/AbstractMessageSink.java
gregw: Resovling this big conversation in favour of the micro discussions above.

gregw (changes requested):

sbordet:
@gregw after a successful WebSocket upgrade, the initial demand is triggered by the "open" event (which means methods annotated with `@OnWebSocketOpen`, or `Session.Listener.onWebSocketOpen()`).
At this point, the `MessageSink` is not in the picture at all: the call to the "open" method does not go through the message sink.

The initial demand may either be automatic or not.
If automatic, the `FrameHandler` is responsible for it, not the `MessageSink`.

_After_ the initial demand, the `MessageSink` will be invoked with _data_ frames (not _control_ frames).
The `MessageSink` calls the application function depending on the `MessageSink` implementation logic.

If the WebSocket endpoint is _not_ autodemanding, then the demand is handled by the application function, which either must demand synchronously, or arrange to demand later asynchronously.
Either case, it is managed by the application function, not by `MessageSink`.

Demand for _control_ frames is managed either by the `FrameHandler` or by other application functions (e.g. for PING, PONG, CLOSE, etc.).

gregw (approved):
I'd still like to see a `ContentSourceMessageSink` that is the implementation behind a `Flow.Publisher` API.  Can you add a TODO for that and we can add in another PR

sbordet:
@gregw I opened https://github.com/eclipse/jetty.project/issues/9722.

lachlan-roberts (approved):

joakime (approved):
:+1:

PR #9664, opened 2023-04-19
Add tests for combining of Constraints as per spec

janbartel:
Add test for all combinations of types of `auth-constraint` as per the servlet spec section 13.8.1


gregw (approved):

janbartel:
I'm holding off on merging this through, as the merge into jetty-12.0.x will be wasted effort until @gregw completes his massive core security check in.

PR #9668, opened 2023-04-24
Issue #9649 ServletApiResponse.addIntHeader must ignore headers after commit

### LINKED ISSUE #9649: jetty-12 ee10 ServletApiResponse.addIntHeader does not ignore headers after response committed
The javadoc for `HttpServletResponse.sendError` says:

`After using this method, the response should be considered to be committed and should not be written to.`

The tck test:

> com/sun/ts/tests/servlet/spec/httpservletresponse/URLClient.java.sendErrorCommitTest

does a `sendError` followed by setting some headers:

```
    PrintWriter pw = response.getWriter();
    response.sendError(HttpServletResponse.SC_OK, "sendErrorCommitTest");
    response.addIntHeader("header1", 12345);
```

 The test expects the header not to be set, but fails because we are allowing the headers to be set after `sendError`.

janbartel:
Closes #9649


joakime:
Do we have a testcase where we test `response.addIntHeader()` after normal commit (eg: `response.flush()`)? (we have the version after `sendError()`)

janbartel:
@joakime no, we don't have a specific test for that, but as the code in `addIntHeader` specifically checks `isCommitted`, I don't think we need a test for `addIntHeader` after `flush`. We should have a test for `flush` correctly setting the committed state however.

joakime (approved):

PR #9670, opened 2023-04-24
Jetty 12 - Fix `HostHeaderCustomizer` and improve tests

joakime:
Re-enable tests for `HostHeaderCustomizer` and improve scope of operation

### DIFF
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HostHeaderCustomizer.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HostHeaderCustomizer.java
@@ -19,7 +19,6 @@ import org.eclipse.jetty.http.HttpHeader;
 import org.eclipse.jetty.http.HttpScheme;
 import org.eclipse.jetty.http.HttpURI;
 import org.eclipse.jetty.http.HttpVersion;
-import org.eclipse.jetty.util.HostPort;
 
 /**
  * Adds a missing {@code Host} header (for example, HTTP 1.0 or 2.0 requests).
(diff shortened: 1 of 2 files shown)


jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HostHeaderCustomizer.java
```diff
@@ -70,6 +71,17 @@ public Request customize(Request request, HttpFields.Mutable responseHeaders)
...
+        ConnectionMetaData connectionMetadata = new ConnectionMetaData.Wrapper(request.getConnectionMetaData())
+        {
+            private HostPort _serverAuthority = new HostPort(host, port);
+
+            @Override
+            public HostPort getServerAuthority()
+            {
+                return _serverAuthority;
+            }
+        };
```
sbordet: I don't think this is correct.
This customizer only needs to add the `Host` header, not change the server authority.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/HostHeaderCustomizerTest.java
```diff
@@ -34,25 +36,38 @@ public void testFixedHostPort() throws Exception
...
+
+                // Test "getHttpURI()"
+                HttpURI httpURI = request.getHttpURI();
+                assertEquals(serverName, httpURI.getHost());
+                assertEquals(serverPort, httpURI.getPort());
+
+                // Test "connectionMetaData.serverAuthority"
+                HostPort serverAuthority = request.getConnectionMetaData().getServerAuthority();
+                assertEquals(serverName, serverAuthority.getHost());
+                assertEquals(serverPort, serverAuthority.getPort());
```
sbordet: I don't think this is right.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/HostHeaderCustomizerTest.java
```diff
@@ -95,21 +109,35 @@ public void testHostPort() throws Exception
...
+
+                // Test "getHttpURI()"
+                HttpURI httpURI = request.getHttpURI();
+                assertEquals(serverName, httpURI.getHost());
+                assertEquals(connector.getLocalPort(), httpURI.getPort());
+
+                // Test "connectionMetaData.serverAuthority"
+                HostPort serverAuthority = request.getConnectionMetaData().getServerAuthority();
+                assertEquals(serverName, serverAuthority.getHost());
+                assertEquals(connector.getLocalPort(), serverAuthority.getPort());
```
sbordet: I don't think this is right.

sbordet (changes requested):

sbordet (approved):

Issue #9637, opened 2023-04-11
jetty-12 ee10 ServletRequestListeners called too many times on sendError


janbartel:
**jetty-12**

Servlet tck error: https://jenkins.webtide.net/job/tck/job/tck-servlet-arquillian-experiment/job/jetty-12-ee10/119/testReport/junit/com.sun.ts.tests.servlet.spec.srlistener/URLClient/error/

The test sends a request to a servlet that does a `sendError(403)`, and then sends another request to check how many times `ServletRequestListener.requestInitialized` and `ServletRequestListener.requestDestroyed` have been called. The test is expecting 3 calls:  an `initialize` and `destroy` for the first request that sends the error, and then another `initialize` for the request that is checking the number of calls. The problem is that instead of 3 calls, we do 5: there is an extra `initialize` and `destroy` pair present before the last `initialize`.

The problem appears to be that `ServletChannel.dispatch` always calls `ServletContextHandler.requestInitialized`.  This means that the first `handle` loop will call `dispatch` (which calls `requestInitialized`), but then the second `handle` loop for the `SEND_ERROR`  case  _also_ calls `dispatch`, leading to the second call to `ServletContextHandler.requestInitialized`.

See line 656 of ServletChannel.dispatch.

Note: this may potentially affect ee9 too?

janbartel:
Here's the relevant test failure output:


```
[main] ERROR org.eclipse.jetty.tck.AppTest - Test: 'checkLogSimple' failed.
java.lang.Exception:  Test Test Case Unable to find the following search string in the server's response: 
'Test PASSED'
 at index: 0
 Server's response:
-------------------------------------------
RequestListener is invoked/deleted more or less times than requiredExpect 3 times, got 5.Got expected value: in requestInitialized method of listener at position 0Got expected value: in requestDestroyed method of listener at position 1Got expected value: in requestInitialized method of listener at position 2Extra access to Listener: in requestDestroyed method of listener at position 3Extra access to Listener: in requestInitialized method of listener at position 4Test FAILED

-------------------------------------------
```

janbartel:
The fix for this will also most likely fix:

>  com/sun/ts/tests/servlet/spec/srlistener/URLClient.java.forwarderror
>  com/sun/ts/tests/servlet/spec/srlistener/URLClient.java.includeerror
>  com/sun/ts/tests/servlet/spec/srlistener/URLClient.java.simpleasyncerror

gregw:
[...]

PR #9677, opened 2023-04-26
Issue #9637 - Update behaviour and add testing for ServletRequestListener

lachlan-roberts:
## Issue https://github.com/eclipse/jetty.project/issues/9637

Add testing for `ServletRequestListener` under various dispatch types and change behaviour so that if there is no error page the `ServletRequestListener` is never called.

I am not sure if this will pass this failing TCK test so we may need to raise a challenge. If the test is dispatching to an error page this will not pass as we will be calling the `ServletRequestListener` again.


lachlan-roberts:
@gregw I have added the test you asked for, but I did not move the calls to the listener to the `MappedServlet.handle` as then the listener will be called multiple times for include and forward dispatches.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -670,6 +670,26 @@ private void dispatch(Dispatchable dispatchable) throws Exception
         }
     }
 
+    private void errorDispatch(Dispatchable dispatchable) throws Exception
```
gregw: I'm not a fan of these two dispatch methods, that are substantially the same.

Whilst this is used for error dispatching, the name really doesn't say what it does.
This really should be called `dispatch` and the other `dispatchWithServletListeners`

Or maybe just have one `dispatch` method that takes a boolean saying if servlet listeners should be called or not.

Or maybe one method calls the other?


Best option would be to arrange for the listeners to be called elsewhere as we discussed... that didn't work??

gregw: Also... remind me how servlet listeners are meant to be called for async requests?  If startAsync is called, should we be calling destroy in the finally block?

gregw (changes requested):

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
lachlan-roberts: > Best option would be to arrange for the listeners to be called elsewhere as we discussed... that didn't work??

I left a comment above describing why this cannot be done as we discussed before, it really needs to be done in the ServletChannel because during include and forward dispatches we don't want to be calling the listener multiple times.

I will add the boolean for this method.

> Also... remind me how servlet listeners are meant to be called for async requests? If startAsync is called, should we be calling destroy in the finally block?

No not when startAsync is called, only when it exits the filterChain.
```
A ServletRequest is defined as coming into scope of a web application when it is about 
to enter the first servlet or filter of the web application, and as going out of scope as it 
exits the last servlet or the first filter in the chain.
```

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ErrorHandler.java
```diff
@@ -106,7 +108,8 @@ public boolean handle(Request request, Response response, Callback callback) thr
         {
             try
             {
-                errorDispatcher.error(servletContextRequest.getHttpServletRequest(), servletContextRequest.getHttpServletResponse());
+                contextHandler.requestInitialized(servletContextRequest, httpServletRequest);
```
janbartel, line 111: I don't see why we would do a `requestInitialized` here. That would mean we're doing it every time we want to send an error via an error page, _even if_ `requestInitialized` has already been called previously?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -423,7 +423,7 @@ public boolean handle()
                             ServletHandler.MappedServlet mappedServlet = _servletContextRequest._mappedServlet;
```
janbartel: Can't we put the `try/finally` of calling the listeners inside this block?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -465,7 +465,7 @@ public boolean handle()
 
                             Dispatcher dispatcher = new Dispatcher(getContextHandler(), uri, decodedPathInContext);
                             dispatcher.async(asyncContextEvent.getSuppliedRequest(), asyncContextEvent.getSuppliedResponse());
-                        });
+                        }, true);
```
janbartel: Can't we put the `try/finally` for calling the listeners inside the `dispatch() ->` block instead of inside `dispatch()` method?

janbartel (changes requested):

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ErrorHandler.java
lachlan-roberts: We can't call `requestInitialized`/`requestDestroyed` for errors from the `ServletChannel` because we don't know if we will dispatch to an error page or not. 

So we avoid calling it from the `SEND_ERROR` state and defer it to the `ErrorHandler` to decide to call it. 

There should be no way for `requestInitialized` to be previously called here, as we specifically didn't call it in `ServletChannel.dispatch`.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
lachlan-roberts: We can.

I thought it was neater with the boolean, because then only 1 try/finally and it stays close to the notification of the combinedListener. 
But I don't mind will change it to this.

lachlan-roberts:
@janbartel @gregw can you please re-review

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ErrorHandler.java
gregw: Perhaps a comment her referencing the javadoc that says we call listeners before/after entering/exiting first filter/servlet of a dispatch?

gregw (approved):

PR #9684, opened 2023-04-27
Refine how Request / Channel / Stream completion works

joakime:
Refine how requests get completed.

* Handling success / failures
* Writing success / failures / last-write
* Callback success / failures
* Error handling cases
* Bad Content-Length during error handling
* Introduce new tests to cover the various combinations that can impact request completion.


gregw:
@sbordet @joakime this is now green, so I think we should consider merging soon so it does not go stale.  There are still somethings that need looking at, but it is better than what is there now.

gregw (approved):
I'm approving, but it is mostly my own work.  Got a thumbs up from @joakime, so that is an OK from him (his PR, so can't approve). @sbordet OK for a merge?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
```diff
@@ -556,6 +556,7 @@ else if (_responseHandler != null && buffer.remaining() >= (HTTP_VERSION_LEN + 1
                 case OTEXT:
                 case SPACE:
                 case HTAB:
+                    System.err.println(BufferUtil.toDetailString(buffer));
```
sbordet: Leftover?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpHeader.java
```diff
@@ -132,6 +133,14 @@
...
     C_PROTOCOL(":protocol");
 
+    public static final EnumSet<HttpHeader> CONTENT_HEADERS = EnumSet.of(
+        CONTENT_TYPE, CONTENT_LENGTH, CONTENT_ENCODING, CONTENT_LANGUAGE, CONTENT_RANGE, CONTENT_MD5, CONTENT_LOCATION, TRANSFER_ENCODING, CACHE_CONTROL, LAST_MODIFIED, EXPIRES, VARY, ETAG
+    );
+
+    public static final EnumSet<HttpHeader> CONTENT_HEADERS_304 = EnumSet.of(
+        CONTENT_TYPE, CONTENT_LENGTH, CONTENT_ENCODING, CONTENT_LANGUAGE, CONTENT_RANGE, CONTENT_MD5, CONTENT_LOCATION, TRANSFER_ENCODING, CACHE_CONTROL, LAST_MODIFIED, EXPIRES, VARY
+        );
+
```
sbordet, line 147: Reduce duplication via `EnumSet.copyOf(CONTENT_HEADERS_304).add(ETAG)`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -77,27 +77,27 @@
...
-    private static final Throwable DO_NOT_SEND = new Throwable("No Send");
-    private static final MetaData.Request ERROR_REQUEST = new MetaData.Request("GET", HttpURI.from("/"), HttpVersion.HTTP_1_0, HttpFields.EMPTY);
-    private static final HttpField SERVER_VERSION = new PreEncodedHttpField(HttpHeader.SERVER, HttpConfiguration.SERVER_VERSION);
-    private static final HttpField POWERED_BY = new PreEncodedHttpField(HttpHeader.X_POWERED_BY, HttpConfiguration.SERVER_VERSION);
-
     /**
      * The state of the written response
      */
-    enum WriteState
+    enum StreamSendState
```
sbordet: Make it `private`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -77,27 +77,27 @@
...
     {
-        /** Not yet written */
-        NOT_LAST,
+        /** Last send not yet sent */
```
sbordet:
```suggestion
        /** Last content not yet sent */
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -443,41 +452,62 @@ private void resetResponse()
         }
     }
 
-    private Throwable lockedCheckWrite(boolean last, long length)
+    private Throwable lockedStreamSend(boolean last, long length)
     {
         assert _request._lock.isHeldByCurrentThread();
```
sbordet: Just use `_lock` here, not `_request._lock`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -443,41 +452,62 @@ private void resetResponse()
...
+            // they are flushed. The DO_NOT_SEND option supports these by turning such writes into a NOOP.
+            case LAST_SENDING, LAST_COMPLETE -> (length > 0)
                 ? new IllegalStateException("last already written")
                 : DO_NOT_SEND;
         };
     }
 
+    private void lockedStreamSendCompleted(boolean success)
+    {
+        assert _request._lock.isHeldByCurrentThread();
```
sbordet: Just use `_lock` here, not `_request._lock`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -443,41 +452,62 @@ private void resetResponse()
...
+        if (_streamSendState == StreamSendState.LAST_SENDING)
+            _streamSendState = success ? StreamSendState.LAST_COMPLETE : StreamSendState.SENDING;
+    }
+
+    private boolean lockedIsLastStreamSendCompleted()
+    {
+        assert _request._lock.isHeldByCurrentThread();
```
sbordet: Just use `_lock` here, not `_request._lock`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -443,41 +452,62 @@ private void resetResponse()
...
+        return _streamSendState == StreamSendState.LAST_COMPLETE;
+    }
+
+    private boolean lockedLastStreamSend()
+    {
+        assert _request._lock.isHeldByCurrentThread();
```
sbordet: Just use `_lock` here, not `_request._lock`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -974,18 +996,57 @@ public String toString()
...
+            // remembering any server or date headers (probably a nicer way of doing this).
+            HttpField serverField = _request._httpChannelState._responseHeaders.getField(HttpHeader.SERVER);
+            HttpField dateField = _request._httpChannelState._responseHeaders.getField(HttpHeader.DATE);
+            _request._httpChannelState._responseHeaders.reset();
+            _request._httpChannelState._committedContentLength = -1;
+            reset();
+            if (serverField != null)
+                _request._httpChannelState._responseHeaders.put(serverField);
+            if (dateField != null)
+                _request._httpChannelState._responseHeaders.put(dateField);
```
sbordet: Reduce duplication by storing `_request._httpChannelState` into a local variable.
Also, you can't directly access it, but you must call `_request.lockedGetHttpChannelState()` because it throws ISE in case the response is used after recycle -- here would throw NPE.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -1213,6 +1292,12 @@ private MetaData.Response lockedPrepareResponse(HttpChannelState httpChannel, bo
                 getTrailersSupplier()
             );
         }
+
+        @Override
+        public String toString()
+        {
+            return "%s@%x{%s %s@%x}".formatted(this.getClass().getSimpleName(), hashCode(), getRequest().getMethod(), getRequest().getHttpURI(), getRequest().hashCode());
```
sbordet: This is all request data, but this is the response's `toString()`. I would at least put in the status code and HTTP version, and then perhaps also add the request's `toString()`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpConnection.java
```diff
@@ -971,8 +971,6 @@ protected void onCompleteSuccess()
         public void onCompleteFailure(final Throwable x)
         {
             failedCallback(release(), x);
-            if (_shutdownOut)
-                getEndPoint().shutdownOutput();
```
sbordet, line 975: Why was this removed? Is it done elsewhere?
We want to gently close in case of TLS.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/CustomRequestLogTest.java
```diff
@@ -675,13 +670,50 @@ public boolean handle(Request request, Response response, Callback callback)
...
+
+            OutputStream output = socket.getOutputStream();
+            output.write("""
+                GET /abort HTTP/1.1
+                Host: localhost
+                    
+                """.getBytes(StandardCharsets.ISO_8859_1));
+            output.flush();
+
+            BufferedReader in = new BufferedReader(new InputStreamReader(socket.getInputStream(), StandardCharsets.ISO_8859_1));
```
sbordet, line 703: Can we use `HttpTester` instead?

sbordet (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpHeader.java
gregw: yeah... but not so nice as add returns a boolean. But OK.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/CustomRequestLogTest.java
gregw: No, because we want to test that no last chunk is sent.    I've added comments to that effect.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpConnection.java
gregw: @sbordet  This was done by @joakime before I "hijacked" the branch.  I think it was done during a collective debugging sessions, resolving the initial race condition, where one thread was shutting does the endpoint whilst another was trying to write out the error page.

Can you more precisely describe the use-case you are concerned about and I'll double check that we test for it and see who is doing the shutdown.

sbordet: The `SHUTDOWN` result is only returned after successful writes (and therefore another iteration on the generator), so I guess it's only a concern on the succeeded path.

I think it's done in 2 places now, `SendCallback.onCompleteSuccess()` and then again in `HTTPStreamOverHTTP1.succeeded()`, so I think the former can be removed.

For the failure case we should always end up in `HTTPStreamOverHTTP1.failed()` so we `abort()` and that should be fine.

sbordet (changes requested):
I'd remove the `shutdownOutput()` call from `SendCallback.onCompleteSuccess()` and see how it goes -- that's the last nit from me.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpConnection.java
gregw: done.

sbordet (approved):

PR #9687, opened 2023-04-29
Fix #9685 Date is not a content header

### LINKED ISSUE #9685: Jetty doesn't set the date header on error responses
**Jetty version(s)**
11.0.15

**OS type/version**

Linux 5.4.0-147-generic

**Description**

Jetty's default error responses don't include the `date` header, but [RFC 9110 specifies that it's mandatory](https://www.rfc-editor.org/rfc/rfc9110.html#field.date) for an "origin server with a clock".

**How to reproduce?**

Run the following class:

```java
package com.example;

public class ErrorServer {
        new Server(8080).start();
    }
}
```
Navigate to http://localhost:8080 in a browser, and view the response headers.  I get:

```
Cache-Control: must-revalidate,no-cache,no-store
Content-Length: 435
Content-Type: text/html;charset=iso-8859-1
Server: Jetty(11.0.15)
```
This appears to be caused by a call to [`Response.resetContent()`](https://github.com/eclipse/jetty.project/blob/jetty-11.0.15/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java#L1225) in the default error handling flow, which erases the `date` header that had previously been set by [`HttpChannel.onRequest(MetaData.Request)`](https://github.com/eclipse/jetty.project/blob/jetty-11.0.15/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java#L892).

gregw:
Fix #9685 resetContent does not reset Date Header

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/Response.java
@@ -1253,7 +1253,6 @@ public class Response implements HttpServletResponse
                 case LAST_MODIFIED:
                 case EXPIRES:
                 case ETAG:
-                case DATE:
                 case VARY:
                     i.remove();
                     continue;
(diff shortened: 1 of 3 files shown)


sbordet (approved):
Remember to merge into 3 places when merging to 12: core, ee10 and ee9.

gregw:
Merged 10->11>12ee9->12ee10   However, there is still an issue in 12, see original issue for discussion.

PR #9715, opened 2023-05-01
deprecate PushSessionCacheFilter

lachlan-roberts:
deprecate PushSessionCacheFilter, will remove in 12 during the merge.

Addresses CVE-2024-6762


jetty-servlets/src/main/java/org/eclipse/jetty/servlets/PushSessionCacheFilter.java
```diff
@@ -37,6 +37,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+@Deprecated
```
joakime, line 43: Should really have a javadoc also ...

``` java
/**
 * @deprecated no replacement for this deprecated http feature
 */
```

joakime (dismissed):

gregw (approved):

PR #9718, opened 2023-05-01
Fix #9686 Configurations usage in websocket test

### LINKED ISSUE #9686: Configurations should not cache Configuration instances
**Jetty version(s)**
10+

**Description**
`oej.eeN.webapp.Configurations.getKnown()` caches the instances of the known and available configurations.

Unfortunately this is wrong because a few of the `Configuration` implementations are stateful.

For example, `WebInfConfiguration` has the non-static field `_preUnpackBaseResource`.

When deploying multiple web applications, the same `WebInfConfiguration` instance is returned by `Configurations`, so the `_preUnpackBaseResource` field is overwritten multiple times, keeping only the last value.

`Configuration` instances should be different for every `WebAppContext` so they cannot be cached by `Configurations` (or only cache those that are stateless).

gregw:
Fix #9686  the tests usage of configurations.  Probably need to improve the configuration API to make this kind of error impossible


joakime (approved):

sbordet (approved):
Magic 🤯

PR #9669, opened 2023-04-24
Issue #9648 ServletApiResponse.sendError must check if already committed

janbartel:
Closes #9648

PR #9719, opened 2023-05-01
Fix #9685 & #9648 test isCommitted in sendError

### LINKED ISSUE #9648: jetty-12 ee10 ServletApiResponse.sendError does not check for the response already being committed
The javadoc for `HttpServletResponse.sendError` says:

```
  * If the response has already been committed, this method throws an IllegalStateException. After using this method, the
  * response should be considered to be committed and should not be written to.
```

Jetty is not checking if the response has already been committed and thus never throws the `IllegalStateException`.

Tck tests affected:

> com/sun/ts/tests/servlet/api/jakarta_servlet_http/httpservletresponse/URLClient.java.sendErrorIllegalStateExceptionTest
> com/sun/ts/tests/servlet/api/jakarta_servlet_http/httpservletresponse/URLClient.java.sendError_StringIllegalStateExceptionTest
> com/sun/ts/tests/servlet/pluggability/api/jakarta_servlet_http/httpservletresponse/URLClient.java.sendErrorIllegalStateExceptionTest
> com/sun/ts/tests/servlet/pluggability/api/jakarta_servlet_http/httpservletresponse/URLClient.java.sendError_StringIllegalStateExceptionTest

### LINKED ISSUE #9685: Jetty doesn't set the date header on error responses

gregw:
Merging #9685 to 12 broke the build as it turned out that the error handling was depending on the attempt to remove the Date header to detect if the response was already committed during a sendError.   This fix makes the test for committed explicit, it also fixes which exception is passed to abort.

### DIFF
--- a/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiResponse.java
+++ b/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiResponse.java
@@ -227,7 +227,12 @@ public class ServletApiResponse implements HttpServletResponse
                     }
                 }
             }
-            default -> _response.getState().sendError(sc, msg);
+            default ->
+            {
+                if (isCommitted())
+                    throw new IllegalStateException("Committed");
+                _response.getState().sendError(sc, msg);
+            }
         }
     }
 
--- a/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
+++ b/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
@@ -480,48 +480,41 @@ public class ServletChannel
                         Throwable cause = errorException instanceof Throwable throwable ? throwable : null;
                         try
                         {
-                            if (_state.isResponseCommitted())
+                            // Get ready to send an error response
+                            getResponse().resetContent();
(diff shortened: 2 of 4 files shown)


joakime (approved):
+1

gregw:
This fix should also fix #9648 and replaces #9669

gregw:
Hmmm let me see if I can merge #9669 to get the unit test.

PR #9724, opened 2023-05-02
Jetty 12 immutable ee10 configurations

gregw:
From #9686 making Configurations immutable


gregw:
@janbartel would be good to get review and/or help as I go on this one.

gregw:
@janbartel nudge

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/IncludeExcludeSet.java
```diff
@@ -68,6 +61,12 @@ public IncludeExcludeSet()
         this(HashSet.class);
     }
 
+    public IncludeExcludeSet<T, P> asImmutable()
+    {
+        // TODO
```
janbartel: A new TODO in new code??!

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/Configurations.java
```diff
@@ -176,7 +176,7 @@ public static Configurations setServerDefault(Server server)
             return configurations;
         configurations = getServerDefault(server);
         server.addBean(configurations);
-        server.setAttribute(Configuration.ATTR, null);
+        server.setAttribute(SERVER_DEFAULT_ATTR, null);
```
janbartel, line 179: This is really not _server_ defaults, they are specifically _environment_ defaults. If an ee9 context setup was to call this method, and then an ee10 called the same method, one would overwrite the other with meaningless class references because they are environment classloader specific.  I'd rather this was something like:

`setEnvironmenDefault(Environment)`

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/AbstractConfiguration.java
```diff
@@ -16,131 +16,148 @@
...
+                    throw new IllegalArgumentException();
+
+                _system.add(c);
+                _server.add("-" + c);
+            }
+            return this;
         }
     }
 
+    protected AbstractConfiguration(Builder builder)
```
janbartel, line 152: This is a bit of a weird builder pattern. Normally you don't construct the object by passing it a builder, you use the builder to construct the object. Can't these fields just be `protected final` blank in `AbstractConfiguration`?

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/MetaInfConfiguration.java
```diff
@@ -605,7 +597,7 @@ public void postConfigure(WebAppContext context) throws Exception
      * @throws IOException if unable to scan the directory
      */
     // TODO: Needs to use resource.
```
janbartel: Maybe remove this TODO?

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationConfiguration.java
```diff
@@ -77,28 +74,36 @@ public class AnnotationConfiguration extends AbstractConfiguration
...
     public static final int DEFAULT_MAX_SCAN_WAIT = 60; /* time in sec */
     public static final boolean DEFAULT_MULTI_THREADED = true;
 
-    protected final List<AbstractDiscoverableAnnotationHandler> _discoverableAnnotationHandlers = new ArrayList<>();
-    protected ClassInheritanceHandler _classInheritanceHandler;
-    protected final List<ContainerInitializerAnnotationHandler> _containerInitializerAnnotationHandlers = new ArrayList<>();
-    protected final List<DiscoveredServletContainerInitializerHolder> _sciHolders = new ArrayList<>();
-
-    protected List<ParserTask> _parserTasks;
+    protected static class State
```
janbartel, line 82: Is this a `record` ?

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationConfiguration.java
```diff
@@ -412,34 +416,29 @@ private void addInheritedTypes(Set<String> results, Map<String, Set<String>> cla
...
-
-    public void addDiscoverableAnnotationHandler(AbstractDiscoverableAnnotationHandler handler)
-    {
-        _discoverableAnnotationHandlers.add(handler);
     }
 
     @Override
     public void configure(WebAppContext context) throws Exception
     {
+        State state = (State)context.getAttribute(STATE);
```
janbartel, line 436: ISE if `state` is null?

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationConfiguration.java
```diff
@@ -478,15 +477,16 @@ public void postConfigure(WebAppContext context) throws Exception
             classMap.clear();
         context.removeAttribute(CLASS_INHERITANCE_MAP);
 
-        _discoverableAnnotationHandlers.clear();
-        _classInheritanceHandler = null;
-        _containerInitializerAnnotationHandlers.clear();
-        _sciHolders.clear();
+        State state = (State)context.removeAttribute(STATE);
```
janbartel, line 475: ISE if `state` is null?

janbartel (changes requested):
Mostly fine, just minor niggles.

I need to be convinced about the `serverDefault` though.

And an alternative to the oddly inverted builder pattern?

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/Configurations.java
gregw: But we can have multiple servers in the same environment and each will have a different set of default configurations.  So I think they still are server defaults

jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/AbstractConfiguration.java
gregw: It is a little different, but I don't think it can be miss-used, so I think it is OK.

jetty-ee10/jetty-ee10-annotations/src/main/java/org/eclipse/jetty/ee10/annotations/AnnotationConfiguration.java
gregw: I was intended to do that, but I'd need to make all fields final, which is a bit of work that risks changing behaviour.

janbartel (approved):

PR #9725, opened 2023-05-02
jetty 12.0.x rename jakarta classes as well

olamy:
- use modify source plugin new snapshot which transform Jakarta classes to Javax classes
work done here https://github.com/jetty-project/jetty-modify-sources-maven-plugin/pull/61
still need a release of the maven plugin


lachlan-roberts (approved):
Looks good to me. All the ee8 websocket classes are now named Javax and not Jakarta.

PR #9727, opened 2023-05-03
Fixes #7091 - Add SOCKS5 support

### LINKED ISSUE #7091: Add SOCKS5 support

sbordet:
Fix #7091

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/Socks5Proxy.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/Socks5Proxy.java
@@ -1,0 +1,459 @@
+//
+// ========================================================================
+// Copyright (c) 1995 Mort Bay Consulting Pty Ltd and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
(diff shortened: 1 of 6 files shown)


jetty-client/src/main/java/org/eclipse/jetty/client/Socks5Proxy.java
```diff
@@ -0,0 +1,416 @@
...
+            super(endPoint, executor);
+            this.connectionFactory = connectionFactory;
+            this.context = context;
+            this.authentications = Map.copyOf(authentications);
+        }
+
+        @Override
+        public ByteBuffer onUpgradeFrom()
+        {
+            return BufferUtil.copy(byteBuffer);
```
huisongma, line 139: Before onUpgradeFrom returns, byteBuffer can be set to null so that it can be collected by GC to save memory. This is useful if user has created many connections.

sbordet: The whole connection will be GCed because it is being replaced by another.

gregw (approved):

PR #9728, opened 2023-05-04
Fixes to QPACK configuration from SETTINGS frames.

lachlan-roberts:


jetty-http3/http3-server/src/main/java/org/eclipse/jetty/http3/server/internal/ServerHTTP3Session.java
```diff
@@ -104,16 +107,88 @@ public HTTP3SessionServer getSessionServer()
...
+        {
+            if (key == SettingsFrame.MAX_TABLE_CAPACITY)
+                decoder.setMaxTableCapacity(value.intValue());
+            else if (key == SettingsFrame.MAX_FIELD_SECTION_SIZE)
+                decoder.setMaxHeadersSize(value.intValue());
+            else if (key == SettingsFrame.MAX_BLOCKED_STREAMS)
+                decoder.setMaxBlockedStreams(value.intValue());
+        });
+
+        encoder.setMaxHeadersSize(configuration.getMaxResponseHeadersSize());
```
lachlan-roberts: Can we get a comment for this line, not obvious what a setting for the encoder is doing here.

Is it possible to receive the settingsFrame from the remote side before this? if so this could override the settings set by that.

jetty-http3/http3-server/src/main/java/org/eclipse/jetty/http3/server/internal/ServerHTTP3Session.java
```diff
@@ -104,16 +107,88 @@ public HTTP3SessionServer getSessionServer()
...
+
+        // Queue the mandatory SETTINGS frame.
         SettingsFrame frame = new SettingsFrame(settings);
```
lachlan-roberts, line 159: Are we no longer waiting for both `onStart()` and `onSettings(SettingsFrame frame)` before calling `onOpen()`?

Can we even send any data frames before receiving the SETTINGS frame? we might send a response larger than what the decoder is willing to accept.

jetty-http3/http3-qpack/src/main/java/org/eclipse/jetty/http3/qpack/QpackEncoder.java
```diff
@@ -131,11 +162,16 @@ public int getCapacity()
      *
      * @param capacity the new capacity.
      */
-    public void setCapacity(int capacity)
+    public void setTableCapacity(int capacity)
     {
-        _context.getDynamicTable().setCapacity(capacity);
-        _handler.onInstructions(List.of(new SetCapacityInstruction(capacity)));
-        notifyInstructionHandler();
+        try (AutoLock ignored = lock.lock())
```
lachlan-roberts, line 167: I didn't even know there was a lock in this class. When I was writing it you said specifically not to have locks and any synchronisation will be done externally.

But since its here now you will also need to use it inside `QPackEncoder.streamCancellation` and `QPackEncoder.insert`.

jetty-http3/http3-server/src/main/java/org/eclipse/jetty/http3/server/internal/ServerHTTP3Session.java
sbordet: Good point I moved the line to the initialization of the encoder.

jetty-http3/http3-server/src/main/java/org/eclipse/jetty/http3/server/internal/ServerHTTP3Session.java
sbordet: Yes the specification explicitly states that implementations should not wait here: https://www.rfc-editor.org/rfc/rfc9114.html#section-7.2.4.2.

As long as we have an initial encoder `maxHeadersSize` that is not `0`, we are able to send.

jetty-http3/http3-qpack/src/main/java/org/eclipse/jetty/http3/qpack/QpackEncoder.java
sbordet: The decoder can be lock free because reads are serialized, and it only reads.

But for the encoder, it can called by the application (write-like operation) at the same time it is reading instructions, so there is the need for internal locking.

Added locking to `streamCancellation()` and `insert()` too, although the latter is only used for testing.

lachlan-roberts: I think we could possibly be a lot more efficient with the locks and lock only when we need to make a modification to the table.

But not something for this PR anyway.

lachlan-roberts:
@sbordet looks good to me, but I can't review with GitHub because it is my PR.

Also the `QpackEncoder` is calling the instruction handler from within the lock, not sure if this will be an issue.

sbordet:
> Also the QpackEncoder is calling the instruction handler from within the lock, not sure if this will be an issue.

The instruction handler only writes to the encoder stream (does not call any user code), so we should be good.

sbordet (approved):

PR #9729, opened 2023-05-04
Simplified QuotedStringTokenizer

gregw:
Now implements a simple subset of `quoted-string` from RFC9110


gregw:
This could be a little bit controversial as it is used in a lot of places and I've just removed the ability to quote anything special. i.e. there is now no support of "\uab12" style unicode escapes, nor "\t" tabs etc.  Also there is no leniency for browsers that do not quote "\".

There was no spec for any of those.  We now just implement RFC9110 quoted-string.
It was also silly that we were lenient on a browser that sent "\Some\File\Name", but if they sent "\some\file\name" we would substitute the "\n".

thoughts?

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/MultiPartFormDataTest.java
```diff
@@ -677,7 +674,7 @@ public void testPartWithBackSlashInFileName() throws Exception
 
         String contents = """
             --AaB03x\r
-            content-disposition: form-data; name="stuff"; filename="Taken on Aug 22 \\ 2012.jpg"\r
+            content-disposition: form-data; name="stuff"; filename="Taken on Aug 22 \\\\ 2012.jpg"\r
```
sbordet: This feels weird. Why there are now 2 backslashes?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QuotedStringTokenizer.java
```diff
@@ -98,7 +93,8 @@ public boolean hasMoreTokens()
 
             switch (state)
             {
-                case 0: // Start
+                case 0 -> // Start
```
sbordet: Can we use an enum instead of `int` for the state?

sbordet (changes requested):

sbordet:
@gregw FTR maybe create another class with stricter behavior and keep the old one around.
And then have a compliance mode in `HttpConfiguration`?

gregw:
After feedback from Simone, I'm going to:
 + get rid of static API
 + get rid of passing the string in constructor 
 + Make the construction API about how to configure only
 + Make the API work with iterators rather than enumerations (welcome to 2000!)
 + Extract an interface
 + Write new implementation, but keep old as an option.

@joakime Can you gather some stats from users about filenames passed to multi-part.  Specifically do any still have non escaped "\" in the filename?   Or do any actually use the '\t' or '\n' style escapes.

Putting this back to draft for now.

gregw:
@joakime @sbordet this is still in draft mode, but can I get a pre-review.  It is a huge PR now and I'd like to get a sense of your feedback before doing too much more work.

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/MultiPartFormDataTest.java
gregw: Because we used to forgive browsers for not properly escaping "\" in filenames.  Now we are going to insist they quote them correctly.    

Could be a bad idea.   @joakime can you do some tests on windows to see how "\" is treated by current browsers?

gregw:
@joakime the tests that you added for multipart file uploads answers the question "do browser correctly quote '\'" as we have tests for msie and edge that expect to handle
```http
Content-Disposition: form-data; name="file"; filename="C:\Users\joakim\Pictures\jetty-avatar-256.png"
```

So I think we do need a mode that doesn't treat '\' as an escape.

Do you you know how such browsers escape characters if they don't treat '\' specially?   Specifically, what do they do if there is a `"` in the filename?

gregw:
I think we may need to implement [rfc8187](https://datatracker.ietf.org/doc/html/rfc8187 ) which replaces [rfc5987](https://datatracker.ietf.org/doc/html/rfc5987) which is suggest as per
https://stackoverflow.com/questions/93551/how-to-encode-the-filename-parameter-of-content-disposition-header-in-http

We will still need a mode that correctly handles the none escaped '\' in filenames, but we should look for filename* for the filename we actually use.

gregw:
@joakime can you gather the raw browser-capture data (like you did for jetty-core/jetty-http/src/test/resources/multipart) for multi part requests that have filenames containing non ASCII characters (eg `€`) and perhaps also with a `"` character in the name (if the FS supports it).

I'm very interested to see if RFC8187 is widely supported.

gregw:
Just doing a test with Chrome Version 113.0.5672.63 (Official Build) (64-bit)

A file called `t€st.txt` is uploaded with:
```
Content-Disposition: form-data; name="userfile1"; filename="tst.txt"
```

I can't tell what character that is instead of the euro.  It pastes into characters app as \u0000 and into intellij as \u0080, which is the unicode control character???
[edit: interestingly github renders that character as the standard replacement character! click edit to see the real character]

A file called `t"est.txt` is uploaded with:
```
Content-Disposition: form-data; name="userfile1"; filename="t%22st.txt"
```

So no RFC 8187 here... but also no usage of slosh escaping rather % escaping.  

Except the very same browser uploads `t%st.txt` as just 
```
Content-Disposition: form-data; name="userfile1"; filename="t%st.txt"
```

So they are not consistent!  No idea what we should do?
I guess if we implemented RFC8187, then at least if we ever saw a `filename*=` parameter, then we'd know for sure how it was encoded.

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/MultiPartFormDataTest.java
gregw: Reverted.   But there is no good solution.  See comments below.

joakime:
> @joakime can you gather the raw browser-capture data (like you did for jetty-core/jetty-http/src/test/resources/multipart) for multi part requests that have filenames containing non ASCII characters (eg `€`) and perhaps also with a `"` character in the name (if the FS supports it).
> 
> I'm very interested to see if RFC8187 is widely supported.

I can, but it'll have to wait till I get back.
I'll also capture the 3 examples in your follow-up comment from multiple os/browser combos.

gregw:
@sbordet I've pushed some javadoc

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QuotedStringTokenizer.java
```diff
@@ -263,362 +42,184 @@ public int countTokens()
...
-
-        StringBuilder b = new StringBuilder(s.length() - 2);
-        boolean escape = false;
-        for (int i = 1; i < s.length() - 1; i++)
+        /**
+         * If called, the built {@link QuotedStringTokenizer} will ignore optional white space characters before
+         * and after delimiters. This is not supported together with {@link #legacy()}.
+         * @return this {@code Builder}
+         */
+        public Builder optionalWhiteSpace()
```
sbordet: I think the name is confusing, I'd prefer `allowOptionalWhiteSpace()` in the style you have for other builder methods <verb><what>, like <return><Delimiters>.

I prefer "allow" rather than "ignore" because allow is what the user wants, "ignore" is what the implementation does.

Also, I don't see anything related to trimming?
If I receive `a = b` and if `allowOptionalWhiteSpace=true`, the name is `a ` or just `a`?
If `allowOptionalWhiteSpace=false` I expect an error?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QuotedStringTokenizer.java
```diff
@@ -263,362 +42,184 @@ public int countTokens()
...
-    {
-        return unquote(s, false);
-    }
+        /**
+         * If called, the built {@link QuotedStringTokenizer} will allow quotes to be within a token, not just as
+         * initial/final characters as per the RFC.  For example the RFC illegal string {@code 'one, two", "three' }
+         * with comma delimiter, would result in two tokens: {@code 'one' & 'two, three'}.
+         * @return this {@code Builder}
+         */
+        public Builder embeddedQuotes()
```
sbordet: Rename as `allowEmbeddedQuotes()`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QuotedStringTokenizer.java
```diff
@@ -263,362 +42,184 @@ public int countTokens()
...
-        if (s == null)
-            return null;
-        if (s.length() < 2)
-            return s;
+        /**
+         * If called, the built {@link QuotedStringTokenizer} will allow single quotes. This can only be used with
+         * {@link #legacy()}.
+         * @return this {@code Builder}
+         */
+        public Builder singleQuotes()
```
sbordet: Rename to `allowSingleQuotes()`, but where? Should it be `allowEmbeddedSingleQuote()`?
Also "single" is confusing... is it referring to single quote `'` or one occurrence of `"`?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QuotedStringTokenizer.java
```diff
@@ -263,362 +42,184 @@ public int countTokens()
...
-        char last = s.charAt(s.length() - 1);
-        if (first != last || (first != '"' && first != '\''))
-            return s;
+        /**
+         * If called, the built {@link QuotedStringTokenizer} will only allow escapes for quote characters.
+         * For example the string {@code '"test\"tokenizer\escape"'} will be unquoted as
+         * {@code 'test"tokenizer\escape'}.
+         * @return this {@code Builder}
+         */
+        public Builder escapeOnlyQuote()
```
sbordet: Not clear what it does? Is this `unescapeEmbeddedEscapedQuotes()`?
Also the javadoc having both single quote `'` and double quote `"` in the example makes it ambiguous. I'd drop the `'`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QuotedStringTokenizer.java
```diff
@@ -263,362 +42,184 @@ public int countTokens()
...
+                    throw new IllegalArgumentException("EscapeOnlyQuote not supported by legacy");
+                if (!_embeddedQuotes)
+                    throw new IllegalArgumentException("EmbeddedQuotes must be used with legacy");
+                return new QuotedStringTokenizerLegacy(_delim, _returnDelimiters, _returnQuotes, _singleQuotes);
             }
+            if (StringUtil.isEmpty(_delim))
+                throw new IllegalArgumentException("Delimiters must be provided");
+            if (_singleQuotes)
+                throw new IllegalArgumentException("Single quotes not supported by RFC9110");
+            return new QuotedStringTokenizerRfc9110(_delim, _optionalWhiteSpace, _returnDelimiters, _returnQuotes, _embeddedQuotes, _escapeOnlyQuote);
```
sbordet: I would pass in the `Builder` instance rather than exploding it, so that if in the future we add more configuration, we don't need to change the constructor. The getters in the builder can be package private.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QuotedStringTokenizerRfc9110.java
```diff
@@ -0,0 +1,379 @@
...
+ * <p>
+ * This class extends {@link StringTokenizer} with partial handling of
+ * <a href="https://www.rfc-editor.org/rfc/rfc9110#name-quoted-strings">RFC9110 quoted-string</a>s.
+ * The deviation from the RFC is that characters are not enforced to be
+ * {@code qdtext = HTAB / SP / %x21 / %x23-5B / %x5D-7E / obs-text} and it is expected
+ * that the caller will enforce any character restrictions.
+ *
+ * @see java.util.StringTokenizer
+ */
+public class QuotedStringTokenizerRfc9110 implements QuotedStringTokenizer
```
sbordet: Rename to `RFC9110QuotedStringTokenizer`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QuotedStringTokenizerRfc9110.java
```diff
@@ -0,0 +1,379 @@
...
+/**
+ * StringTokenizer with Quoting support.
+ * <p>
+ * This class extends {@link StringTokenizer} with partial handling of
+ * <a href="https://www.rfc-editor.org/rfc/rfc9110#name-quoted-strings">RFC9110 quoted-string</a>s.
+ * The deviation from the RFC is that characters are not enforced to be
+ * {@code qdtext = HTAB / SP / %x21 / %x23-5B / %x5D-7E / obs-text} and it is expected
+ * that the caller will enforce any character restrictions.
+ *
+ * @see java.util.StringTokenizer
```
sbordet: Remove this reference and update the javadocs too.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QuotedStringTokenizerRfc9110.java
```diff
@@ -0,0 +1,379 @@
...
+        _embeddedQuotes = embeddedQuotes;
+        _escapeOnlyQuote = escapeOnlyQuote;
+
+        if (_delim.indexOf('"') >= 0)
+            throw new IllegalArgumentException("Can't use quote as delimiters: " + _delim);
+        if (_optionalWhiteSpace && _delim.indexOf(' ') >= 0)
+            throw new IllegalArgumentException("Can't delimit with space with optional white space");
+    }
+
+    protected boolean isOWS(char c)
```
sbordet: Let's be clear and expand to `isOptionalWhiteSpace()`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QuotedStringTokenizerRfc9110.java
```diff
@@ -0,0 +1,379 @@
...
+        return null;
+    }
+
+    /**
+     * Unquote a string.
+     *
+     * @param s The string to unquote.
+     * @return unquoted string
+     */
+    public String unquote(String s)
```
sbordet: Missing `@Override`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QuotedStringTokenizer.java
```diff
@@ -263,362 +42,184 @@ public int countTokens()
      * empty string.
      *
      * @param s The string to quote.
-     * @param delim the delimiter to use to quote the string
      * @return quoted string
      */
-    public static String quoteIfNeeded(String s, String delim)
+    default String quote(String s)
```
sbordet, line 54: I think "quote" is a bit misleading, as really what this does is "escape and quote", and how it escapes is subject to configuration.
Not sure changing the method name to `escapeAndQuote()` is better, but at least mention in javadocs that this method escapes (perhaps with some example?).

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/MultiPartFormDataTest.java
```diff
@@ -677,7 +675,7 @@ public void testPartWithBackSlashInFileName() throws Exception
 
         String contents = """
             --AaB03x\r
-            content-disposition: form-data; name="stuff"; filename="Taken on Aug 22 \\ 2012.jpg"\r
+            content-disposition: form-data; name="stuff"; filename="C:\\Pictures\\2012.jpg"\r
```
sbordet: We can't have a filename with spaces? 
If so we should keep the test and verify we cannot.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpField.java
```diff
@@ -26,6 +26,8 @@
  */
 public class HttpField
 {
+    public static final QuotedStringTokenizer PARAMETER_TOKENIZER = QuotedStringTokenizer.builder().delimiters(";").optionalWhiteSpace().embeddedQuotes().returnQuotes().build();
+    public static final QuotedStringTokenizer NAME_VALUE_TOKENIZER = QuotedStringTokenizer.builder().delimiters("=").optionalWhiteSpace().build();
```
sbordet: Javadocs for these 2, explaining "use this tokenizer when ...".
It's hard to tell when a string is a _value_ or it is a _parameter_.

jetty-ee9/jetty-ee9-security/src/main/java/org/eclipse/jetty/ee9/security/authentication/DigestAuthenticator.java
```diff
@@ -52,6 +53,7 @@
 public class DigestAuthenticator extends LoginAuthenticator
 {
     private static final Logger LOG = LoggerFactory.getLogger(DigestAuthenticator.class);
+    private static final QuotedStringTokenizer __tokenizer = QuotedStringTokenizer.builder().delimiters("=, ").returnDelimiters().embeddedQuotes().build();
```
sbordet: Let's use `TOKENIZER` rather than the old `__tokenizer` syntax.

sbordet (changes requested):

jetty-core/jetty-http/src/test/java/org/eclipse/jetty/http/MultiPartFormDataTest.java
gregw: The test is not about spaces. It is about how '\' is treated. 
I just made the name more realistic... but I'll put spaces back in.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QuotedStringTokenizer.java
gregw: OK with the rename.

You don't need to trim when there is OWS around delimiters.
So `a = b` with delimiter '=' and OWS has 2 tokens `a` and `b`. (Note that is not legal in a QuotedCSV as OWS is only around ',' and ';' but not '=' !!!)
Without OWS there are two token `a ` and ` b`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QuotedStringTokenizer.java
gregw: Allowing single quotes doesn't change where quotes are allowed.
If you want embedded single quotes then you need to do `builder.allowEmbeddedQuotes().allowSingleQuotes()`.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QuotedStringTokenizer.java
gregw: I'd prefer not.   Having it this way makes it very clear what configurations are supported by which browser type.   If we passed the Builder and passed it into the constructors, then the legacy implementation might not be updated to see a new config and would thus not throw if it was set.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QuotedStringTokenizer.java
gregw: I've improved the javadoc

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QuotedStringTokenizer.java
gregw: Nothing to do with embedded.
I've improved the name and javadoc.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/QuotedStringTokenizer.java
gregw: Escaping is part of quoting.  The RFC doesn't define it as `escaped-and-quoted-string`, just as `quoted-string`.

Would we also have to call it `unquoteAndUnEscape`?    So nah!     I'll add something about escapes in the javadoc.

sbordet (approved):

PR #9733, opened 2023-05-04
FCGI improvements.

sbordet:
* Better handling for HTTPS parameter on server side.
* Better handling of unknown frame types.

### DIFF
--- a/jetty-core/jetty-fcgi/jetty-fcgi-client/src/main/java/org/eclipse/jetty/fcgi/parser/Parser.java
+++ b/jetty-core/jetty-fcgi/jetty-fcgi-client/src/main/java/org/eclipse/jetty/fcgi/parser/Parser.java
@@ -59,8 +59,14 @@ public abstract class Parser
     private static final Logger LOG = LoggerFactory.getLogger(Parser.class);
 
     protected final HeaderParser headerParser = new HeaderParser();
+    private final Listener listener;
     private State state = State.HEADER;
     private int padding;
+
+    protected Parser(Listener listener)
+    {
+        this.listener = listener;
+    }
 
     /**
      * @param buffer the bytes to parse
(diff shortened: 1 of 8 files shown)


jetty-core/jetty-fcgi/jetty-fcgi-client/src/main/java/org/eclipse/jetty/fcgi/parser/Parser.java
```diff
@@ -59,83 +59,97 @@
...
-                {
-                    throw new IllegalStateException();
+                    default -> throw new IllegalStateException();
                 }
             }
         }
+        catch (Throwable x)
+        {
+            notifyFailure(headerParser.getRequest(), x);
+            buffer.position(buffer.limit());
```
gregw: Swap these two lines. i.e consume the buffer before notifying the user

jetty-core/jetty-fcgi/jetty-fcgi-client/src/main/java/org/eclipse/jetty/fcgi/parser/Parser.java
```diff
@@ -59,83 +59,97 @@
...
+        }
     }
 
     protected abstract ContentParser findContentParser(FCGI.FrameType frameType);
 
+    private void notifyFailure(int request, Throwable failure)
+    {
+        listener.onFailure(request, failure);
+    }
+
```
gregw: Is this needed?  inline it.

jetty-core/jetty-fcgi/jetty-fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/internal/HttpStreamOverFCGI.java
```diff
@@ -101,15 +102,17 @@ else if (FCGI.Headers.QUERY_STRING.equalsIgnoreCase(name))
...
         else
             processField(field);
     }
 
     public void onHeaders()
     {
         String pathQuery = URIUtil.addPathQuery(_path, _query);
-        // TODO https?
-        MetaData.Request request = new MetaData.Request(_method, HttpScheme.HTTP.asString(), hostPort, pathQuery, HttpVersion.fromString(_version), _headers, -1);
+        HttpScheme scheme = _secure != null ? HttpScheme.HTTPS : HttpScheme.HTTP;
```
gregw: What is the value of `_secure` for?   Could it be "false" i.e. saying that it is not secure?

gregw (changes requested):
Almost approved... but an explanation of the value of the `_secure` field is needed (probably as a comment).   Other things are minor niggles.

jetty-core/jetty-fcgi/jetty-fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/internal/HttpStreamOverFCGI.java
sbordet: RFC 3875 (https://datatracker.ietf.org/doc/html/rfc3875) does not define `HTTPS`, but apparently for PHP it must be set to a non-empty value to indicate that the request was secure (https://www.php.net/manual/en/reserved.variables.server.php).

We use `HTTPS=on` for compatibility with PHP applications such as WordPress, but other applications require `true` or `1`, see for example: https://build.trac.wordpress.org/browser/branches/6.2/wp-includes/load.php#L1473

So `HTTPS=false` as per PHP spec would still mean that the request is secure because it's non-empty, but it really depends on the application.

🤷🏼

gregw: Maybe we should have a list of affirmative values: "true", "1", "on", "secure", "https". Anything else is not secure.

At the very least can we exclude the empty string?

sbordet: -1 for the list, +1 for non-empty.

gregw (approved):

PR #9735, opened 2023-05-04
Issue #9734 fix check for state start in cookie config

### LINKED ISSUE #9734: Cookie config can be set after SessionHandler is started
jetty-12.0.x

It is allowable to call the `CookieConfig.setXX` methods after the `SessionHandler` has started, before  the context has fully started. Ie it is allowable for `ServletContextListeners` to call these methods in `contextInitialized()`.

This change is wrong:

https://github.com/eclipse/jetty.project/blob/jetty-12.0.x/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/SessionHandler.java#L260

Needs to be reverted to:

```java
            if (_servletContextHandlerContext != null && _servletContextHandlerContext.getServletContextHandler().isAvailable())
                throw new IllegalStateException("CookieConfig cannot be set after ServletContext is started");  
```

janbartel:
Closes #9734

### DIFF
--- a/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/SessionHandler.java
+++ b/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/SessionHandler.java
@@ -257,8 +257,10 @@ public class SessionHandler extends AbstractSessionManager implements Handler.Si
 
         private void checkState()
         {
-            if (isStarted())
+            ServletContextHandler handler = ServletContextHandler.getCurrentServletContextHandler();
+            if (handler != null && handler.isAvailable())
                 throw new IllegalStateException("CookieConfig cannot be set after ServletContext is started");
+
         }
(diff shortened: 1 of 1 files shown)


jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/SessionHandler.java
```diff
@@ -257,8 +257,10 @@ public void setSecure(boolean secure)
 
         private void checkState()
         {
-            if (isStarted())
+            ServletContextHandler handler = ServletContextHandler.getCurrentServletContextHandler();
+            if (handler != null && handler.isAvailable())
```
gregw, line 264: You need a comment here to say why `isStarted()` is not sufficient when the error message is "cannot be set after ... is started".

gregw (approved):
OK, but you need to comment why.

PR #9736, opened 2023-05-04
Optimize Content-Length: 0 handling

gregw:
Create and use a pre-encoded HttpFields.CONTENT_LENGTH_0 constant Used the constant in more places
renamed the `putLongField` methods of HttpFields.Mutable to just `put`


gregw:
Screwed by force push to head!

PR #9740, opened 2023-05-05
Jetty 12 content length 0 take3

gregw:
Create and use a pre-encoded HttpFields.CONTENT_LENGTH_0 constant
Used the constant in more places
renamed the `putLongField` methods of HttpFields.Mutable to just `put`

(third attempt after forced push zero event incident!)


joakime:
As I was looking around at our CONTENT_LENGTH usage (in this branch), I was struck by an apparent set of bugs in the jetty-documentation's server/HandlerDocs.

https://github.com/eclipse/jetty.project/blob/897611d7361ac8f13a538d8c86016477d8af5e3e/documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/server/HandlerDocs.java#L45

There's 8 occurrences of this line in the above file.
@sbordet is this intended? or a bug?

joakime:
This one seems like a bug.

https://github.com/eclipse/jetty.project/blob/897611d7361ac8f13a538d8c86016477d8af5e3e/jetty-core/jetty-fcgi/jetty-fcgi-client/src/main/java/org/eclipse/jetty/fcgi/client/transport/internal/HttpSenderOverFCGI.java#L76

Why would we ever want to write a blank CONTENT_LENGTH header?

gregw:
@joakime I'll fix those in this PR as well....

gregw:
Note also that it is kind of a bug to add headers like CONTENT_LENGTH and CONTENT_TYPE as there should only be 1 of those.  Thus put should be used.   Should we produce a warning if somebody adds one of these?

gregw:
@joakime I fixed the CONTENT_LENGTH->CONTENT_TYPE ones, as they are kind of related to this.  But I've left the FCGI one alone, as I'm not sure if that is a bug or not (looks deliberate).

joakime (approved):

PR #9741, opened 2023-05-05
Review of websocket parser, improve testing & comments.

lachlan-roberts:


jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/Parser.java
```diff
@@ -55,6 +55,7 @@
     private int cursor;
     private byte[] mask;
     private int payloadLength;
+    private long longLengthAccumulator;
```
sbordet: Do you really need this extra field?
I think that `payloadLength` can be kept as `int`, and the parser adjusted to read 8 bytes but to throw if they exceed `Integer.MAX_VALUE`, so that we know that the max length we support is at most 2 GiB.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/Parser.java
```diff
@@ -324,28 +338,28 @@ private ParsedFrame parsePayload(ByteBuffer buffer)
...
-                // not enough to complete this frame
-                // Can we auto-fragment
+                // Not enough data to complete this frame, can we auto-fragment?
                 if (configuration.isAutoFragment() && isDataFrame)
                     return autoFragment(buffer, available);
 
-                // No space in the buffer, so we have to copy the partial payload
+                // No space in the buffer, so we have to copy the partial payload.
+                // The size of this allocation is limited by the maxFrameSize.
                 aggregate = bufferPool.acquire(payloadLength, false);
```
sbordet, line 346: Let's use `ByteBufferAccumulator` here rather than an `aggregate` buffer.

sbordet (changes requested):

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/Parser.java
lachlan-roberts: We are not accumulating ByteBuffers here so this would be pointless. We just want a single buffer to back the websocket Frame.

sbordet (approved):

PR #9749, opened 2023-05-08
HTTP/2 improvements for CVE-2023-36478

sbordet:
Addressing CVE-2023-36478

* Implemented a few required error handlings.
* Changed `Parser.init()` to directly take the listener, rather than wrapping it. The reason for this change was to be able to reconfigure the Parser upon receiving a SETTINGS frame.

### DIFF
--- a/jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
+++ b/jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
@@ -61,7 +61,7 @@ public class HTTP2ClientConnectionFactory implements ClientConnectionFactory
         Generator generator = new Generator(byteBufferPool, client.isUseOutputDirectByteBuffers(), client.getMaxHeaderBlockFragment());
         FlowControlStrategy flowControl = client.getFlowControlStrategyFactory().newFlowControlStrategy();
 
-        Parser parser = new Parser(byteBufferPool, client.getMaxDecoderTableCapacity(), client.getMaxResponseHeadersSize());
+        Parser parser = new Parser(byteBufferPool, client.getMaxResponseHeadersSize());
         parser.setMaxFrameSize(client.getMaxFrameSize());
         parser.setMaxSettingsKeys(client.getMaxSettingsKeys());
 
--- a/jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
+++ b/jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
@@ -58,15 +58,16 @@ public class HTTP2ClientConnectionFactory implements ClientConnectionFactory
         @SuppressWarnings("unchecked")
         Promise<Session> sessionPromise = (Promise<Session>)context.get(SESSION_PROMISE_CONTEXT_KEY);
 
-        Generator generator = new Generator(byteBufferPool, client.getMaxEncoderTableSize(), client.getMaxHeaderBlockFragment());
+        Generator generator = new Generator(byteBufferPool, client.isUseOutputDirectByteBuffers(), client.getMaxHeaderBlockFragment());
         FlowControlStrategy flowControl = client.getFlowControlStrategyFactory().newFlowControlStrategy();
 
-        Parser parser = new Parser(byteBufferPool, client.getMaxDecoderTableSize(), client.getMaxResponseHeadersSize());
+        Parser parser = new Parser(byteBufferPool, client.getMaxDecoderTableCapacity(), client.getMaxResponseHeadersSize());
         parser.setMaxFrameSize(client.getMaxFrameSize());
         parser.setMaxSettingsKeys(client.getMaxSettingsKeys());
 
         HTTP2ClientSession session = new HTTP2ClientSession(scheduler, endPoint, parser, generator, listener, flowControl);
         session.setMaxRemoteStreams(client.getMaxConcurrentPushedStreams());
+        session.setMaxEncoderTableCapacity(client.getMaxEncoderTableCapacity());
         long streamIdleTimeout = client.getStreamIdleTimeout();
         if (streamIdleTimeout > 0)
             session.setStreamIdleTimeout(streamIdleTimeout);
@@ -111,8 +112,8 @@ public class HTTP2ClientConnectionFactory implements ClientConnectionFactory
             {
                 if (v == null)
                 {
-                    v = client.getMaxDecoderTableSize();
-                    if (v == HpackContext.DEFAULT_MAX_TABLE_SIZE)
+                    v = client.getMaxDecoderTableCapacity();
+                    if (v == HpackContext.DEFAULT_MAX_TABLE_CAPACITY)
                         v = null;
                 }
                 return v;
--- a/jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
+++ b/jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
@@ -27,6 +27,7 @@ import org.eclipse.jetty.http2.frames.PrefaceFrame;
 import org.eclipse.jetty.http2.frames.SettingsFrame;
 import org.eclipse.jetty.http2.frames.WindowUpdateFrame;
 import org.eclipse.jetty.http2.generator.Generator;
+import org.eclipse.jetty.http2.hpack.HpackContext;
 import org.eclipse.jetty.http2.parser.Parser;
 import org.eclipse.jetty.io.ByteBufferPool;
 import org.eclipse.jetty.io.ClientConnectionFactory;
@@ -111,7 +112,7 @@ public class HTTP2ClientConnectionFactory implements ClientConnectionFactory
                 if (v == null)
                 {
                     v = client.getMaxDecoderTableSize();
-                    if (v == 4096)
+                    if (v == HpackContext.DEFAULT_MAX_TABLE_SIZE)
                         v = null;
                 }
                 return v;
--- a/jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
+++ b/jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
@@ -57,11 +57,11 @@ public class HTTP2ClientConnectionFactory implements ClientConnectionFactory
         @SuppressWarnings("unchecked")
         Promise<Session> sessionPromise = (Promise<Session>)context.get(SESSION_PROMISE_CONTEXT_KEY);
 
-        Generator generator = new Generator(byteBufferPool, client.getMaxDynamicTableSize(), client.getMaxHeaderBlockFragment());
+        Generator generator = new Generator(byteBufferPool, client.getMaxEncoderTableSize(), client.getMaxHeaderBlockFragment());
         FlowControlStrategy flowControl = client.getFlowControlStrategyFactory().newFlowControlStrategy();
 
-        Parser parser = new Parser(byteBufferPool, 4096, 8192);
-        parser.setMaxFrameLength(client.getMaxFrameLength());
+        Parser parser = new Parser(byteBufferPool, client.getMaxDecoderTableSize(), client.getMaxResponseHeadersSize());
+        parser.setMaxFrameSize(client.getMaxFrameSize());
         parser.setMaxSettingsKeys(client.getMaxSettingsKeys());
 
         HTTP2ClientSession session = new HTTP2ClientSession(scheduler, endPoint, parser, generator, listener, flowControl);
@@ -102,7 +102,7 @@ public class HTTP2ClientConnectionFactory implements ClientConnectionFactory
             if (settings == null)
                 settings = new HashMap<>();
 
-            // Here we want to populate any settings to send to the server
+            // Below we want to populate any settings to send to the server
             // that have a different default than what prescribed by the RFC.
             // Changing the configuration is done when the SETTINGS is sent.
 
@@ -110,7 +110,7 @@ public class HTTP2ClientConnectionFactory implements ClientConnectionFactory
             {
                 if (v == null)
                 {
-                    v = client.getMaxDynamicTableSize();
+                    v = client.getMaxDecoderTableSize();
                     if (v == 4096)
                         v = null;
                 }
@@ -131,8 +131,18 @@ public class HTTP2ClientConnectionFactory implements ClientConnectionFactory
             {
                 if (v == null)
                 {
-                    v = client.getMaxFrameLength();
+                    v = client.getMaxFrameSize();
                     if (v == Frame.DEFAULT_MAX_LENGTH)
+                        v = null;
+                }
+                return v;
+            });
+            settings.compute(SettingsFrame.MAX_HEADER_LIST_SIZE, (k, v) ->
+            {
+                if (v == null)
+                {
+                    v = client.getMaxResponseHeadersSize();
+                    if (v <= 0)
                         v = null;
                 }
                 return v;
--- a/jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
+++ b/jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
@@ -22,6 +22,7 @@ import org.eclipse.jetty.http2.FlowControlStrategy;
 import org.eclipse.jetty.http2.HTTP2Connection;
 import org.eclipse.jetty.http2.ISession;
 import org.eclipse.jetty.http2.api.Session;
+import org.eclipse.jetty.http2.frames.Frame;
 import org.eclipse.jetty.http2.frames.PrefaceFrame;
 import org.eclipse.jetty.http2.frames.SettingsFrame;
 import org.eclipse.jetty.http2.frames.WindowUpdateFrame;
@@ -58,23 +59,25 @@ public class HTTP2ClientConnectionFactory implements ClientConnectionFactory
 
         Generator generator = new Generator(byteBufferPool, client.getMaxDynamicTableSize(), client.getMaxHeaderBlockFragment());
         FlowControlStrategy flowControl = client.getFlowControlStrategyFactory().newFlowControlStrategy();
-        HTTP2ClientSession session = new HTTP2ClientSession(scheduler, endPoint, generator, listener, flowControl);
+
+        Parser parser = new Parser(byteBufferPool, 4096, 8192);
+        parser.setMaxFrameLength(client.getMaxFrameLength());
+        parser.setMaxSettingsKeys(client.getMaxSettingsKeys());
+
+        HTTP2ClientSession session = new HTTP2ClientSession(scheduler, endPoint, parser, generator, listener, flowControl);
         session.setMaxRemoteStreams(client.getMaxConcurrentPushedStreams());
         long streamIdleTimeout = client.getStreamIdleTimeout();
         if (streamIdleTimeout > 0)
             session.setStreamIdleTimeout(streamIdleTimeout);
 
-        Parser parser = new Parser(byteBufferPool, session, 4096, 8192);
-        parser.setMaxFrameLength(client.getMaxFrameLength());
-        parser.setMaxSettingsKeys(client.getMaxSettingsKeys());
-
         RetainableByteBufferPool retainableByteBufferPool = byteBufferPool.asRetainableByteBufferPool();
-
         HTTP2ClientConnection connection = new HTTP2ClientConnection(client, retainableByteBufferPool, executor, endPoint,
-            parser, session, client.getInputBufferSize(), promise, listener);
+            session, client.getInputBufferSize(), promise, listener);
         connection.setUseInputDirectByteBuffers(client.isUseInputDirectByteBuffers());
         connection.setUseOutputDirectByteBuffers(client.isUseOutputDirectByteBuffers());
         connection.addEventListener(connectionListener);
+        parser.init(connection);
+
         return customize(connection, context);
     }
 
@@ -84,9 +87,9 @@ public class HTTP2ClientConnectionFactory implements ClientConnectionFactory
         private final Promise<Session> promise;
         private final Session.Listener listener;
 
-        private HTTP2ClientConnection(HTTP2Client client, RetainableByteBufferPool retainableByteBufferPool, Executor executor, EndPoint endpoint, Parser parser, ISession session, int bufferSize, Promise<Session> promise, Session.Listener listener)
+        private HTTP2ClientConnection(HTTP2Client client, RetainableByteBufferPool retainableByteBufferPool, Executor executor, EndPoint endpoint, HTTP2ClientSession session, int bufferSize, Promise<Session> promise, Session.Listener listener)
         {
-            super(retainableByteBufferPool, executor, endpoint, parser, session, bufferSize);
+            super(retainableByteBufferPool, executor, endpoint, session, bufferSize);
             this.client = client;
             this.promise = promise;
             this.listener = listener;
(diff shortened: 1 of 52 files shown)


jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
```diff
@@ -58,23 +59,25 @@ public Connection newConnection(EndPoint endPoint, Map<String, Object> context)
 
         Generator generator = new Generator(byteBufferPool, client.getMaxDynamicTableSize(), client.getMaxHeaderBlockFragment());
         FlowControlStrategy flowControl = client.getFlowControlStrategyFactory().newFlowControlStrategy();
-        HTTP2ClientSession session = new HTTP2ClientSession(scheduler, endPoint, generator, listener, flowControl);
+
+        Parser parser = new Parser(byteBufferPool, 4096, 8192);
+        parser.setMaxFrameLength(client.getMaxFrameLength());
+        parser.setMaxSettingsKeys(client.getMaxSettingsKeys());
+
+        HTTP2ClientSession session = new HTTP2ClientSession(scheduler, endPoint, parser, generator, listener, flowControl);
```
gregw, line 68: It is kind of strange passing in the parser and generator, rather than something like a `Http2Configuration` or directly configuring the session itself after it is constructed.

This is a -0 because we were already doing it for the generator.

jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
```diff
@@ -58,23 +59,25 @@ public Connection newConnection(EndPoint endPoint, Map<String, Object> context)
...
 
-        Parser parser = new Parser(byteBufferPool, session, 4096, 8192);
-        parser.setMaxFrameLength(client.getMaxFrameLength());
-        parser.setMaxSettingsKeys(client.getMaxSettingsKeys());
-
         RetainableByteBufferPool retainableByteBufferPool = byteBufferPool.asRetainableByteBufferPool();
-
         HTTP2ClientConnection connection = new HTTP2ClientConnection(client, retainableByteBufferPool, executor, endPoint,
-            parser, session, client.getInputBufferSize(), promise, listener);
+            session, client.getInputBufferSize(), promise, listener);
```
gregw: Can you rename variable to
```suggestion
            session, client.getInputBufferSize(), sessionPromise, listener);
```

jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
```diff
@@ -84,9 +87,9 @@ private static class HTTP2ClientConnection extends HTTP2Connection implements Ca
         private final Promise<Session> promise;
         private final Session.Listener listener;
 
-        private HTTP2ClientConnection(HTTP2Client client, RetainableByteBufferPool retainableByteBufferPool, Executor executor, EndPoint endpoint, Parser parser, ISession session, int bufferSize, Promise<Session> promise, Session.Listener listener)
+        private HTTP2ClientConnection(HTTP2Client client, RetainableByteBufferPool retainableByteBufferPool, Executor executor, EndPoint endpoint, HTTP2ClientSession session, int bufferSize, Promise<Session> promise, Session.Listener listener)
```
gregw: ditto
```suggestion
        private HTTP2ClientConnection(HTTP2Client client, RetainableByteBufferPool retainableByteBufferPool, Executor executor, EndPoint endpoint, HTTP2ClientSession session, int bufferSize, Promise<Session> sessionPromise, Session.Listener listener)
```

jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
```diff
@@ -98,12 +101,42 @@ public void onOpen()
...
-            settings.computeIfAbsent(SettingsFrame.INITIAL_WINDOW_SIZE, k -> client.getInitialStreamRecvWindow());
-            settings.computeIfAbsent(SettingsFrame.MAX_CONCURRENT_STREAMS, k -> client.getMaxConcurrentPushedStreams());
 
-            Integer maxFrameLength = settings.get(SettingsFrame.MAX_FRAME_SIZE);
-            if (maxFrameLength != null)
-                getParser().setMaxFrameLength(maxFrameLength);
+            // Here we want to populate any settings to send to the server
+            // that have a different default than what prescribed by the RFC.
+            // Changing the configuration is done when the SETTINGS is sent.
+
```
gregw, line 110: remove blank line!

jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
```diff
@@ -98,12 +101,42 @@ public void onOpen()
...
+            settings.compute(SettingsFrame.HEADER_TABLE_SIZE, (k, v) ->
+            {
+                if (v == null)
+                {
+                    v = client.getMaxDynamicTableSize();
```
gregw: again it feels like we should have a `Http2Configuration` interface.  Here you are pulling configuration directly from the client, were elsewhere you did a pull from the client and then pushed into parser/generator.

I'm now also worried about how the value in the client might differ from the value in the generator.

jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
```diff
@@ -98,12 +101,42 @@ public void onOpen()
...
+            {
+                if (v == null)
+                {
+                    v = client.getInitialStreamRecvWindow();
+                    if (v == FlowControlStrategy.DEFAULT_WINDOW_SIZE)
+                        v = null;
+                }
+                return v;
+            });
+            settings.compute(SettingsFrame.MAX_FRAME_SIZE, (k, v) ->
```
gregw, line 132: ditto, pulling configuration from client that was pushed into parser and then into session.

jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Connection.java
```diff
@@ -231,6 +231,76 @@ private Runnable pollTask()
...
+        session.onHeaders(frame);
+    }
+
+    @Override
+    public void onData(DataFrame frame)
+    {
+        NetworkBuffer networkBuffer = producer.networkBuffer;
+        networkBuffer.retain();
+        Callback callback = networkBuffer;
+        session.onData(frame, callback);
```
gregw: Why is the retain needed here?  I'm not seeing an extra reference made to the buffer? Or is this the original reference?  I.e. the parser makes the frame, but doesn't increment retain on network buffer?
If so, add a comment saying that it is the initial retain for the frame, not an additional one.

jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/parser/Parser.java
```diff
@@ -44,33 +43,34 @@
...
     {
         this.byteBufferPool = byteBufferPool;
-        this.listener = listener;
         this.headerParser = new HeaderParser(rateControl == null ? RateControl.NO_RATE_CONTROL : rateControl);
         this.hpackDecoder = new HpackDecoder(maxDynamicTableSize, maxHeaderSize);
         this.bodyParsers = new BodyParser[FrameType.values().length];
     }
 
-    public void init(UnaryOperator<Listener> wrapper)
+    public void init(Listener listener)
```
gregw, line 81: Can you explain why the listener can no longer be injected in the constructor?
It might be better to just pass the listener in with every call to parse?
I'm not really understanding the previous wrapper init either.

gregw (dismissed):
This is only a partial review, as I have lots of questions and am -0 on somethings.
Perhaps we need to hangout, but it would be good if the comments/javadoc was good enough so that explanation of these things was not needed.

jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
sbordet: No, because it is a generic comment that does not apply to the next line, it stands on its own for the rest of the code.

jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Connection.java
sbordet: It is needed because the `DataFrame` implicitly contains a slice of the `networkBuffer`.
This is the first available spot that we have to retain the `networkBuffer` before it is given to applications along with the callback to release it.
Remember that this code is older than the introduction of `Retainable` -- in Jetty 12 we do it differently and fully leverage `Retainable`.

jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/parser/Parser.java
sbordet: It is a chicken/egg problem.
We need the parser to build the session, we need the session to build the connection, and the connection is the parser listener that we would need to build the parser.
Therefore we need a once-only setter that it is `init(Listener)`.

Before, a parser listener (the session) was passed to the constructor of the parser, but then the connection had to wrap it via `init(Function)` -- it was indeed weird.

Now it's clearer because the contract is that you build the parser, and then you have to call `init(Listener)` passing the listener (the connection).

jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
sbordet: The parser needs to be built outside because it is configured with the session as the parser listener, and needs to be given to the session for reconfiguration in case a SETTINGS frame is received.
What are we doing for the generator?

jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
sbordet: The value in the client is the per-connection _initial_ value.
After the connection is established, the connection specific value may be changed by SETTINGS frames, that apply to that specific connection.
A new connection gets again the initial value from the client.
While rare, there may be cases where different connections may need different configurations.

I don't see much value in `Http2Configuration` interface, as we never need to use it polymorphically.

If it is a class/record, it will be partially shared between client and server, because both have specific configurations that don't apply to the other (e.g. max concurrent streams vs max pushed streams, enable connect only valid on server, etc.).

Also, the other components (such as parser and generator) will only use some of the values, and we would need to clone the configuration to keep track of connection-specific values vs connection initial values (or duplicate the relevant component configuration in the component itself).
I'm not opposed to it, and we have actually taken that path in http3 with `HTTP3Configuration` and `QuicConfiguration`, but for http2 feels like a lot of work for little gain and still the need to maintain all the setters for backward compatibility.

Perhaps for 12, but again breaking backwards compatibility (by removing all the setters)?

jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
gregw: > What are we doing for the generator?

passing it in after constructing it with configuration pulled from the client.

jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
gregw: The "populate" applies to the next block of code (4 calls to compute), which are not separated by blank lines.

Either delete the blank line, or don't start the comment with just "Here" (maybe "Beyond here"), as currently it looks isolated and like a left over old comment.

jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
gregw: OK in regards to no Http2Configuration class.

But I still find it confusing to see `client.getMaxDynamicTableSize()`, here, but also in the constructor for the generator above.

If this is the initial value, then where does the canonical current value live?  Surely that should only be initialised once from the `client` and then dynamically managed after that?

jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/parser/Parser.java
gregw: OK. I don't love it, but better than the wrapper.

jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/HTTP2Connection.java
gregw: Can you add a short comment to that effect.

jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
sbordet: Replaced "Here" with "Below".

jetty-http2/http2-client/src/test/java/org/eclipse/jetty/http2/client/SettingsTest.java
```diff
@@ -0,0 +1,357 @@
...
+
+            @Override
+            public void onClose(Session session, GoAwayFrame frame)
+            {
+                clientCloseLatch.countDown();
+            }
+        });
+
+        // Wait until the client has finished the previous writes.
+        Thread.sleep(1000);
```
lachlan-roberts, line 92: Is there a way you could avoid using the Thread.sleep()?

jetty-http2/http2-hpack/src/main/java/org/eclipse/jetty/http2/hpack/HpackContext.java
```diff
@@ -183,26 +183,26 @@
...
+            LOG.debug(String.format("HdrTbl[%x] created max=%d", hashCode(), maxTableSize));
     }
 
     public void resize(int newMaxDynamicTableSize)
     {
         if (LOG.isDebugEnabled())
-            LOG.debug(String.format("HdrTbl[%x] resized max=%d->%d", hashCode(), _maxDynamicTableSizeInBytes, newMaxDynamicTableSize));
-        _maxDynamicTableSizeInBytes = newMaxDynamicTableSize;
+            LOG.debug(String.format("HdrTbl[%x] resized max=%d->%d", hashCode(), _maxTableSize, newMaxDynamicTableSize));
+        _maxTableSize = newMaxDynamicTableSize;
```
lachlan-roberts, line 206: Isn't this more setting the `capacity` rather than the `maxDynamicTableSize`?
or at least `dynamicTableSize` (without max), the max is set by the SETTINGS frame.

For example the spec says: 

> Protocols that use HPACK determine the maximum size that the encoder
>    is permitted to use for the dynamic table.  In HTTP/2, this value is
>    determined by the SETTINGS_HEADER_TABLE_SIZE setting (see
>    Section 6.5.2 of [HTTP2]).
> 
>    An encoder can choose to use less capacity than this maximum size
>    (see Section 6.3), but the chosen size MUST stay lower than or equal
>    to the maximum set by the protocol.


I would also review the other places where max is used to describe the table size.

jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
```diff
@@ -98,12 +101,52 @@ public void onOpen()
...
+            settings.compute(SettingsFrame.MAX_HEADER_LIST_SIZE, (k, v) ->
+            {
+                if (v == null)
+                {
+                    v = client.getMaxResponseHeadersSize();
+                    if (v <= 0)
+                        v = null;
+                }
+                return v;
+            });
```
lachlan-roberts, line 151: I would have expected to see these values be set on the decoder after this like in the HTTP/3 equivalent.
Why is this not done here?

jetty-http2/http2-server/src/main/java/org/eclipse/jetty/http2/server/AbstractHTTP2ServerConnectionFactory.java
```diff
@@ -245,12 +289,16 @@ public HttpConfiguration getHttpConfiguration()
...
     {
         Map<Integer, Integer> settings = new HashMap<>();
-        settings.put(SettingsFrame.HEADER_TABLE_SIZE, getMaxDynamicTableSize());
-        settings.put(SettingsFrame.INITIAL_WINDOW_SIZE, getInitialStreamRecvWindow());
-        int maxConcurrentStreams = getMaxConcurrentStreams();
-        if (maxConcurrentStreams >= 0)
-            settings.put(SettingsFrame.MAX_CONCURRENT_STREAMS, maxConcurrentStreams);
-        settings.put(SettingsFrame.MAX_HEADER_LIST_SIZE, getHttpConfiguration().getRequestHeaderSize());
+        int maxTableSize = getMaxDecoderTableSize();
+        if (maxTableSize != 4096)
```
lachlan-roberts: Can we add a constant for this 4096 value.

lachlan-roberts: Is it possible to combine this logic with the `settings.compute` code you have elsewhere?

lachlan-roberts (changes requested):

jetty-http2/http2-client/src/test/java/org/eclipse/jetty/http2/client/SettingsTest.java
sbordet: No. The test is doing a naked write of the bytes of a badly crafted SETTINGS frame, and the write cannot initiate before the previous writes that are initializing the connection are complete.
Unfortunately there is event that signals when these writes are complete.

jetty-http2/http2-hpack/src/main/java/org/eclipse/jetty/http2/hpack/HpackContext.java
sbordet: No, method `resize(int)` explicitly takes the `maxTableSize` as parameter.
We can rename it to `updateMaxTableSize()` or even `setMaxTableSize()`, but it sets the max, not the actual size.
The actual size is only tracked by `add()` and `evict()`.

jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
sbordet: Because in HTTP/3 the SETTINGS frame is sent only once, while here it can be sent multiple times and at any time.
Here it is only the place for the first, mandatory, SETTINGS frame, for which we need to provide sane initial values.

Because SETTINGS frame can be sent later by the application, the values on the decoder are set in `beforeSend()`.

jetty-http2/http2-server/src/main/java/org/eclipse/jetty/http2/server/AbstractHTTP2ServerConnectionFactory.java
sbordet: Not sure what logic you refer to, but client and server have different logics and pull the values from different places (the `HTTP2Client` and the `AbstractHTTP2ServerConnectionFactory`), so would be difficult to share the code.

sbordet: Now using a constant for the default maxTableSize of 4096.

jetty-http2/http2-hpack/src/main/java/org/eclipse/jetty/http2/hpack/HpackContext.java
lachlan-roberts: So in QPACK we have
1. the table capacity - how much can be stored in the table
2. the max table capacity - a value set ONLY by the settings frame, which the table capacity cannot be set to be larger than this by the remote side.
3. the table size - the combined size of the entries currently stored in the table

So what I am saying is it seems like 1 & 2 have been merged into the one name of `maxDynamicTableSize` here which is a little confusing.

jetty-http2/http2-client/src/main/java/org/eclipse/jetty/http2/client/HTTP2ClientConnectionFactory.java
```diff
@@ -54,27 +56,30 @@ public Connection newConnection(EndPoint endPoint, Map<String, Object> context)
...
         @SuppressWarnings("unchecked")
-        Promise<Session> promise = (Promise<Session>)context.get(SESSION_PROMISE_CONTEXT_KEY);
+        Promise<Session> sessionPromise = (Promise<Session>)context.get(SESSION_PROMISE_CONTEXT_KEY);
 
-        Generator generator = new Generator(byteBufferPool, client.getMaxDynamicTableSize(), client.getMaxHeaderBlockFragment());
+        Generator generator = new Generator(byteBufferPool, client.isUseOutputDirectByteBuffers(), client.getMaxHeaderBlockFragment());
         FlowControlStrategy flowControl = client.getFlowControlStrategyFactory().newFlowControlStrategy();
-        HTTP2ClientSession session = new HTTP2ClientSession(scheduler, endPoint, generator, listener, flowControl);
+
+        Parser parser = new Parser(byteBufferPool, client.getMaxDecoderTableCapacity(), client.getMaxResponseHeadersSize());
```
lachlan-roberts: I think technically the `getMaxDecoderTableCapacity()` should only be set on the decoder when the settings frame is about to be sent. Otherwise the application could specify in `SETTINGS` that they do not want to use the dynamic table and it will be ignored.

So this is too early.


I think we are already doing this in `HTTP2Session.configure()` so you should pass a 0 value for the `maxDecoderTableCapacity` to the `Parser` constructor.

lachlan-roberts (changes requested):

sbordet:
@lachlan-roberts I think there is an inherent race that cannot be fixed unless delaying the use of API until both initial SETTINGS frame have been acknowledged.

For example, the client has no special configuration, all defaults, so `client.decoder.maxTableCapacity=4096`.
The client sends the initial SETTINGS frame that may or may not contain that value.
Then the client can send a request that may or may not contain the instruction to resize the table.

Meanwhile the server has not sent the initial SETTINGS frame, but it is configured so that its decoder has `server.decoder.maxTableCapacity=0`.
The server does not read until it has sent the initial SETTINGS frame; when it reads, it reads the client's SETTINGS, which applies to its encoder, and then it reads the request that has been sent using the default values, which may contain a resize instruction that is invalid for the server decoder configuration.

The specification says, section 3.4 of RFC 9113:

> To avoid unnecessary latency, clients are permitted to send additional frames to the server immediately after sending the client connection preface, without waiting to receive the server connection preface.

So it is clear that the spec allows for this window to happen.
Waiting for the SETTINGS reply does not help, as the client should wait for the server initial SETTINGS, which is not mandatory as the above paragraph says.

So yes it's racy, but it is allowed, and it has been fixed in HTTP/3 where the table initial sizes are 0 and all communication about table sizes happens as parts of HEADERS frames.

However, I filed https://github.com/eclipse/jetty.project/issues/9801 about this.

jetty-http2/http2-hpack/src/main/java/org/eclipse/jetty/http2/hpack/HpackContext.java
sbordet: In HPACK we have the same.
If we agree that "max size" == "capacity", then:

* maxTableCapacity is handled by the encoder or the decoder, not by `HpackContext`.
* `HpackContext.maxTableSize == tableCapacity` so `HpackContext` only deals with `tableCapacity` and the actual `tableSize`.

I can rename the fields and the methods if you like, but I thought to leave them unchanged.

lachlan-roberts (approved):
Looks good. All my comments have been addressed.

We have some pre-existing race conditions which will be address separately as part of #9801 so I am happy to go ahead and merge this one.

PR #9754, opened 2023-05-10
Issue #8885 - Restore HttpChannel.Listener

### LINKED ISSUE #8885: Jetty-12, replacement for HttpChannel.Listener
**Jetty version(s)**
Jetty-12

**Enhancement Description**

In Jetty-9/10/11 A `HttpChannel.Listener` can be added as a bean to an `AbstractConnector` and then efficiently retrieved in the constructor of every `HttpChannel`.

Jetty-12 has good interception points (not least with `HttpStream` wrappers), but has no similar convenience method for installing them.  Instead a handler needs to be used to install a stream wrapper on every request.

We could either revive a listener mechanism in 12,  or perhaps instead add a method that would allow a bean added to a connector to efficiently wrap every `HttpStream`?   Or perhaps a combination of both, i.e. a stream wrapper that invokes a new listener?

joakime:
Issue #8885 - Restore HttpChannel.Listener for Jetty 12.


gregw:
I'm not sure all the events are still relevant, but we might also need some more.  Some notes on each one.
 + `void onRequestBegin(Request request, Response response)`.  This can mostly be obtained by just having a handler
  + `void onBeforeDispatch(Request request, Response response)` Dispatch is now kind of a servlet  thing only, makes little sense for us, as we just have the handle dispatch.
  + `void onDispatchFailure(Request request, Response response, Throwable failure)` ditto
  + `void onAfterDispatch(Request request, Response response)` ditto.... except that it would be good to have an `afterHandle` event, which is currently very difficult to capture (need to extend HttpChannel to wrap the onRequest runnable)
  + `void onRequestContent(Request request, Response response, ByteBuffer content)` obtained by wrapping `request.read`
  + `void onRequestContentEnd(Request request, Response response)` ditto
  + `void onRequestTrailers(Request request, Response response)` ditto
  + `void onRequestEnd(Request request, Response response)` ditto I think???
  + `void onRequestFailure(Request request, Response response, Throwable failure)` ditto
  + `void onResponseBegin(Request request, Response response)` this is wrapping either the response.write or HttpStream.send depending
  + `void onResponseCommit(Request request, Response response)` We don't have aggregation, so commit is really just the first HttpStream.send... or I guess it can be a callback.succeeded without a send???
  + `void onResponseContent(Request request, Response response, ByteBuffer content)` wrapper of response.write or HttpStream.send
  + `void onResponseEnd(Request request, Response response)` ditto
  + `void onResponseFailure(Request request, Response response, Throwable failure)` this is just a callback wrapper
  + `void onComplete(Request request, Response response)`. We now have 2 type of complete: completed handler callback and complete the stream callback

The main thing that is difficult to capture is the `afterHandle`, so I'm thinking that perhaps we just add a simple mechanism for that.   Then we write a handler that provides the Listener interface and does all the wrapping necessary of all things to capture all the events and call the listener.   If you don't have a listener handler, then you don't pay the cost of the mechanism.

joakime:
There's a few things that the `HttpChannel.Listener` concept supported in Jetty 9 thru Jetty 11 that we are going to remove support for.

1. Getting events for HttpParser errors.

_Example: No Host failure._

https://github.com/eclipse/jetty.project/blob/a138bfbe1ab41a422448a7a9d95d5fad3495a063/jetty-server/src/test/java/org/eclipse/jetty/server/HttpChannelEventTest.java#L143-L169

2. Getting events for EndPoint failures during a handling.

_Example: having the EndPoint's be closed during the handling._

https://github.com/eclipse/jetty.project/blob/a138bfbe1ab41a422448a7a9d95d5fad3495a063/jetty-server/src/test/java/org/eclipse/jetty/server/HttpChannelEventTest.java#L220-L253

3. Modifying Response Headers before commit.

_Example: Removing headers in response_

https://github.com/eclipse/jetty.project/blob/a138bfbe1ab41a422448a7a9d95d5fad3495a063/jetty-server/src/test/java/org/eclipse/jetty/server/HttpChannelEventTest.java#L172-L217

gregw:
@sbordet @joakime 

So now this PR has:
```java
    public interface Listener extends EventListener
    {
        void onRequestBegin(Request request);
        default void onBeforeHandling(Request request);
        default void onAfterHandling(Request request, boolean handled, Throwable failure);
        default void onRequestRead(Request request, Content.Chunk chunk);
        default void onResponseCommitted(Request request, int status, HttpFields response);
        default void onResponseWrite(Request request, boolean last, ByteBuffer content, Throwable failure);
        default void onComplete(Request request, Throwable failure);
    }
```

Which is read only.      But we have a proposal in #9173 to have a `onResponseBegin(Response response)` that would allow headers to be mutated.  I think we decided that was the same even as `onResponseCommit[ted]`, so perhaps we really need `onResponseCommit(Request request, in status, HttpFields.Mutable fields)`?

But that's a little strange to have just one method that can be a mutator and all the other are read only.

However, we also have the ability to wrap the `HttpStream` with `Request.addHttpStreamWrapper(Function<HttpStream, HttpStream> wrapper)`.   This is a rather clunky mechanism, but it allows mutations on:
```java
public interface HttpStream extends Callback
{
    String getId();
    long getNanoTime();
    Content.Chunk read();
    void demand();
    void prepareResponse(HttpFields.Mutable headers);
    void send(MetaData.Request request, MetaData.Response response, boolean last, ByteBuffer content, Callback callback);
    void push(MetaData.Request resource);
    boolean isCommitted();
    TunnelSupport getTunnelSupport();
    Throwable consumeAvailable();
}
```

Several of those things really should not be wrapped (e.g. `getId`), but we allow wrapping so a) we can do stats collection (but that can now be done by `HttpChannel.Listener`); and b) to intercept IO and headers to do gzip inflation/deflation.

So I'm thinking why not combine the two mechanism.  Get rid of the ability to wrap `HttpStream` and instead just have a listener that can mutate:
```java
    public interface Listener extends EventListener
    {
        default Request onRequestBegin(Request request) { return request; }
        default Request onBeforeHandling(Request request) { return request; }
        default void onAfterHandling(Request request, boolean handled, Throwable failure);
        default Content.Chunk onRequestRead(Request request) {return request.read();}
        default int onResponseCommit(Request request, int status, HttpFields.Mutable response) {return status;}
        default void onResponseWrite(Request request, boolean last, ByteBuffer content, Callback callback, Content.Sink sink) {sink.write(content, last, callback);}
        default void onComplete(Request request, Throwable failure) {}
    }
```

Note that this still cannot handle a request, but it can mutate most things about how a handler sees a request or produces a response.

thoughts?

Edit:   Hmmm we also need to intercept demand for the gzip inflation. Hard to make nice in the listener form rather than the wrapper form.   Could be `default void demain(Runnable streamDemand) { streamDemand.demand();}` but that looks exactly like the `Content.Source.demand(Runnable)` but it has different semantics.     We have a good wrapper style for intercepting content sources, so perhaps this is not a good idea after all, as it introduces yet another style?

gregw:
@lorban could you try using this listening instead of a HttpStream wrapper for the stats handler?

lorban:
@joakime Could you please modify `AbstractLatencyRecordingHandler` to use the listener, and adapt `LatencyRecordingHandlerTest`?

lorban:
I added an API proposal to this PR. Sorry for hijacking it but I would like the discussion to get started.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -101,6 +107,123 @@ public interface HttpChannel extends Invocable
...
+
+        /**
+         * Invoked after a response content chunk has been written to the network.
+         *
+         * @param request the request object
+         * @param last indicating last write
+         * @param content a {@link ByteBuffer#slice() slice} of the response content chunk
+         * @param failure if there was a failure to write the given content
+         */
+        default void onResponseWrite(Request request, boolean last, ByteBuffer content, Throwable failure)
```
gregw: I don't like the expense of calling this one. I'll do a branch suggesting an alternative... Stand by...

gregw (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
gregw: @joakime see #9872 for a cheaper implementation for write events (split into two). Also I improved some of the comments

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java
```diff
@@ -188,6 +191,23 @@ public AbstractConnector(
         _bufferPool = bufferPool != null ? bufferPool : server.getByteBufferPool();
         addBean(_bufferPool, bufferPool != null);
 
+        addEventListener(new Container.Listener()
+        {
+            @Override
+            public void beanAdded(Container parent, Object bean)
+            {
+                if (bean instanceof HttpChannel.Listener)
+                    _httpChannelListeners.set(getBeans(HttpChannel.Listener.class));
```
sbordet, line 200: Seems a bit clumsy to re-scan all the beans, and re-create all the listeners, works less well concurrently, etc.
Can't we just add the new bean?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -101,6 +107,123 @@ public interface HttpChannel extends Invocable
...
+     * <p>Listener methods are invoked synchronously from the thread that is
+     * performing the request processing, and they should not call blocking code
+     * (otherwise the request processing will be blocked as well).</p>
+     * <p>Listener instances that are set as a bean on the {@link Connector} are
+     * efficiently added to {@link HttpChannel}.
+     */
+    public interface Listener extends EventListener
+    {
+        /**
+         * Invoked just after the HTTP request line and headers have been parsed.
```
sbordet: Wrap with `<p>` (and likewise everywhere else).

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -101,6 +107,123 @@ public interface HttpChannel extends Invocable
...
+         * Invoked just before calling the server handler tree.
+         *
+         * <p>
+         * This is the final state of the request before the handlers are called.
+         * This includes any request customization.
+         * </p>
+         *
+         * @param request the request object
+         */
+        default void onBeforeHandling(Request request)
```
sbordet, line 161: I think we are missing an event -- `onRequestHeaders()` when all the headers have arrived.
For example, with the delayed handlers, there could be a difference between `onRequestHeaders()` and `onBeforeHandling()`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -101,6 +107,123 @@ public interface HttpChannel extends Invocable
...
+         * making it available to the application.
+         *
+         * <p>
+         *     TODO: make notes about Trailers / EOF / Errors
+         * </p>
+         *
+         * @param request the request object
+         * @param chunk a request content chunk TODO: make bytebuffer a slice?
+         */
+        default void onRequestRead(Request request, Content.Chunk chunk)
```
sbordet, line 190: Is it worth to have a `onRequestDemand()` event?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -101,6 +107,123 @@ public interface HttpChannel extends Invocable
...
+        }
+
+        /**
+         * Invoked just before the response line is written to the network.
+         *
+         * @param request the request object
+         * @param status the response status
+         * @param response the response object
+         */
+        default void onResponseCommitted(Request request, int status, HttpFields response)
```
sbordet, line 203: I would rename this method, as it reads as if the response has already been committed, while it is not.
`onResponseBegin()` would be better.

And I think we should have an `onResponseCommitted()` that is fired _after_ the response has been committed.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannelListeners.java
```diff
@@ -13,267 +13,236 @@
...
+            LISTENER_HANDLER_ON_RESPONSE_WRITE = lookup.findVirtual(HttpChannel.Listener.class, "onResponseWrite", LISTENER_TYPE_ON_RESPONSE_WRITE);
+
+            LISTENER_HANDLER_ON_COMPLETE = lookup.findVirtual(HttpChannel.Listener.class, "onComplete", LISTENER_TYPE_ON_COMPLETE);
         }
-        catch (Exception e)
+        catch (NoSuchMethodException e)
+        {
+            throw new RuntimeException(e);
+        }
+        catch (IllegalAccessException e)
```
sbordet: Duplicate catch, fold into previous.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannelListeners.java
```diff
@@ -13,267 +13,236 @@
...
-    public void onBeforeDispatch(Request request)
+    public HttpChannelListeners()
     {
-        onBeforeDispatch.onRequest(request);
+        set(null);
     }
 
-    @Override
-    public void onDispatchFailure(Request request, Throwable failure)
+    public void set(Collection<HttpChannel.Listener> listeners)
```
sbordet, line 104: Weird semantic, I would have expected `add` semantic as beans can be added while running, and there is no need to not notify existing listeners of events because they are being rebuilt.

gregw (changes requested):
See comments on #8885 as we are now doubtful that this listener is needed.

sbordet (commented):
Partial review, but we discussed this online.

PR #9756, opened 2023-05-10
H3: Fix racy read from stream-less channel

### LINKED ISSUE #9655: Flaky test in jetty-12 org.eclipse.jetty.test.client.transport.HttpClientLoadTest#testIterative
**Jetty version(s)**
12

**Description**

When building locally:
```
[ERROR] Failures: 
[ERROR]   HttpClientLoadTest.testIterative:98->run:176 [Result failed Result[HttpRequest[GET /20 HTTP/3.0]@2a159815 > HttpResponse[HTTP/3.0 200 null]@6f170333] java.util.concurrent.TimeoutException: idle timeout 30000 ms elapsed] ==> expected: <true> but was: <false>

```

lorban:
`HttpReceiverOverHTTP3.read()` fails when it is called while `HttpChannelOverHTTP3.getStream()` returns null.

This is caused by a race condition in the client when it is setting the stream, which happens _after_ a new request is created and its headers were sent on the network. If the network and the server happen to be fast enough, the response could arrive at the client before the channel's stream has been set, making `HttpReceiverOverHTTP3.read()` see a `null` stream.

The solution is to set the stream field before any data is sent onto the network, like HTTP/2 is going.

Fixes https://github.com/eclipse/jetty.project/issues/9655

### DIFF
--- a/jetty-core/jetty-http3/jetty-http3-client-transport/src/main/java/org/eclipse/jetty/http3/client/transport/internal/HttpReceiverOverHTTP3.java
+++ b/jetty-core/jetty-http3/jetty-http3-client-transport/src/main/java/org/eclipse/jetty/http3/client/transport/internal/HttpReceiverOverHTTP3.java
@@ -13,6 +13,7 @@
 
 package org.eclipse.jetty.http3.client.transport.internal;
 
+import java.io.EOFException;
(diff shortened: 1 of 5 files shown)


lorban:
Superceded by https://github.com/eclipse/jetty.project/pull/9761

Issue #9682, opened 2023-04-27
RetainableByteBuffer buffer release bug in WebSocket


SerCeMan:
**Jetty version(s)**

The extreme case was observed on _10.0.13_, however, the same symptoms seem to be observable on _10.0.15_.

**Java version/vendor** `(use: java -version)`

```
openjdk version "17.0.5" 2022-10-18 LTS
OpenJDK Runtime Environment Zulu17.38+21-CA (build 17.0.5+8-LTS)
OpenJDK 64-Bit Server VM Zulu17.38+21-CA (build 17.0.5+8-LTS, mixed mode, sharing)
```

**OS type/version**

Linux, containers

**Description**

Hi, Jetty folks! We've observed a few cases where a running instance of Jetty seems to leak native memory through direct ByteBuffers. From our investigation, we've found that the likely cause is a single bucket `RetainedBucket` that ends up holding a large number of `RetainableByteBuffer` instances – 346k of them which amounts to ~5G of native memory. The native memory usage seems to steadily grow over time – e.g. 5G native memory usage was observed after the application was running for a couple of days. 

![image](https://user-images.githubusercontent.com/1780970/234734804-4b3247c0-6792-40bd-a805-e18cc3f904f7.png)

We're continuing the investigation mainly by adding additional instrumentation, however, I wonder if you have any suggestions for whether what we're seeing is expected under any specific circumstances. 

**How to reproduce?**

I don't have a solid reproducer at the moment. However, some notes about the environment that might be relevant: 
* It's a Jetty application that primarily handles WebSockets. 
* It uses Conscrypt for SSL.

gregw:
@SerCeMan is this isolated incidents, or does the memory for this app always grow over time even after a restart?

SerCeMan:
It does grow over time even after restart, although sometimes it decreases (not by much). However, most of the time the growth is manageable, and it typically stays under 1GB of direct memory used. However, in some cases, maybe once a week, all instances in the cluster observe a growth in the amount of direct memory consumed. 

We'll hopefully be able to provide more information as the investigation continues, but I was wondering if you had any ideas of the top of your head for what could be causing this behaviour. Thanks!

gregw:
[...]

PR #9759, opened 2023-05-11
Issue #9682 - fix RetainableByteBuffer release bug in WebSocket

lachlan-roberts:
## Issue #9682

I suspect this is the cause for issue #9682, but not enough information to confirm.

The test `SuspendResumeTest.timeoutWhileSuspended()` reproduces a the leak where the buffer from the RetainableByteBufferPool is not released after the websocket connection is closed.

This could happen in two cases.
1. The application does not demand another frame and there is still data in the network buffer, then an idleTimeout happens and the connection is closed and the network buffer still contains data and is never released from WebSocketConnection.
2. The application suspends with the Jetty WebSocket API, the frame+callback are stored away until resume is called, then the idleTimeout fires and the connection is closed, the delayed callback is never completed so the network buffer is never released.

### DIFF
--- a/jetty-websocket/websocket-jetty-common/src/main/java/org/eclipse/jetty/websocket/common/JettyWebSocketFrameHandler.java
+++ b/jetty-websocket/websocket-jetty-common/src/main/java/org/eclipse/jetty/websocket/common/JettyWebSocketFrameHandler.java
@@ -198,6 +198,7 @@ public class JettyWebSocketFrameHandler implements FrameHandler
                     break;
 
                 case SUSPENDING:
+                    assert (delayedFrame == null && delayedCallback == null);
                     delayedFrame = frame;
                     delayedCallback = callback;
                     state = SuspendState.SUSPENDED;
@@ -283,18 +284,18 @@ public class JettyWebSocketFrameHandler implements FrameHandler
     @Override
     public void onClosed(CloseStatus closeStatus, Callback callback)
     {
-        Callback suspendedCallback;
+        Callback delayedCallback;
         try (AutoLock l = lock.lock())
         {
             // We are now closed and cannot suspend or resume.
             state = SuspendState.CLOSED;
-            delayedFrame = null;
-            suspendedCallback = delayedCallback;
-            delayedCallback = null;
+            this.delayedFrame = null;
+            delayedCallback = this.delayedCallback;
+            this.delayedCallback = null;
         }
 
-        if (suspendedCallback != null)
-            suspendedCallback.failed(new CloseException(closeStatus.getCode(), closeStatus.getCause()));
+        if (delayedCallback != null)
+            delayedCallback.failed(new CloseException(closeStatus.getCode(), closeStatus.getCause()));
 
         notifyOnClose(closeStatus, callback);
         container.notifySessionListeners((listener) -> listener.onWebSocketSessionClosed(session));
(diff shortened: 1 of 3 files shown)


jetty-websocket/websocket-jetty-common/src/main/java/org/eclipse/jetty/websocket/common/JettyWebSocketFrameHandler.java
```diff
@@ -457,13 +465,13 @@ public void resume()
 
                 case SUSPENDED:
                     needDemand = true;
-                    delayedFrame = delayedOnFrame;
-                    delayedOnFrame = null;
+                    frame = delayedFrame;
+                    callback = delayedCallback;
```
sbordet, line 470: Should not you null out the fields, like you did before?

jetty-websocket/websocket-jetty-tests/src/test/java/org/eclipse/jetty/websocket/tests/SuspendResumeTest.java
```diff
@@ -189,4 +200,49 @@ public void testSuspendAfterClose() throws Exception
         // suspend after closed throws ISE
         assertThrows(IllegalStateException.class, () -> clientSocket.session.suspend());
     }
+
+    @Test
+    public void timeoutWhileSuspended() throws Exception
```
sbordet: Rename to `testTimeoutWhileSuspended()` following the pattern we use elsewhere.

sbordet (changes requested):

jetty-websocket/websocket-jetty-common/src/main/java/org/eclipse/jetty/websocket/common/JettyWebSocketFrameHandler.java
```diff
@@ -199,7 +198,8 @@ public void onFrame(Frame frame, Callback callback)
                     break;
 
                 case SUSPENDING:
-                    delayedOnFrame = () -> onFrame(frame, callback);
+                    delayedFrame = frame;
```
gregw, line 202: I'm so glad we are getting rid of suspend/resume!
Can you add an `assert delayedFrame==null` here just to show that we know that there can only be 1 delayed frame.

jetty-websocket/websocket-jetty-common/src/main/java/org/eclipse/jetty/websocket/common/JettyWebSocketFrameHandler.java
```diff
@@ -283,12 +283,19 @@ private void onCloseFrame(Frame frame, Callback callback)
     @Override
     public void onClosed(CloseStatus closeStatus, Callback callback)
     {
+        Callback suspendedCallback;
         try (AutoLock l = lock.lock())
         {
             // We are now closed and cannot suspend or resume.
             state = SuspendState.CLOSED;
+            delayedFrame = null;
+            suspendedCallback = delayedCallback;
```
gregw: Not a big fan of mixing the `suspended` and `delayed` naming here.   How about calling the local variable `theDelayedCallback`.  (Note that is why I like the underscore prefix!)

gregw (dismissed):
Fix looks good, but some niggles.

gregw (approved):

sbordet (approved):

PR #9761, opened 2023-05-11
H3: Fix racy read from stream-less channel

### LINKED ISSUE #9655: Flaky test in jetty-12 org.eclipse.jetty.test.client.transport.HttpClientLoadTest#testIterative

lorban:
There is a race condition in the client when it is setting the stream, which happens _after_ a new request is created and its headers are sent onto the network. If the network and the server happen to be fast enough, the response could arrive at the client before the channel's stream has been set, making `HttpReceiverOverHTTP3` see a `null` stream.

The solution is to set the stream field before any data is sent onto the network, like HTTP/2 is doing.

Fixes https://github.com/eclipse/jetty.project/issues/9655 once merged to 12.

### DIFF
--- a/jetty-http3/http3-client/src/main/java/org/eclipse/jetty/http3/client/internal/HTTP3SessionClient.java
+++ b/jetty-http3/http3-client/src/main/java/org/eclipse/jetty/http3/client/internal/HTTP3SessionClient.java
@@ -106,6 +106,7 @@ public class HTTP3SessionClient extends HTTP3Session implements Session.Client
             return promise;
 
         stream.setListener(listener);
+        notifyNewStream(listener, stream);
 
         stream.writeFrame(frame)
             .whenComplete((r, x) ->
@@ -125,6 +126,21 @@ public class HTTP3SessionClient extends HTTP3Session implements Session.Client
             });
 
         return promise;
+    }
+
+    private void notifyNewStream(Stream.Client.Listener listener, HTTP3StreamClient stream)
+    {
+        if (listener != null)
+        {
+            try
+            {
+                listener.onNewStream(stream);
+            }
+            catch (Throwable x)
+            {
+                LOG.info("Failure while notifying listener {}", listener, x);
+            }
+        }
     }
 
     @Override
--- a/jetty-http3/http3-client/src/main/java/org/eclipse/jetty/http3/client/internal/HTTP3SessionClient.java
+++ b/jetty-http3/http3-client/src/main/java/org/eclipse/jetty/http3/client/internal/HTTP3SessionClient.java
@@ -106,7 +106,7 @@ public class HTTP3SessionClient extends HTTP3Session implements Session.Client
             return promise;
 
         stream.setListener(listener);
-        notifyNewStream(listener, stream);
+        stream.onOpen();
 
         stream.writeFrame(frame)
             .whenComplete((r, x) ->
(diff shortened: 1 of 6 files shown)


jetty-http3/http3-client/src/main/java/org/eclipse/jetty/http3/client/internal/HTTP3SessionClient.java
```diff
@@ -106,6 +106,7 @@ private CompletableFuture<Stream> newRequest(long streamId, HeadersFrame frame,
             return promise;
 
         stream.setListener(listener);
+        notifyNewStream(listener, stream);
```
sbordet: I'd like this to be moved to `HTTP3StreamClient` via a new method `stream.onOpen()`, because `HTTP3StreamClient` has all the other `notifyXYZ()` methods.

jetty-http3/http3-client/src/main/java/org/eclipse/jetty/http3/client/internal/HTTP3SessionClient.java
```diff
@@ -127,6 +128,21 @@ private CompletableFuture<Stream> newRequest(long streamId, HeadersFrame frame,
         return promise;
     }
 
+    private void notifyNewStream(Stream.Client.Listener listener, HTTP3StreamClient stream)
```
sbordet: Move it to `HTTP3StreamClient`.

jetty-http3/http3-http-client-transport/src/main/java/org/eclipse/jetty/http3/client/http/internal/HttpReceiverOverHTTP3.java
```diff
@@ -59,9 +59,24 @@ protected void receive()
...
             responseSuccess(exchange);
+        }
         else
-            getHttpChannel().getStream().demand();
+        {
+            Stream stream = getHttpChannel().getStream();
+            if (LOG.isDebugEnabled())
+                LOG.debug("Demanding from {} in {}", stream, this);
+            if (stream == null)
+                return;
```
sbordet, line 71: Ok for this additional check, but did you remove the similar check in `read(boolean)` that was in #9756?

sbordet (changes requested):

jetty-http3/http3-client/src/main/java/org/eclipse/jetty/http3/client/internal/HTTP3SessionClient.java
lorban: Done.

jetty-http3/http3-http-client-transport/src/main/java/org/eclipse/jetty/http3/client/http/internal/HttpReceiverOverHTTP3.java
lorban: Done.

jetty-http3/http3-client/src/main/java/org/eclipse/jetty/http3/client/internal/HTTP3SessionClient.java
lorban: Done.

jetty-http3/http3-http-client-transport/src/main/java/org/eclipse/jetty/http3/client/http/internal/HttpReceiverOverHTTP3.java
lorban: No, I did not. This change won't merge to 12 and my plan is to replace this change with the `if (stream == null) return Content.Chunk.from(new EOFException())` check.

sbordet (approved):

PR #9768, opened 2023-05-12
Java 17 switch semantics

preveen-stack:
- change to java 17 switch expression wherever possible


gregw:
Thanks for this PR.  However, our policy is not to update jetty-10 to use the new style switch semantics.   
All our active development is on the jetty-12 branch, where we are switching to the new switch style when appropriate.

So please look at that branch if you wish to contribute cleanup like this.

Closing as we will not change in jetty-10 nor jetty-11.

PR #9773, opened 2023-05-15
Fixes #9767 - jetty-12 ee10 ServerPush failures

### LINKED ISSUE #9767: jetty-12 ee10 ServerPush failures
### Tck test: com/sun/ts/tests/servlet/spec/serverpush/Client.java.serverPushNegtiveTest

This test performs a number of checks. The 2 that are failing:

1. `PushBuilder.method(null)` must throw NPE according to the spec, but we throw `IAE` instead
2. `PushBuilder.method(m)`  must throw `IAE` if the method is considered unsafe, which according to the `PushBuilder` javadoc are:  `POST, PUT, DELETE, CONNECT, OPTIONS and TRACE`. However, we do _not_ consider `OPTIONS` and `TRACE` to be unsafe.

sbordet:


janbartel (approved):
+1 if the test failures are nothing to do with this change.

PR #9775, opened 2023-05-16
Issue #9774 Disallow cross context

### LINKED ISSUE #9774: jetty-12 ee10 Cross context dispatch is not supported
jetty-12 does not (yet) support cross-context dispatch.  Thus, we need to return `null` for calls to the method `ServletContextApi.getContext(String)`.

This is also tck related, as there are many cross-context dispatch tests which should really test if the container supports cross-context before proceeding with the test - see https://github.com/jakartaee/servlet/issues/488

janbartel:
Fixes #9774

### DIFF
--- a/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextHandler.java
+++ b/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextHandler.java
@@ -2712,9 +2712,6 @@ public class ServletContextHandler extends ContextHandler implements Graceful
         public ServletContext getContext(String uripath)
         {
             //TODO jetty-12 does not currently support cross context dispatch
-            String myPath = getContextPath();
-            if (myPath != null && myPath.equals(uripath))
-                return this;
             return null;
         }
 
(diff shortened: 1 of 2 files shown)


jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextHandler.java
```diff
@@ -2711,87 +2711,10 @@ public ContextHandler getContextHandler()
...
-                        contexts.add(ch);
-                }
-            }
-
-            if (contexts.size() > 0)
-                return contexts.get(0).getServletContext();
+            //TODO jetty-12 does not currently support cross context dispatch
+            String myPath = getContextPath();
+            if (myPath != null && myPath.equals(uripath))
+                return this;
```
gregw: I think we should always return null as we have suggested that `servletContext.getContext(servletContext.getContextPath())==null` is a good test for if cross context dispatch is supported or not.

gregw (changes requested):

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextHandler.java
janbartel: I did it the same way that ee9 is handling `getContext(String)`, which is to return the current context if there's a match, see: https://github.com/eclipse/jetty.project/blob/jetty-12.0.x/jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/ContextHandler.java#L1750

Do you want me to change both ee9 and ee10 to return null?

gregw: yes please.  This way the TCK tests can simply check if it is supported by trying to lookup themselves

gregw (approved):

PR #9779, opened 2023-05-16
Fix #9777 CrossOriginFilter Vary Header

### LINKED ISSUE #9777: CrossOriginFilter does not return Vary header on no-cors mode
**Jetty version(s)**
11.0.13

**Java version/vendor**
openjdk version "11.0.16" 2022-07-19
OpenJDK Runtime Environment (build 11.0.16+8-post-Ubuntu-0ubuntu118.04)
OpenJDK 64-Bit Server VM (build 11.0.16+8-post-Ubuntu-0ubuntu118.04, mixed mode, sharing)

**OS type/version**
Linux 5.15.0-71-generic #78-Ubuntu SMP Tue Apr 18 09:00:29 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

**Description**
When running your server with the CrossOriginFilter default filter, OPTIONS responses do not return a `Vary` header on `no-cors` requests. 
So, others `cors` requests to the same resource uses the `no-cors` request cached headers and lack cors headers such as `Origin` or `Access-Control-Allow-Origin`. 

As explained here, https://fetch.spec.whatwg.org/#cors-protocol-and-http-caches :
In particular, consider what happens if `Vary` is not used and a server is configured to send `Access-Control-Allow-Origin` for a certain resource only in response to a CORS request. When a user-agent receives a response to a non-CORS request for that resource (for example, as the result of a navigation request), the response will lack `Access-Control-Allow-Origin` and the user agent will cache that response. Then, if the user-agent subsequently encounters a CORS request for the resource, it will use that cached response from the previous non-CORS request, without `Access-Control-Allow-Origin`.

After opening this below issue, the submitted fix doesn't seem to fix this issue. 
https://github.com/eclipse/jetty.project/issues/1927

**How to reproduce?**
1. Configure your server to use Jetty's default `CrossOriginFilter`.
2. Load a resource through your browser with a `no-cors` request, the `Vary: Origin` header is not present on the OPTIONS response due to `no-cors`.
3. Load the same resource with a `cors` request, your user-agent won't include the `Origin` & `Access-Control-Allow-Origin` header due to caching, the request will be rejected.

gregw:
Fix #9777 by always setting the Vary Header.

### DIFF
--- a/jetty-servlets/src/test/java/org/eclipse/jetty/servlets/CrossOriginFilterTest.java
+++ b/jetty-servlets/src/test/java/org/eclipse/jetty/servlets/CrossOriginFilterTest.java
@@ -83,8 +83,8 @@ public class CrossOriginFilterTest
         String rawResponse = connector.getResponse(request);
         HttpTester.Response response = HttpTester.parseResponse(rawResponse);
 
-        assertThat(response.toString(), response.getStatus(), is(HttpStatus.OK_200));
         assertTrue(latch.await(1, TimeUnit.SECONDS));
+        assertThat(response.toString(), response.getStatus(), is(HttpStatus.OK_200));
         assertThat(response.get(HttpHeader.VARY), is(HttpHeader.ORIGIN.asString()));
     }
 
--- a/jetty-servlets/src/test/java/org/eclipse/jetty/servlets/CrossOriginFilterTest.java
+++ b/jetty-servlets/src/test/java/org/eclipse/jetty/servlets/CrossOriginFilterTest.java
@@ -24,6 +24,7 @@ import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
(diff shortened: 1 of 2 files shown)


jetty-servlets/src/test/java/org/eclipse/jetty/servlets/CrossOriginFilterTest.java
```diff
@@ -84,6 +85,7 @@ public void testRequestWithNoOriginArrivesToApplication() throws Exception
 
         assertThat(response.toString(), response.getStatus(), is(HttpStatus.OK_200));
         assertTrue(latch.await(1, TimeUnit.SECONDS));
+        assertThat(response.get(HttpHeader.VARY), is(HttpHeader.ORIGIN.asString()));
```
sbordet, line 88: The asserts on the response be together after the assert on the latch.

sbordet (changes requested):

sbordet (approved):

PR #9781, opened 2023-05-17
Issue #9657 Upgrade request should throw ServletException

### LINKED ISSUE #9657: jetty-12 ee9 & ee10 Request.upgrade returns null
According to the javadoc for `HttpServletRequest` class, the `upgrade` method must throw `ServletException` if the handler fails to be instantiated:

```
     * @exception IOException if an I/O error occurred during the upgrade
     * @exception ServletException if the given <code>handlerClass</code> fails to be instantiated
```

We are returning `null` as it is not yet implemented. We should at least throw a hard `ServletException`.

janbartel:
Closes #9657


olamy:
maybe we could have simple unit tests?

sbordet (approved):

PR #9782, opened 2023-05-17
Remove duplicated classes from test session, fix some to be able to run in parallel and not restarting 3rd party for each tests

olamy:
- ee10 mongo session test in parallel
- cleanup pom
- move common class MongoTestHelper to a tools module
- fix dependencies
- jetty-ee9-test-sessions run in parallel and use common code to start mongodb
- use common class for memcached
- use common class for memcached
- make jdbc tests parallel
- make jdbc session tests work in parallel
- infinispan tests in parellel
- rework infinispan tests to avoid copies of same class
- fix duplicate classes infinispan test, do no restart image for each test
- infinispan reuse classes for ee9 as well
- hazelcast tests in parallel, remove duplicated classes
- init infinispan only at startup and upgrade Docker image to 11.0.14.Final
- reuse code for gcloud test and start container only once then reuse it for parallel tests
- better naming
- extract more common classes
- remove commented content


tests/jetty-test-session-common/pom.xml
```diff
@@ -0,0 +1,143 @@
...
+  <name>Tests :: Session Test Tools</name>
+  <packaging>jar</packaging>
+
+  <properties>
+    <bundle-symbolic-name>${project.groupId}.session.test.tools</bundle-symbolic-name>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>jakarta.servlet</groupId>
```
janbartel: Can't be common if it depends on servlet 6 api?

jetty-integrations/jetty-infinispan/jetty-infinispan-remote-query/pom.xml
```diff
@@ -9,7 +9,6 @@
   <name>Integrations :: Infinispan :: Remote with Querying</name>
   <properties>
     <bundle-symbolic-name>${project.groupId}.infinispan.remote.query</bundle-symbolic-name>
-    <infinispan.docker.image.version>11.0.9.Final</infinispan.docker.image.version>
```
janbartel, line 12: where's the version of infinispan image configured now, I can't find it?

olamy: top pom (https://github.com/eclipse/jetty.project/pull/9782/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R200). I even keep a comment maybe we can remove the comment now? :)

tests/jetty-test-session-common/pom.xml
olamy: I was not sure about this. but it's marked as provided and optional.
It's only few classes from servlet spec.
the real artifact version is used in each ee9/ee10 module

olamy: changed

janbartel:
@olamy is this ready for review?

olamy:
@janbartel yes I fixed your first comments

janbartel (approved):

PR #9783, opened 2023-05-17
Jetty 12 simplify EchoHandler

gregw:
Simplify EchoHandler to make it a better example Handler.

### DIFF
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpField.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpField.java
@@ -398,6 +398,11 @@ public class HttpField
         return _name.equalsIgnoreCase(name);
     }
 
+    public boolean is(HttpHeader header)
+    {
+        return Objects.equals(_header, header);
+    }
+
     private int nameHashCode()
     {
         int h = this._hash;
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpField.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpField.java
@@ -398,11 +398,6 @@ public class HttpField
         return _name.equalsIgnoreCase(name);
     }
 
-    public boolean is(HttpHeader header)
-    {
-        return Objects.equals(_header, header);
-    }
-
     private int nameHashCode()
     {
         int h = this._hash;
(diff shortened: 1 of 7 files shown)


jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpField.java
```diff
@@ -398,6 +398,11 @@ public boolean is(String name)
         return _name.equalsIgnoreCase(name);
     }
 
+    public boolean is(HttpHeader header)
+    {
+        return Objects.equals(_header, header);
+    }
+
```
sbordet: Never used, remove it.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpFields.java
```diff
@@ -90,6 +91,17 @@ default HttpFields asImmutable()
         return HttpFields.build(this).asImmutable();
     }
 
+    /**
+     * Implements the {@link Supplier} of {@code HttpFields} contract so a {@code HttpFields} instance can be
+     * supply itself.
+     * @return The HttpFields itself.
```
sbordet:
```suggestion
     * <p>Supplies this instance, typically used to supply HTTP trailers.</p>
     * @return this instance
```

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/EchoHandler.java
```diff
@@ -32,26 +31,32 @@ public class EchoHandler extends Handler.Abstract.NonBlocking
...
+                    case CONTENT_LENGTH ->
+                    {
+                        response.getHeaders().add(field);
+                        contentLength = field.getLongValue();
+                    }
+                    case CONTENT_TYPE -> response.getHeaders().add(field);
+                    case TRAILER -> response.setTrailersSupplier(HttpFields.build());
+                    case TRANSFER_ENCODING -> contentLength = Long.MAX_VALUE;
+                }
+            }
```
sbordet, line 56: Quite harder to read, because now different concerns are in the same code blob. 
-0 but I won't oppose too much.

sbordet (changes requested):

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/EchoHandler.java
gregw: wow I find it so much easier to read as each concern is now only 1 or 2 lines and not spread over multiple blocks of code.

sbordet (approved):

PR #9789, opened 2023-05-20
Jetty 12 http field processor for #9173

### LINKED ISSUE #9173: Configuring SameSite on a per-cookie basis in Jetty 12

gregw:
Address #9173 with a HttpField processor


gregw:
@sbordet @joakime would be good to get some feedback this week.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
```diff
@@ -592,6 +592,12 @@ public Builder secure(boolean secure)
             return this;
         }
 
+        public Builder sameSite(SameSite sameSite)
+        {
+            _attributes = lazyAttributePut(_attributes, SAME_SITE_ATTRIBUTE, sameSite.attributeValue);
+            return this;
+        }
```
sbordet, line 599: This method does not cover the removal of the attribute? Also other methods don't do it.
I think that it is necessary when you do:

```java
HttpCookie.build(someOtherCookie)
    .sameSite(null) // clear whatever the other cookie had
    .build();
```

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpField.java
```diff
@@ -411,6 +412,11 @@ public boolean is(String name)
         return _name.equalsIgnoreCase(name);
     }
 
+    public boolean is(HttpHeader header)
+    {
+        return Objects.equals(_header, header);
+    }
```
sbordet: This is only used in 1 place, and it's the equivalent of `field.getHeader() == header` so I'd remove it.

sbordet (changes requested):
A couple of bits here, but mostly see the comments I made in general about this approach.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpCookie.java
gregw: fixed

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpField.java
gregw: I know it is not strictly necessary, but it is symmetric with the existing `is(String)` method.
Using `HttpHeader`s should be at least as convenient as using header names, so I'd prefer to keep it.

sbordet: I think it's confusing. Why having a method to replace `==`?
We use the `==` form in other 18 places.

It is not symmetric with `is(String)` because the comparison in that case is case-insensitive.

sbordet:
@gregw see my comment, I don't particularly like this approach -- I think using a channel event is better.

gregw:
@sbordet i agree that a channel listener approach would be better. But i think we should merge the changes i made in this PR that cache the cookie and help reposting it if necessary.   I'll try to rebase on top of event PR.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpField.java
gregw: As i said. Symmetry

gregw:
@sbordet actually we don't need the listener for this (and it doesn't support response mutation anyway (see discussion in #9754)).
I've just pushed an alternative implementation that uses the existing `HttpStream` wrapper, which works OK.

thoughts?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpField.java
gregw: more reasons:
 + I think the code reads better with it.   
 + As I'm used to using `field.is(String)` I instinctively try to use `field.is(HttpHeader)` and am always surprised when it is not there. 
 + I have found and fixed inefficient code that does `field.is(HttpHeader.SOME_HEADER.asString())`
 + it avoids the need to worry about null returns from `field.getHeader()`

But I'm not going to die in a ditch for this.  If you really don't like it, I'll remove it.

gregw:
@joakime @sbordet ready for re-review

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpField.java
sbordet: Yes, I'd prefer it to be removed.

sbordet (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpField.java
gregw: sigh!

sbordet (approved):

PR #9790, opened 2023-05-22
Fixed FCGI content parsing.

sbordet:
WordPress sends the initial content in the same frame as the response headers. StreamContentParser was receiving the frame, delegating to HttpParser to parse the headers, but the buffer was still containing some content, that was lost. Now the content after the headers is correctly retained.


lorban:
Does this bug only affect 12 and not 10/11?

lorban (approved):
LGTM but I wonder if that fix shouldn't go to 10/11 as well.

lorban (approved):

sbordet:
@lorban I verified that this is not a problem in 10/11.
The reason is that in 10/11 we have a "push" model for the content, so we parse the headers, notify the application, then continue parsing, find the content, notify the application, so the buffer is fully consumed and the content is not lost.
In 12 we have a "pull" model for the content, so we stop parsing after the headers because we give control to the application to call `Content.Source.read()`, but by doing so we re-enter the parsing but lost the content.

PR #9791, opened 2023-05-22
Fix flaky test testBlockingReadInADifferentThread

### LINKED ISSUE #9765: Flaky test  testBlockingReadInADifferentThread - org.eclipse.jetty.http3.tests.DataDemandTest
**jetty-12**

https://jenkins.webtide.net/blue/organizations/jenkins/jetty.project/detail/PR-9764/1/tests

lorban:
Fix a race condition that can happen when one thread demands and another reads.

Fixes #9765

### DIFF
--- a/jetty-core/jetty-http3/jetty-http3-common/src/main/java/org/eclipse/jetty/http3/HTTP3StreamConnection.java
+++ b/jetty-core/jetty-http3/jetty-http3-common/src/main/java/org/eclipse/jetty/http3/HTTP3StreamConnection.java
@@ -125,9 +125,8 @@ public abstract class HTTP3StreamConnection extends AbstractConnection
                 case FRAME ->
                 {
                     action.getAndSet(null).run();
-                    // Release the network buffer here (if empty), since the application may
-                    // not be reading more bytes, to avoid to keep around a consumed buffer.
-                    tryReleaseBuffer(false);
+                    // Do not release the buffer here to avoid races with
+                    // user-spawned threads that may call Stream.read().
                 }
             }
         }


jetty-core/jetty-http3/jetty-http3-common/src/main/java/org/eclipse/jetty/http3/HTTP3StreamConnection.java
```diff
@@ -125,9 +125,11 @@ private void processDataFrames(boolean setFillInterest)
...
                 {
                     action.getAndSet(null).run();
-                    // Release the network buffer here (if empty), since the application may
-                    // not be reading more bytes, to avoid to keep around a consumed buffer.
-                    tryReleaseBuffer(false);
+                    // We cannot release the buffer eagerly here as that would race
+                    // with another thread executing user code calling Stream.read();
+                    // that user thread could throw NPE in HTTP3StreamConnection.parseAndFill()
+                    // when trying to access the networkBuffer field that is set
+                    // to null here depending on which thread wins the race - see #9765.
```
sbordet: I'd simplify to:

```suggestion
                    // Do not release the buffer here to avoid races with 
                    // user-spawned threads that may call Stream.read().
```

The reason to say less is to avoid that the comment rots if we change the implementation.

sbordet (changes requested):

jetty-core/jetty-http3/jetty-http3-common/src/main/java/org/eclipse/jetty/http3/HTTP3StreamConnection.java
lorban: Done.

sbordet (approved):

PR #9798, opened 2023-05-25
review and cleanup of HTTP/3 QPACK Integer and String encoding

lachlan-roberts:
- Cleanup the code in `NBitIntegerEncoder` to reduce duplication.
- Change implementation of `NBitIntegerEncoder.octetsNeeded()` to include an extra byte for prefixes less than 8, this means we can get rid of the `+ 1` everywhere it is used.
- Added a `NBitStringEncoder` utility class to simplify where we encode strings and calculate their required lengths.

### DIFF
--- a/jetty-http/src/main/java/org/eclipse/jetty/http/compression/NBitIntegerEncoder.java
+++ b/jetty-http/src/main/java/org/eclipse/jetty/http/compression/NBitIntegerEncoder.java
@@ -25,96 +25,64 @@ public class NBitIntegerEncoder
     }
 
     /**
-     * @param n the prefix used to encode this long.
-     * @param i the integer to encode.
+     * @param prefix the prefix used to encode this long.
+     * @param value the integer to encode.
      * @return the number of octets it would take to encode the long.
      */
-    public static int octetsNeeded(int n, long i)
+    public static int octetsNeeded(int prefix, long value)
     {
-        if (n == 8)
-        {
-            int nbits = 0xFF;
-            i = i - nbits;
-            if (i < 0)
-                return 1;
-            if (i == 0)
-                return 2;
-            int lz = Long.numberOfLeadingZeros(i);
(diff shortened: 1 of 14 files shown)


jetty-http/src/main/java/org/eclipse/jetty/http/compression/NBitIntegerEncoder.java
```diff
@@ -25,96 +25,64 @@ private NBitIntegerEncoder()
...
             {
-                buf.put(p, (byte)((buf.get(p) & ~bits) | i));
-            }
-            else
-            {
-                buf.put(p, (byte)(buf.get(p) | bits));
-
-                long length = i - bits;
-                while (true)
+                if ((length & ~0x7F) == 0)
```
sbordet: I would add a comment saying that `~0x7F` is different from `0x80` because of all the `1`s from the msb.
I would also use `~0x7FL` so that it's clear this is an `&` between 2 `long`s.

jetty-http/src/main/java/org/eclipse/jetty/http/compression/NBitIntegerEncoder.java
```diff
@@ -25,96 +25,64 @@ private NBitIntegerEncoder()
...
+        if (value < 0)
             return 1;
-        int lz = Long.numberOfLeadingZeros(i);
+        if (value == 0)
+            return 2;
+        int lz = Long.numberOfLeadingZeros(value);
         int log = 64 - lz;
-        return (log + 6) / 7;
+
+        // The return value is 1 for the prefix + the number of 7-bit groups used to encode the value.
```
sbordet:
```suggestion
        // The return value is 1 for the prefix + the number of 7-bit groups necessary to encode the value.
```

sbordet (changes requested):

sbordet (approved):

PR #9805, opened 2023-05-30
Do not create `AsynchronousCloseException` unnecessarily

### LINKED ISSUE #9524: InputStreamResponseListener's InputStream creates an exception on close()
**Jetty version(s)**
_[Jetty 9.x is now at End of Community Support](https://github.com/eclipse/jetty.project/issues/7958)_
jetty-10
jetty-11
jetty-12

**Enhancement Description**
It seems like `InputStreamResponseListener#Input` always creates an `AsynchronousCloseException` instance on close(), where in many cases is not even needed. I noticed this takes a relatively high amount of time when running [this Solr test](https://github.com/apache/solr/blob/main/solr/benchmark/src/java/org/apache/solr/bench/search/SimpleSearch.java) with a profiler. Would it make sense to only create such instance in the case of callbacks present? Or am I missing something?

lorban:
Minor perf improvement.

Fixes #9524

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/util/InputStreamResponseListener.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/util/InputStreamResponseListener.java
@@ -359,8 +359,11 @@ public class InputStreamResponseListener extends Listener.Adapter
             if (LOG.isDebugEnabled())
                 LOG.debug("InputStream close");
 
-            Throwable failure = new AsynchronousCloseException();
-            callbacks.forEach(callback -> callback.failed(failure));
+            if (!callbacks.isEmpty())
+            {
+                Throwable failure = new AsynchronousCloseException();
+                callbacks.forEach(callback -> callback.failed(failure));
+            }
 
             super.close();
         }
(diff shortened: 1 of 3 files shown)


sbordet (dismissed):

jetty-client/src/main/java/org/eclipse/jetty/client/util/InputStreamResponseListener.java
```diff
@@ -359,8 +359,11 @@ public void close() throws IOException
             if (LOG.isDebugEnabled())
                 LOG.debug("InputStream close");
 
-            Throwable failure = new AsynchronousCloseException();
-            callbacks.forEach(callback -> callback.failed(failure));
+            if (!callbacks.isEmpty())
+            {
+                Throwable failure = new AsynchronousCloseException();
+                callbacks.forEach(callback -> callback.failed(failure));
```
gregw, line 365: Since this is performance, perhaps don't use for each? Actually we determined that streams are slower than a simple for loop, but does that include forEach ?

sbordet: Callbacks are only present if the stream was not read until the end, so should be a rare case and `forEach` should not matter much.
In the normal case, the callbacks would be empty but we were always creating the exception.

sbordet (approved):

PR #9867, opened 2023-06-04
Jetty 12 graceful contexts

gregw:
@sbordet @lorban I was cleaning up TODO's in ContextHandler and saw that it didn't implement `Graceful`, so I looked at the `GracefulHandler` and saw that it had a bug, as it assumed a throw meant that the callback had not been called - thus a handler that completes the callback and then throws could be counted twice.   Fixing that, I re-remembered the shortfalls of Callback, in that many of them are not truly atomic.       So I was going to fix that just for `GracefulHandler` and `ContextHandler`, but then decided to do it for all the utility callbacks in `Callback.java`.    I'm not 100% sold on this, but am wondering if there is actually any performance impact.  @lorban can you run the perf tests on this branch to see?

If we don't like the extra atomics in `Callback` then we can just make a `Callback.AtomicNested` and use it where we think it is important.   But if there is no significant performance impact, may as well use it everywhere... specially as I can add some debugging to help detect double calls.


jetty-core/jetty-deploy/src/main/java/org/eclipse/jetty/deploy/providers/ContextProvider.java
```diff
@@ -317,6 +317,9 @@ else if (Supplier.class.isAssignableFrom(context.getClass()))
         if (context instanceof Deployable deployable)
             deployable.initializeDefaults(properties);
 
+        if (Boolean.valueOf(properties.get("graceful")))
```
sbordet: Use `parseBoolean()` instead of `valueOf()`.
However, I don't understand this?
`ContextHandler implements Graceful` and the properties likely won't have `graceful=true`, so it won't be set here.

I would have expected these lines to not be necessary and `ContextHandler` just be graceful.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -199,6 +197,40 @@ protected ScopedContext newContext()
...
+     * @return true if {@link #shutdown()} will wait for all requests to complete.
+     */
+    public boolean isGraceful()
+    {
+        return _graceful;
+    }
+
+    /**
+     * Set if this context should gracefully shutdown by waiting for all current requests to
+     * complete.  This can be set via the "graceful" property of the {@code DeploymentManager}.
```
sbordet: `DeploymentManager` has not properties that can be set, so the javadoc should be improved to say that this can be set at deployment time via a properties file with the same name of the webapp to deploy.

However, why we need this? `ContextHandler` implements `Graceful` so why we need this at all?

Furthermore, seems confusing to me to set `graceful=false`, but then be able to call `shutdown()` which is supposed to be graceful but it is not.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -199,6 +197,40 @@ protected ScopedContext newContext()
...
+
+        _requests.reset();
+        if (!graceful && _shutdown != null)
+            _shutdown.complete(null);
+    }
+
+    /**
+     * @return The number of current requests, or -1 if {@link #isGraceful()} is false.
+     */
+    public long getCurrentRequests()
```
sbordet: Annotate with `@ManagedAttribute`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -591,10 +619,34 @@ public boolean isShutdown()
...
+                    checkShutdown();
+                    return _shutdown;
+                }
+            }
+        }
+    }
+
+    protected void checkShutdown()
+    {
+        if (_graceful && !_shutdown.isDone() && _requests.sum() == 0)
```
sbordet: Isn't this mechanism of counting the requests a duplicate of what `GracefulHandler` already does?
The idea was to extract it out from `StatisticsHandler` and have just `GracefulHandler` to all the work, no?
Why do we need it also here?

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ContextHandlerTest.java
```diff
@@ -891,4 +899,162 @@ void assertInContext(Context context, Request request)
                 assertThat(r, sameInstance(request));
         }
     }
+
+    @Test
+    public void testGraceful() throws Exception
```
sbordet, line 905: I would consider this test invalid, as I would expect a `GracefulHandler` to wrap the `ContextHandler` if we want to test gracefulness, no?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Callback.java
```diff
@@ -155,23 +157,33 @@ static Callback from(InvocationType invocationType, Runnable success, Consumer<T
     {
         return new Callback()
         {
+            private final AtomicBoolean _completed = new AtomicBoolean();
```
sbordet: We have `CountingCallback` that can be used with `count=1` with the same behavior, so I would use that and revert these changes to `Callback`.

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/HttpChannel.java
```diff
@@ -1530,7 +1530,7 @@ public void failed(final Throwable x)
             if (x instanceof HttpException httpException)
             {
                 MetaData.Response responseMeta = new MetaData.Response(httpException.getCode(), httpException.getReason(), HttpVersion.HTTP_1_1, HttpFields.build().add(HttpFields.CONNECTION_CLOSE), 0);
-                send(_request.getMetaData(), responseMeta, null, true, new Nested(this)
+                send(_request.getMetaData(), responseMeta, null, true, new Nested(this.getCallback())
```
sbordet: No need for `this.getCallback()`, just `getCallback()`.

sbordet (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Callback.java
lorban, line 1: I am +1 on the idea but -1 on this implementation, as it feels like a half-empty glass.

I'm tempted by the idea of mimicking `AbstractLifeCycle` with `public final` `success` and `fail` methods and `protected` `doSucceed` and `doFail` methods, but that sounds like a huge change so late in the development cycle.

gregw: Yeah, I don't think we should do this now... or even ever.    The ways that callbacks combine are really varied and it is too complex to capture in a single simple abstraction.  Ultimately any actions that are not idempotent need to be protected themselves rather than relying on abstract callbacks that may or may not have been combined with other behaviours that may throw/block etc.

If we were to do something, then an abstract like you suggest would be OK for many situations.

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/HttpChannel.java
gregw: It is needed as I have removed the strange "copy" constructor.    The previous behaviour relied on the fact that the passed callback was a nested, so that instead of being wrapped, it was replaced.  This was entirely non intuitive and relied on the type of the passed callback, which a user might not even know that the callback they have is actually implemented as a nested.... so if they strongly type it, they get different behaviour.  Specifically the following code:
```java
SomeCallback callback = new SomeCallback();
call(new Nested(callback));
```
may not have the same behaviour as:
```java
Callback callback = new SomeCallback();
call(new Nested(callback));
```

This is fragile (cost me many hours trying to figure out), so I removed the "copy" constructor and the replacement of the nested callback is now explicit in the error handling here.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/handler/ContextHandlerTest.java
gregw: Can be done either wrapped or inside.  I've updated the test to use GracefulHandler, so it is kind of a double test of that handler, but at least we test it with ContextHandler now.

gregw:
@lorban @sbordet This PR now removes `Graceful` from `ContextHandler`.  It fixes the bug in `GracefulHandler` and adds a test of `GracefulHandler` with a `ContextHandler`.  It also does some cleanup of `Callback`, but nothing revolutionary.   Ready for re-review when green build again.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Callback.java
```diff
@@ -202,13 +203,37 @@ public void completed()
...
+            @Override
+            public InvocationType getInvocationType()
+            {
+                return invocationType;
+            }
+
+            @Override
+            public String toString()
+            {
+                return "Callback@%x{%s,%s}".formatted(hashCode(), invocationType, completed);
```
sbordet: Should not this be "Completing@%x..."? Or at least "Callback.Completing@%x..."?

jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/HttpChannel.java
sbordet: Sure, I meant to remove `this.` and just have `new Nested(getCallback())`.

sbordet (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulHandler.java
```diff
@@ -32,24 +34,30 @@ public class GracefulHandler extends Handler.Wrapper implements Graceful
...
+                long count = getCurrentRequests();
                 if (LOG.isDebugEnabled())
                     LOG.debug("isShutdownDone: count {}", count);
                 return count == 0;
             }
         };
     }
 
+    @ManagedAttribute("current requests")
+    public long getCurrentRequests()
```
lorban: Maybe rename to `getCurrentRequestCount()`?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Callback.java
```diff
@@ -202,13 +203,37 @@ public void completed()
...
+        return new Completing()
         {
             @Override
             public void completed()
             {
                 completed.run();
             }
+
+            @Override
+            public void succeeded()
```
lorban: It feels like `succeeded()` and `completed()` should have default implementations calling `completed()` in the interface.

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Callback.java
```diff
@@ -437,45 +435,7 @@ static Callback combine(Callback cb1, Callback cb2)
...
-                }
-            }
-
-            @Override
-            public InvocationType getInvocationType()
-            {
-                return Invocable.combine(Invocable.getInvocationType(cb1), Invocable.getInvocationType(cb2));
-            }
-        };
+        return from(cb1, cb2);
```
lorban: The `from(Callback cb1, Callback cb2)` implementation is more brittle than the one it replaces here as it does not try to manage the exceptions.

IMHO the `from(Callback cb1, Callback cb2)` implementation should become what has been deleted here.

lorban (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/Callback.java
gregw: I'm not sure there is a correct behaviour when callbacks start throwing... but yeah, I switched it around

lorban (approved):

sbordet (approved):

PR #9870, opened 2023-06-05
Use default HttpExchange#setStreams implementation

### LINKED ISSUE #9150: jetty-http-spi: Jetty's implementation of HttpExchange.setStreams method faulty
**Jetty version(s)**
_[Jetty 9.x is now at End of Community Support](https://github.com/eclipse/jetty.project/issues/7958)_
Jetty 9.x, 10.x, 11.x (12 not reviewed)

java version "1.8.0_271"
Java(TM) SE Runtime Environment (build 1.8.0_271-b09)
Java HotSpot(TM) 64-Bit Server VM (build 25.271-b09, mixed mode)

**OS type/version**
Linux mclane 5.19.0-26-generic #27-Ubuntu SMP PREEMPT_DYNAMIC Wed Nov 23 20:44:15 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
Ubuntu 22.10

**Description**
This is a bug related to the issue #9125, which I reported earlier. Writing a workaround for that bug, I discovered another bug, which I report here. In respect to the priorities of the Jetty team and the likeliness that this bug (and the other) will ever be fixed, it has been said enough in the other issue. No need to repeat. I create this issue simply to document the bug.

The abstract built-in class `HttpExchange` is implemented in Jetty by the `JettyHttpExchange` class, which delegates all calls to the `JettyHttpExchangeDelegate` class (for whatever reason).

The method `setStreams (in, out)` allows to replace input and output streams, mostly for filter implementations, but not necessarily limited to them. The method has a special calling convention, namely, that if the  `in` or `out` parameter is `null`, the respective stream is left unchanged. Unfortunately, the method `setStreams` in `JettyHttpExchangeDelegate` simply replaces the internal member variables with the given parameters, thus not adhering to the convention.

**How to reproduce?**

* create a HttpHandler or Filter implementation
* call `getRequestBody` or `getResponseBody` on the provided `HttpExchange` instance and verify that the returned streams are not null
* call `setStreams (null, null)`, which should effectively do nothing
* call `getRequestBody` or `getResponseBody`; the returned streams are `null` instead of the values returned earlier

strogiyotec:
Right now the `JettyHttpExchangeDelegate#setStreams` implementation just sets given parameters to respective steams. However, the default JDK implementation of `HttpExchange#setStreams` calls `ExchangeImpl#setStreams` which in turn only changes the respective streams when given parameters aren't null. To follow this convention the  `JettyHttpExchangeDelegate#setStreams` was rewritten to have the same implementation

### DIFF
--- a/jetty-http-spi/src/main/java/org/eclipse/jetty/http/spi/JettyHttpExchangeDelegate.java
+++ b/jetty-http-spi/src/main/java/org/eclipse/jetty/http/spi/JettyHttpExchangeDelegate.java
@@ -209,8 +209,11 @@ public class JettyHttpExchangeDelegate extends HttpExchange
     @Override
     public void setStreams(InputStream i, OutputStream o)
     {
-        _is = i;
-        _os = o;
+        assert _is != null;
+        if (i != null)
+            _is = i;
+        if (o != null)
+            _os = o;
     }
 
     @Override


joakime:
There is no valid Eclipse ECA on file for the email addresses on these commits, and as such this PR is not legally allowed to be merged.

Please review https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/CONTRIBUTING.md#eclipse-contributor-agreement

strogiyotec:
@joakime  sorry I wasn't familiar with it , I signed the agreement , feel free to take a look at the PR when you have time

jetty-http-spi/src/main/java/org/eclipse/jetty/http/spi/JettyHttpExchangeDelegate.java
```diff
@@ -209,8 +209,15 @@ public void setAttribute(String name, Object value)
...
-        _os = o;
+        assert _is != null;
+        if (i != null)
+        {
+            _is = i;
+        }
+        if (o != null)
+        {
+            _os = o;
+        }
```
sbordet: Drop the braces in case of one-liners.

```suggestion
        if (i != null)
            _is = i;
        if (o != null)
            _os = o;
```

sbordet (changes requested):

strogiyotec:
@sbordet  I made the changes please take a look

sbordet (approved):

sbordet:
@strogiyotec for the next PRs, please do not force-push, because for a reviewer it becomes harder to review, particularly with many files to review.

Thanks!

strogiyotec:
got it thanks for taking time and helping me

sbordet:
@strogiyotec Thanks!

PR #9872, opened 2023-06-05
Refine HttpChannel.Listener

gregw:
addendum to #9754 
better comments
split write into two cheaper events

### DIFF
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
@@ -128,47 +128,48 @@ public interface HttpChannel extends Invocable
      * <p>Listener instances that are set as a bean on the {@link Connector} are
      * efficiently added to {@link HttpChannel}.
      */
-    public interface Listener extends EventListener
+    interface Listener extends EventListener
     {
         /**
-         * Invoked just after the HTTP request line and headers have been parsed.
+         * Invoked just after the HTTP request line and headers have been parsed
+         * (i.e. from within the call to {@link HttpChannel#onRequest(MetaData.Request)}).
          *
          * <p>
          * This the state of the request from the network, and does not include
          * any request customizations (eg: forwarding, secure, etc)
          * </p>
          *
-         * <p>
-         * Implementations must not block!
-         * </p>
-         *
-         * @param request the request object
+         * @param request the request object, which should not be mutated in any way.
+         * @see HttpChannel#onRequest(MetaData.Request)
          */
         default void onRequestBegin(Request request)
         {
         }
 
         /**
-         * Invoked just before calling the server handler tree.
+         * Invoked just before calling the server handler tree (i.e. just before the {@link Runnable}
+         * returned from {@link HttpChannel#onRequest(MetaData.Request)} is run).
          *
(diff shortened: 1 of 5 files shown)


jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannelListeners.java
```diff
@@ -45,6 +45,7 @@ public class HttpChannelListeners implements HttpChannel.Listener
 
     private static final MethodType LISTENER_TYPE_ON_RESPONSE_COMMITTED = methodType(Void.TYPE, Request.class, Integer.TYPE, HttpFields.class);
     private static final MethodType LISTENER_TYPE_ON_RESPONSE_WRITE = methodType(Void.TYPE, Request.class, Boolean.TYPE, ByteBuffer.class);
+    public static final MethodHandle EMPTY_ON_RESPONSE_WRITE_HANDLE = MethodHandles.empty(LISTENER_TYPE_ON_RESPONSE_WRITE);
```
joakime: Don't do this, you risk pinning a classloader using `MethodHandles.empty()` as a static final.

gregw: @joakime Huh? This class already has plenty of static final `MethodHandle`s that are on actual classes?  How can an empty handle pin a classloader?

gregw: @joakime I changed to use a boolean anyway.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -128,95 +128,116 @@ public interface HttpChannel extends Invocable
      * <p>Listener instances that are set as a bean on the {@link Connector} are
      * efficiently added to {@link HttpChannel}.
      */
-    public interface Listener extends EventListener
+    interface Listener extends EventListener
     {
         /**
-         * Invoked just after the HTTP request line and headers have been parsed.
+         * Invoked just after the HTTP request line and headers have been parsed
+         * (i.e. from within the call to {@link HttpChannel#onRequest(MetaData.Request)}).
```
sbordet, line 135: Wrap with `<p>`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -128,95 +128,116 @@ public interface HttpChannel extends Invocable
...
          * This the state of the request from the network, and does not include
          * any request customizations (eg: forwarding, secure, etc)
          * </p>
          *
-         * <p>
-         * Implementations must not block!
-         * </p>
-         *
-         * @param request the request object
+         * @param request the request object. The {@code read()} and {@code demand(Runnable)} methods must not be called by the listener.
```
sbordet, line 142: I think calling `read()` and `demand()` should not be a worry -- if called wrongly it's the user fault.

sbordet (changes requested):

sbordet (commented):
Partial review, but we have discussed this online.

PR #9883, opened 2023-06-07
Jetty 12.0.x 9072 move core ee classes

janbartel:
Completes #9072 

Moved `Deployable` into `jetty-core/jetty-server`, moved `ConstraintAware` and `ConstraintMapping` into `jetty-ee10/jetty-ee10-servlet`


gregw:
@janbartel something is still looking for jetty-ee in the CodeQL?

janbartel:
> @janbartel something is still looking for jetty-ee in the CodeQL?
Yes, just pushed a fix - some poms got missed.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Deployable.java
```diff
@@ -11,14 +11,20 @@
...
 import java.util.function.Predicate;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+/**
+ * Deployable
+ *
+ * Communicates to the DeploymentManager that a context can be deployed and should
+ * have its defaults initialized.
+ */
```
gregw:
```suggestion
/**
 * Interface that can be implemented by ContextHandlers within Environments to allow configuration
 * to be passed from the DeploymentManager without dependencies on the Deployment module itself. 
 */
```

gregw (approved):
but with a comment suggestion. Also check on the CI failure looking for jetty-ee

PR #9884, opened 2023-06-07
Fix `QueuedPool`'s performance

### LINKED ISSUE #9881: H2 perf can be pathologically bad
**Jetty version(s)**
12

**Description**
Running the perf test suite against H2C, we sometimes get a successful outcome with the expected performance and sometimes a failure with catastrophic performance 3 or 4 orders of magnitude below the norm.

A quick analysis of [async-profiler flame graphs of both types of runs](https://github.com/eclipse/jetty.project/files/11674441/flamegraphs.zip) indicates a buffer leak as the run with bad performance spends ~80% of its time in `QueuedPool.requeue()`.

lorban:
Manually maintain `QueuedPool`'s size to avoid calling the super-slow `ConcurrentLinkedQueue.size()`. When `QueuedPool` is hammered with a massive concurrent load, the perf difference can be of two orders of magnitude.

Fixes #9881

### DIFF
--- a/jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/internal/QueuedPool.java
+++ b/jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/internal/QueuedPool.java
@@ -18,6 +18,7 @@ import java.util.Collection;
 import java.util.Objects;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicMarkableReference;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
@@ -40,6 +41,7 @@ public class QueuedPool<P> implements Pool<P>
 {
     private final int maxSize;
     private final Queue<Entry<P>> queue = new ConcurrentLinkedQueue<>();
+    private final AtomicInteger queueSize = new AtomicInteger();
 
     // This lock protects the 'terminated' field.
     private final ReadWriteLock rwLock = new ReentrantReadWriteLock();
@@ -56,7 +58,7 @@ public class QueuedPool<P> implements Pool<P>
         rwLock.readLock().lock();
         try
         {
-            if (terminated || queue.size() == maxSize)
+            if (terminated || queueSize.get() == maxSize)
                 return null;
             return new QueuedEntry<>(this);
         }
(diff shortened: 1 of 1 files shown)


gregw (commented):
If we are locking and we know the max size, then why use ConcurrentLinkedQueue?   Surely ArrayList would work fine (perhaps as a stack)?

lorban:
@gregw we use a `ReadWriteLock` to mutually exclude `terminate()` from everything else, so all methods take the read lock before accessing the queue, and `terminate()` is the only one taking the write lock. This lock isn't used for thread safety; all queue access can happen concurrently.

gregw (changes requested):
This class needs a lot more javadoc about how/why the lock works.   It is very confusing to see all the logic inside the read/write lock blocks, yet it is only protecting 1 field and other fields are modified within the write lock.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/internal/QueuedPool.java
```diff
@@ -71,8 +73,9 @@ private boolean requeue(Entry<P> entry)
         rwLock.readLock().lock();
         try
         {
-            if (terminated || queue.size() == maxSize)
+            if (terminated || queueSize.get() == maxSize)
                 return false;
+            queueSize.incrementAndGet();
```
sbordet: Can you turn this into a `while(true)` with a CAS to be precise in updating the size?

sbordet (changes requested):

sbordet (approved):

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/internal/QueuedPool.java
lorban: Done.

lorban:
@gregw If you look at the fields, there are the unprotected ones:

```java
    private final int maxSize;
    private final Queue<Entry<P>> queue = new ConcurrentLinkedQueue<>();
    private final AtomicInteger queueSize = new AtomicInteger();
```
and the only one protected by the lock, together with a comment:
```java
    // This lock protects the 'terminated' field.
    private final ReadWriteLock rwLock = new ReentrantReadWriteLock();
    private boolean terminated;
```

Isn't that clear enough? If not, would you happen to have some suggestions on how I can make that clearer?

gregw:
@lorban I can't say how to improve the comments as I don't really understand the synchronisation.   Why are the atomics being updated holding the read/write lock that protects just one field?  Don't answer here. answer in comments!

gregw (approved):
I'm not sure "fully threadsafe" is the right expression.   Perhaps "racy but of no consequence" would be a better description for any code that relies on 2 atomics.   But OK.

PR #9886, opened 2023-06-07
Remove unused ee10-demo-realm files and add distro test

janbartel:
`ee10-demo-realm.mod`, `ee10-demo-realm.properties` and `ee10-demo-realm.xml` are no longer needed after security moved to core.

Added distribution test to make sure.


janbartel:
The test failure is unrelated to this change.

gregw (approved):

PR #9890, opened 2023-06-07
Fix redirect demo; fix links to sources for demos; fix blog link

janbartel:
* Fixed redirect demo, now redirects to a new static context
* Fixed links to sources of each demo
* Fixed link to jetty blogs


jetty-home/src/main/resources/modules/demo.d/root/index.html
```diff
@@ -104,7 +104,7 @@ <h2>Getting&nbsp;Help</h2>
 
   <div id='blog'>
     <h1>Jetty Developer's Blog</h1>
-    <iframe src="https://webtide.com/blog.jsp" />
+    <iframe src="https://webtide.com/blog" />
```
joakime:
```suggestion
    <iframe src="https://webtide.com/blog/" />
```

joakime (changes requested):
Use the correct blog url.

olamy (approved):

PR #9891, opened 2023-06-08
Issue #9682 - notify WebSocket message sinks of connection close

lachlan-roberts:
## Issue #9682

Notify the websocket MessageSinks of connection close so they can release resources.

### DIFF
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferCallbackAccumulator.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferCallbackAccumulator.java
@@ -89,11 +89,14 @@ public class ByteBufferCallbackAccumulator
 
     public void fail(Throwable t)
     {
-        for (Entry entry : _entries)
+        // In some usages the callback recursively fails the accumulator.
+        // So we copy and clear to avoid double completing the callback.
+        ArrayList<Entry> entries = new ArrayList<>(_entries);
+        _entries.clear();
+        _length = 0;
+        for (Entry entry : entries)
         {
             entry.callback.failed(t);
         }
-        _entries.clear();
-        _length = 0;
     }
 }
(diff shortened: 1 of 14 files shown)


lachlan-roberts:
@sbordet I opened a separate issue for #9892 so it can be addressed in a separate PR.

I think just putting the `activeMessageSink` in a lock doesn't really fix the entire problem, probably should be fixed at the websocket-core level.

And the race would be extremely rare, because being in `FrameHandler.onMessage` means you have just read data so idleTimeout is not about to expire. But I think the fix is out of scope of this PR.

jetty-io/src/main/java/org/eclipse/jetty/io/ByteBufferCallbackAccumulator.java
```diff
@@ -89,11 +89,14 @@ public void writeTo(ByteBuffer buffer)
 
     public void fail(Throwable t)
     {
-        for (Entry entry : _entries)
+        // In some usages the callback recursively fails the accumulator.
+        // So we copy and clear to avoid double completing the callback.
+        ArrayList<Entry> entries = new ArrayList<>(_entries);
+        _entries.clear();
+        _length = 0;
+        for (Entry entry : entries)
```
gregw, line 99: What if `addEntry` is called simultaneously? 
Perhaps a comment is needed saying this class is not threadsafe and callers must do mutual exclusion.

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/messages/MessageInputStream.java
```diff
@@ -218,6 +184,56 @@ private Entry getCurrentEntry() throws IOException
...
+        }
+
+        try
+        {
+            super.close();
+        }
+        catch (IOException e)
+        {
+            LOG.debug("Failure Closing InputStream", e);
+        }
```
gregw: use `IO.close(super::close)` as it does the try catch for you.

gregw (changes requested):

gregw (approved):

PR #9896, opened 2023-06-09
Jetty 12.0.x custom request log  #8819

### LINKED ISSUE #8819: Jetty-12 Improve CustomRequestLog efficiency
**Jetty version(s)**
Jetty-12

**Enhancement Description**

The current handling for servlet attributes to be logged is in `ServletChannel`:
https://github.com/eclipse/jetty.project/blob/jetty-12.0.0.alpha2/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java#L793-L804

This is creating 3 map keys even if they are not used.  Either, they entries should only be created if used; OR create a single key with a Record containing all the attributes that might be logged (or maybe an interface with methods that can fetch the info to be logged, holding a reference to the servlet wrapper).

gregw:
Resolves #8819 CustomRequestLog improvements:
 + only add extra detail if the log is a CustomRequestLog
 + add extra detail as a record
 + get authentication state directly from request attribute

### DIFF
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/CustomRequestLog.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/CustomRequestLog.java
@@ -16,6 +16,7 @@ package org.eclipse.jetty.server;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodType;
+import java.security.Principal;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -73,21 +74,18 @@ import static java.lang.invoke.MethodType.methodType;
  * <th>Format String</th>
  * <th>Description</th>
  * </tr>
- *
  * <tr>
  * <td>X</td>
  * <td>
  * <p>The X character.</p>
  * </td>
  * </tr>
- *
  * <tr>
  * <td>%%</td>
  * <td>
  * <p>The percent character.</p>
  * </td>
  * </tr>
- *
  * <tr>
  * <td>%{format}a</td>
  * <td>
@@ -99,7 +97,6 @@ import static java.lang.invoke.MethodType.methodType;
  * the end-user and the server.</p>
  * </td>
  * </tr>
- *
  * <tr>
  * <td>%{format}p</td>
  * <td>
@@ -111,7 +108,6 @@ import static java.lang.invoke.MethodType.methodType;
  * the end-user and the server.</p>
  * </td>
  * </tr>
- *
  * <tr>
  * <td>%{CLF}I</td>
  * <td>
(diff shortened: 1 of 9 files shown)


mkvl77 (approved):

mkvl77 (approved):

gregw:
@lachlan-roberts @lorban nudge!

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -743,4 +744,39 @@ static HttpURI newHttpURIFrom(Request request, String newEncodedPathInContext)
...
+    /**
+     * A minimal Authentication interface, primarily used for logging.  It is implemented by the
+     * {@code jetty-security} module's {@code AuthenticationState} to provide full authentication services.
+     */
+    interface AuthenticationState
+    {
+        /**
+         * @return The authenticated user {@link Principal}, or null if the Authentication is in a non-authenticated state.
+         */
+        default Principal getUserPrincipal()
```
lorban, line 777: This method is always overridden, which makes me feel it shouldn't have a default implementation.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/CustomRequestLog.java
lorban, line 1: While you're at it, could you please fix `logLatency()` so that it doesn't use `System.currentTimeMillis()` but rather something like `unit.convert(NanoTime.since(request.getNanoTime()), TimeUnit.NANOSECONDS)`?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -743,4 +744,39 @@ static HttpURI newHttpURIFrom(Request request, String newEncodedPathInContext)
...
+    static void setAuthenticationState(Request request, AuthenticationState state)
+    {
+        request.setAttribute(AuthenticationState.class.getName(), state);
+    }
+
+    /**
+     * A minimal Authentication interface, primarily used for logging.  It is implemented by the
+     * {@code jetty-security} module's {@code AuthenticationState} to provide full authentication services.
+     */
+    interface AuthenticationState
```
lorban, line 772: I do not like this name, but I cannot come up with anything better...

lorban (changes requested):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
gregw: I went through several different names.... but ultimately called it what the security layer calls it, as that also went through similar name doubts.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
gregw: There are 5 static usages in ee9 that use this default implementation.  I could introduce an intermediary interface, but this seams simpler.

gregw:
@lorban Using nanotime now for logged latency.

lorban (approved):

Issue #9900, opened 2023-06-12
Improve `Request.getBeginNanoTime()` accuracy


lorban:
**Jetty version(s)**
12

**Enhancement Description**
It currently is very inaccurate for H2, H3, ~~FCGI~~, and when upgrades happen.

See #8885 for when `Request.getBeginNanoTime()` was introduced.

This and `Request.getBeginNanoTime()` should be backported to 10/11 to ease comparing performance across Jetty versions accurately.

sbordet:
Fixed and backported to 10/11.

PR #9901, opened 2023-06-12
Add `EventsHandler` API

### LINKED ISSUE #8885: Jetty-12, replacement for HttpChannel.Listener

lorban:
An abstract `Handler` with overridable `on*` methods fits the Jetty 12 model better than reintroducing `HttpChannel.Listener`.

Closes #8885.
I opened #9900 for the leftover that can wait until after the 12.0.0 release.


documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/client/http3/HTTP3ClientDocs.java
```diff
@@ -117,7 +118,7 @@ public void newStream() throws Exception
             .put(HttpHeader.USER_AGENT, "Jetty HTTP3Client {version}");
 
         // The request metadata with method, URI and headers.
-        MetaData.Request request = new MetaData.Request("GET", HttpURI.from("http://localhost:8444/path"), HttpVersion.HTTP_3, requestHeaders);
+        MetaData.Request request = new MetaData.Request("GET", HttpURI.from("http://localhost:8444/path"), HttpVersion.HTTP_3, requestHeaders, NanoTime.now());
```
gregw: We could have a constructor that doesn't take the timestamp and puts in NanoTime.now() so we would not have all these mods everywhere.   In many cases it is the right value.

jetty-core/jetty-alpn/jetty-alpn-java-client/src/test/java/org/eclipse/jetty/alpn/java/client/JDK9HTTP2ClientTest.java
```diff
@@ -58,7 +58,7 @@ public void testJDK9HTTP2Client() throws Exception
 
             HttpFields.Mutable requestFields = HttpFields.build();
             requestFields.put("User-Agent", client.getClass().getName() + "/" + Jetty.VERSION);
-            MetaData.Request metaData = new MetaData.Request("GET", HttpURI.from("https://" + host + ":" + port + "/"), HttpVersion.HTTP_2, requestFields);
+            MetaData.Request metaData = new MetaData.Request("GET", HttpURI.from("https://" + host + ":" + port + "/"), HttpVersion.HTTP_2, requestFields, 0);
```
gregw: So why 0 here rather than now?
I'm liking the idea of a constructor with now more.

jetty-core/jetty-fcgi/jetty-fcgi-server/src/main/java/org/eclipse/jetty/fcgi/server/internal/HttpStreamOverFCGI.java
```diff
@@ -113,7 +105,7 @@ public void onHeaders()
     {
         String pathQuery = URIUtil.addPathQuery(_path, _query);
         HttpScheme scheme = StringUtil.isEmpty(_secure) ? HttpScheme.HTTP : HttpScheme.HTTPS;
-        MetaData.Request request = new MetaData.Request(_method, scheme.asString(), hostPort, pathQuery, HttpVersion.fromString(_version), _headers, -1);
+        MetaData.Request request = new MetaData.Request(_method, scheme.asString(), hostPort, pathQuery, HttpVersion.fromString(_version), _headers, -1, NanoTime.now()); // TODO #9900 make beginNanoTime accurate
```
gregw: OK to leave now here as a marker for future improvements.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
```diff
@@ -1519,6 +1526,8 @@ public boolean parseNext(ByteBuffer buffer)
                 _methodString = null;
                 _endOfContent = EndOfContent.UNKNOWN_CONTENT;
                 _header = null;
+                if (buffer.hasRemaining())
```
gregw, line 1529: are we ever called with an empty buffer?

To be absolutely correct, it is possible that we are started with some CRLF chars that I think we skip.    Maybe would be better to do this in `setState` and check for leaving the START state?
But I don't want to be too pedantic on this one, so this is OK... but maybe we can do better?

jetty-core/jetty-http2/jetty-http2-common/src/test/java/org/eclipse/jetty/http2/frames/ContinuationParseTest.java
```diff
@@ -68,7 +68,7 @@ public void onConnectionFailure(int error, String reason)
             HttpFields fields = HttpFields.build()
                 .put("Accept", "text/html")
                 .put("User-Agent", "Jetty");
-            MetaData.Request metaData = new MetaData.Request("GET", HttpScheme.HTTP.asString(), new HostPortHttpField("localhost:8080"), "/path", HttpVersion.HTTP_2, fields, -1);
+            MetaData.Request metaData = new MetaData.Request("GET", HttpScheme.HTTP.asString(), new HostPortHttpField("localhost:8080"), "/path", HttpVersion.HTTP_2, fields, -1, 0);
```
gregw: why 0?

jetty-core/jetty-http2/jetty-http2-common/src/test/java/org/eclipse/jetty/http2/frames/FrameFloodTest.java
```diff
@@ -71,7 +71,7 @@ public void testHeadersFrameFlood()
     public void testInvalidHeadersFrameFlood() throws Exception
     {
         // Invalid MetaData (no method, no scheme, etc).
-        MetaData.Request metadata = new MetaData.Request("NULL", null, null, null, HttpVersion.HTTP_2, null, -1)
+        MetaData.Request metadata = new MetaData.Request("NULL", null, null, null, HttpVersion.HTTP_2, null, -1, -1)
```
gregw: Now why -1?   I'd much rather a constructor that doesn't take a value rather than every caller making up a fake time.

gregw (changes requested):
First pass review.

I think you need to add constructors to MetaData that do not take the nanotime and instead insert now themselves.    Most of the changes in this PR are passing the nanotime, and it is sometimes now, sometimes 0 and sometimes -1.    At the very least, this inconsistency needs to be fixed... but having the convenience constructor will avoid the problem entirely.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/client/http3/HTTP3ClientDocs.java
lorban: This doubles the number of constructors in `Metadata.Request` and forces the new `long beginNanoTime` to be at a different position than the last one (otherwise it clashes with the `long contentLength` one) but I don't disagree with you.

jetty-core/jetty-alpn/jetty-alpn-java-client/src/test/java/org/eclipse/jetty/alpn/java/client/JDK9HTTP2ClientTest.java
lorban: It's because the timestamp is not needed in this particular test. But I agree a default constructor would be better.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/HttpParser.java
lorban: I've added a TODO referring to #9900 as when exactly should the begin timestamp be taken should be worked on for all protocols, including H1.

jetty-core/jetty-http2/jetty-http2-common/src/test/java/org/eclipse/jetty/http2/frames/ContinuationParseTest.java
lorban: Replaced with default ctor.

jetty-core/jetty-http2/jetty-http2-common/src/test/java/org/eclipse/jetty/http2/frames/FrameFloodTest.java
lorban: Replaced with default ctor.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
```diff
@@ -0,0 +1,419 @@
...
+            return handled;
+        }
+        catch (Throwable x)
+        {
+            fireOnAfterHandling(roRequest, false, x);
+            throw x;
+        }
+    }
+
+    private void fireOnBeforeHandling(Request request)
```
gregw: I don't mind the "fireOn" names, but elsewhere we have used "notify" for the same purpose, so I'm +0 for changing that... but only if you really want to.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
```diff
@@ -0,0 +1,419 @@
...
+        // Before handling
+        ReadOnlyRequest roRequest = new ReadOnlyRequest(request);
+        fireOnBeforeHandling(roRequest);
+        try
+        {
+            EventsRequest wrappedRequest = new EventsRequest(request, roRequest);
+            EventsResponse wrappedResponse = new EventsResponse(roRequest, response);
+            request.addHttpStreamWrapper(stream -> new HttpStream.Wrapper(stream)
+            {
+                private void fireNotifications(Throwable failure)
```
gregw: "fire" and "notify" are the same thing.  See my other comment about the naming... this could just be "notify".    I'll say no more :)

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
```diff
@@ -0,0 +1,419 @@
...
+     * @param request the request object. The {@code read()}, {@code demand(Runnable)} and {@code fail(Throwable)} methods must not be called by the listener.
+     * @param failure if there was a failure to complete
+     */
+    protected void onComplete(Request request, Throwable failure)
+    {
+    }
+
+    private class EventsResponse extends Response.Wrapper
+    {
+        private boolean notifiedOnResponseBegin;
```
gregw, line 335: OK, I know I said I'd say no more... but again naming is confusing... is notified different to fired? No I'll say no more.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
```diff
@@ -0,0 +1,419 @@
...
+        try
+        {
+            EventsRequest wrappedRequest = new EventsRequest(request, roRequest);
+            EventsResponse wrappedResponse = new EventsResponse(roRequest, response);
+            request.addHttpStreamWrapper(stream -> new HttpStream.Wrapper(stream)
+            {
+                private void fireNotifications(Throwable failure)
+                {
+                    if (!wrappedResponse.notifiedOnResponseBegin)
+                        fireOnResponseBegin(roRequest, wrappedResponse.getHeaders().asImmutable(), wrappedResponse.getStatus());
```
gregw: I know this doesn't need to be atomic, but this test and act looks bad.  Couldn't `fireOnRespnseBegin` be written to check the boolean itself, perhaps returning a boolean to say if it did fire/notify or not?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
```diff
@@ -0,0 +1,419 @@
...
+    {
+        try
+        {
+            Content.Chunk c;
+            if (chunk == null)
+                c = null;
+            else if (chunk.hasRemaining())
+                c = Content.Chunk.asChunk(chunk.getByteBuffer().asReadOnlyBuffer(), chunk.isLast(), chunk);
+            else
+                c = chunk.isLast() ? Content.Chunk.EOF : Content.Chunk.EMPTY;
```
gregw: Is this necessary? why not just pass the chunk:
```suggestion
                c = chunk;
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
```diff
@@ -0,0 +1,419 @@
...
+
+    private void fireOnRequestRead(Request wrapped, Content.Chunk chunk)
+    {
+        try
+        {
+            Content.Chunk c;
+            if (chunk == null)
+                c = null;
+            else if (chunk.hasRemaining())
+                c = Content.Chunk.asChunk(chunk.getByteBuffer().asReadOnlyBuffer(), chunk.isLast(), chunk);
```
gregw: I'm not convinced we need to do this, but good to develop with it and then either remove later or perhaps make optional.   It might be more readable with a utility method:
```suggestion
                c = Content.Chunk.asReadOnly(chunk);
```

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
```diff
@@ -0,0 +1,419 @@
...
+     * This is the final state of the request before the handlers are called.
+     * This includes any request customization.
+     * </p>
+     *
+     * @param request the request object. The {@code read()}, {@code demand(Runnable)} and {@code fail(Throwable)} methods must not be called by the listener.
+     * @see org.eclipse.jetty.server.HttpChannel#onRequest(MetaData.Request)
+     */
+    protected void onBeforeHandling(Request request)
+    {
+    }
```
gregw, line 224: Should these all have LOG.debugs in them?  This would make this handler nice to have for debugging even on its own.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
```diff
@@ -0,0 +1,419 @@
...
+        @Override
+        public Content.Chunk read()
+        {
+            Content.Chunk chunk = super.read();
+            fireOnRequestRead(roRequest, chunk);
+            return chunk;
+        }
+    }
+
+    private static class ReadOnlyRequest extends Request.Wrapper
```
gregw: This should be moved to Request.ReadOnly as it plausible that it will be used elsewhere.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
```diff
@@ -0,0 +1,419 @@
...
+    {
+        private ReadOnlyRequest(Request request)
+        {
+            super(request);
+        }
+
+        @Override
+        public void addIdleTimeoutListener(Predicate<TimeoutException> onIdleTimeout)
+        {
+            throw new UnsupportedOperationException();
```
gregw: Why can't we support this? it is a readOny listener

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
```diff
@@ -0,0 +1,419 @@
...
+        }
+
+        @Override
+        public void addFailureListener(Consumer<Throwable> onFailure)
+        {
+            throw new UnsupportedOperationException();
```
gregw: why can't we support this? it is a readOnly listener.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/LatencyRecordingHandler.java
```diff
@@ -0,0 +1,48 @@
...
+    protected final void onComplete(Request request, Throwable failure)
+    {
+        onRequestComplete(NanoTime.since(request.getBeginNanoTime()));
+    }
+
+    /**
+     * Called back for each completed request with its execution's latency.
+     * @param durationInNs the duration in nanoseconds of the completed request
+     */
+    protected abstract void onRequestComplete(long durationInNs);
```
gregw: I know we cannot pass in the request, but we could pass in the request ID!

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/StatisticsHandler.java
```diff
@@ -295,13 +245,28 @@ public MinimumDataRateHandler(long minimumReadRate, long minimumWriteRate)
             _minimumWriteRate = minimumWriteRate;
         }
 
+        /**
+         * Creates a {@code MinimumDataRateHandler} with the specified read and write rates.
+         * @param minimumReadRate the minimum number of bytes to be read per second, or 0 for not checking the read rate.
+         * @param minimumWriteRate the minimum number of bytes to be written per second, or 0 for not checking the write rate.
+         * @param handler the handler to wrap.
+         */
+        public MinimumDataRateHandler(long minimumReadRate, long minimumWriteRate, Handler handler)
```
gregw, line 254: Does this actually use any of the features of StatsHandler or EventHandler?  Could this not be just a stand alone handler now?

gregw (changes requested):
Mostly niggles or questions

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/LatencyRecordingHandler.java
lorban: Added the request ID.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/StatisticsHandler.java
lorban: It uses `getBytesRead()` and `getBytesWritten()` from `StatsHandler`, so either we duplicate the bytes counting logic, or we keep `MinimumDataRateHandler` as a subclass of `StatsHandler`. I personally prefer to not duplicate the logic.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
lorban: You're right, this does not need to be prevented.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
lorban: You're right, this does not need to be prevented.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
lorban: I've created a `default Request asReadOnly()` method on `Request` instead. Good with you?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
lorban: I've moved the logic to a new `default Chunk asReadOnly()` method in `Chunk`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
lorban: I've simplified this logic and moved it to `Chunk.asReadOnly()`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
lorban: I've moved the test inside the `notify` method.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
lorban: `notify` it shall be.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
lorban: `notify` it shall be.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
lorban: `notify` it shall be.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
lorban: Added debug logs.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
gregw: Oops it is on chunk! that's fine.  Ignore previous comment version

Edited: 
I'm not sure we want another method on the Request that can be wrapped.   A static method would be OK!

gregw (dismissed):
LGTM

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/EventsHandler.java
lorban: Changed `Request.asReadOnly()` to static.

gregw (approved):

Issue #9760, opened 2023-05-11
jetty-12 ee9 Omnibus tck failure analysis


janbartel:
As of build https://jenkins.webtide.net/job/tck/job/servlettck-run-12.0.x-ee9/299/ there are 64 failing ee9 tests. This issue is to track their analysis in a central place.

**Note that:** the CI build is running against `servlet-tck-5`, whilst local testing is running against `servlet-tck-6` as that is the only version that has been extensively modified with arquillian etc to allow individual tests to run.

### works in servlet-6-tck
com/sun/ts/tests/servlet/spec/security/secform/Client.java.test11
com/sun/ts/tests/servlet/spec/security/secform/Client.java.test1
com/sun/ts/tests/servlet/spec/security/secform/Client.java.test10
com/sun/ts/tests/servlet/spec/security/secform/Client.java.test13
com/sun/ts/tests/servlet/spec/security/secform/Client.java.test14
com/sun/ts/tests/servlet/spec/security/secform/Client.java.test17
com/sun/ts/tests/servlet/spec/security/secform/Client.java.test2
com/sun/ts/tests/servlet/spec/security/secform/Client.java.test3
com/sun/ts/tests/servlet/spec/security/secform/Client.java.test4
com/sun/ts/tests/servlet/spec/security/secform/Client.java.test6
com/sun/ts/tests/servlet/spec/security/secform/Client.java.test1_anno
com/sun/ts/tests/servlet/spec/security/secform/Client.java.test2_anno
com/sun/ts/tests/servlet/spec/security/secform/Client.java.test3_anno
com/sun/ts/tests/servlet/spec/security/secform/Client.java.test4_anno
com/sun/ts/tests/servlet/spec/security/secform/Client.java.test6_anno
com/sun/ts/tests/servlet/spec/security/secform/Client.java.test14_anno
com/sun/ts/tests/servlet/spec/requestdispatcher/URLClient.java.getRequestAttributes
com/sun/ts/tests/servlet/spec/requestdispatcher/URLClient.java.getRequestAttributes1
com/sun/ts/tests/servlet/spec/requestdispatcher/URLClient.java.getRequestAttributes3
com/sun/ts/tests/servlet/spec/requestdispatcher/URLClient.java.getRequestAttributes4
com/sun/ts/tests/servlet/spec/requestdispatcher/URLClient.java.getRequestAttributes6
[...]

PR #9902, opened 2023-06-12
Jetty 12.0.x ee9 serverpush tck

janbartel:
Fix the remaining ee9 serverpush tck errors as per #9760


olamy:
good fix for TCK (tested locally and it looks you tested on CI as well)
but there is now a test failure.

janbartel:
@lorban could you take a look at this? Simone has made a relatively simple change in `Http2Stream`, but there are now 2 http2 tests failing. The fix may be as simple as changing the `CountdownLatch` from `2` to `1`, but as you know more about the guts of http2/3 than me, I'd like you to take a look and see if that is reasonable or not.

lorban:
Committed a change to fix H2 trailers that the H2 push fix broke.

olamy (approved):

PR #9903, opened 2023-06-13
Issue #9895 - ensure callback is failed after error in PerMessageDeflateExtension

### LINKED ISSUE #9895: A MessageTooLargeException doesn't close a WebSocket connection
**Jetty version(s)**

11.0.15

openjdk version "11.0.11" 2021-04-20

**OS type/version**

MacOs Ventura (13.2.1)

**Description**

When `org.eclipse.jetty.websocket.core.internal.messages.StringMessageSink#accept` is called with a `Frame` larger than configured the `MessageTooLargeException` is thrown. This behavior is expected.

The problem is that in this state the WS connection is not closed which is not aligned with `org.eclipse.jetty.websocket.api.WebSocketPolicy` JavaDoc - `Text messages over this maximum will result in a close code 1009 {@link StatusCode#MESSAGE_TOO_LARGE}`

The interesting thing is that setting `WebSocketPolicy#setMaxTextMessageSize` with a value lower than the incoming text message solves the problem. The error is propagated as expected. But using a default value doesn't leads to this problem.
 
Right after the exception in DEBUG logging mode, the following information is printed:

```
2023-06-08 22:41:34 [JettyServerThreadPool-35] DEBUG o.e.j.w.c.i.PerMessageDeflateExtension - Decompress finished: true CONTINUATION@58971e91[len=0,fin=false,rsv=000,m=null]HeapByteBuffer@791fd82d[p=0,l=0,c=8192,r=0]={<<<>>>xWZKFpBac...\x00\x00\x00\x00\x00\x00\x00}
2023-06-08 22:41:34 [JettyServerThreadPool-35] DEBUG o.e.j.w.c.i.WebSocketConnection - moreDemand? d=0 fp=true RetainableByteBuffer@112d1841{DirectByteBuffer@542f8ca0[p=4096,l=4096,c=4096,r=0]={/\x88\x15af\x92\xEc\xCd\xA9...VS\x82K\xB2,w<<<>>>},r=2} WebSocketConnection@5d4ab1ad::SocketChannelEndPoint@4c6641e8[{l=/0:0:0:0:0:0:0:1:8839,r=/0:0:0:0:0:0:0:1:64564,OPEN,fill=-,flush=-,to=6347/30000}{io=0/0,kio=0,kro=1}]->[WebSocketConnection@5d4ab1ad[SERVER,p=Parser@f16d1a1[s=FRAGMENT,c=0,o=0x0,m=1dD1217f,l=36967],f=Flusher@71234240[IDLE][queueSize=0,aggregate=null],g=org.eclipse.jetty.websocket.core.internal.Generator@3769740]]
2023-06-08 22:41:34 [JettyServerThreadPool-35] DEBUG o.e.j.u.t.ReservedThreadExecutor - ReservedThread@1a3a4772{RUNNING,thread=Thread[JettyServerThreadPool-35,5,main]} was=RUNNING next=RESERVED size=0+2 capacity=12
2023-06-08 22:41:34 [JettyServerThreadPool-35] DEBUG o.e.j.u.t.ReservedThreadExecutor - ReservedThread@1a3a4772{RESERVED,thread=Thread[JettyServerThreadPool-35,5,main]} waiting ReservedThreadExecutor@6dc1484{reserved=2/12,pending=0}
```
`onWebSocketClose/onWebSocketError` methods are not called. From the client's perspective (Chrome browser in my case) the connection is not closed either. 

Searching through relevant issues I can mention https://github.com/eclipse/jetty.project/issues/7351 as it has similar symptoms.

From my point of view `PerMessageDeflateExtension` should somehow handle the underlying exception (close the connection) but it does nothing like that in this case.

**How to reproduce?**

A reproducer project is available at https://github.com/raymank26/jetty-11.0.15-websocket-permessage-deflate

lachlan-roberts:
fixes #9895

### DIFF
--- a/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/PerMessageDeflateExtension.java
+++ b/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/PerMessageDeflateExtension.java
@@ -464,6 +464,11 @@ public class PerMessageDeflateExtension extends AbstractExtension implements Dem
             chunk.setPayload(payload);
             chunk.setFin(frame.isFin() && complete);
 
+            // If we are complete we return true, then DemandingFlusher.process() will null out the Frame and Callback.
+            // The application may decide to hold onto the buffer and delay completing the callback, so we need to capture
+            // references to these in the payloadCallback and not rely on state of the flusher which may have moved on.
+            // This flusher could be failed while the application already has the payloadCallback, so we need protection against
+            // the flusher failing and the application completing the callback, that's why we use the payload AtomicReference.
             boolean completeCallback = complete;
             AtomicReference<ByteBuffer> payloadRef = _payloadRef;
             Callback payloadCallback = Callback.from(() ->
--- a/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/PerMessageDeflateExtension.java
+++ b/jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/PerMessageDeflateExtension.java
@@ -464,16 +464,18 @@ public class PerMessageDeflateExtension extends AbstractExtension implements Dem
             chunk.setPayload(payload);
             chunk.setFin(frame.isFin() && complete);
 
-            boolean succeedCallback = complete;
+            boolean completeCallback = complete;
             AtomicReference<ByteBuffer> payloadRef = _payloadRef;
             Callback payloadCallback = Callback.from(() ->
             {
                 getBufferPool().release(payloadRef.getAndSet(null));
-                if (succeedCallback)
+                if (completeCallback)
                     callback.succeeded();
             }, t ->
             {
                 getBufferPool().release(payloadRef.getAndSet(null));
+                if (completeCallback)
+                    callback.failed(t);
                 failFlusher(t);
             });
 
--- a/jetty-websocket/websocket-jetty-tests/src/test/java/org/eclipse/jetty/websocket/tests/LargeDeflateTest.java
+++ b/jetty-websocket/websocket-jetty-tests/src/test/java/org/eclipse/jetty/websocket/tests/LargeDeflateTest.java
@@ -33,6 +33,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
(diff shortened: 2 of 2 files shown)


jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/PerMessageDeflateExtension.java
```diff
@@ -464,16 +464,18 @@ private boolean inflate(Frame frame, Callback callback, boolean first) throws Da
...
-                if (succeedCallback)
+                if (completeCallback)
                     callback.succeeded();
             }, t ->
             {
                 getBufferPool().release(payloadRef.getAndSet(null));
+                if (completeCallback)
+                    callback.failed(t);
                 failFlusher(t);
             });
```
gregw, line 485: I'm not exactly following the logic here, which means that it probably needs a comment.
Specifically in the case of complete==true, it is not clear who actually does complete the callback passed into the method.

I also see the atomic payloadRef, is that just protecting from calls to both succeeded and failed?  Or is it racing with something else?
If it is just checking that the other method has not been called, then the other actions should be conditional on a non-null return from the getAndSet, so you don't call the nested callback twice either nor fail the flusher.

gregw (changes requested):

jetty-websocket/websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/internal/PerMessageDeflateExtension.java
lachlan-roberts: If `completeCallback` is true it nulls out the frame callback field after this method so the `payloadCallback` must always complete the frame callback.

If `completeCallback` is false it will not have been cleared out so the `failFlusher()` call will clear the callback when it reaches `onCompleteFailure()`.

The atomic payloadRef is for a different race, can't remember the exact details but will be described in the review comments here #7360.

gregw: probably best to say this in comments rather than here.

The comments in #7360 indicate that the atomic is for the prevention of multiple calls, so other actions should be conditional on a atomic as well, else we do some things correctly once, but others twice.  Since we have the atomic, we should use it!

lachlan-roberts: Note that the `Callback` is an instance of `CountingCallback` and can also be called more than once and uses an `AtomicInteger`.

gregw: @lachlan-roberts I think it could be something like:
```java
            Callback payloadCallback = Callback.from(() ->
            {
                ByteBuffer buffer = payloadRef.getAndSet(null);
                if (buffer == null)
                    return; 
                getBufferPool().release(buffer);
                if (completeCallback)
                    callback.succeeded();
            }, t ->
            {
                ByteBuffer buffer = payloadRef.getAndSet(null);
                if (buffer == null)
                    return; 
                getBufferPool().release(buffer);
                if (completeCallback)
                    callback.failed(t);
                failFlusher(t);
            });

```

gregw: > Note that the `Callback` is an instance of `CountingCallback` and can also be called more than once and uses an `AtomicInteger`.

Yeah but surely it is only called once from each wrapping?

gregw: Either way... this callback needs some javadoc to explain why the atomics are used and who completes the nested callback if completeCallback is false.

lachlan-roberts: This code suggestion will not work. We cannot do the early return if `buffer == null` otherwise the callback will never be completed.

I have now added a lengthy comment above this section of code.

gregw (approved):

PR #9904, opened 2023-06-13
Experiment/jetty 12 chunk isError and warnings

gregw:
Experiment to see:
 + how Chunk works without a known Error type and instead use an isError method.
 + to try transient errors/warnings in a Content.Source


gregw:
@lorban  I have removed the `Content.Chunk.Error` type in favour of a `Content.Chunk.isError` method.    I think that bit is pretty good.

I have also started experimenting with transient failures (calling them warnings for now).   I've updated AsyncSource and the ContentSourceInputStream.  Both mostly trivial updates, but both definitely needed updates.  So on that sample, every `Source` and many `Source.Consumers` will need at least a little change to deal with warnings. 

I've added you as an assignee rather than as a reviewer, so please feel free collaborate and experiment in this branch to see how you like it.   I think there needs to be a lot more tests of mods to the basic `Source`s before we even think of trying to put an idletimeout through as a warning.

gregw:
@sbordet this is ready for your review.

@lorban and I tossed around the naming for a while.   We ended up with:

- All chunks have a `getError()` method, which can return null
- There is a `static boolean Chunk.isError(Chunk)` to make it easy to do a null chunk and null error check
- We use the term "failure" for errors that have last==true
- We use the term "warning" for errors that have last=false

We have updated a few things to handle "warnings":
- `AsyncContent`
- `Content.copy` methods
- `ContentSourceInputStream`
- `ContentDocs`

Our feeling is that this change is good in that it removes the `Error` type and that failure vs warning is pretty clear.  However, almost every caller of `Source.read()` will need careful review about how it handles errors.

Currently nothing we have will inject a warning into any of our sources.   Ultimately `addIdleTimeoutListener` could be changed so that a false return from the listener means use a warning rather than a failure.... or it could be try state?

lorban:
@sbordet here is the gist of this API modification, other changes are about adapting the implementation:

 - Error checking is done with `Chunk.isError()`, not anymore with `instanceof Chunk.Error`, and the latter class got removed from the exposed API.
 - Chunk now has a `Throwable getError()` method:
   - Error chunks must return non-null from `getError()`.
   - Non-error chunks must return null from `getError()`.
 - Error chunks aren't necessarily `last` anymore:
   - When they are last, they are persistent like they used to be. Those are now referred to as _failures_.
   - When they are not last, the following read and demand calls may return non-error data. Those are now referred to as _warnings_.
 -  `Source.warn(Throwable)` has been added next to `Source.fail(Throwable)`.
 - Error chunks cannot be retainable and must be backed by an empty byte buffer.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/ContentDocs.java
```diff
@@ -0,0 +1,148 @@
...
+                            callback.failed(chunk.getError());
+                        else
+                        {
+                            LOG.warn("transient error", chunk.getError());
+                            continue;
+                        }
+                    }
+                    else // otherwise continue with another chunk
+                        continue;
+                    return;
```
sbordet: Very hard to follow.

I would prefer `source.read()` to be the first statement inside the loop, then check for null and return, then check for error and handle+return, then it is a normal chunk and we can write it.

The code above is also weird in case of a last empty chunk, where the write is skipped -- it is really hard to "see" that succeeding the callback also does the last write; for example, the loop above would probably not work for a proxy.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -597,18 +653,46 @@ static Error from(Throwable failure)
...
             if (chunk.isLast())
                 return EOF;
             return null;
         }
 
+        /**
+         * @param chunk The chunk to test for an {@link Chunk#getError() error}.
+         * @return True if the chunk is non-null and {@link Chunk#getError() chunk.getError()} returns non-null.
+         */
+        static boolean isError(Chunk chunk)
```
sbordet: Rename to `isFailure()`.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -597,18 +653,46 @@ static Error from(Throwable failure)
...
+         * <ul>
+         * <li>A {@code chunk} must not have an error and a {@link #getByteBuffer()} with content.</li>
+         * <li>A {@code chunk} with an error may or may not be {@link #isLast() last}.</li>
+         * <li>A {@code chunk} with an error must not be {@link #canRetain() retainable}.</li>
+         * </ul>
+         * @return A {@link Throwable} indicating the error or null if there is no failure or warning.
+         * @see Source#fail(Throwable)
+         * @see Source#warn(Throwable)
+         */
+        default Throwable getError()
```
sbordet: Rename to `getFailure()`.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
```diff
@@ -51,8 +51,8 @@ public String toString()
 
     private final AutoLock.WithCondition lock = new AutoLock.WithCondition();
     private final SerializedInvoker invoker = new SerializedInvoker();
-    private final Queue<AsyncChunk> chunks = new ArrayDeque<>();
-    private Content.Chunk.Error errorChunk;
+    private final Queue<Content.Chunk> chunks = new ArrayDeque<>();
```
sbordet, line 54: This seems just to complicate things, adding more `instanceof`.
Is it really necessary? I'd revert to `AsyncChunk`.

sbordet (dismissed):
@gregw the only think I am worried is a spin loop in case of transient failed Chunks.
Do we have a test case for that?
And is the semantic that once the transient failed Chunk has been read, another read returns null (or something else)?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: @sbordet The names were discussed at length between @lorban and myself.   We did try the renames that you suggest, but they don't really work.

Currently we have `isError` and `getError`, where an "error" can be a failure from `fail(Throwable)` or a warning from `warn(Throwable)`

An alternative is to have `isFailure` and `getFailure`, where a "failure" can be a persistent from `fail(Throwable)` or a either from `fail(Throwable failure, boolean last)`.  @lorban thoughts?

Eitherway, the language used is more than just these 2 methods and updates need to be coordinated through the javadoc and other methods.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: see above

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
gregw: AsyncChunk cannot carry a non-persistent Failure (aka a warning).  So it is necessary.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: @sbordet note that @lorban and I did consider the `fail(Throwable failure, boolean last)` approach, but there are big differences between the semantics of last being true/false.
With last==true, any queued chunks are failed and the source is drained and failed immediately.  With last==false, the "warning" is handled in order with other chunks and no draining occurs.

I guess that can be OK behind the one method with careful javadoc?

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/ContentDocs.java
gregw: @sbordet How about now?

gregw:
> @gregw the only think I am worried is a spin loop in case of transient failed Chunks. Do we have a test case for that? 

Note really.  Since transient failures (aka warnings) are transient, then we'd only loop if something kept raising them over and over again.   Currently we have no mechanism to introduce warnings (aka transient failures) other than in test harnesses.... but if we were, it would be as a result of an `onIdleTimeout` listener callback.  The listener would have to keep saying "treat this as a transient timeout" over and over in order to loop.

The related question, for when we want to actually use these transient failures is should `onIdleTimeout` be bi-state (timeout is permanent or transient) or tri-state (permanent, transient, ignored).       Currently we have (permanent or ignored), so it we just went bi-state, we'd loose the ability to totally ignore a timeout... but I think this is OK.

However, we need to review all our callers of `read()` before we contemplate actually using this mechanism.   This PR looked at a few and each needed small changes, so probably all will need it.      I'm happy to visit all readers... but we need to get the language (transient failure vs warning) sorted out before I do.

> And is the semantic that once the transient failed Chunk has been read, another read returns null (or something else)?

It returns null.  See `Content.Chunk.next(Chunk)`.  It is just a chunk to be consumed like any other non-last chunk.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
lorban: I vote -0 on `fail(Throwable failure, boolean last)` for the reasons @gregw explained: the last flag completely changes the semantic so I'd prefer to have two separate methods, but could live with a single well-documented one.

Error, failure and warning have very specific meanings in this context, so maybe they should be better documented but those three words were carefully chosen.

gregw: I'm also -0 on the single method.   Having said that, I'm also only +0 on the `warn(Throwable)` name.  
It is interesting that to describe what a warning is, I tend to say it is a "transient failure" rather than a "transient error".

So I could definitely be +1 on `isFailure(Chunk)` and `getFailure()` if we could work out a good way of injecting a transient failure.  That way we could talk about "failures" only and their attributes (last or not last), and not have to distinguish between "warning" and "error", as they would not apply.

Perhaps `fail(Throwable)` and `transientFail(Throwable)`?

Or maybe `fail(Throwable failure, boolean last)` is not so bad if we javadoc it really well?

gregw: @lorban thoughts on naming? again?

lorban: I prefer to stick to what we have: `isError`, `getError` and `fail(Throwable)`. I just wouldn't add `warn(Throwable)` on `Content.Source` yet and only have it on `AsyncContent` for now.

gregw: @lorban now you say it... I'm liking less the usage of error, fail and warn,  when we could just do fail and javadoc. so I'm ±0 for the current naming.   @sbordet if you do look at this, what are your thoughts now we've explained the dilema?

gregw:
@lorban can I get a review on this modulo naming?

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/ContentDocs.java
```diff
@@ -0,0 +1,157 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@SuppressWarnings("unused")
+public class ContentDocs
+{
+    private static final Logger LOG = LoggerFactory.getLogger(ContentDocs.class);
+
+    // tag::echo[]
+    void echo(Content.Source source, Content.Sink sink, Callback callback)
```
lorban: This method should be public.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/ContentDocs.java
```diff
@@ -0,0 +1,157 @@
...
+                        // if it is a persistent error, then fail the callback, else log only
+                        if (chunk.isLast())
+                        {
+                            callback.failed(chunk.getError());
+                            return;
+                        }
+                        LOG.warn("transient error", chunk.getError());
+                    }
+
+                    if (chunk.hasRemaining() || chunk.isLast())
```
lorban, line 69: I think this `if` block will trigger an infinite loop once `EOF` is reached.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -327,18 +331,34 @@ default long getLength()
         void demand(Runnable demandCallback);
 
         /**
-         * <p>Fails this content source, possibly failing and discarding accumulated
-         * content chunks that were not yet read.</p>
+         * <p>Fails this content source with a terminal {@link Chunk#getError() error},
```
lorban: A _terminal error_ sounds unclear, I would prefer to read that an error with the last flag set to true will be returned by the next read.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -327,18 +331,34 @@ default long getLength()
...
+         * possibly failing and discarding accumulated content chunks that were not yet read.</p>
```
lorban: _possibly failing and discarding_ is unclear since that is guaranteed. I would just get rid of the _possibly_ word.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -560,9 +580,45 @@ static Chunk asChunk(ByteBuffer byteBuffer, boolean last, Retainable retainable)
...
+                @Override
+                public boolean isLast()
+                {
+                    return last;
+                }
+
+                @Override
+                public String toString()
+                {
+                    return String.format("%s@%x{c=%s,l=%b}", getClass().getSimpleName(), hashCode(), failure, last);
```
lorban: The generated class name will be confusing. How about hardcoding `ErrorContent@%x{...` instead of using `"%s@%x{...", getClassGetSimpleName()`?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -597,18 +653,46 @@ static Error from(Throwable failure)
          */
         static Chunk next(Chunk chunk)
```
lorban, line 667: The javadoc needs to be modified to explain the new non-last error.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceInputStream.java
```diff
@@ -54,8 +54,12 @@ public int read(byte[] b, int off, int len) throws IOException
         {
             if (chunk != null)
             {
-                if (chunk instanceof Content.Chunk.Error error)
-                    throw IO.rethrow(error.getCause());
+                if (Content.Chunk.isError(chunk))
+                {
+                    Content.Chunk c = chunk;
+                    chunk = Content.Chunk.next(chunk); // TODO should we null here and rely on Source to remember state?
```
lorban: I think we should follow what you suggest in the TODO and rely on the source to remember the state.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -69,7 +69,7 @@
  *             return true;
  *         }
  *
- *         if (chunk instanceof Content.Chunk.Error error)
+ *         if (Content.Chunk.isError(chunk))
```
lorban, line 72: The javadoc needs to explain what to do with non-last errors.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/HttpChannelTest.java
```diff
@@ -1214,8 +1213,8 @@ public boolean handle(Request request, Response response, Callback callback)
         Request rq = handling.get().getRequest();
         Content.Chunk chunk = rq.read();
         assertTrue(chunk.isLast());
-        assertInstanceOf(Content.Chunk.Error.class, chunk);
-        assertThat(((Content.Chunk.Error)chunk).getCause(), sameInstance(failure));
+        assertTrue(Content.Chunk.isError(chunk));
```
lorban: `assertTrue(Content.Chunk.isError(chunk) && chunk.isLast());`

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTransformerTest.java
```diff
@@ -259,11 +258,11 @@ public void testTransformThrows()
         chunk.release();
 
         chunk = transformer.read();
-        assertInstanceOf(Content.Chunk.Error.class, chunk);
+        assertTrue(Content.Chunk.isError(chunk));
```
lorban: `assertTrue(Content.Chunk.isError(chunk) && chunk.isLast());`

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTransformerTest.java
```diff
@@ -259,11 +258,11 @@ public void testTransformThrows()
...
 
         // Trying to read again returns the error again.
         chunk = transformer.read();
-        assertInstanceOf(Content.Chunk.Error.class, chunk);
+        assertTrue(Content.Chunk.isError(chunk));
```
lorban: `assertTrue(Content.Chunk.isError(chunk) && chunk.isLast());`

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTransformerTest.java
```diff
@@ -284,11 +283,11 @@ public void testTransformReturnsError()
         chunk.release();
 
         chunk = transformer.read();
-        assertInstanceOf(Content.Chunk.Error.class, chunk);
+        assertTrue(Content.Chunk.isError(chunk));
```
lorban: `assertTrue(Content.Chunk.isError(chunk) && chunk.isLast());`

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTransformerTest.java
```diff
@@ -284,11 +283,11 @@ public void testTransformReturnsError()
...
 
         // Trying to read again returns the error again.
         chunk = transformer.read();
-        assertInstanceOf(Content.Chunk.Error.class, chunk);
+        assertTrue(Content.Chunk.isError(chunk));
```
lorban: `assertTrue(Content.Chunk.isError(chunk) && chunk.isLast());`

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTransformerTest.java
```diff
@@ -306,11 +305,11 @@ public void testSourceReturnsError()
         source.fail(new IOException());
 
         chunk = transformer.read();
-        assertInstanceOf(Content.Chunk.Error.class, chunk);
+        assertTrue(Content.Chunk.isError(chunk));
```
lorban: `assertTrue(Content.Chunk.isError(chunk) && chunk.isLast());`

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTransformerTest.java
```diff
@@ -306,11 +305,11 @@ public void testSourceReturnsError()
...
 
         // Trying to read again returns the error again.
         chunk = transformer.read();
-        assertInstanceOf(Content.Chunk.Error.class, chunk);
+        assertTrue(Content.Chunk.isError(chunk));
```
lorban: `assertTrue(Content.Chunk.isError(chunk) && chunk.isLast());`

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTest.java
```diff
@@ -237,7 +241,7 @@ public void testReadFailReadReturnsError(Content.Source source) throws Exception
 
         // We must read the error.
         chunk = source.read();
-        assertInstanceOf(Content.Chunk.Error.class, chunk);
+        assertTrue(Content.Chunk.isError(chunk));
```
lorban: `assertTrue(Content.Chunk.isError(chunk) && chunk.isLast());`

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTest.java
```diff
@@ -259,7 +263,7 @@ public void testReadErrorDemandInvokesDemandCallback(Content.Source source) thro
         source.fail(new CancellationException());
 
         Content.Chunk chunk = source.read();
-        assertInstanceOf(Content.Chunk.Error.class, chunk);
+        assertTrue(Content.Chunk.isError(chunk));
```
lorban: `assertTrue(Content.Chunk.isError(chunk) && chunk.isLast());`

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ContentSourceTest.java
```diff
@@ -285,7 +289,7 @@ public void testDemandCallbackThrows(Content.Source source) throws Exception
         });
 
         chunk = source.read();
-        assertInstanceOf(Content.Chunk.Error.class, chunk);
+        assertTrue(Content.Chunk.isError(chunk));
```
lorban: `assertTrue(Content.Chunk.isError(chunk) && chunk.isLast());`

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/AsyncContentTest.java
```diff
@@ -101,7 +100,7 @@ public void testFailInvokesDemandCallback() throws Exception
 
             // We must read the error.
             chunk = async.read();
-            assertInstanceOf(Content.Chunk.Error.class, chunk);
+            assertTrue(Content.Chunk.isError(chunk));
```
lorban: `assertTrue(Content.Chunk.isError(chunk) && chunk.isLast());`

jetty-core/jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientAsyncContentTest.java
```diff
@@ -248,11 +247,11 @@ protected void service(Request request, org.eclipse.jetty.server.Response respon
             .onResponseContentSource((response, contentSource) -> response.abort(new Throwable()).whenComplete((failed, x) ->
             {
                 Content.Chunk chunk = contentSource.read();
-                assertInstanceOf(Content.Chunk.Error.class, chunk);
+                assertTrue(Content.Chunk.isError(chunk));
```
lorban: `assertTrue(Content.Chunk.isError(chunk) && chunk.isLast());`

jetty-core/jetty-client/src/test/java/org/eclipse/jetty/client/HttpClientAsyncContentTest.java
```diff
@@ -248,11 +247,11 @@ protected void service(Request request, org.eclipse.jetty.server.Response respon
...
                 contentSource.demand(() ->
                 {
                     Content.Chunk c = contentSource.read();
-                    assertInstanceOf(Content.Chunk.Error.class, c);
+                    assertTrue(Content.Chunk.isError(c));
```
lorban: `assertTrue(Content.Chunk.isError(chunk) && chunk.isLast());`

jetty-core/jetty-tests/jetty-test-client-transports/src/test/java/org/eclipse/jetty/test/client/transport/ServerTimeoutsTest.java
```diff
@@ -130,8 +130,8 @@ public boolean handle(Request request, Response response, Callback callback)
 
         // Reads should yield the idle timeout.
         Content.Chunk chunk = requestRef.get().read();
-        assertThat(chunk, instanceOf(Content.Chunk.Error.class));
-        Throwable cause = ((Content.Chunk.Error)chunk).getCause();
+        assertTrue(Content.Chunk.isError(chunk));
```
lorban: `assertTrue(Content.Chunk.isError(chunk) && chunk.isLast());`

lorban (changes requested):
After pondering this change, I'm now wondering if `isError()` shouldn't do `return chunk != null && chunk.getError() != null && chunk.isLast()`.

All the existing code that used to do `chunk instanceof Error` would then do the right thing both in case of a failure and a warning without any other change than `isError(chunk)`: failures would mark the end of the source, and warnings would be considered like a normal content with a zero-byte buffer; this would all be safer by default.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/ContentDocs.java
gregw: No.  The callback is completed and return from the method in lines 45-46.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: Every time I do that, @sbordet changes it back to `getSimpleName()`!   Why is it confusing?  If it is, then the extended class can always implement toString itself.

gregw:
> After pondering this change, I'm now wondering if `isError()` shouldn't do `return chunk != null && chunk.getError() != null && chunk.isLast()`.

That could be an `isFailure(Chunk)` method, but it would be confusing to have an `isError(Chunk)` that returns false when `getError()` is non-null.   May be it should be `getCause()` again?

All in all, I think all code needs to be revisited to make sure it can handle an non-last error/failure/warning, so I don't think this variation is necessary.

gregw:
@lorban The comment on the call last night regarding a Throwable that is an "error" but not a `java.lang.Error` has convinced me that just using "failure" is the better option.  I'll prepare a variant like that to see.

gregw:
@lorban in preparing the rename I found lots of javadoc that linked through to `java.lang.Error`, convincing me even more.

gregw:
@lorban I have pushed the rename back to `isFailure` and `getFailure`.

I think the distinction between "error", "failure" and "warning" would have been fine if all we were talking about was `Chunk`s, but we are in an environment with `java.lang.Error` and 'Callback.failed(Throwable)` and `XyzListener.onError(Throwable)`, so that any naming purity we established withing a chunk was lost the moment we used that with other code, where we immediately saw things like `callback.failed(chunk.getError())`.
We now still have similar things like `listener.onError(chunk.getFailure())`, but we are no longer trying to make a semantic difference based on "failure" vs "error" vs "warning",  so such code is less confusing.

So I'm now a definitely +1 on the just call it "failure" camp.      Hopefully I've done a reasonable job in the javadoc to explain transient vs persistent failures?    I have also added a `Chunk.isFailure(Chunk chunk, boolean last)` method so you can easily check on the type of failure if you want (many tests do).

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
lorban: I find it confusing in this particular case because it is in an anonymous class, so the logs will look like `Chunk$1@12345{...}`.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
```diff
@@ -33,7 +33,7 @@
 class AsyncContentProducer implements ContentProducer
 {
     private static final Logger LOG = LoggerFactory.getLogger(AsyncContentProducer.class);
-    private static final Content.Chunk.Error RECYCLED_ERROR_CHUNK = Content.Chunk.from(new StaticException("ContentProducer has been recycled"));
+    private static final Content.Chunk RECYCLED_ERROR_CHUNK = Content.Chunk.from(new StaticException("ContentProducer has been recycled"));
```
lorban: This chunk should be a `last` failure, shouldn't it?

jetty-core/jetty-tests/jetty-test-client-transports/src/test/java/org/eclipse/jetty/test/client/transport/HttpClientTest.java
```diff
@@ -983,7 +982,7 @@ public void testParallelContentSourceListenersPartialFailureInSpawnedThread(Tran
         assertThat(chunks3Latch.await(5, TimeUnit.SECONDS), is(true));
         assertThat(chunks3.stream().mapToInt(c -> c.getByteBuffer().remaining()).sum(), is(0));
         assertThat(chunks3.size(), is(1));
-        assertThat(chunks3.get(0), instanceOf(Content.Chunk.Error.class));
+        assertTrue(Content.Chunk.isFailure(chunks3.get(0)));
```
lorban: `assertTrue(Content.Chunk.isFailure(chunks3.get(0), true));` ?

jetty-core/jetty-tests/jetty-test-client-transports/src/test/java/org/eclipse/jetty/test/client/transport/HttpClientTest.java
```diff
@@ -941,7 +940,7 @@ public void testParallelContentSourceListenersPartialFailure(Transport transport
         assertThat(chunks2.stream().mapToInt(c -> c.getByteBuffer().remaining()).sum(), is(totalBytes));
         assertThat(chunks3.stream().mapToInt(c -> c.getByteBuffer().remaining()).sum(), is(0));
         assertThat(chunks3.size(), is(1));
-        assertThat(chunks3.get(0), instanceOf(Content.Chunk.Error.class));
+        assertTrue(Content.Chunk.isFailure(chunks3.get(0)));
```
lorban: `assertTrue(Content.Chunk.isFailure(chunks3.get(0), true));` ?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -327,18 +334,38 @@ default long getLength()
...
+         * and subsequent calls to {@link #read() read} may produce other content.</p>
+         * <p>A reader may treat a transient failure as persistent.</p>
+         *
+         * @param failure A failure.
+         * @param last true if the failure is persistent, false if the failure is transient.
+         * @see Chunk#getFailure()
+         */
+        default void fail(Throwable failure, boolean last)
+        {
+            fail(failure);
```
lorban, line 366: Since the `last` flag is ignored, this breaks the contract specified in the javadoc when last is false... Better throw `UnsupportedOperationException` when last == false for now?

lorban (changes requested):
Just a few small things left.

Overall, I'm happy with this change and you're right about the purity of error vs failure vs warning that doesn't make sense in the overall picture. I'm still convinced `isFailure(chunk, true)` is the right thing to do by default to check for errors, but we can re-review all that later and we don't generate transient failures for now so what we have now currently is safe. And I do agree that `isFailure(chunk)` should not check the last flag. The contract totally LGTM.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
gregw: It is by default. Should I made it explicit?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: I updated the contract to say that a source or its reader can treat a transient failure as permanent.    I.E. it is better to hard fail the source rather than the caller if you don't explicitly support transient.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/AsyncContentProducer.java
lorban: Probably not, but I got confused because `from(Throwable failure)` does not state that the last flag is set to true by default in the javadoc, while the implementation does the right thing.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -555,14 +582,50 @@ static Chunk asChunk(ByteBuffer byteBuffer, boolean last, Retainable retainable)
...
         /**
-         * <p>Creates an {@link Error error chunk} with the given failure.</p>
+         * <p>Creates an {@link Chunk#isFailure(Chunk) failure chunk} with the given failure.</p>
          *
          * @param failure the cause of the failure
-         * @return a new Error.Chunk
+         * @return a new {@link Chunk#isFailure(Chunk) failure chunk}
          */
-        static Error from(Throwable failure)
+        static Chunk from(Throwable failure)
```
lorban, line 591: The Javadoc should explicitly state that the last flag is set to true.

lorban (changes requested):
One last nit.

lorban (dismissed):

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: Currently this method is not used, but will be used for idle timeouts from the other PR?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
```diff
@@ -597,18 +666,57 @@ static Error from(Throwable failure)
...
+            return chunk != null && chunk.getFailure() != null;
+        }
+
+        /**
+         * @param chunk The chunk to test for an {@link Chunk#getFailure() failure}
+         * @param last The {@link Chunk#isLast() last} status to test for.
+         * @return True if the chunk is non-null and {@link Chunk#getFailure()} returns non-null
+         *         and {@link Chunk#isLast()} matches the passed status.
+         */
+        static boolean isFailure(Chunk chunk, boolean last)
```
sbordet, line 693: This method is never called with `last=false`?
Also, I don't think it is necessary, as if `isFailed(Chunk)==true` then you can safely call `chunk.isLast()` without NPE.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
sbordet: I would stick with fail(Throwable), `Throwable getFailure()` and `isFailed(Chunk)`.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/Content.java
gregw: It is used in lots of places, albeit mostly tests. No harm having it.

gregw:
@lorban can I get another review, I lost the green tick fixing some niggles from @sbordet's review

lorban (approved):

PR #9905, opened 2023-06-13
Jetty 12 idletimeout

gregw:
Rework idle timeout mechanism, undisable tests


gregw:
@lorban @lachlan-roberts The key changes in this PR are:
 + lots of simple mechanical changes to pass `TimeoutException` instead of `Throwable` when it is known to be a timeout.
 + `Request` has been updated to have two types of listeners: for failures and for timeouts.  Failures cannot be ignored, timeouts can be.
 + `HttpChannelState` is where the majority of the changes have been made and to invoke the various listeners.
 + The production of an error response now uses a second `ChannelResponse` instance.  This is to avoid the race where the original handler is still writing to the response whilst an error handler is trying to write the error page for an asynchronously delivered error.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
```diff
@@ -1202,6 +1203,23 @@ public void call(Invocable.Callable callable, Request request) throws Exception
             }
         }
 
+        public <T> boolean test(Predicate<T> predicate, T t, Request request)
```
lorban, line 1206: This looks unused. Should we keep it nevertheless?

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/ConnectorTimeoutTest.java
```diff
@@ -583,15 +583,25 @@ protected static class WaitHandler extends Handler.Abstract
...
+                Thread.sleep(MAX_IDLE_TIME * 3 / 2);
             }
             catch (Exception e)
             {
                 e.printStackTrace();
             }
+
+            // TODO what do we do about the failing write?
+            //      should timeout errors be non persistent?
+            Callback ocb = callback;
```
lorban, line 599: Should we keep this wrapping?

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/ServerTest.java
```diff
@@ -212,4 +217,142 @@ public boolean handle(Request request, Response response, Callback callback) thr
...
+                \r
+                """;
+        String rawResponse = _connector.getResponse(request);
+        HttpTester.Response response = HttpTester.parseResponse(rawResponse);
+        assertThat(response.getStatus(), is(HttpStatus.INTERNAL_SERVER_ERROR_500));
+        assertThat(response.getContent(), containsString("HTTP ERROR 500 java.util.concurrent.TimeoutException: Idle timeout expired:"));
+    }
+
+    @Test
+    public void testIdleTimeoutNoListenerHttpConfigurationOnly() throws Exception
```
lorban, line 255: I fail to see what this test asserts that the above one (`testIdleTimeoutNoListener()`) does not.

lorban (dismissed):
There are quite a few TODOs in the tests, but we can tackle them after the 12.0.0 release.

So LGTM.

jetty-core/jetty-fcgi/jetty-fcgi-client/src/main/java/org/eclipse/jetty/fcgi/client/transport/internal/HttpConnectionOverFCGI.java
```diff
@@ -217,7 +217,7 @@ private void shutdown()
     }
 
     @Override
-    public boolean onIdleExpired()
+    public boolean onIdleExpired(TimeoutException timeoutException)
     {
         long idleTimeout = getEndPoint().getIdleTimeout();
         TimeoutException failure = new TimeoutException("Idle timeout " + idleTimeout + " ms");
```
lachlan-roberts: I think we can avoid creating `failure` and just use `timeoutException`.

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ByteArrayEndPointTest.java
```diff
@@ -270,33 +265,6 @@ public void testIdle() throws Exception
...
-            fcb.get();
-            fail("Expected ExecutionException");
-        }
-        catch (ExecutionException t)
-        {
-            assertThat(t.getCause(), instanceOf(TimeoutException.class));
-        }
-        assertThat(NanoTime.millisSince(start), greaterThan(halfIdleTimeout));
-        assertThat("Endpoint open", endp.isOpen(), is(true));
+        assertFalse(endp.isOpen());
```
lachlan-roberts, line 268: Why is the rest of the test removed?

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/ServerTest.java
gregw: This one has the idle timeout configured in HttpConfiguration only, so it is testing that it is correctly applied by HttpChannel.  I.e the idle timeout between requests is 10x the idle timeout during requests.

gregw: (probably could have been parameterised... will do when I next touch this test.)

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/ConnectorTimeoutTest.java
gregw: long term no, but whilst we still have work to do....

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java
gregw: @lorban  It has been used before, but now that we don't have listening predicates.....  oh but wait yes we do.  This should be used by the context addIdleTimeoutListener wrapper that is not implemented!!!! stand by.....

gregw: @lorban I've added the context wrapper so idle timeouts are callback in the context.  I've also added a test to check that.  Please re-review... at least that last commit.

lorban (dismissed):

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -326,15 +346,41 @@ public Invocable.InvocationType getInvocationType()
         return Invocable.getInvocationType(onContent);
     }
 
-    public Runnable onFailure(Throwable x)
+    public Runnable onIdleTimeout(TimeoutException t)
```
lachlan-roberts, line 350: Use `@Override`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java
```diff
@@ -326,15 +346,41 @@ public Invocable.InvocationType getInvocationType()
...
+                    if (task != null)
+                        task.run();
+                }
+            };
+            return _serializedInvoker.offer(onIdle);
+        }
+        return onFailure(t); // TODO can we avoid double lock?
+    }
 
+    public Runnable onFailure(Throwable x)
```
lachlan-roberts, line 377: Use `@Override`.

lachlan-roberts (changes requested):

jetty-core/jetty-io/src/test/java/org/eclipse/jetty/io/ByteArrayEndPointTest.java
gregw: Because we removed the logic in AbstractEndPoint about only closing on idle if we were already half closed.

lorban (approved):

lachlan-roberts (approved):

PR #9911, opened 2023-06-15
 fixing JPMS and reactivating the tests

### LINKED ISSUE #9909: Jetty-12 ee9-annotations module does not work with JPMS
**Jetty version(s)**
12

**Description**
```
Error occurred during initialization of boot layer
java.lang.module.FindException: Module jakarta.cdi not found, required by jakarta.transaction
```

**How to reproduce?**
```
java -jar $JETTY_HOME/start.jar --add-modules=ee9-annotations
java -jar $JETTY_HOME/start.jar --jpms
```

gregw:
Fix #9909  JPMS dependencies and reinstate JPMS tests.

### DIFF
--- a/jetty-core/jetty-http2/jetty-http2-tests/pom.xml
+++ b/jetty-core/jetty-http2/jetty-http2-tests/pom.xml
@@ -12,6 +12,7 @@
 
   <properties>
     <maven.deploy.skip>true</maven.deploy.skip>
+    <h2spec.skip>${skipTests}</h2spec.skip>
   </properties>
 
   <build>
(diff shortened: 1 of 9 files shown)


gregw:
@olamy Can you take this one over? 
The issue appears to be that ee9-plus didn't have the required dependencies for jakarta-transactions (like ee10 does).   I've copied them over and simple command line test like in the issue now work.
But when I re-enabled the demo distribution tests, I still have problems when JPMS is enabled.

Also, I'm not entirely sure what dependencies I should be using, as ee10 and ee9 use the same transactions API version, but appear to have different versions for it's dependencies.   Not sure how to work out exactly which ones are needed?   Also then there is ee8, which I've not yet looked at.

Finally, there are still version properties in the top level pom that are different yet again.  Do we need them in the top level at all?

This test also revealed some other unrelated issues in the demos, so I will run down those rabbit holes if you can take this one over?

olamy (approved):

PR #9912, opened 2023-06-15
Jetty 12 distro test for static content features.

gregw:
Added a test for static content features.


gregw:
@lorban I added a distro test for static content and it is failing badly:

 + Some gzip error for immutable headers (I'll look at this one)
 + favico.ico is not served
 + jetty-dir.css is not served

Can you look at this (other than gzip issue) and perhaps add some more tests to check the distro can really serve static content.

lorban:
@gregw I committed a fix for `jetty-dir.css`, which is now properly served. 

As for `favico.ico`, I just checked that Jetty 11 behaves exactly like Jetty 12:

 - it is served when only the `http` module is added.
 - it is not served when the `demo` (or `ee9-demos`) module is added.

gregw:
@lorban can you explain why the `favicon.ico` behaviour is like that?   Surely it should be served by the root context?

gregw:
Ah it is served by our ErrorHandler when there is no root context.    Hmmm I think we need to add a favicon.ico to our root context and make it the j:// one.  I'll do that.

gregw:
@lorban are you working on this?

lorban:
@gregw now I am

lorban:
Obsoleted by #9927 and #9952

PR #9915, opened 2023-06-15
Issue #9396 - fixes to resolve WebSocket JPMS warnings

### LINKED ISSUE #9396: Improve JPMS testing for websocket in Jetty 12
I added `--jpms` to some of the websocket tests in `DistributionTests` and got some failures. These should be investigated and extra tests should be added to confirm that this works.

Also add a test for the `JakartaWebSocketShutdownContainer` which is an internal class and not in a package exported with JPMS but is exposed for the service loader, and does a `protectAndExpose` in the `JakartaWebSocketConfiguration` class.

lachlan-roberts:
## Issue #9396

- fix some JPMS warnings for websocket
- add testing for websocket with JPMS in `DistributionTests`
- Make `WebSocketShutdownContainer`s publicly exported in module-info files.
- Optional dependencies for `WebSocketShutdownContainer` and `JettyWebSocketClientConfiguration` should be `requires static transitive` as they are in public signatures but also optional.


lachlan-roberts:
I don't think the `requires static transitive` works. I'm getting the following error when running tests for `jetty-ee9-websocket-client`

> Corrupted channel by directly writing to native stream in forked JVM 1. Stream 'java.lang.module.ResolutionException: Module org.eclipse.jetty.ee9.websocket.jetty.client does not read a module that exports org.eclipse.jetty.ee9.webapp'.

The issue is that we must export the package of `JettyWebSocketClientConfiguration` as we use it with the service provider, and because this is publicly exposed we need the `requires transitive`. However we also want to make it an optional dependency so we want the `requires static`. And so the addition of the `static` modifier is what is causing this error.

gregw:
@janbartel help

gregw:
i think this could just be a surefire issue.  I might needs something like:
```xml
      <plugin>
        <artifactId>maven-surefire-plugin</artifactId>
        <configuration>
          <argLine>
            @{argLine} ${jetty.surefire.argLine}
            --add-modules org.eclipse.jetty.ee9.webapp
            --add-reads org.eclipse.jetty.ee9.websocket.jetty.client=org.eclipse.jetty.ee9.webapp
          </argLine>
        </configuration>
      </plugin>
```

gregw:
@lachlan-roberts that fixed it, so I've pushed to this branch.  
@janbartel do you think that looks right?

jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jetty-client/pom.xml
```diff
@@ -16,6 +16,16 @@
 
   <build>
     <plugins>
+      <plugin>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <argLine>
+            @{argLine} ${jetty.surefire.argLine}
+            --add-modules org.eclipse.jetty.ee9.webapp
+            --add-reads org.eclipse.jetty.ee9.websocket.jetty.client=org.eclipse.jetty.ee9.webapp
```
janbartel: You do not need the `read` line. That is only used where we cannot declare an outright `requires` in the `module-info.java`, eg for test-only dependencies. Your `module-info.java` has declared a `requires static` and that _must_ be sufficient and should _not_ need to be forcibly overridden in this way.

jetty-ee8/jetty-ee8-websocket/jetty-ee8-websocket-jetty-client/pom.xml
```diff
@@ -32,6 +32,16 @@
           </instructions>
         </configuration>
       </plugin>
+      <plugin>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <argLine>
+            @{argLine} ${jetty.surefire.argLine}
+            --add-modules org.eclipse.jetty.ee8.webapp
+            --add-reads org.eclipse.jetty.ee8.websocket.jetty.client=org.eclipse.jetty.ee8.webapp
```
janbartel: You do not need the `read` line. That is only used where we cannot declare an outright `requires` in the `module-info.java`, eg for test-only dependencies. Your `module-info.java` has declared a `requires static` and that _must_ be sufficient and should _not_ need to be forcibly overridden in this way.

jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jetty-client/src/main/java/module-info.java
```diff
@@ -11,11 +11,15 @@
...
 //
 
+import org.eclipse.jetty.ee10.websocket.client.config.JettyWebSocketClientConfiguration;
+
 module org.eclipse.jetty.ee10.websocket.jetty.client
 {
     requires org.eclipse.jetty.websocket.client;
 
-    requires static org.eclipse.jetty.ee10.webapp;
+    requires static transitive org.eclipse.jetty.ee10.webapp;
```
janbartel: I don't see the same surefire config in the pom.xml as for ee8 and ee9, and indeed I can't see any unit tests, where are they?

janbartel (changes requested):

jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jetty-client/pom.xml
lachlan-roberts: We don't declare an outright `requires` in the `module-info.java` we use `requires static` because webapp is optional for the client.

I think there would be no issue if we just had  `requires transitive org.eclipse.jetty.ee10.webapp;` in the module info, but we use `requires static transitive`.

jetty-ee10/jetty-ee10-websocket/jetty-ee10-websocket-jetty-client/src/main/java/module-info.java
lachlan-roberts: This is needed for the `JakartaWebSocketShutdownContainer` which is really a fallback if the Jetty `JakartaWebSocketClientContainer` is used from a different servlet container. Otherwise if in Jetty it should just register as a bean on the `ContextHandler`.

Because of this it is hard to test in a unit test.

jetty-ee9/jetty-ee9-websocket/jetty-ee9-websocket-jetty-client/pom.xml
janbartel: @lachlan-roberts you do not need and should not have the `reads` line in pom.xml. 

The `transitive` in `module-info` is not necessary either - you do not expose any methods in your module that expose any webapp classes. If in fact you delete `transitive`,  the distribution tests work fine.

The question is: do you need the `add-modules=org.eclipse.jetty.ee9.webapp` in your `pom.xml` when your `module-info` has declared it to be `requires static`. Normally, the `requires static` would mean that you don't need it to be present. _However_ because you have declared a `provides` of a class that depends upon classes in `org.eclipse.jetty.ee9.webapp` I believe that `jpms` insists that that module is always present. In other words, even if you are just deploying the client outside of jetty, with no webapps, you will always have to deploy the `org.eclipse.jetty.ee9.webapp` module.

I think the only way to solve that (ie to make it truly optional to provide `org.eclipse.jetty.ee9.webapp`) would be to make a separate `jetty-ee9-websocket-jetty-client-webapp` module that declares `requires` dependencies on both `org.eclipse.jetty.ee9.webapp` and `org.eclipse.jetty.ee9.websocket.client`  and `provides` the configuration class. That way, users who are deploying the client and don't care about webapps will deploy `org.eclipse.jetty.ee9.websocket.client`, but those that are deploying in jetty will deploy `org.eclipse.jetty.websocket.client.webapp`.

janbartel (changes requested):

sbordet (dismissed):

gregw (changes requested):
wot @janbartel said!

I think that was my original mistake and I went from "Try A" to "Try A and B" without actually doing "Try B".

lachlan-roberts:
These `--add-modules` and `--add-reads` are removed by #9981.
I will merge these two together now simone is happy with other changes in this PR.

gregw (dismissed):
Look OK to me

janbartel (dismissed):

gregw:
@jmcc0nn3ll @joakime This has not actually been built in CI?

jmcc0nn3ll:
> @jmcc0nn3ll @joakime This has not actually been built in CI?

This needs to push a notification to the build servers, I updated DNS but perhaps there's something else that we need to do to fix that connection.

gregw (approved):

PR #9920, opened 2023-06-16
Remove ee10 HttpChannel.Listener

lorban:
It's only used in tests, so I replaced it with other existing listener mechanisms.


gregw (changes requested):
This looks good... except for checkstyle issue I think?

gregw (approved):

Issue #9910, opened 2023-06-15
Inconsistent handling of welcome files between Jetty 10 and 12


grgrzybek:
**Jetty version(s)**
12.0.0.beta1

**Java version/vendor** `(use: java -version)`
```
$ java -version
java version "11.0.19" 2023-04-18 LTS
Java(TM) SE Runtime Environment 18.9 (build 11.0.19+9-LTS-224)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.19+9-LTS-224, mixed mode)
```

**OS type/version**
```
$ uname -srv
Linux 6.3.6-200.fc38.x86_64 #1 SMP PREEMPT_DYNAMIC Mon Jun  5 15:45:04 UTC 2023
```

**Description**
When working on Pax Web welcome file handling, I found some differences between Jetty 10 and Jetty 12.

**How to reproduce?**

I have a quite complicated setup related to welcome file handling. For Pax Web 9 / Jetty 10 it is: 
https://github.com/ops4j/org.ops4j.pax.web/blob/web-9.0.9/pax-web-jetty/src/test/java/org/ops4j/pax/web/service/jetty/internal/UnifiedJettyTest.java#L659

Pax Web provides a copy of Jetty's DefaultServlet just to unprivate some methods and uses its extension called `org.ops4j.pax.web.service.jetty.internal.web.JettyResourceServlet` which is more aware of OSGi resource handling.

In this setup (`paxWebWelcomePagesWithDifferentContext()` test) I have:
* one `org.eclipse.jetty.ee10.servlet.ServletContextHandler` with `/c` context path and `{ "index.y", "index.x" }` welcome files
* 5 servlets:
  * `org.eclipse.jetty.ee10.servlet.DefaultServlet` mapped to `/` (redirectWelcome=false, welcomeServlets=true, pathInfoOnly=false, baseResource="target/b1")
  * `org.eclipse.jetty.ee10.servlet.DefaultServlet` mapped to `/r/*` (redirectWelcome=false, welcomeServlets=true, pathInfoOnly=true, baseResource="target/b2")
  * `org.eclipse.jetty.ee10.servlet.DefaultServlet` mapped to `/s/*` (redirectWelcome=true, welcomeServlets=true, pathInfoOnly=true baseResource="target/b3")
  * servlet mapped to `{ "*.x", "*.y" }` to print out values of attributes like `jakarta.servlet.forward.request_uri`
[...]

PR #9921, opened 2023-06-16
Fix ee10 path info only

lorban:
Minimalistic fix for #9910

### DIFF
--- a/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
+++ b/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
@@ -600,6 +600,11 @@ public class DefaultServlet extends HttpServlet
                 _uri = Request.newHttpURIFrom(getWrapped(), URIUtil.encodePath(URIUtil.addPaths(_servletRequest.getServletPath(), _servletRequest.getPathInfo())));
         }
 
+        public HttpServletRequest getServletRequest()
+        {
+            return _servletRequest;
+        }
+
         @Override
         public HttpFields getHeaders()
         {
@@ -1042,7 +1047,14 @@ public class DefaultServlet extends HttpServlet
                     // precedence over resources served by Servlets.
                     Resource welcomePath = base.resolve(welcome);
                     if (Resources.isReadableFile(welcomePath))
+                    {
+                        if (isPathInfoOnly() && !isIncluded(request))
+                        {
+                            String servletPath = getServletRequest(coreRequest).getServletPath();
+                            welcomeInContext = URIUtil.addPaths(servletPath, welcomeInContext);
+                        }
                         return welcomeInContext;
+                    }
 
                     // Check whether a Servlet may serve the welcome resource.
                     if (_welcomeServletMode != WelcomeServletMode.NONE && welcomeTarget == null)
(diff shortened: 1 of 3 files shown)


gregw (changes requested):
This looks about right.... but the test failures are related.   It could be bad test?

lorban:
@gregw I don't think it's a bad test. Schematically, this is what it fails on:

 - `context.setWelcomeFiles(new String[]{"index.html", "index.jsp", "index.htm"})`
 - `GET /context/alt/dir/index.html | assert (status == 200)` (passes)
 - `GET /context/alt/dir/ | assert (status == 200)` (fails)

For the 2nd GET, we have:

`encodedPathInContext == /dir/`
`getWrapped().getHttpURI() == http://local/context/alt/dir/`
`Request.newHttpURIFrom(getWrapped(), encodedPathInContext) == http://local/context/dir/`

So the fix for `testPathInfoOnly` breaks `testWelcomeMultipleDefaultServletsDifferentBases`.

gregw:
@lorban Why is `Request.newHttpURIFrom(getWrapped(), encodedPathInContext) == http://local/context/dir/` rather than `http://local/context/alt/dir/`?    If `isPathInfoOnly` is false, then the encodedPathInContext should be `/alt/dir/`,  if it is true, then `/dir/` must resolve to a resource in the alternative base.

lorban:
@gregw `pathInfoOnly` is set to true, so... could it be a bad test?

janbartel:
The `DefaultServletTest.testWelcomeMultipleDefaultServletsDifferentBases` test that is failing seems like a good test to me. The request sends `http://local/context/other`, which should be matched to the directory `docroot/altroot/dir` because the `pathInfo` on the request is `""`.  As that is a directory, we should send a redirect to the welcome of `http://local/context/other/`, not `http://local/context/`.

lorban:
@janbartel thanks for the clarification.

I've modified the redirect handling as well as the serving of static files to account for pathInfoOnly being true.

janbartel:
A few things I don't understand:
* why in `ResourceService.doGet()` are we passing the `pathInContext` as a separate parameter to `sendWelcome()`, considering that the `Request` is now the `ServletCoreRequest` with its path already modified?
* why do we throw away the `Content` that we found in `DefaultServlet.doGet()` inside `ResourceService.sendWelcome()` when it calls `welcome()`?

It seems to me that we want to use the `pathInfoOnly` to help us select the resource that we want to serve, but if we are going to redirect or send it off to some other servlet to handle then we need to keep the full original path info? @gregw?_

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -593,13 +593,18 @@ private static class ServletCoreRequest extends Request.Wrapper
             String includedServletPath = (String)request.getAttribute(RequestDispatcher.INCLUDE_SERVLET_PATH);
             boolean included = includedServletPath != null;
             if (request.getDispatcherType() == DispatcherType.REQUEST)
-                _uri = getWrapped().getHttpURI();
+                _uri = Request.newHttpURIFrom(getWrapped(), encodedPathInContext);
```
gregw, line 596: I'm now cautious about us rewriting the URI.  We should just reconstruct it from servletPath and pathInfo, but not change it for pathInfoOnly.

gregw (changes requested):

lorban:
Superseded by https://github.com/eclipse/jetty.project/pull/9934

PR #9923, opened 2023-06-19
Issue #9919 - Avoid creating two ByteBufferPools for the server

### LINKED ISSUE #9919: Jetty-12 creates two instances of ArrayByteBufferPool
**Jetty version(s)**
12

**How to reproduce?**
```
java -jar $JETTY_HOME/start.jar --add-modules=server
java -jar $JETTY_HOME/start.jar jetty.server.dumpAfterStart=true
```

Then in the dump you can see two different instances of  `ArrayByteBufferPool`.
Unless there is a specific reason why we need 2 I think we should be sharing one for the entire server.

```
...
|     +> qtp92150540-25 TIMED_WAITING tid=25 prio=5 IDLE
|     +> qtp92150540-31 TIMED_WAITING tid=31 prio=5 IDLE
+= oejut.ScheduledExecutorScheduler@3444d69d{STARTED} - STARTED
+- org.eclipse.jetty.io.ArrayByteBufferPool@3911c2a7{min=0,max=65536,buckets=16,heap=0/521142272,direct=0/521142272}
|  +> direct size=16
|  |  +> org.eclipse.jetty.io.ArrayByteBufferPool$RetainedBucket@15c43bd9{capacity=4096,inuse=0(0%)}
|  |  +> org.eclipse.jetty.io.ArrayByteBufferPool$RetainedBucket@3d74bf60{capacity=8192,inuse=0(0%)}
|  |  +> org.eclipse.jetty.io.ArrayByteBufferPool$RetainedBucket@4f209819{capacity=12288,inuse=0(0%)}
...
+~ org.eclipse.jetty.util.resource.FileSystemPool@16ec5519
+- org.eclipse.jetty.io.ArrayByteBufferPool@2f7298b{min=0,max=65536,buckets=16,heap=0/521142272,direct=0/521142272}
|  +> direct size=16
|  |  +> org.eclipse.jetty.io.ArrayByteBufferPool$RetainedBucket@6a192cfe{capacity=4096,inuse=0(0%)}
|  |  +> org.eclipse.jetty.io.ArrayByteBufferPool$RetainedBucket@5119fb47{capacity=8192,inuse=0(0%)}
|  |  +> org.eclipse.jetty.io.ArrayByteBufferPool$RetainedBucket@7193666c{capacity=12288,inuse=0(0%)}
|  |  +> org.eclipse.jetty.io.ArrayByteBufferPool$RetainedBucket@20deea7f{capacity=16384,inuse=0(0%)}
|  |  +> org.eclipse.jetty.io.ArrayByteBufferPool$RetainedBucket@3835c46{capacity=20480,inuse=0(0%)}
|  |  +> org.eclipse.jetty.io.ArrayByteBufferPool$RetainedBucket@1dde4cb2{capacity=24576,inuse=0(0%)}
...
```

lachlan-roberts:
## Closes #9919

Ensure we only create one `ByteBufferPool` and `ScheduledExecutorScheduler` instance in `jetty.xml`.

### DIFF
--- a/jetty-core/jetty-server/src/main/config/etc/jetty.xml
+++ b/jetty-core/jetty-server/src/main/config/etc/jetty.xml
@@ -24,24 +24,15 @@
 <!-- configuration that may be set here.                             -->
 <!-- =============================================================== -->
 <Configure id="Server" class="org.eclipse.jetty.server.Server">
-    <Arg name="threadPool"><Ref refid="threadPool"/></Arg>
-
-    <Call name="addBean">
-      <Arg><Ref refid="byteBufferPool"/></Arg>
-    </Call>
-
-    <!-- =========================================================== -->
-    <!-- Add shared Scheduler instance                               -->
-    <!-- =========================================================== -->
-    <Call name="addBean">
-      <Arg>
-        <New class="org.eclipse.jetty.util.thread.ScheduledExecutorScheduler">
-          <Arg name="name"><Property name="jetty.scheduler.name"/></Arg>
-          <Arg name="daemon" type="boolean"><Property name="jetty.scheduler.daemon" default="false" /></Arg>
-          <Arg name="threads" type="int"><Property name="jetty.scheduler.threads" default="-1" /></Arg>
-        </New>
-      </Arg>
-    </Call>
+  <Arg name="threadPool"><Ref refid="threadPool"/></Arg>
+  <Arg>
+    <New class="org.eclipse.jetty.util.thread.ScheduledExecutorScheduler">
+      <Arg name="name"><Property name="jetty.scheduler.name"/></Arg>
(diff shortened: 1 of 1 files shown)


gregw (changes requested):
I don't think this is the right fix, as it just avoids the problem.
If the server has been configured with a scheduler or buffer pool via a bean, then it should not create another one.

We've never needed this before, so not sure why it is need now.


Note also that `<Configuration>` args are fragile, as if another XML is added to the configuration before this one, then it will create the Server, perhaps with different arguments.   I never really liked it when the ThreadPool was added as a final field, so it is needed as an arg rather than a setter.  I don't think we should triple down on this.

gregw (approved):

gregw:
@lachlan-roberts I still don't like it, but I've changed my review.    The items are now all final fields, so not much else we can do other than constructor injection.

PR #5735, opened 2020-11-27
Jetty 9.4.x optimize parser warning from logs and add spotbug

olamy:
current logs parser are reading multiple time log files so this generate extra load on jenkins server.
this generate aggregated report per jdk as well (more readable)
Screen Shot 2020-11-27 at 6 57 47 pm

PR #9929, opened 2023-06-19
Fix spotbugs packages

zUniQueX:
Spotbugs was disabled for JDK 17+ some time ago in 04acdb72f041a15267bb79c36d8fc4d517beaadb. After investigating, if the new Spotbugs releases allow to re-enable the check in the project, I've seen that some packages in the `<spotbugs.onlyAnalyze>` tags are wrong. Those tags were introduced at the [migration from findbugs](https://github.com/eclipse/jetty.project/pull/5735). I've corrected the packages and made adjustments according to the package splitting of the different EE versions.


olamy:
Good catch.
Please note we use the profile `ci` in Jenkins to record analysis.
But you're right it's time to remove the profile `jdk17`.
I modified the PR to remove this limitation

olamy (approved):

PR #9932, opened 2023-06-20
Issue #9476 - Fix 9.4.51 NullPointerException

### LINKED ISSUE #9476: onCompleteFailure called multiple times
**Jetty version(s)**
Jetty 9.4.51

java version 1.8.0_161

**OS type/version**
CentOS

**Description**

When I upgrade to Jetty 9.4.51, this exception will appear in the log every day.
What caused this anomaly?

```
04:04:21.744 WARN  [qtp903644315-163107] [o.e.j.i.AbstractConnection](AbstractConnection.java:96)
java.lang.NullPointerException: null
        at org.eclipse.jetty.io.AbstractConnection.lambda$failedCallback$0(AbstractConnection.java:92)
        at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:883)
        at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1034)
        at java.lang.Thread.run(Thread.java:748)
```

sid-evolphin:
#9476 
Added a null check for the Callback parameter in o.e.j.iAbstractConnection.failedCallback(Callback, Throwable) so that the subsequent code-flow is not aborted; while the root cause may be ascertained in case debug logging is enabled.

Further, in the case of a RejectedExecutionException when submitting the failCallback, the Runnable is run instead of directly invoking the Callback, so that exceptions in the Callback do not abort the synchronous code-flow.

### DIFF
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/AbstractConnection.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/AbstractConnection.java
@@ -85,14 +85,6 @@ public abstract class AbstractConnection implements Connection
 
     protected void failedCallback(final Callback callback, final Throwable x)
     {
-        if (callback == null)
-        {
-            if (LOG.isDebugEnabled())
-            {
-                LOG.warn("Callback null", x);
-            }
-            return;
-        }
         Runnable failCallback = () ->
         {
             try
--- a/jetty-io/src/main/java/org/eclipse/jetty/io/AbstractConnection.java
+++ b/jetty-io/src/main/java/org/eclipse/jetty/io/AbstractConnection.java
@@ -85,6 +85,12 @@ public abstract class AbstractConnection implements Connection
 
     protected void failedCallback(final Callback callback, final Throwable x)
     {
+        if (callback == null) {
+            if (LOG.isDebugEnabled()) {
+                LOG.warn("Callback null", x);
+            }
+            return;
+        }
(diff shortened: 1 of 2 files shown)


gregw (changes requested):
I don't think this is the right solution.   It is ignoring a null callback that should never have been created/executed.

Instead, I think the fix should be in `onCompleteSuccess` and `onCompleteFailure` on `org.eclipse.jetty.server.HttpConnection.SendCallback`.    Both of these methods should call `release()` and if they get null back, then do no more.

I can see other problems there (those methods are not truly atomic), but we are not going to entirely fix jetty-9.4.

sid-evolphin:
Thank you very much @gregw for the review and comments.

- I have incorporated your suggested changes to the methods in `o.e.j.s.HttpConnection.SendCallback` that use `release()` which may potentially return a `null` `Callback`.
    - They also now have try-finally blocks to ensure that `shutdownOutput()` does get invoked even if there is an exception in the callback.
- I have reverted the null check in `o.e.j.i.AbstractConnection.failedCallback(Callback, Throwable)` to maintain the method parameter contract of `@NonNull`.

Please re-review this PR. Thank you very much.

gregw:
Replaced by #9935

sid-evolphin:
> Replaced by #9935

Thanks a lot @gregw . I'm glad you are fixing the root cause.

Issue #9944, opened 2023-06-21
Remove integer for demand in websocket in Jetty-12


lachlan-roberts:
**Enhancement Description**

We use the signature `void demand(long n);` for demand in websocket, however we only ever use or have a usecase for `demand(1)`. So we should remove the long field and just have a `demand()` interface.

PR #9945, opened 2023-06-22
Issue #9944 - remove integer from demand in websocket core

lachlan-roberts:
Remove the `long` argument from `demand()` in websocket core.

This had already been done for the new core Jetty WebSocket API.


jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/WebSocketConnection.java
```diff
@@ -343,21 +343,18 @@ public void run()
...
         {
             if (LOG.isDebugEnabled())
-                LOG.debug("demand {} d={} fp={} {} {}", n, demand, fillingAndParsing, networkBuffer, this);
+                LOG.debug("demand {} d={} fp={} {}", demand, fillingAndParsing, networkBuffer, this);
 
             if (demand < 0)
                 return;
 
-            demand = MathUtils.cappedAdd(demand, n);
+            demand = MathUtils.cappedAdd(demand, 1);
```
gregw, line 357: I think with our other usages of no-args demand, then a second call to demand is either an ISE or ignored.    This looks like internally we are still keeping a long counter of demand?   Is there a reason for this?   Could this just be a boolean?

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/util/DemandingFlusher.java
```diff
@@ -76,21 +75,21 @@ public DemandingFlusher(IncomingFrames emitFrame)
     protected abstract boolean handle(Frame frame, Callback callback, boolean first);
 
     @Override
-    public void demand(long n)
+    public void demand()
     {
-        _demand.getAndUpdate(d -> Math.addExact(d, n));
+        _demand.incrementAndGet();
```
gregw, line 80: again this is implemented as a long rather than a boolean.   Reason?

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/CoreSession.java
```diff
@@ -152,14 +152,13 @@ public interface CoreSession extends OutgoingFrames, IncomingFrames, Configurati
     void abort();
 
     /**
-     * <p>Manages flow control by indicating demand for WebSocket frames.</p>
+     * <p>Manages flow control by indicating demand for a WebSocket frame.</p>
```
gregw, line 156: The behaviour for calling `demand` when there is already pending `demand` needs to be documented.   You are currently internally tracking a long count, so you have effectively kept the long and could reproduce with:
```java
   public void demand(long n)
   {
       for (long i = 0; i < n; i++)
           demand();
   }
```

I'm not sure this is optimal.   Could we throw ISE if there is already pending demand?

gregw (changes requested):

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/WebSocketConnection.java
lachlan-roberts: Yeah I agree with all these comments.

I think maybe it would be best to get this API change merged in case we want to do a beta3 release soon, and work on the internal changes to not allow multiple demand in another PR. Unless we are delaying beta3 until next week  and then I can work on it monday.

gregw: OK, add a TODO and/or open an issue and I'll approve.

gregw: @lachlan-roberts Note I think beta3 will be Monday... so perhaps you can do it by then, but no need to rush the impl... so long as that is the intent.
(also the javadoc needs to describe the intended contract regardless of implementations)

lachlan-roberts: Opened Issue https://github.com/eclipse/jetty.project/issues/9965

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/CoreSession.java
gregw: I still think you need to javadoc in this PR what is the expected behaviour is of a demand call with pending demand... even if the actual behaviour is implemented in a different PR.

gregw (dismissed):

jetty-core/jetty-websocket/jetty-websocket-core-common/src/main/java/org/eclipse/jetty/websocket/core/CoreSession.java
lachlan-roberts: Should it be ReadPendingException or IllegalStateException?

sbordet: `ReadPendingException` extends `IllegalStateException` so either works.

jetty-core/jetty-websocket/jetty-websocket-core-tests/src/test/java/org/eclipse/jetty/websocket/core/server/WebSocketServerTest.java
```diff
@@ -110,7 +110,8 @@ protected void demand()
             Frame frame = serverHandler.receivedFrames.poll(250, TimeUnit.MILLISECONDS);
             assertNull(frame);
 
-            serverHandler.getCoreSession().demand(2);
+            serverHandler.getCoreSession().demand();
+            serverHandler.getCoreSession().demand();
```
sbordet, line 114: This seems wrong, as the second call should throw.
If it's not causing flakyness, needs a comment explaining why.

jetty-core/jetty-websocket/jetty-websocket-core-tests/src/test/java/org/eclipse/jetty/websocket/core/server/WebSocketServerTest.java
```diff
@@ -248,7 +249,9 @@ public void testTcpCloseNoDemand() throws Exception
             public void onOpen(CoreSession coreSession, Callback callback)
             {
                 super.onOpen(coreSession, callback);
-                coreSession.demand(3);
+                coreSession.demand();
+                coreSession.demand();
+                coreSession.demand();
```
sbordet, line 254: Ditto.

jetty-core/jetty-websocket/jetty-websocket-core-tests/src/test/java/org/eclipse/jetty/websocket/core/server/WebSocketServerTest.java
```diff
@@ -307,7 +310,8 @@ public void onOpen(CoreSession coreSession, Callback callback)
             {
                 super.onOpen(coreSession);
                 callback.succeeded();
-                coreSession.demand(2);
+                coreSession.demand();
+                coreSession.demand();
```
sbordet, line 314: Ditto.

jetty-core/jetty-websocket/jetty-websocket-core-tests/src/test/java/org/eclipse/jetty/websocket/core/server/WebSocketServerTest.java
```diff
@@ -375,7 +379,8 @@ public void onOpen(CoreSession coreSession, Callback callback)
...
```
sbordet, line 383: Ditto.

sbordet (changes requested):

jetty-core/jetty-websocket/jetty-websocket-core-tests/src/test/java/org/eclipse/jetty/websocket/core/server/WebSocketServerTest.java
lachlan-roberts: See comments from https://github.com/eclipse/jetty.project/pull/9945#discussion_r1241862275

I have opened https://github.com/eclipse/jetty.project/issues/9965 to change the implementation.

lachlan-roberts: In the next PR these should fail and I will need to change them.

Hopefully that will make beta3 as well, but if not at least we have the API changed with this.

sbordet (approved):

PR #9948, opened 2023-06-22
#9946 Stop passing Handler in constructor as a parent

### LINKED ISSUE #9946: Handler passed to Handler in constructor a parent or child?
**Jetty version(s)**
12

**Description**

In `Handler.Wrapper` we can now pass a handler in a [constructor](https://github.com/eclipse/jetty.project/blob/jetty-12.0.0.beta2/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Handler.java#L582-L585) that is the child handler.
However, we still have (deprecated) constructors on [some handlers](https://github.com/eclipse/jetty.project/blob/jetty-12.0.0.beta2/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java#L184-L189), that take a passed handler as the parent.

We should do one or the other, as otherwise it is confusing. Probably keep the former as it is not deprecated, and  so we should remove the later and uses of it (in XML as well?).

lorban:
Fixes #9946.


lorban:
@gregw I may have found a bug in `StopTest.testWriteDuringShutdown()`: during graceful shutdown, responses do not have the `Connection: close` header but this test was written assuming the header should be there.

gregw (changes requested):
Wow, amazing how often we passed the parent in the constructor and still called setHandler!

Is the StopTest issue related?  Could you move that out to another PR?  From my understanding the connection: close header should be there whilst the connector is being shutdown, but not whilst the webapp is being shutdown.

lorban:
@gregw No, the StopTest issue isn't related. It just happened that StopTest was disabled but using a ctor passing a handler as a parent, so I updated & re-enabled it and only `testWriteDuringShutdown` fails.

I'll keep it disabled in this PR but open an issue to review it once this gets merged.

lorban:
Here: #9949

gregw (dismissed):

gregw (approved):

PR #9952, opened 2023-06-23
Fix flaky DistributionTests.testIniSectionPropertyOverriddenByCommandLine

lorban:
`DistributionTests` tests bind ports that are hardcoded in the distribution's config files, so they cannot run in parallel otherwise they risk failing to bind the ports.

This is why they've been flaky recently.

### DIFF
--- a/tests/test-distribution/test-distribution-common/pom.xml
+++ b/tests/test-distribution/test-distribution-common/pom.xml
@@ -11,8 +11,7 @@
   <packaging>jar</packaging>
 
   <properties>
-    <!-- DistributionTests tests cannot run in parallel as they bind ports hardcoded in the distribution's config files -->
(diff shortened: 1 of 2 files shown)


olamy:
the distribution tests are using random http/https port.
What sort is failing because of this? do you have any build logs links?

lorban:
@olamy I saw many DistributionTests failures as part of https://github.com/eclipse/jetty.project/pull/9912. I added some extra info on the failure (the extra log in `testIniSectionPropertyOverriddenByCommandLine()`) and saw a bind error for port 8443.

Of course, since I closed the PR the Jenkins logs have been deleted.

olamy:
well in this case when `ssl` module is activated we simply need to use a random free port for `jetty.ssl.port` as we do for `jetty.http.port`
I can add that on Monday.

lorban:
@olamy I did it myself, please re-review.

gregw (approved):

PR #9958, opened 2023-06-23
Update plexus-utils to 4.0.0

janbartel:
Updates to plexus-utils 4.0.0 that was proposed by dependabot, but too complex for it to do.

### DIFF
--- a/jetty-maven-plugin/pom.xml
+++ b/jetty-maven-plugin/pom.xml
@@ -161,6 +161,10 @@
       <artifactId>plexus-utils</artifactId>
     </dependency>
     <dependency>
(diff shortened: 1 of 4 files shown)


jetty-maven-plugin/pom.xml
```diff
@@ -160,6 +160,10 @@
       <groupId>org.codehaus.plexus</groupId>
       <artifactId>plexus-utils</artifactId>
     </dependency>
+    <dependency>
```
olamy, line 163: test-distribution needs it as well.

olamy (changes requested):

olamy (approved):

Issue #265, opened 2016-02-16
--list-config license enhancement


jmcc0nn3ll:
migrated from Bugzilla [#483677](https://bugs.eclipse.org/bugs/show_bug.cgi?id=483677)
status NEW severity _enhancement_ in component _start_ for _9.3.x_
Reported in version _9.3.6_ on platform _PC_
Assigned to: Joakim Erdfelt

On 2015-12-04 09:22:49 -0500, Jesse McConnell wrote:

> would be a nice feature to be able to indicate the license for each jar listed in the --list-config command

joakime:
@gregw this sounds like a good task for lachlan, wdyt?

gregw:
@joakime I think the start code is probably a bit confusing as it's had so many maintainers... but we will see

strogiyotec:
@joakime  @gregw  @sbordet  I made a PR
https://github.com/eclipse/jetty.project/pull/9964
Could you take a look please

sbordet:
@strogiyotec thanks for the PR.
Looking at your implementation, you could use `JarFile.getManifest()`.

However, I think it's opening a can of worms, as looking up `Bundle-License` from the manifest works only for OSGi...
@strogiyotec has designed for other means to lookup the license, but what if there are multiple licenses? 
We risk presenting EPL 2, when can also be AL 2, etc.

I feel we won't be able to be precise, and at that point better to not present anything.

@joakime what do you think?

gregw:
> I feel we won't be able to be precise, and at that point better to not present anything.

We actually have an obligation to present our licenses correctly, as our attempt at doing so is the `NOTICE.txt` file.    It would be nice if we could generate the contents of that through something like this, rather than have it be manually maintained. 

For our own jars, there is no excuse for the manifest not to be accurate with regards to licenses.  For our dependencies, we can raise issues to get them to correct any issues.

Ultimately you care about the the things you measure.   If we don't report on the license listings in the manifest, then they will still be in-precise/wrong... we just wont notice it.   Is that any better?

sbordet:
@gregw from the PR output I was surprised that `slf4j-api.jar` 2.0.5 does not have a license and indeed does not have neither `Bundle-License` nor a LICENSE.txt file.

Turns out in 2.0.7 there is `META-INF/LICENSE.txt`, but it is the MIT one, which would be incredibly difficult to recognize and parse. Fortunately it also has `Bundle-License`.

[...]

PR #9964, opened 2023-06-25
Issue #265 --list-config license enhancement

strogiyotec:
This is a start of a logic that finds the licenses of module jar files
So far it only checks the META-INF/MANIFEST.MF file
Later on, I will add pom.xml and License.txt file checks
@sbordet  could you give me the feedback , I wonder if you agree with my solution
Here is the example output that I got during the testing
```
Jetty Server Classpath:
-----------------------
Version Information on 26 entries in the classpath.
Note: order presented here is how they would appear on the classpath.
      changes to the --module=name command line options will be reflected here.
 0:                  10.0.15 | ${jetty.base}/lib/ext/demo-mock-resources-10.0.15.jar |Eclipse Public License 2.0
 1:         10.0.16-SNAPSHOT | ${jetty.base}/lib/ext/demo-mock-resources-10.0.16-SNAPSHOT.jar |Eclipse Public License 2.0
 2:                    1.3.3 | ${jetty.base}/lib/ext/jakarta.transaction-api-1.3.3.jar |Eclipse Public License 2.0
 3:      1.4.1.v201005082020 | ${jetty.base}/lib/ext/javax.mail.glassfish-1.4.1.v201005082020.jar |(none specified)
 4:                    (dir) | ${jetty.base}/resources |(dir)
 5:                    2.0.5 | ${jetty.home}/lib/logging/slf4j-api-2.0.5.jar |(none specified)
 6:                    4.0.6 | ${jetty.home}/lib/jetty-servlet-api-4.0.6.jar |Apache License 2.0
 7:                    1.3.3 | ${jetty.home}/lib/jakarta.transaction-api-1.3.3.jar |Eclipse Public License 2.0
 8:                      9.5 | ${jetty.home}/lib/annotations/asm-9.5.jar |(none specified)
 9:                      9.5 | ${jetty.home}/lib/annotations/asm-analysis-9.5.jar |(none specified)
10:                      9.5 | ${jetty.home}/lib/annotations/asm-commons-9.5.jar |(none specified)
11:                      9.5 | ${jetty.home}/lib/annotations/asm-tree-9.5.jar |(none specified)
12:                    1.3.5 | ${jetty.home}/lib/annotations/jakarta.annotation-api-1.3.5.jar |Eclipse Public License 2.0
13:    3.26.0.v20210609-0549 | ${jetty.home}/lib/apache-jsp/org.eclipse.jdt.ecj-3.26.0.jar |(none specified)
14:                  10.0.15 | ${jetty.home}/lib/apache-jsp/org.eclipse.jetty.apache-jsp-10.0.15.jar |Eclipse Public License 2.0
15:                   9.0.52 | ${jetty.home}/lib/apache-jsp/org.mortbay.jasper.apache-el-9.0.52.jar |Apache License 2.0
16:                   9.0.52 | ${jetty.home}/lib/apache-jsp/org.mortbay.jasper.apache-jsp-9.0.52.jar |Apache License 2.0
17:                    1.2.5 | ${jetty.home}/lib/apache-jstl/org.apache.taglibs.taglibs-standard-impl-1.2.5.jar |Apache License 2.0
18:                    1.2.5 | ${jetty.home}/lib/apache-jstl/org.apache.taglibs.taglibs-standard-spec-1.2.5.jar |Apache License 2.0
19:                    1.1.2 | ${jetty.home}/lib/websocket/jetty-javax-websocket-api-1.1.2.jar |Apache License 2.0
20:                  10.0.15 | ${jetty.home}/lib/http2/http2-common-10.0.15.jar |Eclipse Public License 2.0
21:                  10.0.15 | ${jetty.home}/lib/http2/http2-hpack-10.0.15.jar |Eclipse Public License 2.0
22:                  10.0.15 | ${jetty.home}/lib/http2/http2-server-10.0.15.jar |Eclipse Public License 2.0
23:                1.70.00.0 | ${jetty.base}/lib/bouncycastle/bcpkix-jdk15on-1.70.jar |(none specified)
24:                   1.70.0 | ${jetty.base}/lib/bouncycastle/bcprov-jdk15on-1.70.jar |(none specified)
25:                1.70.00.0 | ${jetty.base}/lib/bouncycastle/bcutil-jdk15on-1.70.jar |(none specified)

```
All the licenses were taken from [spdx](https://github.com/spdx/license-list-data/blob/main/json/licenses.json)


jetty-start/src/main/java/org/eclipse/jetty/start/AllOSSLicenses.java
```diff
@@ -0,0 +1,536 @@
...
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// ========================================================================
+//
+
+package org.eclipse.jetty.start;
+
+import java.util.AbstractMap;
+import java.util.Map;
+
+public class AllOSSLicenses
```
gregw, line 19: I have an issue with this file.  Firstly it will need to be maintained, but more importantly it is from a generated list that says we need to see the sources to determine the licence for the list itself.   I'm not sure we can check this in and say it is EPL/Apache2

I think we should just report the raw license info we find.... perhaps there is an option to lookup this info, but do so online without checking this is.

jetty-start/src/main/java/org/eclipse/jetty/start/JarLicense.java
```diff
@@ -0,0 +1,123 @@
...
+     * 3. LICENSE file
+     * @param file Jar File
+     * @return License name or (none specified) if license is absent
+     */
+    public static String getLicense(File file)
+    {
+        String nonSpecified = "(none specified)";
+        try (JarFile jar = new JarFile(file))
+        {
+            return findFirstNonNullLicenseOr(
```
gregw, line 47: We need to report all licenses, not just the first.

gregw (changes requested):

jetty-start/src/main/java/org/eclipse/jetty/start/JarLicense.java
```diff
@@ -0,0 +1,123 @@
...
+                List.of(
+                    JarLicense::findFromManifest,
+                    JarLicense::findFromPOM,
+                    JarLicense::findFromLicenseFile
```
joakime, line 51: There should also be a `JarLicense::findFromSPDXIndicator`.
And it should report all of them, not just the first one it finds.

Now, you can present them in order though.
1. `SPDX-License-Identifier` results
2. POM
3. Manifest
4. License Files (NOTICE.*, LICENSE.*) - the output of these should indicate (via URI) where they came from.
If there is any order, it should be present

jetty-start/src/main/java/org/eclipse/jetty/start/AllOSSLicenses.java
joakime: I don't like this file either.
Please remove it.

sbordet: But the "license" we find in `Bundle-License` is a URL like `https://www.eclipse.org/legal/epl-2.0`, so we can't really print the URL, hence this mapping file.

If we remove it, what do we report, the URLs?

joakime: A Bundle-License can be any text. (even non-urls)
Reporting the URLs does make sense, if we have report all of the discovered License information.

joakime: Keep in mind, that with JEP-238, it is possible to have multiple META-INF/MANIFEST.MF files too.
And they could even not agree with each other (we've seen this)

gregw: > If we remove it, what do we report, the URLs?

The URLs are mostly humanly readable.  It is clear what `https://www.eclipse.org/legal/epl-2.0` is.
But if there is need, an option can be provided to download the source json or XML and pretty print it from that. The data is in github, perhaps it is also in maven, in which case we can even cache it locally.

joakime:
Implemented in PR #11993 in branch `jetty-12.0.x`

PR #9968, opened 2023-06-26
Deflake ReservedThreadExecutorTest

gregw:
Use awaitability with longer times to avoid short optimistic waits in tests.

### DIFF
--- a/jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/thread/ReservedThreadExecutorTest.java
+++ b/jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/thread/ReservedThreadExecutorTest.java
@@ -17,22 +17,22 @@ import java.util.ArrayDeque;
 import java.util.Deque;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
-import org.eclipse.jetty.util.NanoTime;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.awaitility.Awaitility.waitAtMost;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.greaterThan;
 import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.fail;
 
 public class ReservedThreadExecutorTest
 {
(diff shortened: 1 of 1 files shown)


jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/thread/ReservedThreadExecutorTest.java
```diff
@@ -174,12 +171,12 @@ public void testEvict() throws Exception
...
 
-        waitForAvailable(2);
+        waitAtMost(10, SECONDS).until(_reservedExecutor::getAvailable, is(2));
 
         int available = _reservedExecutor.getAvailable();
         assertThat(available, is(2));
-        Thread.sleep((5 * IDLE) / 2);
-        assertThat(_reservedExecutor.getAvailable(), is(0));
+
+        waitAtMost(5 * IDLE, MILLISECONDS).until(_reservedExecutor::getAvailable, is(0));
```
lorban, line 179: You're now waiting up to twice as long, 5 times IDLE instead of `(5 * IDLE) / 2`. Is that on purpose?

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/thread/ReservedThreadExecutorTest.java
```diff
@@ -216,38 +213,18 @@ public void testReservedIdleTimeoutWithOneReservedThread() throws Exception
...
 
         assertThat(_reservedExecutor.tryExecute(NOOP), is(false));
         Thread thread = _executor.startThread();
         assertNotNull(thread);
-        waitForAvailable(1);
-
-        Thread.sleep(2 * idleTimeout);
-
-        waitForAvailable(0);
+        waitAtMost(5, SECONDS).until(_reservedExecutor::getAvailable, is(1));
```
lorban, line 224: This is going to be flaky; awaitility may not see the transition to 1.

lorban (changes requested):

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/thread/ReservedThreadExecutorTest.java
gregw: Yep, that shorter wait was always done, hence kept short.   Now we only wait until we see the condition.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/thread/ReservedThreadExecutorTest.java
gregw: CI should not pause for an entire idle timeout period.... if it does, I think more tests would fail.   So I think you are right technically, but I'm hoping that in practise it is fine.... or at least better than now.   Thoughts?

lorban: Okay, there's a natural "long" delay between the two waits, so that should work fine.

Maybe it's worth a comment?

lorban (dismissed):

lorban:
FYI the test failure in the last CI build (`DistributionTests.testIniSectionPropertyOverriddenByCommandLine`) should be fixed by #9967 so it can be ignored.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/thread/ReservedThreadExecutorTest.java
gregw: added

lorban (approved):

PR #9970, opened 2023-06-26
Fix default servlet character encoding directories

### LINKED ISSUE #9966: NullPointerException with default servlet, include and welcome pages
**Jetty version(s)**
12.0.0 SNAPSHOT

**Description**
**How to reproduce?**

I added a comment under #9927, but here's separate issue.

@gregw I can't track the exact part of 0b1c28a8883bb70421a228d1a72644d6125f2baa, but I think it started causing NPE in some special case.

My servlet is invoked with `/gateway/x?what=include&where=/`:
```java
switch (what) {
	case "redirect":
		resp.sendRedirect(where);
		return;
	case "forward":
		return;
	case "include":
		return;
	default:
}
```

`org.eclipse.jetty.ee10.servlet.ServletApiResponse#getWriter()` calls:
```java
String encoding = getServletResponseInfo().getCharacterEncoding(true);
```

`org.eclipse.jetty.ee10.servlet.ServletContextResponse#getCharacterEncoding(boolean)` calls:
```java
if (setContentType)
    setCharacterEncoding(encoding, EncodingFrom.DEFAULT);
```

so `org.eclipse.jetty.ee10.servlet.ServletContextResponse#_characterEncoding` is set to `iso-8859-1`.
Because the forward is for `/`, default servlet is involved (because I need to test welcome files).

`org.eclipse.jetty.ee10.servlet.DefaultServlet#doGet()` has this fragment:
```java
if (contextResponse != null)
{
    String characterEncoding = contextResponse.getRawCharacterEncoding();
    if (characterEncoding != null)
        content = new ForcedCharacterEncodingHttpContent(content, characterEncoding);
}
```

and the constructor is:
```java
public ForcedCharacterEncodingHttpContent(HttpContent content, String characterEncoding)
{
    super(content);
    this.characterEncoding = characterEncoding;
    String mimeType = content.getContentTypeValue();
    int idx = mimeType.indexOf(";charset");
    if (idx >= 0)
        mimeType = mimeType.substring(0, idx);
    this.contentType = mimeType + ";charset=" + this.characterEncoding;
}
```

the `content` is:
```
content = {org.ops4j.pax.web.service.jetty.internal.web.DefaultServlet$UnknownLengthHttpContent@4857} "UnknownLengthHttpContent@c898d8f[CompressedFormatsHttpContent@1c61db46[ResourceHttpContent@471a9af8{r=file:/data/sources/github.com/ops4j/org.ops4j.pax.web-jakarta-new/pax-web-jetty/target/b1/,ct=null}]]"
 _delegate: org.eclipse.jetty.http.content.HttpContent  = {org.eclipse.jetty.http.content.PreCompressedHttpContentFactory$CompressedFormatsHttpContent@4919} "CompressedFormatsHttpContent@1c61db46[ResourceHttpContent@471a9af8{r=file:/data/sources/github.com/ops4j/org.ops4j.pax.web-jakarta-new/pax-web-jetty/target/b1/,ct=null}]"
  _delegate: org.eclipse.jetty.http.content.HttpContent  = {org.eclipse.jetty.http.content.ResourceHttpContent@4922} "ResourceHttpContent@471a9af8{r=file:/data/sources/github.com/ops4j/org.ops4j.pax.web-jakarta-new/pax-web-jetty/target/b1/,ct=null}"
   _contentType: java.lang.String  = null
   _etag: org.eclipse.jetty.http.HttpField  = {org.eclipse.jetty.http.PreEncodedHttpField@4925} "ETag: W/"///+d0tLiXs/////73HTsQ""
   _path: java.nio.file.Path  = {sun.nio.fs.UnixPath@4924} "/data/sources/github.com/ops4j/org.ops4j.pax.web-jakarta-new/pax-web-jetty/target/b1"
   _resource: org.eclipse.jetty.util.resource.Resource  = {org.eclipse.jetty.util.resource.PathResource@4749} "file:/data/sources/github.com/ops4j/org.ops4j.pax.web-jakarta-new/pax-web-jetty/target/b1/"
  compressedFormats: java.util.Set  = {java.util.HashSet@4921}  size = 0
```

the underlying resource is directory, so the extension is `*` and the resulting content type is null.

so I'm getting NullPointerException.

My test you can run (without bothering with OSGi stuff) is https://github.com/ops4j/org.ops4j.pax.web/blob/1b26a2f8ea394d696a6780ff0fa92b2613019dff/pax-web-jetty/src/test/java/org/ops4j/pax/web/service/jetty/internal/UnifiedJettyTest.java#L361

lorban:
Assuming directory listing is allowed when a servlet does the following:
```java
req.getRequestDispatcher("/some/path/to/a/folder").include(req, resp);
```

`resp.getWriter()` sets the character encoding to `iso-8859-1` in `ServletApiResponse.getWriter()` (by calling `ServletContextResponse.getCharacterEncoding(true)`.

This means the response's character encoding is set to `iso-8859-1`. This has some problems:

1) We try to force the character encoding by wrapping the `HttpContent` instance with a  `ForcedCharacterEncodingHttpContent` but since the wrapped content is a directory, it returns null to `getContentTypeValue()` which ends up throwing NPE from the `ForcedCharacterEncodingHttpContent` ctor.

2) The XHTML generated by the include is written as a `utf-8` string to the response on which the character encoding is set to `iso-8859-1`.

3) The response given to the include request can be mutated (status and headers can be modified) and terminated (last write can be performed) while that should be forbidden.

This PR is an attempt at fixing these bugs.

Fixes #9966

### DIFF
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
@@ -13,6 +13,7 @@
 
 package org.eclipse.jetty.server;
 
+import java.nio.charset.Charset;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
@@ -54,7 +55,7 @@ public class ResourceListing
      * @param query query params
      * @return the XHTML as String
      */
-    public static String getAsXHTML(Resource resource, String encoding, String base, boolean parent, String query)
+    public static String getAsXHTML(Resource resource, Charset encoding, String base, boolean parent, String query)
     {
         // This method doesn't check aliases, so it is OK to canonicalize here.
         base = URIUtil.normalizePath(base);
@@ -110,7 +111,7 @@ public class ResourceListing
         StringBuilder buf = new StringBuilder(4096);
 
         // Doctype Declaration + XHTML
-        buf.append("<?xml version=\"1.0\" encoding=\"").append(encoding).append("\"?>\n");
+        buf.append("<?xml version=\"1.0\" encoding=\"").append(encoding.name()).append("\"?>\n");
         buf.append("""
             <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd">
             <html xmlns="http://www.w3.org/1999/xhtml" lang="en" xml:lang="en">
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
@@ -13,7 +13,6 @@
 
 package org.eclipse.jetty.server;
 
-import java.nio.charset.Charset;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
@@ -49,13 +48,12 @@ public class ResourceListing
      * Convert the Resource directory into an XHTML directory listing.
      *
      * @param resource the resource to build the listing from
-     * @param encoding the encoding to specify in the &lt;?xml ... ?&gt; header
      * @param base The base URL
      * @param parent True if the parent directory should be included
      * @param query query params
      * @return the XHTML as String
      */
-    public static String getAsXHTML(Resource resource, Charset encoding, String base, boolean parent, String query)
+    public static String getAsXHTML(Resource resource, String base, boolean parent, String query)
     {
         // This method doesn't check aliases, so it is OK to canonicalize here.
         base = URIUtil.normalizePath(base);
@@ -110,9 +108,10 @@ public class ResourceListing
 
         StringBuilder buf = new StringBuilder(4096);
 
-        // Doctype Declaration + XHTML
-        buf.append("<?xml version=\"1.0\" encoding=\"").append(encoding.name()).append("\"?>\n");
+        // Doctype Declaration + XHTML. The spec says the encoding MUST be "utf-8" in all cases at it is ignored;
+        // see: https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-charset
         buf.append("""
+            <?xml version="1.0" encoding="utf-8"?>
             <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd">
             <html xmlns="http://www.w3.org/1999/xhtml" lang="en" xml:lang="en">
             """);
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
@@ -48,12 +48,13 @@ public class ResourceListing
      * Convert the Resource directory into an XHTML directory listing.
      *
      * @param resource the resource to build the listing from
+     * @param encoding the encoding to specify in the &lt;?xml ... ?&gt; header
      * @param base The base URL
      * @param parent True if the parent directory should be included
      * @param query query params
-     * @return the HTML as String
+     * @return the XHTML as String
      */
-    public static String getAsXHTML(Resource resource, String base, boolean parent, String query)
+    public static String getAsXHTML(Resource resource, String encoding, String base, boolean parent, String query)
     {
         // This method doesn't check aliases, so it is OK to canonicalize here.
         base = URIUtil.normalizePath(base);
@@ -109,8 +110,8 @@ public class ResourceListing
         StringBuilder buf = new StringBuilder(4096);
 
         // Doctype Declaration + XHTML
+        buf.append("<?xml version=\"1.0\" encoding=\"").append(encoding).append("\"?>\n");
         buf.append("""
-            <?xml version="1.0" encoding="utf-8"?>
             <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd">
             <html xmlns="http://www.w3.org/1999/xhtml" lang="en" xml:lang="en">
             """);
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
@@ -17,6 +17,7 @@ import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.channels.ReadableByteChannel;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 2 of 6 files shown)


jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
```diff
@@ -618,16 +618,18 @@ private void sendDirectory(Request request, Response response, HttpContent httpC
             return;
         }
 
+        String characterEncoding = httpContent.getCharacterEncoding() != null ? httpContent.getCharacterEncoding() : "utf-8";
```
janbartel:
```suggestion
        String characterEncoding = httpContent.getCharacterEncoding();
        characterEncoding == null ? "utf-8"  : characterEncoding;
```

janbartel: Minor, just to avoid calling the getter twice.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -1233,11 +1233,22 @@ public ForcedCharacterEncodingHttpContent(HttpContent content, String characterE
         {
             super(content);
             this.characterEncoding = characterEncoding;
-            String mimeType = content.getContentTypeValue();
-            int idx = mimeType.indexOf(";charset");
-            if (idx >= 0)
-                mimeType = mimeType.substring(0, idx);
-            this.contentType = mimeType + ";charset=" + this.characterEncoding;
+            if (content.getResource().isDirectory())
```
janbartel: Are we sure resource can never be null here?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -1233,11 +1233,22 @@ public ForcedCharacterEncodingHttpContent(HttpContent content, String characterE
...
+            {
+                this.contentType = "text/html;charset=" + characterEncoding;
```
janbartel: If it's a directory, but we're going to serve a welcome that then goes on to be handled by a servlet, is it ok to force the content type here?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
lorban: Done.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
lorban: At this point, yes, because there is an explicit null check in the method that calls the `ForcedCharacterEncodingHttpContent` constructor to take a different branch.
But I'm going to add an extra explicit null check here for clarity.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
gregw: Should we use `StandardCharsets.UTF_8.name()` instead of "utf-8"?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
lorban: My understanding is that if the content is a directory, the `HttpContent`'s content type is going to be ignored. I've just set it here to something somewhat reasonable to have `characterEncoding` set somewhere. But maybe I should have set `contentType` to null.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
```diff
@@ -48,12 +48,13 @@ public class ResourceListing
...
      * @param resource the resource to build the listing from
+     * @param encoding the encoding to specify in the &lt;?xml ... ?&gt; header
      * @param base The base URL
      * @param parent True if the parent directory should be included
      * @param query query params
-     * @return the HTML as String
+     * @return the XHTML as String
      */
-    public static String getAsXHTML(Resource resource, String base, boolean parent, String query)
+    public static String getAsXHTML(Resource resource, String encoding, String base, boolean parent, String query)
```
gregw: Should be pass around a `Charset` rather than a `String`.  In the common case we can then just hard code that that `StandardCharsets.UTF_8`.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -861,18 +861,22 @@ private static class ServletCoreResponse implements Response
...
             _coreResponse = ServletContextResponse.getServletContextResponse(response);
             _httpFields = new HttpServletResponseHttpFields(response);
+            _included = include;
         }
 
         @Override
         public HttpFields.Mutable getHeaders()
         {
+            if (_included)
+                return HttpFields.build(_httpFields);
```
gregw: Hmmm this is not exactly the right semantic.   It will not propagate sets to the real headers, but a set then a get would see the wrong updated value.   We need a special `HttpFields` wrapper that just ignores sets.

gregw (changes requested):
These are good changes... just need a little more.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
lorban: You're right, let me fix that.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceService.java
lorban: At this point, we're getting the character encoding as a string from `HttpContent`. Keeping the type a string is more natural.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/ResourceListing.java
lorban: The encoding is only used for string concatenation, so is it worth passing a `Charset` object? How about the case of the charset name, should it be then forced to be lowercase or do we use the name reported by the `Charset` instance verbatim, even if it's a slight behavior change?

gregw: The [spec](https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-charset) says it is case-insensitive, so it is change it now or never!
I still prefer passing the `Charset` as that makes the case decision in only one place, else we may have "utf-8", "utf8", "UTF-8" and "UTF8" scattered all over the code.

lorban: I'm convinced, let me do it.

lorban: Hey, thanks for the [spec link](https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-charset)! I've learnt that it says:

 _its value must be an ASCII case-insensitive match for the string "utf-8"._ 

and 

 _The charset attribute on the meta element has no effect in XML documents_

so I've reverted these changes to keep the originally hardcoded "utf-8" string.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -863,20 +863,37 @@ private static class ServletCoreResponse implements Response
...
+                    {
+                        return null;
+                    }
+
+                    @Override
+                    public boolean onRemoveField(HttpField field)
+                    {
+                        return false;
+                    }
+                };
```
gregw, line 888: Hey that's pretty neat for a newly added "feature" on the wrapper!

gregw (approved):

PR #9971, opened 2023-06-26
Fix #9960 Improve Request methods, including NPE protection.

### LINKED ISSUE #9960: Custom logging in Jetty 12 beta2 can fail due to NullPointerException in org.eclipse.jetty.server.Request
**Jetty version**
Jetty 12 beta2 - Embedded

openjdk version "17.0.7" 2023-04-18
OpenJDK Runtime Environment Temurin-17.0.7+7 (build 17.0.7+7)
OpenJDK 64-Bit Server VM Temurin-17.0.7+7 (build 17.0.7+7, mixed mode, sharing)

**OS type/version**
Windows 11 - Microsoft Windows [Version 10.0.22631.1906]

**Description**
In case of errors in the code of the Jetty 12 emebbor, the custom log attempts to write out the request, but will fail due to NullPointerException's. Of course, Jetty is not to blame for the implementors problematic code, but this is nicely handled by try-catches and console logging, except if you add a custom logger to the server.

I have seens various such errors, and most of them are located in the `org.eclipse.jetty.server.Request` class such as get methods returning a String. For example:

```java
    static String getRemoteAddr(Request request)
    {
        if (request == null)
            return null;
        SocketAddress remote = request.getConnectionMetaData().getRemoteSocketAddress();
        if (remote instanceof InetSocketAddress inetSocketAddress)
        {
            if (inetSocketAddress.isUnresolved())
                return inetSocketAddress.getHostString();

            InetAddress address = inetSocketAddress.getAddress();
            String result = address == null
                ? inetSocketAddress.getHostString()
                : address.getHostAddress();
            return HostPort.normalizeHost(result);
        }
        return remote.toString();
    }
```

could go `NullPointerException` at the last line. Perhaps this should be rewritten to something like:

```java
        return (remote != null)? remote.toString(): "null" /* or something else? */;
```

This applies to some potential methods in the `Request` class, from what I have seen until now:

- getLocalAddr
- getRemoteAddr
- getServerName

In contrast, you have methods such as `getRemotePort` not having this problem as it doesn't attempt to use the `remote`variable for method calls in its class:

```java
    static int getRemotePort(Request request)
    {
        if (request == null)
            return -1;
        SocketAddress remote = request.getConnectionMetaData().getRemoteSocketAddress();
        if (remote instanceof InetSocketAddress)
            return ((InetSocketAddress)remote).getPort();
        return -1;
    }
```

Above you see the `return -1;` call not attempting to use `remote` that could potentially be `null`.

**How to reproduce?**
Having an embedded Jetty with a custom logger and causing exceptions in the embeddors code that are then caught by Jetty handlers.

The one that really came up to be the most troublesome was `getRemoteAddr`.

### LINKED ISSUE #9966: NullPointerException with default servlet, include and welcome pages

gregw:
Fix #9960 with NPE protection for bad requests.

### DIFF
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
@@ -25,8 +25,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
-import java.util.function.UnaryOperator;
-import java.util.stream.Collectors;
+import java.util.Set;
 
 import org.eclipse.jetty.util.FileID;
 import org.eclipse.jetty.util.Index;
@@ -40,39 +39,17 @@ import org.slf4j.LoggerFactory;
 public class MimeTypes
 {
     static final  Logger LOG = LoggerFactory.getLogger(MimeTypes.class);
-    private static final Index<Locale> KNOWN_LOCALES = new Index.Builder<Locale>()
-        .caseSensitive(false)
-        .withAll(() -> Arrays.stream(Locale.getAvailableLocales())
-            .collect(Collectors.toMap(Locale::toString, UnaryOperator.identity())))
-        .build();
+    private static final Set<Locale> KNOWN_LOCALES = Set.copyOf(Arrays.asList(Locale.getAvailableLocales()));
 
     /**
-     * @param languageDashCountry The language and optional country of the local. For example "en" or "en-GB".
-     * @return The {@link Locale} matching the string or null if a matching Locale is not known;
+     * @param languageTag The language and optional country of the local. For example "en" or "en-GB".
+     * @return A {@link Locale} known to the JVM matching the languageTage or null if the Locale is not known;
+     * @see Locale#forLanguageTag(String)
      */
-    public static Locale getLocale(String languageDashCountry)
+    public static Locale getLocale(String languageTag)
     {
-        // First try for a direct hit in the known Locale cache
-        Locale locale = KNOWN_LOCALES.get(languageDashCountry);
-        if (locale != null)
-            return locale;
-
-        // failing that, create a new Locale instance to normalize the name
-        String language = HttpField.stripParameters(languageDashCountry);
-        int dash = languageDashCountry.indexOf('-');
-        if (dash < 0)
-        {
-            locale = new Locale(language);
-        }
-        else
-        {
-            language = languageDashCountry.substring(0, dash).trim();
-            String country = languageDashCountry.substring(dash + 1).trim();
-            locale = new Locale(language, country);
-        }
-
-        // look up normalized name in the known Locale cache.
-        return KNOWN_LOCALES.get(locale.toString());
+        Locale locale =  Locale.forLanguageTag(languageTag);
+        return KNOWN_LOCALES.contains(locale) ? locale : null;
     }
 
     /** Enumeration of predefined MimeTypes. This is not exhaustive */
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
@@ -18,14 +18,12 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
-import java.util.Set;
 
 import org.eclipse.jetty.util.FileID;
 import org.eclipse.jetty.util.Index;
@@ -39,18 +37,6 @@ import org.slf4j.LoggerFactory;
 public class MimeTypes
 {
     static final  Logger LOG = LoggerFactory.getLogger(MimeTypes.class);
-    private static final Set<Locale> KNOWN_LOCALES = Set.copyOf(Arrays.asList(Locale.getAvailableLocales()));
-
-    /**
-     * @param languageTag The language and optional country of the local. For example "en" or "en-GB".
-     * @return A {@link Locale} known to the JVM matching the languageTage or null if the Locale is not known;
-     * @see Locale#forLanguageTag(String)
-     */
-    public static Locale getLocale(String languageTag)
-    {
-        Locale locale =  Locale.forLanguageTag(languageTag);
-        return KNOWN_LOCALES.contains(locale) ? locale : null;
-    }
 
     /** Enumeration of predefined MimeTypes. This is not exhaustive */
     public enum Type
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
@@ -37,6 +37,43 @@ import org.slf4j.LoggerFactory;
 public class MimeTypes
 {
     static final  Logger LOG = LoggerFactory.getLogger(MimeTypes.class);
+    private static final Index<Locale> KNOWN_LOCALES;
+
+    static
+    {
+        Index.Builder<Locale> builder = new Index.Builder<>();
+        builder = builder.caseSensitive(false);
+        for (Locale locale : Locale.getAvailableLocales())
+            builder.with(locale);
+        KNOWN_LOCALES = builder.build();
+    }
+
+    /**
+     * @param languageDashCountry The language and optional country of the local. For example "en" or "un-uk".
+     * @return The {@link Locale} matching the string or null if a matching Locale is not known;
+     */
+    public static Locale getLocale(String languageDashCountry)
+    {
+        // First try for a direct hit in the known Locale cache
+        Locale locale = KNOWN_LOCALES.get(languageDashCountry);
+        if (locale != null)
+            return locale;
+
+        // failing that, create a new Locale instance to normalize the name
+        String language = HttpField.stripParameters(languageDashCountry);
+        int dash = languageDashCountry.indexOf('-');
+        if (dash < 0)
+            locale = new Locale(language);
+        else
+        {
+            language = languageDashCountry.substring(0, dash).trim();
+            String country = languageDashCountry.substring(dash + 1).trim();
+            locale = new Locale(language, country);
+        }
+
+        // look up normalized name in the known Locale cache.
+        return KNOWN_LOCALES.get(locale.toString());
+    }
 
     /** Enumeration of predefined MimeTypes. This is not exhaustive */
     public enum Type
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
@@ -18,12 +18,15 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
+import java.util.function.UnaryOperator;
+import java.util.stream.Collectors;
 
 import org.eclipse.jetty.util.FileID;
 import org.eclipse.jetty.util.Index;
@@ -37,19 +40,14 @@ import org.slf4j.LoggerFactory;
 public class MimeTypes
 {
     static final  Logger LOG = LoggerFactory.getLogger(MimeTypes.class);
-    private static final Index<Locale> KNOWN_LOCALES;
-
-    static
-    {
-        Index.Builder<Locale> builder = new Index.Builder<>();
-        builder = builder.caseSensitive(false);
-        for (Locale locale : Locale.getAvailableLocales())
-            builder.with(locale);
-        KNOWN_LOCALES = builder.build();
-    }
+    private static final Index<Locale> KNOWN_LOCALES = new Index.Builder<Locale>()
+        .caseSensitive(false)
+        .withAll(() -> Arrays.stream(Locale.getAvailableLocales())
+            .collect(Collectors.toMap(Locale::toString, UnaryOperator.identity())))
+        .build();
 
     /**
-     * @param languageDashCountry The language and optional country of the local. For example "en" or "un-uk".
+     * @param languageDashCountry The language and optional country of the local. For example "en" or "en-GB".
      * @return The {@link Locale} matching the string or null if a matching Locale is not known;
      */
     public static Locale getLocale(String languageDashCountry)
@@ -63,7 +61,9 @@ public class MimeTypes
         String language = HttpField.stripParameters(languageDashCountry);
         int dash = languageDashCountry.indexOf('-');
         if (dash < 0)
+        {
             locale = new Locale(language);
+        }
         else
         {
             language = languageDashCountry.substring(0, dash).trim();
--- a/jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
+++ b/jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
@@ -17,7 +17,9 @@ import java.io.ByteArrayOutputStream;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Locale;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Stream;
 
 import org.eclipse.jetty.http.HttpCookie;
 import org.eclipse.jetty.http.HttpHeader;
@@ -30,6 +32,9 @@ import org.eclipse.jetty.util.component.LifeCycle;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
@@ -370,6 +375,37 @@ public class RequestTest
         HttpTester.Response response = HttpTester.parseHeadResponse(connector.getResponse(rawRequest));
         assertNotNull(response);
         assertThat(response.getStatus(), is(HttpStatus.OK_200));
+    }
+
+    public static Stream<Arguments> localeTests()
+    {
+        return Stream.of(
+            Arguments.of(null, List.of(Locale.getDefault()).toString()),
+            Arguments.of("zz", List.of(Locale.US.toString()).toString()),
+            Arguments.of("en", "[en]"),
+            Arguments.of("en-us", List.of(Locale.US.toString()).toString()),
+            Arguments.of("EN-US", List.of(Locale.US.toString()).toString()),
+            Arguments.of("en-us,en-gb", List.of(Locale.US.toString(), Locale.UK).toString()),
+            Arguments.of("en-us;q=0.5,fr;q=0.0,en-gb;q=1.0", List.of(Locale.UK.toString(), Locale.US).toString())
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("localeTests")
+    public void testAcceptableLocales(String acceptLanguage, String expectedLocales) throws Exception
+    {
+        acceptLanguage = acceptLanguage == null ? "" : (HttpHeader.ACCEPT_LANGUAGE.asString() + ": " + acceptLanguage + "\n");
+        String rawRequest = """
+                GET / HTTP/1.1
+                Host: tester
+                Connection: close
+                %s
+                """.formatted(acceptLanguage);
+
+        HttpTester.Response response = HttpTester.parseResponse(connector.getResponse(rawRequest));
+        assertNotNull(response);
+        assertThat(response.getStatus(), is(HttpStatus.OK_200));
+        assertThat(response.getContent(), containsString("locales=" + expectedLocales));
     }
 
     private static void checkCookieResult(String containedCookie, String[] notContainedCookies, String response)
--- a/jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
+++ b/jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
@@ -380,13 +380,14 @@ public class RequestTest
     public static Stream<Arguments> localeTests()
     {
         return Stream.of(
-            Arguments.of(null, List.of(Locale.getDefault()).toString()),
-            Arguments.of("zz", List.of(Locale.US).toString()),
+            Arguments.of(null, List.of(Locale.getDefault().toLanguageTag()).toString()),
+            Arguments.of("zz", List.of(Locale.US.toLanguageTag()).toString()),
             Arguments.of("en", "[en]"),
-            Arguments.of("en-us", List.of(Locale.US).toString()),
-            Arguments.of("EN-US", List.of(Locale.US).toString()),
-            Arguments.of("en-us,en-gb", List.of(Locale.US, Locale.UK).toString()),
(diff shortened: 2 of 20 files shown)


jetty-ee9/jetty-ee9-nested/src/main/java/org/eclipse/jetty/ee9/nested/Request.java
```diff
@@ -858,7 +858,7 @@ public Locale getLocale()
             country = language.substring(dash + 1).trim();
             language = language.substring(0, dash).trim();
         }
-        return new Locale(language, country);
+        return Locale.of(language, country);
```
lorban: `Locale.of()` only appeared in JDK 19: https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/util/Locale.html#of(java.lang.String,java.lang.String)

lorban: Not fixed?
12.0.x requires JDK 17 so you can't use that API.

lorban: Not fixed?
12.0.x requires JDK 17 so you can't use that API.

lorban (changes requested):

gregw:
I'm going to do a more comprehensive fix for the locale... Stand by...

gregw:
@lorban I replaced with a better Locale implementations that actually handles the quality list that we go to the trouble of parsing (except for ee9, where I keep behaviour the same).
Impl could be better in java19, but I use an `Index` to catch the most common cases.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
```diff
@@ -37,6 +37,43 @@
...
+    {
+        Index.Builder<Locale> builder = new Index.Builder<>();
+        builder = builder.caseSensitive(false);
+        for (Locale locale : Locale.getAvailableLocales())
+            builder.with(locale);
+        KNOWN_LOCALES = builder.build();
+    }
+
+    /**
+     * @param languageDashCountry The language and optional country of the local. For example "en" or "un-uk".
```
lorban: I guess you meant `en-uk` instead of `un-uk`. But that doesn't even exist, the right form is actually `en-gb`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
```diff
@@ -37,6 +37,43 @@
 public class MimeTypes
 {
     static final  Logger LOG = LoggerFactory.getLogger(MimeTypes.class);
+    private static final Index<Locale> KNOWN_LOCALES;
+
+    static
+    {
+        Index.Builder<Locale> builder = new Index.Builder<>();
```
lorban: Stream equivalent:

```java
KNOWN_LOCALES = new Index.Builder<Locale>()
    .caseSensitive(false)
    .withAll(() -> Arrays.stream(Locale.getAvailableLocales()).collect(Collectors.toMap(Locale::toString, UnaryOperator.identity())))
    .build();
```

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
```diff
@@ -37,6 +37,43 @@
...
+        // First try for a direct hit in the known Locale cache
+        Locale locale = KNOWN_LOCALES.get(languageDashCountry);
+        if (locale != null)
+            return locale;
+
+        // failing that, create a new Locale instance to normalize the name
+        String language = HttpField.stripParameters(languageDashCountry);
+        int dash = languageDashCountry.indexOf('-');
+        if (dash < 0)
+            locale = new Locale(language);
```
lorban: Missing `{` and `}` over the `if` block.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
```diff
@@ -372,6 +377,37 @@ public boolean handle(org.eclipse.jetty.server.Request request, Response respons
         assertThat(response.getStatus(), is(HttpStatus.OK_200));
     }
 
+    public static Stream<Arguments> localeTests()
+    {
+        return Stream.of(
```
lorban: There is no need to call `toString()` on the locales that are added to the lists, that just adds confusion.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
```diff
@@ -37,6 +37,43 @@
...
+        else
+        {
+            language = languageDashCountry.substring(0, dash).trim();
+            String country = languageDashCountry.substring(dash + 1).trim();
+            locale = new Locale(language, country);
+        }
+
+        // look up normalized name in the known Locale cache.
+        return KNOWN_LOCALES.get(locale.toString());
```
lorban: I do not understand why a `Locale` is created, then turned into a string, then use that string to look up a locale from an index.

Either there is no need to create the new Locale object, or I'm missing something subtle that's lacking a test and/or some comment.

lorban (changes requested):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
gregw: @lorban Read the comments!

gregw: It is to normalize the name if we miss the first cache try.

gregw: I think we could do better with java 19

lorban: Normalize how?

gregw: uk to GB and other transformations that are magically done inside the Locale class.   I've done code coverage and the test covers all paths.

gregw: @lorban see the private `convertOldISOCodes` method within the Locale constructor.  That's what I'm accessing.

lorban: Got it, thanks.

lorban (dismissed):

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
```diff
@@ -37,6 +40,40 @@
...
+        }
+        else
+        {
+            language = languageDashCountry.substring(0, dash).trim();
+            String country = languageDashCountry.substring(dash + 1).trim();
+            locale = new Locale(language, country);
+        }
+
+        // look up normalized name in the known Locale cache.
+        return KNOWN_LOCALES.get(locale.toString());
```
joakime:
```suggestion
       Locale locale = Locale.forLanguageTag(languageDashCountry);
       return KNOWN_LOCALES.get(locale.toLanguageTag());
```

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
```diff
@@ -37,6 +40,40 @@
 public class MimeTypes
 {
     static final  Logger LOG = LoggerFactory.getLogger(MimeTypes.class);
+    private static final Index<Locale> KNOWN_LOCALES = new Index.Builder<Locale>()
+        .caseSensitive(false)
+        .withAll(() -> Arrays.stream(Locale.getAvailableLocales())
+            .collect(Collectors.toMap(Locale::toString, UnaryOperator.identity())))
```
joakime: Don't use the unreliable `Locale.toString()` here which is for end users, use the `Locale.toLanguageTag()` which uses reliable language tags.

```suggestion
            .collect(Collectors.toMap(Locale::toLanguageTag, UnaryOperator.identity())))
```

joakime (dismissed):
There's a better way to use `Locale`

gregw:
> There's a better way to use `Locale`

@joakime which is? The API was improved for java 19, which we can't use.

Edit: oh i see the `forLanguageTag`, I'll give that a go....

gregw:
The language tag is not suitable, as it is the long form and will never match anything coming from a HTTP header. I tried that first as i also don't like using to string, but it was unsuitable.

joakime:
> The language tag is not suitable, as it is the long form and will never match anything coming from a HTTP header. I tried that first as i also don't like using to string, but it was unsuitable.

What kind of value are you having problems with?
I cannot find an example of a Locale header that fails with `Locale.forLanguageTag(String)`.

``` java
package locales;

import java.util.Locale;

public class LocaleDemo
{
    public static void main(String[] args)
    {
        dump("en");
        dump("en-gb");
        dump("en-US");
        dump("fr-CA");
        dump("sr-Latn");
        dump("de-DE-1996");
    }

    private static void dump(String input)
    {
        Locale locale = Locale.forLanguageTag(input);

        System.out.println("----");
        System.out.println("Locale: " + locale);
        System.out.println("Locale.toString(): " + locale.toString());
        System.out.println("Locale.toLanguageTag(): " + locale.toLanguageTag());
        System.out.println("Locale.getDisplayCountry(): " + locale.getDisplayCountry());
        System.out.println("Locale.getDisplayLanguage(): " + locale.getDisplayLanguage());
        System.out.println("Locale.getDisplayName(): " + locale.getDisplayName());
        System.out.println("Locale.getDisplayScript(): " + locale.getDisplayScript());
        System.out.println("Locale.getDisplayVariant(): " + locale.getDisplayVariant());
    }
}
```

with output (from Java 17.0.5)

```
----
Locale: en
Locale.toString(): en
Locale.toLanguageTag(): en
Locale.getDisplayCountry(): 
Locale.getDisplayLanguage(): English
Locale.getDisplayName(): English
Locale.getDisplayScript(): 
Locale.getDisplayVariant(): 
----
Locale: en_GB
Locale.toString(): en_GB
Locale.toLanguageTag(): en-GB
Locale.getDisplayCountry(): United Kingdom
Locale.getDisplayLanguage(): English
Locale.getDisplayName(): English (United Kingdom)
Locale.getDisplayScript(): 
Locale.getDisplayVariant(): 
----
Locale: en_US
Locale.toString(): en_US
Locale.toLanguageTag(): en-US
Locale.getDisplayCountry(): United States
Locale.getDisplayLanguage(): English
Locale.getDisplayName(): English (United States)
Locale.getDisplayScript(): 
Locale.getDisplayVariant(): 
----
Locale: fr_CA
Locale.toString(): fr_CA
Locale.toLanguageTag(): fr-CA
Locale.getDisplayCountry(): Canada
Locale.getDisplayLanguage(): French
Locale.getDisplayName(): French (Canada)
Locale.getDisplayScript(): 
Locale.getDisplayVariant(): 
----
Locale: sr__#Latn
Locale.toString(): sr__#Latn
Locale.toLanguageTag(): sr-Latn
Locale.getDisplayCountry(): 
Locale.getDisplayLanguage(): Serbian
Locale.getDisplayName(): Serbian (Latin)
Locale.getDisplayScript(): Latin
Locale.getDisplayVariant(): 
----
Locale: de_DE_1996
Locale.toString(): de_DE_1996
Locale.toLanguageTag(): de-DE-1996
Locale.getDisplayCountry(): Germany
Locale.getDisplayLanguage(): German
Locale.getDisplayName(): German (Germany, German orthography of 1996)
Locale.getDisplayScript(): 
Locale.getDisplayVariant(): German orthography of 1996
```

joakime:
If I include the 3 examples from RFC2616 - https://datatracker.ietf.org/doc/html/rfc2616#section-3.10
```
en-cockney, i-cherokee, x-pig-latin
```

Then you have the following output ...

```
----: en-cockney
Locale: en__cockney
Locale.toString(): en__cockney
Locale.toLanguageTag(): en-cockney
Locale.getDisplayCountry(): 
Locale.getDisplayLanguage(): English
Locale.getDisplayName(): English (cockney)
Locale.getDisplayScript(): 
Locale.getDisplayVariant(): cockney
----: i-cherokee
Locale: 
Locale.toString(): 
Locale.toLanguageTag(): und
Locale.getDisplayCountry(): 
Locale.getDisplayLanguage(): 
Locale.getDisplayName(): 
Locale.getDisplayScript(): 
Locale.getDisplayVariant(): 
----: x-pig-latin
Locale: 
Locale.toString(): 
Locale.toLanguageTag(): x-pig-latin
Locale.getDisplayCountry(): 
Locale.getDisplayLanguage(): 
Locale.getDisplayName(): 
Locale.getDisplayScript(): 
Locale.getDisplayVariant(): 
```

You will get a `Locale` object from them, but in the case of `i-cherokee` it is `und` for Undetermined (as it is not well-formed), and `x-pig-latin` is just accepted as-is, without display information.
Notice that the `toString()` results are blank too, but not the `toLanguageTag()` results.

joakime:
> The language tag is not suitable, as it is the long form and will never match anything coming from a HTTP header. I tried that first as i also don't like using to string, but it was unsuitable.

The `Index<Locale> KNOWN_LOCALES` will not match contents from the HTTP Header either, as the contents of that `Index` is using `toString()` which will use underscore `_` (sometimes doubled), and `#`, and other characters not seen on the HTTP Header.

Example:

```
Accept-Language: en-US
```

The `Index<Locale> KNOWN_LOCALES` is stored with `en_US` and will never match the header `en-US`.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
gregw: Strange, I had tried that, but was originally getting long form strings like getDisplayName.   Must have been my-bad somehow.   OK will use this!

joakime:
Just run this and you'll see the results of what `Locale.toString()` is doing to the Index.

``` java
package locales;

import java.util.Arrays;
import java.util.Comparator;
import java.util.Locale;

public class LocaleDump
{
    public static void main(String[] args)
    {
        Arrays.stream(Locale.getAvailableLocales())
            .sorted(Comparator.comparing(Locale::toString))
            .forEach((locale) ->
                System.out.printf("%s -> languageTag[%s]%n", locale.toString(), locale.toLanguageTag()));
    }
}
```

Also, if you want a challenging set of languages for your test cases, use these ...

```
ja-JP-u-ca-japanese-x-lvariant-JP
ja-JP
th-TH-u-nu-thai-x-lvariant-TH
th-TH
```

Those should all result in different `Locale` objects.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
gregw: @joakime yep, that's much much better.  We can even change the Index to just a Set

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
```diff
@@ -37,6 +39,18 @@
...
+
+    /**
+     * @param languageTag The language and optional country of the local. For example "en" or "en-GB".
+     * @return A {@link Locale} known to the JVM matching the languageTage or null if the Locale is not known;
+     * @see Locale#forLanguageTag(String)
+     */
+    public static Locale getLocale(String languageTag)
+    {
+        Locale locale =  Locale.forLanguageTag(languageTag);
+        return KNOWN_LOCALES.contains(locale) ? locale : null;
```
joakime: This wont handle well-formed, but unimplemented Locales, like `x-pig-latin` (from the HTTP RFCs)
This is a valid `Locale` to Java, it just doesn't have built-in calendar or number variants.
Nor is it listed in the results of `Locale.getAvailableLocales()`

Oddly, it will handle not well-formed Locales, like `i-cherokee`.

Why do we need the cache here at all?
The `Locale` class in the JVM has it's own cache.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
```diff
@@ -372,6 +377,38 @@ public boolean handle(org.eclipse.jetty.server.Request request, Response respons
...
+            Arguments.of(null, List.of(Locale.getDefault().toLanguageTag()).toString()),
+            Arguments.of("zz", List.of(Locale.US.toLanguageTag()).toString()),
+            Arguments.of("en", "[en]"),
+            Arguments.of("en-gb", List.of(Locale.UK.toLanguageTag()).toString()),
+            Arguments.of("en-us", List.of(Locale.US.toLanguageTag()).toString()),
+            Arguments.of("EN-US", List.of(Locale.US.toLanguageTag()).toString()),
+            Arguments.of("en-us,en-gb", List.of(Locale.US.toLanguageTag(), Locale.UK.toLanguageTag()).toString()),
+            Arguments.of("en-us;q=0.5,fr;q=0.0,en-gb;q=1.0", List.of(Locale.UK.toLanguageTag(), Locale.US.toLanguageTag()).toString())
```
joakime: This is missing the check for the `fr` (French) tag.

joakime (commented):
We should see if this remaining cache still has any purpose over the built-in cache already present in the JVM's `Locale` class.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
```diff
@@ -37,6 +39,18 @@
 public class MimeTypes
 {
     static final  Logger LOG = LoggerFactory.getLogger(MimeTypes.class);
+    private static final Set<Locale> KNOWN_LOCALES = Set.copyOf(Arrays.asList(Locale.getAvailableLocales()));
+
+    /**
+     * @param languageTag The language and optional country of the local. For example "en" or "en-GB".
+     * @return A {@link Locale} known to the JVM matching the languageTage or null if the Locale is not known;
```
joakime: I'm not sure returning `null` is valid here.

All inputs for `languageTag` (that are not `null`), even empty `""`, dash `"-"`, and control codes will return the `und` `Locale` entry from the internal `Locale` cache (which is present in the `Locale.getAvailableLocales()` results too)

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
gregw: Well we never handled anything other than language dash country before, so no need to handle every variant.

The cache is not there as a cache, but as a check for known locales.  We use this (and the null return) because the `Accept-Language` header is a quality sorted list.   Previously we have always just taken the first locale from the list and used that even if it is not known to the JVM.    But if we check with known Locales, then we can skip a high q unknown locale in favour of using a lower q value known locale.  Perhaps we should fall back to using the first unknown if there are no known locales?

Note that this was just simple fix of a deprecated locale API in a PR that was otherwise about NPEs.   Perhaps I should revert the Locale stuff to get the NPE fixes in and we can make a Locale specific PR to discuss exactly what we should do with the quality list of Locales?

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
gregw: because it is "q=0.0" and thus excluded from the list!

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
gregw: @joakime see discussion above about quality lists

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/RequestTest.java
joakime: > because it is "q=0.0" and thus excluded from the list!

Ah, good point.

joakime (dismissed):
Lets get this merged, as the NPE fix is important.
We can have a followup PR that addresses some of the Locale specific issues.

gregw:
> Lets get this merged, as the NPE fix is important. We can have a followup PR that addresses some of the Locale specific issues.

Correct answer!

I'll strip out the Locale changes later tonight and open a new PR with them. Standby.....

gregw:
@joakime @lorban I've reverted all the Locale changes so this is just the NPE fixes and some javadoc fixes now.

lorban (approved):

PR #9975, opened 2023-06-26
Experiment with a fully async ContentSourceCompletableFuture

gregw:
Currently `MultiPartFormData`  is only partially used and `MultiPartByteRanges` is not used at all.   Both contain duplicate code that does not have a clear split between the responsibilities of managing the `Content.Source`, parsing the data and managing the `CompletableFuture`.  

The usage of `MultiPartFormData` is in blocking style any bypasses the `Content.Source`. If the `DelayedHandler.UntilMultiPartFormData` was also used, then 2 attempts to parse the content would be made, one asynchronous that would be successful, but ignored and the second would just see EOF!

This PR introduces a `ContentSourceCompletableFuture` which attempts to extract all the common aspects of `MultiPartFormData` and `MultiPartByteRanges`. Once updated, then the usage needs to be updated so that it will follow the pattern of `org.eclipse.jetty.server.FormFields#from(org.eclipse.jetty.server.Request, int, int)` so that content can be read asynchronously before a blocking servlet is invoked.   However it needs to be a little more complex so that context or servlet specific configuration can be used.


jetty-core/jetty-client/src/test/java/org/eclipse/jetty/client/util/MultiPartRequestContentTest.java
```diff
@@ -439,9 +439,9 @@ public boolean handle(Request request, Response response, Callback callback) thr
             String contentType = request.getHeaders().get(HttpHeader.CONTENT_TYPE);
             assertEquals("multipart/form-data", HttpField.valueParameters(contentType, null));
             String boundary = MultiPart.extractBoundary(contentType);
-            MultiPartFormData formData = new MultiPartFormData(boundary);
+            MultiPartFormData formData = new MultiPartFormData(request, boundary);
             formData.setFilesDirectory(tmpDir);
-            formData.parse(request);
+            formData.parse();
```
sbordet: Don't like this because it makes `MultiPartFormData` non-reusable, and it is also weird because a constructor dependency typically implies "I need this component to work", while here the source it's not necessary at all -- it is only necessary when parsing it.

It is also confusing when using `MultiPartFormData` to parse chunks, as it would happen in the client.
What would be the source passed to the constructor used for?

sbordet (changes requested):

sbordet:
@gregw sorry but `MultiPartBytesRanges` is critically used in `ResourceService` to handle ranged responses.
It will also be used on the client to parse those, so it's not true that it is not used.

> The usage of MultiPartFormData is in blocking style any bypasses the Content.Source

Not sure what you mean, as all `MultiPartFormData` APIs are non-blocking?

I'm pretty much -1 on this PR, I don't see the savings, the two classes used in different contexts, and while there is a little duplication, I don't think it's worth an abstract class, also considering that the current duplication may not be a duplication anymore if other parameters are added -- for example max parse length, etc.

I also don't see your point about double parsing?
If `DelayedHandler.UntilMultiPartFormData` does the parsing, it should store the parts as an attribute, and other code should probe for the attribute first.
Trying to read the source again should correctly yield EOF, like it does for forms fields and `getParameters()`.

gregw:
@sbordet please take time to understand before you -1.  Currently in too many places we are taking back to blocking... Which is so stupid after making so the effort to make our async so much better.     By all means give feed back, but hold the veto until I've been able to answer your questions.

gregw:
> @sbordet wrote:
> Don't like this because it makes `MultiPartFormData` non-reusable

`MultiPartFormData` is a `CompletableFuture`, so it has never been reusable.

> and it is also weird because a constructor dependency typically implies "I need this component to work", while here the source it's not necessary at all

What I find weird is that it contains code to efficiently/asynchronously consume input from a `Content.Source` in the class, but then that is not actually used and instead we have hard coded blocking IO code or other `Content.Source` code that does the reading, creates fake chunks and then calls `parse(Chunk)` directly.    Either the class is an IO class or it is not.  It should not be used by code that is doing it's own IO.   Either we have just a pure parser, or we always let the class do the IO.    Since we have some great utility methods to turn other input sources into `Content.Source`s, then I think that is the best way to go.  The less often we have to write the same IO loops is a good thing.

> it is only necessary when parsing it.
 
That's what `MultiPartFormData` is! A parser!   It is the `MultiPartFormData.Parts` class that is the thing used afterwards which is not-a parser and doesn't contain IO code.   Perhaps it is a little strange that the parser class is not called "XxxParser" and contains the end product class.... but that predates this PR and I'm open to renaming and restructuring that if it makes sense.

> It is also confusing when using `MultiPartFormData` to parse chunks, as it would happen in the client. What would be the source passed to the constructor used for?

Nope, it is the otherway around.  It is confusing that we have code that is directly managing chunks to be parsed.  We already have an abstraction for a sequences of `Chunks`, it is the `Content.Source` and we should just use that rather than inventing other ways to feed in chunks.

>  `MultiPartBytesRanges` is critically used in `ResourceService` to handle ranged responses. It will also be used on the client to parse those, so it's not true that it is not used.

See I missed that because it was using the class in a strange way.  It was not calling `parse(Content.Source)`.

I'm not saying that we get rid of the class. I'm saying that we factor out the duplicate code with `MultiPartFormData` and we avoid having it be used in two different modes.

> > The usage of MultiPartFormData is in blocking style any bypasses the Content.Source
> Not sure what you mean, as all `MultiPartFormData` APIs are non-blocking?

The code that is reading an InputStream, converting it to Chunks and then calling `parse(Chunk)` is blocking.  OK it will always be blocking because it is an `InputStream`, but it should just use `InputStreamContentSource` and avoid duplicating the IO code.

> I'm pretty much -1 on this PR, I don't see the savings, the two classes used in different contexts, 

Hey - as I said before. No veto until you've understood and given me a chance to answer your questions.  Besides, it is an experiment tagged for 12.0.1, so no need to decide either way now.   

> and while there is a little duplication, I don't think it's worth an abstract class, also considering that the current duplication may not be a duplication anymore if other parameters are added -- for example max parse length, etc.

The IO code is duplicated in both classes, and then not used in several places where more IO code is written, duplicating either the same IO code again or other utility IO code that we have.  Perhaps the IO code should not be in either class? but either way, we have far too much duplication of very similar code that is absolutely crucial to both performance and correctness to get right.  Having that code in well maintained utility or base classes is a good thing to do.     Managing the whole conversation between turning a `Content.Source` into a `CompletableFuture` via some kind of parser is a pattern repeated several times (even just creating a UTF8 string) and can benefit from having a single implementation.   Perhaps this PR is not the right abstraction.... that's why it is labelled experimental and tagged for 12.0.1.

> I also don't see your point about double parsing? If `DelayedHandler.UntilMultiPartFormData` does the parsing, it should store the parts as an attribute, and other code should probe for the attribute first. Trying to read the source again should correctly yield EOF, like it does for forms fields and `getParameters()`.

OK I missed that `ServletMultiPartFormData` is doing the attribute lookup for the `MultiPartFormData`, so in that case it is would not double parse if the delayed handler (or similar) was used.   But that is the wrong class to put that code in (hence I didn't find it).   It should be in `MultiPartFormData`.    Also we have a problem of how to get the multipart config to any delay style handler, so that it is correctly used if we asynchronously parse before calling the servlet.  We need to solve this problem... somewhat orthogonal, but related.

sbordet:
`MultiPartFormData` was supposed to be a "container" class for both parsing and generation, like `MultiPart` is.

Agreed on single-use of `MultiPartFormData` -- it should be used on-the-fly, so passing the source to constructor is fine, or perhaps even going for `static` if we introduce a `Parser` class.

I'm not opposed to having `MultiPartFormData` as a "container" class and inner classes `Parser` and `ContentSource` (for generation) and `Parts` for results.

Agreed that `parse(Chunk)` is wrong -- I think it was there to be used by the client, but we have since converted the client to a pull model as well, so rather than `onContent(Chunk)` now has `onContentSource(Content.Source)`, which is symmetrical to the server.

> See I missed that because it was using the class in a strange way. It was not calling parse(Content.Source).

It's not using it in a strange way 😃 
The server does not parse `MultiPartByteRanges` (typically), it generates them, so it is natural that it uses `MultiPartByteRanges.ContentSource` which is the generator.
The client should parse `MultiPartByteRanges` but we apparently don't have tests with a clear client usage.

About the veto, I think I'm using -1 too liberally as in "I don't agree with this PR, but open to discussion" -- I'll use -0 in future.

> It [putting `Parts` as a request attribute] should be in MultiPartFormData. 

I disagree. `MultiPartFormData` is used by the client too (as well as `MultiPartByteRanges`) so it should not have anything related to server-side specific stuff.
I maintain that `ServletMultiPartFormData` is the right place because it's server-side and Servlet specific.
For server-side core APIs, IIRC we settled on the fact that the application must do form-data parsing explicitly because we don't provide explicit APIs for parts (while Servlets do), so there is always this pseudocode for those apps:

```java
if (content-type is multipart/form-data) {
  try lookup attribute
  else explicit parsing
} else {
  // some other request type
}
```

If this style is confirmed, then we either have a `ServerMultiPartFormData` for core where we store the attribute, or we document the single line necessary to look up the attribute, which I think is simpler, at least for now.

On the client, we don't need the attribute as nothing parses multipart if not the application itself.

Accessing the Servlet multipart config from a delayed handler I think would be hard... not a big fan of delaying TBH (I prefer lazy for this case).
If `DelayedHandler.UntilMultiPartFormData` can exist after `ServletContextHandler` great (would be able to access Servlet configuration), otherwise I think I would prefer to remove `DelayedHandler.UntilMultiPartFormData`, possibly to reintroduce it later after more thinking.

gregw:
> `MultiPartFormData` was supposed to be a "container" class for both parsing and generation, like `MultiPart` is.

There is something a bit strange, or at least different in how we have a class that is a container, a parser and a generator.    We might needs some restructuring here, or perhaps just some renaming? or maybe a bit of both.

However, that might be for another PR.  I started this PR because of the duplication I saw in the IO patterns.   Thus my main goal is to generalise the IO that is done around these types of classes.     Once that is done and agreed, I think we can look to see if this PR should also include some restructuring/renaming or if that is separate.

> Agreed on single-use of `MultiPartFormData` -- it should be used on-the-fly, so passing the source to constructor is fine, or perhaps even going for `static` if we introduce a `Parser` class.

Or is the CompletableFuture abstraction something separate that just happens to use the parser to provide a container in the future?


> I'm not opposed to having `MultiPartFormData` as a "container" class and inner classes `Parser` and `ContentSource` (for generation) and `Parts` for results.
> 
> Agreed that `parse(Chunk)` is wrong -- I think it was there to be used by the client, but we have since converted the client to a pull model as well, so rather than `onContent(Chunk)` now has `onContentSource(Content.Source)`, which is symmetrical to the server.
> 
> > See I missed that because it was using the class in a strange way. It was not calling parse(Content.Source).
> 
> It's not using it in a strange way smiley The server does not parse `MultiPartByteRanges` (typically), it generates them, so it is natural that it uses `MultiPartByteRanges.ContentSource` which is the generator. The client should parse `MultiPartByteRanges` but we apparently don't have tests with a clear client usage.

OK - bad choice of words.  But see above, it can be a surprise to see a class that looks like it is primarily a parser, but then discover it is primarily used as a generator.  Elsewhere we are very explicit with the Parser/Generator naming.   We have a different pattern here, without clarity or consistency.  The pattern might be fine, but we need better consistency to help with clarity.

> About the veto, I think I'm using -1 too liberally as in "I don't agree with this PR, but open to discussion" -- I'll use -0 in future.

Definitely too liberal with the -1's.   I also think that a draft PR of an experimental idea targeted for 2 or 3 releases in the future also doesn't need a -0.  This PR or anything that comes from it is weeks if not months away from needing any ± decisions. Questions, dislikes, clarifications are all worthwhile at this stage, but a -0 helps little with a draft of an experiment.


> > It [putting `Parts` as a request attribute] should be in MultiPartFormData.
> 
> I disagree. `MultiPartFormData` is used by the client too (as well as `MultiPartByteRanges`) so it should not have anything related to server-side specific stuff. I maintain that `ServletMultiPartFormData` is the right place because it's server-side and Servlet specific. 

Yes and no.  Anything that moves to servlet land immediately needs at least 2 versions of it, so trying to keep things core is worthwhile.   But then the config class itself it a servlet API class.   So we need an simple/elegant way for a core handler to suck servlet specific configuration out of a context in a typeless way.     Currently no good ideas for this, so 2 API dependent versions might be the best.
 
> Accessing the Servlet multipart config from a delayed handler I think would be hard... not a big fan of delaying TBH (I prefer lazy for this case). If `DelayedHandler.UntilMultiPartFormData` can exist after `ServletContextHandler` great (would be able to access Servlet configuration), otherwise I think I would prefer to remove `DelayedHandler.UntilMultiPartFormData`, possibly to reintroduce it later after more thinking.

Lazy is not good enough.   Once we get to servlet-land, then we are naturally blocking. We can go back to async, but not without dispatch type changes that makes anything we do less transparent.

I'm not a big fan of the DelayedHandler in its current form, as I think it is trying to do too much for too many different situations.     I think we should just give up on the whole delay dispatch until the first data optimisation.. or at last just have a handler dedicated to just that.       What this PR is about is ensuring that we can delay dispatch until the entire content is asynchronously consumed (and written to temporary files etc.) before invoking the servlet.  Perhaps we should call this "pre-parsing" or "pre-loading" rather than "delaying"? 

Pre-loading is best done in a core handler before hitting the `ServletHandler`.  Now, after recent changes, it can be done after the `ServletContextHandler`, so the servlet itself will be known along with any multipart configuration.   
This has always been the target of my long game for ee10: to do as much as possible in core handlers before hitting the `ServletHandler`.  Thus we now now core sessions and core security and we can wrap core request/responses after the `ServletContextHandler`.   All that work has been done precisely so we can do fast simple handling in core handlers for things like asynchronous reading of multipart.    Hey eventually I want to get rid of the blocking code that invokes the `ResourceHandler` from the `DefaultServlet`, so that in the same way the `ResourceHandler` can asynchronously serve static content if it has not been filtered/wrapped. 

But there is no hurry on this stuff, it is all internal and can be done after 12.0.0.  I only worked on it yesterday and today as I had 4 hours on a train and didn't want to work on anything more critical.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceCompletableFuture.java
```diff
@@ -0,0 +1,91 @@
...
+
+import java.io.EOFException;
+import java.util.concurrent.CompletableFuture;
+
+import org.eclipse.jetty.io.Content;
+
+/**
+ *
+ */
+public abstract class ContentSourceCompletableFuture<X> extends CompletableFuture<X>
```
sbordet, line 54: I think this class should be either package private or moved to internal.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
```diff
@@ -409,251 +182,445 @@ protected HttpFields customizePartHeaders(MultiPart.Part part)
...
+
+        // Only used for testing.
+        int getPartsSize()
+        {
+            return listener.getPartsSize();
+        }
+
+        private class PartsListener extends MultiPart.AbstractPartsListener
+        {
+            private final AutoLock lock = new AutoLock(); // TODO why do we need this lock?
```
sbordet: We need the lock because `fail()` can be called at any time asynchronously from the parsing.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/AsyncContent.java
```diff
@@ -79,7 +79,7 @@ public void write(boolean last, ByteBuffer byteBuffer, Callback callback)
      * or succeeded if and only if the chunk is terminal, as non-terminal
      * chunks have to bind the succeeding of the callback to their release.
      */
-    private void offer(Content.Chunk chunk)
+    protected void offer(Content.Chunk chunk)
```
sbordet: I remember @lorban and I discussing whether to expose this method, and IIRC we made it `private` because we did not want users to be able to add other implementations of `Chunk`, mess with the `last`-ness and mess with the reference counting.

The implementation seems to have some relic of the historical implementation, when all types of chunk were allowed and can be cleaned up.

I would make it `private` again, as it is only used in a couple of tests.

sbordet (changes requested):
Looks good.
Needs `MultiPartByteRanges` to be updated in the same style.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
gregw: But can it be now?  Failures come via source.read().  If somebody externally fails the CompletableFuture, then the atomic nature of that helps with memory barriers and making things only happen once.   What exactly is the scenario that two threads can be in the parser listener at the same time?

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceCompletableFuture.java
gregw: It is in jetty-io and used in jetty-http.  I think it is a genuinely useful utility class, along with the other adaptor classes we have for Content.Source in jetty-io.

gregw:
@lorban @sbordet I think this is getting closer to being ready to merge.    I still don't have an `EagerHandler`, but the from method is better setup to support that with servlet configuration now.   I've removed the multipart aspect to the DelayedHandler

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceCompletableFuture.java
```diff
@@ -0,0 +1,130 @@
...
+import org.eclipse.jetty.io.Content;
+
+/**
+ * A utility class to convert content from a {@link Content.Source} to an instance
+ * available via a {@link CompletableFuture}.
+ * <p>
+ * An example usage to asynchronously read UTF-8 content is:
+ * </p>
+ * <pre>
+ *     public static class FutureUtf8String extends ContentSourceCompletableFuture<String>
```
lorban: You need to replace `<` and `>` with `&lt;` and `&gt;` respectively in the template type declaration.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/FormFields.java
```diff
@@ -90,24 +88,28 @@ public static CompletableFuture<Fields> get(Request request)
...
-        Object attr = request.getAttribute(FormFields.class.getName());
+        return from(request, getFormEncodedCharset(request), maxFields, maxLength);
+    }
+
+    public static CompletableFuture<Fields> from(Request request, Charset charset, int maxFields, int maxLength)
+    {
+        return from(request, request, charset, maxFields, maxLength);
+    }
+
+    static CompletableFuture<Fields> from(Content.Source source, Attributes attributes, Charset charset, int maxFields, int maxLength)
```
lorban, line 159: This needs some serious javadoc'ing.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/FormFields.java
```diff
@@ -138,6 +140,7 @@ private static int getRequestAttribute(Request request, String attribute)
 
     public FormFields(Content.Source source, Charset charset, int maxFields, int maxSize)
```
lorban: Since it looks like this class is meant to be instantiated with the `static from*()` methods, maybe the ctor should be private?

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/FormFields.java
```diff
@@ -33,7 +35,7 @@
  * A {@link CompletableFuture} that is completed once a {@code application/x-www-form-urlencoded}
  * content has been parsed asynchronously from the {@link Content.Source}.
  */
-public class FormFields extends CompletableFuture<Fields> implements Runnable
+public class FormFields extends ContentSourceCompletableFuture<Fields>
```
lorban, line 38: All the `static` methods lack javadoc.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartByteRanges.java
```diff
@@ -54,73 +54,8 @@
  */
 public class MultiPartByteRanges extends CompletableFuture<MultiPartByteRanges.Parts>
```
sbordet: Should not be a CF anymore, right? Also the javadoc should be changed.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceCompletableFuture.java
```diff
@@ -0,0 +1,131 @@
...
+
+import org.eclipse.jetty.io.Content;
+
+/**
+ * A utility class to convert content from a {@link Content.Source} to an instance
+ * available via a {@link CompletableFuture}.
+ * <p>
+ * An example usage to asynchronously read UTF-8 content is:
+ * </p>
+ * <pre>
```
sbordet: Please do:

```
<pre>{@code
...
}</pre>
```

In this way you don't have to escape `<` or `>` and the code is readable (no HTML entities in it).
Also, align the code left.

I would rename the class to `CompletableUTF8String`, as `Future` seems more like a `java.util.concurrent.Future`, which is kind of not what it is -- it's more a `CompletableFuture`.

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/FormFieldsTest.java
```diff
@@ -0,0 +1,86 @@
...
+        AsyncContent source = new AsyncContent();
+        Attributes attributes = new Attributes.Mapped();
+        CompletableFuture<Fields> futureFields = FormFields.from(source, attributes, charset, maxFields, maxLength);
+        assertFalse(futureFields.isDone());
+
+        int last = chunks.size() - 1;
+        for (int i = 0; i <= last; i++)
+            source.write(i == last, BufferUtil.toBuffer(chunks.get(i), charset), Callback.NOOP);
+
+        assertTrue(futureFields.isDone());
```
lorban: If my understanding is correct, the current implementation of `AsyncContent` guarantees that when `write()` is called with the `last` flag set to true, `futureFields` will be completed.

But nothing in the `Content.Source` contract provides this guarantee so if we ever change the `AsyncContent` impl, we could make this test flaky without noticing.

I'd add a wait on a latch initialized with `chunks.size()` permits passed as the callback to write().

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/ContentDocs.java
```diff
@@ -146,8 +153,94 @@ public static void testEcho() throws Exception
...
+        Content.Sink.write(source, true, "Three", writeCallback);
+        if (!writeCallback.isDone() || !future.isDone())
+            throw new IllegalStateException("Should be consumed");
+
+        System.err.println(future.get());
+    }
+
+    public static class FutureUtf8String extends ContentSourceCompletableFuture<String>
+    {
+        Utf8StringBuilder builder = new Utf8StringBuilder();
```
lorban: This field should be `private final`.

documentation/jetty-documentation/src/main/java/org/eclipse/jetty/docs/programming/ContentDocs.java
```diff
@@ -146,8 +153,94 @@ public static void testEcho() throws Exception
...
+        if (!writeCallback.isDone() || future.isDone())
+            throw new IllegalStateException("Should be consumed");
+        Content.Sink.write(source, false, "Two", writeCallback);
+        if (!writeCallback.isDone() || future.isDone())
+            throw new IllegalStateException("Should be consumed");
+        Content.Sink.write(source, true, "Three", writeCallback);
+        if (!writeCallback.isDone() || !future.isDone())
+            throw new IllegalStateException("Should be consumed");
+
+        System.err.println(future.get());
```
lorban: Remove that `System.err`

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceCompletableFuture.java
```diff
@@ -0,0 +1,131 @@
...
+     * @throws Throwable Thrown if there is an error parsing
+     */
+    protected abstract X parse(Content.Chunk chunk) throws Throwable;
+
+    /**
+     * @param cause A {@link Content.Chunk#isLast() non-last}
+     *             {@link org.eclipse.jetty.io.Content.Chunk#isFailure(Content.Chunk) failure chunk}
+     * @return True if the chunk can be ignored.
+     */
+    protected boolean ignoreTransientFailure(Throwable cause)
```
sbordet: I'd rename to `onTransientFailure(Throwable)`, as the current name seems more like a command than a query.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceCompletableFuture.java
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Called to parse a {@link org.eclipse.jetty.io.Content.Chunk}
+     * @param chunk The chunk containing content to parse. The chunk will
+     * never be a
+     * {@link org.eclipse.jetty.io.Content.Chunk#isFailure(Content.Chunk) failure chunk}.
+     * @return The parsed {@code X} instance or null if parsing is not yet complete
+     * @throws Throwable Thrown if there is an error parsing
+     */
+    protected abstract X parse(Content.Chunk chunk) throws Throwable;
```
lorban, line 130: That method is supposed to do both parsing and accumulation. I'm tempted to suggest a new name like  `parseAndAccumulate` or simply `accumulate`.

At least, this should be more clearly javadoc'ed.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceCompletableFuture.java
lorban: I would not expose this for now, so I'd prefer to see this method be `private`.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/PreloadFormHandler.java
```diff
@@ -0,0 +1,87 @@
...
+        };
+
+        if (future == null)
+            return super.handle(request, response, callback);
+
+        if (future.isDone())
+        {
+            if (!super.handle(request, response, callback))
+                callback.failed(new IllegalStateException("Not Handled"));
+            return true;
```
sbordet: Can't we just return the `boolean` from `super`?
The client made a request to the wrong URI, so I'd say it's a 404.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/FormFields.java
```diff
@@ -138,6 +140,7 @@ private static int getRequestAttribute(Request request, String attribute)
 
     public FormFields(Content.Source source, Charset charset, int maxFields, int maxSize)
     {
+        super(source);
```
lorban, line 202: This makes the `_source` field useless.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/PreloadFormHandler.java
```diff
@@ -0,0 +1,87 @@
...
+        {
+            case FORM_ENCODED -> FormFields.from(request);
+            case MULTIPART_FORM_DATA -> ServletMultiPartFormData.from(Request.as(request, ServletContextRequest.class).getServletApiRequest(), contentType);
+            default -> null;
+        };
+
+        if (future == null)
+            return super.handle(request, response, callback);
+
+        if (future.isDone())
```
sbordet: I'm not sure it's worth splitting the case for `isDone()`, as the code is exactly identical.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
```diff
@@ -489,72 +489,93 @@ public Collection<Part> getParts() throws IOException, ServletException
...
-            String contentType = getContentType();
-            if (contentType == null || !MimeTypes.Type.MULTIPART_FORM_DATA.is(HttpField.valueParameters(contentType, null)))
-                throw new ServletException("Unsupported Content-Type [%s], expected [%s]".formatted(contentType, MimeTypes.Type.MULTIPART_FORM_DATA.asString()));
-
-            MultipartConfigElement config = (MultipartConfigElement)getAttribute(ServletContextRequest.MULTIPART_CONFIG_ELEMENT);
-            if (config == null)
-                throw new IllegalStateException("No multipart config for servlet");
+            try
+            {
+                CompletableFuture<ServletMultiPartFormData.Parts> futureServletMultiPartFormData = ServletMultiPartFormData.from(this);
```
sbordet, line 494: The wrong content type was producing a ServletException before, now an ISE.
Is this tested by the TCK?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/PreloadFormHandler.java
```diff
@@ -0,0 +1,87 @@
...
+ * @see ServletMultiPartFormData#from(ServletRequest)
+ */
+public class PreloadFormHandler extends Handler.Wrapper
+{
+    public PreloadFormHandler()
+    {
+        this(null);
+    }
+
+    public PreloadFormHandler(Handler handler)
```
lorban: :+1:

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
```diff
@@ -489,72 +489,93 @@ public Collection<Part> getParts() throws IOException, ServletException
...
+                    cause = ee.getCause();
+                else if (t instanceof ServletException se)
+                    cause = se.getCause();
+                else
+                    cause = t;
+
+                if (cause instanceof IOException ioException)
+                    throw ioException;
+
+                throw new ServletException(new BadMessageException("bad multipart", cause));
```
lorban, line 577: Why not directly throwing `BadMessageException`? Wrapping it in `ServletException` to later unwrap it looks strange.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletHolder.java
```diff
@@ -155,6 +155,11 @@ public ServletHolder(Class<? extends Servlet> servlet)
         setHeldClass(servlet);
     }
 
+    public Object getMultipartConfig()
```
sbordet: Why returning `Object`?

Also `getMultiPartConfig()` capital P.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletMultiPartFormData.java
```diff
@@ -49,106 +50,103 @@
...
-     * @return a {@link Parts} object to access the individual {@link Part}s
-     * @throws IOException if reading the request content fails
-     * @see org.eclipse.jetty.server.handler.DelayedHandler
+     * Get future {@link ServletMultiPartFormData.Parts} from a servlet request.
+     * @param servletRequest A servlet request
+     * @return A future {@link ServletMultiPartFormData.Parts}, which may have already been created and/or completed.
+     * @see #from(ServletRequest, String)
      */
-    public static Parts from(ServletApiRequest request) throws IOException
+    public static CompletableFuture<Parts> from(ServletRequest servletRequest)
```
sbordet, line 58: Should not this be `HttpServletRequest`? As parts are HTTP things.

lorban (changes requested):

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/MultiPartServletTest.java
```diff
@@ -93,32 +93,27 @@ public void before() throws Exception
         tmpDirString = tmpDir.toAbsolutePath().toString();
     }
 
-    private void start(HttpServlet servlet) throws Exception
+    private void start(HttpServlet servlet, MultipartConfigElement config, boolean preload) throws Exception
```
sbordet: preload -> eager

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/PreloadFormHandler.java
```diff
@@ -0,0 +1,87 @@
...
+        if (future.isDone())
+        {
+            if (!super.handle(request, response, callback))
+                callback.failed(new IllegalStateException("Not Handled"));
+            return true;
+        }
+
+        future.whenComplete((result, failure) ->
+        {
+            try
```
sbordet: There must be a check on `failure` to not call super if we are already failed.

In case of success, we want to store the parts as an attribute, no?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MultiPartFormData.java
sbordet: It is possible because we return the CF to the application, and they can fail it at any time, and failing the CF must cleanup what's been partially parsed.

sbordet (changes requested):

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceCompletableFuture.java
gregw: I think it just parses... and many parsers have an element of accumulation.  However, an implementation is free to call chunk.retain rather than accumulate.  I'll javadoc that.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceCompletableFuture.java
gregw: The whole point of these changes is to allow us to experiment with transient failures.  
Specifically, it is removing many different read loops that might not be transient failure safe, and replacing them with this loop that is.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
gregw: Because Servlet API says that is what we should throw to allow app to catch and handle.   We wrap a BadMessage to make sure we send a 400 if the app does not catch.
Yes it is cumbersome... and probably fragile with regards to the TCK for failures.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/PreloadFormHandler.java
gregw: No, because we have already returned true.

jetty-core/jetty-io/src/main/java/org/eclipse/jetty/io/content/ContentSourceCompletableFuture.java
lorban: Retaining the chunks is a form of accumulation, but you're right that parsing implicitly implies some sort of accumulation too.

Ok to keep the `parse` name, and a bit of javadoc stating this method is supposed to do some accumulation should be enough.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
gregw: Converted to a ServletException by the `ServletApiRequest`

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletMultiPartFormData.java
gregw: It will have to be-a HttpServletRequest, but just going for the lowest common denominator as elsewhere in the code.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/PreloadFormHandler.java
gregw: result or failure are handled by ServletApiRequest#getParts.  I'll comment.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletHolder.java
gregw: I've added the type, but it is lowercase p to match the Servlet API type now: `MultipartConfigElement`

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/FormFields.java
```diff
@@ -59,43 +59,102 @@ public static Charset getFormEncodedCharset(Request request)
...
+     */
     public static CompletableFuture<Fields> from(Request request, int maxFields, int maxLength)
     {
         return from(request, getFormEncodedCharset(request), maxFields, maxLength);
     }
 
+    /**
+     * Find or create a {@link FormFields} from a {@link Content.Source}.
+     * @param request The {@link Request} in which to look for an existing {@link FormFields} attribute,
+     *                using the classname as the attribute name, else the request is used
```
lorban, line 135: The implementation is not looking for a `FormFields` attribute, is it?

jetty-core/jetty-server/src/test/java/org/eclipse/jetty/server/FormFieldsTest.java
```diff
@@ -61,9 +62,11 @@ public void testFormFields(List<String> chunks, Charset charset, int maxFields,
         assertFalse(futureFields.isDone());
 
         int last = chunks.size() - 1;
+        FutureCallback eof = new FutureCallback();
         for (int i = 0; i <= last; i++)
-            source.write(i == last, BufferUtil.toBuffer(chunks.get(i), charset), Callback.NOOP);
+            source.write(i == last, BufferUtil.toBuffer(chunks.get(i), charset), i == last ? eof : Callback.NOOP);
 
+        assertTrue(eof.isDone());
```
lorban: You're missing:

```java
eof.get(5, TimeUnit.SECONDS);
```

lorban (changes requested):
Almost good to go.

lorban (dismissed):
LGTM

sbordet (approved):

PR #9983, opened 2023-06-28
Implement quality lists for Locales

gregw:
Implement quality lists for Locales that orders known locales before unknown.

### DIFF
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
@@ -20,22 +20,22 @@ import java.net.SocketAddress;
 import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 import java.security.Principal;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
-import java.util.stream.Collectors;
 
 import org.eclipse.jetty.http.CookieCache;
 import org.eclipse.jetty.http.HttpCookie;
-import org.eclipse.jetty.http.HttpField;
 import org.eclipse.jetty.http.HttpFields;
 import org.eclipse.jetty.http.HttpHeader;
 import org.eclipse.jetty.http.HttpScheme;
 import org.eclipse.jetty.http.HttpURI;
 import org.eclipse.jetty.http.MetaData;
+import org.eclipse.jetty.http.MimeTypes;
 import org.eclipse.jetty.http.Trailers;
 import org.eclipse.jetty.io.Content;
 import org.eclipse.jetty.server.internal.HttpChannelState;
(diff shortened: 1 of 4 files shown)


gregw:
@lorban @joakime I spun this out of the NPE PR.

Now we produce a list of all Locales, but we order known ones before unknown.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -452,26 +453,27 @@ static List<Locale> getLocales(Request request)
...
-            String country = "";
-            int dash = language.indexOf('-');
-            if (dash > -1)
+            case 0 -> DEFAULT_LOCALES;
+            case 1 -> List.of(Locale.forLanguageTag(acceptable.get(0)));
+            default ->
             {
-                country = language.substring(dash + 1).trim();
-                language = language.substring(0, dash).trim();
+                List<Locale> locales = acceptable.stream().map(Locale::forLanguageTag).toList();
```
lorban, line 467: Another place where `Locale.forLanguageTag` is called with end-user data.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
```diff
@@ -452,26 +453,27 @@ static List<Locale> getLocales(Request request)
...
-        return acceptable.stream().map(language ->
+        // return sorted list of locals, with known locales in quality order before unknown locales in quality order
+        return switch (acceptable.size())
         {
-            language = HttpField.stripParameters(language);
-            String country = "";
-            int dash = language.indexOf('-');
-            if (dash > -1)
+            case 0 -> DEFAULT_LOCALES;
+            case 1 -> List.of(Locale.forLanguageTag(acceptable.get(0)));
```
lorban, line 464: We cannot call `Locale.forLanguageTag` for any string coming from an end-user as that's a potential CVE.

`Locale.forLanguageTag` will accept any 2+ letters alphabetical string and internally create and permanently store a `Locale` instance, without any limit.

The following test will OOME without any way to recover:
```java
    @Test
    public void testOomeLocale()
    {
        int index = 26; // start at 'AA'
        for (int i = 0; i < Integer.MAX_VALUE; i++)
        {
            String tag = toAlpha(index + i);
            Locale.forLanguageTag(tag);
        }
    }

    private static String toAlpha(int cp)
    {
        int div = cp / 26;
        int remainder = cp % 26;
        char letter = (char)((int)'A' + remainder);

        if (div != 0)
            return toAlpha(div - 1) + letter;
        return Character.toString(letter);
    }
```

lorban (changes requested):
I think this change can lead to an exploit. 
_Update: no, it cannot._

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java
gregw: OH!  yes! Good spot, as the javadoc is not clear that caching is happening.    

So perhaps we go back to the impl I had where we just look up in a Trie index.   But we lose the ability to do deprecated name conversion then.
Hmmm must be a non allocating way to access that logic?  standby

lorban: Actually the internal `Locale` cache seems to be using `SoftReference`s, so it's going to stress the GC but won't necessarily OOME.

lorban (approved):

PR #9986, opened 2023-06-28
Added test case to show how to handle zip content in AsyncMiddleManSe…

sbordet:
…rvlet.

### DIFF
--- a/jetty-proxy/src/test/java/org/eclipse/jetty/proxy/AsyncMiddleManServletTest.java
+++ b/jetty-proxy/src/test/java/org/eclipse/jetty/proxy/AsyncMiddleManServletTest.java
@@ -13,6 +13,7 @@
 
 package org.eclipse.jetty.proxy;
 
+import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


jetty-proxy/src/test/java/org/eclipse/jetty/proxy/AsyncMiddleManServletTest.java
```diff
@@ -1551,6 +1556,82 @@ protected ContentTransformer newServerResponseContentTransformer(HttpServletRequ
...
+            {
+                return new AfterContentTransformer()
+                {
+                    @Override
+                    public boolean transform(Source source, Sink sink) throws IOException
+                    {
+                        // Un-zip and re-zip.
+
+                        ZipInputStream zipIn = new ZipInputStream(source.getInputStream());
+                        ZipOutputStream zipOut = new ZipOutputStream(sink.getOutputStream());
```
gregw, line 1596: Is this the wrong spot to do decompression and compression?  Shoudn't the client compress the data inbound and the server compress the outbound data normally ?

joakime (approved):
This is only a test case, and it looks good to me.

PR #9988, opened 2023-06-28
Add constructors accepting the handler to wrap to all core handler wrappers

lorban:
Add constructors that accept the wrapped handler to all core `Handler.Wrapper` subclasses

This allows the chaining of hander wrappers in a more elegant way; this is how we currently do it:

```java
GzipHandler handler = new GzipHandler();
handler.setHandler(new EchoHandler());
server.setHandler(handler);
```

and this is what this change allows us to write:

```java
server.setHandler(new GzipHandler(new EchoHandler()));
```

### DIFF
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DebugHandler.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DebugHandler.java
@@ -41,6 +41,15 @@ public class DebugHandler extends Handler.Wrapper implements Connection.Listener
     private OutputStream _out;
     private PrintStream _print;
 
+    public DebugHandler()
+    {
+    }
+
+    public DebugHandler(Handler handler)
+    {
+        super(handler);
+    }
+
     @Override
     public boolean handle(Request request, Response response, Callback callback) throws Exception
     {
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DebugHandler.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DebugHandler.java
@@ -43,6 +43,7 @@ public class DebugHandler extends Handler.Wrapper implements Connection.Listener
 
     public DebugHandler()
     {
+        this(null);
     }
 
     public DebugHandler(Handler handler)
--- a/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ShutdownHandler.java
+++ b/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ShutdownHandler.java
@@ -91,7 +91,7 @@ public class ShutdownHandler extends Handler.Wrapper
      */
     public ShutdownHandler(String shutdownToken)
     {
-        this(null, shutdownToken, false, null);
+        this(null, null, shutdownToken, false);
     }
 
     /**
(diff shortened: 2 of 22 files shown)


sbordet:
Seems reasonable.
However, newly introduced parameterless constructors should be implemented by calling `this(null)`, so that if the other constructor is changed, we properly delegate.
Also, I would like the `Handler` parameter to be the first, as it is canonical in Java (subclass has the same list of parameters as superclass, plus eventually others).

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DebugHandler.java
```diff
@@ -41,6 +41,15 @@ public class DebugHandler extends Handler.Wrapper implements Connection.Listener
     private OutputStream _out;
     private PrintStream _print;
 
+    public DebugHandler()
+    {
```
sbordet, line 45: I'd prefer this to call `this(null)`.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ShutdownHandler.java
```diff
@@ -111,9 +111,11 @@ public ShutdownHandler(String shutdownToken, boolean exitJVM)
      * @param shutdownPath the path to respond to shutdown requests against (default is "{@code /shutdown}")
      * @param shutdownToken a secret password to avoid unauthorized shutdown attempts
      * @param exitJVM If true, when the shutdown is executed, the handler class System.exit()
+     * @param handler the handler to wrap
      */
-    public ShutdownHandler(String shutdownPath, String shutdownToken, boolean exitJVM)
+    public ShutdownHandler(String shutdownPath, String shutdownToken, boolean exitJVM, Handler handler)
```
sbordet: I'd like `Handler` as first parameter better.

lorban: I've put that parameter at the end because `MinimumDataRateHandler` and `Handler.Wrapper` already accepts the handler to wrap as the last ctor parameter.

But I'm not against changing this ordering.

jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/DebugHandler.java
lorban: Ok.

janbartel:
I have a little bit of a doubt after the very long history of (at least some) handler constructors taking the parent handler as the argument, whereas this flips it around and passes the child. But I do think it looks neat. Maybe we solve that with very good documentation.

lorban:
@janbartel this PR is a followup to https://github.com/eclipse/jetty.project/pull/9948 that changed all handlers to only take the child handler as an argument as it was decided in https://github.com/eclipse/jetty.project/issues/9946.

Said differently, passing the child handler to the constructor of a `Handler.Wrapper` is the new standard.

janbartel:
@lorban hhhmm'ok. So you're saying that if I look in the documentation I'll find this difference with foregoing versions of jetty clearly spelt out? :)

lorban:
@janbartel now you should :-)

gregw (commented):
I'm +1 on this once it actually compiles.

gregw (approved):

lachlan-roberts (approved):

PR #9989, opened 2023-06-28
Fix Path to URL conversion for windows users

joakime:
Fixing UrlResourceFactoryTest to convert from Path to URL properly for users building on Windows.


lorban (approved):
Of course, Windows `file:` URIs.

Issue #9973, opened 2023-06-26
Creating a Resource for an entry in a nested jar file in Jetty 12


wilkinsona:
**Jetty version**

Jetty 12.0.0.Beta2

**Java version**

17

**Question**

I'm working on upgrading Spring Boot to Jetty 12. Spring Boot supports executable jars and wars that can be run using `java -jar`. This results in URLs pointing to jars nested within a jar or war. For example, this URL taken from our test suite:

```
jar:file:/Users/awilkinson/dev/spring-projects/spring-boot/main/spring-boot-tests/spring-boot-integration-tests/spring-boot-server-tests/build/spring-boot-server-tests-app/build/libs/spring-boot-server-tests-app-jetty.war!/WEB-INF/lib/spring-boot-server-tests-app-resources.jar!/
```

Much of the path doesn't matter. A shorter example would be:

```
jar:file:/example.war!/WEB-INF/lib/spring-boot-server-tests-app-resources.jar!/
```

With Jetty 11 and earlier, I could create a `Resource` that points to the `META-INF/resources` entry within `spring-boot-server-tests-app-resources.jar!` by calling `Resource.newResource(url + "META-INF/resources")`. I am struggling to find the Jetty 12 equivalent of this code.

I have a `WebAppContext` to hand so I have tried calling `ResourceFactory resourceFactory = ResourceFactory.of(handler)` and calling `ResourceFactory.newResource`. This works for situations where `spring-boot-server-tests-app-resources.jar` is available directly on the file system but not for the case where it's nested within a war.

How can I get this working for nested jars using Jetty 12 please?

joakime:
Quick answer / example:

Create/Open `Resources` via the `ResourceFactory`, but be aware of resource management.

``` java
Path exampleJar = Path.of("example.jar");

try (ResourceFactory.Closeable resourceFactory = ResourceFactory.closeable())
{
    // Create Resource from jar/zip, mounted to root of jar/zip
    Resource jarFileResource = resourceFactory.newJarFileResource(exampleJar.toUri());

    // access as Resource
    Resource manifestResource = jarFileResource.resolve("/META-INF/MANIFEST.MF");
    try (InputStream inputStream = manifestResource.newInputStream())
    {
        // read from input stream
    }

    // Create Resource from jar/zip URI
    URI uri = URI.create("jar:file:/example.jar!/");
    Resource uriResource = resourceFactory.newResource(uri);
    // access as Path object
[...]

PR #9991, opened 2023-06-28
#9973 prototype JarURLResourceFactory

lorban:
Example `ResourceFactory` alternative for https://github.com/eclipse/jetty.project/issues/9973

Not to be merged.

### DIFF
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/JarURLResourceFactory.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/JarURLResourceFactory.java
@@ -1,0 +1,274 @@
+//
+// ========================================================================
+// Copyright (c) 1995 Mort Bay Consulting Pty Ltd and others.
+//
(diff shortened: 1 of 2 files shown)


joakime:
No, this is unnecessary, and I don't like it. (this introduces deprecated code usage in the JVM and a host of bugs around zip-slip).

The URLResourceFactory is sufficient and should be able to function.

It also cannot handle the use case in #9973 properly, both because JarFile doesn't handle nested jars, and as they have a custom URLStreamHandler to handle the nested jar behavior.

lorban:
This PR is just meant to show how the old Jetty 11 `JarURLResourceFactory` could look like if it were to be revived, but I'm definitely not arguing to include it in the Jetty codebase.

If this can help the Spring team, they're welcome to adopt this code in their codebase and maintain it, until they come up with a `java.nio.file.FileSystem` implementation that supports their nested jars requirements.

joakime:
Issue #9973 addressed entirely within URLResourceFactory

PR #9995, opened 2023-06-29
Issue #9973 - Unwrap URI.scheme-specific-parts until we find a path we can resolve against.

### LINKED ISSUE #9973: Creating a Resource for an entry in a nested jar file in Jetty 12

joakime:
URLResourceFactory will unwrap opaque URIs (eg: `jar:file:///example.jar!/`) until it finds a URI with a path it can apply the `resolve(String)` against, and the rebuild the URI.

Not very efficient, but it is URI correct behavior.

### DIFF
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/URLResourceFactory.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/URLResourceFactory.java
@@ -25,6 +25,7 @@ import java.nio.file.Path;
 import java.time.Instant;
 
 import org.eclipse.jetty.util.FileID;
+import org.eclipse.jetty.util.URIUtil;
 
 /**
  * {@link ResourceFactory} for {@link java.net.URL} based resources.
@@ -122,7 +123,7 @@ public class URLResourceFactory implements ResourceFactory
         @Override
         public URI getURI()
         {
-            return uri;
+            return URIUtil.correctFileURI(uri);
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/URLResourceFactory.java
```diff
@@ -151,6 +152,20 @@ public Resource resolve(String subUriPath)
             }
         }
 
+        // This could probably live in URIUtil, but it's awefully specific to URLResourceFactory.
+        private static URI resolve(URI parent, String path)
+        {
+            if (parent.getPath() == null)
```
joakime: might want to use URI.isOpaque() here.  hmmm.

lorban (dismissed):

lorban (approved):

Issue #10057, opened 2023-07-03
Jetty12 URLResource efficiency


gregw:
**Jetty version(s)**
12

**Enhancement Description**

The current jetty-12 implementation of `URLResourceFactory` creates a very inefficient resource that will create many `URLConnection`s and not retain them.   We should revert to the jetty<12 technique of caching the latest connection in the resource an reusing it when possible.      Even though `URL` and `URLResource` are being replaced, there will still be custom usages of them for some time.

This is raised from discussion in #9973

PR #10058, opened 2023-07-03
Issue #9973 - use URIUtil.addPaths() instead of URI.resolve()

### LINKED ISSUE #9973: Creating a Resource for an entry in a nested jar file in Jetty 12

joakime:


gregw (changes requested):
Whilst this individual fix looks good, we need to address #10057 and make the URLResource a first class citizen again.   Please compare against the version in jetty-11 and make sure that it is as compatible as possible.

Also it would be good to see if we could get a test that was closer to the spring-boot usage and maybe even use a custom URLStreamHandler?

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/URLResourceFactory.java
```diff
@@ -62,18 +90,39 @@ public void setUseCaches(boolean useCaches)
...
+                }
+            }
         }
+        return urlResource;
     }
 
-    private static class URLResource extends Resource
+    private void cleanupReferences()
     {
+        // TODO
```
gregw: We should probably do the cleanup before 12.0.0, but perhaps not for beta3.
I was thinking it should be like the JVM's LocaleObjectCache, in that the factory should keep a list of soft references to all resources given out and that we always clean the stale objects on any new resource (which will call close) and then if the factory is closed, we should close all remaining resources, even if still referenced.

gregw (dismissed):
There are still differences with jetty-11, for example, isDurectory checks for existence as well as the trailing '/'.  `getName` is also different?

Please open an issue for the cleanup 

Other than that looks good.

gregw:
We should probably do the cleanup before 12.0.0, but perhaps not for beta3.
I was thinking it should be like the JVM's LocaleObjectCache, in that the factory should keep a list of soft references to all resources given out and that we always clean the stale objects on any new resource (which will call close) and then if the factory is closed, we should close all remaining resources, even if still referenced.

joakime:
This branch makes the spring-boot build green.

The spring-boot branch I'm working with can be found at
https://github.com/joakime/spring-boot-jetty-12/tree/jetty-12-joakim
(complete with code changes to spring-boot)

lorban (approved):

lorban:
Note that this also solves https://github.com/eclipse/jetty.project/issues/10057

PR #10060, opened 2023-07-04
DefaultServlet combinations test

lorban:
Add tests for `DefaultServlet` combinations, and remove the `pathInfoOnly` parameter of `DefaultServlet`, replacing it with some automatic detection based on the servlet mapping.

See #9910


jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
```diff
@@ -0,0 +1,191 @@
...
+        context.setWelcomeFiles(new String[]{"index.html", "index.welcome"});
+
+        ServletHolder welcomeExtHolder = context.addServlet(WelcomeServlet.class, "*.welcome");
+        welcomeExtHolder.setInitParameter("mapping", "welcome extension");
+    }
+
+    private void startServer(boolean pathInfoOnly, ResourceService.WelcomeMode welcomeMode) throws Exception
+    {
+        ServletHolder defaultHolder = context.addServlet(DefaultServlet.class, "/");
+        defaultHolder.setInitParameter("pathInfoOnly", String.valueOf(pathInfoOnly));
```
gregw: This is not really testing the pathInfoOnly, because there is no servletPath.   
If pathInfoOnly is true, then the DefaultServlet needs to be registered at something like "/static/*" and all the URIs from the test need to include that prefix.

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
```diff
@@ -0,0 +1,191 @@
...
+
+    public static Stream<Data> data()
+    {
+        return Stream.of(
+            new Data(false, SERVE, "/", HttpStatus.OK_200, "Static index.html at root"),
+            new Data(false, REDIRECT, "/", HttpStatus.FOUND_302, "http://local/index.html"),
+            new Data(false, REHANDLE, "/", HttpStatus.OK_200, "Static index.html at root"),
+            new Data(true, SERVE, "/", HttpStatus.FOUND_302, "http://local/index.html/"),
+            new Data(true, REDIRECT, "/", HttpStatus.FOUND_302, "http://local/index.html"),
+            new Data(true, REHANDLE, "/", HttpStatus.FOUND_302, "http://local/index.html/"),
```
gregw:
```suggestion
            new Data(true, SERVE, "/static/", HttpStatus.FOUND_302, "http://local/index.html/"),
            new Data(true, REDIRECT, "/static/", HttpStatus.FOUND_302, "http://local/index.html"),
            new Data(true, REHANDLE, "/static/", HttpStatus.FOUND_302, "http://local/index.html/"),
```

You can do that explicitly here or perhaps prefix the "/static" in the Data constructor if pathInfoOnly is true?

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    record Data(boolean pathInfoOnly, ResourceService.WelcomeMode welcomeMode, String requestPath, int expectedStatus, String expected)
+    {
+    }
+
+    public static Stream<Data> data()
+    {
+        return Stream.of(
+            new Data(false, SERVE, "/", HttpStatus.OK_200, "Static index.html at root"),
```
gregw: I don't like flattened lists like this as it is impossible to tell if all combinations have been handled, I'd much rather see:
```javav
    for(boolean pathInfoOnly : new boolean[]{ true, false})
        for (WelcomeMode m : new WelcomeMode[]{REDIRECT, REHANDLE, SERVE})
           for (String uri : .... etc.)
```

I know this means you can't list out the expected results, but you can write a little expert system to produce the expected result... i.e a few if elses 

```

File: jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java [hunk 4]
```diff
@@ -0,0 +1,191 @@
...
+    public void init() throws Exception
+    {
+        docRoot = workDir.getEmptyPathDir().resolve("docroot");
+        FS.ensureDirExists(docRoot);
+        Path subdirPath = docRoot.resolve("subdir");
+        FS.ensureDirExists(subdirPath);
+        Files.writeString(docRoot.resolve("index.html"), "Static index.html at root", UTF_8);
+        Files.writeString(docRoot.resolve("foo.welcome"), "Static foo.welcome at root", UTF_8);
+        Files.writeString(subdirPath.resolve("index.html"), "Static index.html at subdir", UTF_8);
+        Files.writeString(subdirPath.resolve("foo.welcome"), "Static foo.welcome at subdir", UTF_8);
```
gregw: Can we have subdirs that have an index.html, but not a foo.welcome file and vice versa, also a subdir with no welcome files.

Remember we need to test the combinations: no file & no servlet; no file & servlet; file & no servlet; file & servlet

It would also be good to test the fact that a welcome with a file takes priority over a welcome servlet.

gregw (changes requested):
good start... but....

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
```diff
@@ -0,0 +1,191 @@
...
+            new Data(true, SERVE, "/subdir/foo.welcome", HttpStatus.OK_200, "Servlet at welcome extension"),
+            new Data(true, REDIRECT, "/subdir/foo.welcome", HttpStatus.OK_200, "Servlet at welcome extension"),
+            new Data(true, REHANDLE, "/subdir/foo.welcome", HttpStatus.OK_200, "Servlet at welcome extension"),
+
+            new Data(false, SERVE, "/nothing/index.welcome", HttpStatus.OK_200, "Servlet at welcome extension"),
+            new Data(false, REDIRECT, "/nothing/index.welcome", HttpStatus.OK_200, "Servlet at welcome extension"),
+            new Data(false, REHANDLE, "/nothing/index.welcome", HttpStatus.OK_200, "Servlet at welcome extension"),
+            new Data(true, SERVE, "/nothing/index.welcome", HttpStatus.OK_200, "Servlet at welcome extension"),
+            new Data(true, REDIRECT, "/nothing/index.welcome", HttpStatus.OK_200, "Servlet at welcome extension"),
+            new Data(true, REHANDLE, "/nothing/index.welcome", HttpStatus.OK_200, "Servlet at welcome extension"),
```
janbartel: Where's the case where we do a `REDIRECT` to the `WelcomeServlet`?

janbartel (changes requested):
I think there might be a few use-cases left out.

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
```diff
@@ -0,0 +1,191 @@
...
+            new Data(true, REHANDLE, "/subdir/", HttpStatus.FOUND_302, "http://local/subdir/index.html/"),
+
+            new Data(false, SERVE, "/subdir/foo.welcome", HttpStatus.OK_200, "Servlet at welcome extension"),
+            new Data(false, REDIRECT, "/subdir/foo.welcome", HttpStatus.OK_200, "Servlet at welcome extension"),
+            new Data(false, REHANDLE, "/subdir/foo.welcome", HttpStatus.OK_200, "Servlet at welcome extension"),
+            new Data(true, SERVE, "/subdir/foo.welcome", HttpStatus.OK_200, "Servlet at welcome extension"),
+            new Data(true, REDIRECT, "/subdir/foo.welcome", HttpStatus.OK_200, "Servlet at welcome extension"),
+            new Data(true, REHANDLE, "/subdir/foo.welcome", HttpStatus.OK_200, "Servlet at welcome extension"),
+
+            new Data(false, SERVE, "/nothing/index.welcome", HttpStatus.OK_200, "Servlet at welcome extension"),
```
gregw: as well as "/nothing/", which is a directory that does not exist, we need "/empty/" which is what it says it is!

gregw (changes requested):

gregw:
@lorban status?

lorban:
@gregw I've added the missing cases, and they are now generated. I also need to add some tests for when servlet include is used, but this suite already is already large enough to be confusing, so that'll come in a 2nd step.

Some results don't look right:
 - the `REDIRECT` to the `WelcomeServlet` case @janbartel speaks about should be covered by `/static/subdirWelcome/` but I don't think it should get a 403. 
 - the `/static/` and `/static/subdirHtml/` `REDIRECT` when `pathInfoOnly` is `true` redirect to a surprising location.

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
```diff
@@ -0,0 +1,258 @@
...
+                {
+                    int expectedStatus;
+                    String expected;
+
+                    switch (requestPath)
+                    {
+                        case "/foo.welcome" ->
+                        {
+                            expectedStatus = HttpStatus.OK_200;
+                            expected = "Servlet at welcome extension";
```
gregw: good

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
```diff
@@ -0,0 +1,258 @@
...
+                        }
+                        case "/static/foo.welcome" ->
+                        {
+                            expectedStatus = HttpStatus.OK_200;
+                            expected = pathInfoOnly ? "Static foo.welcome at root" : "Static foo.welcome at static";
```
gregw: good!  nicely clear what pathInfoOnly is doing

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
```diff
@@ -0,0 +1,258 @@
...
+                        }
+                        case "/static/" ->
+                        {
+                            if (welcomeMode == REDIRECT)
+                            {
+                                expectedStatus = HttpStatus.FOUND_302;
+                                expected = pathInfoOnly ? "http://local/static/static/index.html" : "http://local/static/index.html";
```
gregw: The pathInfoOnly location is nonsensical as it contain `/static/static`. So the pathInfo in this case is "/" and the found welcome is thus docroot + "/" + "index.html", so we need to redirect to "http://local/static/index.html"

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
```diff
@@ -0,0 +1,258 @@
...
+        context.setContextPath("/");
+        context.setWelcomeFiles(new String[]{"index.html", "index.welcome"});
+
+        ServletHolder welcomeExtHolder = context.addServlet(WelcomeServlet.class, "*.welcome");
+        welcomeExtHolder.setInitParameter("mapping", "welcome extension");
+    }
+
+    private void startServer(boolean pathInfoOnly, ResourceService.WelcomeMode welcomeMode) throws Exception
+    {
+        ServletHolder defaultHolder = context.addServlet(DefaultServlet.class, "/static/*");
```
gregw: I think this should be:
```suggestion
        ServletHolder defaultHolder = context.addServlet(DefaultServlet.class, pathInfoOnly ? "/static/*" : "/");
```

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
```diff
@@ -0,0 +1,258 @@
...
+        FS.ensureDirExists(subdirEmptyPath);
+
+        server = new Server();
+
+        connector = new LocalConnector(server);
+        connector.getConnectionFactory(HttpConfiguration.ConnectionFactory.class).getHttpConfiguration().setSendServerVersion(false);
+
+        context = new ServletContextHandler();
+        context.setBaseResourceAsPath(docRoot);
+        context.setContextPath("/");
```
gregw: let's have a real context to check we are not leaving it out anywhere:
```suggestion
        context.setContextPath("/ctx");
```

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
```diff
@@ -0,0 +1,258 @@
...
+                        {
+                            if (welcomeMode == REDIRECT)
+                            {
+                                expectedStatus = HttpStatus.FOUND_302;
+                                expected = pathInfoOnly ? "http://local/static/static/index.html" : "http://local/static/index.html";
+                            }
+                            else
+                            {
+                                expectedStatus = HttpStatus.OK_200;
+                                expected = pathInfoOnly ? "Static index.html at root" : "Static index.html at static";
```
gregw:
```suggestion
                                expected = "Static index.html at static";
```

gregw (changes requested):
This is really good..., but more necessary!

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
lorban: Done.

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
lorban: This would confuse the combinations, wouldn't it?
For instance, `/static/foo.welcome` would change from:

```java
expected = pathInfoOnly ? "Static foo.welcome at root" : "Servlet at welcome extension";
```
to:
```java
expected = pathInfoOnly ? "Static foo.welcome at root" : "Static foo.welcome at static";
```

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
lorban: This is a bug in DefaultServlet, let me fix it.

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
lorban: In my understanding, when `pathInfoOnly == true`, the servlet path (`/static`) is supposed to be stripped from the path, so `docroot/index.html` should be served instead of `docroot/static/index.html`.

Or did I misunderstand?

gregw: No - I did!

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
gregw: Ah, I see, you actually have 2 DefaultServlet instances, one that really is a default servlet always at "/" and another that is configured at "/static/*"

I think the test should be changed to only ever have 1 `DefaultServlet`, which is at "/" when `pathInfoOnly` is false. otherwise it is at "/static/*" and some error servlet should be at "/".

Then there needs to be no tests that explicitly have "/static" in their URI. Instead, if `pathInfoOnly` is true, you should prefix "/static" to all the tests and check all the combinations work if they are not at the root and have pathInfo only.


Hmmm but then we might need another combination.  Testing a non default DefaultServlet, but that has `pathInfoOnly` false!     Ugh sooo many variations!  But that is the point of this unit test.

lorban: Yes, your suggestion would make the combinations more readable. Let me try this...

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
```diff
@@ -117,44 +118,79 @@ public void init() throws Exception
...
             resp.getWriter().print("Servlet at " + mapping);
         }
     }
 
+    public static class TeapotServlet extends HttpServlet
+    {
+        @Override
+        protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException
+        {
+            resp.setStatus(HttpStatus.IM_A_TEAPOT_418);
```
gregw, line 163: Love it!

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/DefaultServletCombinationsTest.java
```diff
@@ -117,44 +118,79 @@ public void init() throws Exception
...
+        {
+            defaultHolder = context.addServlet(DefaultServlet.class, "/static/*");
+            defaultHolder.setInitParameter("pathInfoOnly", "true");
+            context.addServlet(TeapotServlet.class,  "/");
+        }
+        else
+        {
+            defaultHolder = context.addServlet(DefaultServlet.class, "/");
+            defaultHolder.setInitParameter("pathInfoOnly", "false");
+        }
```
gregw: This looks good, but we are still missing the scenario of a `DefaultServelt` at non root path, but with `pathInfoOnly` false.  Not sure this is a good configuration to support?

I wonder if we just get rid of the configuration and always act as `pathInfoOnly`, so if req.getHttpServletMapping().getMappingMatch() == MappingMatch.DEFAULT, then we use the servletPath, otherwise we use the pathInfo.   Is that simpler? Is that sufficient?

lorban: I've added the _DefaultServlet at non root path_ case as an extra test method as I found it confusing to mix these combinations with the _DefaultServlet at root path_ ones.

Eliminating one parameter is tempting, but is your suggestion spec-compliant?

gregw: The spec says nothing about non-default servlet.   It just says the default default servlet should server resources, but may be replaced by an application provided default servlet that may do anything.
It is entirely a jetty feature that we let our `DefaultServlet` be used in non-default ways (It should have been called `ResourceServlet`).

So spec is of no concern here.

Only concern is if there is a valid use-case to have our DefaultServlet` deployed as non-default, but using the full servletPath + pathInfo ???

lorban: You're probably onto something, but that seems to be too large for the scope of this PR. Or this PR should be moved to after the 12.0.0 release.

lorban:
@gregw nudge

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -1227,6 +1195,11 @@ private static boolean isIncluded(HttpServletRequest request)
         return request.getAttribute(RequestDispatcher.INCLUDE_REQUEST_URI) != null;
     }
 
+    private static boolean isPathInfoOnly(HttpServletRequest req)
```
gregw: I think `isDefaultServlet(Request)` or `isDefaultMapping(Request)` would be a clearer name.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
```diff
@@ -442,20 +433,15 @@ protected ServletContextHandler initContextHandler(ServletContext servletContext
...
-    {
-        return _isPathInfoOnly;
-    }
-
     @Override
     protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException
     {
         String includedServletPath = (String)req.getAttribute(RequestDispatcher.INCLUDE_SERVLET_PATH);
         boolean included = includedServletPath != null;
         String encodedPathInContext;
```
gregw: maybe put the whole encodedPathInContext calculation into a protected method, so if somebody really wants something different, they can simply extend it?

gregw (changes requested):
I like it, but a couple of quibbles

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
lorban: `isDefaultMapping()` sounds good but it requires inverting the logic; so
`if (isPathInfoOnly(req))` tests become `if (!isDefaultMapping(req))`.

Let me do it.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/DefaultServlet.java
lorban: Done.

gregw (approved):

PR #10061, opened 2023-07-04
Simplified URLResource cleaner

gregw:
The reference to the InputStream does not need to be soft/weak as a hard reference to it does not prevent the URLResource itself from being collected.  Also by making a runnable record, there is one less allocation per input stream creation.

### DIFF
--- a/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/URLResourceFactory.java
+++ b/jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/URLResourceFactory.java
@@ -24,6 +24,7 @@ import java.nio.channels.Channels;
 import java.nio.channels.ReadableByteChannel;
 import java.nio.file.Path;
 import java.time.Instant;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 
 import org.eclipse.jetty.util.FileID;
@@ -114,7 +115,7 @@ public class URLResourceFactory implements ResourceFactory
         private final int readTimeout;
         private final boolean useCaches;
         private URLConnection connection;
-        private InputStream in = null;
+        private InputStreamReference inputStreamReference = null;
 
         public URLResource(URI uri, int connectTimeout, int readTimeout, boolean useCaches) throws MalformedURLException
         {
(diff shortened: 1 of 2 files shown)


jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/URLResourceFactory.java
```diff
@@ -223,13 +221,7 @@ public boolean exists()
...
                     {
                         in = connection.getInputStream();
-                        Reference<InputStream> ref = new SoftReference<>(in);
-                        CLEANER.register(this, () ->
-                        {
-                            IO.close(ref.get());
-                            if (ON_SWEEP_LISTENER != null)
-                                ON_SWEEP_LISTENER.accept(ref);
-                        });
+                        CLEANER.register(this, new InputStreamCleaner(in));
```
lorban: Rereading this code, I noticed the input stream gets closed if the `URLResource` gets garbage collected.

This means we get this:

```java
Resource webResource = resource.resolve("/web.xml");
assertTrue(webResource.exists());
InputStream inputStream = webResource.newInputStream();
webResource = null;

await().atMost(5, TimeUnit.SECONDS).until(() ->
{
    System.gc();
    return cleanedRefCount.get() > 0;
});

assertThat(IO.toString(inputStream), notNullValue()); //  this throws IOException: Stream closed
```

lorban (changes requested):

jetty-core/jetty-util/src/main/java/org/eclipse/jetty/util/resource/URLResourceFactory.java
gregw: Hmmm perhaps we need to weak reference the input stream itself? 

But it will generally be fine as I don't the reference to the input stream outlives the resource reference often.

Either way, this is better than what is committed.

lorban (dismissed):

gregw:
@lorban thanks for the green tick... But I think I have an idea to simply solve that problem... I'll try tonight....

gregw:
@lorban what about now?

gregw:
@lorban I think the test is a bit flakey.  sometimes it does not complete when I run it from the IDE

gregw:
@lorban I fixed the tests and now test for the inputstream not being closed if it is taken

lorban (approved):
LGTM

PR #10064, opened 2023-07-05
Various Cleanup in ServletChannel

gregw:
Various Cleanup in ServletChannel
 + [x] remove lambdas for clarity
 + [x] TODO Non-blocking error dispatch
 + [x] TODO isHandled does not exist
 + [x] TODO checkAndPrepareUpgrade and implement servlet upgrade?
 + [x] remove unused variables
 + [x] review javadoc (including any warnings)
 + [x] review any compiler or findbug warnings (if any)


gregw:
@lachlan-roberts can you collaborate with me on this one.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -525,16 +519,11 @@ public boolean handle()
...
-                                    // We do not notify ServletRequestListener on this dispatch because it might not
-                                    // be dispatched to an error page, so we delegate this responsibility to the ErrorHandler.
-                                    dispatch(() -> errorHandler.handle(_servletContextRequest, getServletContextResponse(), blocker));
-                                    blocker.block();
-                                }
+                                // We do not notify ServletRequestListener on this dispatch because it might not
+                                // be dispatched to an error page, so we delegate this responsibility to the ErrorHandler.
+                                // We must ignore the callback because the error servlet could go async.
+                                dispatch(() ->
+                                    errorHandler.handle(_servletContextRequest, getServletContextResponse(), Callback.NOOP));
```
gregw: @lachlan-roberts I don't think this going to work... or at least it should not work and if it does it is probably a race or only working because the error page is small and does not congest the connection.

The continuation of the handling loop needs to be delayed until the callback is called. Something like what was suggested in the TODO comment previously: `Callback completeCallback = Callback.from(() -> _state.completed(null), _state::completed);`    although I don't think that is exactly correct.     It might just need a call to get the handling going again... which might need to be executed?  We might also need an optimisation to see if the callback has been completed in the call to errorHandler.handle, in which case we can continue, if not then the callback has to continue handling.

lachlan-roberts: The problem is that in `ErrorHandler` we do:

```java
errorDispatcher.error(httpServletRequest, httpServletResponse);
callback.succeeded();
```

so after callback is succeeded but we could still be asynchronously writing the response with an async dispatch. So in that case of a servlet error page the callback means effectively nothing. So calling `state.Completing()` or `state.completed()` is wrong.

I think this change I made will work for any servlet error page but not a Jetty core one. So maybe we need to do a `scheduleDispatch()` in the callback?

lachlan-roberts: Take a look at the further changes I made to this.

gregw: Yeah, even if we "know" the thing we are calling is currently synchronous, if we are passing a callback, we should strive to call it in correct style.   Your changes are better, but I'll comment inline how to improve.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -584,16 +520,22 @@ public boolean handle()
...
+                                AtomicInteger count = new AtomicInteger();
+                                Callback errorCallback = Callback.from(() ->
                                 {
-                                    // We do not notify ServletRequestListener on this dispatch because it might not
-                                    // be dispatched to an error page, so we delegate this responsibility to the ErrorHandler.
-                                    dispatch(() -> errorHandler.handle(_servletContextRequest, getServletContextResponse(), blocker));
-                                    blocker.block();
-                                }
+                                    if (count.incrementAndGet() == 2)
+                                        _state.scheduleDispatch();
```
gregw: @lachlan-roberts  I think the count is a bit confusing and it can be done with an AtomicBoolean:
```java
  AtomicBoolean asyncCompletion = new AtomicBoolean(false)
  Callback errorCallback = Callback.from(() ->
  {
      if (!asyncCompletion.compareAndSet(false, true))
        _state.scheduleDispatch() 
  }
  dispatch(() -> errorHandler.handle(_servletContextRequest, getServletContextResponse(), errorCallback));
  if (asyncCompletion.compareAndSet(false, true))                                   
    return false;
```

lachlan-roberts: done

gregw:
This got too out of date.  starting again....

gregw:
attempt two

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -117,16 +112,17 @@ public Callback getCallback()
...
+     * a servlet mapping. The association remains until {@link #recycle()} is called.
      * @param servletContextRequest The servlet context request to associate
      * @see #recycle()
      */
     public void associate(ServletContextRequest servletContextRequest)
     {
-        _state.recycle();
+        // TODO This is needed as requests that are handled before the ServletHandler are not recycled.
+        if (_servletContextRequest != null)
+            recycle();
```
gregw: We really need a cheap and easy mechanism to trigger on request completion... i.e both callback complete and handle method exited.

gregw:
@lachlan-roberts nudge

lachlan-roberts (dismissed):

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -480,13 +480,15 @@ public boolean handle()
 
                     case DISPATCH:
                     {
-                        dispatch(_requestDispatchable);
+                        reopen();
```
lachlan-roberts, line 491: Why do we need this `reopen()`?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -654,6 +646,12 @@ else if (ExceptionUtil.areNotAssociated(cause, x))
         return !suspended;
     }
 
+    private void reopen()
```
janbartel, line 657: Why are there so many calls to this method?

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
```diff
@@ -117,16 +112,17 @@ public Callback getCallback()
...
-     * a servlet mapping.
+     * This method is called by the {@link ServletContextHandler} when a core {@link Request} is accepted and associated with
+     * a servlet mapping. The association remains until {@link #recycle()} is called.
      * @param servletContextRequest The servlet context request to associate
      * @see #recycle()
      */
     public void associate(ServletContextRequest servletContextRequest)
     {
-        _state.recycle();
+        // TODO This is needed as requests that are handled before the ServletHandler are not recycled.
```
janbartel: Oh the irony - the comment on the PR says its purpose is to remove TODOs, and here's a new one being added! :grin:

janbartel (dismissed):

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
gregw: Oh it didn't need the TODO part in the comment :)

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
gregw: Because previously there was a `dispatch(Dispatchable)` method that once upon a time did lots of stuff but ultimately was reduced down to just doing a reopen and calling the dispatchable.... but I have inlined this method to get rid of the creation of the lambda dispatchable... but that means I need to call reopen every where there used to be a call to dispatch.   A small price to pay for less garbage on every dispatch.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletChannel.java
gregw: perhaps not.  We were doing it before in dispatch method, but I'll try without and see what happens....

gregw: it wasn't necessary

gregw: oh yes it was.  It is needed for multiple requests on the same connection.

janbartel (approved):

PR #10071, opened 2023-07-06
add SizeLimitHandler to Jetty-12

lachlan-roberts:
adapt the `SizeLimitHandler` from jetty 9.4 to the Jetty 12 core Handler style

### DIFF
--- a/jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/SizeLimitHandlerServletTest.java
+++ b/jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/SizeLimitHandlerServletTest.java
@@ -128,10 +128,6 @@ public class SizeLimitHandlerServletTest
         ContentResponse response = _client.POST(uri)
             .body(new StringRequestContent(content)).send();
 
-        System.err.println(response);
-        System.err.println(response.getHeaders());
-        System.err.println(response.getContentAsString());
-
         assertThat(response.getStatus(), equalTo(HttpStatus.PAYLOAD_TOO_LARGE_413));
     }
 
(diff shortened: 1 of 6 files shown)


gregw (changes requested):
This looks good, but can you also add a test where this is put inside an ee10 `ServletContextHandler`, perhaps with a `GzipHandler` also in the context.   It is not needed to test this handler, but I want to make sure we test the ability to nest core handlers inside ee10 contexts.   We probably also need to work out simple methods to configure/deploy

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/SizeLimitHandlerServletTest.java
```diff
@@ -0,0 +1,166 @@
...
+        sizeLimitHandler.setHandler(gzipHandler);
+        contextHandler.insertHandler(sizeLimitHandler);
+
+        contextHandler.addServlet(new HttpServlet()
+        {
+            @Override
+            protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOException
+            {
+                String requestContent = IO.toString(req.getInputStream());
+                resp.getWriter().print(requestContent);
```
gregw, line 79: It would be good to test both if the size limit is broken incrementally and if it is broken by an initial large write.

gregw (changes requested):
almost!  one more thing!

jetty-ee10/jetty-ee10-servlet/src/test/java/org/eclipse/jetty/ee10/servlet/SizeLimitHandlerServletTest.java
lachlan-roberts: We do have this test, but not in the servlet version of the test using GZIP.
It is in `SizeLimitHandlerTest`.

gregw (approved):

PR #10073, opened 2023-07-06
Add a module to produce a standard OSGi Repository

laeubi:
Jetty currently already provides a P2 Update-Site as maven artifact, but there is also now a [standardized repository format](https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.repository.html) available.

This adds a new module 'jetty-osgi-repository' that generates such a standard repository.

This currently references the tycho snapshot repository but I plan to do a new release soon so it would be great if it already can be reviewed and discussed if it could be an option for jetty.

As with the P2 module this does not really require anything special or maintained, everything is performed by the maven plugin.

### DIFF
--- a/jetty-osgi-repository/pom.xml
+++ b/jetty-osgi-repository/pom.xml
@@ -1,0 +1,33 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>org.eclipse.jetty</groupId>
+		<artifactId>jetty-project</artifactId>
+		<version>10.0.16-SNAPSHOT</version>
+	</parent>
+	<artifactId>jetty-osgi-repository</artifactId>
+	<name>Jetty OSGi Repository</name>
+	<description>Provides a OSGi Repository for all Jetty</description>
+	<packaging>repository</packaging>
+	<properties>
+		<tycho-version>4.0.1-SNAPSHOT</tycho-version>
+	</properties>
+	<pluginRepositories>
+		<pluginRepository>
+			<id>tycho-snapshots</id>
+			<url>https://repo.eclipse.org/content/repositories/tycho-snapshots/</url>
+		</pluginRepository>
+	</pluginRepositories>
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.eclipse.tycho</groupId>
+				<artifactId>tycho-repository-plugin</artifactId>
+				<version>${tycho-version}</version>
+				<extensions>true</extensions>
+			</plugin>
+		</plugins>
+	</build>
+</project>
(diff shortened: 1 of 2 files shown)


janbartel:
@laeubi thanks for the PR. I'm a bit mystified as I can't see how this module and/or the tycho plugin works out which jetty modules to include in the osgi repository? Also, what is the repository that is produced? Is it a zip file? Or are the jetty modules automagically uploaded to an Eclipse repository that follows the osgi repository format?

BTW this doesn't seem to build currently? Can you change this to a `draft` PR, as it relies on snapshot versions of the tycho plugin?

laeubi:
> I'm a bit mystified as I can't see how this module and/or the tycho plugin works out which jetty modules to include in the osgi repository?

It works by analyzing all projects in the reactor an collecting those build artifacts that are a bundle (what is most of the jetty artifacts).

> Also, what is the repository that is produced? Is it a zip file? 

The result is a `repository.xml` in the `target` folder I have attached the one I produced by running `mvn clean package -DskipTests` from this pr ([repository.xml.zip](https://github.com/eclipse/jetty.project/files/11967022/repository.xml.zip)).

> Or are the jetty modules automagically uploaded to an Eclipse repository that follows the osgi repository format?

Everything is still referenced as maven artifact (you can search for `<attribute name="url"` in the xml) so the repository.xml only contains the metadata but no actual files.

> BTW this doesn't seem to build currently?

It seems that some parts are using Java 11 but sadly Tycho requires Java 17 to build :-(

> Can you change this to a draft PR, as it relies on snapshot versions of the tycho plugin?

will do that.

janbartel:
@laeubi thanks for those explanations, makes sense.

I'm wondering if you might not be better off targeting jetty-12? Pretty soon we're going to release 12.0.0 and that will become our main development branch because it can do everything jetty-10 and jetty-11 do. Plus it uses jdk-19 as a minimum.

joakime:
We've been moving away from all of the variations of OSGI repository over the years.

Group A wants it in one format. 
Group B wants it in a different format.
Group C doesn't want all of the artifacts. (it complicates things for them in some way)
Group D wants only the minimum bytecode support. (every release of Jetty has a minimum and recommended version to use all features of Jetty - eg: Jetty 10 is minimum Java 11, but needs Java 17 for specific features, and Java 20 for even more features)
Group E wants it available on a public website.
Group F wants support from a maven repository.
Group G wants the signatures compatible with the limited set of technologies in Java 8.

This mess is the reality of OSGI repository stuff over the years.

So now we leave the OSGI related build in Eclipse Jetty to the Eclipse Tycho team.
They have recommended (and maintain) the `jetty-p2` artifact on our build.

https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-p2/pom.xml

This apparently uses `org.eclipse.tycho:tycho-p2-repository-plugin`

laeubi:
@joakime yes I know (I initially added the p2 stuff), the `tycho-repository-plugin` will hopefully replace the p2 part some time but unless then it would be good to just use both things in parallel, so I think targetting jetty-12 is a good approach.

joakime:
Perhaps all a small reorg is warranted now as well.

* move `/jetty-p2/` into `/jetty-osgi/jetty-p2/`
* move this `/jetty-osgi-repository/` into `jetty-osgi/jetty-osgi-repository/`

joakime:
The `jetty-p2` repo has been active in Jetty 10/11 for a while now.
We are working on `jetty-p2` for Jetty 12 in PR #12015

PR #10077, opened 2023-07-06
Issue #10068 allow Resource configuration in jetty maven plugin

### LINKED ISSUE #10068: Jetty 12: instantiation of `HashLoginService`
**Jetty version(s)**
12.0.0.beta2

**Description**
Hey Jetty team. The beta versions of Jetty 12 work well so far. However, while trying to upgrade from 11, I've experienced a problem with instantiation in a Maven integration test.

When using the Maven plugin for Jetty 11, the `config` property of the `HashLoginService` was a String and could be set via the Maven configuration directly as described in the docs. In Jetty 12 the config is now an instance of `Resource`. Since no implementation of `Resource` provides a no-arg constructor, the existing implementations cannot be used in the Maven configuration.

Maybe we could add a utility module with classes supporting Maven configurations to solve the instantiation problem without changing the current resource semantics.

janbartel:
Closes #10068

### DIFF
--- a/jetty-ee10/jetty-ee10-maven-plugin/src/it/jetty-start-mojo-it/jetty-simple-webapp/src/config/login.xml
+++ b/jetty-ee10/jetty-ee10-maven-plugin/src/it/jetty-start-mojo-it/jetty-simple-webapp/src/config/login.xml
@@ -1,0 +1,6 @@
+jetty: MD5:164c88b302622e17050af52c89945d44,user
+admin: CRYPT:adpexzg3FUZAk,server-administrator,content-administrator,admin
+other: OBF:1xmk1w261u9r1w1c1xmq,user
+plain: plain,user
+user: password,user
+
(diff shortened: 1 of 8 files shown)


gregw (approved):
This looks good, but probably should update some doco somewhere?

jetty-ee10/jetty-ee10-maven-plugin/src/it/jetty-start-mojo-it/jetty-simple-webapp/src/config/login.xml
```diff
@@ -0,0 +1,6 @@
+jetty: MD5:164c88b302622e17050af52c89945d44,user
```
basil, line 1: Why is this file named `login.xml` when its contents are not an XML file?

janbartel: Looks like  a typo to me.

PR #10081, opened 2023-07-07
Fix replacement logic for `Configuration` lists

janbartel:
jetty-12 ee10

When a list of `Configuration` classes was passed to the `Configurations(List<Configuration>)` constructor, it was not processing the list for replacements.


jetty-ee10/jetty-ee10-webapp/src/main/java/org/eclipse/jetty/ee10/webapp/Configurations.java
```diff
@@ -231,7 +231,9 @@ public Configurations()
 
     public Configurations(List<Configuration> configurations)
     {
-        _configurations = new ArrayList<>(configurations == null ? Collections.emptyList() : configurations);
+        _configurations = new ArrayList<>();
```
gregw, line 234: add a comment like: 
```java
    // add configurations individually so that replacements can be handled
```

gregw (dismissed):
LGTM, but a comment about why would be nice.

gregw (approved):

PR #10082, opened 2023-07-07
Various cleanups of StringUtil and TypeUtil

gregw:
Removed deprecated and unused methods
Moved charset handling to MimeTypes
resolve IDE warnings

### DIFF
--- a/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
+++ b/jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
@@ -16,8 +16,10 @@ package org.eclipse.jetty.http;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.nio.charset.UnsupportedCharsetException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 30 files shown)


jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
```diff
@@ -1049,7 +1035,7 @@ public BufferedReader getReader() throws IOException
 
         String encoding = getCharacterEncoding();
         if (encoding == null)
-            encoding = StringUtil.__ISO_8859_1;
+            encoding = MimeTypes.ISO_8859_1;
```
joakime, line 1038: Why is a Charset constant on the `MimeTypes` class?  Seems like a weird place for that.
Charset deals with encodings of bytes to Strings (and back), why not have these Charset constants on `StringUtil`?

gregw: @joakime Because the charset has nothing to do with strings in general, and everything to do with mime types that often have a charset argument: e.g. `text/html;charset=utf-8`
If I was going to be more radical, I'd rename `MimeTypes` to `ContentTypes` as strictly speaking a content type is a mimetype plus an optional charset.   But this PR is about util, not http.

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
```diff
@@ -235,6 +244,62 @@ public static Type getBaseType(String contentType)
...
+    }
+
+    /**
+     * Convert alternate charset names (eg utf8) to normalized
+     * name (eg UTF-8).
+     *
+     * @param s the charset to normalize
+     * @return the normalized charset (or null if normalized version not found)
+     */
+    public static String normalizeCharset(String s)
```
lorban: Rename the `s` parameter to `charsetName`?

jetty-core/jetty-http/src/main/java/org/eclipse/jetty/http/MimeTypes.java
```diff
@@ -235,6 +244,62 @@ public static Type getBaseType(String contentType)
...
+    /**
+     * Convert alternate charset names (eg utf8) to normalized
+     * name (eg UTF-8).
+     *
+     * @param s the charset to normalize
+     * @param offset the offset in the charset
+     * @param length the length of the charset in the input param
+     * @return the normalized charset (or null if not found)
+     */
+    public static String normalizeCharset(String s, int offset, int length)
```
lorban: Rename the `s` parameter to `charsetName`?

lorban (dismissed):
I like it, except for a couple of nits.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletApiRequest.java
joakime: That's only true for mime types that have a character representation.

Take the mime-type `application/octet-stream`, that has no charset, and never will.
Nor will `image/png` or `image/jpeg` and similar.

gregw: Yeah but charset has a lot more to do with mymetypes than it does general strings

lorban (approved):

PR #9974, opened 2023-06-26
Fix ServletContext.getResourcePaths()

lorban:
The `ServletContextApi.getResourceAsStream()`, `ServletContextApi.getResource()` and `ServletContextApi.getResourcePaths()` are canonicalizing the path parameter while they should be normalizing it instead.

Fixes #9972

PR #10085, opened 2023-07-07
Issue #10084 - Directory results from `getResourcePaths(String)` should include trailing slash

### LINKED ISSUE #10084: ServletApiContext.getResourcePaths() doesn't respect the spec
**Jetty version(s)**

12.0.0.beta3

openjdk version "17.0.7" 2023-04-18
OpenJDK Runtime Environment (Red_Hat-17.0.7.0.7-4.fc38) (build 17.0.7+7)
OpenJDK 64-Bit Server VM (Red_Hat-17.0.7.0.7-4.fc38) (build 17.0.7+7, mixed mode, sharing)

**OS type/version**

Fedora 38

**Description**

I am trying to list files and directories within a `WEB-INF/` folder which has a structure similar to the one below:

```
/WEB-INF/
  lib/
  custom-taglibs/
    my-tablib.tld
  web.xml
```

According to the spec, a call to `servletApiContext.getResourcePaths(/WEB-INF/)` should return:

* /WEB-INF/lib/
* /WEB-INF/custom-taglibs/
* /WEB-INF/web.xml

Instead, the method returns directory names without a trailing `/`:

* /WEB-INF/lib
* /WEB-INF/custom-taglibs
* /WEB-INF/web.xml

This makes it much harder to determine which entries are subdirectories and which ones aren't.

**How to reproduce?**

Create subdirectories in an application's `/WEB-INF/` folder and call `ServletApiContext.getResourcePaths("/WEB-INF/")`.

joakime:
* Adding support for trailing slashes on `getResourcePaths(String)` results.
* Backporting tests from ee10 to ee9
* Fixing bug in `getRealPath(String)` as well


jetty-ee9/jetty-ee9-webapp/src/test/java/org/eclipse/jetty/ee9/webapp/WebAppContextTest.java
```diff
@@ -546,6 +551,132 @@ public void testBaseResourceAbsolutePath() throws Exception
...
+            "/WEB-INF/zero.xml",
+            "/WEB-INF/one.xml",
+            "/WEB-INF/bar/",
+            "/WEB-INF/foo/"
+        };
+        assertThat(results, containsInAnyOrder(expected));
+    }
+
+    @Test
+    @Disabled("There is extra decoding of the nested-reserved paths that is getting in the way")
```
joakime, line 618: This can be fixed with @lorban backporting his work in PR #9974 back to ee9 tree.

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextHandler.java
```diff
@@ -810,12 +811,15 @@ public Set<String> getResourcePaths(String path)
...
+                path = path + '/';
 
             HashSet<String> set = new HashSet<>();
             for (Resource item: resource.list())
             {
-                set.add(path + item.getFileName());
+                String entry = path + item.getFileName();
+                if (item.isDirectory())
+                    entry = entry + '/';
+                set.add(entry);
```
gregw, line 822: Is it possible to resolve this within the resource classes?   Previously the resource classes would have the trailing '/' on directories so it would be everywhere.  Doing it just here as a post process risks not covering all usages.

gregw (dismissed):

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextHandler.java
joakime: The Resource object returns a `List<Resource>` so I think perhaps the place would be `Resource.getFileName()` ?

gregw: Yes getFileName, toString and anything else that returns the path as a string. Ideally even the URI should have the trailing slash. But
I don't think Path retains it.

joakime: Turns out fixing `getFileName()` breaks a lot of test cases. 
See latest Jenkins build on this PR.
I'll noodle through them on monday.

gregw: Would be good to check if any of these texts existed in jetty 11 and see if they were changed to not expect the trailing slash

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/PathResourceTest.java
```diff
@@ -353,6 +356,25 @@ public void testNullCharEndingFilename(WorkDir workDir) throws Exception
...
+    public void testGetFileName(WorkDir workDir) throws IOException
+    {
+        Path tmpPath = workDir.getEmptyPathDir();
+        Path dir = tmpPath.resolve("foo-dir");
+        FS.ensureDirExists(dir);
+        Path file = dir.resolve("bar.txt");
+        Files.writeString(file, "This is bar.txt", StandardCharsets.UTF_8);
+
+        Resource baseResource = new PathResource(tmpPath);
+        assertThat(baseResource.getFileName(), endsWith(File.separator));
```
lorban: I don't think you should use `File.separator` as the trailing character isn't platform-dependent.

jetty-ee9/jetty-ee9-webapp/src/test/java/org/eclipse/jetty/ee9/webapp/WebAppContextTest.java
lorban: That would also require backporting the fix to 10/11.

lorban (changes requested):

jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/ServletContextHandler.java
joakime: The Jetty 11 `Resource.getName()` does not return the trailing slash on directories.

However, the Jetty 11 `Resource.list()` returns a String array, and that array has the trailing slash.
The `Resource.list()` method in Jetty 11 is used ...

* ContextHandler.getResourcePaths(String) which returns a `Set<String>` on the Servlet API
* Resource.getListHTML - used to iterate a directory, and build a new set of Resources, that eventually are presented (in HTML)
* MetaInfConfiguration - iterates over WEB-INF/lib/*.jar files, and builds a new set of Resources that eventually wind up in the WebAppClassLoader
* WebAppClassLoader - used in addJars(Resource) to iterate a directory and build a new set of Resources that are added to the WebAppClassLoader
* Runner.Classpath.addJars(Resource) - same a WebAppClassLoader.addJars(Resource)

Turns out that in Jetty 12 we skip the String array step (eg: the _Resource -> String array -> List of Resource -> iterate/process_ flow).
In Jetty 11, we don't rely on the results of `Resource.getName()` for anything critical, it seems to be used only for things like debugging, exception messages, keys in the annotation parser, etc.

jetty-core/jetty-util/src/test/java/org/eclipse/jetty/util/resource/PathResourceTest.java
joakime: Using the same values as what `Path.getFileName()` return now.

jetty-ee9/jetty-ee9-webapp/src/test/java/org/acme/webapp/GetRealPathsServlet.java
```diff
@@ -0,0 +1,49 @@
...
+        collectResourcePaths("/").stream()
+            .map(p -> getServletContext().getRealPath(p))
+            .forEach(resp.getWriter()::println);
+        resp.getWriter().flush();
+    }
+
+    private Set<String> collectResourcePaths(String path)
+    {
+        Set<String> allResourcePaths = new LinkedHashSet<>();
+        Set<String> pathsForPath = getServletContext().getResourcePaths(path);
```
lorban, line 38: I think you meant to call `getServletContext().getRealPath(path)` here. 

You should probably rename this method and variables for clarity too.

jetty-ee9/jetty-ee9-webapp/src/test/java/org/eclipse/jetty/ee9/webapp/WebAppContextTest.java
```diff
@@ -546,6 +551,132 @@ public void testBaseResourceAbsolutePath() throws Exception
...
+    public void testGetResourcePathsWithDirsFromCollection() throws Exception
+    {
+        Server server = newServer();
+
+        WebAppContext context = new WebAppContext();
+        context.setContextPath("/");
+        context.setBaseResource(ResourceFactory.combine(
+            ResourceFactory.root().newResource(MavenPaths.findTestResourceDir("wars/layer0/")),
+            ResourceFactory.root().newResource(MavenPaths.findTestResourceDir("wars/layer1/")),
+            ResourceFactory.root().newResource(MavenPaths.findTestResourceDir("wars/with_dirs/"))
```
lorban: Please avoid using `ResourceFactory.root()` when `context.getResourceFactory()` is as easy as well as better practice.

lorban (changes requested):

jetty-ee9/jetty-ee9-webapp/src/test/java/org/acme/webapp/GetRealPathsServlet.java
joakime: This is doing the right thing.
It collects all of the servlet references first with `getResourcePaths()`, then iterates through each result and uses `getRealPath()` against it.
Outputting the `getRealPath()` of each result to the `HttpServletResponse.getWriter()`.

Besides, you wrote this, for ee10, and this is just copied into place for ee9. ;-)

jetty-ee9/jetty-ee9-webapp/src/test/java/org/eclipse/jetty/ee9/webapp/WebAppContextTest.java
joakime: Done.

jetty-ee9/jetty-ee9-webapp/src/test/java/org/acme/webapp/GetRealPathsServlet.java
lorban: You are right, I mis-read the code.

lorban (approved):

gregw:
@joakime whoa! you're dismissing my reviews now and merging anyway?!?!?!  
I still believe this is the wrong approach of just fixing listing.  It looks like the dropping of the trailing "/" for directories was made when you switched from `FileResource` to `PathResource`.
Fixing it like this just feels like playing what-a-mole with resource bugs.

gregw:
@joakime I have pushed an extra test in c435bfdbe65110546399df0b05b0c0f3ccbd8f97 that checks that for a `getResource(String)` the result includes a trailing slash for a directory.    This test passes, because it turns out that we already have the code in `PathResource` to ensure there is a trailing slash on the URI.

So I'm less concerned about this PR now... other than dismissing negative review and that it feels wrong that we already have a `Resource` that internally has the trailing '/' and that returns that in the URI, but that here we are using `getFilename()` that doesn't return the trailing slash, so we have to re-add it.  Looks like make work to me and a chance to get something wrong.  Surely we can ask a resource for it's segment relative to another and not have to do the work again externally?

joakime:
See https://github.com/eclipse/jetty.project/pull/10085#discussion_r1258185318

The behavior in Jetty 9/10/11 for this ...

* Trailing slash was present only in `ServletContext.getResourcePaths(String)` usages
* `Resource.list()` was the place a trailing slash was done.
* `Resource.list()` returned a `String[]`
* `Resource.list()` was only used as-is for `ServletContext.getResourcePaths(String)` (all other uses of `Resource.list()` took the values and created another collection of `Resource` objects, which meant the trailing slash wasn't meaningful)
* The `Resource.getName()` method did not return a trailing slash for JarFile, File, and Path implementations.
* `URLResource.getName()` returned the trailing slash, but only if it was present on the input (it didn't add it)
* `URLResource` behavior with `getName()` was not applied to `JarFile` implementation.

joakime:
> @joakime I have pushed an extra test in [c435bfd](https://github.com/eclipse/jetty.project/commit/c435bfdbe65110546399df0b05b0c0f3ccbd8f97) that checks that for a `getResource(String)` the result includes a trailing slash for a directory. This test passes, because it turns out that we already have the code in `PathResource` to ensure there is a trailing slash on the URI.

The `getResource(String)` returns a value in URL path terms, right? (not filesystem path terms).

The `getRealPath(String)` likely has the same requirements for trailing slash, but the returned value is in native file system path terms.

Examples:
We have a webapp (exploded directory) on windows, in path `C:\jetty\webapps\context\`.
* `getRealPath("/WEB-INF/web.xml")` should return `C:\jetty\webapps\context\WEB-INF\web.xml`
* `getRealPath("/css/")` should return `C:\jetty\webapps\context\css\`

Right?

gregw:
Correct, `getResource(String)` returns a `URL`.    This is correctly including a trailing slash for directories because our resource implementations correctly contain an internal URI that includes the trailing slash for directories.

It seams wrong/fragile that in order to implement `getResourcePaths(String)` that there is not some `API` in resource that can give reasonable access to this trailing slash.  Instead this PR used a filename method, which probably rightly doesn't include the trailing slash and then has to recreate the trailing slash repeating existing logic.

It is not explicit what space that the return set of `getResourcePaths(String)` is in, but given that it doesn't have "real" in it's method name and the examples all include '/', then I think it is safe to assume it is a set of partial URI path segments, decoded (like most other servlet APIs).  It should be possible to extract that from the URI in the resource.

But, this got merged without approval, so I guess it is done.   But let's remember this when doing any future resource work and look to see if we can avoid duplicating this logic.

PR #10090, opened 2023-07-10
Improved `start.jar` dry-run command line quoting

gregw:
Use shell double quoting for arguments.
Replaces #9999 for https://github.com/eclipse/jetty.docker/issues/148

PR #10099, opened 2023-07-12
Quoting for start arguments

gregw:
backport of #10090

probably should be merged first and then merged through 11 & 12 before applying #10090

### DIFF
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
@@ -127,7 +127,7 @@ public class CommandLineBuilder
             return input;
 
         StringBuilder builder = new StringBuilder(input.length() * 2);
-        builder.append('\'');
+        builder.append("'");
         builder.append(input, 0, --i);
 
         while (i < input.length())
@@ -136,10 +136,10 @@ public class CommandLineBuilder
             if (c == '\'')
             {
                 // There is no escape for a literal single quote, so we must leave the quotes
-                // and then escape the single quote. We test for the start/end of the string so
+                // and then escape the single quote. We test for the start/end of the string, so
                 // we can be less ugly in those cases.
                 if (i == 1)
-                    builder.insert(0, '\\').append('\'');
+                    builder.insert(0, "\\").append("'");
                 else if (i == input.length())
                     builder.append("'\\");
                 else
@@ -148,7 +148,7 @@ public class CommandLineBuilder
             else
                 builder.append(c);
         }
-        builder.append('\'');
+        builder.append("'");
 
         return builder.toString();
     }
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
@@ -53,31 +53,81 @@ public class CommandLineBuilder
         return "java";
     }
 
-    /**
-     * Perform an optional quoting of the argument, being intelligent with spaces and quotes as needed. If a subString is set in quotes it won't the subString
-     * won't be escaped.
-     *
-     * @param arg the argument to quote
-     * @return the quoted and escaped argument
-     * @deprecated no replacement, quoting is done by {@link #toQuotedString()} now.
-     */
-    @Deprecated
-    public static String quote(String arg)
-    {
-        return "'" + arg + "'";
-    }
-
-    private List<String> args;
+    private final StringBuilder commandLine = new StringBuilder();
+    private final List<String> args = new ArrayList<>();
+    private final String separator;
 
     public CommandLineBuilder()
     {
-        args = new ArrayList<String>();
+        this(false);
     }
 
-    public CommandLineBuilder(String bin)
+    public CommandLineBuilder(boolean multiline)
     {
-        this();
-        args.add(bin);
+        separator = multiline ? (" \\" + System.lineSeparator() + "  ") : " ";
+    }
+
+    /**
+     * Quote a string suitable for use with a command line shell using double quotes.
+     * <p>This method applies doubles quoting as described for the unix {@code sh} commands:
+     * Enclosing characters within double quotes preserves the literal meaning of all characters except
+     * dollarsign ($), backquote (`), and backslash (\).
+     * The backslash inside double quotes is historically weird, and serves
+     * to quote only the following characters: {@code $ ` " \ newline}.
+     * Otherwise, it remains literal.
+     *
+     * @param input The string to quote if needed
+     * @return The quoted string or the original string if quotes are not necessary
+     */
+    public static String shellQuoteIfNeeded(String input)
+    {
+        if (input == null || input.length() == 0)
+            return input;
+
+        int i = 0;
+        boolean needsQuoting = false;
+        while (!needsQuoting && i < input.length())
+        {
+            char c = input.charAt(i++);
+
+            // needs quoting unless a limited set of known good characters
+            needsQuoting = !(
+                (c >= 'A' && c <= 'Z') ||
+                (c >= 'a' && c <= 'z') ||
+                (c >= '0' && c <= '9') ||
+                c == '/' ||
+                c == ':' ||
+                c == '.' ||
+                c == '-' ||
+                c == '_'
+                );
+        }
+
+        if (!needsQuoting)
+            return input;
+
+        StringBuilder builder = new StringBuilder(input.length() * 2);
+        builder.append('"');
+        builder.append(input, 0, --i);
+
+        while (i < input.length())
+        {
+            char c = input.charAt(i++);
+            switch (c)
+            {
+                case '"':
+                case '\\':
+                case'`':
+                case'$':
+                    builder.append('\\').appendCodePoint(c);
+                    break;
+                default: builder.appendCodePoint(c);
+            }
+        }
+
+        builder.append('"');
+
+        return builder.toString();
     }
 
     /**
@@ -92,29 +142,25 @@ public class CommandLineBuilder
         if (arg != null)
         {
             args.add(arg);
+            if (commandLine.length() > 0)
+                commandLine.append(separator);
+            commandLine.append(shellQuoteIfNeeded(arg));
         }
     }
 
     /**
-     * Similar to {@link #addArg(String)} but concats both name + value with an "=" sign, quoting were needed, and excluding the "=" portion if the value is
-     * undefined or empty.
-     *
-     * <pre>
-     *   addEqualsArg("-Dname", "value") = "-Dname=value"
-     *   addEqualsArg("-Djetty.home", "/opt/company inc/jetty (7)/") = "-Djetty.home=/opt/company\ inc/jetty\ (7)/"
-     *   addEqualsArg("-Djenkins.workspace", "/opt/workspaces/jetty jdk7/") = "-Djenkins.workspace=/opt/workspaces/jetty\ jdk7/"
-     *   addEqualsArg("-Dstress", null) = "-Dstress"
-     *   addEqualsArg("-Dstress", "") = "-Dstress"
-     * </pre>
-     *
      * @param name the name
      * @param value the value
      */
-    public void addEqualsArg(String name, String value)
+    public void addArg(String name, String value)
     {
+        if (commandLine.length() > 0)
+            commandLine.append(separator);
+        commandLine.append(shellQuoteIfNeeded(name));
         if ((value != null) && (value.length() > 0))
         {
             args.add(name + "=" + value);
+            commandLine.append('=').append(shellQuoteIfNeeded(value));
         }
         else
         {
@@ -123,17 +169,31 @@ public class CommandLineBuilder
     }
 
     /**
-     * Add a simple argument to the command line.
-     * <p>
-     * Will <b>NOT</b> quote/escape arguments that have a space in them.
-     *
-     * @param arg the simple argument to add
+     * @param option the option
+     * @param name the name
+     * @param value the value
      */
-    public void addRawArg(String arg)
+    public void addArg(String option, String name, String value)
     {
-        if (arg != null)
+        if (commandLine.length() > 0)
+            commandLine.append(separator);
+        commandLine.append(option);
+        if (name == null || name.length() == 0)
         {
-            args.add(arg);
+            args.add(option);
+        }
+        else
+        {
+            commandLine.append(shellQuoteIfNeeded(name));
+            if ((value != null) && (value.length() > 0))
+            {
+                args.add(option + name + "=" + value);
+                commandLine.append('=').append(shellQuoteIfNeeded(value));
+            }
+            else
+            {
+                args.add(option + name);
+            }
         }
     }
 
@@ -145,19 +205,11 @@ public class CommandLineBuilder
     @Override
     public String toString()
     {
-        return toString(" ");
-    }
-
-    public String toString(String delim)
-    {
         StringBuilder buf = new StringBuilder();
-
         for (String arg : args)
         {
             if (buf.length() > 0)
-            {
-                buf.append(delim);
-            }
+                buf.append(' ');
             buf.append(arg); // we assume escaping has occurred during addArg
         }
 
@@ -169,23 +221,9 @@ public class CommandLineBuilder
      *
      * @return the toString but each arg that has spaces is surrounded by {@code '} (single-quote tick)
      */
-    public String toQuotedString()
+    public String toCommandLine()
     {
-        StringBuilder buf = new StringBuilder();
-
-        for (String arg : args)
-        {
-            if (buf.length() > 0)
-                buf.append(' ');
-            boolean needsQuotes = (arg.contains(" "));
-            if (needsQuotes)
-                buf.append("'");
-            buf.append(arg);
-            if (needsQuotes)
-                buf.append("'");
-        }
-
-        return buf.toString();
+        return commandLine.toString();
     }
 
     public void debug()
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
@@ -95,8 +95,8 @@ public class CommandLineBuilder
      */
     public static String shellQuoteIfNeeded(String input)
     {
-        // Single quotes are used because double quotes are process differently by some shells and the xarg
-        // command used by jetty.xh
+        // Single quotes are used because double quotes are processed differently by some shells and the xarg
+        // command used by jetty.sh
 
         if (input == null)
             return null;
@@ -136,8 +136,8 @@ public class CommandLineBuilder
             if (c == '\'')
             {
                 // There is no escape for a literal single quote, so we must leave the quotes
-                // and then escape the single quote.  If we are at the start or end of the string
-                // we can be less ugly and avoid an empty ''
+                // and then escape the single quote. We test for the start/end of the string so
+                // we can be less ugly in those cases.
                 if (i == 1)
                     builder.insert(0, '\\').append('\'');
                 else if (i == input.length())
@@ -218,11 +218,7 @@ public class CommandLineBuilder
      */
     public void addOption(String option)
     {
-        Objects.requireNonNull(option);
-        if (commandLine.length() > 0)
-            commandLine.append(separator);
-        commandLine.append(option);
-        args.add(option);
+        addOption(option, null, null);
     }
 
     /**
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
@@ -118,6 +118,7 @@ public class CommandLineBuilder
                 c == ':' ||
                 c == '.' ||
                 c == ',' ||
+                c == '+' ||
                 c == '-' ||
                 c == '_'
                 );
@@ -213,7 +214,7 @@ public class CommandLineBuilder
     }
 
     /**
-     * Adds a "-O" style option to the command line with no quoting, for example `--help`.
+     * Adds a "-OPTION" style option to the command line with no quoting, for example `--help`.
      * @param option the option
      */
     public void addOption(String option)
@@ -222,7 +223,7 @@ public class CommandLineBuilder
     }
 
     /**
-     * Adds a "-Oname=value" style option to the command line with
+     * Adds a "-OPTIONname=value" style option to the command line with
      * name and value quoted if necessary, for example "-Dprop=value".
      * @param option the option
      * @param name the name
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
@@ -69,23 +69,21 @@ public class CommandLineBuilder
     }
 
     /**
-     * Quote a string suitable for use with a command line shell using double quotes.
-     * <p>This method applies doubles quotes suitable for a POSIX compliant shell:
-     * Enclosing characters within double quotes preserves the literal meaning of all characters except
-     * dollarsign ($), backquote (`), and backslash (\).
-     * The backslash inside double quotes is historically weird, and serves
-     * to quote only the following characters: {@code $ ` " \ newline}.
-     * Otherwise, it remains literal.
+     * This method applies single quotes suitable for a POSIX compliant shell if
+     * necessary.
      *
      * @param input The string to quote if needed
      * @return The quoted string or the original string if quotes are not necessary
      */
     public static String shellQuoteIfNeeded(String input)
     {
+        // Single quotes are used because double quotes are process differently by some shells and the xarg
+        // command used by jetty.xh
+
         if (input == null)
             return null;
         if (input.length() == 0)
-            return "\"\"";
+            return "''";
 
         int i = 0;
         boolean needsQuoting = false;
@@ -111,26 +109,28 @@ public class CommandLineBuilder
             return input;
 
         StringBuilder builder = new StringBuilder(input.length() * 2);
-        builder.append('"');
+        builder.append('\'');
         builder.append(input, 0, --i);
 
         while (i < input.length())
         {
             char c = input.charAt(i++);
-            switch (c)
+            if (c == '\'')
             {
-                case '"':
-                case '\\':
-                case '`':
-                case '$':
-                case '\n':
-                    builder.append('\\').append(c);
-                    break;
-                default: builder.append(c);
+                // There is no escape for a literal single quote, so we must leave the quotes
+                // and then escape the single quote.  If we are at the start or end of the string
+                // we can be less ugly and avoid an empty ''
+                if (i == 1)
+                    builder.insert(0, '\\').append('\'');
+                else if (i == input.length())
+                    builder.append("'\\");
+                else
+                    builder.append("'\\''");
             }
+            else
+                builder.append(c);
         }
-
-        builder.append('"');
+        builder.append('\'');
 
         return builder.toString();
     }
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
@@ -16,6 +16,7 @@ package org.eclipse.jetty.start;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 public class CommandLineBuilder
 {
@@ -81,8 +82,10 @@ public class CommandLineBuilder
      */
     public static String shellQuoteIfNeeded(String input)
     {
-        if (input == null || input.length() == 0)
-            return input;
+        if (input == null)
+            return null;
+        if (input.length() == 0)
+            return "\"\"";
 
         int i = 0;
         boolean needsQuoting = false;
@@ -118,11 +121,11 @@ public class CommandLineBuilder
             {
                 case '"':
                 case '\\':
-                case'`':
-                case'$':
-                    builder.append('\\').appendCodePoint(c);
+                case '`':
+                case '$':
+                    builder.append('\\').append(c);
                     break;
-                default: builder.appendCodePoint(c);
+                default: builder.append(c);
             }
         }
 
@@ -132,9 +135,7 @@ public class CommandLineBuilder
     }
 
     /**
-     * Add a simple argument to the command line.
-     * <p>
-     * Will quote arguments that have a space in them.
+     * Add a simple argument to the command line, quoted if necessary.
      *
      * @param arg the simple argument to add
      */
@@ -142,59 +143,66 @@ public class CommandLineBuilder
     {
         if (arg != null)
         {
-            args.add(arg);
             if (commandLine.length() > 0)
                 commandLine.append(separator);
+            args.add(arg);
             commandLine.append(shellQuoteIfNeeded(arg));
         }
     }
 
     /**
+     * Add a "name=value" style argument to the command line with
+     * name and value quoted if necessary.
      * @param name the name
      * @param value the value
      */
     public void addArg(String name, String value)
     {
+        Objects.requireNonNull(name);
+
         if (commandLine.length() > 0)
             commandLine.append(separator);
-        commandLine.append(shellQuoteIfNeeded(name));
+
         if ((value != null) && (value.length() > 0))
         {
             args.add(name + "=" + value);
-            commandLine.append('=').append(shellQuoteIfNeeded(value));
+            commandLine.append(shellQuoteIfNeeded(name)).append('=').append(shellQuoteIfNeeded(value));
         }
         else
         {
             args.add(name);
+            commandLine.append(shellQuoteIfNeeded(name));
         }
     }
 
     /**
+     * Add a "-Oname=value" style argument to the command line with
+     * name and value quoted if necessary.
      * @param option the option
      * @param name the name
      * @param value the value
      */
     public void addArg(String option, String name, String value)
     {
+        Objects.requireNonNull(option);
+
         if (commandLine.length() > 0)
             commandLine.append(separator);
-        commandLine.append(option);
+
         if (name == null || name.length() == 0)
         {
+            commandLine.append(option);
             args.add(option);
+        }
+        else if ((value != null) && (value.length() > 0))
+        {
+            args.add(option + name + "=" + value);
+            commandLine.append(option).append(shellQuoteIfNeeded(name)).append('=').append(shellQuoteIfNeeded(value));
         }
         else
         {
-            commandLine.append(shellQuoteIfNeeded(name));
-            if ((value != null) && (value.length() > 0))
-            {
-                args.add(option + name + "=" + value);
-                commandLine.append('=').append(shellQuoteIfNeeded(value));
-            }
-            else
-            {
-                args.add(option + name);
-            }
+            args.add(option + name);
+            commandLine.append(option).append(shellQuoteIfNeeded(name));
         }
     }
 
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
@@ -213,7 +213,7 @@ public class CommandLineBuilder
     }
 
     /**
-     * Add a "-Oname" style option to the command line with no quoting.
+     * Adds a "-O" style option to the command line with no quoting, for example `--help`.
      * @param option the option
      */
     public void addOption(String option)
@@ -222,8 +222,8 @@ public class CommandLineBuilder
     }
 
     /**
-     * Add a "-Oname=value" style option to the command line with
-     * name and value quoted if necessary.
+     * Adds a "-Oname=value" style option to the command line with
+     * name and value quoted if necessary, for example "-Dprop=value".
      * @param option the option
      * @param name the name
      * @param value the value
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
@@ -98,6 +98,7 @@ public class CommandLineBuilder
                 c == '/' ||
                 c == ':' ||
                 c == '.' ||
+                c == ',' ||
                 c == '-' ||
                 c == '_'
                 );
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
@@ -95,8 +95,8 @@ public class CommandLineBuilder
      */
     public static String shellQuoteIfNeeded(String input)
     {
-        // Single quotes are used because double quotes are processed differently by some shells and the xarg
-        // command used by jetty.sh
+        // Single quotes are used because double quotes
+        // are evaluated differently by some shells.
 
         if (input == null)
             return null;
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
@@ -70,7 +70,7 @@ public class CommandLineBuilder
 
     /**
      * Quote a string suitable for use with a command line shell using double quotes.
-     * <p>This method applies doubles quoting as described for the unix {@code sh} commands:
+     * <p>This method applies doubles quotes suitable for a POSIX compliant shell:
      * Enclosing characters within double quotes preserves the literal meaning of all characters except
      * dollarsign ($), backquote (`), and backslash (\).
      * The backslash inside double quotes is historically weird, and serves
@@ -123,6 +123,7 @@ public class CommandLineBuilder
                 case '\\':
                 case '`':
                 case '$':
+                case '\n':
                     builder.append('\\').append(c);
                     break;
                 default: builder.append(c);
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
@@ -54,6 +54,17 @@ public class CommandLineBuilder
         return "java";
     }
 
+    /**
+     * @param arg string
+     * @return Quoted string
+     * @deprecated use {@link #shellQuoteIfNeeded(String)}
+     */
+    @Deprecated
+    public static String quote(String arg)
+    {
+        return "'" + arg + "'";
+    }
+
     private final StringBuilder commandLine = new StringBuilder();
     private final List<String> args = new ArrayList<>();
     private final String separator;
@@ -61,6 +72,13 @@ public class CommandLineBuilder
     public CommandLineBuilder()
     {
         this(false);
+    }
+
+    @Deprecated
+    public CommandLineBuilder(String bin)
+    {
+        this();
+        args.add(bin);
     }
 
     public CommandLineBuilder(boolean multiline)
@@ -152,6 +170,15 @@ public class CommandLineBuilder
     }
 
     /**
+     * @deprecated use {@link #addArg(String, String)}
+     */
+    @Deprecated
+    public void addEqualsArg(String name, String value)
+    {
+        addArg(name, value);
+    }
+
+    /**
      * Add a "name=value" style argument to the command line with
      * name and value quoted if necessary.
      * @param name the name
@@ -174,6 +201,15 @@ public class CommandLineBuilder
             args.add(name);
             commandLine.append(shellQuoteIfNeeded(name));
         }
+    }
+
+    /**
+     * @deprecated use {@link #addArg(String)}
+     */
+    @Deprecated
+    public void addRawArg(String arg)
+    {
+        addArg(arg);
     }
 
     /**
@@ -224,6 +260,15 @@ public class CommandLineBuilder
         }
 
         return buf.toString();
+    }
+
+    /**
+     * @deprecated use {@link #toCommandLine()}
+     */
+    @Deprecated
+    public String toQuotedString()
+    {
+        return toCommandLine();
     }
 
     /**
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
@@ -213,13 +213,26 @@ public class CommandLineBuilder
     }
 
     /**
-     * Add a "-Oname=value" style argument to the command line with
+     * Add a "-Oname" style option to the command line with no quoting.
+     * @param option the option
+     */
+    public void addOption(String option)
+    {
+        Objects.requireNonNull(option);
+        if (commandLine.length() > 0)
+            commandLine.append(separator);
+        commandLine.append(option);
+        args.add(option);
+    }
+
+    /**
+     * Add a "-Oname=value" style option to the command line with
      * name and value quoted if necessary.
      * @param option the option
      * @param name the name
      * @param value the value
      */
-    public void addArg(String option, String name, String value)
+    public void addOption(String option, String name, String value)
     {
         Objects.requireNonNull(option);
 
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
@@ -780,19 +780,23 @@ public class StartArgs
                 }
                 for (Map.Entry<String, Set<String>> entry : jmodPatch.entrySet())
                 {
-                    cmd.addArg("--patch-module", entry.getKey(), String.join(File.pathSeparator, entry.getValue()));
+                    cmd.addArg("--patch-module", null, null);
+                    cmd.addArg(entry.getKey(), String.join(File.pathSeparator, entry.getValue()));
                 }
                 for (Map.Entry<String, Set<String>> entry : jmodOpens.entrySet())
                 {
-                    cmd.addArg("--add-opens", entry.getKey(), String.join(",", entry.getValue()));
+                    cmd.addArg("--add-opens", null, null);
+                    cmd.addArg(entry.getKey(), String.join(",", entry.getValue()));
                 }
                 for (Map.Entry<String, Set<String>> entry : jmodExports.entrySet())
                 {
-                    cmd.addArg("--add-exports", entry.getKey(), String.join(",", entry.getValue()));
+                    cmd.addArg("--add-exports", null, null);
+                    cmd.addArg(entry.getKey(), String.join(",", entry.getValue()));
                 }
                 for (Map.Entry<String, Set<String>> entry : jmodReads.entrySet())
                 {
-                    cmd.addArg("--add-reads", entry.getKey(), String.join(",", entry.getValue()));
+                    cmd.addArg("--add-reads", null, null);
+                    cmd.addArg(entry.getKey(), String.join(",", entry.getValue()));
                 }
             }
             else
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
@@ -816,7 +816,7 @@ public class StartArgs
             {
                 for (Prop p : properties)
                 {
-                    if (!p.key.startsWith("java."))
+                    if (!p.key.startsWith("java.version."))
                         cmd.addArg(p.key, properties.expand(p.value));
                 }
             }
--- a/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
+++ b/jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
@@ -208,6 +208,7 @@ public class StartArgs
     private boolean listConfig = false;
     private boolean version = false;
     private boolean dryRun = false;
+    private boolean multiLine = false;
     private final Set<String> dryRunParts = new HashSet<>();
     private boolean jpms = false;
     private boolean createStartD = false;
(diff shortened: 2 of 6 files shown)


jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -53,31 +53,82 @@ public static String findJavaBin()
...
+     * Otherwise, it remains literal.
+     *
+     * @param input The string to quote if needed
+     * @return The quoted string or the original string if quotes are not necessary
+     */
+    public static String shellQuoteIfNeeded(String input)
     {
-        this();
-        args.add(bin);
+        if (input == null || input.length() == 0)
```
sbordet: Use `StringUtil` instead.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -53,31 +53,82 @@ public static String findJavaBin()
...
+
+        while (i < input.length())
+        {
+            char c = input.charAt(i++);
+            switch (c)
+            {
+                case '"':
+                case '\\':
+                case'`':
+                case'$':
```
sbordet: Add space after `case`.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -53,31 +53,82 @@ public static String findJavaBin()
...
+                    builder.append('\\').appendCodePoint(c);
```
sbordet: Just `append(c)`, as `appendCodePoint()` converts to `int` and I'm not sure they are the same.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -53,31 +53,82 @@ public static String findJavaBin()
...
+                    break;
+                default: builder.appendCodePoint(c);
```
sbordet: Ditto.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -123,17 +170,31 @@ public void addEqualsArg(String name, String value)
...
+     * @param value the value
      */
-    public void addRawArg(String arg)
+    public void addArg(String option, String name, String value)
     {
-        if (arg != null)
+        if (commandLine.length() > 0)
+            commandLine.append(separator);
+        commandLine.append(option);
+        if (name == null || name.length() == 0)
```
sbordet, line 239: Use `StringUtil` here.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -123,17 +170,31 @@ public void addEqualsArg(String name, String value)
...
         {
-            args.add(arg);
+            args.add(option);
+        }
+        else
+        {
+            commandLine.append(shellQuoteIfNeeded(name));
+            if ((value != null) && (value.length() > 0))
```
sbordet: Use `StringUtil` here.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -123,17 +170,31 @@ public void addEqualsArg(String name, String value)
...
+     * @param name the name
+     * @param value the value
      */
-    public void addRawArg(String arg)
+    public void addArg(String option, String name, String value)
     {
-        if (arg != null)
+        if (commandLine.length() > 0)
+            commandLine.append(separator);
+        commandLine.append(option);
```
sbordet: The logic to add `option`, `name` and `value` to `commandLine` and `args` seems different.
Should they not be paired and be equal?
E.g. above we always add `option` to `commandLine`, why not to `args` too?

jetty-start/src/main/resources/org/eclipse/jetty/start/usage.txt
```diff
@@ -60,17 +60,20 @@ Report Commands:
...
   --dry-run=<part>(,<part>)*
-                   Prints specific parts of the command line. The parts are:
+                   Prints specific parts of the command line in a format usable by
+                   a the unix sh command. The parts are:
                      o  "java" - the JVM to run
                      o  "opts" - the JVM options (e.g. -D, -X and -XX flags)
                      o  "path" - the JVM class-path and/or the JPMS module-path
                      o  "main" - the main class to run
                      o  "args" - the arguments passed to the main class
+                     o  "multipart" - a fake part that causes the command to be
```
sbordet: It's `multiline`! 😂

jetty-start/src/test/java/org/eclipse/jetty/start/CommandLineBuilderTest.java
```diff
@@ -23,43 +28,83 @@ public class CommandLineBuilderTest
...
+        return Stream.of(
+            Arguments.of(null, null),
+            Arguments.of("", ""),
+            Arguments.of("Hello", "Hello"),
+            Arguments.of("Hell0", "Hell0"),
+            Arguments.of("H-llo_world", "H-llo_world"),
+            Arguments.of("H:llo/world", "H:llo/world"),
+            Arguments.of("Hello World", "\"Hello World\""),
+            Arguments.of("foo\\bar", "\"foo\\\\bar\""),
+            Arguments.of("'single'", "\"'single'\"")
```
sbordet: Add other cases for all escaped characters, especially `$`.

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
```diff
@@ -208,6 +208,7 @@ public class StartArgs
     private boolean listConfig = false;
     private boolean version = false;
     private boolean dryRun = false;
+    private boolean multiLine = false;
```
sbordet, line 211: This field is never read.

sbordet (changes requested):
I like the idea of being minimal to quote, and only if necessary.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
gregw: I initially used `StringUtil`, but we need to shade it into jetty-start, which I tried, but then it needed `Index` shaded, which needed `AbstractTrie`, etc. etc. So I stopped and just used this directly.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
gregw: oops, left over from earlier version. Good catch!

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
gregw: We do.  All three branches below add option to the args one way or another.     The code is a bit convoluted because it is minimal.  Let me make it more explicit even if it duplicates a few lines....

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
gregw: ditto

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
gregw: It is passed in the constructor of CommandLineBuilder at line 523... ah no, missed in jetty-10

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
gregw: is that better?

joakime:
This branch has broken `jetty.sh` handling of complex properties that was pointed out in Issue #9309

```
$ ~/code/jetty/jetty.project-10.0.x/jetty-home/target/jetty-home/bin/jetty.sh start
Starting Jetty: xargs: unmatched double quote; by default quotes are special to xargs unless you use the -0 option
2023-07-12 07:53:14.956:WARN :oejx.XmlConfiguration:main: Unable to execute XmlConfiguration
java.nio.file.NoSuchFileException: %s
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
	at java.base/sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:218)
	at java.base/java.nio.file.Files.newByteChannel(Files.java:380)
	at java.base/java.nio.file.Files.newByteChannel(Files.java:432)
	at java.base/java.nio.file.spi.FileSystemProvider.newInputStream(FileSystemProvider.java:422)
	at java.base/java.nio.file.Files.newInputStream(Files.java:160)
	at org.eclipse.jetty.util.resource.PathResource.getInputStream(PathResource.java:456)
	at org.eclipse.jetty.xml.XmlConfiguration.<init>(XmlConfiguration.java:218)
	at org.eclipse.jetty.xml.XmlConfiguration.main(XmlConfiguration.java:1845)
Exception in thread "main" java.nio.file.NoSuchFileException: %s
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
	at java.base/sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:218)
	at java.base/java.nio.file.Files.newByteChannel(Files.java:380)
	at java.base/java.nio.file.Files.newByteChannel(Files.java:432)
	at java.base/java.nio.file.spi.FileSystemProvider.newInputStream(FileSystemProvider.java:422)
	at java.base/java.nio.file.Files.newInputStream(Files.java:160)
	at org.eclipse.jetty.util.resource.PathResource.getInputStream(PathResource.java:456)
	at org.eclipse.jetty.xml.XmlConfiguration.<init>(XmlConfiguration.java:218)
	at org.eclipse.jetty.xml.XmlConfiguration.main(XmlConfiguration.java:1845)
```

Attached is the jetty.base to reproduce this.
[requestlog-custom-base.tar.gz](https://github.com/eclipse/jetty.project/files/12028139/requestlog-custom-base.tar.gz)

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -53,87 +54,155 @@ public static String findJavaBin()
...
+     * @return The quoted string or the original string if quotes are not necessary
+     */
+    public static String shellQuoteIfNeeded(String input)
     {
-        this();
-        args.add(bin);
+        if (input == null)
+            return null;
+        if (input.length() == 0)
+            return "\"\"";
```
joakime:
```suggestion
            return "''";
```

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -53,87 +54,155 @@ public static String findJavaBin()
...
 
-    public CommandLineBuilder(String bin)
+    /**
+     * Quote a string suitable for use with a command line shell using double quotes.
+     * <p>This method applies doubles quoting as described for the unix {@code sh} commands:
+     * Enclosing characters within double quotes preserves the literal meaning of all characters except
+     * dollarsign ($), backquote (`), and backslash (\).
+     * The backslash inside double quotes is historically weird, and serves
+     * to quote only the following characters: {@code $ ` " \ newline}.
+     * Otherwise, it remains literal.
```
joakime:
```suggestion
     * Quote a string suitable for use with a command line shell using strong quotes {@code '}.
     * <p>This method applies strong quoting as described for the unix {@code sh} commands:
     * Enclosing characters within strong quotes preserves the literal meaning of all characters.
```

We don't want to use double quotes with sh / bash / zsh, we want to use strong quotes `'`, as that will not interpret the contents between the quotes.

Examples:

**`sh` on ubuntu**

```
[joakim@hyperion ~]$ which sh
/usr/bin/sh
[joakim@hyperion ~]$ ls -la /usr/bin/sh
lrwxrwxrwx 1 root root 4 Mar 23  2022 /usr/bin/sh -> dash
[joakim@hyperion ~]$ sh -i -l
$ echo "Hello $USER"
Hello joakim
$ echo 'Hello $USER'
Hello $USER
```

**`bash` on ubuntu**

```
[joakim@hyperion ~]$ echo "Hello $USER"
Hello joakim
[joakim@hyperion ~]$ echo 'Hello $USER'
Hello $USER
```

**`zsh` on ubuntu**

```
[joakim@hyperion ~]$ zsh -i -l
joakim@hyperion ~ % echo "Hello $USER"
Hello joakim
joakim@hyperion ~ % echo 'Hello $USER'
Hello $USER
```

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -53,87 +54,155 @@ public static String findJavaBin()
...
+                c == '-' ||
+                c == '_'
+                );
+        }
+
+        if (!needsQuoting)
+            return input;
+
+        StringBuilder builder = new StringBuilder(input.length() * 2);
+        builder.append('"');
```
joakime:
```suggestion
        builder.append('\'');
```

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -53,87 +54,155 @@ public static String findJavaBin()
...
+                case '\\':
+                case '`':
+                case '$':
+                    builder.append('\\').append(c);
+                    break;
+                default: builder.append(c);
+            }
+        }
+
+        builder.append('"');
```
joakime:
```suggestion
        builder.append('\'');
```

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -53,87 +54,155 @@ public static String findJavaBin()
...
+            switch (c)
+            {
+                case '"':
+                case '\\':
+                case '`':
+                case '$':
+                    builder.append('\\').append(c);
+                    break;
+                default: builder.append(c);
+            }
```
joakime:
```suggestion
            switch (c)
            {
                case '\'':
                    builder.append("'\''");
                    break;
                default: builder.append(c);
            }
```

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -53,87 +54,155 @@ public static String findJavaBin()
...
-     * <pre>
-     *   addEqualsArg("-Dname", "value") = "-Dname=value"
-     *   addEqualsArg("-Djetty.home", "/opt/company inc/jetty (7)/") = "-Djetty.home=/opt/company\ inc/jetty\ (7)/"
-     *   addEqualsArg("-Djenkins.workspace", "/opt/workspaces/jetty jdk7/") = "-Djenkins.workspace=/opt/workspaces/jetty\ jdk7/"
-     *   addEqualsArg("-Dstress", null) = "-Dstress"
-     *   addEqualsArg("-Dstress", "") = "-Dstress"
-     * </pre>
-     *
+     * Add a "name=value" style argument to the command line with
+     * name and value quoted if necessary.
```
joakime: Why get rid of this javadoc?

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -53,87 +54,155 @@ public static String findJavaBin()
...
      * @param name the name
      * @param value the value
      */
-    public void addEqualsArg(String name, String value)
+    public void addArg(String name, String value)
```
joakime, line 188: Cannot rename this in Jetty 10.
We can rename this for Jetty 12+ tho.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -53,87 +54,155 @@ public static String findJavaBin()
...
-     *
-     * @param arg the simple argument to add
+     * Add a "-Oname=value" style argument to the command line with
+     * name and value quoted if necessary.
+     * @param option the option
+     * @param name the name
+     * @param value the value
      */
-    public void addRawArg(String arg)
+    public void addArg(String option, String name, String value)
```
joakime: Cannot rename in Jetty 10.
We can rename in Jetty 12+ tho.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -144,20 +213,12 @@ public List<String> getArgs()
 
     @Override
     public String toString()
-    {
-        return toString(" ");
-    }
-
-    public String toString(String delim)
```
joakime, line 151: This public method cannot be removed in Jetty 10.
Safe for Jetty 12+ tho.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -169,23 +230,9 @@ public String toString(String delim)
      *
      * @return the toString but each arg that has spaces is surrounded by {@code '} (single-quote tick)
      */
-    public String toQuotedString()
+    public String toCommandLine()
```
joakime, line 289: Another public method that cannot be renamed in Jetty 10.
Good for Jetty 12 tho.

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
```diff
@@ -1225,6 +1223,12 @@ public void parse(final String rawarg, String source)
             int colon = arg.indexOf('=');
             for (String part : arg.substring(colon + 1).split(","))
             {
+                if ("multiline".equalsIgnoreCase(part))
+                {
+                    multiLine = true;
+                    continue;
+                }
+
```
joakime, line 1235: Not necessary if you use "Strong Quoting" per the various shells.

jetty-start/src/main/resources/org/eclipse/jetty/start/usage.txt
```diff
@@ -60,17 +60,20 @@ Report Commands:
...
-                   Prints specific parts of the command line. The parts are:
+                   Prints specific parts of the command line in a format usable by
+                   the unix sh command. The parts are:
                      o  "java" - the JVM to run
                      o  "opts" - the JVM options (e.g. -D, -X and -XX flags)
                      o  "path" - the JVM class-path and/or the JPMS module-path
                      o  "main" - the main class to run
                      o  "args" - the arguments passed to the main class
+                     o  "multiline" - a fake part that causes the command to be
+                                      generated over multiple lines
```
joakime: Not needed.

jetty-start/src/test/java/org/eclipse/jetty/start/CommandLineBuilderTest.java
joakime: Most cases are not needed if using "Strong Quoting" per the shells

joakime (changes requested):
Don't use Double Quotes if you want to support `sh`

Use "Strong Quoting" instead, which is supported across the board.
And then you have only 1 character to worry about to escape.

**bash**

```
[joakim@hyperion ~]$ java -jar ~/java/args-dump-1.1.jar 'Hello internal quotes' foo bar
args.length=3
args[0]: [Hello internal quotes]
args[1]: [foo]
args[2]: [bar]
[joakim@hyperion ~]$ java -jar ~/java/args-dump-1.1.jar 'Hello '\''internal'\'' quotes' foo bar
args.length=3
args[0]: [Hello 'internal' quotes]
args[1]: [foo]
args[2]: [bar]
[joakim@hyperion ~]$ java -jar ~/java/args-dump-1.1.jar 'Hello $USER env var' zed foo
args.length=3
args[0]: [Hello $USER env var]
args[1]: [zed]
args[2]: [foo]
```

**sh**

```
$ java -jar ~/java/args-dump-1.1.jar 'Hello internal quotes' foo bar
args.length=3
args[0]: [Hello internal quotes]
args[1]: [foo]
args[2]: [bar]
$ java -jar ~/java/args-dump-1.1.jar 'Hello '\''internal'\'' quotes' foo bar
args.length=3
args[0]: [Hello 'internal' quotes]
args[1]: [foo]
args[2]: [bar]
$ java -jar ~/java/args-dump-1.1.jar 'Hello $USER env var' zed foo
args.length=3
args[0]: [Hello $USER env var]
args[1]: [zed]
args[2]: [foo]
```

**zsh**

```
joakim@hyperion ~ % java -jar ~/java/args-dump-1.1.jar 'Hello internal quotes' foo bar
args.length=3
args[0]: [Hello internal quotes]
args[1]: [foo]
args[2]: [bar]
joakim@hyperion ~ % java -jar ~/java/args-dump-1.1.jar 'Hello '\''internal'\'' quotes' foo bar
args.length=3
args[0]: [Hello 'internal' quotes]
args[1]: [foo]
args[2]: [bar]
joakim@hyperion ~ % java -jar ~/java/args-dump-1.1.jar 'Hello $USER env var' zed foo
args.length=3
args[0]: [Hello $USER env var]
args[1]: [zed]
args[2]: [foo]
```

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
gregw: Using single quotes is ugly/complex because we have to escape literal single quotes outside of the scope of the quotes, so quoting something like `my.property='foo bar'` (where the single quotes are literal) will end up with the unreadable `my.property=\''foo bar'\'` 

Double quotes a simpler and with 4 characters that need escaping are well enough defined and simple to support.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
gregw: sure we can. This is an internal method.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
gregw: because the -D part no longer applied.

gregw:
@joakime your custom requestlog base is broken.  It contains the line:
```
jetty.requestlog.formatString=%{client}a - %u %{dd/MMM/yyyy:HH:mm:ss ZZZ|GMT}t "%r $JETTY_HOME`' %s %O "%{Referer}i" "%{User-Agent}i"
```
So the unmatched quotes are from your ini file!

If I take the OPs reproducer from #9309 and put it in my ini file as:
```
jetty.requestlog.formatString=%{client}a - %u %{dd/MMM/yyyy:HH:mm:ss ZZZ|GMT}t "%r" %s %O "%{Referer}i" "%{User-Agent}i"
```

then it starts fine and --dry-run=multiline reports:
```
/usr/lib/jvm/java-17-openjdk-amd64/bin/java \
  -Djava.io.tmpdir=/tmp \
  -Djetty.home=/home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home \
  -Djetty.base=/home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/base \
  --class-path \
  /home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/base/resources:/home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/lib/logging/slf4j-api-2.0.5.jar:/home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/lib/logging/jetty-slf4j-impl-10.0.16-SNAPSHOT.jar:/home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/lib/jetty-servlet-api-4.0.6.jar:/home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/lib/jetty-http-10.0.16-SNAPSHOT.jar:/home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/lib/jetty-server-10.0.16-SNAPSHOT.jar:/home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/lib/jetty-xml-10.0.16-SNAPSHOT.jar:/home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/lib/jetty-util-10.0.16-SNAPSHOT.jar:/home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/lib/jetty-io-10.0.16-SNAPSHOT.jar:/home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/lib/jetty-security-10.0.16-SNAPSHOT.jar:/home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/lib/jetty-servlet-10.0.16-SNAPSHOT.jar:/home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/lib/jetty-webapp-10.0.16-SNAPSHOT.jar:/home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/lib/jetty-deploy-10.0.16-SNAPSHOT.jar \
  org.eclipse.jetty.xml.XmlConfiguration \
  jetty.base=/home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/base \
  jetty.base.uri=file:///home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/base \
  jetty.home=/home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home \
  jetty.home.uri=file:///home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home \
  jetty.requestlog.formatString="%{client}a - %u %{dd/MMM/yyyy:HH:mm:ss ZZZ|GMT}t \"%r\" %s %O \"%{Referer}i\" \"%{User-Agent}i\"" \
  jetty.webapp.addServerClasses=org.eclipse.jetty.logging.,file:///home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/lib/logging/,org.slf4j. \
  runtime.feature.alpn=true \
  slf4j.version=2.0.5 \
  /home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/etc/jetty-bytebufferpool.xml \
  /home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/etc/jetty-threadpool.xml \
  /home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/etc/jetty.xml \
  /home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/etc/jetty-webapp.xml \
  /home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/etc/jetty-deploy.xml \
  /home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/etc/jetty-http.xml \
  /home/gregw/src/jetty-10.0.x/jetty-home/target/jetty-home/etc/jetty-requestlog.xml
```

This looks perfectly correct

gregw:
I think we are [posix shell](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html) compliant: 
```
2.2.2 Single-Quotes
Enclosing characters in single-quotes ( '' ) shall preserve the literal value of each character within the single-quotes. A single-quote cannot occur within single-quotes.

2.2.3 Double-Quotes
Enclosing characters in double-quotes ( "" ) shall preserve the literal value of all characters within the double-quotes, with the exception of the characters backquote, <dollar-sign>, and <backslash>, as follows:

$
The <dollar-sign> shall retain its special meaning introducing parameter expansion (see [Parameter Expansion](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02)), a form of command substitution (see [Command Substitution](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_03)), and arithmetic expansion (see [Arithmetic Expansion](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_04)).
The input characters within the quoted string that are also enclosed between "$(" and the matching ')' shall not be affected by the double-quotes, but rather shall define that command whose output replaces the "$(...)" when the word is expanded. The tokenizing rules in [Token Recognition](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_03), not including the alias substitutions in [Alias Substitution](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_03_01), shall be applied recursively to find the matching ')'.

Within the string of characters from an enclosed "${" to the matching '}', an even number of unescaped double-quotes or single-quotes, if any, shall occur. A preceding <backslash> character shall be used to escape a literal '{' or '}'. The rule in [Parameter Expansion](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02) shall be used to determine the matching '}'.

`
The backquote shall retain its special meaning introducing the other form of command substitution (see [Command Substitution](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_03)). The portion of the quoted string from the initial backquote and the characters up to the next backquote that is not preceded by a <backslash>, having escape characters removed, defines that command whose output replaces "`...`" when the word is expanded. Either of the following cases produces undefined results:
A single-quoted or double-quoted string that begins, but does not end, within the "`...`" sequence

A "`...`" sequence that begins, but does not end, within the same double-quoted string

\
The <backslash> shall retain its special meaning as an escape character (see [Escape Character (Backslash)](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_02_01)) only when followed by one of the following characters when considered special:
$   `   "   \   <newline>

The application shall ensure that a double-quote is preceded by a <backslash> to be included within double-quotes. The parameter '@' has special meaning inside double-quotes and is described in [Special Parameters](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_05_02) .
```

gregw:
I'll change our doco to say posix shell...

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
joakime: The example property `my.property='foo bar'` would, as if it appeared in `start.d/my.ini` would mean that the end user wants the single quotes in the value as the end result (meaning if they do `properties.get("my.property")` they would expect a String with those single-tick characters: `"'foo bar'"`)

The end result would be for strong quotes ...
```
'-Dmy.property='\''foo bar'\'''
```

The entire argument is strong quoted, not just a section.

Example:

```
[joakim@hyperion ~]$ java -Dmy.property='foo bar' -XshowSettings:properties --version 2>&1 | grep my.prop
    my.property = foo bar
[joakim@hyperion ~]$ java '-Dmy.property='\''foo bar'\''' -XshowSettings:properties --version 2>&1 | grep my.prop
    my.property = 'foo bar'
```

This Strong Quoting also eliminates ALL interpretation of the contents of the quoted string by the shell (on all shells)

From the GNU dash man page (what the standard `sh` is on Ubuntu)

>   **Single Quotes**
>     Enclosing characters in single quotes preserves the literal meaning
>     of all the characters (except single quotes, making it impossible
>     to put single-quotes in a single-quoted string).

Something that Double Quotes does not do, and escaping will need to be SUPER robust (things you are ignoring in this PR will fail on various shells) if we go that route.

Yes, I'm aware that `dash` only mentions a handful of characters for Double Quotes, the other shells have far more special characters when encountered within Double Quotes (a behavior that Strong Quoting doesn't have)

joakime:
> I'll change our doco to say posix shell...

The current released versions of `dash` are _not_ POSIX shell compliant btw (but it's being worked on).

gregw:
@joakime ultimately, both double quotes and single quotes are well documented in `sh`, `bash`, `posix` etc.   We can use either, so long as we fully implement either spec.

I chose double quotes because I prefer the simplicity of just 5 characters needing escape within the quotes, whilst single quotes needs to have two mechanism: single quotes; escaped chars outside of quotes; and sometimes a combination.   Ultimately for the few difficult cases we might have, double quotes result in a lot more readable and understandable output.

Ultimately, for our purposes single and double quotes are equivalent, as we can convert one to the other, since we never include any unexpanded parameters.    Thus there is no compelling reason to change.

gregw:
> > I'll change our doco to say posix shell...
> 
> The current released versions of `dash` are _not_ POSIX shell compliant btw (but it's being worked on).

Then who knows how it implements single quotes? If they are not POSIX compliant, then neither quoting style is ok.

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
```diff
@@ -819,7 +816,8 @@ public CommandLineBuilder getMainArgs(Set<String> parts) throws IOException
             {
                 for (Prop p : properties)
                 {
-                    cmd.addRawArg(CommandLineBuilder.quote(p.key) + "=" + CommandLineBuilder.quote(properties.expand(p.value)));
+                    if (!p.key.startsWith("java."))
```
joakime: This broke the redefinement of `java.io.tmpdir`

jetty-start/src/main/resources/org/eclipse/jetty/start/usage.txt
joakime: `multiline` is not necessary.

If you are working on a shell you can accomplish this output on any already released version of Jetty now.

```
$ java -jar /path/to/jetty-home/start.jar --dry-run | xargs printf "%s\n"
```

A useful technique when troubleshooting the behavior of our quoting on various shells too (as the string to arg conversion for that shell is performed through the pipe to xargs)

joakime:
> @joakime your custom requestlog base is broken. It contains the line:
> 
> ```
> jetty.requestlog.formatString=%{client}a - %u %{dd/MMM/yyyy:HH:mm:ss ZZZ|GMT}t "%r $JETTY_HOME`' %s %O "%{Referer}i" "%{User-Agent}i"
> ```

Unmatched quotes are a perfectly legitimate thing to appear in one of our INIs.
Almost anything can appear as a property value in our INI. (the only exception being control characters, as we don't have the same escaping rules as a Java properties file).

BTW, this example base is supported properly on PR #10100 now.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -54,86 +55,204 @@ public static String findJavaBin()
...
-     * Will quote arguments that have a space in them.
+     * This method applies single quotes suitable for a POSIX compliant shell if
+     * necessary.
+     *
+     * @param input The string to quote if needed
+     * @return The quoted string or the original string if quotes are not necessary
+     */
+    public static String shellQuoteIfNeeded(String input)
+    {
+        // Single quotes are used because double quotes are process differently by some shells and the xarg
```
gregw:
```suggestion
        // Single quotes are used because double quotes are processed differently by some shells and the xarg
```

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -54,86 +55,204 @@ public static String findJavaBin()
...
+        // command used by jetty.xh
```
gregw:
```suggestion
        // command used by jetty.sh
```

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -54,86 +55,204 @@ public static String findJavaBin()
...
+        builder.append(input, 0, --i);
+
+        while (i < input.length())
+        {
+            char c = input.charAt(i++);
+            if (c == '\'')
+            {
+                // There is no escape for a literal single quote, so we must leave the quotes
+                // and then escape the single quote.  If we are at the start or end of the string
+                // we can be less ugly and avoid an empty ''
```
gregw:
```suggestion
                // There is no escape for a literal single quote, so we must leave the quotes
                // and then escape the single quote. We test for the start/end of the string so
                // we can be less ugly in those cases.
```

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -54,86 +55,204 @@ public static String findJavaBin()
...
+                (c >= 'A' && c <= 'Z') ||
+                (c >= 'a' && c <= 'z') ||
+                (c >= '0' && c <= '9') ||
+                c == '/' ||
+                c == ':' ||
+                c == '.' ||
+                c == ',' ||
+                c == '-' ||
+                c == '_'
+                );
```
sbordet, line 124: Should we have a static table for this lookup?

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -54,86 +55,204 @@ public static String findJavaBin()
...
+                    builder.insert(0, '\\').append('\'');
+                else if (i == input.length())
+                    builder.append("'\\");
+                else
+                    builder.append("'\\''");
+            }
+            else
+                builder.append(c);
+        }
+        builder.append('\'');
```
sbordet: The builder is used here with a `String` parameter and `char` parameter, which is difficult to read.
Is is possible to just use the `String` method?

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -54,86 +55,204 @@ public static String findJavaBin()
...
+     * Add a "-Oname" style option to the command line with no quoting.
+     * @param option the option
+     */
+    public void addOption(String option)
+    {
+        Objects.requireNonNull(option);
+        if (commandLine.length() > 0)
+            commandLine.append(separator);
+        commandLine.append(option);
+        args.add(option);
```
sbordet: To avoid duplication:

```suggestion
        addOption(option, null, null);
```

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
```diff
@@ -756,60 +757,60 @@ public CommandLineBuilder getMainArgs(Set<String> parts) throws IOException
...
                         .collect(Collectors.joining(File.pathSeparator));
-                    cmd.addRawArg(directories);
+                    cmd.addArg(directories);
                 }
 
                 if (!jmodAdds.isEmpty())
                 {
-                    cmd.addRawArg("--add-modules");
-                    cmd.addRawArg(String.join(",", jmodAdds));
+                    cmd.addArg("--add-modules");
```
sbordet: Following the others, this should be `addOption(...)`.

jetty-start/src/main/resources/org/eclipse/jetty/start/usage.txt
gregw: neat technique, but the multiline output is correctly slashed so it is still executable.  Frequently in docker style processing it is necessary to modify the output of the dry-run in some way, and having it broken up into multiple lines will make scripts like sed simpler to apply

jetty-start/src/test/java/org/eclipse/jetty/start/CommandLineBuilderTest.java
```diff
@@ -23,43 +28,87 @@ public class CommandLineBuilderTest
...
+            Arguments.of("Hell0", "Hell0"),
+            Arguments.of("Hello$World", "'Hello$World'"),
+            Arguments.of("Hello\\World", "'Hello\\World'"),
+            Arguments.of("Hello`World", "'Hello`World'"),
+            Arguments.of("'Hello World'", "\\''Hello World'\\'"),
+            Arguments.of("\"Hello World\"", "'\"Hello World\"'"),
+            Arguments.of("H-llo_world", "H-llo_world"),
+            Arguments.of("H:llo/world", "H:llo/world"),
+            Arguments.of("Hello World", "'Hello World'"),
+            Arguments.of("foo\\bar", "'foo\\bar'")
```
sbordet: Add `foo'bar`.

jetty-start/src/test/java/org/eclipse/jetty/start/CommandLineBuilderTest.java
```diff
@@ -23,43 +28,87 @@ public class CommandLineBuilderTest
...
+        cmd.addArg("-Djetty.home", "/opt/jetty");
+        cmd.addArg("monetary.symbol", "€");
+        assertThat(cmd.toCommandLine(), is("java -Djetty.home=/opt/jetty monetary.symbol='€'"));
+    }
+
+    public static Stream<Arguments> shellQuoting()
+    {
+        return Stream.of(
+            Arguments.of(null, null),
+            Arguments.of("", "''"),
```
sbordet, line 80: This is weird, why it is not the empty string `""` but has single quotes?

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
gregw: @joakime I think what you have been trying to say is: "xargs doesn't grok double quotes like y'd think".  Once I worked that out, then OK let's go with single quotes.   updated.

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
gregw: I'm not just excluding `java.version.`, which obviously we cannot set!

jetty-start/src/test/java/org/eclipse/jetty/start/CommandLineBuilderTest.java
```diff
@@ -23,43 +28,87 @@ public class CommandLineBuilderTest
...
+            Arguments.of("Hello", "Hello"),
+            Arguments.of("Hell0", "Hell0"),
+            Arguments.of("Hello$World", "'Hello$World'"),
+            Arguments.of("Hello\\World", "'Hello\\World'"),
+            Arguments.of("Hello`World", "'Hello`World'"),
+            Arguments.of("'Hello World'", "\\''Hello World'\\'"),
```
sbordet, line 86: This is weird. I would have expected to begin and end with a single quote.

sbordet (changes requested):

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
gregw: No.
This is in the start process, so it is only used to process one command line.   There are probably more characters in the lookup table than in the command line itself.

Besides, the option `--dry-run` is precisely for those who want quick startup and to avoid running start.jar at all, so those that want a quick start will not be using this code at all. I.e. don't optimise the slow path and this is definitely the slow path.

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
gregw:
```suggestion
                    cmd.addOption("--add-modules");
```

jetty-start/src/test/java/org/eclipse/jetty/start/CommandLineBuilderTest.java
gregw: because somehow somebody gave us an argument that was the empty string.  If we just return the empty string then we will lose the argument.  Returning empty quotes preserves that there in an argument of zero length

jetty-start/src/test/java/org/eclipse/jetty/start/CommandLineBuilderTest.java
gregw: Escaping `'` only works outside of quotes so it could be `''\\''Hello World'\\'''`, but the leading and trailing `''` are just useless noise. 

If we really cared we would quote something like `'Hello'` as `\'Hello\'` and never enter the quotes.  But I'm guessing that most times there are quotes, then there will be a character than needs quoting.

jetty-start/src/test/java/org/eclipse/jetty/start/CommandLineBuilderTest.java
gregw:
```suggestion
            Arguments.of("foo\\bar", "'foo\\bar'"),
            Arguments.of("foo'bar", "'foo'bar'")
```

jetty-start/src/test/java/org/eclipse/jetty/start/CommandLineBuilderTest.java
joakime: @sbordet this is particularly important when the argument has secondary arguments.

Like the `--classpath` argument on Java itself, that argument alone isn't useful, it needs the next argument to be valid.
 `--classpath <value>` would be two arguments (1. to indicate the desire to use classpath, and 2. to indicate the value of classpath).

So the use of `--classpath ''` would be two arguments, a classpath that is empty (I know, not terribly useful for Java, but it shows an example of an empty argument)

In jetty `start.jar` terms, this particular example (`--classpath ''`) shows up when you use `start.jar --dry-run` in a directory that has no libs defined (eg: an empty directory, or no directly declared libs, or no enabled modules that have libs)

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -54,86 +55,200 @@ public static String findJavaBin()
...
      */
+    @Deprecated
     public void addRawArg(String arg)
     {
-        if (arg != null)
+        addArg(arg);
+    }
+
+    /**
+     * Add a "-Oname" style option to the command line with no quoting.
```
sbordet:
```suggestion
     * Adds a "-O" style option to the command line with no quoting, for example `--help`.
```

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -54,86 +55,200 @@ public static String findJavaBin()
...
+     * @param option the option
+     */
+    public void addOption(String option)
+    {
+        addOption(option, null, null);
+    }
+
+    /**
+     * Add a "-Oname=value" style option to the command line with
+     * name and value quoted if necessary.
```
sbordet:
```suggestion
     * Adds a "-Oname=value" style option to the command line with
     * name and value quoted if necessary, for example "-Dprop=value".
```

sbordet (dismissed):
Small javadocs nits if you want to fix them, but LGTM.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
gregw: > @joakime I think what you have been trying to say is: "xargs doesn't grok double quotes like y'd think". Once I worked that out, then OK let's go with single quotes. updated.

@joakime what I meant to say was... thanks for persisting and I understood the issue eventually.

sbordet (dismissed):

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -54,86 +55,200 @@ public static String findJavaBin()
...
-     * @return the quoted and escaped argument
-     * @deprecated no replacement, quoting is done by {@link #toQuotedString()} now.
+     * @param arg string
+     * @return Quoted string
+     * @deprecated use {@link #shellQuoteIfNeeded(String)}
      */
     @Deprecated
     public static String quote(String arg)
     {
         return "'" + arg + "'";
```
joakime, line 65: I know it's deprecated, but at least we can be consistent in Jetty 10.

```suggestion
        return shellQuoteIfNeeded(arg);
```

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -54,86 +55,200 @@ public static String findJavaBin()
...
+            needsQuoting = !(
+                (c >= 'A' && c <= 'Z') ||
+                (c >= 'a' && c <= 'z') ||
+                (c >= '0' && c <= '9') ||
+                c == '/' ||
+                c == ':' ||
+                c == '.' ||
+                c == ',' ||
+                c == '-' ||
+                c == '_'
```
joakime, line 123: Add the `+` sign for the allowable list.

Example:

```
'/home/joakim/java/jvm/jdk-17.0.6+10/bin/java' -Djava.io.tmpdir=/tmp
```

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -54,86 +55,200 @@ public static String findJavaBin()
...
+     * @param option the option
+     */
+    public void addOption(String option)
+    {
+        addOption(option, null, null);
+    }
+
+    /**
+     * Adds a "-Oname=value" style option to the command line with
+     * name and value quoted if necessary, for example "-Dprop=value".
```
joakime: Is something is off here?
First reference is `-O` and the next reference is `-D` ?

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
joakime, line 211: This should be a separate concept, not as a "part" or "section" on `--dry-run=<parts>`

Consider instead a `--dry-run-format=<type>` where we have a few options.

* `sh` - what we do now (and is the default format)
* `windows` - an alternate mode for windows
* `newline` - what this `multiLine` is doing

And in the future we might even be able to support the various Daemon tooling and produce JSON or XML of the args that those tools consume.

joakime (changes requested):
Is it intentional that all `--<value>` arguments are passed to the JVM side of the `--dry-run` ?

Example:

`start.d/foo.ini`

```
--long-arg
--zed=zora
```

Those two will be present in `--dry-run` before the classpath.

----

If we have a directory with an equals sign, this is interpreted as a property, not an XML.

`start.d/mode.ini`
```
etc/mode=dev/example.xml
```

I think we check for equals before we check for `endsWith(".xml")`?

----

It's not possible to have an xml or lib with a directory with a backslash defined in an `start.d/<name>.ini`

Example:

```
etc/oh\no/example.xml
--lib=lib/rut\roh/main.jar
```

----

An argument that starts with an equals sign produces some odd output on `--dry-run`

`start.d/zed.ini`

```
=1234
```

results in the output

```
''=1234
```

If there is no key with the use of the `=` sign, that quoting is not necessary.

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
gregw: But how do you select which parts to output with this new option?

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
gregw: I thought about that, but anybody that is calling this method is happy with the existing behavior. We should not change it.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
gregw: First is representing the general style with O standing for option. The second is a specific example.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
joakime: > @joakime I think what you have been trying to say is: "xargs doesn't grok double quotes like y'd think". Once I worked that out, then OK let's go with single quotes. updated.

It's not `xargs` doing that. 
It is `bash` doing that, `bash` (and `zsh`) will evaluate on any redirection/pipe before redirection/pipe (unlike `dash`).
This was a design decision by `bash` to get around deficiencies in `sh` not allowing variables to be set on redirection/pipe.

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
joakime: The existing `--dry-run=<parts>` still exists.
Select what you want from that.

How the parts are presented are impacted by the new arg.

default behavior: (as `--dry-run-format=sh` is default, that's what is used)

```
$ java -jar start.jar --dry-run=opts,path,main,args
```

or on windows

```
> java -jar start.jar --dry-run=opts,args --dry-run-format=windows
```

or for new lines

```
$ java -jar start.jar --dry-run=opts,path,main,args --dry-run-format=newline
```

and in the future

```
$ java -jar start.jar --dry-run=opts,path,main,args --dry-run-format=json
```

gregw:
These may well be issues, but they have nothing to do with quoting or `--dry-run`

> Is it intentional that all `--<value>` arguments are passed to the JVM side of the `--dry-run` ?

Any unknown options are considered JVM options and are thus ordered as they are.

> If we have a directory with an equals sign, this is interpreted as a property, not an XML.

Indeed, any arg with an = will be recognised as a property and it has been like this since the beginning.  
If it every becomes an issue, then we can add an optional option called '--file" which makes the next arg always be handled as a file.

> It's not possible to have an xml or lib with a directory with a backslash defined in an `start.d/<name>.ini`

Ditto. If this ever becomes an issue, then we can add an optional `--xml` option.
But I think having a backslash in a filename is asking for problems.

> An argument that starts with an equals sign produces some odd output on `--dry-run`

Well perhaps we should make this an error, but it has not been a problem because there is no need for an arg starting with =

jetty-start/src/main/java/org/eclipse/jetty/start/StartArgs.java
gregw: This is too complex and verbose for a minor feature

gregw:
@sbordet I've run out of time to deal with this PR.  Can you take over this PR to get it approved and merged 10,11, and to 12.    The PR #10090 is out of date and I've not updated whilst waiting for this to be approved.   It might be simpler to abandon that and just merge this to 12 and fix up the env handling differences.

With regards to the multiline feature, I do not think this should be turned into anything as complex as has been suggested.   It is just a simple little tool and we have no need to support json or non posix shells.    If there is no agreement on it, then just remove it. We don't want to hold up progress over such a simple thing.

jetty-start/src/main/java/org/eclipse/jetty/start/CommandLineBuilder.java
```diff
@@ -54,86 +55,201 @@ public static String findJavaBin()
...
+     * This method applies single quotes suitable for a POSIX compliant shell if
+     * necessary.
+     *
+     * @param input The string to quote if needed
+     * @return The quoted string or the original string if quotes are not necessary
+     */
+    public static String shellQuoteIfNeeded(String input)
+    {
+        // Single quotes are used because double quotes are processed differently by some shells and the xarg
+        // command used by jetty.sh
```
joakime:
```suggestion
        // Single quotes are used because double quotes are evaluated differently by some shells such as bash
        // which is used by jetty.sh and the Jetty docker image
```

sbordet (approved):

joakime (approved):
LGTM, lets merge this to `jetty-12.0.x` asap

Issue #10055, opened 2023-07-03
Deployment of static files does not work with --dry-run Jetty-12


lachlan-roberts:
**Jetty version(s)**
12.0.x

**Description**
The output of dry run does not work when pasted into command line when using static content with `core-deploy` module.

**How to reproduce?**

``` bash
java -jar $JETTY_HOME/start.jar --add-to-start=core-deploy,http
mkdir webapps/ROOT
echo "hello world" > webapps/ROOT/index.html
java -jar $JETTY_HOME/start.jar --dry-run
# paste the output of the dry run
```

Output of --dry-run:
[...]

PR #10100, opened 2023-07-12
Fix Quoting of arguments from Jetty `start.jar --dry-run`

joakime:
This addresses the concerns brought up by Issue #10055 with a minimal fix.

### DIFF
--- a/jetty-home/src/main/resources/bin/jetty.sh
+++ b/jetty-home/src/main/resources/bin/jetty.sh
@@ -499,7 +499,8 @@ case "$ACTION" in
           disown \$!
           echo \$! > \"$JETTY_PID\""
       else
-        echo ${RUN_ARGS[*]} | xargs ${JAVA} > /dev/null &
+        # echo ${RUN_ARGS[*]} | xargs ${JAVA} > /dev/null &
+        echo ${RUN_ARGS[*]} | xargs ${JAVA} & # TODO: don't check in this version only used for cli debugging
         disown $!
         echo $! > "$JETTY_PID"
       fi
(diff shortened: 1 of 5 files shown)


joakime:
This is DRAFT, as more work on `jetty.sh` is being done to avoid re-evaluating the `RUN_ARGS` once they are set, to avoid double-expansion of arguments by the shell.

jetty-home/src/main/resources/bin/jetty.sh
```diff
@@ -499,7 +499,8 @@ case "$ACTION" in
           disown \$!
           echo \$! > \"$JETTY_PID\""
       else
-        echo ${RUN_ARGS[*]} | xargs ${JAVA} > /dev/null &
+        # echo ${RUN_ARGS[*]} | xargs ${JAVA} > /dev/null &
+        echo ${RUN_ARGS[*]} | xargs ${JAVA} & # TODO: don't check in this version only used for cli debugging
```
joakime, line 503: This is temporary (just avoiding the redirection to `/dev/null` during testing)

joakime:
After a conversation with @sbordet and @lorban to show them what we are dealing with, it might be possible to write unit tests that actually execute `jetty.sh` in a safe way.

Stay tuned, going to write up a few examples Thursday.

gregw (changes requested):
Why has this forked my branch and lost recent work? There is no need to do things like lookup tables, as this does not need to be super fast, plus it is only used at startup, so likely more time spent building lookup table than using it.

If there is an issue with jetty.sh, then I would prefer that work was done in #10099 and it is fixed, rather than redoing the work on quoting.

joakime:
Closing in favor of #10099

PR #10110, opened 2023-07-14
HttpChannel, AdaptiveExecutionStrategy: improve profiler-friendliness

stevenschlansker:
In an application with multiple Jetty instances in one JVM (e.g. integration test), when examining stack frames using the debugger or profiler, most samples that involve user code will have these two frames in their stack.

Unfortunately, with a lambda, the different Jetty instances actually have different class names for different lambdas, which causes stack analysis to falsely think the frames are different.

It's a little uglier, but by replacing these two specific lambdas with anonymous classes with a stable name, the profiler is able to see that the frames are in fact the same and collapse them, improving observability.

Here is an example Java Mission Control flame graph showing before, notice how the flame splits exactly at the `AdaptiveExecutionStrategy$$Lambda$` due to different names:
image

and showing after, now our stack traces only split where the handler chains actually differ (GzipHandler vs not) and thus are more understandable:
image

### DIFF
--- a/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
+++ b/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
@@ -494,15 +494,19 @@ public abstract class HttpChannel implements Runnable, HttpOutput.Interceptor
                         if (!_request.hasMetaData())
                             throw new IllegalStateException("state=" + _state);
 
-                        dispatch(DispatcherType.REQUEST, () ->
+                        dispatch(DispatcherType.REQUEST, new Dispatchable()
                         {
-                            for (HttpConfiguration.Customizer customizer : _configuration.getCustomizers())
+                            @Override
+                            public void dispatch() throws IOException, ServletException
                             {
-                                customizer.customize(getConnector(), _configuration, _request);
-                                if (_request.isHandled())
-                                    return;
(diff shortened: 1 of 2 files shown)


stevenschlansker:
I updated my ECA; I am not sure why the check still fails.

image

stevenschlansker:
I updated my ECA; I am not sure why the check still fails
image

sbordet:
Can you please re-issue this PR against the `jetty-10.0.x` branch? We will take care of merging forward.

sbordet:
It is really a bummer that lambda class names are unspecified, but I guess there is no way around it.

stevenschlansker:
Yes, I agree, it is a bummer. But changing the JVM behavior here seems unlikely. Thankfully, there seem to only be two places where it matters.

I rebased this PR on top of `jetty-10.0.x`

gregw (commented):
If making inner classes instead of lambdas, it is probably worthwhile adding toString method to then as well to make debugging easier

sbordet:
@stevenschlansker I have taken your changes and improved upon them, can you please check that branch `fix/jetty-10-avoid-lambda-flamegraph-split` works for you?

sbordet (approved):

sbordet:
Incorporated into #10115.

stevenschlansker:
Great, thank you!

sbordet:
@stevenschlansker did you try #10115? Yields a good flamegraph?

PR #10115, opened 2023-07-17
Avoid lambda flame-graph split (improved version of #10110)

sbordet:


jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannel.java
```diff
@@ -549,11 +544,7 @@ public boolean handle()
                                 break;
                             }
 
-                            dispatch(DispatcherType.ERROR, () ->
-                            {
-                                errorHandler.handle(null, _request, _request, _response);
-                                _request.setHandled(true);
-                            });
+                            dispatch(DispatcherType.ERROR, new ErrorDispatchable(errorHandler));
```
joakime, line 547: This ErrorDispatcher can't be a final field due to the errorHandler being able to be set later?
What about just having the ErrorDispatcher be a normal field?

joakime (changes requested):
Don't we have the same problem with multiple instances of the ErrorDispatcher in the flamegraph with this implementation?

joakime (approved):
Reason for `ErrorDispatchable` the way it is was explained to me.
I'm good with this PR now.

PR #10116, opened 2023-07-17
Fixes #10086 - Revisiting ProxyConfiguration.getProxies()

### LINKED ISSUE #10086: Revisiting ProxyConfiguration.getProxies()
@sbordet @joakime Revisiting https://github.com/eclipse/jetty.project/issues/8723, I'm wondering whether it was wise/necessary to deprecate `getProxies()`.

Maybe we could keep the `getProxies()` method but specify that it returns an unmodifiable `List` or `Collection` instead? This would restore the users' ability to look up the proxy list without exposing any implementation details. Users would still be forced to use the `addProxy()`, `removeProxy()` methods for mutation.

sbordet:
Removed "for removal" deprecation, added clarifying javadocs.

### DIFF
--- a/jetty-client/src/main/java/org/eclipse/jetty/client/ProxyConfiguration.java
+++ b/jetty-client/src/main/java/org/eclipse/jetty/client/ProxyConfiguration.java
@@ -41,10 +41,18 @@ public class ProxyConfiguration
     private final List<Proxy> proxies = new BlockingArrayQueue<>();
 
     /**
+     * <p>Returns the list of proxies added to this configuration.</p>
+     * <p>This method is deprecated because it has historically been
+     * used to add/remove proxies directly to/from the returned list,
+     * but this has now been replaced by {@link #addProxy(Proxy)} and
+     * {@link #removeProxy(Proxy)} instead.</p>
+     * <p>In the future, this method will return an unmodifiable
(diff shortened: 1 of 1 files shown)


jetty-client/src/main/java/org/eclipse/jetty/client/ProxyConfiguration.java
```diff
@@ -41,10 +41,18 @@ public class ProxyConfiguration
     private final List<Proxy> proxies = new BlockingArrayQueue<>();
 
     /**
+     * <p>Returns the list of proxies added to this configuration.</p>
+     * <p>This method is deprecated because it has historically been
+     * used to add/remove proxies directly to/from the returned list,
+     * but this has now been replaced by {@link #addProxy(Proxy)} and
+     * {@link #removeProxy(Proxy)} instead.</p>
+     * <p>In the future, this method will return an unmodifiable
+     * list of proxies just for querying purposes.</p>
```
joakime, line 50: If the future version will return an unmodifiable list, why is this marked for removal?

joakime (changes requested):
Confused state of the javadoc about the future of this method, and then marking it for removal.

joakime (approved):
Yes, I see the change the right way now.
Approved.

